PR #12, opened 2020-08-26
build improvements

bertwesarg:
Using a non-system HWLOC was only possible when setting `CFLAGS` and `LIBS` on the `configure` line. Using `pkg-config` to get the flags from the `.pc` file makes this more convenient. Utilizing `libtool` for the linking step also eases this, as `libtool` uses `rpath` for libraries not in system paths.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Install Dependencies
-      run: sudo apt-get install libncursesw5-dev
+      run: sudo apt-get install pkg-config libncursesw5-dev
     - name: Bootstrap
       run: ./autogen.sh
     - name: Configure
@@ -30,7 +30,7 @@ jobs:
         sudo add-apt-repository 'deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-10 main' -y
         sudo apt-get update -q
     - name: Install Dependencies
-      run: sudo apt-get install clang-10 libncursesw5-dev
+      run: sudo apt-get install clang-10 pkg-config libncursesw5-dev
     - name: Bootstrap
       run: ./autogen.sh
     - name: Configure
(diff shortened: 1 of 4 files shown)


natoscott:
@bertwesarg re "always require `pkg-config`" - AIUI there is no pkg-config in the Apple developer tools on Mac OS X (ignoring brew and other add-ons).

Hisham was working on an approach for using pkg-config conditionally - have a look at the pkgconfig-on-linux branch.

bertwesarg:
> @bertwesarg re "always require `pkg-config`" - AIUI there is no pkg-config in the Apple developer tools on Mac OS X (ignoring brew and other add-ons).

I have the impression, that either the default autotools are ancient or there are no autotools in the Apple developer tools at all. So what is the point to restrict htop here, if the mac developers need to install a recent autotools anyway via homebrew, then they should also be able to install pkg-config via homebrew

> 
> Hisham was working on an approach for using pkg-config conditionally - have a look at the pkgconfig-on-linux branch.

thanks, this looks like it was already merged

natoscott:
> I have the impression, that either the default autotools are ancient or there are no autotools in the Apple developer tools at all.

Released htop sources are self-contained in terms of autotools - i.e. there's no requirement for a local (eg brew-based) install.

bertwesarg:
> > I have the impression, that either the default autotools are ancient or there are no autotools in the Apple developer tools at all.
> 
> Released htop sources are self-contained in terms of autotools - i.e. there's no requirement for a local (eg brew-based) install.

I'm well aware of that fact and this also holds for `pkg-config`. If it is not available on the build machine configure should not use it. And the previous state also assumes this, as it was only about creating a release, not building a release.

natoscott:
> > Released htop sources are self-contained in terms of autotools - i.e. there's no requirement for a local (eg brew-based) install.
> 
> I'm well aware of that fact and this also holds for `pkg-config`.

Ah, OK.

> If it is not available on the build machine configure should not use it.

Are we sure about this?  Can we confirm that?  Keen to not break the non-Linux platform builds here - thanks.

BTW, there's conflicts in this PR compared to HEAD now.

bertwesarg:
> > If it is not available on the build machine configure should not use it.
> 
> Are we sure about this? Can we confirm that? Keen to not break the non-Linux platform builds here - thanks.

I could only test with `PKG_CONFIG=false` on a macOS, because that already had the Homebrew `pkg-config` formula installed. And it did not failed:

```
checking pkg-config is at least version 0.9.0... no
```

bertwesarg:
The package I used was created on a Linux system.

natoscott:
Closing this one for now - personally, I think the general idea of a non-default pkg-config use is OK, but this PR has become quite dated now and there's very little demand for the change.  Ideally we'd solve the ncurses[w] pkg-config issue in the same way, at the same time.

PR #23, opened 2020-08-28
Coverity fixes

cgzones:
There is one strange issue left:

```
399static Htop_Reaction actionStrace(State* st) {
400   Process* p = (Process*) Panel_getSelected(st->panel);
   1. Condition !p, taking false branch.
401   if (!p) return HTOP_OK;
   2. alloc_fn: Storage is returned from allocation function TraceScreen_new. [show details]
   3. var_assign: Assigning: ts = storage returned from TraceScreen_new(p).
402   TraceScreen* ts = TraceScreen_new(p);
   4. noescape: Resource ts is not freed or pointed-to in TraceScreen_forkTracer. [show details]
403   bool ok = TraceScreen_forkTracer(ts);
   5. Condition ok, taking false branch.
404   if (ok) {
405      InfoScreen_run((InfoScreen*)ts);
406   }
   6. noescape: There is at least one path in TraceScreen_delete on which resource ts is not freed or pointed-to.
407   TraceScreen_delete((Object*)ts);
408   clear();
409   CRT_enableDelay();
   CID 213660 (#1 of 1): Resource leak (RESOURCE_LEAK)7. leaked_storage: Variable ts going out of scope leaks the storage it points to.
410   return HTOP_REFRESH | HTOP_REDRAW_BAR;
411}
```

But I do not see any path in `TraceScreen_delete()` that does not free `ts`:

```
void TraceScreen_delete(Object* cast) {
   TraceScreen* this = (TraceScreen*) cast;
   if (this->child > 0) {
      kill(this->child, SIGTERM);
      waitpid(this->child, NULL, 0);
      fclose(this->strace);
   }
   CRT_enableDelay();
   free(InfoScreen_done((InfoScreen*)cast));
}
```

### DIFF
--- a/InfoScreen.c
+++ b/InfoScreen.c
@@ -80,7 +80,7 @@ void InfoScreen_drawTitled(InfoScreen* this, const char* fmt, ...) {
    va_start(ap, fmt);
    attrset(CRT_colors[METER_TEXT]);
    mvhline(0, 0, ' ', COLS);
-   wmove(stdscr, 0, 0);
+   (void) wmove(stdscr, 0, 0);
    vw_printw(stdscr, fmt, ap);
    attrset(CRT_colors[DEFAULT_COLOR]);
    this->display->needsRedraw = true;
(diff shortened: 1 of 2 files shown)


natoscott:
> But I do not see any path in `TraceScreen_delete()` that does not free `ts`:

@cgzones agreed - perhaps covscan is getting confused by the return type of InfoScreen_done being an InfoScreen pointer? (it doesn't matter though AFAICS - free should still do the right thing in this case)

PR #30, opened 2020-08-29
add toggle for vim mode in options

Ckath:
seems weird the only way to enable vim mode is changing `vim_mode` in the htoprc file, this just adds a toggle like enabling the mouse in display optiions


7415963987456321:
On a related note, is there generally any interest in expanding the vim-like functionality in htop? A number of bindings come to mind, for instance I use <C-b> and <C-u> a lot in vim for scrolling more than one line, would those kind of additions be welcomed to htop?

Ckath:
im not sure but im not super happy about how its implemented here with the whole "change keys into old bindings" type of deal. I could add whatever htop currently offers as binds but as vim binds if thats desired

PR #37, opened 2020-08-29
ignore vim keys in prompts

### LINKED ISSUE #34: Vim keys break search
When searching for a process with vim keys enabled, you cannot use HJKL keys. It instead closes the search box and starts navigating instead.

Ckath:
this makes it so vimmode(rewriting hjkl to arrow keys) will be ignored when inside a prompt, such as search or filter. to detect this im simply looking at when the functionbar only has 3 items (esc, cancel, prompt) as I couldnt find a more straight forward way to find out.

this would resolve #34

### DIFF
--- a/ScreenManager.c
+++ b/ScreenManager.c
@@ -190,7 +190,7 @@ void ScreenManager_run(ScreenManager* this, Panel** lastFocus, int* lastKey) {
       set_escdelay(25);
       ch = getch();
 
-      if (this->settings->vimMode) {
+      if (this->settings->vimMode && panelFocus->currentBar->size != 3) {
          switch (ch) {
             case 'h': ch = KEY_LEFT; break;
             case 'j': ch = KEY_DOWN; break;
--- a/ScreenManager.c
+++ b/ScreenManager.c
@@ -190,6 +190,7 @@ void ScreenManager_run(ScreenManager* this, Panel** lastFocus, int* lastKey) {
       set_escdelay(25);
       ch = getch();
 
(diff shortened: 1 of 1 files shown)


cgzones:
One could maybe add the global state into the `struct ScreenManager` via `const struct State *` and initialize it in `ScreenManager_new()` and the check via `((MainPanel*)st->panel)->inc->active == NULL` ?
Don't know if it's worth the trouble.

I would however prefer a comment before the check `if (this->settings->vimMode && panelFocus->currentBar->size != 3)`

Ckath:
a comment to explain why this is here seems like a good idea, as its very vague.

>Don't know if it's worth the trouble.
if it really clears up the situation it might be, I dont see that global state keeping helping a lot there, especially with it only being used for this one edgecase. in my mind it'd make the most sense to keep track of if something is a prompt or not in the function bars, and then check that one bool.

cgzones (approved):

BenBE:
While the comment explains the check I think it should be cleaned up in the long run as this is error prone (in case we were to add a second bar at the bottom, or whatever). The code as is is not conservative in its effects in regards to changes of the underlying assumptions. Okay for short term; prefer cleanup longterm.

Ckath:
to clear up confusion, maybe best to close this and the other vim_mode related prs?

I've discussed some options and will be working towards a improved vim_mode implementation that wont have any of these bugs caused by the way the old one was implemented. when that time comes any discussion related to that can just be moved to one (draft) pr

PR #39, opened 2020-08-29
never use glyphs for checkmarks

### LINKED ISSUE #29: bring back the regular checkmarks
pls
its great and all this is still maintained, but I dont enjoy things becoming less readable. you could argue here "use a better font" but htop is used in a lot of minimal installs that might be lacking all and any fancy glyphs, which is why the old method of checkmarks is used in nearly every cli program

Ckath:
dont really gain anything from ever using these, it seems this part of the source isnt great for getting options from settings as it initializes once at startup, but the more I thought about it the more it made me not care about ever seeing these glyphs for checkmarks.

this would resolve #29, see the issue for more problems with ever using these

### DIFF
--- a/CRT.c
+++ b/CRT.c
@@ -194,12 +194,6 @@ const char *CRT_treeStrUtf8[TREE_STR_COUNT] = {
    "\xe2\x94\x80", // TREE_STR_SHUT ─
 };
 
-const char *CRT_checkStrUtf8[CHECK_STR_COUNT] = {
-   "\xe2\x98\x90", // CHECK_STR_NONE    ☐
-   "\xe2\x98\x92", // CHECK_STR_PARTIAL ☒
-   "\xe2\x98\x91", // CHECK_STR_FULL    ☑
-};
-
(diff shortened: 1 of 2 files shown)


bertwesarg:
this is by far not complete regarding the reversal, as I still have the original commit which introduced the box drawings, I will open a new PR soonish

bertwesarg:
> this is by far not complete regarding the reversal, as I still have the original commit which introduced the box drawings, I will open a new PR soonish

here is the original commit https://github.com/bertwesarg/htop/commit/5d5913d355b3a9f03da589b3542b8f55467b4ed6

natoscott:
> this is by far not complete regarding the reversal, as I still have the original commit which introduced the box drawings, I will open a new PR soonish

@bertwesarg thanks Bert!

Ckath:
I agree reverting it completely is the better option here, I just undid what I thought was the full change over to glyphs. I'll close this in favor of the to-be-submitted PR of @bertwesarg

PR #48, opened 2020-08-31
align single digit cpu id to center

eworm-de:


bertwesarg:
Do you have any motivation for this to share? AFAIC that would be the first center alignment in htop.

eworm-de:
With just a single digit cpu id the number is closer to what ever is left of it instead of the meter it belongs to...
For me this has always looked strange.

Not a big deal if this is not merged, but I wanted to propose at least.
Let me know if you have an idea to solve this even better.

bertwesarg:
So with >99 CPUs it should now look this way:
```
 1 […]
 2 […]
99 […]
100[…]
```
Why not make it right align to begin with?
```
  1[…]
  2[…]
 99[…]
100[…]
```

eworm-de:
Fine with me if you prefer it to be aligned to the right.

cgzones:
I have no preference whether to align the cpu id to center or to right.

fasterit:
+1 for right aligned

Rationale:
- Consistent with other labels
- Aligned with Mem / Swp and friends

cgzones:
Right alignment it is.

PR #50, opened 2020-08-31
add option (-U, --no-unicode) to disable unicode at runtime

eworm-de:


CRT.c
```diff
@@ -763,12 +763,15 @@ void CRT_init(int delay, int colorScheme) {
    setlocale(LC_CTYPE, "");
 
 #ifdef HAVE_LIBNCURSESW
-   if(strcmp(nl_langinfo(CODESET), "UTF-8") == 0) {
+   if (allowUnicode == true &&
```
cgzones, line 766: `allowUnicode == true` seems a bit redundant

cgzones (commented):
LGTM

CRT.c
eworm-de: I can remove that if you prefer...

Ckath:
doesnt seem needed anymore after #54, which I think would be the preferable solution

bertwesarg:
> doesnt seem needed anymore after #54, which I think would be the preferable solution

I think they match up perfectly. Disabling Unicode, even if the terminal is capable, looks like a valid case.

eworm-de:
Let me explain my use case: The graph meters use braille characters to display. This works fine if braille characters are available (X terminal), but fails if not: Only squares are displayed then. I use this to switch to ASCII mode in virtual (non-X) terminal. UTF-8 is supported there, just glyphs are missing.

natoscott (changes requested):
This needs a man page update I think...?  Thanks.

eworm-de:
Ah, right... Will add that.

eworm-de:
Rebased and added the man page change.

natoscott (commented):
Thanks - another small thing, it looks like all the #ifdef conditionals in htop.c could be removed around the new code?  The default behaviour without HAVE_LIBNCURSESW is to not 'allowUnicode', right?  - it seems that code can be simplified further, always allowing the -U/--no-unicode option to be entered by the user, no matter how htop is built.

eworm-de:
Probably you are right. :)
Updated (and rebased) again.

PR #56, opened 2020-09-01
align cpu id to right

eworm-de:


eworm-de:
Take this as an alternative to #48

cgzones:
Thanks, applied.

See discussion in #48.

PR #42, opened 2020-08-30
Improving Command display/sort

gnarendran:
Addresses https://github.com/htop-dev/htop/issues/40; Rebased from https://github.com/hishamhm/htop/pull/809; Originally discussed in https://github.com/hishamhm/htop/issues/801

With inputs from @Explorer09, @BenBE, @marxin

1. On Linux a process is able to modify its own command name (/proc/pid/comm) say using prctl, and its own command line (/proc/pid/cmdline) say by writing into its arguments vector. But /proc/[pid]/exe is set by the kernel and provides the definitive location of the executable image. While the Command column currently displays cmdline, for the above reason it is desirable to display the executable image path and command name, along with cmdline, merging them where possible.

- In this pull request, two new optional columns are added:
i) "Comm" - displays /proc/[pid]/comm if readable. It may not be readable for zombies.
ii) "Exe" - displays 15 characters of the basename of /proc/[pid]/exe if readable. htop is able to read the /proc/[pid]/exe of ALL processes only when htop is either run with root privilleges (done with care) or given the capability CAP_SYS_PTRACE (`sudo setcap 'cap_sys_ptrace=ep' /usr/bin/htop`).

- A new option is added: "Merge exe, comm and cmdline in Command", toggled by the key 'm'
If this option is unset (default for backwards compatibility), Command column displays cmdline as usual. If this option is set, exe, comm and cmdline are merged in the Command column, which is to be interpreted as follows:
	i) If no token is colorized, it implies htop didn't have permission to read /proc/[pid]/exe of the process, and htop has fallen back to displaying only cmdline.

	ii) If a token (which may have embedded spaces) is colorized, upto 15 bytes of it is understood to be comm. This implies htop was able to read the process' /proc/[pid]/exe and /proc/[pid]/comm. During display, htop first tries to find/merge comm in exe's basename; If that fails and also if the new option 'Try to find comm in cmdline' is set (default), htop tries to find/merge comm in cmdline (this may mis-identify a string in cmdline in very rare cases, say if comm or cmdline had been unsuitably modified).
		a) If comm was not merged into either exe or cmdline, three fields are displayed (with "│" as the field separator), the first being exe, the second comm, and the last cmdline.
		b) If comm was merged into exe or cmdline: If exe could not be merged with cmdline, two fields are displayed, the first being exe and the last cmdline;
		c) Otherwise, exe and cmdline are merged into a single field.

[...]

PR #67, opened 2020-09-02
use separate window/screen to show long commands in multiple lines

ryenus:
With this change the command of the highlighted process can be shown in 
a separate window, and wrapped into multiple lines.

Here's a screenshot showing a Chrome process with a long command:

image

### DIFF
--- a/Action.c
+++ b/Action.c
@@ -13,6 +13,7 @@ in the source distribution for its full text.
 #include "CategoriesPanel.h"
 #include "CRT.h"
 #include "EnvScreen.h"
+#include "CommandScreen.h"
 #include "MainPanel.h"
 #include "OpenFilesScreen.h"
 #include "Process.h"
@@ -422,8 +423,8 @@ static const struct { const char* key; const char* info; } helpRight[] = {
    { .key = "      e: ", .info = "show process environment" },
    { .key = "      i: ", .info = "set IO priority" },
    { .key = "      l: ", .info = "list open files with lsof" },
+   { .key = "      M: ", .info = "show process command in multiple lines" },
    { .key = "      s: ", .info = "trace syscalls with strace" },
-   { .key = "         ", .info = "" },
    { .key = " F2 C S: ", .info = "setup" },
    { .key = "   F1 h: ", .info = "show this help screen" },
    { .key = "  F10 q: ", .info = "quit" },
@@ -530,6 +531,16 @@ static Htop_Reaction actionShowEnvScreen(State* st) {
    return HTOP_REFRESH | HTOP_REDRAW_BAR;
 }
 
+static Htop_Reaction actionShowCommandScreen(State* st) {
+   Process* p = (Process*) Panel_getSelected(st->panel);
+   if (!p) return HTOP_OK;
+   CommandScreen* cmdScr = CommandScreen_new(p);
+   InfoScreen_run((InfoScreen*)cmdScr);
+   CommandScreen_delete((Object*)cmdScr);
+   clear();
+   CRT_enableDelay();
+   return HTOP_REFRESH | HTOP_REDRAW_BAR;
+}
 
 void Action_setBindings(Htop_Action* keys) {
    keys[KEY_RESIZE] = actionResize;
@@ -584,4 +595,5 @@ void Action_setBindings(Htop_Action* keys) {
    keys['U'] = actionUntagAll;
    keys['c'] = actionTagAllChildren;
    keys['e'] = actionShowEnvScreen;
+   keys['M'] = actionShowCommandScreen;
 }
--- a/CommandScreen.c
+++ b/CommandScreen.c
@@ -1,0 +1,74 @@
+#include "CommandScreen.h"
+
+#include "config.h"
+#include "CRT.h"
+#include "IncSet.h"
+#include "ListItem.h"
+#include "Platform.h"
+#include "StringUtils.h"
+
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+
+static void CommandScreen_addLine(InfoScreen* this, char* line, const char* p, int line_offset, int len) {
+   memcpy(line, p - line_offset, len);
+   line[len] = '\0';
+   InfoScreen_addLine(this, line);
+}
+
+static void CommandScreen_scan(InfoScreen* this) {
+   Panel* panel = this->display;
+   int idx = MAXIMUM(Panel_getSelectedIndex(panel), 0);
+
+   Panel_prune(panel);
+
+   const char* p = this->process->comm;
+   char* line = xMalloc(COLS + 1);
+   int line_offset = 0, last_spc = -1, len;
+   for (; *p != '\0'; p++, line_offset++) {
+      if (*p == ' ') last_spc = line_offset;
+
+      if (line_offset == COLS) {
+         len = (last_spc == -1) ? line_offset : last_spc;
+         CommandScreen_addLine(this, line, p, line_offset, len);
+         line_offset -= len;
+         last_spc = -1;
+      }
+   }
+
+   if (line_offset > 0) CommandScreen_addLine(this, line, p, line_offset, line_offset);
+
+   free(line);
+   Panel_setSelected(panel, idx);
+}
+
+static void CommandScreen_draw(InfoScreen* this) {
+   char* title = xMalloc(COLS + 1);
+   int len = snprintf(title, COLS + 1, "Command of process %d - %s", this->process->pid, this->process->comm);
+   if (len > COLS) {
+      memset(&title[COLS - 3], '.', 3);
+   }
+
+   InfoScreen_drawTitled(this, "%s", title);
+   free(title);
+}
+
+InfoScreenClass CommandScreen_class = {
+   .super = {
+      .extends = Class(Object),
+      .delete = CommandScreen_delete
+   },
+   .scan = CommandScreen_scan,
+   .draw = CommandScreen_draw
+};
+
+CommandScreen* CommandScreen_new(Process* process) {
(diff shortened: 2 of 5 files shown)


Action.c
```diff
@@ -454,8 +455,8 @@ static const struct { const char* key; const char* info; } helpRight[] = {
    { .key = "      e: ", .info = "show process environment" },
    { .key = "      i: ", .info = "set IO priority" },
    { .key = "      l: ", .info = "list open files with lsof" },
+   { .key = "      m: ", .info = "show process command in multiple lines" },
```
cgzones, line 458: Utilizing the m key conflicts with #42

CommandScreen.c
```diff
@@ -0,0 +1,79 @@
...
+   return (CommandScreen*) InfoScreen_init(&this->super, process, NULL, LINES-3, " ");
+}
+
+void CommandScreen_delete(Object* this) {
+   free(InfoScreen_done((InfoScreen*)this));
+}
+
+void CommandScreen_draw(InfoScreen* this) {
+   char* title = xMalloc(COLS + 1);
+   snprintf(title, COLS - 4, "Command of process %d - %s", this->process->pid, this->process->comm);
```
cgzones, line 43: Why allocate `COLS + 1` and then use only `COLS - 4`?

cgzones (changes requested):

Action.c
ryenus: Maybe `M` then?

CommandScreen.c
ryenus: Right, I improved the title truncation, now it only show trailing dots when needed.

ryenus:
I just realized that the tab char (`\t`) is displayed as a question mark (`?`), so we only wrap by space now:

image

ryenus:
@cgzones rebase done as well.

CommandScreen.c
```diff
@@ -0,0 +1,83 @@
...
+#include "IncSet.h"
+#include "ListItem.h"
+#include "Platform.h"
+#include "StringUtils.h"
+
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+/*{
```
cgzones, line 14: Please drop this part; auto-generated headers have been removed.

CommandScreen.c
```diff
@@ -0,0 +1,83 @@
...
+   .super = {
+      .extends = Class(Object),
+      .delete = CommandScreen_delete
+   },
+   .scan = CommandScreen_scan,
+   .draw = CommandScreen_draw
+};
+
+CommandScreen* CommandScreen_new(Process* process) {
+   CommandScreen* this = xMalloc(sizeof(CommandScreen));
```
cgzones, line 32: `AllocThis(CommandScreen)` (includes malloc + Object_setClass)

CommandScreen.c
```diff
@@ -0,0 +1,83 @@
...
+   free(line);
+}
+
+void CommandScreen_scan(InfoScreen* this) {
+   Panel* panel = this->display;
+   int idx = MAXIMUM(Panel_getSelectedIndex(panel), 0);
+
+   Panel_prune(panel);
+
+   char *p = this->process->comm;
```
cgzones, line 66: maybe `const char *` ?

CommandScreen.h
```diff
@@ -0,0 +1,24 @@
+/* Do not edit this file. It was automatically generated. */
```
cgzones, line 1: Please drop this line; auto-generated headers have been removed.

CommandScreen.h
```diff
@@ -0,0 +1,24 @@
...
+
+typedef struct CommandScreen_ {
+   InfoScreen super;
+} CommandScreen;
+
+extern InfoScreenClass CommandScreen_class;
+
+CommandScreen* CommandScreen_new(Process* process);
+
+void CommandScreen_delete(Object* this);
```
cgzones, line 16: Do these four functions need to be global ones, can't they be static ones?

cgzones (changes requested):

CommandScreen.c
ryenus: done

CommandScreen.c
ryenus: done

CommandScreen.c
ryenus: done

CommandScreen.h
ryenus: done

CommandScreen.h
ryenus: I guess that's possible? I just followed the convention in OpenFilesScreen.h and EnvScreen.h, though.

ryenus: OK, I gave it a try, and found that I also have to:
1. remove the static functions from CommandScreen.h due to unused-function warning
2. subsequently move up the static functions in CommandScreen.c to ensure they're defined before use.

See the yet to be merged commits here: https://github.com/ryenus/htop/compare/4b02f0d...static-fn

BTW, `CommandScreen_delete` can't be static as it's called from `Action.c`. 

@cgzones How about that? Should we proceed from here?

cgzones: Sorry, I missed that `CommandScreen_delete()` is getting called from `Action.c`.
The static-fn branch looks good to me.
Please squash all the commits into a single one.

ryenus: done

CommandScreen.c
```diff
@@ -0,0 +1,74 @@
...
+#include "StringUtils.h"
+
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+
+static void CommandScreen_addLine(InfoScreen* this, char* line, const char* p, int line_offset, int len) {
+   memcpy(line, p - line_offset, len);
+   line[len] = '\0';
```
BenBE, line 17: What about `strncpy` instead?

cgzones (approved):

cgzones:
Thanks, applied.

CommandScreen.c
```diff
@@ -0,0 +1,74 @@
...
+   for (; *p != '\0'; p++, line_offset++) {
+      if (*p == ' ') last_spc = line_offset;
+
+      if (line_offset == COLS) {
+         len = (last_spc == -1) ? line_offset : last_spc;
+         CommandScreen_addLine(this, line, p, line_offset, len);
+         line_offset -= len;
+         last_spc = -1;
+      }
+   }
```
BenBE, line 39: My head hurts trying to read this … I'd prefer if `CommandScreen_addLine` didn't input the end of the string it should add, but its beginning. With some clever restructuring this even removes the need for `CommandScreen_addLine`:

```c
   char* line = xCalloc(COLS + 1, sizeof(char));
   int line_offset = 0
   int last_space = -1;
   for (const char* p = this->process->comm; *p != 0; p++) {
      if (*p == ' ') {
         last_spc = line_offset;
      }

      line[line_offset++] = *p;

      if (COLS == line_offset) {
         if(-1 != last_space) {
            line[last_space] = 0;
            line_offset = last_space;
         }
         InfoScreen_addLine(this, line);
         if(COLS != line_offset) {
            memmov(line, line + line_offset, COLS - line_offset);
         }
         line_offset = COLS - line_offset;
         last_space = -1;
      }
   }
   if(0 != line_offset) {
      line[line_offset] = 0;
      InfoScreen_addLine(this, line);
   }
   free(line);
```
(Implementation untested, but should do)

ryenus: Hey @BenBE, thanks! And yeah, it's definitely an option to [fill the line while
scanning][1]. Just make sure to correctly [carrier over][2] the already scanned
character(s) to the next line.

On the other hand by doing that, the possible optimization in **memcpy**, namely
filling the line by words no longer applies. But I'm not sure if that's really 
a gain or lose before further checking the detail, or that doesn't even matter?

BTW, in your comment you mentioned the idea to skip leading whitespace(s) when
starting next line, I'd prefer not to do that, because spaces in a quoted string
ought to be preserved. And by not doing that we can be sure that the original
command can be reconstructed by just removing the line endings. Anyway that's not 
really a big deal.


[1]: https://github.com/ryenus/htop/commit/7f955db52fb602303711714df0a3387d8fc0e941#diff-5e605b3d4fdff9c400a8330b918626feR20-R43
[2]: https://github.com/ryenus/htop/commit/7f955db52fb602303711714df0a3387d8fc0e941#diff-5e605b3d4fdff9c400a8330b918626feR34

ryenus: Hey @BenBE, thanks! And yeah, it's definitely an option to [fill the line while
scanning][1]. Just make sure to correctly [carrier over][2] the already scanned
character(s) to the next line.

On the other hand by doing that, the possible optimization in **memcpy**, at 
least filling the line by words no longer applies. But I'm not sure if that's 
really a gain or lose before further checking the detail, or that doesn't even 
matter?

BTW, in your comment you mentioned the idea to skip leading whitespace(s) when
starting next line, I'd prefer not to do that, because spaces in a quoted string
ought to be preserved. And by not doing that we can be sure that the original
command can be reconstructed by just removing the line endings. Anyway that's not 
really a big deal.


[1]: https://github.com/ryenus/htop/commit/7f955db52fb602303711714df0a3387d8fc0e941#diff-5e605b3d4fdff9c400a8330b918626feR20-R43
[2]: https://github.com/ryenus/htop/commit/7f955db52fb602303711714df0a3387d8fc0e941#diff-5e605b3d4fdff9c400a8330b918626feR34

ryenus: OK, I [gave it a try][1] and measured the time, for a command with 7k+ 
characters, it's about `30` vs `28` (ms) on average before and after in
the best case. And the copy-on-scan approach won with a little margin.

[1]: https://github.com/ryenus/htop/commit/d96a931de437ea027d51af393523b21d4b2bea25

PR #70, opened 2020-09-02
Axe automated header generation.

zevweiss:
Reasoning:
 - implementation was unsound -- broke down when I added a fairly
   basic macro definition expanding to a struct initializer in a *.c
   file.

 - made it way too easy (e.g. via otherwise totally innocuous git
   commands) to end up with timestamps such that it always ran
   MakeHeader.py but never used its output, leading to overbuild noise
   when running what should be a null 'make'.

 - but mostly: it's just an awkward way of dealing with C code.

[Addresses #9.]


bertwesarg:
Please remove `extern` from all pure function declarations in the headers. They are superfluous.

Makefile.am
```diff
@@ -58,7 +58,7 @@ all_platform_headers += $(linux_platform_headers)
 
 if HTOP_LINUX
 AM_LDFLAGS += -rdynamic
-myhtopplatsources = linux/Platform.c linux/IOPriorityPanel.c linux/IOPriority.c \
+myhtopplatsources = linux/Platform.c linux/IOPriorityPanel.c \
```
bertwesarg, line 61: Why was `linux/IOPriority.c` removed?

zevweiss: Because aside from being a source from which the corresponding header was generated, it included nothing of actual substance -- an enum declaration, a few macros, and a typedef (zero variable or function definitions).

zevweiss:
Agreed re: `extern`; bulk removal performed in 822154a.

Makefile.am
bertwesarg: sorry, missed, that the `.c` file was removed in reallity too.

bertwesarg:
The extra targets in `Makefile.am` should now be removed too:

```
.PHONY: htop-headers clean-htop-headers

htop-headers: $(myhtopheaders) $(all_platform_headers)

clean-htop-headers:
	-rm -f $(myhtopheaders) $(all_platform_headers)
```

Maybe the `all_platform_headers` variable too

zevweiss:
Ah, good point -- and also `SUFFIXES` and `BUILT_SOURCES` I believe.

zevweiss:
Rebased to resolve merge conflicts; also added a tiny commit to remove a duplicate declaration introduced in #86.

PR #93, opened 2020-09-04
apply htop temperature patch

### LINKED ISSUE #49: Feature request: add temperature support patch?
Hi

It is possible to add the temperature patch, so the AUR package [htop-temperature](https://aur.archlinux.org/packages/htop-temperature/) won't be needed anymore?

The patch itself is available [here](https://aur.archlinux.org/cgit/aur.git/tree/htop-temperature.patch?h=htop-temperature).

Thank you very much!

DX37:
Closes #49.

Example of TemperatureMeter using from comment below:
изображение
изображение
изображение
изображение

Also should be working on Raspberry Pi and maybe other ARM devices, which report `cpu_thermal` zone to `sensors` command (tested on Arch Linux ARM).

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -117,6 +117,11 @@ myhtopheaders = \
 	Vector.h \
 	XAlloc.h
 
+if HAVE_LIBSENSORS
+myhtopsources += TemperatureMeter.c
+myhtopheaders += TemperatureMeter.h
+endif
+
 # Linux
 # -----
 
--- a/TemperatureMeter.c
+++ b/TemperatureMeter.c
@@ -1,0 +1,170 @@
+/*
+htop - TemperatureMeter.c
+(C) 2013 Ralf Stemmer
+(C) 2014 Blair Bonnett
+(C) 2020 Maxim Kurnosenko
+Released under the GNU GPL, see the COPYING file
+in the source distribution for its full text.
+*/
+
+#include "TemperatureMeter.h"
+
+#include "ProcessList.h"
+#include "CRT.h"
+#include "StringUtils.h"
+#include "XAlloc.h"
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <limits.h>
+#include <errno.h>
+#include <sensors/sensors.h>
+
+typedef struct
+{
+   int size;
+   int* current_temp;
+   int* max_temp;
+} Dynarr;
+
+static int temp_converter(const Meter* this, double temp) {
+   if (this->pl->settings->fahrenheitTemperature)
+      return temp * 9 / 5 + 32;
+   else
+      return temp;
+}
+
+static Dynarr* get_cpu_temp(const Meter* this) {
+   int core = 0, n = 0, count = 0;
+   Dynarr* result = xMalloc(sizeof(Dynarr));
+
+   for (const sensors_chip_name *chip = sensors_get_detected_chips(NULL, &n); chip; chip = sensors_get_detected_chips(NULL, &n)) {
+      char buffer[64];
+      if (sensors_snprintf_chip_name(buffer, sizeof(buffer), chip) < 0)
+         continue;
+
+      // cpu_thermal is for Raspberry Pi sensor support
+      if (!String_startsWith(buffer, "coretemp") && !String_startsWith(buffer, "cpu_thermal"))
+         continue;
+
+      int m = 0;
+      for (const sensors_feature *feature = sensors_get_features(chip, &m); feature; feature = sensors_get_features(chip, &m)) {
+         if (feature->type == SENSORS_FEATURE_TEMP)
+            count++;
+      }
+
+      result->current_temp = xCalloc(count, sizeof(int));
+      result->max_temp = xCalloc(count, sizeof(int));
+      result->size = count;
+
+      m = 0;
+      for (const sensors_feature *feature = sensors_get_features(chip, &m); feature; feature = sensors_get_features(chip, &m)) {
+         if (feature->type != SENSORS_FEATURE_TEMP)
+            continue;
+
+         double local_temp;
+         const sensors_subfeature *sub_feature;
+
+         sub_feature = sensors_get_subfeature(chip, feature, SENSORS_SUBFEATURE_TEMP_INPUT);
+         if (sub_feature) {
+            if (sensors_get_value(chip, sub_feature->number, &local_temp) == 0)
+               result->current_temp[core] = temp_converter(this, local_temp);
+         }
+
+         sub_feature = sensors_get_subfeature(chip, feature, SENSORS_SUBFEATURE_TEMP_MAX);
+         if (sub_feature) {
+            if (sensors_get_value(chip, sub_feature->number, &local_temp) == 0)
+               result->max_temp[core] = temp_converter(this, local_temp);
+         }
+         ++core;
+      }
+   }
+
+   return result;
+}
+
+int TemperatureMeter_attributes[] = {
+   TEMPERATURE_COOL,
+   TEMPERATURE_MEDIUM,
+   TEMPERATURE_HOT,
+};
+
+// Shows only package temperature in bars meter mode
+static void TemperatureMeter_updateValues(Meter* this, char* buffer, int len) {
+   Dynarr* core_temperature = get_cpu_temp(this);
+   double temperature = core_temperature->current_temp[0];
+
+   if (temperature < 0) {
+      xSnprintf(buffer, len, "N/A");
+      free(core_temperature);
+      return;
+   }
+
+   this->total = core_temperature->max_temp[0];
+   this->values[0] = temperature;
+
+   if (this->pl->settings->fahrenheitTemperature)
+      xSnprintf(buffer, len, "%d%sF", (int) temperature, CRT_degreeSign());
+   else
+      xSnprintf(buffer, len, "%d%sC", (int) temperature, CRT_degreeSign());
+   free(core_temperature);
+}
+
+// Shows every physical core temperature in text meter mode
+static void TemperatureMeter_display(const Object* cast, RichString* out) {
+   const Meter* this          = (const Meter*) cast;
+   Dynarr* cores_temperatures = get_cpu_temp(this);
+   int temp_medium            = temp_converter(this, 55);
+   int temp_hot               = temp_converter(this, 75);
+
+   if (cores_temperatures->current_temp[0] < 0 || cores_temperatures->size == 0) {
+      RichString_append(out, CRT_colors[METER_VALUE], "n/a");
+      free(cores_temperatures);
+      return;
+   }
+
+   int core_count_begin = cores_temperatures->size == 1 ? 0 : 1;
+
+   for (int core = core_count_begin; core < cores_temperatures->size; ++core) {
+      // convert double to integer
+      int temperature = (int) cores_temperatures->current_temp[core];
+
+      // choose the color for the temperature
+      int tempColor;
+      if (temperature < temp_medium)
+         tempColor = CRT_colors[TEMPERATURE_COOL];
+      else if (temperature < temp_hot)
+         tempColor = CRT_colors[TEMPERATURE_MEDIUM];
+      else
+         tempColor = CRT_colors[TEMPERATURE_HOT];
+
+      // output the temperature
+      char buffer[4];
+      snprintf(buffer, 4, "%3d", temperature);
+      RichString_append(out, tempColor, buffer);
+      snprintf(buffer, 3, "%s", CRT_degreeSign());
+      RichString_append(out, CRT_colors[METER_TEXT], buffer);
+      if (this->pl->settings->fahrenheitTemperature)
+         RichString_append(out, CRT_colors[METER_TEXT], "F ");
+      else
+         RichString_append(out, CRT_colors[METER_TEXT], "C ");
+   }
+   free(cores_temperatures);
+}
+
+MeterClass TemperatureMeter_class = {
+   .super = {
+      .extends = Class(Meter),
+      .display = TemperatureMeter_display,
+      .delete = Meter_delete,
+   },
+   .updateValues = TemperatureMeter_updateValues,
+   .defaultMode = TEXT_METERMODE,
+   .maxItems = 1,
+   .total = 100.0,
+   .attributes = TemperatureMeter_attributes,
+   .name = "CPUTemperature",
+   .uiName = "CPU Temperature",
+   .caption = "CPU Temperature: "
+};
--- a/TemperatureMeter.h
+++ b/TemperatureMeter.h
@@ -1,0 +1,18 @@
+#ifndef HEADER_TemperatureMeter
+#define HEADER_TemperatureMeter
+/*
+htop - TemperatureMeter.h
+(C) 2013 Ralf Stemmer
+(C) 2014 Blair Bonnett
+(C) 2020 Maxim Kurnosenko
+Released under the GNU GPL, see the COPYING file
+in the source distribution for its full text.
+*/
+
+#include "Meter.h"
+
+extern int TemperatureMeter_attributes[];
+
+extern MeterClass TemperatureMeter_class;
+
+#endif
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,15 +45,15 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Install Dependencies
-      run: sudo apt-get install libncursesw5-dev libhwloc-dev libnl-3-dev libnl-genl-3-dev
+      run: sudo apt-get install libncursesw5-dev libhwloc-dev libnl-3-dev libnl-genl-3-dev libsensors4-dev
     - name: Bootstrap
       run: ./autogen.sh
     - name: Configure
-      run: ./configure --enable-werror --enable-openvz --enable-cgroup --enable-vserver --enable-ancient-vserver --enable-taskstats --enable-unicode --enable-hwloc --enable-setuid --enable-delayacct
+      run: ./configure --enable-werror --enable-openvz --enable-cgroup --enable-vserver --enable-ancient-vserver --enable-taskstats --enable-unicode --enable-hwloc --enable-setuid --enable-delayacct --enable-libsensors
     - name: Build
       run: make -k
     - name: Distcheck
-      run: make distcheck DISTCHECK_CONFIGURE_FLAGS='--enable-werror --enable-openvz --enable-cgroup --enable-vserver --enable-ancient-vserver --enable-taskstats --enable-unicode --enable-hwloc --enable-setuid --enable-delayacct'
+      run: make distcheck DISTCHECK_CONFIGURE_FLAGS='--enable-werror --enable-openvz --enable-cgroup --enable-vserver --enable-ancient-vserver --enable-taskstats --enable-unicode --enable-hwloc --enable-setuid --enable-delayacct --enable-libsensors'
 
   build-ubuntu-latest-clang-analyzer:
     runs-on: ubuntu-latest
--- a/CRT.c
+++ b/CRT.c
@@ -101,6 +101,11 @@ int CRT_colorSchemes[LAST_COLORSCHEME][LAST_COLORELEMENT] = {
       [METER_VALUE_IOWRITE] = ColorPair(Blue,Black),
       [LED_COLOR] = ColorPair(Green,Black),
       [TASKS_RUNNING] = A_BOLD | ColorPair(Green,Black),
+#ifdef HAVE_LIBSENSORS
+      [TEMPERATURE_COOL] = A_DIM | ColorPair(Green,Black),
+      [TEMPERATURE_MEDIUM] = A_NORMAL | ColorPair(Yellow,Black),
+      [TEMPERATURE_HOT] = A_BOLD | ColorPair(Red,Black),
+#endif
       [PROCESS] = A_NORMAL,
       [PROCESS_SHADOW] = A_BOLD | ColorPairGrayBlack,
       [PROCESS_TAG] = A_BOLD | ColorPair(Yellow,Black),
@@ -177,6 +182,11 @@ int CRT_colorSchemes[LAST_COLORSCHEME][LAST_COLORELEMENT] = {
       [METER_VALUE_IOWRITE] = A_NORMAL,
       [LED_COLOR] = A_NORMAL,
       [TASKS_RUNNING] = A_BOLD,
+#ifdef HAVE_LIBSENSORS
+      [TEMPERATURE_COOL] = A_DIM,
+      [TEMPERATURE_MEDIUM] = A_NORMAL,
+      [TEMPERATURE_HOT] = A_BOLD,
+#endif
       [PROCESS] = A_NORMAL,
       [PROCESS_SHADOW] = A_DIM,
       [PROCESS_TAG] = A_BOLD,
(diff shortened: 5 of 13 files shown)


DX37:
Fixing errors now.

TemperatureMeter.c
```diff
@@ -0,0 +1,102 @@
...
+
+      // convert the string into an integer, this is necessary for further steps
+      temperature = strtol(tstart, &tend, 10);
+      if (temperature == INT_MAX || temperature == INT_MIN) continue;
+      if (tstart == tend) continue;
+
+      // choose the color for the temperature
+      int tempColor;
+      if      (temperature < 60)                      tempColor = coolColor;
+      else if (temperature >= 60 && temperature < 70) tempColor = mediumColor;
```
cgzones, line 75: `temperature >= 60` is always true here

TemperatureMeter.c
```diff
@@ -0,0 +1,102 @@
...
+   ProcessList* pl = this->pl;
+   this->total = pl->totalTasks;
+   this->values[0] = pl->runningTasks;
+   snprintf(buffer, len, "%d/%d", (int) this->values[0], (int) this->total);
+}
+
+static void TemperatureMeter_display(Object* cast, RichString* out) {
+   (void) cast;
+   FILE *p;
+   p = popen("sensors", "r");
```
cgzones, line 39: Is there a non-exec way of getting the cup temperature portable?
(`/sys/class/thermal/thermal_zone0/temp`)

DX37: Maybe. I used this path to get temperature of my Raspberry Pi (htop-temperature-rpi in AUR). But I believe it'll be limited or may somehow not work on some systems, like `sensors` not working on some of them.

TemperatureMeter.c
DX37: Alright, got it.

TemperatureMeter.c
DX37: For example, I have on my ASUS UX433FN ten thermal zones. But on other notebooks and devices they can differ. And `sensors` covers almost all of them.

cgzones:
As an alternative one could get the cpu temperature over libsensors(3):

```
#include <errno.h>
#include <stdlib.h>
#include <string.h>

#include <sensors/sensors.h>

static int get_cpu_temp(double *temp) {
    int n = 0;

    for (const sensors_chip_name *chip = sensors_get_detected_chips(NULL, &n); chip; chip = sensors_get_detected_chips(NULL, &n)) {
        int r;
        char buffer[256];
        r = sensors_snprintf_chip_name(buffer, sizeof(buffer), chip);
        if (r < 0)
            continue;

        if (strncmp(buffer, "coretemp", strlen("coretemp")) != 0)
            continue;

        int m = 0;
        for (const sensors_feature *feature = sensors_get_features(chip, &m); feature; feature = sensors_get_features(chip, &m)) {
            if (feature->type != SENSORS_FEATURE_TEMP)
                continue;

            if (feature->number != 0)
                continue;

            const sensors_subfeature *sub_feature = sensors_get_subfeature(chip, feature, SENSORS_SUBFEATURE_TEMP_INPUT);
            if (sub_feature) {
                double local_temp;
                r = sensors_get_value(chip, sub_feature->number, &local_temp);
                if (r == 0) {
                    *temp = local_temp;
                    return 0;
                }
            }
        }
    }

    return -ENOENT;
}

int main() {
    int r;
    double temp;

    r = sensors_init(NULL);
    if (r < 0) {
        fprintf(stderr, "Failed to initilaze sensors: %m\n");
        return EXIT_FAILURE;
    }

    r = get_cpu_temp(&temp);
    if (r < 0) {
        fprintf(stderr, "Failed to get cpu temperature\n");
        sensors_cleanup();
        return EXIT_FAILURE;
    }

    printf("%f\n", temp);

    sensors_cleanup();

    return EXIT_SUCCESS;
}
```

natoscott:
+1 to using libsensors

Makefile.am
```diff
@@ -47,6 +47,7 @@ linux_platform_headers = \
 	linux/LinuxProcessList.h \
 	linux/LinuxCRT.h \
 	linux/Battery.h \
+	TemperatureMeter.h \
```
cgzones, line 50: This should probably go under `myhtopheaders`

Makefile.am
```diff
@@ -55,7 +56,7 @@ linux_platform_headers = \
 if HTOP_LINUX
 AM_LDFLAGS += -rdynamic
 myhtopplatsources = linux/Platform.c linux/IOPriorityPanel.c \
-linux/LinuxProcess.c linux/LinuxProcessList.c linux/LinuxCRT.c linux/Battery.c \
+linux/LinuxProcess.c linux/LinuxProcessList.c linux/LinuxCRT.c linux/Battery.c TemperatureMeter.c \
```
cgzones, line 59: This should probably go under `myhtopsources`

TemperatureMeter.c
```diff
@@ -0,0 +1,129 @@
...
+#include "CRT.h"
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <limits.h>
+#include <errno.h>
+#include <sensors/sensors.h>
+
+/*{
```
cgzones, line 22: please drop this part, auto-generated headers have been removed

TemperatureMeter.c
```diff
@@ -0,0 +1,129 @@
...
+    return -ENOENT;
+}
+
+int TemperatureMeter_attributes[] = {
+   TEMPERATURE_COOL,
+   TEMPERATURE_MEDIUM,
+   TEMPERATURE_HOT,
+};
+
+static void TemperatureMeter_setValues(Meter* this, char* buffer, int len) {
```
cgzones, line 68: This function seems to be just copy-pasted.
The `_updateValues` functions print the line for `BAR_METERMODE` (`[|||                        ]`).

Rename to `TemperatureMeter_updateValues`, and only set `this->values[0]` to `cached_cpu_temp` (see below).
Also use `xSnprintf`

TemperatureMeter.c
```diff
@@ -0,0 +1,129 @@
...
+   this->values[0] = pl->runningTasks;
+   snprintf(buffer, len, "%d/%d", (int) this->values[0], (int) this->total);
+}
+
+static void TemperatureMeter_display(Object* cast, RichString* out) {
+   (void) cast;
+   int r, temperature;
+   double temp;
+
+   r = sensors_init(NULL);
```
cgzones, line 80: Do not call `sensors_init`+`sensors_cleanup` every second.
Call it once in `htop.c` or some other top-level code.

TemperatureMeter.c
```diff
@@ -0,0 +1,129 @@
...
+
+   sensors_cleanup();
+
+   int textColor   = CRT_colors[METER_TEXT];
+   int coolColor   = CRT_colors[TEMPERATURE_COOL];
+   int mediumColor = CRT_colors[TEMPERATURE_MEDIUM];
+   int hotColor    = CRT_colors[TEMPERATURE_HOT];
+
+   // convert double to integer
+   temperature = (int)temp;
```
cgzones, line 101: use a file local `static` variable for the temperature, so it can be accessed in `TemperatureMeter_updateValues`

TemperatureMeter.c
```diff
@@ -0,0 +1,129 @@
...
+   int tempColor;
+   if      (temperature < 60) tempColor = coolColor;
+   else if (temperature < 70) tempColor = mediumColor;
+   else                       tempColor = hotColor;
+
+   // output the temperature
+   char buffer[30];
+   sprintf(buffer, "%d", temperature);
+   RichString_append(out, tempColor, buffer);
+   RichString_append(out, textColor, "°C ");
```
cgzones, line 113: Maybe we should add a configuration switch to use `°F`?

TemperatureMeter.c
```diff
@@ -0,0 +1,129 @@
...
+
+MeterClass TemperatureMeter_class = {
+   .super = {
+      .extends = Class(Meter),
+      .display = TemperatureMeter_display,
+      .delete = Meter_delete,
+   },
+   .updateValues = TemperatureMeter_setValues,
+   .defaultMode = TEXT_METERMODE,
+   .total = 100.0,
```
cgzones, line 124: `100` should be generally a fine max value, however my chip says `crit = +105.0°C`.
We can also get a second temperature in `get_cpu_temp` from `SENSORS_SUBFEATURE_TEMP_MAX` and override the default value set here in `TemperatureMeter_display`
Also when using Fahrenheit this needs to be adjusted.

TemperatureMeter.h
```diff
@@ -0,0 +1,20 @@
+/* Do not edit this file. It was automatically generated. */
```
cgzones, line 1: Drop this comment as auto-generated headers have been removed.

cgzones (changes requested):
Please rebase onto the master branch and squash into a single commit.

DX37:
Okay, this will take time, since I was only bumping version of package on Arch Linux User Repository and not touched original patch since 2017, I guess.

cgzones:
See #111 as an alternative implementation.

DX37:
> 
> 
> See #111 as an alternative implementation.

Looks interesting, but does it really needed to be inline? Overall temperature of CPU should be enough, I guess.
And load bars will mess up temperature values look, just as it messes with load percents look.

cgzones:
> Looks interesting, but does it really needed to be inline? Overall temperature of CPU should be enough, I guess.
And load bars will mess up temperature values look, just as it messes with load percents look.

I don't quite understand your point.
Added an example screenshot under load.
And do we need to spend an extra line just to display a single number?

DX37:
> 
> And do we need to spend an extra line just to display a single number?
> 
Battery meters do that, for example. And uptime and time meters.
Maybe I can write additional code to display temperatures of every core, but I'm quite new in libsensors and it'll take a while.

DX37:
You know, if you already added frequency meter in CPU load bars and all already agreed with it, I think I can stop to working on this pull request and let you do yours.

DX37:
Aaand I messed up. Nice.

Makefile.am
DX37: Then it'll fail to build on FreeBSD and macOS.

Makefile.am
DX37: Same as in comment below.

Makefile.am
DX37: Nevermind, managed to add --enable-sensors.

Makefile.am
DX37: Nevermind, managed to add --enable-sensors.

DX37:
Implementation is ready. Shows all cores temperatures in all meter modes and can be switched to Fahrenheit.

изображение
изображение
изображение
изображение

Also found out, that my Raspberries support `libsensors`, so I managed to add showing theirs CPU temperature too. Definitely works on Arch Linux on 5.4.51 and 5.4.64 kernels. On other distributions I didn't tested it, but if it's not supported there, it'll just print `n/a`.

DX37:
In origin, this patch purpose was to parse `sensors` command and print all cores temperature, not single number. And I wasn't author of this patch.
By rewriting it to `libsensors`, I returned to same purpose, step-by-step.

DX37:
Okay, this looks solid to me.

DX37:
> Please rebase onto the master branch and squash into a single commit.

Er... Done? (testing github answer to changes request)

TemperatureMeter.c
```diff
@@ -0,0 +1,155 @@
...
+static int get_cpu_sensors_amount() {
+   int n = 0, count = 0;
+
+   for (const sensors_chip_name *chip = sensors_get_detected_chips(NULL, &n); chip; chip = sensors_get_detected_chips(NULL, &n)) {
+      char buffer[64];
+      if (sensors_snprintf_chip_name(buffer, sizeof(buffer), chip) < 0)
+         continue;
+
+      // cpu_thermal is for Raspberry Pi sensor support
+      if ((strncmp(buffer, "coretemp", strlen("coretemp")) != 0) && (strncmp(buffer, "cpu_thermal", strlen("cpu_thermal")) != 0))
```
cgzones, line 33: Maybe use `String_startsWith()`

TemperatureMeter.c
```diff
@@ -0,0 +1,155 @@
...
+
+      // choose the color for the temperature
+      int tempColor;
+      if      (temperature < temp_medium) tempColor = CRT_colors[TEMPERATURE_COOL];
+      else if (temperature < temp_hot)    tempColor = CRT_colors[TEMPERATURE_MEDIUM];
+      else                                tempColor = CRT_colors[TEMPERATURE_HOT];
+
+      // output the temperature
+      char buffer[4];
+      sprintf(buffer, "%3d", temperature);
```
cgzones, line 132: Please use `snprintf(buffer, sizeof(buffer), ...`.
It's fine here, but `sprintf` smells

TemperatureMeter.c
```diff
@@ -0,0 +1,155 @@
...
+static int get_cpu_temp(Meter* this, double *temp, int core, sensors_subfeature_type type) {
+   int n = 0;
+
+   for (const sensors_chip_name *chip = sensors_get_detected_chips(NULL, &n); chip; chip = sensors_get_detected_chips(NULL, &n)) {
+      char buffer[64];
+      if (sensors_snprintf_chip_name(buffer, sizeof(buffer), chip) < 0)
+         continue;
+
+      // cpu_thermal is for Raspberry Pi sensor support
+      if ((strncmp(buffer, "coretemp", strlen("coretemp")) != 0) && (strncmp(buffer, "cpu_thermal", strlen("cpu_thermal")) != 0))
```
cgzones, line 55: Maybe use `String_startsWith()`

cgzones (changes requested):
Could you add
```
index 00c49ce..b60170f 100644
--- a/Meter.c
+++ b/Meter.c
@@ -375,10 +375,13 @@ static void LEDMeterMode_draw(Meter* this, int x, int y, int w) {
    int xx = x + strlen(this->caption);
    int len = RichString_sizeVal(out);
    for (int i = 0; i < len; i++) {
-      char c = RichString_getCharVal(out, i);
+      unsigned char c = RichString_getCharVal(out, i);
       if (c >= '0' && c <= '9') {
          LEDMeterMode_drawDigit(xx, y, c-48);
          xx += 4;
+      } else if (c == 176) {
+         mvaddch(yText, xx, ACS_DEGREE);
+         xx += 1;
       } else {
          mvaddch(yText, xx, c);
          xx += 1;
```
so the LED mode looks fine

DX37:
> Could you add
> 
> ```
> diff --git a/Meter.c b/Meter.c
> index 00c49ce..b60170f 100644
> --- a/Meter.c
> +++ b/Meter.c
> @@ -375,10 +375,13 @@ static void LEDMeterMode_draw(Meter* this, int x, int y, int w) {
>     int xx = x + strlen(this->caption);
>     int len = RichString_sizeVal(out);
>     for (int i = 0; i < len; i++) {
> -      char c = RichString_getCharVal(out, i);
> +      unsigned char c = RichString_getCharVal(out, i);
>        if (c >= '0' && c <= '9') {
>           LEDMeterMode_drawDigit(xx, y, c-48);
>           xx += 4;
> +      } else if (c == 176) {
> +         mvaddch(yText, xx, ACS_DEGREE);
> +         xx += 1;
>        } else {
>           mvaddch(yText, xx, c);
>           xx += 1;
> ```
> 
> so the LED mode looks fine

All requested changes are done.

DX37 (commented):
Changes done (tried to tell GitHub about approved changes).

cgzones (approved):
LGTM code wise :+1:

I'll do some final testing.

DX37:
God, I forgot about second srncmp replacement, thought it was a duplicate. Fixing.

DX37:
Checked code again. All fixed now.

cgzones:
Further comments: 
- The degree character (`°` - 176) is an extended ASCII character and causes problems, e.g. on systems with POSIX locale. (see #111 `CRT_degreeSign()`). It affects besides runtime also the checked-out source code.
- On systems with a big core count we are space limited by the Meter width.

BenBE:
Do you see a chance to cache the `sensors_get_detected_chips()` and `sensors_get_features()` to be cached before processing `get_cpu_temp()` for each core?

TemperatureMeter.c
```diff
@@ -0,0 +1,154 @@
...
+static int get_cpu_sensors_amount(void) {
+   int n = 0, count = 0;
+
+   for (const sensors_chip_name *chip = sensors_get_detected_chips(NULL, &n); chip; chip = sensors_get_detected_chips(NULL, &n)) {
+      char buffer[64];
+      if (sensors_snprintf_chip_name(buffer, sizeof(buffer), chip) < 0)
+         continue;
+
+      // cpu_thermal is for Raspberry Pi sensor support
+      if ((String_startsWith(buffer, "coretemp") == 0) && (String_startsWith(buffer, "cpu_thermal") == 0))
```
BenBE, line 32: `String_startsWith` is defined as a macro that already checks for `== 0` on the return value of `strncmp`, thus returning `bool`. Thus this line better would read `!String_startsWith(...)` or `String_startsWith(...) == false`.

TemperatureMeter.c
```diff
@@ -0,0 +1,154 @@
...
+            continue;
+
+         if (feature->number != core)
+            continue;
+
+         const sensors_subfeature *sub_feature = sensors_get_subfeature(chip, feature, type);
+         if (sub_feature) {
+            double local_temp;
+            if (sensors_get_value(chip, sub_feature->number, &local_temp) == 0) {
+               *temp = this->pl->settings->fahrenheitTemperature ? local_temp * 9 / 5 + 32 : local_temp;
```
BenBE, line 69: More some nitpicking, but the Fahrenheit conversion should be a separate function to decouple the value acquisition and display conversion logically.

TemperatureMeter.c
```diff
@@ -0,0 +1,154 @@
...
+      // output the temperature
+      char buffer[4];
+      snprintf(buffer, 4, "%3d", temperature);
+      RichString_append(out, tempColor, buffer);
+      if (this->pl->settings->fahrenheitTemperature)
+         RichString_append(out, CRT_colors[METER_TEXT], "°F ");
+      else
+         RichString_append(out, CRT_colors[METER_TEXT], "°C ");
+   }
+}
```
BenBE, line 138: Remodel the `get_cpu_sensors_amount()` function to both count the CPUs AND retrieve status information for each core at once. The returned structure should best be indexed in a wait for the actual information display step to directly lookup the proper information without additional searching in the returned data, but direct array access. Current O(n³) complexity is deadly for systems with many CPU cores when basically O(n) or O(n*log(n)) should easily be achievable.

BenBE (changes requested):

TemperatureMeter.c
DX37: Reworked to structure with dynamic array.

DX37:
Resolved everything, except for Meter width - don't have any big Linux servers at home :smile:

TemperatureMeter.c
```diff
@@ -0,0 +1,153 @@
...
+static int temp_converter(Meter* this, double temp) {
+   if (this->pl->settings->fahrenheitTemperature)
+      return temp * 9 / 5 + 32;
+   else
+      return temp;
+}
+
+static Dynarr* get_cpu_temp(Meter* this, sensors_subfeature_type type) {
+   int i = 0, n = 0, count = 0;
+   Dynarr* result = malloc(sizeof(Dynarr));
```
BenBE, line 38: Use `xMalloc` from `XAlloc.h` or check for proper return value.

TemperatureMeter.c
```diff
@@ -0,0 +1,153 @@
...
+      if (!String_startsWith(buffer, "coretemp") && !String_startsWith(buffer, "cpu_thermal"))
+         continue;
+
+      int m = 0;
+      for (const sensors_feature *feature = sensors_get_features(chip, &m); feature; feature = sensors_get_features(chip, &m)) {
+         if (feature->type == SENSORS_FEATURE_TEMP)
+            count++;
+      }
+
+      result->array = calloc(count, sizeof(int));
```
BenBE, line 55: Use `xCalloc` from `XAlloc.h` or check proper return value.

TemperatureMeter.c
```diff
@@ -0,0 +1,153 @@
...
+   TEMPERATURE_MEDIUM,
+   TEMPERATURE_HOT,
+};
+
+static void TemperatureMeter_updateValues(Meter* this, char* buffer, int len) {
+   Dynarr* core_temperature = get_cpu_temp(this, SENSORS_SUBFEATURE_TEMP_INPUT);
+   double temperature = core_temperature->array[0];
+
+   // Show only package temperature in bars meter mode
+   if (temperature < 0) {
```
BenBE, line 88: Comment doesn't match code it preceeds.

TemperatureMeter.c
```diff
@@ -0,0 +1,153 @@
...
+      xSnprintf(buffer, len, "N/A");
+      return;
```
BenBE, line 90: This leaks the `core_temperature` array.

TemperatureMeter.c
```diff
@@ -0,0 +1,153 @@
...
+   }
+   this->total = get_cpu_temp(this, SENSORS_SUBFEATURE_TEMP_MAX)->array[0];
```
BenBE, line 92: Memory leak of pointer returned by `get_cpu_temp`.

TemperatureMeter.c
```diff
@@ -0,0 +1,153 @@
...
+   else
+      xSnprintf(buffer, len, "%d%sC", (int) temperature, CRT_degreeSign());
+   free(core_temperature);
+}
+
+static void TemperatureMeter_display(Object* cast, RichString* out) {
+   Meter* this                = (Meter*)cast;
+   Dynarr* cores_temperatures = get_cpu_temp(this, SENSORS_SUBFEATURE_TEMP_INPUT);
+   int temp_medium            = temp_converter(this, 60);
+   int temp_hot               = temp_converter(this, 70);
```
BenBE, line 106: Not sure what common values nowadays for these should be. But given we are talking about 10K of 70K effective range (35°C - 105°C), we probably should make this a bit wider: 50-55°C for medium, 75-85°C for hot?

TemperatureMeter.c
```diff
@@ -0,0 +1,153 @@
...
+
+   if (cores_temperatures->array[0] < 0 || cores_temperatures->size == 0) {
+      RichString_append(out, CRT_colors[METER_VALUE], "n/a");
+      return;
+   }
```
BenBE, line 111: Leaks `core_temperatures`.

TemperatureMeter.c
```diff
@@ -0,0 +1,153 @@
...
+
+   for (int core = core_count_begin; core < cores_temperatures->size; ++core) {
+      // convert double to integer
+      int temperature = (int) cores_temperatures->array[core];
+
+      // choose the color for the temperature
+      int tempColor;
+      if      (temperature < temp_medium) tempColor = CRT_colors[TEMPERATURE_COOL];
+      else if (temperature < temp_hot)    tempColor = CRT_colors[TEMPERATURE_MEDIUM];
+      else                                tempColor = CRT_colors[TEMPERATURE_HOT];
```
BenBE, line 123: We don't need to save on lines; I'd prefer linebreak (and optionally braces) after the conditions.

BenBE (changes requested):

DX37:
Resolved suggested changes, rebased onto master and reworked use of subfeature type to save max_temp of every core into structure.

BenBE:
LGTM. At least didn't notice anything new. :grin:

DX37:
Nothing new, except for rebase onto master.

BenBE:
> Nothing new, except for rebase onto master.

Any open points/loose ends left?

DX37:
> > Nothing new, except for rebase onto master.
> 
> Any open points/loose ends left?

Well since `Meter` and `Object` vars were changed to `const`, changed all corresponding vars to `const`. That's all.
All loose ends was closed in previous requested changes.

PR #94, opened 2020-09-05
add support for building with meson

The-King-of-Toasters:
This is just #80 again because I screwed up rebasing, thanks git.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,226 @@
+project(
+	'htop',
+	'c',
+	version: '3.2.0',
+	meson_version: '>=0.53.0',
+	default_options: [
+		'c_std=c99',
+		'warning_level=2',
+	],
+)
+
+add_project_arguments(
+	'-DSYSCONFDIR="@0@"'.format(get_option('sysconfdir')),
+	language: 'c'
+)
+
+if get_option('debug')
+	add_project_arguments('-DDEBUG', language: 'c')
+endif
+
+cc   = meson.get_compiler('c')
+os   = host_machine.system()
+conf = configuration_data()
+
+conf.set_quoted('COPYRIGHT', '(C) 2004-2018 Hisham Muhammad')
+conf.set_quoted('VERSION', meson.project_version())
+
+zfs_sources = []
+zfs_inc = ''
+
+math   = cc.find_library('m', required: false)
+
+# Here goes the curses detection, ugh...
+nc_libs    = []
+nc_headers = []
+
+if get_option('unicode')
+	nc_libs += ['ncursesw']
+	nc_headers += [
+		'ncursesw/ncurses.h',
+		'ncursesw.h'
+	]
+	# Special header incantations for wide char support
+	add_project_arguments([
+		'-D_XOPEN_SOURCE_EXTENDED',
+		'-DNCURSES_WIDECHAR'],
+		language: 'c')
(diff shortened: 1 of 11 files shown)


meson.build
```diff
@@ -0,0 +1,208 @@
+project(
+        'htop',
+	'c',
```
hugomg, line 3: It looks like it is mixing tabs and spaces here?

The-King-of-Toasters: oops, lemme fix that.

The-King-of-Toasters:
I need to go through the last couple commits and see if I need to change anything (e.g. HTOP_LINUX et al.).

The-King-of-Toasters:
Ok, I made hwloc and delay accounting a feature, so that if they're disabled at build time they won't be included in the lib list. Also added checks for linux-affinity and hwloc, along with adding the header for setuid support. Will test on other machines.

BenBE:
@The-King-of-Toasters Can you check for #126 to be properly handled by meson? TIA.

The-King-of-Toasters:
Thanks to Ninja you get header checking for free. Touching `CRT.c` only compiles `CRT.o`, but touching `CRT.h` re-compiles everything.

BenBE:
We're not supporting alternative build systems as of now.

Salamandar:
@BenBE do you still refuse this MR ?
Since 2020, Meson became more and more present in the Linux landscape, low-level libs are using it.
Among the arguments, let me cite:
* easy-to-find build options (just read the meson_options.txt file) for maintainers/new users
* out-of-tree build
* out of the box support for advanced build options like LTO, unity build etc.

If you'd reconsider this, I'd love to pick this MR where it was left.

fasterit:
@Salamandar: We still see no reason to support alternative build systems. They all have upsides and downsides and none are compelling enough to consider anything beyond Autotools for now.

eli-schwartz:
> Among the arguments, let me cite:
> 
>  * easy-to-find build options (just read the meson_options.txt file) for maintainers/new users
> 
>  * out-of-tree build
> 
>  * out of the box support for advanced build options like LTO, unity build etc.

To be fair to autotools, it already supports the first two.

All build options are available in the `./configure --help` output (just like `meson configure`). This actually should not be hard for a build system to do, but cmake still fails at it... :D 

autotools does support out of tree builds as well as in-tree builds. This assumes a correctly-written Makefile.am, which is not actually guaranteed since most people don't really check that (if you use `make distcheck`, that will automatically check that VPATH builds succeed, but many people do not distcheck).

Mixing out of tree and in-tree builds on the same sourcedir is a recipe for disaster and will bomb out. Meson simply refuses to support in-tree builds at all.

fasterit:
Autotools will complain as well, if you mix the two: `configure: error: source directory already configured; run "make distclean" there first`.
And FTR: htop builds fine out-of-tree.

I think this is a case of "but I like $otherbuildsystem better". That is fair and fine. Just not reason for us to put any work in.
Autotools is old and not beautiful in any way. But it still works well enough.

Salamandar:
Alright, understood.

eli-schwartz:
> Autotools will complain as well, if you mix the two: `configure: error: source directory already configured; run "make distclean" there first`.

Only if you build in tree first, I think.

> And FTR: htop builds fine out-of-tree.

That is always good to hear. :)

PR #100, opened 2020-09-07
Hide userland threads by default

### LINKED ISSUE #99: Suggestion: Hide userland threads by default
People casually using htop usually have no idea what userland threads are for.
People who actually need to see them can easily enable them via SHIFT+H.

With them currently enabled by default, it is very inconvenient to go through the list and see what is running, taking up RAM, CPU usage and whatnot, therefore I think it'd be more user-friendly to not show them by default.

These values are 1 by default, change to 0 upon pressing SHIFT+H, I couldn't figure out what exactly to change in code otherwise I'd send a PR directly:
```
hide_userland_threads
hide_threads
```

Current situation:
image

Userland threads off (SHIFT+H):
image

C0rn3j:
Fixes #99

### DIFF
--- a/Settings.c
+++ b/Settings.c
@@ -356,11 +356,11 @@ Settings* Settings_new(int cpuCount) {
 
    this->sortKey = PERCENT_CPU;
    this->direction = 1;
-   this->hideThreads = false;
+   this->hideThreads = true;
    this->shadowOtherUsers = false;
    this->showThreadNames = false;
    this->hideKernelThreads = false;
-   this->hideUserlandThreads = false;
+   this->hideUserlandThreads = true;
    this->treeView = false;
    this->highlightBaseName = false;
    this->highlightMegabytes = false;


Settings.c
```diff
@@ -356,11 +356,11 @@ Settings* Settings_new(int cpuCount) {
 
    this->sortKey = PERCENT_CPU;
    this->direction = 1;
-   this->hideThreads = false;
+   this->hideThreads = true;
```
BenBE, line 359: This is a different setting actually. Just checked in the source. It's referenced at https://github.com/htop-dev/htop/blob/master/linux/LinuxProcessList.c#L865-L867 to implement some RedHat magic …

BenBE (changes requested):
Sorry for pointing to the wrong setting name in #99.

Settings.c
C0rn3j: The SHIFT+H keycombo changes both hideThreads nad hideUserlandThreads, I wasn't entirely sure what it did, but wanted it to be consistent with the shortcut.

Should I then only change hideUserLandThreads to true?

BenBE: Good question. Maybe see what the devs have to say about this.

cgzones:
I have no strong opinion an whether to disable or enable userland threads by default.

Looking at this change, what is the setting variable `hideThreads` even for?
The only switch based on it is in *LinuxProcessList.c*:
```
// The RedHat kernel hides threads with a dot.
// I believe this is non-standard.
if ((!settings->hideThreads) && name[0] == '.') {
   name++;
}
```
which should imho be based on `hideKernelThreads`.

BenBE:
Maybe the better name for `hideThreads` would have been `hideDotNamedThreads` …

@natoscott @cgzones What about renaming this configuration option (or remove it altogether?

cgzones:
The setting `hideThreads` got removed in ad72b747.

Now do we want this default configuration change or not?

fasterit:
No as per https://github.com/htop-dev/htop/issues/99#issuecomment-688247052

C0rn3j:
Closing this as both current maintainers and the original owner argued against it.

PR #101, opened 2020-09-07
No float equal compare

### LINKED ISSUE #78: Avoid floating point comparison with == / !=
There are several places where the source does an explicit comparison of floating point numbers with constants (in particular the battery state feature). As comparing floating point numbers with `==`/`!=` is unsafe, and the compared to values are for error cases (i.e. value unavailable) the better solution would be using `NAN` and check for the error case with `isnan()` instead of abusing those magic values.

Compile with `gcc -Wfloat-equal` to see those issues in the source.

If you like I can try to whip up some patch, though for consistency the semantic of some APIs might need slight changes (N/A and errors as `NAN` instead of `-1` for e.g. the battery level).

BenBE:
Fixes #78.

Please note comment regarding https://github.com/htop-dev/htop/commit/2c3e8eec528cd6579de7bb01fe1af912742d14d1; you might want to check if further changes are required.

### DIFF
--- a/CPUMeter.c
+++ b/CPUMeter.c
@@ -42,7 +42,7 @@ static void CPUMeter_updateValues(Meter* this, char* buffer, int size) {
    if (this->pl->settings->showCPUFrequency) {
       double cpuFrequency = this->values[CPU_METER_FREQUENCY];
       char cpuFrequencyBuffer[16];
-      if (cpuFrequency < 0) {
+      if (isnan(cpuFrequency)) {
          xSnprintf(cpuFrequencyBuffer, sizeof(cpuFrequencyBuffer), "N/A");
       } else {
          xSnprintf(cpuFrequencyBuffer, sizeof(cpuFrequencyBuffer), "%.0fMHz", cpuFrequency);
--- a/CPUMeter.c
+++ b/CPUMeter.c
@@ -83,12 +83,12 @@ static void CPUMeter_display(Object* cast, RichString* out) {
       xSnprintf(buffer, sizeof(buffer), "%5.1f%% ", this->values[CPU_METER_SOFTIRQ]);
       RichString_append(out, CRT_colors[METER_TEXT], "si:");
       RichString_append(out, CRT_colors[CPU_SOFTIRQ], buffer);
-      if (this->values[CPU_METER_STEAL]) {
+      if (!isnan(this->values[CPU_METER_STEAL])) {
          xSnprintf(buffer, sizeof(buffer), "%5.1f%% ", this->values[CPU_METER_STEAL]);
          RichString_append(out, CRT_colors[METER_TEXT], "st:");
          RichString_append(out, CRT_colors[CPU_STEAL], buffer);
       }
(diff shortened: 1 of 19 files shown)


BenBE:
Addresses #78.

cgzones:
Probably `-Wfloat-equal` should be added to the default warning flags. I'll take a deeper look after hopefully #105 got merged.

cgzones:
Please rebase and add `-Wfloat-equal` to default warnings in configure.ac .
(Both [GCC](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html) and [Clang](https://clang.llvm.org/docs/DiagnosticsReference.html) supports it.)

CPUMeter.c
```diff
@@ -83,12 +83,12 @@ static void CPUMeter_display(Object* cast, RichString* out) {
       xSnprintf(buffer, sizeof(buffer), "%5.1f%% ", this->values[CPU_METER_SOFTIRQ]);
       RichString_append(out, CRT_colors[METER_TEXT], "si:");
       RichString_append(out, CRT_colors[CPU_SOFTIRQ], buffer);
-      if (this->values[CPU_METER_STEAL]) {
+      if (this->values[CPU_METER_STEAL] < 0.005) {
```
cgzones, line 86: This seems odd. Don't we want to display significant values?

linux/Battery.c
```diff
@@ -299,20 +301,23 @@ void Battery_getData(double* level, ACPresence* isOnAC) {
...
+      if (isnan(*level)) {
          Battery_method = BAT_ERR;
       }
    }
    if (Battery_method == BAT_ERR) {
-      *level = -1;
+      *level = NAN;
       *isOnAC = AC_ERROR;
    }
+   if (*level < 0.0) {
```
cgzones, line 318: Maybe combine with the following condition to `CLAMP(*level, 0.0, 100.0)` ?

CPUMeter.c
BenBE: For this value I couldn't find documentation on how to distinguish no return from the system/kernel and just no CPU time spent on this. Suggestions on how to do this in a cleaner way welcome.

cgzones: https://github.com/htop-dev/htop/blob/5ea13e7ea9ca19130d5114a745d337b052cf2199/linux/Platform.c#L180 ?

linux/Battery.c
```diff
@@ -299,22 +301,21 @@ void Battery_getData(double* level, ACPresence* isOnAC) {
...
    }
    if (Battery_method == BAT_ERR) {
-      *level = -1;
+      *level = NAN;
       *isOnAC = AC_ERROR;
-   }
-   if (*level > 100.0) {
-      *level = 100.0;
+   } else {
+      CLAMP(*level, 0.0, 100.0);
```
cgzones, line 318: `*level = CLAMP(*level, 0.0, 100.0);`

cgzones (approved):

cgzones:
Thanks, applied.

PR #105, opened 2020-09-08
Add some base compiler warnings

cgzones:
Includes #102

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,4 +9,9 @@ os:
   - linux
   - osx
 
-script: ./autogen.sh && ./configure && make
+script:
+  - ./autogen.sh
+  # clang might warn about C11 generic selections in isnan()
+  - CFLAGS=-Wno-c11-extensions ./configure --enable-werror
+  - make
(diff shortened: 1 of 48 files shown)


BenBE:
For the unused arguments patch I'd suggest having a globally available macro `HTOP_UNUSED` that is used inside the function body. Using the `__attribute__((unused))` is mostly GCC syntax (and incidentally also supported on LLVM for GCC compat).

```c
// In some global header:
#define HTOP_UNUSED(v) (void)(v)

//Inside functions with unused arguments:
void example_foo(int unused_var) {
    HTOP_UNUSED(unused_var);

    // Rest of the function code.
}
```

Using that macro definition is portable even outside GCC/LLVM.

BenBE:
I'm not quite happy with the naming of those macros, as names with leading underscores are normally reserved for the compiler/standard library to use. Any particular reason for doing the current naming?

fasterit:
Could you separate out the fixes (e.g.2b98e3f) from the new precompiler macros? 
We just discussed this PR in the core team meeting (you had an invite but did not respond). May be you can join next week, I'll send you another invite email.

cgzones:
Separated hwlock-linux_affinity commit into #107

PR #106, opened 2020-09-08
minor vim bind improvements

### LINKED ISSUE #69: arrow keys broken in vim mode
The arrow keys probably should always work regardless the vim mode is on or off.

Ckath:
this addresses two things mainly:

1. keep arrow keys working in vim mode
2. allow `hjkl` movement in panels (open files, environment vars, etc)

the arrow keys are restored in vimmode by dropping `H` and `K` keys when vimmode is on and simply remapping `HJKL` to `hjkl` directly without the need to ruin arrow keys. `hjkl` movement in panels is simply added regardless of vimmode, as this doesnt collide with anything here and in the existing codebase it isnt trivial to access the settings from within panel.c

this would fix #69

### DIFF
--- a/Panel.c
+++ b/Panel.c
@@ -385,10 +385,12 @@ bool Panel_onKey(Panel* this, int key) {
    switch (key) {
    case KEY_DOWN:
    case KEY_CTRL('N'):
+   case 'j':
       this->selected++;
       break;
    case KEY_UP:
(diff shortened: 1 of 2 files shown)


natoscott:
See also: https://github.com/htop-dev/htop/pull/37

PR #111, opened 2020-09-10
Inline Temperature

cgzones:
Alternative implementation of #93 

Show the temperature inline as the CPU frequency.

temp1
temp2
temp3

### DIFF
--- a/CPUMeter.c
+++ b/CPUMeter.c
@@ -56,25 +56,44 @@ static void CPUMeter_updateValues(Meter* this, char* buffer, int size) {
       return;
    }
    memset(this->values, 0, sizeof(double) * CPU_METER_ITEMCOUNT);
+
+   char cpuUsageBuffer[8] = { 0 };
+   char cpuFrequencyBuffer[16] = { 0 };
+   char cpuTemperatureBuffer[16] = { 0 };
+
    double percent = Platform_setCPUValues(this, cpu);
+
+   if (this->pl->settings->showCPUUsage) {
+      xSnprintf(cpuUsageBuffer, sizeof(cpuUsageBuffer), "%5.1f%%", percent);
+   }
+
    if (this->pl->settings->showCPUFrequency) {
       double cpuFrequency = this->values[CPU_METER_FREQUENCY];
-      char cpuFrequencyBuffer[16];
       if (isnan(cpuFrequency)) {
          xSnprintf(cpuFrequencyBuffer, sizeof(cpuFrequencyBuffer), "N/A");
       } else {
          xSnprintf(cpuFrequencyBuffer, sizeof(cpuFrequencyBuffer), "%4uMHz", (unsigned)cpuFrequency);
       }
-      if (this->pl->settings->showCPUUsage) {
-         xSnprintf(buffer, size, "%5.1f%% %s", percent, cpuFrequencyBuffer);
-      } else {
-         xSnprintf(buffer, size, "%s", cpuFrequencyBuffer);
-      }
-   } else if (this->pl->settings->showCPUUsage) {
-      xSnprintf(buffer, size, "%5.1f%%", percent);
-   } else if (size > 0) {
-      buffer[0] = '\0';
    }
+
+   #ifdef HAVE_LIBSENSORS
+   if (this->pl->settings->showCPUTemperature) {
+      double cpuTemperature = this->values[CPU_METER_TEMPERATURE];
+      if (isnan(cpuTemperature))
+         xSnprintf(cpuTemperatureBuffer, sizeof(cpuTemperatureBuffer), "N/A");
+      else if (this->pl->settings->degreeFahrenheit)
+         xSnprintf(cpuTemperatureBuffer, sizeof(cpuTemperatureBuffer), "%3d%sF", (int)(cpuTemperature * 9 / 5 + 32), CRT_degreeSign);
+      else
+         xSnprintf(cpuTemperatureBuffer, sizeof(cpuTemperatureBuffer), "%d%sC", (int)cpuTemperature, CRT_degreeSign);
+   }
+   #endif
+
+   xSnprintf(buffer, size, "%s%s%s%s%s",
+             cpuUsageBuffer,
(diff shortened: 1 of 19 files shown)


CPUMeter.c
```diff
@@ -38,25 +38,41 @@ static void CPUMeter_updateValues(Meter* this, char* buffer, int size) {
...
-      }
-   } else if (this->pl->settings->showCPUUsage) {
-      xSnprintf(buffer, size, "%5.1f%%", percent);
-   } else if (size > 0) {
-      buffer[0] = '\0';
    }
+
+   if (this->pl->settings->showCPUTemperature) {
+      double cpuTemperature = this->values[CPU_METER_TEMPERATURE];
+      if (cpuTemperature < 0)
```
BenBE, line 62: What about very cold rooms? Prefer to use NAN for unavailable values. Cf. #101 for a similar change for the frequency stuff.

CPUMeter.c
```diff
@@ -109,6 +125,19 @@ static void CPUMeter_display(Object* cast, RichString* out) {
          RichString_append(out, CRT_colors[CPU_GUEST], buffer);
       }
    }
+
+   if (this->pl->settings->showCPUTemperature) {
+      char cpuTemperatureBuffer[10];
+      double cpuTemperature = this->values[CPU_METER_TEMPERATURE];
+      if (cpuTemperature < 0)
```
BenBE, line 132: Why not prefer `NAN`/`isnan` over magic numbers? Cf. #101.

Settings.c
```diff
@@ -312,10 +318,12 @@ Settings* Settings_new(int cpuCount) {
    this->countCPUsFromZero = false;
    this->showCPUUsage = true;
    this->showCPUFrequency = false;
+   this->showCPUTemperature = false;
    this->updateProcessNames = false;
    this->cpuCount = cpuCount;
    this->showProgramPath = true;
    this->highlightThreads = true;
+   this->degreeFahrenheit = false;
```
BenBE, line 326: What about display in Kelvin? /joke

CPUMeter.c
cgzones: converted

CPUMeter.c
cgzones: converted

BenBE:
Can you skip the blank after `°C` (before the closing `]` in the bar view)?

If the temperature is usually integer degrees I'd suggest dropping the fraction part in the display.

BenBE (commented):
Patch LGTM.

cgzones:
> Can you skip the blank after °C (before the closing ] in the bar view)?

That's not that simple. The `°` character is not an ASCII 7-bit character and somehow the length computation changes.
So for now I think the bar is more readable with the blank than with omitting the degree sign (`40C`).

>If the temperature is usually integer degrees I'd suggest dropping the fraction part in the display.

Changed the format to `%d` (Celsius) and `%3d` (Fahrenheit).

BenBE:
Why not `%3d` for both? Making them both align right.

cgzones:
> Why not %3d for both? Making them both align right.

Cause degree Celsius are normally two digit numbers.

BenBE:
> > Why not %3d for both? Making them both align right.
> 
> Cause degree Celsius are normally two digit numbers.

Asking the other way around: Why than bother with making Fahrenheit values 3 digits wide, when they usually are 3 digits anyway?

Just asking for why introducing the difference of `%d°C` vs. `%3dF` when both should produce valid results with `%d` (probably better in our case) or `%3d` (if we wanted to enforce alignment to the right).

cgzones:
> Asking the other way around: Why than bother with making Fahrenheit values 3 digits wide, when they usually are 3 digits anyway?

99°F is ~37°C, so I'd like to have a fixed position if the value hovers around that.

DX37:
Since on systems with Hyper-Threading `libsensors` shows temperature of only physical cores, inline temperature in CPU meters will look like this:
изображение

BenBE:
How much work would it be to get the logical and physical cores linked; i.e. duplicate the phys core 0 temp also shown for the HT core of that?

Having half the values display as N/A looks strange.

DX37:
What about [this implementation](https://github.com/htop-dev/htop/pull/93#issuecomment-693210723), which has triggered this pull request to happen, @BenBE?

cgzones:
Needs to be refactored for SMT/HT and multi-socket systems.

BenBE:
I'd somehow prefer the order `Temp Freq Load` in the display of the meter asthis causes the least change towards previous versions that only had the `Load` in the meter. This shouldn't cause too many issues regarding the `°C` issue either as there's a natural space guaranteed in that case.

Any other opinions?

CRT.c
```diff
@@ -69,6 +69,14 @@ bool CRT_utf8 = false;
 
 const char *const *CRT_treeStr = CRT_treeStrAscii;
 
+const char* CRT_degreeSign() {
+   #ifdef HAVE_LIBNCURSESW
+   if (CRT_utf8)
+      return "\xc2\xb0";
+   #endif
```
fasterit, line 76: With libcurses you can still use ° as it is within most non-UTF8 charsets (ISO-8859-x), even CP437, have it at 0x90.

cgzones: @fasterit can you take a look at the added commit

fasterit:
Without `--enable-sensors` the option in setup should not be available as it will always be "N/A".

CRT.c
```diff
@@ -74,6 +74,22 @@ const char *const *CRT_treeStr = CRT_treeStrAscii;
...
+const char* CRT_degreeSign;
+
+static const char* initDegreeSign(void) {
+#ifdef HAVE_LIBNCURSESW
+   if (CRT_utf8)
+      return "\xc2\xb0";
+#endif
+
+   static char buffer[4];
+   int r = snprintf(buffer, sizeof(buffer), "%lc", 176);
```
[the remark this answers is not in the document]
cgzones: no (maybe i'll add a comment): this might actually fail on platforms not supporting wide characters, e.g. with POSIX locale (occurred on a raspberry pi)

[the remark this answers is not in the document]
BenBE: I feared something like that. ;-) Comment will be appreciated.

Issue #113, opened 2020-09-11
Feature request: Add context switches


xkill:
It could be great if htop also displays the context switches per process

cgzones:
Could you take a look and try out #114 ?

xkill:
> Could you take a look and try out #114 ?

Replied on the Pull Request

xkill:
Seems that the new commit added the new feature and works really fine.
Thanks so much!!!!
;)

PR #114, opened 2020-09-11
Add Linux process column for total context switches

cgzones:
Displays the incremental sum of voluntary_ctxt_switches and nonvoluntary_ctxt_switches.

Closes: #113

### DIFF
--- a/htop.1.in
+++ b/htop.1.in
@@ -386,6 +386,9 @@ Which cgroup the process is in.
 .B OOM
 OOM killer score.
 .TP
+.B CTXT
+Incremental sum of voluntary and nonvoluntary context switches.
+.TP
 .B IO_PRIORITY (IO)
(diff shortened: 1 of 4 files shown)


xkill:
The problem with this patch is that it shows the accumulated context switches. It should show the difference between the actual status and the previous refresh.

cgzones:
Added commit to show the difference instead of accumulated total

xkill:
Great, works fine!!!

cgzones:
Should values above a certain value (>1000 ?) be highlighted (bold ?) ?

cgzones:
Should values above a certain value (>1000 ?) be highlighted (bold ?) ?

BenBE:
I think usual big number highlighting (starting blue for thousands, red for millions) would be fine …

cgzones:
Entries above 1000 are now display in bold.

Issue #112, opened 2020-09-10
Quiet failure on xSnprintf failure


cgzones:
If `xSnprintf` fails, htop will exit just with a meesage: `htop: Success`
Maybe we should use `abort` or similar instead or `err(1, NULL);` ?

BenBE:
Maybe reuse the SIGSEGV handler?

BenBE:
This issue should be fixed by #204.

cgzones:
Closed by #204

PR #115, opened 2020-09-11
Rework memory error handling and enable debug mode by default

cgzones:
Print exact location on a memory allocation or write (snprintf) failure,
instead of just exiting with: `htop: Success`

Enable asserts by not using `-DNDEBUG` by default.
Let users/package maintainers specify it if they want to.

Closes: #112


BenBE:
LTGM so far …

Makefile.am
```diff
@@ -12,9 +12,8 @@ applications_DATA = htop.desktop
 pixmapdir = $(datadir)/pixmaps
 pixmap_DATA = htop.png
 
-AM_CFLAGS += -pedantic -Wall $(wextra_flag) -std=c99 -D_XOPEN_SOURCE_EXTENDED -DSYSCONFDIR=\"$(sysconfdir)\" -I"$(top_srcdir)/$(my_htop_platform)"
```
fasterit, line 15: -(W)pedantic is not included in -Wall, why drop it?

fasterit (commented):
Why do you remove the inlines? Have you checked on the performance impact?

I'm not sure we want to invert the debug logic as dozens of distros need to adjust their packaging then and - typically - if you run `make` on what you just grabbed from git, you'd not expect a debug build either.

Makefile.am
cgzones: The statement expressions, `({ ... })`, inside the macros are not C99 (see https://gcc.gnu.org/onlinedocs/gcc/Statement-Exprs.html).
So `-pedantic` warns about them.

cgzones:
The `inline` specifier is quite useless. Compilers are free to inline any function they like and to not inline any function they don't like, whether the function is annotated with `inline` or not.
The specifier needs to be dropped to allow the `assert` statements inside.

Regarding `NDEBUG`: For example systemd gets build on Debian without it and they use asserts heavily.

Makefile.am
fasterit: Hm, and you think this patch is important enough to drop C99 compatibility for some loops in preprocessor instructions? (honest question)

fasterit:
> The `inline` specifier is quite useless. Compilers are free to inline any function they like and to not inline any function they don't like, whether the function is annotated with `inline` or not.

It tells the compiler _you_ want it inlined and sane compilers will obey.

> The specifier needs to be dropped to allow the `assert` statements inside.

Similar to the -pedantic above, you drop standards in the code base to modernize it just in the place that you touch.

I think this needs a more general approach to either stay with the current style / language level or go beyond that to some _defined_ level.

> Regarding `NDEBUG`: For example systemd gets build on Debian without it and they use asserts heavily.

I do not see why that is an argument for you to invert the logic of the code base and put work on every downstream packager.

BenBE:
Unless there's a really good reason to do so (and I currently see none) I'd prefer to avoid any GNU extensions and thus keep `-pedantic`.

For the `-DNDEBUG`: I'm okay with dropping that define, as this makes some potential issues more visible. And with the few asserts in the code base this shouldn't change too much overall either, even if downstream didn't re-add those for their builds.

Makefile.am
cgzones: Reworked in a C99 compliant fashion

XAlloc.h
```diff
@@ -5,34 +5,88 @@
...
-#undef xStrdup
-#undef xStrdup_
-#ifdef NDEBUG
-# define xStrdup_ xStrdup
-#else
-# define xStrdup(str_) (assert(str_), xStrdup_(str_))
-#endif
+#define xAsprintf(strp, fmt, ...)                   \
+   do {                                             \
+      int _ret = asprintf(strp, fmt, __VA_ARGS__);  \
```
BenBE, line 77: Prefer `vasprintf` as shown with my implementation in #155.

XAlloc.h
```diff
@@ -5,34 +5,88 @@
...
+      if (_ret < 0) {                               \
+         oom_failure(__FILE__, __LINE__);           \
+      }                                             \
+   } while(0)
 
-char* xStrdup_(const char* str) ATTR_NONNULL;
+#define xSnprintf(fmt, len, ...)                         \
```
BenBE, line 83: The macro's signature is wrong. Cf. #155.

XAlloc.h
cgzones: There is no need inside a macro for manual argument unpacking

XAlloc.h
cgzones: Zero sized variable arguments are a C extension, so `#define xSnprintf(buf, len, fmt, ...)` does not work in case of `xSnprintf(foo, 42, "Hello World");`

BenBE: That's not what #155 uses (partly for that reason).
Also, the current macro calls the first argument `fmt`, while it should be `buf`.
BTW: That naming issue was what led me to rewrite the macro as a proper function.

cgzones:
Superseded by #204

PR #116, opened 2020-09-11
Add Valgrind script

cgzones:
Add script to run valgrind with ncurses suppressions.

### DIFF
--- a/scripts/htop_suppressions.valgrind
+++ b/scripts/htop_suppressions.valgrind
@@ -1,0 +1,63 @@
+{
+   <ncurses internal memory allocated at startup>
+   Memcheck:Leak
+   match-leak-kinds: reachable
+   ...
+   fun:CRT_init
+   fun:main
+}
+
+{
+   <ncurses internal memory allocated at startup>
+   Memcheck:Leak
+   match-leak-kinds: reachable
+   ...
+   fun:CRT_init
+}
+
+{
+   <ncurses internal memory>
+   Memcheck:Leak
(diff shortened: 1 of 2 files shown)


BenBE (commented):
LGTM. Is the reason for the wgetch leak known somehow?

cgzones:
> Is the reason for the wgetch leak known somehow?

No. The man page of wgetch does not say anything about memory allocations.

The-King-of-Toasters:
Ncurses likes to keep memory around for performance' sake, but you can compile it so that it frees everything. More info [here](https://invisible-island.net/ncurses/ncurses.faq.html#config_leaks).

cgzones:
> Ncurses likes to keep memory around for performance' sake, but you can compile it so that it frees everything. More info here.

Yes, but supplying suppressions is more user friendly and less invasive than requiring a ncurses rebuild just for cleaner valgrind output.

scripts/htop_suppressions.valgrind
```diff
@@ -0,0 +1,63 @@
+{
+   <insert_a_suppression_name_here>
```
fasterit, line 2: It would be nice to have good suppression_names for the ones that should stay,
e.g. explain that ncurses retains memory and we cannot change that.
Not everybody is a valgrind expert :-).

BenBE:
On MacOS there seems to be some dylib loader leak (cf. investigation in #118), as well as potential issues with `getpwuid` and `localtime_r`. If someone could try to confirm these, it'd be nice.

PR #117, opened 2020-09-11
Call character checking function with unsigned char

cgzones:
See https://wiki.sei.cmu.edu/confluence/display/c/STR37-C.+Arguments+to+character-handling+functions+must+be+representable+as+an+unsigned+char


BenBE:
Shouldn't all those checks include `ch` to be between `0` and `255`?

cgzones:
> Shouldn't all those checks include ch to be between 0 and 255?

Applied

PR #121, opened 2020-09-12
Avoid modifying optarg

cgzones:
It might be working, but lets rather not modify getopt's global variable
`optarg`.

Also there is no need to call `getenv("USER")`, just use `geteuid()`.

### DIFF
--- a/htop.c
+++ b/htop.c
@@ -131,21 +131,21 @@ static CommandLineSettings parseArguments(int argc, char** argv) {
             }
             break;
          case 'u':
-            if (!optarg && optind < argc && argv[optind] != NULL &&
+         {
+            const char *username = optarg;
+            if (!username && optind < argc && argv[optind] != NULL &&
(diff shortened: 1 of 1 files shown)


htop.c
```diff
@@ -99,6 +99,7 @@ static CommandLineSettings parseArguments(int argc, char** argv) {
    int opt, opti=0;
    /* Parse arguments */
    while ((opt = getopt_long(argc, argv, "hvmCs:td:u::Up:", long_opts, &opti))) {
+      const char *username;
```
BenBE, line 102: I'd prefer to limit the scope for `username` to only the case label that it is used in. This requires some `{}` inside the `case 'u'`, but makes it clearer that it's only ever used there. Also I kinda dislike defining an uninitialized variable that only get's initialized much later (if at all).

cgzones: Fixed

PR #124, opened 2020-09-13
Use getpwuid_r to avoid static OS buffers

BenBE:


natoscott:
Hmm ... but why?  AFAICS there's no reason to do this (htop is not threaded, using static buffers is fine) and adding malloc/free of a local 16KB buffer just slows things down and adds more code.

BenBE:
This patch is mostly written to check the MacOS memleak issue.

We can make the 16K buffer static too. Intention was mostly to avoid any dynamic allocations in system libraries.

You can mark this WIP for now if you like.

cgzones:
I don't like this neither: `getpwuid()` is a totally usable function in a single-threaded program.
We should not complicate code, due to the Darwin's libc implementation not cleaning up a one-time static buffer.

cgzones:
NAK (for now)

PR #130, opened 2020-09-15
configure.ac: axe python check

The-King-of-Toasters:
Now that the automated header script is gone, there's no need for python
now.

Too late for 3.0.2, maybe a new release?


natoscott:
> Now that the automated header script is gone, there's no need for python
> now.

+1

> Too late for 3.0.2, maybe a new release?

No urgency, it's doing no harm.

PR #134, opened 2020-09-15
Use PROCDIR throughout instead of /proc on Linux

cgzones:


BenBE:
Shouldn't it be better be called `PROCDIR`?

cgzones:
> Shouldn't it be better be called PROCDIR?

Yes, fixed commit message

linux/Platform.c
```diff
@@ -230,8 +230,8 @@ void Platform_setZfsCompressedArcValues(Meter* this) {
    ZfsCompressedArcMeter_readStats(this, &(lpl->zfs));
 }
 char* Platform_getProcessEnv(pid_t pid) {
-   char procname[32+1];
-   xSnprintf(procname, 32, "/proc/%d/environ", pid);
+   char procname[128+1];
+   xSnprintf(procname, 128, PROCDIR "/%d/environ", pid);
```
BenBE, line 234: It's somehow bad style to repeat the buffer size, instead of using `sizeof(buffer)-1` as this creates multiple places that need updating if a buffer size is changed. Using the `sizeof` version avoids this pitfall.

BenBE (commented):
LGTM. Maybe decide if to use `sizeof` instead of repeating magic numbers.

linux/Platform.c
cgzones: True, I just wanted to keep the patch small

BenBE: When both lines are touched there's not much diff here ;-)

PR #135, opened 2020-09-15
Add DiskIOMeter for IO read/write usage

cgzones:
diskio


BenBE:
Just out of curiousity: Does the kernel somehow provide the information how much data was read from caches vs. from actual storage media? If it is available would be cool to have it displayed.

Patch is fine without that information too.

BenBE (approved):
LGTM. Changes seem reasonable.

cgzones:
> Just out of curiosity: Does the kernel somehow provide the information how much data was read from caches vs. from actual storage media? If it is available would be cool to have it displayed.

I am not sure. And cache is probably much faster than real disk IO and I think the real and slow disk IO is what user expects here.

PR #143, opened 2020-09-17
Feature to normalize process CPU percent.

PatrickGMM:
Feature divides the reported process CPU percent by the number of CPU
cores. Selected by command-line '-n' or '--normalize-cpu' and can be
toggled at main screen with the 'z' key. Disabled by default.


cgzones:
I think such a small feature should not have a command line option (maybe not even not a key binding).

PatrickGMM:
No problem...Just offering up what I have locally.

cgzones:
Please rebase and drop the command line argument (I don't see any strong reason for a command line option and we should not add one for every configuration option.)

fasterit:
@cgzones , @PatrickGMM : Why isn't this just a new column (NCPU%)?

BenBE:
Implemented by alternative provided in #292.

PR #146, opened 2020-09-17
Add --enable-debug configure option to enable asserts

cgzones:
asserts are still disabled by default.


natoscott:
@cgzones looks like this has a conflict on AffinityPanel.c at present?

cgzones:
Lets wait until #154 has been decided, cause I don't like the export of `Hashtable_isConsistent()` and `Vector_isConsistent()` for `extern inline` compatibility.

PR #147, opened 2020-09-17
Convert short version option to capital V and sort help message options

cgzones:


BenBE:
What about also taking care of #138 and make `--no-mouse` use `-M`?

Issue #72, opened 2020-09-02
Unreachable code


maddin200:
htop-master/solaris/SolarisProcess.c: line 154
` case ZONE:{
      xSnprintf(buffer, n, "%-*s ", ZONENAME_MAX/4, sp->zname); break;

      if (buffer[ZONENAME_MAX/4] != '\0') {    // <-- unreachable
         buffer[ZONENAME_MAX/4] = ' ';
         buffer[(ZONENAME_MAX/4)+1] = '\0';
      }
      break;
   }`

PR #148, opened 2020-09-17
Drop dead code after break

cgzones:
The code is also unnecessary as a trailing space is always printed via
the format string.

Closes: #72


BenBE (approved):
LGTM.

cgzones:
The commit description is wrong.

Why is only `ZONENAME_MAX/4` printed? And then the last printed character changed to a space?

PR #155, opened 2020-09-18
Xalloc dedup&cleanup

BenBE:
This cleans up some of the mess in `XAlloc.h` …

This also fixes an argument list issue the original macro had (`buf` was missing and wrongly labelled `fmt`).

### DIFF
--- a/XAlloc.c
+++ b/XAlloc.c
@@ -5,6 +5,8 @@
 #ifndef _GNU_SOURCE
 #define _GNU_SOURCE
 #endif
+
+#include <stdarg.h>
 #include <stdlib.h>
 #include <string.h>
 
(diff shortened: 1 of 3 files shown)


fasterit (changes requested):
Thank you very much for cleaning up. Much appreciated!

Can we have these in StringUtils please?
Nobody would expect them in XAlloc.

BenBE:
Given the naming I somewhat expected them to be there. Inside `XAlloc.*` all the functions start with a lowercase `x`. For those functions to properly fit the naming for `StringUtils.c`/`.h` They should be named `String_asnprintf`, `String_snprintf` and `String_strdup` instead (to stay consistent).

I could rename and move them over to `StringUtils.c`/`.h`, but doing so is better suited in a separate PR IMHO.

BenBE:
> Thank you very much for cleaning up. Much appreciated!
> 
> Can we have these in StringUtils please?
> Nobody would expect them in XAlloc.

I did a dryrun with just moving these functions (without rename) and that patch already causes quite a disruption. And with proper renaming (which IMHO should be done when moving these functions over to `StringUtils.h`) causes not just about 20 files, but nearly 100 that would need changes; across all platforms. Furthermore I'd have to redo quite a bit of #156 and #160; thus moving those functions is best done after #160 (which also allows to better clean up after this change).

Thus I'd prefer to go along with #155 as-is for now.

PR #156, opened 2020-09-18
Use ATTR_NONNULL to annotate declarations

BenBE:
If #155 gets merged before this, a minor change in `XAlloc.h` is recommended to include these attributes for the newly created/renamed functions `xAsnprintf`, `xSnprintf` and `xStrdup`.

I can also rebase this PR onto #155. If okay, I'd add the necessary changes to the above mentioned functions.

### DIFF
--- a/XAlloc.c
+++ b/XAlloc.c
@@ -5,6 +5,8 @@
 #ifndef _GNU_SOURCE
 #define _GNU_SOURCE
 #endif
+
+#include <stdarg.h>
 #include <stdlib.h>
 #include <string.h>
 
@@ -37,6 +39,32 @@ void* xRealloc(void* ptr, size_t size) {
       fail();
    }
    return data;
+}
+
+int xAsprintf(char** strp, const char* fmt, ...) {
+   va_list vl;
+   va_start(vl, fmt);
+   int _r = vasprintf(strp, fmt, vl);
+   va_end(vl);
+
+   if (_r < 0) {
+      fail();
+   }
+
+   return _r;
+}
+
+int xSnprintf(char* buf, int len, const char* fmt, ...) {
+   va_list vl;
+   va_start(vl, fmt);
+   int _n=vsnprintf(buf, len, fmt, vl);
+   va_end(vl);
+
+   if (!(_n > -1 && _n < len)) {
+      fail();
+   }
+
+   return _n;
 }
 
 char* xStrdup_(const char* str) {
--- a/XAlloc.c
+++ b/XAlloc.c
@@ -67,7 +67,7 @@ int xSnprintf(char* buf, int len, const char* fmt, ...) {
    return _n;
 }
 
-char* xStrdup_(const char* str) {
+char* xStrdup(const char* str) {
    char* data = strdup(str);
    if (!data) {
       fail();
--- a/Affinity.c
+++ b/Affinity.c
@@ -6,6 +6,8 @@ Released under the GNU GPL, see the COPYING file
 in the source distribution for its full text.
 */
 
+#include "config.h"
+
 #include "Affinity.h"
 
 #include <stdlib.h>
--- a/AvailableMetersPanel.h
+++ b/AvailableMetersPanel.h
@@ -7,10 +7,10 @@ Released under the GNU GPL, see the COPYING file
 in the source distribution for its full text.
 */
 
-#include "Settings.h"
 #include "Panel.h"
-#include "ScreenManager.h"
 #include "ProcessList.h"
+#include "ScreenManager.h"
+#include "Settings.h"
 
 typedef struct AvailableMetersPanel_ {
    Panel super;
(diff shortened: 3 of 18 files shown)


BenBE:
Could someone have a look at this `configure.ac` issue? Can't seem to get it work that it ignores the `-Wnull-dereference` flag, if unsupported by the compiler.

natoscott:
I personally think this is becoming quite invasive for very marginal gain - the FORMAT and NORETURN macros I'm OK with, but those NULL macros *everywhere* is making my eyes bleed, and IMO doesn't really help.  Coverity, cppcheck, et al do a fine job of checking these things without the noise.  I reckon we can rely on static checkers for most of what's proposed here - could you rework it to use just FORMAT and NORETURN? (and maybe not even the latter TBH - does this really help us to prevent/solve bugs?)

BenBE:
This patch already only adds `ATTR_NONNULL`, all functions with `ATTR_NORETURN` and `ATTR_FORMAT` had them before this patch.

Regarding bugs: While adding the specifiers I started off with all functions marked `ATTR_NONNULL_ALL` and got several warnings about functions where `NULL` was passed in explicitly. After checking these functions I found all call sites where okay in passing `NULL` as the callee was specialcasing a `NULL` argument. For one function I noticed that it allowed the first OR the second+third argument to be `NULL`, but not all.

Thus regarding effectiveness: While we don't currently have `NULL` pointer issues, it quite helps to have these annotated, as it shows the contract of the function. Also: Most functions these attributes were added to do not have a `NULL` check as part of the function body; thus hard-crashing at runtime.

What can be done is putting these attributes into the same line as the actual declaration, which should reduce the visual noise a bit.

natoscott:
> This patch already only adds `ATTR_NONNULL`, all functions with `ATTR_NORETURN` and `ATTR_FORMAT` had them before this patch.

Yep, understood.  Judicious, strategic use of attributes on key functions is fine and helpful IMO, but no more than that.

> Regarding bugs: [...]

I think these are issues that static code checkers would find anyway.

> Thus regarding effectiveness: While we don't currently have (obvious) `NULL` pointer issues, it quite helps to have these annotated, as it shows the contract of the function.

IMO the signal-to-noise ratio is not in favour of any of the proposed NULL annotations.

> What can be done is putting these attributes into the same line as the actual declaration, which should reduce the visual noise a bit.

That still seems to me like its going to be alot of noise for very little gain.  The static checker tools are able to verify these same problem scenarios (and many more) without the code changes.  Its a NAK for me - lets leave it at the simpler level of annotation that's already been merged.

BenBE:
NP. Thus let's drop https://github.com/htop-dev/htop/pull/156/commits/e5c987de9ac6b076287a590b6884888918e5726c.

Remains the question of https://github.com/htop-dev/htop/pull/156/commits/8e7ada4bc632abe9be8d49d3eabc1c7cbcc2abb3, i.e. if we enable the compiler checks/warnings (independent of `ATTR_NONNULL`)?

natoscott:
> Remains the question of [8e7ada4](https://github.com/htop-dev/htop/commit/8e7ada4bc632abe9be8d49d3eabc1c7cbcc2abb3), i.e. if we enable the compiler checks/warnings

That one's OK by me - @cgzones ?

cgzones:
8e7ada4 looks good.

natoscott:
@BenBE can you update this to resolve the conflict on XAlloc.c (both this and #155 seem to have a conflict there?) - thanks!

BenBE:
> @BenBE can you update this to resolve the conflict on XAlloc.c (both this and #155 seem to have a conflict there?) - thanks!

Should be resolved. The conflict originated from some recent work in `XAlloc.[ch]` on `master` and as #156 is based on #155 (as is #161) all these 3 PRs have that conflict. For #161 there was some more work to do to accomodate for some new call sites.

@natoscott Would be nice if you could review+merge these 3 PRs ASAP. The next PR which is kinda urgend after receiving a rebase is #160 as it contains quite a few changes to track when rebasing; though for #160 I'll wait until #155, #156 & #161 are merged until I do the next rebase there.

XAlloc.c
```diff
@@ -39,6 +41,32 @@ void* xRealloc(void* ptr, size_t size) {
...
 }
 
+int xAsprintf(char** strp, const char* fmt, ...) {
+   va_list vl;
+   va_start(vl, fmt);
+   int _r = vasprintf(strp, fmt, vl);
+   va_end(vl);
+
+   if (_r < 0) {
+      fail();
```
cgzones, line 51: I do not like these `fail()` calls:
In case of an allocation error (or in the` xSnprintf()` case a too small buffer) we just exit cleanly, with no message or hint for the user **why** and no information for us developers **where**.
See also #115

BenBE: Replacing this call by some error handler shouldn't be an issue.

With the current base for this PR there's only `fail()` available as a sensible place to call, but replacing it with the functions introduced in #115 should be fine.

I'd prefer not touching this PR right now, as bugfixes remarked on in #115 are already addressed in #155 which this PR is based upon.

PR #157, opened 2020-09-18
minor fixes for Solaris 11

senjan:
I hit two issues when running the current htop on Solaris 11.4:
1. it crashes (due the NULL pointer dereference) in SolarisProcessList_scanZfsArcstats() because the following kstats are not present in Solaris 11.4:

    - mfu_size
    - mru_size
    - anon_size
    - hdr_size
    - other_size

    They are missing at least also from recent Solaris 11.3 and Solaris 10 1/13 (aka Update 11). I suppose they may exist in some OpenSolaris forks so I think we can keep them in the code. Unfortunately it looks like there are no equivalents in Solaris we can use instead of these or to calculate them.  

2. When running in a Non-Global Zone, htop does not display any memory used. It looks to me that we should use "freemem" kstat rather than "pageslocked" as the latter is always 0 in the zone.


cgzones (approved):

cgzones:
Thanks, applied.

Issue #154, opened 2020-09-18
Discussion: Replace function-like macros by inline functions?


BenBE:
While working my way through several headers I noticed quite a few `#define`s that better should be inline functions.

This suggestion does not cover the various cast helpers, but things like e.g. `Settings_cpuId` or `xSnprintf`.

Detecting support for LTO in configure.ac we even could implement all those function outside the header files without loosing any performance.

cgzones:
Me preference is also using type-safe (inline) function than macros (except generic types, like in `MAXIMUM`, are needed or the code location should not change, e.g. `__FILE__` for debugging).
The second point affects `xSnprintf`, when implemented like in #115.

BenBE:
Regarding xSnprintf I just noticed that the macro is duplicated in both `XAlloc.h` and `XAlloc.c`. Patch to remove the duplication incoming.

cgzones:
Considering the function `String_startsWith()` from *StringUtils.h* I see four ways to implement it:

**I (preprocessor macro)**
*StringUtils.h*:
```
#define String_startsWith(s, match) (strncmp((s),(match),strlen(match)) == 0)
```

**II (header inline function)**
*StringUtils.h*:
```
static inline bool String_startsWith(const char *s, const char *match) {
    return strncmp(s, match, strlen(match)) == 0;
}
```

**III (external inline function)**
*StringUtils.h*:
```
bool String_startsWith(const char *s, const char *match);
```
*StringUtils.c*:
```
extern inline bool String_startsWith(const char *s, const char *match) {
    return strncmp(s, match, strlen(match)) == 0;
}
```

**IV (normal function)**
*StringUtils.h*:
```
bool String_startsWith(const char *s, const char *match);
```
*StringUtils.c*:
```
bool String_startsWith(const char *s, const char *match) {
    return strncmp(s, match, strlen(match)) == 0;
}
```

IV probably depends on using LTO/IPO to get inlined.

BenBE:
I'd prefer `IV` or `II`. Not sure about the semantics of `III`* though. And `I` should be avoided due to type safety issues.

*The gist I got from looking it up seems like an invitation to yourself for a hell lot of linking issues …

cgzones:
Mentioned in https://github.com/htop-dev/htop/blob/master/docs/styleguide.md#symbol-exports

PR #160, opened 2020-09-19
Cleanup includes

BenBE:
This updates all source files to mention exactly those includes required for compiling.

Tested with IWYU 0.12 + clang 9. Please report issues with other setups.

NB: IWYU sometimes has issues with finding certain declarations and reports compile issues in these cases. If plain `clang` compiles then this is an issue with IWYU. Such issues can usually be ignored.

FWIW: IWYU sees functions used inside defines as being used where the macro is used, not where the define is declared. This is a good reason to adopt #154 to avoid this issue.

As I'm sure there will be compile issues on some platforms with this as is I'm just drafting this for now.

### DIFF
--- a/Action.c
+++ b/Action.c
@@ -8,28 +8,32 @@ in the source distribution for its full text.
 #include "config.h"
 
 #include "Action.h"
+
+#include <pwd.h>
+#include <stdbool.h>
+#include <stdlib.h>
+
 #include "Affinity.h"
 #include "AffinityPanel.h"
 #include "CategoriesPanel.h"
+#include "CommandScreen.h"
 #include "CRT.h"
 #include "EnvScreen.h"
-#include "CommandScreen.h"
+#include "FunctionBar.h"
+#include "IncSet.h"
+#include "InfoScreen.h"
+#include "ListItem.h"
+#include "Macros.h"
 #include "MainPanel.h"
 #include "OpenFilesScreen.h"
 #include "Process.h"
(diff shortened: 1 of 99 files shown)


BenBE:
The order of includes I established in this PR is:

1. `"config.h"` with a IWYU pragma to always keep it (required for conditional defines and certain system macros)
2. header for the file being compiled
3. all system headers in alphabetical order, sub directories at the end
4. all our headers in alphabetical order, sub directories at the end
5. all conditionally included headers, i.e. checked by `#if defined(…)`

AFTER the block of includes two blank lines should be present before anything else.

For header files item 2 is simply skipped.

**Important note**: For functions used as part of macro defines you must add an IWYU pragma to keep these headers. Otherwise IWYU will suggest to remove them as unused AND ask each and every user of a define to include those headers. This is BTW another good reason to avoid function-like defines in headers and replace them as inline functions.

natoscott:
> The order of includes I established in this PR is:
> [...]

LGTM.  The 'run_iwyu.sh' script looks like it uses bash-isms - perhaps it should use a /bin/bash shebang line instead of /bin/sh?  (the run_valgrind script has the same issue, so I'm gonna ignore it for now - feel free to follow up though)

> [...]
> This is BTW another good reason to avoid function-like defines in headers and replace them as inline functions.

Hmm - that sounds more like a bug in IWYU than a good reason to do anything, TBH.

PR #161, opened 2020-09-19
Xsnprintf to stringutils

BenBE:
This is a follow-up patch after #155 as asked for by @fasterit …

I'd appreciate it if we could do this one after #155, #156 and #160.


StringUtils.c
```diff
@@ -140,3 +143,37 @@ char* String_readLine(FILE* fd) {
       at = buffer + bufSize - step;
    }
 }
+
+int xAsprintf(char** strp, const char* fmt, ...) {
+   va_list vl;
+   va_start(vl, fmt);
+   int _r = vasprintf(strp, fmt, vl);
```
cgzones, line 150: Please remove the leading underscore. (Originally `xAsprintf` was a macro, so it was about avoiding shadowing)

StringUtils.c
```diff
@@ -140,3 +143,37 @@ char* String_readLine(FILE* fd) {
...
+      fail();
+   }
+
+   return _r;
+}
+
+int xSnprintf(char* buf, int len, const char* fmt, ...) {
+   va_list vl;
+   va_start(vl, fmt);
+   int _n=vsnprintf(buf, len, fmt, vl);
```
cgzones, line 163: Please remove the leading underscore (and add spaces around `=`). (Originally `xAsprintf` was a macro, so it was about avoiding shadowing)

StringUtils.c
```diff
@@ -140,3 +143,37 @@ char* String_readLine(FILE* fd) {
...
+   va_end(vl);
+
+   if (!(_n > -1 && _n < len)) {
```
cgzones, line 166: Please simplify this condition to: `if (_n < 0 || _n >= len)`

cgzones:
Personally I do not like the names `String_asprintf`, `String snprintf` and `String_strdup`, because it might not be obvious what the differences to the standard functions are.
The **x** prefix for allocation wrappers is somewhat common to signal a checked or garbage-collected allocation, so `xasprintf` might be easier translated into *checked asprintf*.

But if @fasterit or @natoscott is fine with the naming, I'll go along.

fasterit:
> Personally I do not like the names `String_asprintf`, `String snprintf` and `String_strdup`, because it might not be obvious what the differences to the standard functions are.
> The **x** prefix for allocation wrappers is somewhat common to signal a checked or garbage-collected allocation, so `xasprintf` might be easier translated into _checked asprintf_.
> 
> But if @fasterit or @natoscott is fine with the naming, I'll go along.

How about renaming `StringUtils.{h|c}` to `XUtils.{h|c}` and joining the remainder of `XAlloc.{h|c}` into this (i.e. `XAlloc.{h|c}` __files__ (not functions) cease to exist)?
This way we could easily stay with naming functions xSnprintf and alike. That means this PR has much less noise where the functions are used, just cleaned and more logical includes.

fasterit:
>     4. Join `xSnprintf`, `xAsprintf`, `xStrdup` and `StringUtils.[ch]` into new `XUtils.[ch]`, but keep current names

My favorite.
/DLange

BenBE:
Could someone please check the deepcode issue? I don't understand what TF it want's to be done …

XUtils.c
```diff
@@ -5,14 +5,55 @@ Released under the GNU GPLv2, see the COPYING file
...
 #include <string.h>
 #include <strings.h>
-#include <stdlib.h>
+#include <unistd.h>
+
+#include "RichString.h"
+
+
+void fail() {
+   curs_set(1);
```
cgzones, line 22: maybe use `CRT_done()` ? (includes `curs_set(1);` and `endwin();`)

cgzones (commented):
LGTM

cgzones (approved):

fasterit:
Merged, thank you very much. Much appreciated cleanup work!

PR #162, opened 2020-09-19
HTOP in a Docker Container

snpranav:
Created a Dockerfile that creates a docker image (<=3MB) which can run htop.

I've created a README inside the docker folder (along with the Dockerfile) which provides instructions on how to run htop in a docker container.
Containerizing htop makes htop more cross compatible on many operating systems and devices that can run Docker containers.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,7 @@
+FROM alpine:3.12
+MAINTAINER Pranav Shikarpur <contact@snpranav.com>
+
+RUN apk --update --no-cache add htop && \
+	rm -rf /var/lib/apt/lists/*
+
+ENTRYPOINT ["htop"]
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.12
+FROM alpine:edge
 MAINTAINER Pranav Shikarpur <contact@snpranav.com>
 
 RUN apk --update --no-cache add htop && \
--- a/docker/README.md
+++ b/docker/README.md
@@ -26,4 +26,3 @@ For example, to run version 3.0.2
 ```sh
 docker run --rm -it --pid=host snpranav/htop:3.0.2
 ```
-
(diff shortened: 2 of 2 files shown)


BenBE:
What's the use-case running `htop` in an empty container? I doubt there's much to see …

snpranav:
Thanks for responding @BenBE 
Although I agree that there are various package managers that provide easy installation of htop, if a person would like the latest version of htop and chooses to builds from source, then they might encounter some issues as they would require some dependencies that need to be installed (like GCC, make, etc).

To answer your question, it's technically not an empty container. If you notice in the [README](https://github.com/snpranav/htop/blob/master/docker/README.md) in the docker folder, I've used a flag `--pid=host` in the instructions on how to run the container. This flag displays the processes running on the host machine in the docker container. Furthermore, from my tests, the **functionality of htop in the container is the same** as running htop from your host machine.

The image primarily improves cross-platform support as a user doesn't need to recreate an environment on their machines to build htop from source. I believe there are various developers (like myself) that would benefit from having a docker image - making htop more cross-compatible.

BenBE:
FWIW: I'm not a big fan of Docker (and most containerization stuff in general).

But apart from this: If this patch was upstreamed it would be wise if the official builds of those htop docker images where Docker-hubbed on some htop-related namespace, thus preferrably `htop-dev/htop` for the image name. Changing this in the README is not a big deal and can easily be changed. The bigger question would be is it worth the maintainance effort to provide proper such image builds as part of the release process (Can be automated via CI/CD probably). I think the "if" is something for others to decide …

Remains the isolation issue: AFAIK does Docker base on LXC, thus there's a bit of kernel namespacing going on to separate things from the host system. This can probably be done for the Process Namespace (cf. your README), but there are other information (libsensors, libnl3/RTNETLINK, vserver stuff) that may be out of sync inside the container (Not sure on the details). Thus for the docker version there might be incentives to disable some of the features, that do not work nicely with containers.

But let's see what others think about this general idea.

fasterit:
I don't see any use case for this either. @ginggs would you want to use docker for testing somewhere?

ginggs:
@fasterit No, I have no use for it at this stage.

PR #164, opened 2020-09-20
Fix minor regression in number highlighting

### LINKED ISSUE #163: Regression overflow of VIRT if ~100G
The human readable sizes were displayed in base 1000 up to 3.0.2 and are now (master) displayed in base 1024 – is that right?

If the value of VIRT (vitual memory) is between 100GB and 100GiB it should be displayed as "100G" or "98.5G" but they are now displayed as "100.2G" and in 3.0.2 they were displayed as "98G". The new format results in a visual overflow. -> Picture

Above: htop 3.0.2
Below: master branch
Bildschirmfoto

The origin is this line: https://github.com/htop-dev/htop/blob/5ea13e7ea9ca19130d5114a745d337b052cf2199/Process.c#L96

BenBE:
Fixes #163

### DIFF
--- a/Process.c
+++ b/Process.c
@@ -93,7 +93,7 @@ void Process_humanNumber(RichString* str, unsigned long number, bool coloring) {
       number %= 1000;
       len = snprintf(buffer, 10, "%03luM ", number);
       RichString_appendn(str, processMegabytesColor, buffer, len);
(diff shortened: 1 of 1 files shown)


cgzones (approved):

cgzones:
Thanks, applied.

PR #165, opened 2020-09-21
Drop configure options

cgzones:
Drop configure options, which have no compile time dependency and are only checked at runtime if specific process columns are selected.

Dropping these reduces the maintenance cost.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,11 +49,11 @@ jobs:
     - name: Bootstrap
       run: ./autogen.sh
     - name: Configure
-      run: ./configure --enable-werror --enable-openvz --enable-cgroup --enable-vserver --enable-ancient-vserver --enable-taskstats --enable-unicode --enable-hwloc --enable-setuid --enable-delayacct --enable-sensors
+      run: ./configure --enable-werror --enable-openvz --enable-vserver --enable-ancient-vserver --enable-taskstats --enable-unicode --enable-hwloc --enable-setuid --enable-delayacct --enable-sensors
     - name: Build
       run: make -k
     - name: Distcheck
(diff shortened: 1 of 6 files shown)


BenBE:
I don't see `HAVE_ANCIENT_VSERVER` being used anywhere. What does it do? Do we still need it?

cgzones:
https://github.com/htop-dev/htop/blob/e0e5997c53ca9b61171d01b5a30c40d915622de2/linux/LinuxProcessList.c#L561-L569

During a quick search I was unable to find any information when/if `s_context` got deprecated/removed.

BenBE:
That code seems to have been introduced in a5dfaa23 … Quite ancient indeed!

[Looking around](https://en.wikipedia.org/wiki/Linux-VServer) it seems this is related to the gr-security kernel patches. As the code is minimal (if it's really just this one `#ifdef`) I'd prefer to just drop both VServer related configure options and compile things unconditionally.

cgzones:
Debian enables all four options: https://salsa.debian.org/debian/htop/-/blob/master/debian/rules
archlinux enables OpenVZ, cgroup and vserver: https://github.com/archlinux/svntogit-packages/blob/packages/htop/trunk/PKGBUILD
Fedora enables all four options: https://src.fedoraproject.org/rpms/htop/blob/master/f/htop.spec

.github/workflows/ci.yml
```diff
@@ -49,11 +49,11 @@ jobs:
     - name: Bootstrap
       run: ./autogen.sh
     - name: Configure
-      run: ./configure --enable-werror --enable-openvz --enable-cgroup --enable-vserver --enable-ancient-vserver --enable-taskstats --enable-unicode --enable-linux-affinity --enable-hwloc --enable-setuid --enable-delayacct
+      run: ./configure --enable-werror --enable-ancient-vserver --enable-unicode --enable-linux-affinity --enable-hwloc --enable-setuid --enable-delayacct
```
BenBE, line 52: What about dropping `--enable-ancient-vserver` defaulting to always on?

BenBE (approved):

PR #167, opened 2020-09-21
Use checked allocation wrappers

cgzones:


linux/LinuxProcessList.c
```diff
@@ -89,15 +89,15 @@ static void LinuxProcessList_initTtyDrivers(LinuxProcessList* this) {
    }
    int numDrivers = 0;
    int allocd = 10;
-   ttyDrivers = malloc(sizeof(TtyDriver) * allocd);
+   ttyDrivers = xMalloc(sizeof(TtyDriver) * allocd);
```
BenBE, line 92: `xCalloc`?

BenBE (changes requested):

linux/LinuxProcessList.c
cgzones: Why? The variable might be realloc'ed later, so the code must not assume a zero'ed memory anyway.

BenBE: After closer inspection ack. I usually prefer `calloc` if you can match the pattern `malloc(numelements * size)` as using `calloc` in these cases avoids potential integer overflow issues (in particular if the `numelements` is user controlled. As that's not the case and the size is reasonably small (& constant), `xMalloc` here is fine. Mostly marked for review if the other option would have been more suitable.

PR #170, opened 2020-09-22
Read CPU frequency from sysfs by default

cgzones:
Use the more portable sysfs node /sys/devices/system/cpu/cpuX/cpufreq/scaling_cur_freq
to get the CPU frequency.
In case of an error fall back to /proc/cpuinfo .

Also use a fixed width of 4 for the frequency to avoid position jumps
in case the frequency moves in the range 900-1100 MHz.

Closes: #132


BenBE:
How much trouble would it be to make the string as wide as the CPU's max frequency?
(Even it 10+GHz CPUs are probably still off for a while)

cgzones:
I think (real) 10 GHz CPU are physically impossible.


(And even if the reported frequency is >= 10 GHz, there is no crash or severe graphical glitch.)

PR #179, opened 2020-09-24
Add eight-column CPU meters

multiplexd:
This series of commits refactors the CPU meter code to call into common code for multi-column meters from wrappers instead of inferring the number of columns from the meter name, adds the four-column meters to non-Linux platform code (which was only added to Linux when that code was added originally in 7fdd8d3!), and adds a series of eight-column CPU meters.

I occasionally work on machines with a very high (>32) core count, and displaying all the CPU meters in eight columns (between the left and right four-column meters) isn't dense enough to leave enough screen space to show a useful amount of the process tree. I've had a patch to fix this in my local tree for a while now, so hopefully this is useful for other people.


cgzones:
This looks great!

Questions:

**I.**
Should we update the initial creation of meters from
https://github.com/htop-dev/htop/blob/18b1e9fba95ce50d47c2199920f6727f26de4063/Settings.c#L67-L80

to something like

```
if (this->cpuCount > 16) {
      this->columns[0].names[0] = xStrdup("LeftCPUs4");
      this->columns[0].modes[0] = BAR_METERMODE;
      this->columns[1].names[r] = xStrdup("RightCPUs4");
      this->columns[1].modes[r++] = BAR_METERMODE;
} else ....
```
?

**II.**
(Not caused by this changes.)
The CPUMeter descriptions are not accurate, e.g. with 32 cores the `CPUs (1-4/8)` meter displays the first 16 cores in 4 rows with 4 columns and the `CPUs (5-8/8)` meter displays the second 16 cores in 4 rows with 4 columns.
Maybe `CPUs (first ½ / 4 colums)` and `CPUs (second ½ / 4 columns)`?

cgzones (approved):

multiplexd:
> *I.* Should we update the initial creation of meters...

That's something which could be considered, yes, though I don't personally have any particularly strong feelings about this; any time I start htop on a new machine, I apply all my preferred configuration and tweaks from muscle memory :slightly_smiling_face:.

> *II.* The CPUMeter descriptions are not accurate...

I think these descriptions are referring to columns, and not individual cores; e.g. `CPUs (1-4/8)` means "meter *columns* one to four, out of eight in total`. There's no harm in relabelling things for clarity, though I'd personally avoid using multi-byte code points which may not render on non-Unicode terminals.

BenBE:
What about `CPU Meter (8 cols, part 1 of 4)` …

cgzones:
>> I. Should we update the initial creation of meters...

>That's something which could be considered, yes, though I don't personally have any particularly strong feelings about this; any time I start htop on a new machine, I apply all my preferred configuration and tweaks from muscle memory slightly_smiling_face.

As the number of columns one wants to use depends also strongly on the terminal width, the current default settings are probably good enough.

>> II. The CPUMeter descriptions are not accurate...

> What about CPU Meter (8 cols, part 1 of 4) …

I'll prepare a patch.

Thanks for your contribution.

PR #180, opened 2020-09-24
Consider the ZFS ARC as cache on Linux

multiplexd:
This is based on a [patch](https://gist.github.com/edef1c/73cc02efcdad39fbcc15f577dbc1922a) which @edef1c and I wrote against htop 2.x, which I've adapted for htop 3.x to use the existing ZFS support code which was merged in 3.0.

The Linux kernel doesn't report ZFS ARC usage as cache memory, which means that the ARC is included in the total used system memory. This means that tools such as `free(1)` will misreport the total memory consumed, as will htop without this patch. With this commit, if ZFS is detected, then htop will adjust the kernel-reported overall used/cache memory using the ARC size read from procfs.

### DIFF
--- a/linux/Platform.c
+++ b/linux/Platform.c
@@ -202,6 +202,8 @@ double Platform_setCPUValues(Meter* this, int cpu) {
 
 void Platform_setMemoryValues(Meter* this) {
    ProcessList* pl = (ProcessList*) this->pl;
+   LinuxProcessList* lpl = (LinuxProcessList*) this->pl;
+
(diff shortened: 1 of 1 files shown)


multiplexd:
The CI failure there looks like the `apt-get update` step on the builder hit a mirror in the middle of syncing with inconsistent state.

multiplexd:
(I've also realised that I've misnamed the source branch here, it should really be `arc-is-cache`. Oops...)

BenBE (approved):
LGTM.

PR #185, opened 2020-09-28
command screen: fill current line when scanning

ryenus:
After pr #67 got merged, @BenBE proposed the option to fill the current line 
while scanning the command, rather than using a separate `memcpy` afterward.
So I gave it a try and [measured the difference][1].

For short commands there's not much difference, but when a command is relatively 
long, e.g., the longest I could find, of a java process, is about 7k+ 
characters, per my test, in the best case, it's about 30 vs 28 (ms) on average
before and after, So the copy-on-scan approach won with a little margin.

[1]: https://github.com/ryenus/htop/commit/d96a931de437ea027d51af393523b21d4b2bea25

### DIFF
--- a/CommandScreen.c
+++ b/CommandScreen.c
@@ -12,33 +12,33 @@
 #include <unistd.h>
 
 
-static void CommandScreen_addLine(InfoScreen* this, char* line, const char* p, int line_offset, int len) {
-   memcpy(line, p - line_offset, len);
-   line[len] = '\0';
-   InfoScreen_addLine(this, line);
-}
-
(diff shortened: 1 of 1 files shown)


ryenus:
@cgzones would you please take a look?

Nevertheless the numbers are machine specific, YMMV.

cgzones (approved):
LGTM

BenBE (approved):
LGTM.

Don't quite like the back-addressing in `p - line_offset`, but overall it's okay nonetheless.

PR #188, opened 2020-09-28
Add security attribute process column

cgzones:


XAlloc.h
```diff
@@ -35,4 +35,10 @@ void* xRealloc(void* ptr, size_t size);
 
 char* xStrdup_(const char* str) ATTR_NONNULL;
 
+static inline void* xFree(void* ptr) {
+   if (ptr)
+      free(ptr);
```
BenBE, line 40: `free(NULL);` is no-op as per the C standard, no check needed.
Cf. https://stackoverflow.com/a/1938758

Runtimes that crash here are broken in the first place and should be fixed instead.

linux/LinuxProcess.c
```diff
@@ -289,6 +291,7 @@ void LinuxProcess_writeField(Process* this, RichString* str, ProcessField field)
          attr |= A_BOLD;
       xSnprintf(buffer, n, "%5lu ", lp->ctxt_diff);
       break;
+   case SECATTR: snprintf(buffer, n, "%-30s   ", lp->secattr ? lp->secattr : "?"); break;
```
BenBE, line 294: Use `xSnprintf`/`String_snprintf` as the checked variant?

linux/LinuxProcessList.c
```diff
@@ -626,6 +626,30 @@ static void LinuxProcessList_readCtxtData(LinuxProcess* process, const char* dir
    process->ctxt_total = ctxt;
 }
 
+static void LinuxProcessList_readSecattrData(LinuxProcess* process, const char* dirname, const char* name) {
+   char filename[MAX_NAME+1];
+   xSnprintf(filename, MAX_NAME, "%s/%s/attr/current", dirname, name);
```
BenBE, line 631: Use `sizeof(filename)`, no need to account for that final `\0`, already done by `xSnprintf`.

XAlloc.h
```diff
@@ -35,4 +35,10 @@ void* xRealloc(void* ptr, size_t size);
 
 char* xStrdup_(const char* str) ATTR_NONNULL;
 
+static inline void* xFree(void* ptr) {
```
BenBE, line 38: I don't quite overall like this function, but see where it comes from. The classical idiom is writing `free(x); x = NULL;`, thus hiding the `NULL` in the function call is not the best style IMHO.

XAlloc.h
cgzones: dropped check

linux/LinuxProcessList.c
cgzones: yep, fixed

linux/LinuxProcess.c
cgzones: The security attribute can and will be quite long (especially with SELinux).
AppArmor example: `/usr/sbin/cups-browsed (enforce)`

So in case it exceeds the buffer limit, htop should not stop.

BenBE (approved):

PR #190, opened 2020-09-28
Adjust colors

cgzones:
- do not reverse CPU steal and guest in monochrome
- black on black in Light Terminal is not visible, use blue on black
- white on blue in Light Terminal is display as blue on black, use
  yellow on black
- re-draw FunctionBar after color change

### DIFF
--- a/CRT.c
+++ b/CRT.c
@@ -39,7 +39,7 @@ in the source distribution for its full text.
 #define ColorPairGrayBlack ColorPair(Magenta,Magenta)
 #define ColorIndexGrayBlack ColorIndex(Magenta,Magenta)
 
-const char *CRT_treeStrAscii[TREE_STR_COUNT] = {
+static const char *const CRT_treeStrAscii[TREE_STR_COUNT] = {
    "-", // TREE_STR_HORZ
    "|", // TREE_STR_VERT
    "`", // TREE_STR_RTEE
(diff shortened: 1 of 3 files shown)


CRT.c
```diff
@@ -203,8 +203,8 @@ int CRT_colorSchemes[LAST_COLORSCHEME][LAST_COLORELEMENT] = {
       [CPU_IOWAIT] = A_NORMAL,
       [CPU_IRQ] = A_BOLD,
       [CPU_SOFTIRQ] = A_BOLD,
-      [CPU_STEAL] = A_REVERSE,
-      [CPU_GUEST] = A_REVERSE,
+      [CPU_STEAL] = A_BOLD,
+      [CPU_GUEST] = A_BOLD,
```
BenBE, line 207: What about `A_DIM` here instead?

BenBE (commented):
LGTM.

CRT.c
cgzones: seems reasonable; done

deviant:
> - black on black in Light Terminal is not visible, use blue on black
> - white on blue in Light Terminal is display as blue on black, use
yellow on black

Yes, because it's meant for a *light* terminal, i.e. where the background is light. It's in the name! This commit actually makes things less readable than they were before. The default colour scheme uses a bunch of bright colours (e.g. cyan) that are difficult to read on a light background, which is why this one exists. It's not broken. (Or, wasn't before this commit, at any rate.)

htop 3.0.2:





htop 3.0.4:

PR #196, opened 2020-09-29
Process.{h,c}: Use integer types that are more portable

mfwitten:
When building on a 32-bit system, the compiler warned that the following line uses a constant whose value is the overflow result of a compile-time computation:

    Process.c (line 109):   } else if (number < 10000 * ONE_M) {

Namely, this constant expression:

    10000 * ONE_M

was intended to produce the following value:

    10485760000

However, the result overflowed to produce:

     1895825408

The reason for this overflow is as follows:

* The macros are expanded:

      10000 * (ONE_K * ONE_K)
      10000 * (1024L * 1024L)

* The untyped constant expression "10000" is typed:

      10000U * (1024L * 1024L)

* The parenthesized expression is evaluated:

      10000U * (1048576L)

* The left operand ("10000U") is converted:

      10000L * (1048576L)

    Unbound by integer sizes, that last multiplication would produce the following value:

      10485760000

    However, on a 32-bit machine, where a long is 32 bits (really 31 bits when talking about positive numbers), the maximum value that can be computed is 2**31-1:

      2147483647

    Consequently, the computation overflows.

* The compiler produces a long int value that is the the result of overflow (10485760000 % 2**31):

      1895825408L

    Actually, I think this overflow is implementation-defined, so it's not even a portable description of what happens.

The solution is to use a long long int (or, even better, an unsigned long long int) type for the constant expression; the C standard mandates a sufficiently large maximum value for such types.

Hence, the following change is made to the bad line:

    -   } else if (number < 10000 * ONE_M) {
    +   } else if (number < 10000ULL * ONE_M) {

However, the whole line is now patently silly, because the variable "number" is typed "unsigned long", and so it will always be less than the constant expression (the compiler will warn about this, too).

Hence, "number" must be typed "unsigned long long"; however, this necessitates changing all of the string formats from something like "%lu" to something like "%llu".

***Et voila!* This commit is born.**

Then, for the sake of completeness, the declared types of the
constant-expression macros are updated:

* `ONE_K` is made unsigned (a `UL` instead of `L`)
* `ONE_T` is computed by introducing `1ULL *`
* Similar changes are made for `ONE_DECIMAL_{K,T}`

Also, a non-portable overflow-conversion to a signed value
has been replaced with a portable comparison:

    -   if ((long long) number == -1LL) {
    +   if (number == ULLONG_MAX) {

It might be worth reviewing the rest of the code for other cases where overflows are not handled correctly; even at runtime, it's often necessary to check for overflow unless such behavior is expected (especially for signed integer values, for which overflow has implementation-defined behavior).

### DIFF
--- a/Process.h
+++ b/Process.h
@@ -138,19 +138,19 @@ typedef struct ProcessClass_ {
 #define Process_sortState(state) ((state) == 'I' ? 0x100 : (state))
 
 
-#define ONE_K 1024L
+#define ONE_K 1024UL
 #define ONE_M (ONE_K * ONE_K)
 #define ONE_G (ONE_M * ONE_K)
-#define ONE_T ((long long)ONE_G * ONE_K)
+#define ONE_T (1ULL * ONE_G * ONE_K)
 
-#define ONE_DECIMAL_K 1000L
+#define ONE_DECIMAL_K 1000UL
 #define ONE_DECIMAL_M (ONE_DECIMAL_K * ONE_DECIMAL_K)
 #define ONE_DECIMAL_G (ONE_DECIMAL_M * ONE_DECIMAL_K)
-#define ONE_DECIMAL_T ((long long)ONE_DECIMAL_G * ONE_DECIMAL_K)
+#define ONE_DECIMAL_T (1ULL * ONE_DECIMAL_G * ONE_DECIMAL_K)
 
 void Process_setupColumnWidths(void);
 
-void Process_humanNumber(RichString* str, unsigned long number, bool coloring);
+void Process_humanNumber(RichString* str, unsigned long long number, bool coloring);
 
 void Process_colorNumber(RichString* str, unsigned long long number, bool coloring);
 
--- a/Process.c
+++ b/Process.c
@@ -55,7 +55,7 @@ void Process_setupColumnWidths() {
    xSnprintf(Process_pidFormat, sizeof(Process_pidFormat), "%%%dd ", digits);
 }
 
-void Process_humanNumber(RichString* str, unsigned long number, bool coloring) {
+void Process_humanNumber(RichString* str, unsigned long long number, bool coloring) {
    char buffer[11];
    int len;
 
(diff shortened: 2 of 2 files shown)


Process.h
```diff
@@ -138,19 +138,19 @@ typedef struct ProcessClass_ {
 #define Process_sortState(state) ((state) == 'I' ? 0x100 : (state))
 
 
-#define ONE_K 1024L
+#define ONE_K 1024UL
```
BenBE, line 141: What about consistency and making this `ULL` too?

cgzones (approved):
LGTM :+1:

BenBE (commented):
Actually I'd be even more happy, if we just could use C99 and use 64bit `size_t` for all amounts representing data.

cgzones:
> Actually I'd be even more happy, if we just could use C99 and use 64bit size_t for all amounts representing data.

`long long` is C99: https://en.cppreference.com/w/c/language/arithmetic_types
and `size_t` won't fix the issue on 32-bit.

Process.h
mfwitten: My thinking is that a type should only be as large as necessary or as practical. For instance:

* Use a `short` when you need to cram things into limited memory, and it's large enough to hold the values in question.

* However, a `short` is often promoted to an `int` automatically anyway, so use an `int` (or often `unsigned int`) even if the values you care about can all be represented by a `short`.

Similarly:

* The constant `1024` can be represented by a plain int type; so, I first used `1024U`, because I just wanted to declare the constant to be unsigned.

* However, the C standard actually only guarantees that an `unsigned int` can represent up to at least 2**16-1, which means you'd only be able to calculate safely a constant as big as `63 * ONE_K`, which is impractical; rather than requiring most uses of the multiplier to specify a size (*e.g.,* `64UL * ONE_K`), I figured it would be much more practical just to declare `ONE_K` with such a declaration instead.

Here are some other notes on the matter:

* This makes the declared type of the constant `ONE_M` also `unsigned long`, but that's not as impractical as making `unsigned int` the declared type of the constant `ONE_K`; indeed, I only had to specify `ULL` on one rather large multiplier (`10000ULL * ONE_M`).

* I actually do not like the idea of using `unsigned long long` at all; though I haven't looked, I'm guessing it introduces a not-insignificant overhead to the calculations on a 32-bit machine; it would be much better to check an input (*e.g.,* the value of `number`) for its size, and then pass it to code that uses `unsigned long long` only if necessary.

    Heck, maybe a 32-bit machine will never show numbers big enough to warrant `unsigned long long`, anyway, and so the offending code should just not be built at all on a 32-bit machine. I don't know enough to make that determination, so this commit is a good fix until someone can tailor the code more tightly to each target architecture.

mfwitten:
* The exact-width types (`uint64_t`, etc.) are in general optional but practically required (if a system can support them, then it must).
* The types `uint_least64_t` and `int_fast64_t` (etc.) have been required since C99; these would be preferable, since they don't enforce certain semantics (*e.g.,* a `uint32_t` *must* result in overflows for values that require more than 32 bits to represent, which is behavior that might not be necessary).

Anyway, the key is to really know what your numbers are, or just drop in an arbitrary-precision library and pretend size doesn't matter :-).

cgzones:
In case of `Process_humanNumber()` `float`/`double` as an argument type is an alternative, as this function is about displaying and the maximum significant digits displayed is 5 (`float` has 7).

Issue #26, opened 2020-08-29
add screenshot


passenger0245:
can we have a screenshot in the readme?

fasterit:
Done with #199

Issue #71, opened 2020-09-02
Drop margin around header by default


BenBE:
Although the blank margin around the header can be removed by a configuration setting already I think it's a waste of valueable screen estate to have it enabled by default, as it wastes two rows of your terminal height placing no information. As the density of information of most other parts of htop is quite high this seems a bit counter-intuitive to have this blank area on by default.

eworm-de:
Never understood why the margin is there, so full ack.

fasterit:
In the core team call we have decided to stay with Hisham's design decisions for the time being.
It is easy enough to change for people that want their screen layout as dense as possible.

PR #199, opened 2020-10-01
Add screenshot of htop to readme

nzbart:
Added a basic screenshot of htop in action to the readme so that visitors to the page can quickly get a rough idea about what htop does.

Hopefully addresses #26.


BenBE:
Please use a screenshot without the border around the header as this is no longer the default design.

fasterit:
> Please use a screenshot without the border around the header as this is no longer the default design.

That's wrong, #71 is your request to do so and we have agreed to stick with Hisham's design aesthetics for the time being, so I'll close that one.

PR #203, opened 2020-10-03
limit max screen title length to window width

ryenus:
Applies screen title truncating to all InfoScreen classes.

@cgzones Possible to take a look? Thank you!

### DIFF
--- a/InfoScreen.c
+++ b/InfoScreen.c
@@ -42,14 +42,21 @@ InfoScreen* InfoScreen_done(InfoScreen* this) {
 void InfoScreen_drawTitled(InfoScreen* this, const char* fmt, ...) {
    va_list ap;
    va_start(ap, fmt);
+
+   char* title = xMalloc(COLS + 1);
+   int len = vsnprintf(title, COLS + 1, fmt, ap);
+   if (len > COLS) {
+      memset(&title[COLS - 3], '.', 3);
+   }
+
    attrset(CRT_colors[METER_TEXT]);
    mvhline(0, 0, ' ', COLS);
-   (void) wmove(stdscr, 0, 0);
-   vw_printw(stdscr, fmt, ap);
+   mvwprintw(stdscr, 0, 0, title);
    attrset(CRT_colors[DEFAULT_COLOR]);
    this->display->needsRedraw = true;
    Panel_draw(this->display, true);
    IncSet_drawBar(this->inc);
+   free(title);
    va_end(ap);
 }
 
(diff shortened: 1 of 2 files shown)


InfoScreen.c
```diff
@@ -42,14 +42,19 @@ InfoScreen* InfoScreen_done(InfoScreen* this) {
 void InfoScreen_drawTitled(InfoScreen* this, const char* fmt, ...) {
    va_list ap;
    va_start(ap, fmt);
+
+   char* title = xMalloc(COLS + 1);
+   int len = vsnprintf(title, COLS + 1, fmt, ap);
+   if (len > COLS) memset(&title[COLS - 3], '.', 3);
+
```
BenBE, line 49: Could we try to move this allocation into the lifetime of the window?

InfoScreen.c
```diff
@@ -42,14 +42,19 @@ InfoScreen* InfoScreen_done(InfoScreen* this) {
 void InfoScreen_drawTitled(InfoScreen* this, const char* fmt, ...) {
    va_list ap;
    va_start(ap, fmt);
+
+   char* title = xMalloc(COLS + 1);
+   int len = vsnprintf(title, COLS + 1, fmt, ap);
+   if (len > COLS) memset(&title[COLS - 3], '.', 3);
```
BenBE, line 48: Please avoid braceless if-statements.

InfoScreen.c
cgzones: I think this is not needed:
`InfoScreen_drawTitled` is called in the `_draw` callbacks of the Screen classes.
The `_draw` callback is called (besides the initial one) on window-resize, F5 key and Crtl-L key.
These events are rather rare, so caching seems not worth it and on resize we would to re-compute anyway.

InfoScreen.c
cgzones: I am not sure if we have agreed on a brace-style for one-line branches. @natoscott @ginggs @fasterit 
But at least put the branch into a separate line.

BenBE: The original source had it on multiple lines. ;-)

cgzones: Good point, another topic for a future style guide.

InfoScreen.c
BenBE: Okay. Sounds reasonable.

InfoScreen.c
ryenus: OK, braces added, though I'm fine with both, but having an agreed style guide is great. 👍

cgzones:
Applied, thanks.

PR #214, opened 2020-10-05
Add key to pause process list updates

cgzones:
Process updates are paused/resumed with `Z`.
An info that updates are paused is shown at the end of the function bar (bottom-left).


Action.c
```diff
@@ -491,12 +499,12 @@ static Htop_Reaction actionHelp(State* st) {
...
    attrset(CRT_colors[PROCESS_THREAD]);
-   mvaddstr(16, 32, "threads");
-   mvaddstr(17, 26, "threads");
+   mvaddstr(17, 32, "threads");
+   mvaddstr(18, 26, "threads");
    attrset(CRT_colors[DEFAULT_COLOR]);
 
    attrset(CRT_colors[HELP_BOLD]);
-   mvaddstr(23,0, "Press any key to return.");
+   mvaddstr(24,0, "Press any key to return.");
```
BenBE, line 507: More a long-term issue, but could we _please_ automate these lines?

BenBE (commented):
The many API changes make this PR kinda noisy. But LGTM.

ScreenManager.c
```diff
@@ -101,7 +104,7 @@ static void checkRecalculation(ScreenManager* this, double* oldTime, int* sortTi
    *timedOut = (newTime - *oldTime > this->settings->delay);
    *rescan = *rescan || *timedOut;
    if (newTime < *oldTime) *rescan = true; // clock was adjusted?
-   if (*rescan) {
+   if (*rescan && !this->state->pauseProcessUpdate) {
       *oldTime = newTime;
       ProcessList_scan(pl);
```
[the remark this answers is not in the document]
cgzones: No, obviously we need to redraw if we update the processes, but in case we don't we do not need to set `redraw`.
It is set at
https://github.com/htop-dev/htop/blob/55eafd3b39eabdf60b5196d8c728b6eaf190650d/ScreenManager.c#L235-L238

Action.c
cgzones: Yes, I'll take a look at it.

Issue #216, opened 2020-10-05
A column for LXC and LXD containers


jb-boin:
There was already an issue on the original project about that : https://github.com/hishamhm/htop/issues/328

I did a simple patch to add the feature but it's not clean enough for a PR request at the moment.

I updated the patch to run on htop 3.0.2, tested on Debian Buster with LXC containers (/lxc/$NAME) and Ubuntu 20.04 with LXD containers with LXC 3 (/lxc.payload/$NAME) and also with LXC 4 (/lxc.payload.$NAME) : 
```patch
--- htop-3.0.2.orig/linux/LinuxProcess.c	2020-09-15 08:00:00.000000000 +0200
+++ htop-3.0.2/linux/LinuxProcess.c	2020-10-05 18:12:51.157239834 +0200
@@ -96,6 +96,7 @@
 #endif
 #ifdef HAVE_CGROUP
    [CGROUP] = { .name = "CGROUP", .title = "    CGROUP ", .description = "Which cgroup the process is in", .flags = PROCESS_FLAG_LINUX_CGROUP, },
+   [LXC] = { .name = "LXC", .title = "       LXC ", .description = "Which LXC constainer the process is in", .flags = PROCESS_FLAG_LINUX_CGROUP, },
 #endif
    [OOM] = { .name = "OOM", .title = " OOM ", .description = "OOM (Out-of-Memory) killer score", .flags = PROCESS_FLAG_LINUX_OOM, },
    [IO_PRIORITY] = { .name = "IO_PRIORITY", .title = "IO ", .description = "I/O priority", .flags = PROCESS_FLAG_LINUX_IOPRIO, },
@@ -145,6 +146,7 @@
    Process_done((Process*)cast);
 #ifdef HAVE_CGROUP
    free(this->cgroup);
+   free(this->lxc);
 #endif
    free(this->ttyDevice);
    free(this);
@@ -251,6 +253,7 @@
    #endif
    #ifdef HAVE_CGROUP
    case CGROUP: xSnprintf(buffer, n, "%-10s ", lp->cgroup); break;
+   case LXC:    xSnprintf(buffer, n, "%-10s ", lp->lxc); break;
    #endif
    case OOM: xSnprintf(buffer, n, "%4u ", lp->oom); break;
    case IO_PRIORITY: {
@@ -346,6 +349,8 @@
    #ifdef HAVE_CGROUP
    case CGROUP:
[...]

PR #197, opened 2020-09-30
Handle parsing envID & VPid from process status file

BenBE:
Fixes #55
Fixes #192

PR #217, opened 2020-10-06
New column for LXC containers name using the CGroups

jb-boin:
Fixed the issues listed on https://github.com/htop-dev/htop/issues/216.

I feel like there should be a `break` to exit the while loop parsing each line of the CGroup file if the LXC container name has been found and if the CGroup column is not used (or at least skip the LXC part if both columns are used and the LXC name has been found).
But it seems that the CGroup files are always parsed even if the CGroup column is not used which is not ideal.

### DIFF
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -533,6 +533,33 @@ static void LinuxProcessList_readCGroupFile(LinuxProcess* process, const char* d
       if (!ok) break;
       char* group = strchr(buffer, ':');
       if (!group) break;
+
+      if (!process->lxc) {
+         // Each line have 3 columns separated by a ':', the 3rd column contains the pathname
+         char* groupPathname = strchr(group + 1, ':') + 1;
+         if (groupPathname) {
+            if ( String_startsWith(groupPathname, "/lxc.payload/") || String_startsWith(groupPathname, "/lxc.payload.") ) {
+               // The process is inside a LXC container using CGroup V2 (/lxc.payload.) or a modern CGroup V1 terminology (/lxc.payload/), for a better readability, only the container name is kept
+               char *slashpostion = strchr((groupPathname + 13), '/');
+               if (slashpostion) {
+                  // There is a '/' in the CGroup string (after the initial "/lxc.payload"), a '\0' will truncate the string at this position
+                  groupPathname[(slashpostion - groupPathname)] = '\0';
+               }
+               free(process->lxc);
+               process->lxc = String_trim(groupPathname + 13);
+            } else if (String_startsWith(groupPathname, "/lxc/")) {
+               // The process is inside a LXC container using a legacy CGroup V1 (/lxc/), for a better readability, only the container name is kept
+               char *slashpostion = strchr((groupPathname + 5), '/');
+               if (slashpostion) {
+                  // There is a '/' in the CGroup string (after the initial "/lxc/"), a '\0' will truncate the string at this position
+                  groupPathname[(slashpostion - groupPathname)] = '\0';
+               }
+               free(process->lxc);
+               process->lxc = String_trim(groupPathname + 5);
+            }
+         }
+      }
+
       if (at != output) {
          *at = ';';
          at++;
@@ -542,6 +569,12 @@ static void LinuxProcessList_readCGroupFile(LinuxProcess* process, const char* d
       left -= wrote;
    }
    fclose(file);
+   if (!process->lxc) {
+      // The process is not in a LXC container
+      free(process->lxc);
+      // To show an empty value instead of (null)
+      process->lxc = NULL;
+   }
    free(process->cgroup);
    process->cgroup = xStrdup(output);
 }
--- a/linux/LinuxProcess.c
+++ b/linux/LinuxProcess.c
@@ -97,6 +97,7 @@ ProcessFieldData Process_fields[] = {
 #endif
 #ifdef HAVE_CGROUP
    [CGROUP] = { .name = "CGROUP", .title = "    CGROUP ", .description = "Which cgroup the process is in", .flags = PROCESS_FLAG_LINUX_CGROUP, },
+   [LXC] = { .name = "LXC", .title = "       LXC ", .description = "Which LXC constainer the process is in", .flags = PROCESS_FLAG_LINUX_CGROUP, },
 #endif
    [OOM] = { .name = "OOM", .title = " OOM ", .description = "OOM (Out-of-Memory) killer score", .flags = PROCESS_FLAG_LINUX_OOM, },
    [IO_PRIORITY] = { .name = "IO_PRIORITY", .title = "IO ", .description = "I/O priority", .flags = PROCESS_FLAG_LINUX_IOPRIO, },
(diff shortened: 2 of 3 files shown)


linux/LinuxProcessList.c
```diff
@@ -533,6 +533,33 @@ static void LinuxProcessList_readCGroupFile(LinuxProcess* process, const char* d
       if (!ok) break;
       char* group = strchr(buffer, ':');
       if (!group) break;
+
+      if (!process->lxc) {
```
cgzones, line 537: Does the LXC cgroup never change for a running process? (Cause it's only computed once)

linux/LinuxProcessList.c
```diff
@@ -533,6 +533,33 @@ static void LinuxProcessList_readCGroupFile(LinuxProcess* process, const char* d
...
+         // Each line have 3 columns separated by a ':', the 3rd column contains the pathname
+         char* groupPathname = strchr(group + 1, ':') + 1;
+         if (groupPathname) {
+            if ( String_startsWith(groupPathname, "/lxc.payload/") || String_startsWith(groupPathname, "/lxc.payload.") ) {
+               // The process is inside a LXC container using CGroup V2 (/lxc.payload.) or a modern CGroup V1 terminology (/lxc.payload/), for a better readability, only the container name is kept
+               char *slashpostion = strchr((groupPathname + 13), '/');
+               if (slashpostion) {
+                  // There is a '/' in the CGroup string (after the initial "/lxc.payload"), a '\0' will truncate the string at this position
+                  groupPathname[(slashpostion - groupPathname)] = '\0';
```
cgzones, line 546: Here we are modifying the string behind `groupPathname`, which points into the string `group`, which get written later in line 568 (`int wrote = snprintf(at, left, "%s", group);`).
Do we really want to modify `group`?
Or maybe move the LXC block to the end of the while loop, so any changes to `group` are discarded.

linux/LinuxProcessList.c
```diff
@@ -533,6 +533,33 @@ static void LinuxProcessList_readCGroupFile(LinuxProcess* process, const char* d
...
+               }
+               free(process->lxc);
+               process->lxc = String_trim(groupPathname + 13);
+            } else if (String_startsWith(groupPathname, "/lxc/")) {
+               // The process is inside a LXC container using a legacy CGroup V1 (/lxc/), for a better readability, only the container name is kept
+               char *slashpostion = strchr((groupPathname + 5), '/');
+               if (slashpostion) {
+                  // There is a '/' in the CGroup string (after the initial "/lxc/"), a '\0' will truncate the string at this position
+                  groupPathname[(slashpostion - groupPathname)] = '\0';
```
cgzones, line 555: dito

linux/LinuxProcessList.c
```diff
@@ -542,6 +569,12 @@ static void LinuxProcessList_readCGroupFile(LinuxProcess* process, const char* d
       left -= wrote;
    }
    fclose(file);
+   if (!process->lxc) {
```
cgzones, line 572: This whole block is now a no-op (it is only entered if `process->lxc` is NULL)

cgzones (changes requested):

linux/LinuxProcessList.c
BenBE: Technically it could (changed process namespace), but practically it doesn't.

linux/LinuxProcessList.c
jb-boin: In all fairness, the current implementation of the CGROUP column is fairly useless on modern systems as it will show lengty SystemD CGroups for services started by it that breaks the alignment of columns as the values are too long compared to the column size.


I initially did put my code inside the `LinuxProcessList_readCGroupFile()` function in the middle of the CGROUP column code for simplicity but maybe it could be separated in another function as there probably wont be anybody using both columns at the same time or create two sub function called inside `LinuxProcessList_readCGroupFile()` for each of the columns.

BenBE (changes requested):
I'm not quite happy with the current CGroup->LXC container name code, both from a code cleanliness perspective (compare with e.g. the CTID/VPID patch (#197) and from handling the disambuguity that may arise from the read value.

Here's some example `cgroup` file from a LXC host:
```
# cat /proc/60054/cgroup 
11:cpuset:/lxc.payload/example
10:devices:/lxc.payload/example/init.scope
9:net_cls,net_prio:/lxc.payload/example
8:blkio:/lxc.payload/example
7:perf_event:/lxc.payload/example
6:rdma:/lxc.payload/example
5:pids:/lxc.payload/example/init.scope
4:freezer:/lxc.payload/example
3:cpu,cpuacct:/lxc.payload/example
2:memory:/lxc.payload/example/init.scope
1:name=systemd:/lxc.payload/example/init.scope
0::/lxc.payload/example/init.scope
```

From this example it's quite clear, that that container is probably called `example`. It's parent looks like this:
```
# cat /proc/60051/cgroup 
11:cpuset:/lxc.monitor/example
10:devices:/lxc.monitor/example
9:net_cls,net_prio:/lxc.monitor/example
8:blkio:/lxc.monitor/example
7:perf_event:/lxc.monitor/example
6:rdma:/lxc.monitor/example
5:pids:/lxc.monitor/example
4:freezer:/lxc.monitor/example
3:cpu,cpuacct:/lxc.monitor/example
2:memory:/lxc.monitor/example
1:name=systemd:/lxc.monitor/example
0::/lxc.monitor/example
```

A process of a service started within that container looks like this:
```
# cat /proc/3686687/cgroup 
11:cpuset:/lxc.payload/example
10:devices:/lxc.payload/example/system.slice/cron.service
9:net_cls,net_prio:/lxc.payload/example
8:blkio:/lxc.payload/example
7:perf_event:/lxc.payload/example
6:rdma:/lxc.payload/example
5:pids:/lxc.payload/example/system.slice/cron.service
4:freezer:/lxc.payload/example
3:cpu,cpuacct:/lxc.payload/example
2:memory:/lxc.payload/example/system.slice/cron.service
1:name=systemd:/lxc.payload/example/system.slice/cron.service
0::/lxc.payload/example/system.slice/cron.service
```

What would be good to have here is some consistent parsing of the file, that doesn't randomly select ANY of the cgroup names for further processing, but would be looking for a specific one (e.g. the `pids` namespace). Another bonus could be a new column `SYSTEMD SERVICE` or `SLICE` providing the slice a given process was started under (independent of LXC or not). Parsing code would be essentially the same; just another part of the string being used.

Overall the `LXC` column should provide 3 information:
1. The container name (`pids` cgroup preferred)
2. If it's a container monitor (cf. second example dump) -> `[m]`
3. If names of cgroups for this container are consistent, i.e. add `[i]` or `[!]` if any of the cgroups doesn't point to the name shown

ser:
it would be cool to have it as we have now incus.

BenBE:
@jb-boin Would you mind updating this PR to base off the latest commit from `main`? Also AFAICS there are still several open discussion items that should be addressed.

NB: Regarding the lengthy CGROUP names from systemd there's a compressed cgroup (`CCGROUP`) column. The source for this could be adapted for this feature.

PR #219, opened 2020-10-06
Provide initial tree filter (matching commands)

layderv:
For this PR see: https://github.com/hishamhm/htop/issues/836#issuecomment-480027645

### DIFF
--- a/htop.1.in
+++ b/htop.1.in
@@ -4,7 +4,7 @@ htop \- interactive process viewer
 .SH "SYNOPSIS"
 .LP
 .B htop
-.RB [ \-dChpustv ]
+.RB [ \-dCFhpustv ]
 .SH "DESCRIPTION"
 .LP
 .B htop
@@ -35,6 +35,9 @@ is greater than 100, it is decreased to 100, i.e. 10 seconds.
 Start
 .B htop
 in monochrome mode
+.TP
+\fB\-F \-\-filter=FILTER
+Filter processes by command
 .TP
 \fB\-h \-\-help
 Display a help message and exit
--- a/htop.c
+++ b/htop.c
@@ -39,6 +39,7 @@ static void printHelpFlag(void) {
          "Released under the GNU GPLv2.\n\n"
          "-C --no-color               Use a monochrome color scheme\n"
          "-d --delay=DELAY            Set the delay between updates, in tenths of seconds\n"
+         "-F --filter=FILTER          Show only the commands matching the given filter\n"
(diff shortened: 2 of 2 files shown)


Settings.h
```diff
@@ -3,12 +3,13 @@
 /*
 htop - Settings.h
 (C) 2004-2011 Hisham H. Muhammad
-Released under the GNU GPLv2, see the COPYING file
+Released under the GNU GPL, see the COPYING file
```
fasterit, line 6: rebase to HEAD please

fasterit (changes requested):
I think this should integrate with the Filter functionality (F4 key), so that -F pre-sets a filter that can be cleared / amended in the UI. If that filter is regex, it should be regex in both use cases to not confuse users.
To me just plain string matching would be fine as well.

layderv:
> I think this should integrate with the Filter functionality (F4 key), so that -F pre-sets a filter that can be cleared / amended in the UI. If that filter is regex, it should be regex in both use cases to not confuse users.
> To me just plain string matching would be fine as well.

@fasterit Your comment make sense, this PR implements it. (By "-F pre-sets..." I think you meant "-P" as in the option.)

fasterit:
> @fasterit Your comment make sense, this PR implements it. (By "-F pre-sets..." I think you meant "-P" as in the option.)

No, I meant -F / --filter as that is consistent with the existing Filter (F4) GUI option.

fasterit:
The second iteration looks good except for

- Old license header (work of a fresh master / HEAD please)
- Command line flag should be -F / --filter
- Man page needs update

layderv:
@fasterit updated, thank you for the comments

fasterit:
@layderv Excellent PR. Merged. Thank you very much.

layderv:
@fasterit would it be possible to approve the PR also? :)

fasterit:
I can't click approve on a PR that is already merged and thus closed.
Merging is the best approval you can get as it confirms the quality of your work.

layderv:
@fasterit ok, I was asking because for the hacktoberfest the PR is pending with reason "Your PR has not yet been accepted by a maintainer. A maintainer can accept your PR by merging it, adding the "hacktoberfest-accepted" label to it, or approving it" (and I believe the update should happen within seconds). Not sure why, but no problem :)

fasterit:
hm, "your PR by merging it". I've add the label, just in case.

layderv:
It worked, thank you! @fasterit

Issue #220, opened 2020-10-06
Optimize ProcessList_buildTree


BenBE:
Doing a short test with `valgrind --tool=callgrind` I noticed that about 23% of the overall runtime (20,6% without sub calls) is spent within `ProcessList_buildTree`.

One of the main contributors to this seem to be the repeated calls to `Process_isChildOf` that is called within the functions main loop.

Further locations for optimizing are the calls `LinuxProcessList_readCmdlineFile` (10.29%/6.56%) and `Vector_take` (4.40%).

Apart from this we're quite heavy on calls to the `snprintf` function family as well as string routines handling processing of numbers (no surprises here, I guess).

cgzones:
I also noticed `LinuxProcessList_readCmdlineFile()` taking a top spot, but at first look I didn't see why. (is the `xread()` functions invoking two `read()`s?)

One optimization about `snprintf` might be instead of constructing thousands of process specific paths, like `"%s/%s/status, dir, name`, opening the process-dir once, and the using openat() ?

BenBE:
As far as I understand the report for `LinuxProcessList_readCmdLineFile` there are two issues in that function:
1. the `snprintf` (~25.7%)
2. the token parsing loop (`for(int i = 0; i < amtRead; i++)`) at the bottom (>56%).

Didn't look at point 2 on the instruction level, but one minor optimization could be `if(command[i] < 32)` for the check for blanks. Sometimes avoiding the explicit indicing into an array may help too, saving on up to 3 pointer calculations. Optimization in-depth would be prefetching stuff from that buffer, but IMHO that's kinda overkill.

BenBE:
The `openat` idea is implemented in #321. Not yet used everywhere though.

New performance measurements needed for overall performance as most bottlenecks should have been addressed at least once..

cgzones:
Anything left to do here?

BenBE:
I think we are fine for now. Unless any grave bottlenecks are known we can close this. If we find any later we can still open a new issue.

PR #221, opened 2020-10-06
Optimize vector

BenBE:
This contains two small patches based on values gathered by using `callgrind`.
The overall effect should be a ~3-4% performance improvement.

This is split out from #220, as the major savings would require rewriting two core functions/core data structures in htop.


cgzones (approved):
LGTM, I like performance improvements :+1:

cgzones:
Thanks, applied.

PR #226, opened 2020-10-08
Implement configure flag to activate LTO for builds

BenBE:
This is a somewhat handy shorthand to specifying this when calling `./configure`, but also bails if it's not supported by the compiler.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -321,6 +321,36 @@ AS_IF([test "x$enable_debug" = "xyes"], , [AM_CPPFLAGS="$AM_CPPFLAGS -DNDEBUG"])
 
 AC_SUBST([AM_CPPFLAGS])
 
+# Link time optimization feature in newer gcc/g++
+#    based on http://svn.r-project.org/R/trunk/configure.ac
+AC_ARG_ENABLE([lto],
+    [AS_HELP_STRING([--enable-lto],[enable link-time optimization @<:@no@:>@])],
+    [if test "x${enableval}" = xyes -o "x${enableval}" = x; then
+        enable_lto=yes
+     elif test "x${enableval}" = xno; then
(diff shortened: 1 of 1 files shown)


cgzones:
While I like Link Time Optimization, I think optimization flags `[-OX -g -flto -march=xy]` should be configured by the end-users/package-maintainers.
(Autotools sets CFLAGS by default to `-O2 -g`, but that's common knowledge.)
Also from my experience there are some issues with older compilers (gcc < 8), and currently we require only a C99 compliant one.

BenBE:
I didn't intend to go into the `-O3` area and all the various `-f` and `-m` optimizations. And yes, I remember some kinda nasty LTO bug you could reliably trigger using `ntohs` on gcc.

Intention for the LTO patch was to have a short-hand for the setup including some basic checks if it's supported. Doesn't free you from testing your final build to be working; but that's with all compiler settings.

Explorer09:
Just let package builders enable LTO by themselves by overriding CFLAGS="-O2 -g -flto" and LDFLAGS="-O2 -g -flto". Why are you bothered to implement this as a configure script option?
People who know the LTO feature of their compilers would already know how to enable it properly. There's no need to add a redundant configure flag.

BenBE:
> Why are you bothered to implement this as a configure script option?

Mostly for laziness … The less I have to deal with `configure` and manually overriding flags, the happier I am … ;-)

natoscott:
FWIW, I tend to agree with the other folks saying this one is well intentioned but probably not worth doing ourselves.  I noticed in Fedora33 -flto has become the toolchain default, and I expect this will be the same in other distros if not already then very soon.  So in due course we can expect this option to become pervasive and forcing it on now for older gcc versions is not really buying us anything.

PR #129, opened 2020-09-14
Graph meter coloring & dynamic scaling!

eworm-de:
This work is by @Explorer09, see original pull request (https://github.com/hishamhm/htop/pull/487).
I rebased to current master to prepare this pull request.

PR #228, opened 2020-10-08
Set reasonable total value for Load Meters

cgzones:
Needed for graph and bar mode

### DIFF
--- a/LoadAverageMeter.c
+++ b/LoadAverageMeter.c
@@ -23,6 +23,7 @@ static const int LoadMeter_attributes[] = {
 
 static void LoadAverageMeter_updateValues(Meter* this, char* buffer, int size) {
    Platform_getLoadAverage(&this->values[0], &this->values[1], &this->values[2]);
+   this->total = MAXIMUM(this->values[0], 1.5 * this->pl->cpuCount);
    xSnprintf(buffer, size, "%.2f/%.2f/%.2f", this->values[0], this->values[1], this->values[2]);
 }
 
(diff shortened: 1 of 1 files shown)


fasterit:
Can you explain the rationale? Why 1.5?

cgzones:
Currently the total value for the LoadMeters is 100.
Thereby in MeterMode Graph there is (almost always) no data printed, cause the Load (in case of the LoadAverageMeter the accumulated 1+5+15 load) is far less than 100.
`1.5 *` is just a guess for a sensible maximum value, do you prefer `1.0 *` or something higher?
(In case the actual load is higher than the guessed maximum, it is still displayed as 100%, since it's getting adjusted with MAXIMUM.)

fasterit:
I would have used 2 (two) as the initial maximum (as it will adjust upwards when real maxima are hit as you said). Load is a measure for the amount of processes & threads waiting in (uninterruptible) sleep (for resources). Hence it is not CPU (core) bound per se. You seem to use the common sense experience that lots of cores have lots to do but that may only be true on servers. I assume most desktop CPUs will be idling with a load around 1-2 typically. So using 2 as a common upper bound seems reasonable to me and on machines with real load it will quickly adjust upwards.

BenBE:
Typically when I look at the load value (numerically) I check if it's above about twice of the CPU count to check if the machine is overloaded. That's also when really compiling a lot of stuff what I usually give for the parallelism as you typically have some proceses waiting for I/O, others waiting for other resources.

So putting the load at 1*CPUcount or 2*CPUcount are two values you could easily compute in your head, being 1*CPUcount being "there's always some work for every core in the system" being a meaningful ground. Having the value statically at e.g. 1 or 2 would already drown the load meter in static noise on my desktop (currently idle at about 0.5); probably getting only worse on systems with many more cores. I think 1 would be a good value if we normalized by CPU count, as than it would roughly represent the amount of time the power of the CPU was used (by any process).

Explorer09:
Load average value is not supposed to have an upper limit. For bar display, it's better to have a limit that's dynamically scaled.
I remember I experimented the feature before. See #129 if you guys are interested to test.

cgzones:
Superseded by revamp of total values in #129

PR #229, opened 2020-10-08
Add SystemdMeter

cgzones:
systemd_meter


fasterit:
Why not parse `systemctl show` which would not add a libsystemd dependency and works runtime?

fasterit:
`systemctl show --property=NInstalledJobs --property=NFailedJobs --property=NNames --property=NJobs` for the ones you currently get via DBus

SystemdMeter.c
```diff
@@ -0,0 +1,198 @@
...
+   free(systemState);
+   systemState = NULL;
+   nFailedUnits = nInstalledJobs = nNames = nJobs = (uint32_t)-1;
+
+   /* Connect to the system bus */
+   r = sd_bus_open_system(&bus);
+   if (r < 0) {
+      //fprintf(stderr, "Failed to connect to system bus: %s\n", strerror(-r));
+      goto finish;
+   }
```
BenBE, line 36: What about connecting once at startup?

SystemdMeter.c
```diff
@@ -0,0 +1,198 @@
...
+
+   r = sd_bus_get_property_string(bus,
+                                  "org.freedesktop.systemd1",           /* service to contact */
+                                  "/org/freedesktop/systemd1",          /* object path */
+                                  "org.freedesktop.systemd1.Manager",   /* interface name */
+                                  "SystemState",                        /* property name */
+                                  &error,                               /* object to return error in */
+                                  &systemState);
+   if (r < 0) {
+      //fprintf(stderr, "Failed to issue method call: %s\n", error.message);
```
BenBE, line 46: Should this debug code really be kept? (Same for the other locations)

SystemdMeter.c
```diff
@@ -0,0 +1,198 @@
...
+#include <systemd/sd-bus.h>
+
+static const int SystemdMeter_attributes[] = {
+   SYSTEMD_FAILED,
+   SYSTEMD_NOTICE,
+   SYSTEMD_OK
+};
+
+static char *systemState = NULL;
+static uint32_t nFailedUnits = 0, nInstalledJobs = 0, nNames = 0, nJobs = 0;
```
BenBE, line 14: Shouldn't this be part of the meter local data (`values` array)?

DiskIOMeter.h
```diff
@@ -2,7 +2,7 @@
 #define HEADER_DiskIOMeter
 /*
  h top - DiskIOMeter*.h
-(C) 2020 Christian Göttsche
+(C) 2020 Christian Goettsche
```
BenBE, line 5: Unrelated change?

DiskIOMeter.c
```diff
@@ -1,6 +1,6 @@
 /*
 htop - DiskIOMeter.c
-(C) 2020 Christian Göttsche
+(C) 2020 Christian Goettsche
```
BenBE, line 3: Unrelated change

cgzones:
Re-implemented via executing `systemctl`

BenBE:
Personally I don't like having to rely on parsing the output of external programs when a proper API for it exists. Even more so if this external API probably is much faster to access than repeatedly starting a new process.

fasterit:
Please add (c) headers to the new files.
Otherwise LGTM. Please feel free to merge.

fasterit:
> Personally I don't like having to rely on parsing the output of external programs when a proper API for it exists. Even more so if this external API probably is much faster to access than repeatedly starting a new process.

But this way we do not need to hard-depend on libsystemd which would mean either of:

- htop variants
- htop without systemd support
- htop pulling in systemd

for distros that support multiple init systems.

And tbh., we parse many more strings than these...

SystemdMeter.c
```diff
@@ -0,0 +1,188 @@
...
+#include "StringUtils.h"
+
+static const int SystemdMeter_attributes[] = {
+   SYSTEMD_FAILED,
+   SYSTEMD_NOTICE,
+   SYSTEMD_OK
+};
+
+static char *systemState = NULL;
+static unsigned int nFailedUnits = 0, nInstalledJobs = 0, nNames = 0, nJobs = 0;
```
BenBE, line 20: Shouldn't this be `(unsigned int)-1` as used in other places for undefined values?

BenBE:
> > Personally I don't like having to rely on parsing the output of external programs when a proper API for it exists. Even more so if this external API probably is much faster to access than repeatedly starting a new process.
> 
> But this way we do not need to hard-depend on libsystemd which would mean either of:
> 
> * htop variants
> * htop without systemd support
> * htop pulling in systemd
> 
> for distros that support multiple init systems.

We still could `dlopen`/`dlsym` it without causing the link time dependency. Wouldn't even be much overhead code-wise to do.

> And tbh., we parse many more strings than these...

Good point. :grinning:

fasterit:
> We still could `dlopen`/`dlsym` it without causing the link time dependency. Wouldn't even be much overhead code-wise to do.

That would be fine for me as well.
I measured the overhead of the execl call and it is negligible on an i7 CPU. If somebody want to measure on their ARMv5, feel free :-).

pthfdr-42:
Please not.
Doing this would likely to let distros to make systemd as a dependency of htop.

fasterit:
> Please not.
> Doing this would likely to let distros to make systemd as a dependency of htop.

That's not true. The last two iterations of this PR soft depend on systemd only, i.e. you can use the meter if you have systemd running anyways but htop does not depend in an installed libsystemd otherwise. That was a deliberate choice especially to not force systemd on people that do not want to run it.

fasterit:
1. The systemd meter only makes sense on Linux.
2. Is there a use case where systemd is active but the dlopen call fails and you need to do the parsing of `systemctl` output?
     I suggested the latter only to prevent the hard dependency on libsystemd. That is already prevented by dlopen use now.

cgzones:
On minimal systems without D-Bus the library method does not work.
There systemctl succeeds by using */run/systemd/private*, which does not have a stability promise: https://systemd.io/PORTABILITY_AND_STABILITY/

fasterit:
> On minimal systems without D-Bus the library method does not work.
> There systemctl succeeds by using _/run/systemd/private_, which does not have a stability promise: https://systemd.io/PORTABILITY_AND_STABILITY/

thanks, that resolves the question 2. from above

linux/SystemdMeter.c
```diff
@@ -0,0 +1,328 @@
...
+*/
+
+#include "SystemdMeter.h"
+
+#include <dlfcn.h>
+#include <fcntl.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <sys/wait.h>
+#include <unistd.h>
```
BenBE, line 15: After `fcntl.h`

linux/SystemdMeter.c
```diff
@@ -0,0 +1,328 @@
...
+      if (dlerror() != NULL)
+         goto dlfailure;
+
+      *(void **)(&sym_sd_bus_get_property_trivial) = dlsym(dlopenHandle, "sd_bus_get_property_trivial");
+      if (dlerror() != NULL)
+         goto dlfailure;
+
+      *(void **)(&sym_sd_bus_unref) = dlsym(dlopenHandle, "sd_bus_unref");
+      if (dlerror() != NULL)
+         goto dlfailure;
```
BenBE, line 76: `dlsym` may return `NULL` without indicating an error, thus the resolved symbols should be checked to be non-NULL.

configure.ac
```diff
@@ -83,6 +83,8 @@ AC_FUNC_CLOSEDIR_VOID
 AC_FUNC_STAT
 AC_CHECK_FUNCS([memmove strncasecmp strstr strdup])
 
+AC_SEARCH_LIBS([dlopen], [dl dld])
```
BenBE, line 86: Probably should also check `dlsym`, `dlclose` and `dlerror`.

BenBE (changes requested):

linux/SystemdMeter.c
cgzones: Isn't it the other way round? (it might return NULL on success)

https://manpages.debian.org/unstable/manpages-dev/dlsym.3.en.html
> There are several scenarios when the address of a global symbol is NULL. For example, a symbol can be placed at zero address by the linker, via a linker script or with --defsym command-line option. Undefined weak symbols also have NULL value. Finally, the symbol value may be the result of a GNU indirect function (IFUNC) resolver function that returns NULL as the resolved value. In the latter case, dlsym() also returns NULL without error. However, in the former two cases, the behavior of GNU dynamic linker is inconsistent: relocation processing succeeds and the symbol can be observed to have NULL value, but dlsym() fails and dlerror() indicates a lookup error.

BenBE: Exactly. Bad phrasing on my side. ;-)

cgzones: ahh, i missed your point at first. resolved

configure.ac
cgzones: It's mainly not about checking if `dlopen()` exists, but if we need to link with `-ldl` or `-ldld` to make the function available.

linux/SystemdMeter.c
```diff
@@ -0,0 +1,328 @@
...
+      if (!sym_sd_bus_get_property_string || dlerror() != NULL)
+         goto dlfailure;
+
+      *(void **)(&sym_sd_bus_get_property_trivial) = dlsym(dlopenHandle, "sd_bus_get_property_trivial");
+      if (!sym_sd_bus_get_property_trivial || dlerror() != NULL)
+         goto dlfailure;
+
+      *(void **)(&sym_sd_bus_unref) = dlsym(dlopenHandle, "sd_bus_unref");
+      if (!sym_sd_bus_unref || dlerror() != NULL)
+         goto dlfailure;
```
BenBE, line 76:
```suggestion
#define resolve(symbolname) do {\
      *(void **)(&sym##symbolname) = dlsym(dlopenHandle, #symbolname); \
      if (!sym##symbolname || dlerror() != NULL) \
         goto dlfailure;

      resolve(sd_bus_open_system);
      resolve(sd_bus_get_property_string);
      resolve(sd_bus_get_property_trivial);
      resolve(sd_bus_unref);

#undef resolve
```

linux/SystemdMeter.c
```diff
@@ -0,0 +1,328 @@
...
+                                      "SystemState",                        /* property name */
+                                      NULL,                                 /* object to return error in */
+                                      &systemState);
+   if (r < 0)
+      goto busfailure;
+
+   r = sym_sd_bus_get_property_trivial(bus,
+                                       "org.freedesktop.systemd1",           /* service to contact */
+                                       "/org/freedesktop/systemd1",          /* object path */
+                                       "org.freedesktop.systemd1.Manager",   /* interface name */
```
BenBE, line 101: What about having constants for those 3 strings?

linux/SystemdMeter.c
cgzones: adopted

linux/SystemdMeter.c
cgzones: adopted

linux/SystemdMeter.c
```diff
@@ -0,0 +1,328 @@
...
+
+   if (bus && dlopenHandle) {
+      sym_sd_bus_unref(bus);
+   }
+   bus = NULL;
+
+   if (dlopenHandle) {
+      dlclose(dlopenHandle);
+      dlopenHandle = NULL;
+   }
```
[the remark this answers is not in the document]
cgzones: My intention was to set `bus` to NULL in every case.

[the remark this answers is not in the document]
BenBE: If we ensure `bus` can only ever be non-NULL if all function pointers resolved successfully, I'm fine with dropping the `dlopenhandle` check in the first of your condition blocks.

Issue #222, opened 2020-10-06
Calculate library size of the process by hand [Linux]


cgzones:
Origin: #189 

On Linux the proc/[pid]/statm field `lib` ("The library size of the process") is unused since Linux 2.6 (always 0).

> BenBe:
Couldn't we "calculate" the lib size based on `/proc/<PID>/maps` by adding up all ranges with `INODE != 0` and at least one segment for that inode marked executable?

fasterit:
May be the calculations of [smem](https://www.selenic.com/smem/) are useful?
Basically their "Proportional Set Size (PSS)" sums up library memory consumption but divides it by the processes having linked to that library.

natoscott:
Just something to be aware of - the smaps file (used in PSS calcs) is relatively expensive to access, compared to maps.  There's a smaps_rollup file in more recent kernels, which offsets some of the costs but its still quite expensive.  Some other tools (e.g. atop) have a command line option to opt-in to accessing the more costly info...

```
$ sudo time cat /proc/*/smaps > /dev/null
0.01user 0.27system 0:00.29elapsed 97%CPU
$ sudo time cat /proc/*/smaps_rollup > /dev/null
0.00user 0.13system 0:00.14elapsed 97%CPU
$ sudo time cat /proc/*/maps > /dev/null
0.00user 0.03system 0:00.03elapsed 94%CPU
```

PR #231, opened 2020-10-08
Fix Hashtable_put to allow storing the same pointer

StoneBam:


cgzones:
This look OK.
Are there instances where this makes a change for you?
Cause if currently this codepath would be taken, we'll have an use-after-free, but I didn't notice any yet.

BenBE:
@StoneBam and I noticed this issue yesterday while working on #222. To simplyfy the code on our end we had the hashtable manage all the pointers and just `Hashtable_put` the updated records into the hashtable after we updated the items in the value.

I'm not aware of any places currently in the source where this bug would be triggered, but I think it's a good defense-in-depth as the use-case where it triggers can be produced by accident quite easily.

cgzones:
Thanks, applied.

PR #234, opened 2020-10-09
Add NetworkIOMeter

cgzones:
networkiometer


BenBE (commented):
Overall patch LGTM.

But isn't having such a prominent feature display 0 on every platform except Linux somewhat strange? Quick search returned for FreeBSD to use `systat -ifstat`, which should work on most other BSD flavors too.

cgzones:
This is just a first draft. I'll try to setup a FreeBSD vm on the weekend and play around with [sysctl(**3**)](https://www.freebsd.org/cgi/man.cgi?query=ifmib&sektion=4&manpath=freebsd).

cgzones:
Added implementation for FreeBSD.

PR #238, opened 2020-10-09
Calculate library size (M_LRS column) from maps file

### LINKED ISSUE #222: Calculate library size of the process by hand [Linux]

StoneBam:
Fixes #222.


cgzones:
This calculation looks kinda expensive: probably it should be guarded by a new flag
(see https://github.com/htop-dev/htop/blob/e5fdb80c7d13d836ec244390976741dd99bc6535/linux/LinuxProcess.h#L11-L18)
, so it's only computed if the M_LRS column is actually enabled.

linux/LinuxProcessList.c
```diff
@@ -404,6 +404,84 @@ static void LinuxProcessList_readIoFile(LinuxProcess* process, const char* dirna
...
+
+   char buffer[1024];
+   while(fgets(buffer, sizeof(buffer), mapsfile)) {
+      uint64_t map_start;
+      uint64_t map_end;
+      char map_perm[16];
+      uint32_t map_dummy_1;
+      int map_dummy_2;
+      int map_dummy_3;
+      uint32_t map_inode;
```
natoscott, line 450: Pretty sure this is a 64bit number on 64bit platforms, so 'uint64_t' here for the win.

(looks like the kernel uses "seq_put_decimal_ull(m, " ", ino);" over in fs/proc/task_mmu.c for example)

linux/LinuxProcessList.c
```diff
@@ -404,6 +404,84 @@ static void LinuxProcessList_readIoFile(LinuxProcess* process, const char* dirna
...
+
+      if(7 != sscanf(buffer, "%lx-%lx %4s %x %d:%d %u",
+         &map_start, &map_end, map_perm,
+         &map_dummy_1, &map_dummy_2, &map_dummy_3,
+         &map_inode)
+      ) {
+         continue;
+      }
+
```
natoscott, line 459: No curly braces needed above.

linux/LinuxProcessList.c
```diff
@@ -404,6 +404,84 @@ static void LinuxProcessList_readIoFile(LinuxProcess* process, const char* dirna
...
+   while(fgets(buffer, sizeof(buffer), mapsfile)) {
+      uint64_t map_start;
+      uint64_t map_end;
+      char map_perm[16];
+      uint32_t map_dummy_1;
+      int map_dummy_2;
+      int map_dummy_3;
+      uint32_t map_inode;
+
+      if(7 != sscanf(buffer, "%lx-%lx %4s %x %d:%d %u",
```
natoscott, line 452: whitespace missing after 'if'

linux/LinuxProcessList.c
```diff
@@ -404,6 +404,84 @@ static void LinuxProcessList_readIoFile(LinuxProcess* process, const char* dirna
...
+      ) {
+         continue;
+      }
+
+      if(!map_inode)
+         continue;
+
+      LibraryData * libdata;
+      libdata = Hashtable_get(ht, map_inode);
+      if(!libdata) {
```
natoscott, line 465: whitespace missing after 'if'

linux/LinuxProcessList.c
```diff
@@ -404,6 +404,84 @@ static void LinuxProcessList_readIoFile(LinuxProcess* process, const char* dirna
...
+         libdata = xCalloc(1, sizeof(LibraryData));
+         libdata->size=0;
+         libdata->exec=false;
```
natoscott, line 468: calloc will auto-initialize memory to zero so above two assignments seem unnecessary?  They also have whitespace inconsistent with the rest of the code FWIW.

linux/LinuxProcessList.c
```diff
@@ -404,6 +404,84 @@ static void LinuxProcessList_readIoFile(LinuxProcess* process, const char* dirna
...
 
 
+typedef struct LibraryData {
+    uint64_t size;
+    bool exec;
+} LibraryData;
+
+static void LinuxProcessList_calcLibSize_helper(
+   ATTR_UNUSED int key, void* value, void* data
+) {
```
natoscott, line 414: This isn't the usual htop coding style for functions.

linux/LinuxProcessList.c
```diff
@@ -404,6 +404,84 @@ static void LinuxProcessList_readIoFile(LinuxProcess* process, const char* dirna
...
+   if(!data)
```
natoscott, line 415: missing whitespace after the 'if'

linux/LinuxProcessList.c
```diff
@@ -404,6 +404,84 @@ static void LinuxProcessList_readIoFile(LinuxProcess* process, const char* dirna
...
+      return;
+
+   if(!value)
+      return;
+
+   LibraryData * v = (LibraryData *)value;
+   uint64_t * d = (uint64_t *)data;
+   if(!v->exec)
```
natoscott, line 423: missing whitespace after the 'if'

linux/LinuxProcessList.c
```diff
@@ -404,6 +404,84 @@ static void LinuxProcessList_readIoFile(LinuxProcess* process, const char* dirna
...
+
+static uint64_t LinuxProcessList_calcLibSize(const char* dirname, const char* name) {
+   char filename[MAX_NAME+1];
+   xSnprintf(filename, sizeof(filename), "%s/%s/maps", dirname, name);
+   FILE* mapsfile = fopen(filename, "r");
+   if (!mapsfile)
+      return 0;
+
+   Hashtable * ht = Hashtable_new(64, true);
+   if(!ht) {
```
natoscott, line 437: missing whitespace after the 'if'

natoscott (changes requested):

Process.h
```diff
@@ -21,6 +22,13 @@ in the source distribution for its full text.
 #define SYS_ioprio_set __NR_ioprio_set
 #endif
 
+// On Linux, this works only with glibc 2.1+. On earlier versions
```
cgzones, line 25: PAGE_SIZE has been refactored, see https://github.com/htop-dev/htop/commit/361877454fe56c95a995d5bdbb7eb70c21e39d62

Process.h
```diff
@@ -106,7 +114,7 @@ typedef struct ProcessFieldData_ {
    const char* name;
    const char* title;
    const char* description;
-   int flags;
+   uint_least32_t flags;
```
cgzones, line 117: i think @natoscott said he does not like fancy integers?

cgzones (changes requested):

Process.h
BenBE: Using `int` in this case would be wrong for platforms with `sizeof(int)==sizeof(short)` as this patchset introduces a 17th flag, thus we need at least 32 bit here.

Process.h
BenBE: Oops, probably an oversight while rebasing the patches …

Process.h
natoscott: > Using `int` in this case would be wrong for platforms with `sizeof(int)==sizeof(short)` [...]

What platform (that htop runs on) has that property?

( +1 to using uint32_t anyway )

linux/LinuxProcessList.c
```diff
@@ -457,7 +459,78 @@ static void LinuxProcessList_readIoFile(LinuxProcess* process, const char* dirna
...
+}
+
+static uint64_t LinuxProcessList_calcLibSize(const char* dirname, const char* name) {
+   char filename[MAX_NAME+1];
+   xSnprintf(filename, sizeof(filename), "%s/%s/maps", dirname, name);
+   FILE* mapsfile = fopen(filename, "r");
+   if (!mapsfile)
+      return 0;
+
+   Hashtable* ht = Hashtable_new(64, true);
```
cgzones, line 489: `Hashtable_new()` does never return `NULL`

cgzones:
While testing with default flags (`-O2 -g`) I noticed htop takes up to 40% CPU usage constantly (on 4*3.4 GHz).
Also with valgrind/callgrind htop is so slow I do not even get a screen rendered.

Seems like
```c
if (7 != sscanf(buffer, "%"PRIx64"-%"PRIx64" %4s %"PRIx32" %d:%d %" PRIu64,
         &map_start, &map_end, map_perm, &map_dummy,
         &map_devmaj, &map_devmin, &map_inode))
         continue;
```
is the bottleneck (gets called millions of times.

BenBE:
Do we have any maschinery to update certain values only every X iterations?

StoneBam:
In the near future I sadly will not have time to contribute to this issue. @BenBE will continue the work.

linux/LinuxProcessList.c
BenBE: Fixed.

BenBE:
@cgzones @natoscott Should we go forward with this PR despite the performance issue? Any idea for optimizing it?

fasterit:
40% CPU usage for this is ridiculous, don't merge things with such impact. We want to observe systems, not burn them for component tests.
How about checking expired time since last update and only update every 5-10 seconds or so, similar to
https://github.com/htop-dev/htop/blob/742e610f1db4212d5bb5ca6ae56147fbf6f70f28/NetworkIOMeter.c#L38
https://github.com/htop-dev/htop/blob/742e610f1db4212d5bb5ca6ae56147fbf6f70f28/DiskIOMeter.c#L44

linux/LinuxProcess.c
```diff
@@ -69,7 +69,7 @@ ProcessFieldData Process_fields[] = {
    [M_SHARE] = { .name = "M_SHARE", .title = "  SHR ", .description = "Size of the process's shared pages", .flags = 0, },
    [M_TRS] = { .name = "M_TRS", .title = " CODE ", .description = "Size of the text segment of the process", .flags = 0, },
    [M_DRS] = { .name = "M_DRS", .title = " DATA ", .description = "Size of the data segment plus stack usage of the process", .flags = 0, },
-   [M_LRS] = { .name = "M_LRS", .title = " LIB ", .description = "The library size of the process (unused since Linux 2.6; always 0)", .flags = 0, },
+   [M_LRS] = { .name = "M_LRS", .title = "  LIB ", .description = "The library size of the process (unused since Linux 2.6; always 0)", .flags = PROCESS_FLAG_LINUX_LRS_FIX, },
```
fasterit, line 72:
```suggestion
   [M_LRS] = { .name = "M_LRS", .title = "  LIB ", .description = "The library size of the process (calculated from memory maps)", .flags = PROCESS_FLAG_LINUX_LRS_FIX, },
```

Process.h
```diff
@@ -113,7 +115,7 @@ typedef struct ProcessFieldData_ {
    const char* name;
    const char* title;
    const char* description;
-   int flags;
+   uint32_t flags;
```
cgzones, line 118: Please also change the type at
https://github.com/htop-dev/htop/blob/46a2e8ac63fb758a589a82d4dca60de1aa0c1eb2/Settings.c#L93
https://github.com/htop-dev/htop/blob/46a2e8ac63fb758a589a82d4dca60de1aa0c1eb2/Settings.h#L30

linux/LinuxProcessList.c
```diff
@@ -479,22 +481,187 @@ static void LinuxProcessList_readIoFile(LinuxProcess* process, const char* dirna
...
    fclose(statmfile);
+
+   if (r == 7) {
+      if (tmp_m_lrs) {
+         process->m_lrs = tmp_m_lrs;
+      } else if (performLookup) {
+         // Check if we really should recalculate the M_LRS value for this process
+         struct timeval tv;
+         gettimeofday(&tv, NULL);
+         uint64_t timeInMilliSeconds = (uint64_t)tv.tv_sec * 1000ULL + (uint64_t)tv.tv_usec / 1000ULL;
```
cgzones, line 651: maybe use the `now` argument of `LinuxProcessList_recurseProcTree()`

natoscott (approved):
LGTM, thanks @BenBE

PR #239, opened 2020-10-09
Implement screen for active file locks

### LINKED ISSUE #184: feature wish: list locks of a process
I wonder whether "list locks for a process" functionality
can be integrated into htop. See lslocks --pid

Thanks,
Karsten

StoneBam:
Fixes #184.


Action.c
```diff
@@ -423,6 +435,7 @@ static const struct { const char* key; const char* info; } helpRight[] = {
    { .key = "      e: ", .info = "show process environment" },
    { .key = "      i: ", .info = "set IO priority" },
    { .key = "      l: ", .info = "list open files with lsof" },
+   { .key = "      x: ", .info = "list file locks with lslocks" },
```
cgzones, line 438: "list file locks of process" ?

ProcessLocksScreen.c
```diff
@@ -0,0 +1,181 @@
...
+(C) 2005-2006 Hisham H. Muhammad
+Released under the GNU GPLv2, see the COPYING file
+in the source distribution for its full text.
+*/
+
+#include "config.h" // IWYU pragma: keep
+
+#include "ProcessLocksScreen.h"
+
+#include "CRT.h"
```
cgzones, line 12: Please sort this include block

ProcessLocksScreen.c
```diff
@@ -0,0 +1,181 @@
...
+#include "ProcessList.h"
+#include "IncSet.h"
+#include "StringUtils.h"
+#include "FunctionBar.h"
+
+#include <string.h>
```
cgzones, line 18: dito

ProcessLocksScreen.c
```diff
@@ -0,0 +1,181 @@
...
+      this->pid = process->pid;
+   return (ProcessLocksScreen*) InfoScreen_init(&this->super, process, NULL, LINES-3, "        ID  TYPE       EXCLUSION  READ/WRITE NODE                            START                  END");
+}
+
+void ProcessLocksScreen_delete(Object* this) {
+   free(InfoScreen_done((InfoScreen*)this));
+}
+
+static void ProcessLocksScreen_draw(InfoScreen* this) {
+   InfoScreen_drawTitled(this, "Snapshot of file locks in process %d - %s", ((ProcessLocksScreen*)this)->pid, this->process->comm);
```
cgzones, line 63: *of* process instead of *in* process ?

Action.c
```diff
@@ -585,6 +598,7 @@ void Action_setBindings(Htop_Action* keys) {
    keys['S'] = actionSetup;
    keys['C'] = actionSetup;
    keys[KEY_F(2)] = actionSetup;
+   keys['x'] = actionLslocks;
```
fasterit, line 601: actionShowLocks to be consistent with other function names and there is no lslocks used in this (thankfully)

Action.c
```diff
@@ -365,6 +366,17 @@ static Htop_Reaction actionLsof(State* st) {
    return HTOP_REFRESH | HTOP_REDRAW_BAR;
 }
 
+static Htop_Reaction actionLslocks(State* st) {
```
fasterit, line 369: actionShowLocks as per the comment below

Action.c
BenBE: Output had no consistend maschine-readable format that would not simply warrant to read `/proc/locks` directly.

BenBE:
The CI failure stems from a difference between format string inconsistencies across platforms. Use of `PRIu64` from `<inttypes.h>` should resolve this.

ProcessLocksScreen.c
```diff
@@ -0,0 +1,180 @@
...
+
+static void ProcessLocksScreen_scan(InfoScreen* this) {
+   Panel* panel = this->display;
+   int idx = Panel_getSelectedIndex(panel);
+   Panel_prune(panel);
+   FileLocks_ProcessData* pdata = ProcessLocksScreen_getProcessData(((ProcessLocksScreen*)this)->pid);
+   if (pdata->error == 127) {
+      InfoScreen_addLine(this, "Could not read file locks.");
+   } else {
+      FileLocks_LockData* ldata = pdata->locks;
```
cgzones, line 138: Maybe print a comment in case the process holds no lock? (I am not sure though)

ProcessLocksScreen.c
```diff
@@ -0,0 +1,180 @@
...
+      while (ldata) {
+         FileLocks_Data* data = &ldata->data;
+
+         char entry[128];
+         if (ULLONG_MAX == data->end) {
+            xSnprintf(entry, sizeof(entry), "%10d: %-10s %-10s %-10s %02x:%02x:%010u %20"PRIu64" %20s",
```
cgzones, line 144: `"%10d: %-10s %-10s %-10s %02x:%02x:%010u %20"PRIu64" %20s"`
this`..^..` colon looks weird to me; maybe just two spaces between ID and type?

cgzones:
Additionally it might make sense to resolve the inode number to the file path, see lslocks implementation:
https://sources.debian.org/src/util-linux/2.36-3/misc-utils/lslocks.c/#L162

StoneBam:
CI / build-ubuntu-latest (pull_request) Check had an internal error, please retry or ignore this one.

cgzones (approved):
LGTM :+1:

Makefile.am
```diff
@@ -103,6 +104,7 @@ myhtopheaders = \
 	Panel.h \
 	Process.h \
 	ProcessList.h \
+	ProcessLocksScreen.h \
```
natoscott, line 107: Ditto.

Makefile.am
```diff
@@ -52,6 +52,7 @@ myhtopsources = \
 	Panel.c \
 	Process.c \
 	ProcessList.c \
+	ProcessLocksScreen.c \
```
natoscott, line 55: Just occurred to me this functionality is Linux-specific, so the source should live below the linux/ subdirectory.

ProcessLocksScreen.c
```diff
@@ -0,0 +1,247 @@
...
+   char sym[PATH_MAX];
+   char* ret = NULL;
+
+   memset(path, 0, sizeof(path));
+   memset(sym, 0, sizeof(sym));
+
+   /*
+    * We know the pid so we don't have to
+    * iterate the *entire* filesystem searching
+    * for the damn file.
```
natoscott, line 91: damn - let's keep the language clean please (your mum might read this code someday).

ProcessLocksScreen.c
```diff
@@ -0,0 +1,247 @@
...
+
+   while ((de = readdir(dirp))) {
+      if (!strcmp(de->d_name, ".") || !strcmp(de->d_name, ".."))
+         continue;
+
+      /* care only for numerical descriptors */
+      if (!strtoull(de->d_name, (char **) NULL, 10))
+         continue;
+
+      if (!fstatat(fd, de->d_name, &sb, 0) && inode != sb.st_ino)
```
natoscott, line 111: Do we (or should we?) have a minimal supported Linux kernel version I wonder?  fstatat is not in kernels before 2.6.something-or-other from (vague) memory.  If we want to continue to support older kernels (I suspect we do), this could be wrapped up in configure.ac based HAVE_FSTATAT style macros.

ProcessLocksScreen.c
```diff
@@ -0,0 +1,247 @@
...
+   for (size_t i = 0; i < ARRAYSIZE(data->data); i++)
+      free(data->data[i]);
+}
+
+static void ProcessLocksScreen_scan(InfoScreen* this) {
+   Panel* panel = this->display;
+   int idx = Panel_getSelectedIndex(panel);
+   Panel_prune(panel);
+   FileLocks_ProcessData* pdata = ProcessLocksScreen_getProcessData(((ProcessLocksScreen*)this)->pid);
+   if (pdata->error == 127) {
```
natoscott, line 197: 127 seems a little magical ... any reason its not just '1' (truth)?  Or perhaps use 'errno' and test for non-zero here?

ProcessLocksScreen.c
```diff
@@ -0,0 +1,247 @@
...
+      data->id = lock_id;
+      data->data[0] = xStrdup(lock_type);
+      data->data[1] = xStrdup(lock_excl);
+      data->data[2] = xStrdup(lock_rw);
+      data->data[3] = ProcessLocksScreen_getInodeFilename(lock_pid, lock_inode);
+      data->dev[0] = lock_dev[0];
+      data->dev[1] = lock_dev[1];
+      data->inode = lock_inode;
+      data->start = strtoull(lock_start, NULL, 10);
+      if (0 != strcmp(lock_end, "EOF")) {
```
natoscott, line 173: No need for curly braces here FWIW.

natoscott (changes requested):

ProcessLocksScreen.c
BenBE: What would be the alternative if `fstatat` is unsupported? Emulate using `fstat` and the textual directory name?

And wouldn't this be a runtime check anyway: I mean you can compile for one kernel and run on another. The larger issue is with the symbol being present in your libc, I guess.

From the man pages:
> fstatat() was added to Linux in kernel 2.6.16; library support was added to glibc in version 2.4.

Given 2.6.16 was released somewhere in 2006 and no supported LTS version reaches further back than 2016 (4.4), with 3.16 (from 2014) EOL'd this June, I think we are quite clear, if we label anything added to the kernel before (let's say) 2010 as okay. Otherwise we'd have to check for each and every function we call in the `configure.ac`, which is doable, but rather unpleasant. To compare: `sigaction` first appeared in 2.6.14 (end of 2005); should by that logic be subject to the same checks …

Thus for sake of simplicity: Let's keep it simple here for now and take another look at the source in general with the kernel history in mind later on, as there might be a few more surprises already in the codebase.

ProcessLocksScreen.c
BenBE: Adapted from the `strace` screen, same magic there.

ProcessLocksScreen.c
BenBE: Also could use the new `String_eq` from `XUtils.h` …

ProcessLocksScreen.c
natoscott: > What would be the alternative if `fstatat` is unsupported? Emulate using `fstat` and the textual directory name?

Just disable the feature (via configure.ac macros as suggested).

natoscott: > Otherwise we'd have to check for each and every function we call in the `configure.ac`,

All four of them...
AC_CHECK_FUNCS([memmove strncasecmp strstr strdup])

TBH, not sure why we're configure-checking those four - they're ubiquitous, and used unconditionally in the code.  *shrug*

It's not difficult at all to support older platforms - there are very few APIs htop is likely to use that are going to be problematic.  I expect the number is actually zero and this is the first 'modern' libc/kernel API we've started to use.

> To compare: `sigaction` first appeared in 2.6.14 (end of 2005); should by that logic be subject to the same checks …

Hmm, where are you getting that information?  I don't think it's correct.

sigaction(2) is a POSIX.1-2001 API and has been around for much longer than that (see the man page references to some sigaction si_code macros being supported 'up to 2.2', some "newer" values added in '2.4 onward', etc)

> [...] as there might be a few more surprises already in the codebase.

I doubt that, and there's no evidence I can see so far to suggest that's the case.

ProcessLocksScreen.c
natoscott: > Adapted from the `strace` screen, same magic there.

That's running an external command though, and 127 is a common exit status convention for 'command is not found on PATH'.  So, not quite the same thing here where we're dealing with an internal error code from a failed syscall.  @cgzones is actually reworking this kinda thing over in #272 - that's a nicer solution IMO.

ProcessLocksScreen.c
BenBE: Doesn't really make it more readable …

```diff
diff --git a/ProcessLocksScreen.c b/ProcessLocksScreen.c
index 9598d40..eff5efe 100644
--- a/ProcessLocksScreen.c
+++ b/ProcessLocksScreen.c
@@ -108,11 +108,26 @@ static char *ProcessLocksScreen_getInodeFilename(pid_t pid, ino_t inode) {
       if (!strtoull(de->d_name, (char **) NULL, 10))
          continue;
 
+#if !defined(HAVE_FSTATAT) || !defined(HAVE_READLINKAT)
+      char filepath[PATH_MAX + 1];
+      xSnprintf(filepath, sizeof(filepath), "%s/%s", path, de->d_name);
+#endif
+
+#ifdef HAVE_FSTATAT
       if (!fstatat(fd, de->d_name, &sb, 0) && inode != sb.st_ino)
          continue;
+#else
+      if (!stat(filepath, &sb)) && inode != sb.st_ino)
+         continue;
+#endif
 
+#ifdef HAVE_READLINKAT
       if ((len = readlinkat(fd, de->d_name, sym, sizeof(sym) - 1)) < 1)
          goto out;
+#else
+      if ((len = readlink(filepath, sym, sizeof(sym) - 1)) < 1)
+         goto out;
+#endif
 
       sym[len] = '\0';
 
diff --git a/configure.ac b/configure.ac
index b10d819..97260d3 100644
--- a/configure.ac
+++ b/configure.ac
@@ -81,7 +81,7 @@ AC_TYPE_UID_T
 # ----------------------------------------------------------------------
 AC_FUNC_CLOSEDIR_VOID
 AC_FUNC_STAT
-AC_CHECK_FUNCS([memmove strncasecmp strstr strdup])
+AC_CHECK_FUNCS([fstatat memmove readlinkat strdup strncasecmp strstr])
 
 save_cflags="${CFLAGS}"
 CFLAGS="${CFLAGS} -std=c99"
```

And yes, you need this third buffer in that case …

ProcessLocksScreen.c
BenBE: Unlikely, but removed anyway …

Action.c
```diff
@@ -34,6 +34,10 @@ in the source distribution for its full text.
 #include "Vector.h"
 #include "XUtils.h"
 
+#ifdef __linux__
```
cgzones, line 37: Pleas use `HTOP_LINUX` from https://github.com/htop-dev/htop/commit/ddbb0c2c35ca1070387b975e85f4fd12f4aaf93a

Action.c
```diff
@@ -371,6 +375,19 @@ static Htop_Reaction actionLsof(State* st) {
    return HTOP_REFRESH | HTOP_REDRAW_BAR;
 }
 
+#ifdef __linux__
```
cgzones, line 378: dito

Action.c
```diff
@@ -435,6 +452,9 @@ static const struct { const char* key; const char* info; } helpRight[] = {
    { .key = "      e: ", .info = "show process environment" },
    { .key = "      i: ", .info = "set IO priority" },
    { .key = "      l: ", .info = "list open files with lsof" },
+#ifdef __linux__
```
cgzones, line 455: dito

Action.c
```diff
@@ -620,6 +640,9 @@ void Action_setBindings(Htop_Action* keys) {
    keys['S'] = actionSetup;
    keys['C'] = actionSetup;
    keys[KEY_F(2)] = actionSetup;
+#ifdef __linux__
```
cgzones, line 643: dito

linux/ProcessLocksScreen.c
```diff
@@ -0,0 +1,259 @@
...
+
+   while ((de = readdir(dirp))) {
+      if (String_eq(de->d_name, ".") || String_eq(de->d_name, ".."))
+         continue;
+
+      /* care only for numerical descriptors */
+      if (!strtoull(de->d_name, (char **) NULL, 10))
+         continue;
+
+#if !defined(HAVE_FSTATAT) || !defined(HAVE_READLINKAT)
```
cgzones, line 108: please use `Compat_fstatat` from https://github.com/htop-dev/htop/commit/bbf01054bf943db4394027d77915f9625ebde81e

linux/ProcessLocksScreen.h
```diff
@@ -0,0 +1,26 @@
+#ifndef HEADER_ProcessLocksScreen
+#define HEADER_ProcessLocksScreen
+/*
+htop - ProcessLocksScreen.h
+(C) 2005-2006 Hisham H. Muhammad
```
cgzones, line 5: Please use correct authorship

linux/ProcessLocksScreen.c
```diff
@@ -0,0 +1,259 @@
+/*
+htop - ProcessLocksScreen.c
+(C) 2005-2006 Hisham H. Muhammad
```
cgzones, line 3: Please use correct authorship

StoneBam:
In the near future I sadly will not have time to contribute to this issue. @BenBE will continue the work.

linux/ProcessLocksScreen.c
BenBE: Updated to htop dev team.

ProcessLocksScreen.c
BenBE: Changed `pdata->error` to `bool`.

ProcessLocksScreen.c
BenBE: Compat wrapper added for `readlinkat`.

Action.c
BenBE: Updated.

natoscott (dismissed):
@BenBE @StoneBam this is much improved, thanks.

One area that I think is not quite right though is the way the code is laid out.  When I said earlier "this is Linux-specific code so should live below linux/" I wasn't at all clear - sorry.  The way this should be done to fit with the existing code base is to use a platform_getProcessLocks() function (or some similar, betterer name?), have the Linux-specific code in the (existing) linux/Platform.c file and the generic LocksScreen code remain platform-neutral.  Have a look at the way EnvScreen.c and Platform_getProcessEnv() are implemented, for a concrete example.

BenBE:
That's why I was asking.

Do you have by chance some information for FreeBSD and Darwin for retrieving the list of locked files?

BenBE:
> @BenBE @StoneBam this is much improved, thanks.
> 
> One area that I think is not quite right though is the way the code is laid out. When I said earlier "this is Linux-specific code so should live below linux/" I wasn't at all clear - sorry. The way this should be done to fit with the existing code base is to use a platform_getProcessLocks() function (or some similar, betterer name?), have the Linux-specific code in the (existing) linux/Platform.c file and the generic LocksScreen code remain platform-neutral. Have a look at the way EnvScreen.c and Platform_getProcessEnv() are implemented, for a concrete example.

Gave it a shot in the latest commit. Not quite happy due to the way things need to be duplicated across platforms even when nothing actually implements it, besides Linux. Made an issue #306 to discuss possibly better ways to handle this.

natoscott:
One last suggestion - take it or leave it - this structure field:

```
typedef struct FileLocks_Data_ {
   char* data[4];
```

"data" could be four separate, meaningful field names there (e.g. "type", "exclusion", "readwrite", "filename").

But either way, LGTM now - thanks.

BenBE:
Change prepared locally (field names as suggested) for inclusion as this adds valuable implicit documentation inside the code. Will add (and merge) later today.

Issue #76, opened 2020-09-02
Feature Request: Filter to show only certain sub tree(s)


BenBE:
Repost from: https://github.com/hishamhm/htop/issues/836

Short Description:
Allow the user to select some processes/threads of interest and after activating the filter view only show those or any children.

Detailed explanation in https://github.com/hishamhm/htop/issues/836#issuecomment-422931464

layderv:
Hi. I implemented the solution in the most upvoted comment in the link above. See https://github.com/htop-dev/htop/pull/219

layderv:
Can this be closed?

BenBE:
The intention of the original feature request was for more or less free selection (space bar) for the trees to show after filtering. So the basic infrastructure to have this is there, but if I remember the patchset correctly it's currently bound to some explicit filter string.

singalen:
I would appreciate even more if I could set a parent process PID on a command line.

ashamza:
I wanted to do something that's similar to this FR. I wanted to add a feature to include children and/or parents for a process when filter. A use case for it is when using a tmux/screen session on a busy server when usually I am interested in processes that's running on those sessions instead of everything.

I have made https://github.com/hishamhm/htop/pull/973#issue-356794743 to show what I was looking for. Is this already fixed? is it possible now to have such a filter?

singalen:
It doesn't look like the current filter supports it.

reportaman:
@BenBE Btw, something struck me regarding this. Given the top pane shows CPU activity, etc it would be nice to have a keybinding to toggle between viewing the active filter and no filter view. That would be useful if user sees unusual activity, and wants to toggle back to unfiltered view.

BenBE:
Such a key binding should probably best go into its own issue, as that's not directly part of this issue. But it has its merits.

yurenchen000:
> I wanted to do something that's similar to this FR. I wanted to add a feature to include children and/or parents for a process when filter. A use case for it is when using a tmux/screen session on a busy server when usually I am interested in processes that's running on those sessions instead of everything.
> 
> I have made [hishamhm/htop#973 (comment)](https://github.com/hishamhm/htop/pull/973#issue-356794743) to show what I was looking for. Is this already fixed? is it possible now to have such a filter?

That's exactly what I'm looking for. :+1:  :+1: 

When I  do `F4 Filter`, I wanna know not only the matched process but also it's parent or children, 
So that I can clearly know where I am. 
[...]

PR #240, opened 2020-10-10
Add support for sorting in Tree mode

### LINKED ISSUE #237: Feature request: Sorted tree view
Any update about this issue?
https://github.com/hishamhm/htop/issues/694

zhiburt:
Hey bump into here by *hacktoberfest* :)
I am exited doing something in such a well-known project.

(I want to believe) It closes  #237

I hope it does what it expected to do but I would spend some time to verify it and test.
And if there are any tests they may need to be updated.

Thank you for the `htop`

### DIFF
--- a/MainPanel.c
+++ b/MainPanel.c
@@ -27,10 +27,8 @@ void MainPanel_updateTreeFunctions(MainPanel* this, bool mode) {
    FunctionBar* bar = MainPanel_getFunctionBar(this);
    if (mode) {
       FunctionBar_setLabel(bar, KEY_F(5), "Sorted");
-      FunctionBar_setLabel(bar, KEY_F(6), "Collap");
    } else {
       FunctionBar_setLabel(bar, KEY_F(5), "Tree  ");
-      FunctionBar_setLabel(bar, KEY_F(6), "SortBy");
    }
 }
 
@@ -68,7 +66,6 @@ static HandlerResult MainPanel_eventHandler(Panel* super, int ch) {
       ProcessField field = ProcessList_keyAt(pl, hx);
       if (field == settings->sortKey) {
          Settings_invertSortOrder(settings);
-         settings->treeView = false;
       } else {
          reaction |= Action_setSortKey(settings, field);
       }
--- a/MainPanel.c
+++ b/MainPanel.c
@@ -25,11 +25,7 @@ static const char* const MainFunctions[]  = {"Help  ", "Setup ", "Search", "Filt
 
 void MainPanel_updateTreeFunctions(MainPanel* this, bool mode) {
    FunctionBar* bar = MainPanel_getFunctionBar(this);
-   if (mode) {
-      FunctionBar_setLabel(bar, KEY_F(5), "Sorted");
-   } else {
-      FunctionBar_setLabel(bar, KEY_F(5), "Tree  ");
-   }
+   FunctionBar_setLabel(bar, KEY_F(5), mode ? "Sorted" : "Tree  ");
 }
 
 void MainPanel_pidSearch(MainPanel* this, int ch) {
--- a/Process.h
+++ b/Process.h
@@ -109,6 +109,11 @@ typedef struct Process_ {
 
    unsigned long int minflt;
    unsigned long int majflt;
+
+   unsigned int tree_left;
+   unsigned int tree_right;
+   unsigned int tree_depth;
+   unsigned int tree_index;
 } Process;
 
 typedef struct ProcessFieldData_ {
--- a/ProcessList.c
+++ b/ProcessList.c
@@ -382,6 +382,11 @@ static void ProcessList_buildTree(ProcessList* this) {
          if (process->pid == ppid)
             r = 0;
 
+         // On Linux both the init process (pid 1) and the root UMH kernel thread (pid 2)
+         // use a ppid of 0. As that PID can't exist, we can skip searching for it.
+         if (!ppid)
+            r = 0;
+
          while (l < r) {
             int c = (l + r) / 2;
             pid_t pid = ((Process*)Vector_get(this->processes, c))->pid;
--- a/ProcessList.c
+++ b/ProcessList.c
@@ -564,11 +564,13 @@ void ProcessList_scan(ProcessList* this, bool pauseProcessUpdate) {
       }
    }
 
-   // Clear out the hashtable to avoid any left-over processes from previous build
-   //
-   // The sorting algorithm relies on the fact that
-   // len(this->displayTreeSet) == len(this->processes)
-   Hashtable_clear(this->displayTreeSet);
+   if (this->settings->treeView) {
+      // Clear out the hashtable to avoid any left-over processes from previous build
+      //
+      // The sorting algorithm relies on the fact that
+      // len(this->displayTreeSet) == len(this->processes)
+      Hashtable_clear(this->displayTreeSet);
 
-   ProcessList_buildTree(this);
+      ProcessList_buildTree(this);
+   }
 }
--- a/ProcessList.c
+++ b/ProcessList.c
@@ -20,16 +20,16 @@ in the source distribution for its full text.
 
 ProcessList* ProcessList_init(ProcessList* this, const ObjectClass* klass, UsersTable* usersTable, Hashtable* pidMatchList, uid_t userId) {
    this->processes = Vector_new(klass, true, DEFAULT_SIZE);
+   this->processes2 = Vector_new(klass, true, DEFAULT_SIZE); // tree-view auxiliary buffer
+
    this->processTable = Hashtable_new(200, false);
-   this->usersTable = usersTable;
-   this->pidMatchList = pidMatchList;
-   this->userId = userId;
-
-   // tree-view auxiliary buffer
-   this->processes2 = Vector_new(klass, true, DEFAULT_SIZE);
-
    this->displayTreeSet = Hashtable_new(200, false);
    this->draftingTreeSet = Hashtable_new(200, false);
+
+   this->usersTable = usersTable;
+   this->pidMatchList = pidMatchList;
+
+   this->userId = userId;
 
    // set later by platform-specific code
    this->cpuCount = 0;
@@ -65,11 +65,13 @@ void ProcessList_done(ProcessList* this) {
       hwloc_topology_destroy(this->topology);
    }
 #endif
-   Hashtable_delete(this->displayTreeSet);
+
    Hashtable_delete(this->draftingTreeSet);
+   Hashtable_delete(this->displayTreeSet);
    Hashtable_delete(this->processTable);
-   Vector_delete(this->processes);
+
    Vector_delete(this->processes2);
+   Vector_delete(this->processes);
 }
 
 void ProcessList_setPanel(ProcessList* this, Panel* panel) {
--- a/ProcessList.c
+++ b/ProcessList.c
@@ -241,6 +241,10 @@ static void ProcessList_updateTreeSetLayer(ProcessList* this, unsigned int leftB
       }
 
       Hashtable_put(this->draftingTreeSet, proc->tree_index, proc);
+
+      // It's not strictly necessary to do this, but doing so anyways
+      // allows for checking the correctness of the inner workings.
+      Hashtable_remove(this->displayTreeSet, newLeftBound);
    }
 
    Vector_delete(layer);
@@ -249,15 +253,20 @@ static void ProcessList_updateTreeSetLayer(ProcessList* this, unsigned int leftB
 static void ProcessList_updateTreeSet(ProcessList* this) {
    unsigned int index = 0;
    unsigned int tree_index = 1;
-   int vsize = Vector_size(this->processes);
 
-   ProcessList_updateTreeSetLayer(this, 0, vsize, 0, 0, vsize*2+1, &index, &tree_index, -1);
-   assert((int)Hashtable_count(this->draftingTreeSet) == vsize);
+   const int vsize = Vector_size(this->processes);
 
-   for(int i = 0; i < vsize; i++) {
-      Process* proc = (Process*)Hashtable_remove(this->draftingTreeSet, i);
-      Hashtable_put(this->displayTreeSet, i, proc);
-   }
+   assert(Hashtable_count(this->draftingTreeSet) == 0);
+   assert((int)Hashtable_count(this->displayTreeSet) == vsize);
+
+   ProcessList_updateTreeSetLayer(this, 0, vsize, 0, 0, vsize * 2 + 1, &index, &tree_index, -1);
+
+   Hashtable* tmp = this->draftingTreeSet;
+   this->draftingTreeSet = this->displayTreeSet;
+   this->displayTreeSet = tmp;
+
+   assert(Hashtable_count(this->draftingTreeSet) == 0);
+   assert((int)Hashtable_count(this->displayTreeSet) == vsize);
 }
 
 static void ProcessList_buildTreeBranch(ProcessList* this, pid_t pid, int level, int indent, int direction, bool show, int* node_counter, int* node_index) {
@@ -529,6 +538,12 @@ void ProcessList_scan(ProcessList* this, bool pauseProcessUpdate) {
          p->updated = false;
       }
    }
+
+   // Clear out the hashtable to avoid any left-over processes from previous build
+   //
+   // The sorting algorithm relies on the fact that
+   // len(this->displayTreeSet) == len(this->processes)
+   Hashtable_clear(this->displayTreeSet);
 
    ProcessList_buildTree(this);
 }
--- a/ProcessList.c
+++ b/ProcessList.c
@@ -20,7 +20,7 @@ in the source distribution for its full text.
 
 ProcessList* ProcessList_init(ProcessList* this, const ObjectClass* klass, UsersTable* usersTable, Hashtable* pidMatchList, uid_t userId) {
    this->processes = Vector_new(klass, true, DEFAULT_SIZE);
-   this->processTable = Hashtable_new(140, false);
+   this->processTable = Hashtable_new(200, false);
    this->usersTable = usersTable;
    this->pidMatchList = pidMatchList;
    this->userId = userId;
@@ -28,8 +28,8 @@ ProcessList* ProcessList_init(ProcessList* this, const ObjectClass* klass, Users
    // tree-view auxiliary buffer
    this->processes2 = Vector_new(klass, true, DEFAULT_SIZE);
 
-   this->displayTreeSet = Hashtable_new(4096, false);
-   this->draftingTreeSet = Hashtable_new(4096, false);
+   this->displayTreeSet = Hashtable_new(200, false);
+   this->draftingTreeSet = Hashtable_new(200, false);
 
    // set later by platform-specific code
    this->cpuCount = 0;
--- a/ProcessList.c
+++ b/ProcessList.c
@@ -78,7 +78,9 @@ void ProcessList_setPanel(ProcessList* this, Panel* panel) {
 
 void ProcessList_printHeader(ProcessList* this, RichString* header) {
    RichString_prune(header);
+
    const ProcessField* fields = this->settings->fields;
+
    for (int i = 0; fields[i]; i++) {
       const char* field = Process_fields[fields[i]].title;
       if (!field) {
@@ -130,11 +132,11 @@ void ProcessList_remove(ProcessList* this, Process* p) {
 }
 
 Process* ProcessList_get(ProcessList* this, int idx) {
-   return (Process*) (Vector_get(this->processes, idx));
+   return (Process*)Vector_get(this->processes, idx);
 }
 
 int ProcessList_size(ProcessList* this) {
-   return (Vector_size(this->processes));
+   return Vector_size(this->processes);
 }
 
 // ProcessList_updateTreeSetLayer sorts this->displayTreeSet,
@@ -188,13 +190,15 @@ static void ProcessList_updateTreeSetLayer(ProcessList* this, unsigned int leftB
       Process* proc = (Process*)Hashtable_get(this->displayTreeSet, i);
       if (proc->tree_depth == deep && proc->tree_left > left && proc->tree_right < right) {
          if (Vector_size(layer) > 0) {
-            Process* previous_process = (Process*)Vector_get(layer, Vector_size(layer)-1);
-            // Make a 'right_bound' of previous_process in a layer a current's process index.
+            Process* previous_process = (Process*)Vector_get(layer, Vector_size(layer) - 1);
+
+            // Make a 'right_bound' of previous_process in a layer the current process's index.
             //
             // Use 'tree_depth' as a temporal variable.
             // it is save to do as later 'tree_depth' will be renovated.
             previous_process->tree_depth = proc->tree_index;
          }
+
          Vector_add(layer, proc);
       }
    }
@@ -204,7 +208,7 @@ static void ProcessList_updateTreeSetLayer(ProcessList* this, unsigned int leftB
    //
    // Thus, if present, set the `rightBound` to the last process on the layer
    if (Vector_size(layer) > 0) {
-      Process* previous_process = (Process*)Vector_get(layer, Vector_size(layer)-1);
+      Process* previous_process = (Process*)Vector_get(layer, Vector_size(layer) - 1);
       previous_process->tree_depth = rightBound;
    }
 
@@ -217,13 +221,13 @@ static void ProcessList_updateTreeSetLayer(ProcessList* this, unsigned int leftB
       unsigned int idx = (*index)++;
       int newLeft = (*treeIndex)++;
 
-      int level = deep == 0 ? 0 : (int)deep-1;
+      int level = deep == 0 ? 0 : (int)deep - 1;
       int currentIndent = indent == -1 ? 0 : indent | (1 << level);
-      int nextIndent = indent == -1 ? 0 : (i < size - 1) ? currentIndent : indent;
+      int nextIndent = indent == -1 ? 0 : ((i < size - 1) ? currentIndent : indent);
 
       unsigned int newLeftBound = proc->tree_index;
       unsigned int newRightBound = proc->tree_depth;
-      ProcessList_updateTreeSetLayer(this, newLeftBound, newRightBound, deep+1, proc->tree_left, proc->tree_right, index, treeIndex, nextIndent);
+      ProcessList_updateTreeSetLayer(this, newLeftBound, newRightBound, deep + 1, proc->tree_left, proc->tree_right, index, treeIndex, nextIndent);
 
       int newRight = (*treeIndex)++;
 
@@ -273,9 +277,9 @@ static void ProcessList_buildTreeBranch(ProcessList* this, pid_t pid, int level,
    Vector* children = Vector_new(Class(Process), false, DEFAULT_SIZE);
 
    for (int i = Vector_size(this->processes) - 1; i >= 0; i--) {
-      Process* process = (Process*) (Vector_get(this->processes, i));
+      Process* process = (Process*)Vector_get(this->processes, i);
       if (process->show && Process_isChildOf(process, pid)) {
-         process = (Process*) (Vector_take(this->processes, i));
+         process = (Process*)Vector_take(this->processes, i);
          Vector_add(children, process);
       }
    }
@@ -283,7 +287,7 @@ static void ProcessList_buildTreeBranch(ProcessList* this, pid_t pid, int level,
    int size = Vector_size(children);
    for (int i = 0; i < size; i++) {
       int index = (*node_index)++;
-      Process* process = (Process*) (Vector_get(children, i));
+      Process* process = (Process*)Vector_get(children, i);
 
       int lft = (*node_counter)++;
 
@@ -298,10 +302,10 @@ static void ProcessList_buildTreeBranch(ProcessList* this, pid_t pid, int level,
          Vector_insert(this->processes2, 0, process);
       }
 
-      assert(Vector_size(this->processes2) == s+1); (void)s;
+      assert(Vector_size(this->processes2) == s + 1); (void)s;
 
       int nextIndent = indent | (1 << level);
-      ProcessList_buildTreeBranch(this, process->pid, level+1, (i < size - 1) ? nextIndent : indent, direction, show ? process->showChildren : false, node_counter, node_index);
+      ProcessList_buildTreeBranch(this, process->pid, level + 1, (i < size - 1) ? nextIndent : indent, direction, show ? process->showChildren : false, node_counter, node_index);
       if (i == size - 1) {
          process->indent = -nextIndent;
       } else {
@@ -312,7 +316,7 @@ static void ProcessList_buildTreeBranch(ProcessList* this, pid_t pid, int level,
 
       process->tree_left = lft;
       process->tree_right = rht;
-      process->tree_depth = level+1;
+      process->tree_depth = level + 1;
       process->tree_index = index;
       Hashtable_put(this->displayTreeSet, index, process);
    }
@@ -333,43 +337,52 @@ static long ProcessList_treeProcessCompareByPID(const void* v1, const void* v2) 
    return SPACESHIP_NUMBER(p1->pid, p2->pid);
 }
 
+// Builds a sorted tree from scratch, without relying on previously gathered information
 static void ProcessList_buildTree(ProcessList* this) {
    int node_counter = 1;
    int node_index = 0;
    int direction = this->settings->direction;
+
    // Sort by PID
    Vector_quickSortCustomCompare(this->processes, ProcessList_treeProcessCompareByPID);
    int vsize = Vector_size(this->processes);
+
    // Find all processes whose parent is not visible
    int size;
    while ((size = Vector_size(this->processes))) {
       int i;
       for (i = 0; i < size; i++) {
-         Process* process = (Process*)(Vector_get(this->processes, i));
-         // Immediately consume not shown processes
+         Process* process = (Process*)Vector_get(this->processes, i);
+
+         // Immediately consume processes hidden from view
          if (!process->show) {
-            process = (Process*)(Vector_take(this->processes, i));
+            process = (Process*)Vector_take(this->processes, i);
             process->indent = 0;
             process->tree_depth = 0;
-            process->tree_left = (node_counter)++;
-            process->tree_index = (node_index)++;
+            process->tree_left = node_counter++;
+            process->tree_index = node_index++;
             Vector_add(this->processes2, process);
             ProcessList_buildTreeBranch(this, process->pid, 0, 0, direction, false, &node_counter, &node_index);
-            process->tree_right = (node_counter)++;
+            process->tree_right = node_counter++;
             Hashtable_put(this->displayTreeSet, process->tree_index, process);
             break;
          }
+
          pid_t ppid = Process_getParentPid(process);
+
          // Bisect the process vector to find parent
-         int l = 0, r = size;
+         int l = 0;
+         int r = size;
+
          // If PID corresponds with PPID (e.g. "kernel_task" (PID:0, PPID:0)
          // on Mac OS X 10.11.6) cancel bisecting and regard this process as
          // root.
          if (process->pid == ppid)
             r = 0;
+
          while (l < r) {
             int c = (l + r) / 2;
-            pid_t pid = ((Process*)(Vector_get(this->processes, c)))->pid;
+            pid_t pid = ((Process*)Vector_get(this->processes, c))->pid;
             if (ppid == pid) {
                break;
             } else if (ppid < pid) {
@@ -378,29 +391,34 @@ static void ProcessList_buildTree(ProcessList* this) {
                l = c + 1;
             }
          }
-         // If parent not found, then construct the tree with this root
+
+         // If parent not found, then construct the tree with this node as root
          if (l >= r) {
-            process = (Process*)(Vector_take(this->processes, i));
+            process = (Process*)Vector_take(this->processes, i);
             process->indent = 0;
             process->tree_depth = 0;
-            process->tree_left = (node_counter)++;
-            process->tree_index = (node_index)++;
+            process->tree_left = node_counter++;
+            process->tree_index = node_index++;
             Vector_add(this->processes2, process);
             Hashtable_put(this->displayTreeSet, process->tree_index, process);
             ProcessList_buildTreeBranch(this, process->pid, 0, 0, direction, process->showChildren, &node_counter, &node_index);
-            process->tree_right = (node_counter)++;
+            process->tree_right = node_counter++;
             break;
          }
       }
+
       // There should be no loop in the process tree
       assert(i < size);
    }
-   assert(Vector_size(this->processes2) == vsize); (void)vsize;
-   assert(Vector_size(this->processes) == 0);
+
    // Swap listings around
    Vector* t = this->processes;
    this->processes = this->processes2;
    this->processes2 = t;
+
+   // Check consistency of the built structures
+   assert(Vector_size(this->processes) == vsize); (void)vsize;
+   assert(Vector_size(this->processes2) == 0);
 }
 
 void ProcessList_sort(ProcessList* this) {
--- a/ProcessList.c
+++ b/ProcessList.c
@@ -8,10 +8,13 @@ in the source distribution for its full text.
 #include "ProcessList.h"
 
 #include <assert.h>
+#include <stdlib.h>
 #include <string.h>
 #include <time.h>
 
 #include "CRT.h"
+#include "Hashtable.h"
+#include "Vector.h"
 #include "XUtils.h"
 
 
@@ -24,6 +27,9 @@ ProcessList* ProcessList_init(ProcessList* this, const ObjectClass* klass, Users
 
    // tree-view auxiliary buffer
    this->processes2 = Vector_new(klass, true, DEFAULT_SIZE);
+
+   this->displayTreeSet = Hashtable_new(4096, false);
+   this->draftingTreeSet = Hashtable_new(4096, false);
 
    // set later by platform-specific code
    this->cpuCount = 0;
@@ -59,6 +65,8 @@ void ProcessList_done(ProcessList* this) {
       hwloc_topology_destroy(this->topology);
    }
 #endif
+   Hashtable_delete(this->displayTreeSet);
+   Hashtable_delete(this->draftingTreeSet);
    Hashtable_delete(this->processTable);
    Vector_delete(this->processes);
    Vector_delete(this->processes2);
@@ -77,7 +85,7 @@ void ProcessList_printHeader(ProcessList* this, RichString* header) {
          field = "- ";
       }
 
-      int color = (!this->settings->treeView && this->settings->sortKey == fields[i]) ?
+      int color = (this->settings->sortKey == fields[i]) ?
          CRT_colors[PANEL_SELECTION_FOCUS] : CRT_colors[PANEL_HEADER_FOCUS];
       RichString_append(header, color, field);
       if (COMM == fields[i] && this->settings->showMergedCommand) {
@@ -129,7 +137,130 @@ int ProcessList_size(ProcessList* this) {
    return (Vector_size(this->processes));
 }
 
-static void ProcessList_buildTree(ProcessList* this, pid_t pid, int level, int indent, int direction, bool show) {
+// ProcessList_updateTreeSetLayer sorts this->displayTreeSet,
+// relying only on itself.
+//
+// Algorithm
+//
+// The algorithm is based on `depth-first search`,
+// even though `breadth-first search` approach may be more efficient on first glance,
+// after comparision it may be not, as it's not save to go deeper without first updating the tree structure.
+// If it would be save that approach would likely bring an advantage in performance.
+//
+// Each call of the function looks for a 'layer'. A 'layer' is a list of processes with the same depth.
+// First it sorts a list. Then it runs the function recursively for each element of the sorted list.
+// After that it updates the settings of processes.
+//
+// It relies on `leftBound` and `rightBound` as an optimization to cut the list size at the time it builds a 'layer'.
+//
+// It uses a temporary Hashtable `draftingTreeSet` because it's not save to traverse a tree
+// and at the same time make changes in it.
+static void ProcessList_updateTreeSetLayer(ProcessList* this, unsigned int leftBound, unsigned int rightBound, unsigned int deep, unsigned int left, unsigned int right, unsigned int* index, unsigned int* treeIndex, int indent) {
+   // It's guaranted that layer_size is enough space
+   // but most likely it needs less. Specifically on first iteration.
+   int layerSize = (right - left) / 2;
+
+   // check if we reach `children` of `leafes`
+   if (layerSize == 0)
+      return;
+
+   Vector* layer = Vector_new(this->processes->type, false, layerSize);
+
+   // Find all processes on the same layer (process with the same `deep` value
+   // and included in a range from `leftBound` to `rightBound`.
+   //
+   // This loop also keeps track of left_bound and right_bound of these processes
+   // in order not to lose this information once the list is sorted.
+   //
+   // The variables left_bound and right_bound are different from what the values lhs and rhs represent,
+   // While left_bound and right_bound define a range of processes to look at, the values given by lhs and rhs are indices into an array
+   //
+   // In the below example note how filtering a range of indices i is different from filtering for processes in the bounds left_bound < x < right_bound …
+   //
+   // The nested tree set is sorted by left value, which is guaranteed upon entry/exit of this function.
+   //
+   // i | l | r
+   // 1 | 1 | 9
+   // 2 | 2 | 8
+   // 3 | 4 | 5
+   // 4 | 6 | 7
+   for (unsigned int i = leftBound; i < rightBound; i++) {
+      Process* proc = (Process*)Hashtable_get(this->displayTreeSet, i);
+      if (proc->tree_depth == deep && proc->tree_left > left && proc->tree_right < right) {
+         if (Vector_size(layer) > 0) {
+            Process* previous_process = (Process*)Vector_get(layer, Vector_size(layer)-1);
+            // Make a 'right_bound' of previous_process in a layer a current's process index.
+            //
+            // Use 'tree_depth' as a temporal variable.
+            // it is save to do as later 'tree_depth' will be renovated.
+            previous_process->tree_depth = proc->tree_index;
+         }
+         Vector_add(layer, proc);
+      }
+   }
+
+   // The loop above changes process-1 so the last process on the layer
+   // isn't updated by the that loop.
+   //
+   // Thus, if present, set the `rightBound` to the last process on the layer
+   if (Vector_size(layer) > 0) {
+      Process* previous_process = (Process*)Vector_get(layer, Vector_size(layer)-1);
+      previous_process->tree_depth = rightBound;
+   }
+
+   Vector_quickSort(layer);
+
+   int size = Vector_size(layer);
+   for (int i = 0; i < size; i++) {
+      Process* proc = (Process*)Vector_get(layer, i);
+
+      unsigned int idx = (*index)++;
+      int newLeft = (*treeIndex)++;
+
+      int level = deep == 0 ? 0 : (int)deep-1;
+      int currentIndent = indent == -1 ? 0 : indent | (1 << level);
+      int nextIndent = indent == -1 ? 0 : (i < size - 1) ? currentIndent : indent;
+
+      unsigned int newLeftBound = proc->tree_index;
+      unsigned int newRightBound = proc->tree_depth;
+      ProcessList_updateTreeSetLayer(this, newLeftBound, newRightBound, deep+1, proc->tree_left, proc->tree_right, index, treeIndex, nextIndent);
+
+      int newRight = (*treeIndex)++;
+
+      proc->tree_left = newLeft;
+      proc->tree_right = newRight;
+      proc->tree_index = idx;
+      proc->tree_depth = deep;
+
+      if (indent == -1) {
+         proc->indent = 0;
+      } else if (i == size - 1) {
+         proc->indent = -currentIndent;
+      } else {
+         proc->indent = currentIndent;
+      }
+
+      Hashtable_put(this->draftingTreeSet, proc->tree_index, proc);
+   }
+
+   Vector_delete(layer);
+}
+
+static void ProcessList_updateTreeSet(ProcessList* this) {
+   unsigned int index = 0;
+   unsigned int tree_index = 1;
+   int vsize = Vector_size(this->processes);
+
+   ProcessList_updateTreeSetLayer(this, 0, vsize, 0, 0, vsize*2+1, &index, &tree_index, -1);
+   assert((int)Hashtable_count(this->draftingTreeSet) == vsize);
+
+   for(int i = 0; i < vsize; i++) {
+      Process* proc = (Process*)Hashtable_remove(this->draftingTreeSet, i);
+      Hashtable_put(this->displayTreeSet, i, proc);
+   }
+}
+
+static void ProcessList_buildTreeBranch(ProcessList* this, pid_t pid, int level, int indent, int direction, bool show, int* node_counter, int* node_index) {
    Vector* children = Vector_new(Class(Process), false, DEFAULT_SIZE);
 
    for (int i = Vector_size(this->processes) - 1; i >= 0; i--) {
@@ -139,9 +270,14 @@ static void ProcessList_buildTree(ProcessList* this, pid_t pid, int level, int i
          Vector_add(children, process);
       }
    }
+
    int size = Vector_size(children);
    for (int i = 0; i < size; i++) {
+      int index = (*node_index)++;
       Process* process = (Process*) (Vector_get(children, i));
+
+      int lft = (*node_counter)++;
+
       if (!show) {
          process->show = false;
       }
@@ -156,13 +292,20 @@ static void ProcessList_buildTree(ProcessList* this, pid_t pid, int level, int i
       assert(Vector_size(this->processes2) == s+1); (void)s;
 
       int nextIndent = indent | (1 << level);
-      ProcessList_buildTree(this, process->pid, level+1, (i < size - 1) ? nextIndent : indent, direction, show ? process->showChildren : false);
-
+      ProcessList_buildTreeBranch(this, process->pid, level+1, (i < size - 1) ? nextIndent : indent, direction, show ? process->showChildren : false, node_counter, node_index);
       if (i == size - 1) {
          process->indent = -nextIndent;
       } else {
          process->indent = nextIndent;
       }
+
+      int rht = (*node_counter)++;
+
+      process->tree_left = lft;
+      process->tree_right = rht;
+      process->tree_depth = level+1;
+      process->tree_index = index;
+      Hashtable_put(this->displayTreeSet, index, process);
    }
    Vector_delete(children);
 }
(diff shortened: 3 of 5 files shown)


ProcessList.c
```diff
@@ -122,6 +122,12 @@ static void ProcessList_buildTree(ProcessList* this, pid_t pid, int level, int i
          Vector_add(children, process);
       }
    }
+
+   // sort children in by default comparator
+   // so all children will be sorted by it.
+   // only a nodes with deep == 0 will be sorted by PID
```
BenBE, line 128: `a nodes` -> `nodes` vs. `a node`

BenBE, line 128: Any reason to skip sorting items with `deep!=0`? Or am I reading this comment wrong?

MainPanel.c
```diff
@@ -19,10 +19,8 @@ void MainPanel_updateTreeFunctions(MainPanel* this, bool mode) {
    FunctionBar* bar = MainPanel_getFunctionBar(this);
    if (mode) {
       FunctionBar_setLabel(bar, KEY_F(5), "Sorted");
-      FunctionBar_setLabel(bar, KEY_F(6), "Collap");
    } else {
       FunctionBar_setLabel(bar, KEY_F(5), "Tree  ");
-      FunctionBar_setLabel(bar, KEY_F(6), "SortBy");
```
BenBE, line 25: Why remove those?

zhiburt: As I these lines changes the default bar. But there's no reason to do that anymore because the label is always set to `SortBy`

ProcessList.c
zhiburt: Actually the it means a little different; it doesn't sort items with `initial` deep (basiclally the top processis like `init`).

zhiburt:
So basically it doesn't sort a main line (init on screenshot).
image

If there's more `init`-like processies they will be not sorted

If it sort everything it works like this 

image

ProcessList.c
zhiburt: Take a look at my last comment

zhiburt: Probably the comment should be removed

BenBE:
The highest level of processes may contain kernel threads alongside the init process if they are visible. Also when filtering visible processes (cf. e.g. #76) The main level may or may not contain multiple items.

Also I noticed the tree sorting currently is kinda expensive (calculation-wise) as noted in #220. As building the tree basically already gives you all sub-sets of processes that need to be sorted this might be a good starting point to optimize things. What about tracking the process subsets by assigning "location ranges", where children always are located between their parent's "location range". The basic idea is e.g. used in Joomla to display [nested sets of items](https://docs.joomla.org/Using_nested_sets).

When sorting all inserts into this nested set structure at buildTree time you should be able to get to about O(n*log(n)) for overall sorting (provided amortized constant access to the range values for any given process (which should be doable using a sufficiently large hashtable).

zhiburt:
I've examined [wiki](https://en.wikipedia.org/wiki/Nested_set_model) and [your Joomla link](https://docs.joomla.org/Using_nested_sets).

I hope I understood what it's supposed to look like.

image

_The image almost a copy paste from wiki._

But unfortunately I haven't got an complete view why it's a beneficial approach yet :disappointed: .

Do you propose to recalculate everything in `ProcessList_sort` or it's supposed to be like a brand new data structure which will be kept up to date. (There will be always a valid processes)?

BenBE:
I mostly thought of this data structure as a new structure kept up to date whenever there's a change to the process list (so basically in `ProcessList_buildTree`). When changing the sort order you can then basically rebuild this tree structure there and the insert of each process into the tree using this structure will automatically enforce the new sort ordering (doing the sort at tree insert time).

zhiburt:
@BenBE Sorry If this question will appear a bit out of place but ... :disappointed_relieved: 

As I understand each insert/remove in `ProcessList` will require to recalculate `location ranges` right? And this list get updated a lot as I understand. 
Therefore I just doubt that it will really be more efficient :)

> Nested sets are very slow for inserts because it requires updating left and right domain values for all records in the table after the insert. This can cause a lot of database stress as many rows are rewritten and indexes rebuilt. However, if it is possible to store a forest of small trees in table instead of a single big tree, the overhead may be significantly reduced, since only one small tree must be updated.

BenBE:
> @BenBE Sorry If this question will appear a bit out of place but ... 
> 
No need to be shy. :) When not here, where to place it instead? I think this place is fine.

> As I understand each insert/remove in `ProcessList` will require to recalculate `location ranges` right? And this list get updated a lot as I understand.
>
AFAIC the only actual updates that need to really rebuild the ordering is
1. when the process list got re-scanned in `ProcessList_scan`
2. when the process list needs to be sorted

Any action that just filters the visible items does not need to care about these nested sets.

> Therefore I just doubt that it will really be more efficient :)
> 
TBH: I'm still in quite early stages of my confidence on this too. ;) Was mostly an idea that might help us with this issue and maybe some related work that is still on the tracker. No guarantee that it actually will.

zhiburt:
Just publishing an update.

I did manage filling of new Hashmap<pid, (left, right)> by small changes in `ProcessList_buildTree` (just add a counter that's it :smile: ). All though I tried to rewrite it once and didn't succeed  :disappointed: .

```
pid=1400; lft=1073; rhs=1074
pid=0; lft=0; rhs=1519
pid=1; lft=1; rhs=1308
pid=1402; lft=1071; rhs=1072
pid=2; lft=1309; rhs=1310
pid=3; lft=1311; rhs=1312
pid=12604; lft=235; rhs=236
pid=4; lft=1313; rhs=1314
pid=1405; lft=1089; rhs=1090
pid=2461; lft=296; rhs=1055
```

The main point as I understand was to simplify sorting but I didn't remove any stuff in actual function so I just thinking :thinking:.

Overall how this information has to be used? I don't thing about sorting it as well about keeping a new process vector keep up to date acording to this information.

I suppose you may have expected something different in the approach so I decided to put it here

Thanks

BenBE:
The idea was more the other way around, so you have the lhs/rhs values in the process, but the map gives you `map<order, process>`. This way, when a new process is inserted, you can look up the parent of where it needs to go and walk the direct children by looking them up in the map (between the parent's lhs/rhs values skipping to the child's rhs+1 value for the next child to look at).

Will look at the source later.

zhiburt:
It will be fair to say that I stuck here.

These my small changes to buildTree function to collect `lhs`, `rht`, `deep` information. There also changes in sorting which does pretty the same as building function so basically I am not using the new fields at all. 
 
https://github.com/zhiburt/htop/blob/3c82dea7321567eba57c6519f004409551c8cddd/ProcessList.c#L168-L212


Barely see the way how to keep list *sorted*.

Was it supposed to build a tree from the ground in `ProcessList_add`? And could you enghliten me what `order` means in your map. Since as it is a basic index why not just using a `Vector<process>`

To be candor I was hesitating to write since I was trying to understand this question as well as have it done for this period :disappointed: .

zhiburt:
PS: is it save delete?

https://github.com/htop-dev/htop/blob/9a277f25fae6913057707e2810b9338b73be4394/ProcessList.c#L149

We don't take the value but get here

BenBE:
> It will be fair to say that I stuck here.
> 
Don't worry. :)

> These my small changes to buildTree function to collect `lhs`, `rht`, `deep` information. There also changes in sorting which does pretty the same as building function so basically I am not using the new fields at all.
> 
> https://github.com/zhiburt/htop/blob/3c82dea7321567eba57c6519f004409551c8cddd/ProcessList.c#L168-L212
> 
> Barely see the way how to keep list _sorted_.
> 
> Was it supposed to build a tree from the ground in `ProcessList_add`?

Building incrementally using the nesting information might work as well _IMHO_ (and also reduces maintainance if just a few processes are added/removed). In `ProcessList_delete` as a counterpart you'd have to update the information accordingly too. When the whole list is sorted anew you'd rebuild the new view by iterating the existing one (sans `order==right` as you'd have already added them) and perform the sorting just by inserting them in the new sorted order in `ProcessList_add`.

> And could you enghliten me what `order` means in your map. Since as it is a basic index why not just using a `Vector<process>`
> 

`order` is just the display order you get when you walk the nested structure. The first item to show has 0, the next one 1, the one after that 2, and so on … With each item having a left and right boundary you visit each process twice in this manner (once for the left boundary - before any of its children - and once for the right boundary - after any of its children).

As the values for `order` should be contiguous a vector should be suitable, but keep in mind the current vector implementation isn't really optimized for bulk insertions/removals (but this can be changed easily). Note that this vector MUST NOT be the owner of the process objects, as each process appears twice and thus you'd break memory management (Just the obvious reason ;-)).

> To be candor I was hesitating to write since I was trying to understand this question as well as have it done for this period  .
> 
> I'll spend a little more time on it will see what happened

Don't hesitate to keep asking if things still remain unclear.

BenBE:
> PS: is it save delete?
> 
> https://github.com/htop-dev/htop/blob/9a277f25fae6913057707e2810b9338b73be4394/ProcessList.c#L149
> 
> We don't `take` the value but `get` it. As I see delete will free it's values ultimately. But we use this pointers later.
> 
> https://github.com/htop-dev/htop/blob/9a277f25fae6913057707e2810b9338b73be4394/ProcessList.c#L133

The list of children is initialized to not become the owner of pointers you insert, cf.:
https://github.com/htop-dev/htop/blob/9a277f25fae6913057707e2810b9338b73be4394/ProcessList.c#L116

See also the implementation of `Vector_new`:
https://github.com/htop-dev/htop/blob/9a277f25fae6913057707e2810b9338b73be4394/Vector.c#L16-L29

Process.c
```diff
@@ -477,6 +477,8 @@ long Process_compare(const void* v1, const void* v2) {
       return (p1->processor - p2->processor);
    case SESSION:
       return (p1->session - p2->session);
+   case TREE_SET:
+      return (p1->tree_lft < p2->tree_lft) ? -1 : 1;
```
BenBE, line 487: Why not `p1->tree_lft - p2->tree_lft`?

Process.h
```diff
@@ -103,6 +104,10 @@ typedef struct Process_ {
 
    unsigned long int minflt;
    unsigned long int majflt;
+
+   unsigned int tree_lft;
+   unsigned int tree_rht;
```
BenBE, line 109: We do not need to save on characters. Why not `tree_left` and `tree_right`?

Process.h
```diff
@@ -103,6 +104,10 @@ typedef struct Process_ {
...
+   unsigned int tree_deep;
```
BenBE, line 110: `tree_level` or `tree_depth`?

ProcessList.c
```diff
@@ -25,6 +26,8 @@ ProcessList* ProcessList_init(ProcessList* this, const ObjectClass* klass, Users
    // tree-view auxiliary buffer
    this->processes2 = Vector_new(klass, true, DEFAULT_SIZE);
 
+   this->treeSet = Hashtable_new(140, false);
```
BenBE, line 28: Any chance you could better guess-timate this initial size?

As an hashtable should usually be filled to 2/3 or 3/4 and we know the rought number of items we have, we might work from there?

ProcessList.c
```diff
@@ -60,6 +63,7 @@ void ProcessList_done(ProcessList* this) {
    Hashtable_delete(this->processTable);
    Vector_delete(this->processes);
    Vector_delete(this->processes2);
+   Hashtable_delete(this->treeSet);
```
BenBE, line 65: Delete weak pointers (the treeSet) before the hard (owning) links (actual process list). That way you avoid a race condition where there are pointers around for memory that has been freed.

BenBE (changes requested):
No complete review, just the notes what I noticed directly …

ProcessList.c
zhiburt: at `ProcessList_init` time it's impossible?
But at the time we use the map we know the exact size (`size == Vector_size(this->processes)`).

BenBE: k, mea culpa ;-) Thought we had some Vector there already to derive this information.
Anyway, 140 is quite optimistic: Just on my current system I reacht about 1000 threads and I have hardly anything™ open (except for chromium+some shells). Going from there and knowing that we basically have to manage a few pointers we could do 4096 for the initial size of that HashSet and will most likely stay within the constraints for the HashSet to keep good performance.

ProcessList.c
```diff
@@ -112,16 +115,73 @@ int ProcessList_size(ProcessList* this) {
...
 
-static void ProcessList_buildTree(ProcessList* this, pid_t pid, int level, int indent, int direction, bool show) {
+static int ProcessList_updateTreeSetLayer(ProcessList* this, unsigned int deep, unsigned int lft, unsigned int rht, int* index, int* tree_index, int indent) {
+   Vector* layer = Vector_new(this->processes->type, false, DEFAULT_SIZE);
+   for (int i = 0; i < Vector_size(this->processes); i++) {
+      Process* proc = (Process*)Vector_get(this->processes, i);
+      if (proc->tree_deep == deep && proc->tree_lft > lft && proc->tree_rht < rht) {
+         Vector_add(layer, proc);
+      }
+   }
```
BenBE: IF you ensure the nested set is up-to-date you can optimize this to start from `parent->tree_left+1` and skipping to `currProcess->tree_right+1` until you reach `parent->tree_right`. Also initial size for that Vector could be estimated (reserved) as `(parent->tree_right - parent->tree_left)/2` and you would never need to re-allocate while extracting that list.

Process.c
zhiburt: Actually it's not the same. 
It turned out it that it's `p2->tree_lft - p1->tree_lft`

ProcessList.c
```diff
@@ -112,16 +115,73 @@ int ProcessList_size(ProcessList* this) {
...
+      else
+         proc->indent = currentIndent;
+   }
+
+   Vector_delete(layer);
+
+   return size;
+}
+
+static void ProcessList_updateTreeSet(ProcessList* this) {
```
BenBE, line 158: Some short documentation what the various new functions do and how they interact might be of help. E.g. what's the difference between `ProcessList_updateTreeSet` and `ProcessList_buildTree`. What are some pre-conditions these functions expect?

zhiburt: A good question. Both of these build a treeSet.

But `ProcessList_buildTree` doesn't use a privious map so basically it does look up the children by pid. Where `ProcessList_updateTreeSet` rebuild map only on knowledge of map.

https://github.com/htop-dev/htop/blob/a0c24b809dd7970019148b12d34c998f62b33bdc/ProcessList.c#L176-L182

https://github.com/htop-dev/htop/blob/a0c24b809dd7970019148b12d34c998f62b33bdc/ProcessList.c#L120-L125

That's the whole difference I suppose

ProcessList.c
zhiburt: An optimisation could be possible if we would know that `this->processes` was sorted already by `TREE_SET`. Since currently I assume it's may be false

BenBE: Yes. But only given the dirent child processes for one parent. Retrieving all processes on level x still would have you search recursively everything on layers above. But this should be linear complexity.

zhiburt:
I tried an optimization where we would rebuild children first which seemed more efficient but it turns out not (an issue where we need to update a lhs of children after we resigned parents layer (as I remember something it was related to something like that)).
So it still sort a parents first but uses `right_bound` and `left_bound`.
 
To implement any of the optimization it was necessary to have a tmp array with index of process. I decided to resolve it as a tree_index in Process structure.

The index though it's broken after sorting.
It took me 2 days of debugging .... (:disappointed:) to try to understand why we can't update a map... And I still has no idea.

Overall I am not sure if it's still correct after the optimization

ProcessList.c
zhiburt: Look at the comment.

zhiburt:
It turns out after 4 days of debug.... :disappointed:  (Yes I am so bad (kinda :smiley: ))

That there's main issues in optimization:
* we cann't rebuild a tree at the time we are traversing it. There's 2 main reason here.
    * We may shadow not traversed yet node
    * And something else which I forgot
* We need also some aditional storage for tmp variables.

That's why now I allocate a tempraly hashtable to put changes there. And I use tree_depth as temporaraly variable while traversing

BenBE:
I'll give it a go on the weekend. Will report back my findings and suggestions.

zhiburt:
ping

BenBE:
Could you rebase this onto current master? There's been some changes to API (in particular making several arguments `const`) that might otherwise accumulate merge issues later on.

BenBE:
I tried rebasing this onto current master, but this ended up with a few compile issues mid-way:

```
$ LC_ALL=C make
make  all-am
make[1]: Entering directory '/usr/src/htop'
depbase=`echo ProcessList.o | sed 's|[^/]*$|.deps/&|;s|\.o$||'`;\
gcc -DHAVE_CONFIG_H -I.  -DNDEBUG  -Wall -Wcast-align -Wcast-qual -Wextra -Wfloat-equal -Wmissing-format-attribute -Wmissing-noreturn -Wmissing-prototypes -Wpointer-arith -Wshadow -Wstrict-prototypes -Wundef -Wunused -Wwrite-strings -Wnull-dereference -pedantic -std=c99 -D_XOPEN_SOURCE_EXTENDED -DSYSCONFDIR=\"/usr/local/etc\" -I"./linux" -D_DEFAULT_SOURCE -D_XOPEN_SOURCE=600 -fsanitize=undefined -I/usr/include/libnl3 -I/usr/include/libnl3 -MT ProcessList.o -MD -MP -MF $depbase.Tpo -c -o ProcessList.o ProcessList.c &&\
ProcessList.c: In function 'ProcessList_sortTreeSetLayer':
ProcessList.c:125:30: warning: comparison of integer expressions of different signedness: 'unsigned int' and 'int' [-Wsign-compare]
  125 |       if (process->tree_deep == deep && Process_isChildOf(process, pid)) {
      |                              ^~
ProcessList.c: In function 'ProcessList_sortTreeSet':
ProcessList.c:154:8: warning: unused variable 'size' [-Wunused-variable]
  154 |    int size = ProcessList_sortTreeSetLayer(this, 0, 0);
      |        ^~~~
ProcessList.c:152:8: warning: unused variable 'vsize' [-Wunused-variable]
  152 |    int vsize = Vector_size(this->processes);
      |        ^~~~~
ProcessList.c: In function 'HashTableView':
ProcessList.c:164:54: warning: unused parameter '_' [-Wunused-parameter]
  164 | static void HashTableView(int key, void* data, void* _) {
      |                                                ~~~~~~^
ProcessList.c: In function 'ProcessList_buildTree':
ProcessList.c:225:28: error: assignment of member 'sortKey' in read-only object
  225 |    this->settings->sortKey = PID;
      |                            ^
ProcessList.c:226:30: error: assignment of member 'direction' in read-only object
  226 |    this->settings->direction = 1;
      |                              ^
ProcessList.c:229:28: error: assignment of member 'sortKey' in read-only object
  229 |    this->settings->sortKey = sortKey;
      |                            ^
ProcessList.c:230:30: error: assignment of member 'direction' in read-only object
  230 |    this->settings->direction = direction;
      |                              ^
At top level:
ProcessList.c:164:13: warning: 'HashTableView' defined but not used [-Wunused-function]
  164 | static void HashTableView(int key, void* data, void* _) {
      |             ^~~~~~~~~~~~~
make[1]: *** [Makefile:1097: ProcessList.o] Error 1
make[1]: Leaving directory '/usr/src/htop'
make: *** [Makefile:769: all] Error 2
```

The essence here is the refusal of the compiler in `ProcessList_buildTree` to change the user settings. This was introduced recently and I think having the compiler bark in this situation is actually a good sign. ;-)

Relevant location in PR source version:
https://github.com/zhiburt/htop/blob/5002e4f386681c908d959cb3a14f16eca0d549fe/ProcessList.c#L251-L256

Apart from this I noticed several issues with trailing whitespace in the patchset (didn't check if they persist in most recent commit), but would be nice to clean up the intermediate patches too.

I did a quick&dirty rebase on [my fork](https://github.com/BenBE/htop/tree/sort_in_tree_mode). Basically just corrects the issues to run this on current master. No cleanup done apart from this.

ProcessList.c
```diff
@@ -111,7 +114,90 @@ int ProcessList_size(ProcessList* this) {
    return (Vector_size(this->processes));
 }
 
-static void ProcessList_buildTree(ProcessList* this, pid_t pid, int level, int indent, int direction, bool show) {
+static void ProcessList_updateTreeSetLayer(ProcessList* this, Hashtable* newTree, unsigned int left_bound, unsigned int right_bound, unsigned int deep, unsigned int left, unsigned int right, unsigned int* index, unsigned int* tree_index, int indent) {
```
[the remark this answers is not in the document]
BenBE: Could you add this for the next poor soul trying to understand this code?

ProcessList.c
```diff
@@ -111,7 +114,90 @@ int ProcessList_size(ProcessList* this) {
...
+   // but most likely it needs less
+   int layer_size = (right - left) / 2;
+   Vector* layer = Vector_new(this->processes->type, false, layer_size);
+
+   for (unsigned int i = left_bound; i < right_bound; i++) {
+      Process* proc = (Process*)Hashtable_get(this->treeSet, i);
+      if (proc->tree_depth == deep && proc->tree_left > left && proc->tree_right < right) {
+         if (Vector_size(layer) > 0) {
+            // use as tmp variable
+            ((Process*)Vector_get(layer, Vector_size(layer)-1))->tree_depth = proc->tree_index;
```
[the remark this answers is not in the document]
zhiburt: I am not as provichient in C :disappointed: . 
I could create a variable which I did instead of macros :smile: is there any benefit in macros?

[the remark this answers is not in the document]
BenBE: Depends on the situation. In this case the macro would have been merely a short-hand for that long expression:

```c
#define LastProcess ((Process*)Vector_get(layer, Vector_size(layer)-1))
// ...
            // use as tmp variable
            LastProcess->tree_depth = proc->tree_index;
// ...
#undef LastProcess
```

But once there is a certain amount of repetition with slight variation involved, this can become [quite handy](https://github.com/htop-dev/htop/pull/229/files#r512234066):
```c
#define resolve(symbolname) do {\
      *(void **)(&sym_##symbolname) = dlsym(dlopenHandle, #symbolname); \
      if (!sym_##symbolname || dlerror() != NULL) \
         goto dlfailure;

      resolve(sd_bus_open_system);
      resolve(sd_bus_get_property_string);
      resolve(sd_bus_get_property_trivial);
      resolve(sd_bus_unref);

#undef resolve
```

Downside with macros usually is that you create repeated code that the optimizer needs to find its way around, whereas variables usually bear no additional runtime cost. One advantage with macros can be, that if written properly (and depending on what you need to do) you can write them so they work as lvalues (and thus can be assigned to) even through several layers of abstraction. For variables you'd have to handle the writing back through the abstraction yourself (depending on what the variable contains. If you can show the pointer for that `vector_get` to be constant you throughout its use here, a variable will be the better solution, if the last element in the vector may change (e.g. by adding/removing items) a macro would be preferred.

ProcessList.c
```diff
@@ -114,31 +114,66 @@ int ProcessList_size(ProcessList* this) {
    return (Vector_size(this->processes));
 }
 
+// ProcessList_updateTreeSetLayer sort this->treeSet,
+// relaying only on itself.
```
BenBE, line 118:
```suggestion
// relying only on itself.
```

ProcessList.c
```diff
@@ -114,31 +114,66 @@ int ProcessList_size(ProcessList* this) {
...
+// and at the same time make changes in it.
+//
+// It could though be removed with sort by TREE_SET if we could savely update this->processes in place.
+//
+// left_bound and right_bound mainly used as a optimization to cut the list where we are building a 'layer'.
 static void ProcessList_updateTreeSetLayer(ProcessList* this, Hashtable* newTree, unsigned int left_bound, unsigned int right_bound, unsigned int deep, unsigned int left, unsigned int right, unsigned int* index, unsigned int* tree_index, int indent) {
-   // it's guaranted that layer_size is enough space
-   // but most likely it needs less
+   // It's guaranted that layer_size is enough space
+   // but most likely it needs less. Specificaly on first iteration.
```
BenBE, line 130:
```suggestion
   // but most likely it needs less. Specifically on first iteration.
```

ProcessList.c
```diff
@@ -114,31 +114,66 @@ int ProcessList_size(ProcessList* this) {
...
    int layer_size = (right - left) / 2;
    Vector* layer = Vector_new(this->processes->type, false, layer_size);
 
+   // Find all process on the same layer (process with the same deep value
+   // and included in a range from left_bound to right_bound.
+   //
+   // Also this loops keeps track of left_bound and right_bound of these proceses to not lose this information
+   // after the list is sorted.
```
BenBE, line 138:
```suggestion
   // This loop also keeps track of left_bound and right_bound of these processes
   // in order not to lose this information once the list is sorted.
```

ProcessList.c
```diff
@@ -114,31 +114,66 @@ int ProcessList_size(ProcessList* this) {
...
+   //
+   // left_bound and right_bound doesn't represent the same what lhs and rhs represents.
+   // Here's an example why. The nested tree set sorted by left value. Which guaranted have been called when
+   // this function is called.  
+   // 
+   // i | l | r
+   // 1 | 1 | 9 
+   // 2 | 2 | 8
+   // 3 | 4 | 5
+   // 4 | 6 | 7
```
BenBE, line 148: The explanation for the example is not very intuitive.

Let me rephrase a bit, maybe it's easier to digest with another wording:

```
The variables left_bound and right_bound are different from what the values lhs and rhs represent,
While left_bound and right_bound define a range of processes to look at, the values given by lhs and rhs are indices into an array …

In the below example note how filtering a range of indices i is different from filtering for processes in the bounds left_bound < x < right_bound …

The nested tree set is sorted by left value, which is guaranteed upon entry/exit(???) of this function.
```

Maybe something more along those lines? I hope I got it somewhat correct. Also the naming of things in the comment should follow the variables used in the code to avoid confusion when trying to understand the comment.

ProcessList.c
```diff
@@ -114,31 +114,66 @@ int ProcessList_size(ProcessList* this) {
...
+            previous_process->tree_depth = proc->tree_index; 
          }
          Vector_add(layer, proc);
       }
    }
 
+   // The lope above changes process-1 so the last process on the layer
+   // isn't updated by the that loop.
+   //
+   // This if set the right_bound to the last process on the layer
```
BenBE, line 167:
```suggestion
   // The loop above changes process-1 so the last process on the layer
   // isn't updated by the that loop.
   //
   // Thus, if present, set the right_bound to the last process on the layer
```

ProcessList.c
```diff
@@ -162,15 +197,18 @@ static void ProcessList_updateTreeSetLayer(ProcessList* this, Hashtable* newTree
 
       if (indent == -1) {
          proc->indent = 0;
-      } else if (i == size - 1)
+      } else if (i == size - 1) {
          proc->indent = -currentIndent;
-      else
+      }
+      else {
```
BenBE, line 203: Should be only one line:

```suggestion
      } else {
```

BenBE (changes requested):

ProcessList.c
zhiburt: I've never have been good at spelling 😞

BenBE:
@zhiburt Please prefer rebasing commits over repeated merges of the master branch. This helps to keep the history in the PR kinda clean and the final history on the master easier to follow. A rebase also helps to minimize noise commits by fixing e.g. typos back where they were introduced. `git commit --fixup=<commit>` and `git rebase -i --autosquash <commit_or_branch_to_base_on>` are quite effective here.

ProcessList.c
zhiburt: > The nested tree set is sorted by left value, which is guaranteed upon entry/exit(???) of this function.

Yes

So as I said in the doc comment for function if there's a save way to sort `ProcessList::processes` in place we could remove newTreeSet and a sort at the end of update algorithm

zhiburt: I could you've got more versed in these things so if you could `suggest` your variant of a comment I could commit it 😄

zhiburt:
Can I ask you @BenBE.

Is the algorithm and the PR overall meets your initial idea of usage a nested tree set? 
And do you understand why we actually use a temporally Hash-table? 

And according to this table we could actually keep it in ProcessList as well to reduce amount of allocations but it would increase a general size of a structure.

ProcessList.c
BenBE: The only in-place variant to sort I can think of right now would cause sorting to be O(n²) or worse …

To avoid having to update tree_left/tree_right for each process (and its tree children) for each move, just track how much each parent is moved starting from a first, consistent state to reach its final place. In a final run over all processes sum up the offsets for each process' parents and update tree_left/tree_right accordingly.

Given:
```
pid | ppid | depth |  left | right | move | xpnd
------------------------------------------------
  1 |    0 |     0 |     1 |     2 |    0 |    0
  2 |    1 |     0 |     3 |     4 |    0 |    0
  3 |    2 |     0 |     5 |     6 |    0 |    0
  4 |    1 |     0 |     7 |     8 |    0 |    0
  5 |    2 |     0 |     9 |    10 |    0 |    0
```

Wanted:
```
pid | ppid | depth |  left | right | move | xpnd
------------------------------------------------
  1 |    0 |     0 |     1 |    10 |    0 |    4
  2 |    1 |     1 |     2 |     7 |   -1 |    4
  3 |    2 |     2 |     3 |     4 |   -1 |    0
  4 |    1 |     1 |     8 |     9 |   -3 |    0
  5 |    2 |     2 |     5 |     6 |   -3 |    0
```

move: distance moved relative to parent, absolute move distance results from adding move value including all parent's.
xpnd: change of distance between left and right. total change by adding xpnd from all children.

Moving one process at a time to the proper place yields said O(n²) behaviour. You can do better, if you can predict the correct insert location for each process by:
1. only inserting at the end of a child list
2. consolidating that sort order (eliminates xpnd variation, resets move)
3. sort sub-lists by determining moves (updates move only)
4. consolidate movements

Consolidation can always be done in linear time if you iterate the structure and have consistent move/xpnd values to work with. The complexity goes into preparing those values.

Wondering how much this already resembles the current PRs implementation.

BenBE:
> Can I ask you @BenBE.

Sure.

> Is the algorithm and the PR overall meets your initial idea of usage a nested tree set?
> And do you understand why we actually use a temporally Hash-table?

I think the PR is going in the right direction. See my move/xpnd comment above. Hope that comment explains a bit better, what my idea with the nested set is and where it helps with the sorting/reordering of items.

For the Hashtable: Basically to re-fill the tree structure afresh and keeping track of which sorted processes have already been re-added. (If I understood the flow correctly).

> And according to this table we could actually keep it in ProcessList as well to reduce amount of allocations but it would increase a general size of a structure.

Could you please elaborate? I'm not sure I can follow you with your thoughts.

zhiburt:
> Could you please elaborate? I'm not sure I can follow you with your thoughts.

I meant to say that currently a temporally Hashtable allocated on each call of `ProcessList_updateTreeSet`.
It could be allocated only once if we would keep it inside `ProcessList`

ProcessList.c
zhiburt: Aa maybe there's something in keeping track of processing by saving offset value.
But currently nothings like that there 😄 . Even though I think that current implementation is generally faster then O(n²) as we reduce an process array considerably by using `left_bound`, `right_stuff`.

Is it save to remove elements from `this->processes`? If it is sorting the array at the end could be deleted.

BenBE:
> > Could you please elaborate? I'm not sure I can follow you with your thoughts.
> 
> I meant to say that currently a temporally Hashtable allocated on each call of `ProcessList_updateTreeSet`.
> It could be allocated only once if we would keep it inside `ProcessList`

The benefit of saving several memory management calls per iteration is greater than those few bytes required for one pointer (and the hashtable memory, we'd need anyway.

So I'd suggest putting that hashmap into the process list as needed.

ProcessList.h
```diff
@@ -43,6 +43,7 @@ typedef struct ProcessList_ {
    UsersTable* usersTable;
 
    Hashtable* treeSet;
+   Hashtable* treeSet2;
```
BenBE: Could you use consistent naming (both with a number), but preferably with the name hinting at their actual use?

zhiburt: Nothing better then `tmpTreeSet` is comming to my ming :smile:

zhiburt:
ping

BenBE:
What about naming them `displayTreeSet` and `draftingTreeSet`? That way their name roughtly conveys what the are used for and the naming is mostly consistent (`treeSet1`+`treeSet2` would have been an option all along, just with much less intuition what these variables actually do. HTH.

On the next steps: Cleaning the naming, maybe look for bottlenecks where cycles could be saved (optional), document the algorithm on some high (abstract) level (have a look at [this comment](https://github.com/BenBE/htop/blob/improved-command-display/linux/LinuxProcessList.c#L900-L911) for an example how this might look like), and finally rebase everything to master (I don't see any merge conflicts right now, thus should be fine).

zhiburt:
rebased in #240 (The best how I could handle this :( )

ProcessList.c
```diff
@@ -116,22 +116,29 @@ int ProcessList_size(ProcessList* this) {
...
+//
+// The algorithm is based on `Depth-first search`,
+// even though `Breadth-first search` approach may be more eficient from first perspetive,
+// after comparision it may be not. As it's not save to not update the tree structure after getting deeper.
+// If it would be save that approach would likely bring an advantage in performance.
 //
-// It could though be removed with sort by TREE_SET if we could savely update this->processes in place.
+// Each call of the function looks for a 'layer'. A 'layer' is a list of processes with the same deep.
+// It sorts a list. Then it runs the function recursevely for each element of sorted list.
+// And then it updates settings of processes.
```
BenBE, line 131:
```suggestion
// First it sorts a list. Then it runs the function recursively for each element of the sorted list.
// After that it updates the settings of processes.
```
With settings you probably mean "position"/"order" kind of information of each process?

ProcessList.c
```diff
@@ -116,22 +116,29 @@ int ProcessList_size(ProcessList* this) {
...
 //
-// left_bound and right_bound mainly used as a optimization to cut the list where we are building a 'layer'.
-static void ProcessList_updateTreeSetLayer(ProcessList* this, unsigned int left_bound, unsigned int right_bound, unsigned int deep, unsigned int left, unsigned int right, unsigned int* index, unsigned int* tree_index, int indent) {
+// It relays on left_bound and right_bound as a optimization to cut the list size at the time it build a 'layer'.
```
BenBE, line 133:
```suggestion
// It relies on `leftBound` and `rightBound` as an optimization to cut the list size at the time it builds a 'layer'.
```
Relay: forward something to somewhere else
Rely: Need something to do something else.

Also updated spelling of the parameters.

ProcessList.c
```diff
@@ -116,22 +116,29 @@ int ProcessList_size(ProcessList* this) {
...
+//
+// It use a temporaraly Hashtable `draftingTreeSet` because it's not save to traverse a tree
```
BenBE, line 135:
```suggestion
// It uses a temporary Hashtable `draftingTreeSet` because it's not save to traverse a tree
```

ProcessList.c
```diff
@@ -116,22 +116,29 @@ int ProcessList_size(ProcessList* this) {
...
+// and at the same time make changes in it.
+static void ProcessList_updateTreeSetLayer(ProcessList* this, unsigned int leftBound, unsigned int rightBound, unsigned int deep, unsigned int left, unsigned int right, unsigned int* index, unsigned int* treeIndex, int indent) {
    // It's guaranted that layer_size is enough space
    // but most likely it needs less. Specifically on first iteration.
-   int layer_size = (right - left) / 2;
-   Vector* layer = Vector_new(this->processes->type, false, layer_size);
+   int layerSize = (right - left) / 2;
+   Vector* layer = Vector_new(this->processes->type, false, layerSize);
 
    // Find all process on the same layer (process with the same deep value
```
BenBE, line 143:
```suggestion
   // Find all processes on the same layer (process with the same `deep` value
```

ProcessList.c
```diff
@@ -116,22 +116,29 @@ int ProcessList_size(ProcessList* this) {
...
    // and included in a range from left_bound to right_bound.
```
BenBE, line 144:
```suggestion
   // and included in a range from `leftBound` to `rightBound`.
```

ProcessList.c
```diff
@@ -172,7 +179,7 @@ static void ProcessList_updateTreeSetLayer(ProcessList* this, unsigned int left_
    // Thus, if present, set the right_bound to the last process on the layer
```
BenBE, line 179:
```suggestion
   // Thus, if present, set the `rightBound` to the last process on the layer
```
Probably needs to be changed in some more places inside comments.

BenBE (changes requested):

BenBE:
> rebased in #307 (The best how I could handle this :( I am not sure if I could somehow reopen that here)

I commented over there with a small howto to get things onto the correct branch again. Hope this helps.

I'd prefer to continue the discussion here for ease, but if there's no luck getting things back on track properly we can do so there too as a last resort …

ProcessList.c
zhiburt: yes `lhs`, `rhs`, `deep` fields

zhiburt:
I hope that what you were looking for @BenBE ?

zhiburt:
ping

BenBE:
Currently crashes right on startup:

```
htop: Vector.c:24: Vector_new: Assertion `size > 0' failed.

Program received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50      ../sysdeps/unix/sysv/linux/raise.c: Datei oder Verzeichnis nicht gefunden.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7248859 in __GI_abort () at abort.c:79
#2  0x00007ffff7248729 in __assert_fail_base (fmt=0x7ffff73de588 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", 
    assertion=0x555555633429 "size > 0", file=0x555555633420 "Vector.c", line=24, function=<optimized out>) at assert.c:92
#3  0x00007ffff7259f36 in __GI___assert_fail (assertion=0x555555633429 "size > 0", file=0x555555633420 "Vector.c", line=24, 
    function=0x5555556335e0 <__PRETTY_FUNCTION__.5843> "Vector_new") at assert.c:101
#4  0x0000555555607b7a in Vector_new (type=0x55555563ef80 <LinuxProcess_class>, owner=false, size=0) at Vector.c:24
#5  0x00005555555f3d85 in ProcessList_updateTreeSetLayer (this=0x5555556a1650, leftBound=1227, rightBound=1228, deep=1, left=2455, 
    right=2456, index=0x7fffffffdb80, treeIndex=0x7fffffffdb84, indent=0) at ProcessList.c:158
#6  0x00005555555f4409 in ProcessList_updateTreeSetLayer (this=0x5555556a1650, leftBound=0, rightBound=1530, deep=0, left=0, 
    right=3061, index=0x7fffffffdb80, treeIndex=0x7fffffffdb84, indent=-1) at ProcessList.c:218
#7  0x00005555555f4a32 in ProcessList_updateTreeSet (this=0x5555556a1650) at ProcessList.c:251
#8  0x00005555555f68bf in ProcessList_sort (this=0x5555556a1650) at ProcessList.c:401
#9  0x00005555555fbe2f in checkRecalculation (this=0x5555556a2ab0, oldTime=0x7fffffffdcd0, sortTimeout=0x7fffffffdcac, 
    redraw=0x7fffffffdca9, rescan=0x7fffffffdcaa, timedOut=0x7fffffffdca8) at ScreenManager.c:117
#10 0x00005555555fc624 in ScreenManager_run (this=0x5555556a2ab0, lastFocus=0x0, lastKey=0x0) at ScreenManager.c:168
#11 0x00005555555d56af in main (argc=1, argv=0x7fffffffdf08) at htop.c:361
(gdb)
```

To reproduce:
1. Build with debug asserts: `make clean all AM_CPPFLAGS="-ggdb -DDEBUG"`
2. Run in gdb: `gdb --args ./htop` --> `r`
3. Get stacktrace with `bt`

zhiburt:
When I was fixing it I've noticed that it `this-processes` list updated somewhere but no in `ProcessList_scan`
So I deleted some debug lines which was affected by this fact.

The fix of the issue you have bump into is here.

https://github.com/htop-dev/htop/blob/21c3164c34d2d9eb9a8483a67ab11e59dbe55e57/ProcessList.c#L163-L166

ProcessList.c
```diff
@@ -128,7 +137,131 @@ int ProcessList_size(ProcessList* this) {
...
+   Vector_delete(layer);
+}
+
+static void ProcessList_updateTreeSet(ProcessList* this) {
+   unsigned int index = 0;
+   unsigned int tree_index = 1;
+   int vsize = Vector_size(this->processes);
+
+   ProcessList_updateTreeSetLayer(this, 0, vsize, 0, 0, vsize*2+1, &index, &tree_index, -1);
+   assert((int)Hashtable_count(this->draftingTreeSet) == vsize);
```
BenBE, line 256: Why has the other post-exec assertion on `this->displayTreeSet == 0` been removed? Where and why is it violated?

I know that your previously used assertion on `Hashtable_count(this->displayTreeSet) == vsize` does not hold for new/dead processes since last scan, but at least the `this->draftingTreeSet` should be empty before the ProcessList_updateTreeSetLayer` call; or am I missing something?

zhiburt: > Why has the other post-exec assertion on this->displayTreeSet == 0 been removed? Where and why is it violated?

As I've said.

> When I was fixing it I've noticed that for some reason which I didn't know about this-processes list updated somewhere but no in ProcessList_scan
So I deleted some debug lines which is affected by this fact.

But I am not sure where.

> I know that your previously used assertion on Hashtable_count(this->displayTreeSet) == vsize does not hold for new/dead processes since last scan, but at least the this->draftingTreeSet should be empty before the ProcessList_updateTreeSetLayer` call; or am I missing something?

Correct I am take it correctly.

It was removed by the same reason as above.

Essentially  displayTreeSet  `sometimes` is not equal to 0 after calling updateTreeSetLayer  as well as `sometimes` `Hashtable_count(this->displayTreeSet) == vsize` isn't true because on in the beggining of function `Vector_size(this->processes) != Hashtable_count(this->displayTreeSet)` which is I am not sure why.

ProcessList.c
```diff
@@ -128,7 +137,131 @@ int ProcessList_size(ProcessList* this) {
...
+
+   for(int i = 0; i < vsize; i++) {
+      Process* proc = (Process*)Hashtable_remove(this->draftingTreeSet, i);
+      Hashtable_put(this->displayTreeSet, i, proc);
+   }
```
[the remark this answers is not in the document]
zhiburt: Sorry for the lack of knowledge I have but I hardly see what sort of swap I mean here 😞

ProcessList.c
zhiburt: BTW ` this->displayTreeSet == 0` shouldn't related to that. I think.

ProcessList.c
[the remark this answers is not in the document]
BenBE: If I understand the intention of that loop correctly AND if `this->displayTreeSet` is empty before that loop, you can achieve the same effect instead of copying all entries like this:

```suggestion
   Hashtable* tmp = this->draftingTreeSet;
   this->draftingTreeSet = this->displayTreeSet;
   this->displayTreeSet = tmp;
```

ProcessList.c
zhiburt: I've found the issue `ProcessList_buildTreeBranch` seems sometimes addes copies on initial build

zhiburt: I seems to be fixed

zhiburt:
The issues seems to be fixed I restore the asserts which initially have been present.
I also run it in a debug mode seems OK

ProcessList.c
```diff
@@ -128,7 +137,137 @@ int ProcessList_size(ProcessList* this) {
...
+      if (indent == -1) {
+         proc->indent = 0;
+      } else if (i == size - 1) {
+         proc->indent = -currentIndent;
+      } else {
+         proc->indent = currentIndent;
+      }
+
+      Hashtable_put(this->draftingTreeSet, proc->tree_index, proc);
+      // // it's unnecesary to do but later we can check that
```
BenBE, line 245: No need to double-comment comments … One pair of slashes fully suffices.

ProcessList.c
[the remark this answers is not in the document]
zhiburt: Seems perfect solution.

ProcessList.c
```diff
@@ -8,12 +8,16 @@ in the source distribution for its full text.
...
+#include <stdlib.h>
 #include <string.h>
 #include <time.h>
 
 #include "CRT.h"
+#include "Hashtable.h"
+#include "Vector.h"
 #include "XUtils.h"
 
+#include <stdio.h>
```
cgzones, line 20: Merge into above system includes

ProcessList.c
```diff
@@ -25,6 +29,9 @@ ProcessList* ProcessList_init(ProcessList* this, const ObjectClass* klass, Users
    // tree-view auxiliary buffer
    this->processes2 = Vector_new(klass, true, DEFAULT_SIZE);
 
+   this->displayTreeSet = Hashtable_new(4096, false);
```
cgzones, line 32: Probably use better initial size than 4096 (`Vector_size(this->processes)  0.7 /* load factor */` ?)

ProcessList.c
```diff
@@ -170,73 +320,96 @@ static long ProcessList_treeProcessCompare(const void* v1, const void* v2) {
    const Process *p1 = (const Process*)v1;
    const Process *p2 = (const Process*)v2;
 
-   return p1->pid - p2->pid;
+   // return p1->tree_left - p2->tree_left;
+   return (p1->tree_left < p2->tree_left ? -1 : 1);
```
cgzones, line 324: `SPACESHIP_NUMBER()` ?

ProcessList.c
```diff
@@ -170,73 +320,96 @@ static long ProcessList_treeProcessCompare(const void* v1, const void* v2) {
...
-               process = (Process*)(Vector_take(this->processes, i));
-               process->indent = 0;
-               Vector_add(this->processes2, process);
-               ProcessList_buildTree(this, process->pid, 0, 0, direction, process->showChildren);
+static long ProcessList_treeProcessCompareByPID(const void* v1, const void* v2) {
+   const Process *p1 = (const Process*)v1;
+   const Process *p2 = (const Process*)v2;
+
+   // return p1->tree_left - p2->tree_left;
+   return (p1->pid - p2->pid);
```
cgzones, line 332: dito?

ProcessList.c
```diff
@@ -364,4 +537,14 @@ void ProcessList_scan(ProcessList* this, bool pauseProcessUpdate) {
...
       }
    }
+
+   // clean up hashtable in order to not have
+   // processes from privious build in it
+   //
+   // sorting algorithm relayes on the fact that
+   // len(this->displayTreeSet) == len(this->processes)
+   Hashtable_delete(this->displayTreeSet);
+   this->displayTreeSet = Hashtable_new(4096, false);
```
cgzones, line 547: Maybe introduce `Hashtable_clear()` instead of delete + new 

```c
void Hashtable_clear(Hashtable* this) {
   for (int i = 0; i < this->size; i++) {
      if (this->owner) {
            free(this->buckets[i].value);
      this->buckets[i].value = NULL;
   }
   this->items = 0;
}
```

ProcessList.c
```diff
@@ -364,4 +537,14 @@ void ProcessList_scan(ProcessList* this, bool pauseProcessUpdate) {
...
+
+   ProcessList_buildTree(this);
```
cgzones, line 549: I guess this should not be called in sorted-mode.

ProcessList.c
BenBE: Probably `Vector_size(this->processes) * 1.5` (`size = Vector_size(this->processes; size += size >> 1;`)

ProcessList.c
cgzones: `Hashtable_delete` does delete the buckets (and `Hashtable_new` creates them). `Hashtable_clear` would not deallocate and allocate them.

ProcessList.c
```diff
@@ -128,7 +137,136 @@ int ProcessList_size(ProcessList* this) {
...
+   Vector_quickSort(layer);
+
+   int size = Vector_size(layer);
+   for (int i = 0; i < size; i++) {
+      Process* proc = (Process*) (Vector_get(layer, i));
+
+      unsigned int idx = (*index)++;
+      int newLeft = (*treeIndex)++;
+
+      int level = deep-1;
```
cgzones, line 221: Not UB, but shows up with sanitizers:
```
ProcessList.c:221:23: runtime error: unsigned integer overflow: 0 - 1 cannot be represented in type 'unsigned int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ProcessList.c:221:23 in 
ProcessList.c:221:19: runtime error: implicit conversion from type 'unsigned int' of value 4294967295 (32-bit, unsigned) to type 'int' changed the value to -1 (32-bit, signed)
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ProcessList.c:221:19 in
```

```c
int level = (int)deep - 1;
```

ProcessList.c
zhiburt: I probably haven't get the question correctly but it should as here me clean the previous hashtable in order to build new one and don't have any artifacts from old one.

Imagine the first build have 100 processes but the second 50. If we wouldn't clean this up we would have a incorrect hashtable

ProcessList.c
zhiburt: It makes sense

zhiburt:
I'd like to point out that likely I'll not able to resolve the issues that may appear and the ones which aren't resolved so far in the nearest.

If you think it appropriate you can merge it on your own.
Otherwise I am trying to do it in the nearest feature.

ProcessList.c
zhiburt: What it actually means?
I am not very educated in these acronyms :(

ProcessList.c
zhiburt: What is your proposal actually is?
Sorry for my lack of knowledge.

cgzones: There is a helper macro https://github.com/htop-dev/htop/blob/95f553b10c3276c731bb7e3d9a917e9d98322e85/Macros.h#L23 which does a [three way comparison](https://en.wikipedia.org/wiki/Three-way_comparison) on numbers.

ProcessList.c
cgzones: In *sorted*-mode (opposed to *tree*-mode (toggle-able with F5)) I think we do not need to build the tree (cause it is not used).

ProcessList.c
BenBE: dito means "the same as before". In this case: to use the `SPACESHIP_NUMBER` macro as with the previous functions.

ProcessList.c
BenBE: Upstreamed to master.

ProcessList.c
BenBE: All assertions in this context restored and checked with the code.

BenBE:
If I didn't miss anything while working through all the comments and unresolved discussions in this thread the PR should now be in a state for review.

ProcessList.c
```diff
@@ -122,26 +134,165 @@ void ProcessList_remove(ProcessList* this, Process* p) {
...
+}
+
+// ProcessList_updateTreeSetLayer sorts this->displayTreeSet,
+// relying only on itself.
+//
+// Algorithm
+//
+// The algorithm is based on `depth-first search`,
+// even though `breadth-first search` approach may be more efficient on first glance,
+// after comparision it may be not, as it's not save to go deeper without first updating the tree structure.
```
natoscott, line 151: save -> safe?

ProcessList.c
```diff
@@ -122,26 +134,165 @@ void ProcessList_remove(ProcessList* this, Process* p) {
...
+// If it would be save that approach would likely bring an advantage in performance.
```
natoscott, line 152: save -> safe?

ProcessList.c
```diff
@@ -122,26 +134,165 @@ void ProcessList_remove(ProcessList* this, Process* p) {
...
+//
+// Each call of the function looks for a 'layer'. A 'layer' is a list of processes with the same depth.
+// First it sorts a list. Then it runs the function recursively for each element of the sorted list.
+// After that it updates the settings of processes.
+//
+// It relies on `leftBound` and `rightBound` as an optimization to cut the list size at the time it builds a 'layer'.
+//
+// It uses a temporary Hashtable `draftingTreeSet` because it's not save to traverse a tree
```
natoscott, line 160: save -> safe?

ProcessList.c
```diff
@@ -122,26 +134,165 @@ void ProcessList_remove(ProcessList* this, Process* p) {
...
+// and at the same time make changes in it.
+static void ProcessList_updateTreeSetLayer(ProcessList* this, unsigned int leftBound, unsigned int rightBound, unsigned int deep, unsigned int left, unsigned int right, unsigned int* index, unsigned int* treeIndex, int indent) {
+   // It's guaranted that layer_size is enough space
```
natoscott, line 163: guaranteed

ProcessList.c
```diff
@@ -122,26 +134,165 @@ void ProcessList_remove(ProcessList* this, Process* p) {
...
+   // but most likely it needs less. Specifically on first iteration.
+   int layerSize = (right - left) / 2;
+
+   // check if we reach `children` of `leafes`
```
natoscott, line 167: leaves?

ProcessList.c
```diff
@@ -122,26 +134,165 @@ void ProcessList_remove(ProcessList* this, Process* p) {
...
+   if (layerSize == 0)
+      return;
+
+   Vector* layer = Vector_new(this->processes->type, false, layerSize);
+
+   // Find all processes on the same layer (process with the same `deep` value
+   // and included in a range from `leftBound` to `rightBound`.
```
natoscott, line 174: No closing brace.

ProcessList.c
```diff
@@ -122,26 +134,165 @@ void ProcessList_remove(ProcessList* this, Process* p) {
...
+   //
+   // This loop also keeps track of left_bound and right_bound of these processes
+   // in order not to lose this information once the list is sorted.
+   //
+   // The variables left_bound and right_bound are different from what the values lhs and rhs represent,
+   // While left_bound and right_bound define a range of processes to look at, the values given by lhs and rhs are indices into an array
```
natoscott, line 180: this comment line ends in a comma but next line is capitalised like start of a new sentence.

ProcessList.c
```diff
@@ -122,26 +134,165 @@ void ProcessList_remove(ProcessList* this, Process* p) {
...
+   //
+   // In the below example note how filtering a range of indices i is different from filtering for processes in the bounds left_bound < x < right_bound …
```
natoscott, line 182: "i is" - should that be just "is"?

ProcessList.c
```diff
@@ -122,26 +134,165 @@ void ProcessList_remove(ProcessList* this, Process* p) {
...
+// If it would be save that approach would likely bring an advantage in performance.
+//
+// Each call of the function looks for a 'layer'. A 'layer' is a list of processes with the same depth.
+// First it sorts a list. Then it runs the function recursively for each element of the sorted list.
+// After that it updates the settings of processes.
+//
+// It relies on `leftBound` and `rightBound` as an optimization to cut the list size at the time it builds a 'layer'.
+//
+// It uses a temporary Hashtable `draftingTreeSet` because it's not save to traverse a tree
+// and at the same time make changes in it.
```
natoscott, line 161: maybe insert another "blank" comment line here so the function name and parameters are a bit more visible?  *shrug*

ProcessList.c
```diff
@@ -122,26 +134,165 @@ void ProcessList_remove(ProcessList* this, Process* p) {
...
+   for (unsigned int i = leftBound; i < rightBound; i++) {
+      Process* proc = (Process*)Hashtable_get(this->displayTreeSet, i);
+      if (proc->tree_depth == deep && proc->tree_left > left && proc->tree_right < right) {
+         if (Vector_size(layer) > 0) {
+            Process* previous_process = (Process*)Vector_get(layer, Vector_size(layer) - 1);
+
+            // Make a 'right_bound' of previous_process in a layer the current process's index.
+            //
+            // Use 'tree_depth' as a temporal variable.
+            // it is save to do as later 'tree_depth' will be renovated.
```
natoscott, line 200: save -> safe, missing capitilisation.

ProcessList.c
```diff
@@ -122,26 +134,165 @@ void ProcessList_remove(ProcessList* this, Process* p) {
...
+            previous_process->tree_depth = proc->tree_index;
+         }
+
+         Vector_add(layer, proc);
+      }
+   }
+
+   // The loop above changes process-1 so the last process on the layer
+   // isn't updated by the that loop.
```
natoscott, line 209: "the that"?

ProcessList.c
```diff
@@ -122,26 +134,165 @@ void ProcessList_remove(ProcessList* this, Process* p) {
...
+      proc->tree_depth = deep;
+
+      if (indent == -1) {
+         proc->indent = 0;
+      } else if (i == size - 1) {
+         proc->indent = -currentIndent;
+      } else {
+         proc->indent = currentIndent;
+      }
+
```
natoscott, line 248: So much braces.  ;-)

ProcessList.c
```diff
@@ -122,26 +134,165 @@ void ProcessList_remove(ProcessList* this, Process* p) {
...
+      Hashtable_put(this->draftingTreeSet, proc->tree_index, proc);
+
+      // It's not strictly necessary to do this, but doing so anyways
+      // allows for checking the correctness of the inner workings.
+      Hashtable_remove(this->displayTreeSet, newLeftBound);
```
natoscott, line 253: +1 - good idea.

natoscott (approved):

ProcessList.c
BenBE: The i is a reference to the column heading `i` below in the example table.

MainPanel.c
```diff
@@ -27,10 +27,8 @@ void MainPanel_updateTreeFunctions(MainPanel* this, bool mode) {
    FunctionBar* bar = MainPanel_getFunctionBar(this);
    if (mode) {
```
cgzones, line 28: maybe (but not a must)
```c
FunctionBar_setLabel(bar, KEY_F(5), mode ? "Sorted", "Tree  ");
```

ProcessList.c
```diff
@@ -8,22 +8,28 @@ in the source distribution for its full text.
...
 #include <time.h>
 
 #include "CRT.h"
+#include "Hashtable.h"
+#include "Vector.h"
 #include "XUtils.h"
 
 
 ProcessList* ProcessList_init(ProcessList* this, const ObjectClass* klass, UsersTable* usersTable, Hashtable* pidMatchList, uid_t userId) {
    this->processes = Vector_new(klass, true, DEFAULT_SIZE);
```
cgzones, line 22: Should we also start with a Vector size of 200 (instead of DEFAULT_SIZE=10)?

ProcessList.c
```diff
@@ -171,73 +331,113 @@ static long ProcessList_treeProcessCompare(const void* v1, const void* v2) {
...
+   this->processes = this->processes2;
+   this->processes2 = t;
+
+   // Check consistency of the built structures
+   assert(Vector_size(this->processes) == vsize); (void)vsize;
+   assert(Vector_size(this->processes2) == 0);
 }
 
+void ProcessList_sort(ProcessList* this) {
+   if (!this->settings->treeView) {
```
cgzones, line 434: maybe swap the branches to avoid the negation?

cgzones (commented):
LGTM :+1:

ProcessList.c
BenBE: It's rare we have soo few processes on a system … Maybe increase `DEFAULT_SIZE`? But that's probably something for a follow-up patchset.

BenBE (approved):

Issue #74, opened 2020-09-02
Feature Request: Allow for marking of new/old entries


BenBE:
Repost of https://github.com/hishamhm/htop/issues/837

Short description:
Allow for marking new/old entries in the process view similar to what SysInternals Process Explorer does.

Screenshot:
![](https://messengergeek.files.wordpress.com/2018/02/processesstartandstop.gif)

Ckath:
are these just cleared/added ever redraw? or is there some delay/decay time(either in ms or amount of redraws later)?

BenBE:
A freshly added process entry is marked green for a configurable time (default 5 seconds IIRC, configurable up to 10 seconds). An entry of a process that just died is marked red and kept there for the same time, before it vanishes.

The redraw delay is independent of that highlighting time.
But in the case of htop it could well be just synced to redraws.

PR #241, opened 2020-10-10
Highlight new and old processes (#74)

adsr:
(Hacktoberfest entry.) Relies on `starttime_ctime` which does not appear to be set in Linux. (The `starttime` field in `/proc/<pid>/stat` is not immediately useable.)


htop.c
```diff
@@ -48,6 +48,7 @@ static void printHelpFlag(void) {
          "-u --user[=USERNAME]        Show only processes for a given user (or $USER)\n"
          "-U --no-unicode             Do not use unicode but plain ASCII\n"
          "-V --version                Print version info\n"
+         "-I --highlight-new-old=N    Highlight new and old processes for N seconds\n"
```
fasterit, line 51: How about -H? (-I seems non-intuitive to me)

ProcessList.c
```diff
@@ -300,9 +301,22 @@ void ProcessList_scan(ProcessList* this) {
...
+      if (p->tomb_time > 0) {
+         // remove tombed process
+         if (time(NULL) >= p->tomb_time) {
+            ProcessList_remove(this, p);
+         }
+      } else if (p->updated == false) {
+         // process no longer exists
+         if (this->settings->highlightNewOldNsec > 0) {
+            // mark tombed
+            p->tomb_time = time(NULL) + this->settings->highlightNewOldNsec;
```
fasterit, line 313: This will fail in case of leap seconds and adjtime calls.
Is that corner case enough to care for something like htop (i.e. use difftime instead)?

Process.c
```diff
@@ -408,6 +414,17 @@ void Process_toggleTag(Process* this) {
    this->tag = this->tag == true ? false : true;
 }
 
+bool Process_isNew(const Process* this, const time_t now) {
+    if (this->starttime_ctime > 0) {
+        return (now - this->starttime_ctime) <= 5;
```
fasterit, line 419: Same as the comment on the tomb_time calculation (not time jump safe as not using monotonic clock and safe time calculation).

fasterit (changes requested):
General change requests:

- This should also be a setting in the setup menu
- This should highlight the whole line like the search bar and the current process and not only the length of the text

Please also document the command line flag (preferably -H, see above) in the main page (`htop.1.in`).

htop.c
```diff
@@ -189,6 +193,10 @@ static CommandLineSettings parseArguments(int argc, char** argv) {
 
             break;
          }
+         case 'I': {
+            flags.highlightNewOldNsec = optarg ? atoi(optarg) : 0;
```
fasterit, line 197: This will set highlightNewOldNsec = -1 on invalid numeric input (e.g. to large numbers, or negative numbers).
Some basic error handling as in the -d (--delay) case would be nice.

ProcessList.c
BenBE: Prefer [https://linux.die.net/man/3/clock_gettime](`clock_gettime`9 with `CLOCK_MONOTONIC`. This avoids the leap second issue AND is intended for such cases.

Process.c
BenBE: Better have an internal field in htop marking when htop first saw each process (using some monotonic timer).

adsr:
@fasterit @BenBE Thanks for the review. I addressed most of the review comments in cf1f9a7024fab14e15bd89ab82c5b07bbc651c19. Remaining work:

* Highlight full line
* Avoid _scanTs global
* Prevent highlighting processes on start-up more elegantly
* Add configurable highlight time

The first three will take some refactoring I think. Pausing here for review.

Process.h
```diff
@@ -26,6 +26,7 @@ in the source distribution for its full text.
 #include "Object.h"
 
 #define PROCESS_FLAG_IO 0x0001
+#define HIGHLIGHT_NSEC 5
```
BenBE, line 29: Shouldn't this be `HIGHLIGHT_SECS` as we are highlighting a bit™ longer than 5 nanoseconds?

ProcessList.h
```diff
@@ -97,4 +99,6 @@ Process* ProcessList_getProcess(ProcessList* this, pid_t pid, bool* preExisting,
 
 void ProcessList_scan(ProcessList* this);
 
+extern long _scanTs;
```
cgzones, line 102: Please don't use a leading underscore (maybe `ProcessList_scanTs`?)

Process.h
adsr: Haha, I use `nsec` as in `nbytes`. I'll change it.

ProcessList.h
adsr: > Please don't use a leading underscore (maybe `ProcessList_scanTs`?)

Ack, will change.

> Probably the type should also be y2038 safe on 32-bit architectures -> time_t/uint64_t?

I think any of these types will be fine. I used long because we assign it to timespec.tv_sec which is a long.

whthT (approved):

BenBE:
What's the difference between the global ProcessList_scanTs and the scanTs value in the ProcessList structure?

adsr:
> What's the difference between the global ProcessList_scanTs and the scanTs value in the ProcessList structure?

I removed the global in 7caabfc.

adsr:
Hello. Can someone add a `hacktoberfest-accepted` label to this issue so that it counts toward Hacktoberfest?

fasterit:
> @fasterit @BenBE Thanks for the review. I addressed most of the review comments in [cf1f9a7](https://github.com/htop-dev/htop/commit/cf1f9a7024fab14e15bd89ab82c5b07bbc651c19). Remaining work:
> 
>     * Highlight full line
>     * Avoid _scanTs global
>     * Prevent highlighting processes on start-up more elegantly
>     * Add configurable highlight time
> The first three will take some refactoring I think. Pausing here for review.

What is the status of these? I.e. why don't you finish the PR so we can merge it?

adsr:
> * Avoid _scanTs global

This is resolved by 7caabfc.

> * Add configurable highlight time

I can add this if you'd like. 

> * Highlight full line
> * Prevent highlighting processes on start-up more elegantly

I believe these should be tracked in separate issues as they require heavier refactoring.

fasterit:
> > * Add configurable highlight time
> I can add this if you'd like.

Please do.

> > * Highlight full line
> > * Prevent highlighting processes on start-up more elegantly
> 
> I believe these should be tracked in separate issues as they require heavier refactoring.

That doesn't make sense as we won't merge it as is.
So feel free to push to this branch of yours.

Process.h
```diff
@@ -60,6 +62,7 @@ typedef struct ProcessPidColumn_ {
 
 typedef struct Process_ {
    Object super;
+   struct ProcessList_* processList;
```
BenBE, line 65: As this is non-owning it should probably be `const`.

htop.c
```diff
@@ -48,6 +48,7 @@ static void printHelpFlag(void) {
          "-u --user[=USERNAME]        Show only processes for a given user (or $USER)\n"
          "-U --no-unicode             Do not use unicode but plain ASCII\n"
          "-V --version                Print version info\n"
+         "-H --highlight-new-old      Highlight new and old processes\n"
```
BenBE, line 51: Not quite happy with `--highlight-new-old` yet. What about something like `--highlight-diff`?

BenBE: Also this could take the optional delay to keep things highlighted, default as configured in `.htoprc` if not given.

BenBE (changes requested):
While testing the branch recently I noticed the highlighted line markup doesn't reach the end of the line. This is kinda irritating. Also with 5 seconds it feels kinda hectic, thus having a configuration for this would be nice.

Apart from this, this PR should be rebased onto current master to resolve the merge conflicts..

htop.c
fasterit: > Not quite happy with `--highlight-new-old` yet. What about something like `--highlight-diff`?

or `--highlight-changes`

fasterit:
Still working on this @adsr? Need any help?

Process.h
adsr: Fixed

htop.c
adsr: Changed to `--highlight-changes`

adsr:
The flag is now `--highlight-changes[=DELAY]`. Settings get written to htoprc. Full line gets highlighted.

fasterit (approved):
Very nice, thank you very much @adsr.

Just as a comment:
As we save `--delay` values in `htoprc` and not allow setting them via the setup GUI, this seems fine for `--highlight-changes` as well.
Medium term we may want a setup screen for these numeric values as well.

Process.c
```diff
@@ -6,6 +6,7 @@ Released under the GNU GPLv2, see the COPYING file
 in the source distribution for its full text.
 */
 
+
```
BenBE, line 9: No need for two blanks here.

Process.c
```diff
@@ -413,6 +420,16 @@ void Process_toggleTag(Process* this) {
    this->tag = this->tag == true ? false : true;
 }
 
+bool Process_isNew(const Process* this) {
+   if (this->processList && this->processList->scanTs >= this->seenTs)
+      return (this->processList->scanTs - this->seenTs <= this->processList->settings->highlightDelaySecs);
```
BenBE, line 425: No parens needed.

Process.c
```diff
@@ -413,6 +420,16 @@ void Process_toggleTag(Process* this) {
...
+   return false;
+}
+
+bool Process_isTomb(const Process* this) {
+    return (this->tombTs > 0);
```
BenBE, line 430: dito.

Process.h
```diff
@@ -10,17 +10,18 @@ in the source distribution for its full text.
 
 #include <stdbool.h>
 #include <sys/types.h>
+#include <time.h>
```
BenBE, line 13: sorts before `<sys/types.h>` (subdirs after parents, cf. #295)

ProcessList.c
```diff
@@ -9,11 +9,11 @@ in the source distribution for its full text.
 
 #include <assert.h>
 #include <string.h>
+#include <time.h>
 
 #include "CRT.h"
 #include "XUtils.h"
 
-
```
BenBE, line 16: Two blanks after headers, cf. #295.

Settings.c
```diff
@@ -307,6 +313,7 @@ Settings* Settings_new(int initialCpuCount) {
    this->updateProcessNames = false;
    this->showProgramPath = true;
    this->highlightThreads = true;
+   this->highlightDelaySecs = DEFAULT_HIGHLIGHT_SECS;
```
BenBE, line 316: Shouldn't `highlightChanges` also be initialized (to `false`) when setting the defaults?

htop.c
```diff
@@ -46,17 +46,18 @@ static void printHelpFlag(void) {
...
-         "-t --tree                   Show the tree view by default\n"
-         "-u --user[=USERNAME]        Show only processes for a given user (or $USER)\n"
-         "-U --no-unicode             Do not use unicode but plain ASCII\n"
-         "-V --version                Print version info\n"
+         "-C --no-color                   Use a monochrome color scheme\n"
+         "-d --delay=DELAY                Set the delay between updates, in tenths of seconds\n"
+         "-F --filter=FILTER              Show only the commands matching the given filter\n"
+         "-h --help                       Print this help screen\n"
+         "-M --no-mouse                   Disable the mouse\n"
+         "-p --pid=PID,[,PID,PID...]      Show only the given PIDs\n"
```
BenBE, line 54: Minor unrelated typo:

Should be `--pid=PID[,PID,PID...]` (no comma before brackets)

htop.c
```diff
@@ -198,6 +204,23 @@ static CommandLineSettings parseArguments(int argc, char** argv) {
...
          }
+         case 'H': {
+            const char *delay = optarg;
+            if (!delay && optind < argc && argv[optind] != NULL &&
+                (argv[optind][0] != '\0' && argv[optind][0] != '-')) {
+                delay = argv[optind++];
+            }
+            if (delay) {
+                if (sscanf(delay, "%16d", &(flags.highlightDelaySecs)) == 1) {
+                   if (flags.highlightDelaySecs < 1) flags.highlightDelaySecs = 1;
```
BenBE, line 215:
```suggestion
                   if (flags.highlightDelaySecs < 1)
                      flags.highlightDelaySecs = 1;
```

BenBE (approved):
Small issues I noticed:
1. Thombstones are shown for threads even when threads are hidden
2. Minor code issues

Future PRs could also include:
1. Scrolling right on a thumbstone still shows the issue with the incomplete line markup (this seems to be a general issue with h-scrolling though).
2. Would be nice to have a key mapping (`C`???) to quickly enable/disable at runtime

I maybe found a nice way around your `firstScanTs` issue: If you manage to get the `scanTs` for the first process list scan set to 0 you should be fine (htop does one scan which is AFAIR never displayed). Would allow to remove the `firstScanTs` field.

htop.c
```diff
@@ -46,17 +46,18 @@ static void printHelpFlag(void) {
...
+         "-F --filter=FILTER              Show only the commands matching the given filter\n"
+         "-h --help                       Print this help screen\n"
+         "-M --no-mouse                   Disable the mouse\n"
+         "-p --pid=PID,[,PID,PID...]      Show only the given PIDs\n"
+         "-s --sort-key=COLUMN            Sort by COLUMN (try --sort-key=help for a list)\n"
+         "-t --tree                       Show the tree view by default\n"
+         "-u --user[=USERNAME]            Show only processes for a given user (or $USER)\n"
+         "-U --no-unicode                 Do not use unicode but plain ASCII\n"
+         "-V --version                    Print version info\n"
+         "-H --highlight-changes[=DELAY]  Highlight new and old processes\n"
```
cgzones, line 60: Please keep the help list sorted.

adsr:
Thanks for the reviews. a83f515 addresses everything except for removing `firstScanTs`. @BenBE can you explain the idea again?

BenBE (approved):
LGTM.

BenBE:
> @BenBE can you explain the idea again?

The idea is basically to use the first process scan of htop (which never gets displayed AFAIR) to set all pre-existing processes to a timestamp far in the past. The first regular scan then uses just the normal timestamps. Each process that existed in the first scan would then have a scanTs so small, that it is considered not changed.

Suggestion for a patch:
```diff
diff --git a/ProcessList.c b/ProcessList.c
index 364de1e..8b244df 100644
--- a/ProcessList.c
+++ b/ProcessList.c
@@ -29,7 +29,6 @@ ProcessList* ProcessList_init(ProcessList* this, const ObjectClass* klass, Users
    this->cpuCount = 0;
 
    this->scanTs = 0;
-   this->firstScanTs = 0;
 
 #ifdef HAVE_LIBHWLOC
    this->topologyOk = false;
@@ -87,12 +86,8 @@ void ProcessList_add(ProcessList* this, Process* p) {
    assert(Hashtable_get(this->processTable, p->pid) == NULL);
    p->processList = this;
 
-   if (this->scanTs == this->firstScanTs) {
-      // prevent highlighting processes found in first scan
-      p->seenTs = this->firstScanTs - this->settings->highlightDelaySecs - 1;
-   } else {
-      p->seenTs = this->scanTs;
-   }
+   // highlighting processes found in first scan by first scan marked "far in the past"
+   p->seenTs = this->scanTs;
 
    Vector_add(this->processes, p);
    Hashtable_put(this->processTable, p->pid, p);
@@ -318,10 +313,11 @@ void ProcessList_scan(ProcessList* this, bool pauseProcessUpdate) {
 
 
    // set scanTs
-   if (clock_gettime(CLOCK_MONOTONIC, &now) == 0) {
-      if (this->firstScanTs == 0) {
-         this->firstScanTs = now.tv_sec;
-      }
+   static bool firstScanDone = false;
+   if (!firstScanDone) {
+      this->scanTs = 0;
+      firstScanDone = true;
+   } else if (clock_gettime(CLOCK_MONOTONIC, &now) == 0) {
       this->scanTs = now.tv_sec;
    }
 
diff --git a/ProcessList.h b/ProcessList.h
index db0549a..3e35130 100644
--- a/ProcessList.h
+++ b/ProcessList.h
@@ -71,7 +71,6 @@ typedef struct ProcessList_ {
    int cpuCount;
 
    time_t scanTs;
-   time_t firstScanTs;
 } ProcessList;
 
 ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* pidMatchList, uid_t userId);
```

Issue #242, opened 2020-10-10
Set `Process.starttime_ctime` in Linux


adsr:
(Related to #241.) Somewhere in or around `LinuxProcessList_readStatFile` we should set `Process.starttime_ctime`.

PR #246, opened 2020-10-12
Do not use extra starttime process field on Linux

cgzones:
Closes: #242

### DIFF
--- a/unsupported/UnsupportedProcessList.c
+++ b/unsupported/UnsupportedProcessList.c
@@ -56,8 +56,8 @@ void ProcessList_goThroughEntries(ProcessList* super) {
     proc->priority = 0;
     proc->nice = 0;
     proc->nlwp = 1;
-    strncpy(proc->starttime_show, "Jun 01 ", sizeof(proc->starttime_show));
     proc->starttime_ctime = 1433116800; // Jun 01, 2015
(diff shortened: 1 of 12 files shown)


BenBE (approved):
LGTM.

unsupported/UnsupportedProcessList.c
```diff
@@ -56,8 +56,8 @@ void ProcessList_goThroughEntries(ProcessList* super) {
     proc->priority = 0;
     proc->nice = 0;
     proc->nlwp = 1;
-    strncpy(proc->starttime_show, "Jun 01 ", sizeof(proc->starttime_show));
     proc->starttime_ctime = 1433116800; // Jun 01, 2015
```
BenBE, line 59: What's so magic about this date? Shouldn't this be either automagically be updated or some value we can detect elsewhere?

cgzones: This is for (yet) unsupported platforms. Thus we don't know how to get the starttime of a process.
I don't know what's special about Jun 1, maybe a birthday?

BenBE: I'd prefer to have 0 here, so some easy "unknown value" check could be done.

PR #248, opened 2020-10-12
Do not hard-code line numbers in help screen building code

cgzones:
Suggested by @BenBe in #214

### DIFF
--- a/Action.c
+++ b/Action.c
@@ -28,6 +28,7 @@ in the source distribution for its full text.
 #include <pwd.h>
 #include <stdlib.h>
 #include <stdbool.h>
+#include <string.h>
 #include <sys/param.h>
 #include <sys/time.h>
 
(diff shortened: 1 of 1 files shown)


Action.c
```diff
@@ -474,37 +478,52 @@ static Htop_Reaction actionHelp(State* st) {
...
    }
-   mvaddstr( 8, 0, " Status: R: running; S: sleeping; T: traced/stopped; Z: zombie; D: disk sleep");
-   for (int i = 0; helpLeft[i].info; i++) { mvaddstr(9+i, 9,  helpLeft[i].info); }
-   for (int i = 0; helpRight[i].info; i++) { mvaddstr(9+i, 49, helpRight[i].info); }
+   line++;
+   mvaddstr(line++, 0, " Status: R: running; S: sleeping; T: traced/stopped; Z: zombie; D: disk sleep");
+
+   int lineCommandStart = line;
+   for (int i = 0; helpLeft[i].info; i++) { mvaddstr(line++, 9,  helpLeft[i].info); }
+   line = lineCommandStart;
```
BenBE, line 504: Why not combine the `.key` and `.info` output loops?

Action.c
```diff
@@ -474,37 +478,52 @@ static Htop_Reaction actionHelp(State* st) {
...
+      mvaddstr(line, 0,  helpLeft[i].key);
+      if (0 == strcmp(helpLeft[i].key, "      H: ")) {
+         attrset(CRT_colors[PROCESS_THREAD]);
+         mvaddstr(line, 32, "threads");
+         attrset(CRT_colors[HELP_BOLD]);
+      } else if (0 == strcmp(helpLeft[i].key, "      K: ")) {
+         attrset(CRT_colors[PROCESS_THREAD]);
+         mvaddstr(line, 26, "threads");
+         attrset(CRT_colors[HELP_BOLD]);
+      }
```
BenBE, line 518: Do we really need to keep this special casing?

Action.c
cgzones: done

Explorer09:
I'm not sure I like this. The reason for hard-coding line numbers was to prevent the help content from accidentally overflowing the 80×25 terminal display. Htop didn't have a "next page" or "previous page" buttons in its help screen. I think it's better to lift the limit only after we implement paging in the help screen UI.

BenBE:
We already are hitting 80x25 in several locations (meters setup is one of them) and for the configuration screen we are at least close depending on headers activated. So it's not really a new problem, but rather something we would need to resolve in general.

PR #249, opened 2020-10-12
Misc XAlloc cleanup

cgzones:


XAlloc.c
```diff
@@ -1,41 +1,48 @@
-
 #include "XAlloc.h"
-#include "RichString.h"
 
+#include "CRT.h"
+
+/* The function vasprintf is a GNU extension. */
 #ifndef _GNU_SOURCE
 #define _GNU_SOURCE
 #endif
```
BenBE, line 8: Isn't this set by `configure`/the `Makefile`?

cgzones: Indeed; replaced with a`#include "config.h"`

cgzones:
Scheduled after #160 and #161

BenBE (approved):
LGTM.

PR #250, opened 2020-10-12
Add SELinuxMeter

cgzones:
Show state and if enabled the current mode

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -119,23 +119,31 @@ myhtopheaders = \
 # -----
 
 linux_platform_headers = \
-	linux/Platform.h \
-	linux/IOPriorityPanel.h \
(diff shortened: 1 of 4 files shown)


Makefile.am
```diff
@@ -132,6 +132,8 @@ linux_platform_headers = \
 
 if HTOP_LINUX
 AM_LDFLAGS += -rdynamic
+myhtopsources += SELinuxMeter.c
+myhtopheaders += SELinuxMeter.h
```
natoscott, line 136: Should these new sources reside below the linux/ directory?  I guess SE*Linux* isn't going to exist on other platforms ;)
Otherwise LGTM.

cgzones: moved into the linux/ directory.

PR #259, opened 2020-10-15
Misc Vector updates

cgzones:
- Move swap() macro to source file and implement as function
- Implement Vector_get() and Vector_size() as inline functions
  to make them type safe and avoid lhs usage
- Comment comparison statistics, they are only needed for performance
  testing
- Use malloc, the code does not assume zero initialization and we might
  later use realloc() on the data

### DIFF
--- a/Vector.c
+++ b/Vector.c
@@ -18,6 +18,7 @@ Vector* Vector_new(const ObjectClass* type, bool owner, int size) {
 
    if (size == DEFAULT_SIZE)
       size = 10;
+   assert(size > 0);
    this = xMalloc(sizeof(Vector));
    this->growthRate = size;
    this->array = (Object**) xCalloc(size, sizeof(Object*));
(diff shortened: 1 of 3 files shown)


Vector.c
```diff
@@ -18,9 +18,10 @@ Vector* Vector_new(const ObjectClass* type, bool owner, int size) {
 
    if (size == DEFAULT_SIZE)
       size = 10;
+   assert(size > 0);
    this = xMalloc(sizeof(Vector));
    this->growthRate = size;
-   this->array = (Object**) xCalloc(size, sizeof(Object*));
+   this->array = xMalloc(size * sizeof(Object*));
```
BenBE, line 24: Please overflow-check this multiplication.

Vector.c
```diff
@@ -87,14 +88,22 @@ void Vector_prune(Vector* this) {
    this->items = 0;
 }
 
-static int comparisons = 0;
+//static int comparisons = 0;
+
+static void swap(Object** array, int indexA, int indexB) {
```
BenBE, line 93: Could as well be marked `inline` …

BenBE (changes requested):

Vector.c
cgzones: dropped this change; it seems not worth it after all

Vector.c
cgzones: Optimizers should do the *right* thing on static/file-local functions

PR #262, opened 2020-10-17
Fix various file descriptor leaks

BenBE:


TraceScreen.c
```diff
@@ -44,11 +44,9 @@ const InfoScreenClass TraceScreen_class = {
 };
 
 TraceScreen* TraceScreen_new(Process* process) {
-   TraceScreen* this = xMalloc(sizeof(TraceScreen));
+   TraceScreen* this = xCalloc(1, sizeof(TraceScreen));
```
fasterit, line 47:
```suggestion
   // This initializes all TraceScreen variables to "false" so only default = true ones need to be set below
   TraceScreen* this = xCalloc(1, sizeof(TraceScreen));
```

TraceScreen.c
```diff
@@ -74,43 +74,71 @@ void TraceScreen_draw(InfoScreen* this) {
...
-   int ok = fcntl(this->fdpair[0], F_SETFL, O_NONBLOCK);
-   if (ok == -1) return false;
-   this->strace = fdopen(this->fdpair[0], "r");
-   this->fd_strace = fileno(this->strace);
+
+   close(fdpair[1]);
+
+   FILE* fd = fdopen(fdpair[0], "r");
+   if (!fd)
+      goto err;
```
cgzones, line 117: on this err label, we close `fdpair[1]` twice

TraceScreen.c
```diff
@@ -74,43 +74,71 @@ void TraceScreen_draw(InfoScreen* this) {
...
+err:
+   close(fdpair[1]);
+   close(fdpair[0]);
+   return false;
 }
 
 void TraceScreen_updateTrace(InfoScreen* super) {
    TraceScreen* this = (TraceScreen*) super;
    char buffer[1025];
+   int fd_strace = fileno(this->strace);
```
cgzones, line 132: maybe add an `assert(fd_strace != -1);`

TraceScreen.h
```diff
@@ -12,10 +12,8 @@ in the source distribution for its full text.
 typedef struct TraceScreen_ {
    InfoScreen super;
    bool tracing;
-   int fdpair[2];
    int child;
```
cgzones, line 15: Maybe change the type of child to `pid_t`?

TraceScreen.c
BenBE: Wasn't there before, but sure.

cgzones (approved):

PR #263, opened 2020-10-17
Fix misaligned access inside taskstats structure

BenBE:
Reported by UB sanitizer (alongside several other messages):
```
linux/LinuxProcessList.c:782:25: runtime error: member access within misaligned address 0x614000000264 for type 'struct taskstats', which requires 8 byte alignment
0x614000000264: note: pointer points here
  64 01 03 00 0a 00 00 00  00 00 00 00 02 00 00 00  00 00 00 00 4b c8 2e 00  00 00 00 00 3e 45 3c fd
              ^
```

The issue doesn't cause trouble on x86, but any architecture with stricter memory alignment requirements may inadvertedly break.

### DIFF
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -766,7 +766,7 @@ static int handleNetlinkMsg(struct nl_msg *nlmsg, void *linuxProcess) {
    struct nlmsghdr *nlhdr;
    struct nlattr *nlattrs[TASKSTATS_TYPE_MAX + 1];
    struct nlattr *nlattr;
-   struct taskstats *stats;
+   struct taskstats stats;
    int rem;
    unsigned long long int timeDelta;
    LinuxProcess* lp = (LinuxProcess*) linuxProcess;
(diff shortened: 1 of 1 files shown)


linux/LinuxProcessList.c
```diff
@@ -766,7 +766,7 @@ static int handleNetlinkMsg(struct nl_msg *nlmsg, void *linuxProcess) {
    struct nlmsghdr *nlhdr;
    struct nlattr *nlattrs[TASKSTATS_TYPE_MAX + 1];
    struct nlattr *nlattr;
-   struct taskstats *stats;
```
cgzones, line 769: Please drop the pointer variable `stats` and replace its usages with `&stats_data`

cgzones (commented):
LGTM

p.s.: how did you encounter these alignment warnings?
I am using clang-11 with UBSAN on Debian sid (amd64) and do not get any alignment issues.

cgzones (approved):

BenBE:
> p.s.: how did you encounter these alignment warnings?

I encountered them using ./configure CFLAGS+="-fsanitize=address,undefined" while using gcc (9.3.0).

I noticed them when trying to start htop with `sudo` (short flicker before the UI showed). When redirecting the normal output to /dev/null I got a clear list of issues with the above mentioned report showing. My guess is that due to missing privileges a normal user doesn't get to that code path.

linux/LinuxProcessList.c
BenBE: It was easier to just make `stats` a stack variable and change `->` to `.`

cgzones:
Ah, as root I can confirm this alignment warnings.

PR #265, opened 2020-10-19
IWYU update

cgzones:
- Add Settings forward declaration in Process.h
- Add conditional stdbool.h include in Object.h
- Drop unneeded stddef.h include in Richstring.c
- improve script to run custom iwyu version


BenBE (approved):
LGTM.

P.S.: Minor note: For DragonflyBSD the headers are not properly sorted. Might have missed that in my original PR. Please also check for Solaris.

PR #271, opened 2020-10-20
Increase print buffer in NetworkIOMeter_display

cgzones:
In case the packet values wrap-around or have other weird values, the
current buffer might be to small


BenBE (approved):
LGTM.

PR #272, opened 2020-10-20
Improve handling of no data in Disk and Network IO Meters

cgzones:


linux/Platform.c
```diff
@@ -347,32 +344,28 @@ void Platform_getDiskIO(unsigned long int *bytesRead, unsigned long int *bytesWr
...
    while (fgets(lineBuffer, sizeof(lineBuffer), fd)) {
       char interfaceName[32];
       unsigned long int bytesReceivedParsed, packetsReceivedParsed, bytesTransmittedParsed, packetsTransmittedParsed;
-      if (fscanf(fd, "%31s %lu %lu %*u %*u %*u %*u %*u %*u %lu %lu %*u %*u %*u %*u %*u %*u",
-                     interfaceName,
-                     &bytesReceivedParsed,
-                     &packetsReceivedParsed,
-                     &bytesTransmittedParsed,
-                     &packetsTransmittedParsed) != 5)
+      if (sscanf(lineBuffer, "%31s %lu %lu %*u %*u %*u %*u %*u %*u %lu %lu %*u %*u %*u %*u %*u %*u",
```
natoscott, line 363: Good catch with the fscanf vs sscanf there.  I think you can safely drop the trailing "%*u %*u %*u %*u %*u %*u" in the above string - that's going to scan those unsigned integers for nothing and throw the result away, so it should be OK to stop scanning at the final %lu there.

cgzones: yep; dropped trailing ignored format specifiers

PR #273, opened 2020-10-20
Fix wrong strncmp replacement

cgzones:
Fix wrong strncmp replacement


BenBE (approved):
LGTM.

But the PR title could be more descriptive …

PR #274, opened 2020-10-21
Misc Cleanup

cgzones:
Mark several structure members const.
E.g. settings should only be changed by the Settings Screen and not through a Process pointer.

(Dropped the hashtable changes)

### DIFF
--- a/ProcessList.c
+++ b/ProcessList.c
@@ -67,7 +67,7 @@ void ProcessList_setPanel(ProcessList* this, Panel* panel) {
 
 void ProcessList_printHeader(ProcessList* this, RichString* header) {
    RichString_prune(header);
-   ProcessField* fields = this->settings->fields;
+   const ProcessField* fields = this->settings->fields;
    for (int i = 0; fields[i]; i++) {
       const char* field = Process_fields[fields[i]].title;
       if (!field) field = "- ";
@@ -142,20 +142,21 @@ static void ProcessList_buildTree(ProcessList* this, pid_t pid, int level, int i
    Vector_delete(children);
 }
 
+static long ProcessList_treeProcessCompare(const void* v1, const void* v2) {
+   const Process *p1 = (const Process*)v1;
+   const Process *p2 = (const Process*)v2;
+
+   return p1->pid - p2->pid;
+}
+
 void ProcessList_sort(ProcessList* this) {
    if (!this->settings->treeView) {
       Vector_insertionSort(this->processes);
    } else {
       // Save settings
       int direction = this->settings->direction;
-      int sortKey = this->settings->sortKey;
       // Sort by PID
-      this->settings->sortKey = PID;
-      this->settings->direction = 1;
-      Vector_quickSort(this->processes);
-      // Restore settings
-      this->settings->sortKey = sortKey;
-      this->settings->direction = direction;
+      Vector_quickSortCustomCompare(this->processes, ProcessList_treeProcessCompare);
       int vsize = Vector_size(this->processes);
       // Find all processes whose parent is not visible
       int size;
@@ -214,7 +215,7 @@ void ProcessList_sort(ProcessList* this) {
 
 ProcessField ProcessList_keyAt(ProcessList* this, int at) {
    int x = 0;
-   ProcessField* fields = this->settings->fields;
+   const ProcessField* fields = this->settings->fields;
    ProcessField field;
    for (int i = 0; (field = fields[i]); i++) {
       const char* title = Process_fields[field].title;
--- a/Vector.c
+++ b/Vector.c
@@ -161,10 +161,10 @@ static void insertionSort(Object** array, int left, int right, Object_Compare co
    }
 }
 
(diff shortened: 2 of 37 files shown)


ProcessList.c
```diff
@@ -87,7 +87,7 @@ void ProcessList_add(ProcessList* this, Process* p) {
 
    assert(Vector_indexOf(this->processes, p, Process_pidCompare) != -1);
    assert(Hashtable_get(this->processTable, p->pid) != NULL);
-   assert(Hashtable_count(this->processTable) == Vector_count(this->processes));
+   assert(Hashtable_count(this->processTable) == (unsigned int)Vector_count(this->processes));
```
BenBE, line 90: Shouldn't be the types for amounts in `*_count` functions be compatible?

ProcessList.c
```diff
@@ -100,7 +100,7 @@ void ProcessList_remove(ProcessList* this, Process* p) {
    assert(idx != -1);
    if (idx >= 0) Vector_remove(this->processes, idx);
    assert(Hashtable_get(this->processTable, pid) == NULL); (void)pid;
-   assert(Hashtable_count(this->processTable) == Vector_count(this->processes));
+   assert(Hashtable_count(this->processTable) == (unsigned int)Vector_count(this->processes));
```
BenBE, line 103: Dito.

ProcessList.c
```diff
@@ -142,20 +142,21 @@ static void ProcessList_buildTree(ProcessList* this, pid_t pid, int level, int i
...
+static long ProcessList_treeProcessCompare(const void* v1, const void* v2) {
+   const Process *p1 = (const Process*)v1;
+   const Process *p2 = (const Process*)v2;
+
+   return p1->pid - p2->pid;
+}
+
 void ProcessList_sort(ProcessList* this) {
    if (!this->settings->treeView) {
       Vector_insertionSort(this->processes);
```
BenBE, line 154: Couldn't `Vector_quickSortCustomCompare` be also used for normal sorting (given the right comparison function)?

darwin/DarwinProcess.c
```diff
@@ -37,7 +37,7 @@ DarwinProcess* DarwinProcess_new(Settings* settings) {
    this->stime = 0;
    this->taskAccess = true;
 
-   return this;
+   return &this->super;
```
BenBE, line 40: Use some central `AsProcess` magic that enforces `&(*this)` and `&this->super` to be the same pointers (mod type).

BenBE (commented):
Do we have some place for central (static) checking of type constraints? E.g. with this change we really require the .super field to be at offset 0 in the struct. While this is true for all current platforms I'd feel better if this was checked somewhere (statically).

BenBE:
Any reason to copy `settings` over to each process instead of having each process just know which process list it belongs to and reference from there?

ProcessList.c
cgzones: `Vector_quickSort` uses the *right* one?

darwin/DarwinProcess.c
cgzones: All the Object pseudo inheritance bases on the fact that a struct address equals the address of the first field.

cgzones:
> Any reason to copy settings over to each process instead of having each process just know which process list it belongs to and reference from there?

What's the difference of storing the ProcessList instead of Settings?

BenBE:
> > Any reason to copy settings over to each process instead of having each process just know which process list it belongs to and reference from there?
> 
> What's the difference of storing the ProcessList instead of Settings?

With ProcessList* you can store more things in a central place.
The ProcessList is added to Process for #240 too.

ProcessList.c
BenBE: Just wondered if we have code duplication there.

Vector.c
```diff
@@ -54,13 +54,13 @@ static bool Vector_isConsistent(const Vector* this) {
...
-int Vector_count(const Vector* this) {
-   int items = 0;
+unsigned int Vector_count(const Vector* this) {
+   unsigned int items = 0;
    for (int i = 0; i < this->items; i++) {
       if (this->array[i])
          items++;
    }
-   assert(items == this->items);
+   assert(items == (unsigned int)this->items);
```
BenBE, line 63: Chance to avoid this cast too?

cgzones: That will cascade into many `int` -> `unsigned int` changes throughout the code; maybe at some point

darwin/DarwinProcess.c
cgzones: It's guaranteed by the C Standard: 

6.7.2.1-13 states:
> Apointer to astructure  object,  suitably  converted,  points  to  its  initial  member  (or  if  that  member  is  abit-field,  then  to  the  unit  in  which  it  resides),  and  vice  versa.   Theremay  be  unnamedpadding within a structure object, but not at its beginning.

http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf page 115

PR #278, opened 2020-10-22
Parse POWER_SUPPLY_CAPACITY

jpalus:
If POWER_SUPPLY_{CHARGE,ENERGY}_NOW is missing then try to use
POWER_SUPPLY_CAPACITY to determine current charge level.

### DIFF
--- a/linux/Battery.c
+++ b/linux/Battery.c
@@ -230,12 +230,18 @@ static void Battery_getSysData(double* level, ACPresence* isOnAC) {
          char *line = NULL;
          bool full = false;
          bool now = false;
+         int fullSize = 0;
+         double capacityLevel = NAN;
          while ((line = strsep(&buf, "\n")) != NULL) {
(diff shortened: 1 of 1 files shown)


linux/Battery.c
```diff
@@ -192,6 +192,8 @@ static void Battery_getSysData(double* level, ACPresence* isOnAC) {
 
    unsigned long int totalFull = 0;
    unsigned long int totalRemain = 0;
+   bool now = false;
```
cgzones, line 195: This scope change does change the logic, as now `now` does not get reset to `false` for the second and following batteries.

jpalus: Ouch you're right, I got confused by all the scopes around. Updated however now it's limited to case when full is available while now is not (which I'm affected by).

cgzones (approved):
LGTM

cgzones:
Thanks, applied.

PR #283, opened 2020-10-26
Refactor code for reading process environment from procfs

BenBE:
Some cleanup and streamlining of the implementation.

This uses the fact, that `realloc(NULL, size)` is equal to simply calling `malloc(size)`.

### DIFF
--- a/linux/Platform.c
+++ b/linux/Platform.c
@@ -263,24 +263,35 @@ char* Platform_getProcessEnv(pid_t pid) {
    char procname[128];
    xSnprintf(procname, sizeof(procname), PROCDIR "/%d/environ", pid);
    FILE* fd = fopen(procname, "r");
+   if(!fd)
+      return NULL;
+
    char *env = NULL;
-   if (fd) {
-      size_t capacity = 4096, size = 0, bytes;
-      env = xMalloc(capacity);
-      while ((bytes = fread(env+size, 1, capacity-size, fd)) > 0) {
-         size += bytes;
-         capacity *= 2;
-         env = xRealloc(env, capacity);
-      }
-      fclose(fd);
-      if (size < 2 || env[size-1] || env[size-2]) {
-         if (size + 2 < capacity) {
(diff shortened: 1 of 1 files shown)


linux/Platform.c
```diff
@@ -263,24 +263,34 @@ char* Platform_getProcessEnv(pid_t pid) {
...
+   do {
+      size += bytes;
+      capacity += 4096;
+      env = xRealloc(env, capacity);
+   } while ((bytes = fread(env + size, 1, capacity - size, fd)) > 0);
+
+   fclose(fd);
+
+   if (bytes < 0) {
+      free(env);
```
cgzones, line 284: should there be a `return NULL;` in this block?

linux/Platform.c
```diff
@@ -263,24 +263,34 @@ char* Platform_getProcessEnv(pid_t pid) {
...
    }
+
+   size += bytes;
+
+   env = xRealloc(env, size + 2);
+
+   env[size] = 0;
```
cgzones, line 291: Maybe use '\0' to make it more obvious that we set a NUL-byte (and not some integer value or NULL-pointer).

linux/Platform.c
```diff
@@ -263,24 +263,34 @@ char* Platform_getProcessEnv(pid_t pid) {
...
+
+   fclose(fd);
+
+   if (bytes < 0) {
+      free(env);
    }
+
+   size += bytes;
+
+   env = xRealloc(env, size + 2);
```
cgzones, line 289: Does it make sense to allocate `capacity + 2` or read `capacity - size -2` to avoid this realloc step?

BenBE: Not really, because unless you read `>=capacity-2` bytes from the environment you can always shrink the memory here, which is mostly free in terms of CPU time. And if reading up until `capacity-2` it makes the code somewhat less intuitive IMO.

cgzones (approved):

PR #286, opened 2020-10-27
Refactor LinuxProcessList_readSmapsFile to work line-oriented

BenBE:
Some cleanup to make the code readable and comply to the rough coding style.

### DIFF
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -479,54 +479,43 @@ static bool LinuxProcessList_readSmapsFile(LinuxProcess* process, const char* di
    //http://elixir.free-electrons.com/linux/v4.10/source/fs/proc/task_mmu.c#L719
    //kernel will return data in chunks of size PAGE_SIZE or less.
 
-   char buffer[CRT_pageSize];// 4k
-   ssize_t nread=0;
+   char buffer[256];
+
    if(haveSmapsRollup) {// only available in Linux 4.14+
       xSnprintf(buffer, sizeof(buffer), "%s/%s/smaps_rollup", dirname, name);
    } else {
       xSnprintf(buffer, sizeof(buffer), "%s/%s/smaps", dirname, name);
    }
-   int fd = open(buffer, O_RDONLY);
-   if (fd == -1)
+
+   FILE* f = fopen(buffer, "r");
+   if (!f)
       return false;
 
    process->m_pss   = 0;
    process->m_swap  = 0;
    process->m_psswp = 0;
 
-   while ( ( nread =  read(fd,buffer, sizeof(buffer)) ) > 0 ){
-        char* start = buffer;
-        char* end   = start + nread;
-        do{//parse 4k block
-            int tmp;
+   while (fgets(buffer, sizeof(buffer), f)) {
+      if(!strchr(buffer, '\n')) {
+         // Partial line, skip to end of this line
+         while(fgets(buffer, sizeof(buffer), f)) {
+            if(strchr(buffer, '\n')) {
+               break;
+            }
+         }
+         continue;
+      }
 
-            if( (tmp = (end - start)) > 0 &&
-                (start = memmem(start,tmp,"\nPss:",5)) != NULL )
-            {
-              process->m_pss += strtol(start+5, &start, 10);
-              start += 3;//now we must be at the end of line "Pss:                   0 kB"
-            }else
-              break; //read next 4k block
+      if (String_startsWith(buffer, "Pss:")) {
+         process->m_pss += strtol(buffer + 4, NULL, 10);
+      } else if (String_startsWith(buffer, "Swap:")) {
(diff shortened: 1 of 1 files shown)


linux/LinuxProcessList.c
```diff
@@ -479,54 +479,33 @@ static bool LinuxProcessList_readSmapsFile(LinuxProcess* process, const char* di
    //http://elixir.free-electrons.com/linux/v4.10/source/fs/proc/task_mmu.c#L719
    //kernel will return data in chunks of size PAGE_SIZE or less.
 
-   char buffer[CRT_pageSize];// 4k
-   ssize_t nread=0;
+   char buffer[128];
```
cgzones, line 482: 128 might not be enough
E.g. `cat /proc/$$/smaps` contains (length 120)
```
7fea52867000-7fea5286a000 r--p 00000000 fe:01 929866                     /usr/lib/x86_64-linux-gnu/libnss_files-2.31.so
```

And too long lines might break the `fgets()` logic

BenBE: Doubled buffer size and copied line skip logic from my OpenVZ status file parser.
Long files would only have broken things if you had one of the keywords exactly after a buffer boundary.

linux/LinuxProcessList.c
```diff
@@ -479,54 +479,43 @@ static bool LinuxProcessList_readSmapsFile(LinuxProcess* process, const char* di
...
-              process->m_psswp += strtol(start+9, &start, 10);
-              start += 3;
-            }else
-              break;
-
-        }while(1);
-   }//while read
-   close(fd);
+   while (!fgets(buffer, sizeof(buffer), f)) {
+      if(!strchr(buffer, '\n')) {
```
cgzones, line 499: It somehow displeases me that we now need to iterate over each line to its end.
Can we refactor (maybe add a wrapper function) by using old style descriptors (**open**(2) and **read**(2))

BenBE: Gave it a try, but the result just feels horrible …

```c
static bool LinuxProcessList_readSmapsFile(LinuxProcess* process, const char* dirname, const char* name, bool haveSmapsRollup) {
   //http://elixir.free-electrons.com/linux/v4.10/source/fs/proc/task_mmu.c#L719
   //kernel will return data in chunks of size PAGE_SIZE or less.

   char buffer[4096];
   size_t buffill = 0;

   if(haveSmapsRollup) {// only available in Linux 4.14+
      xSnprintf(buffer, sizeof(buffer), "%s/%s/smaps_rollup", dirname, name);
   } else {
      xSnprintf(buffer, sizeof(buffer), "%s/%s/smaps", dirname, name);
   }

   int fd = open(buffer, O_RDONLY);
   if (fd < 0)
      return false;

   process->m_pss   = 0;
   process->m_swap  = 0;
   process->m_psswp = 0;

   ssize_t nread;
   do {
      nread = read(fd, buffer + buffill, sizeof(buffer) - buffill - 1);

      // Check for EOF or error
      if(nread < 0)
         break;

      buffill += (size_t)nread;
      buffer[buffill] = '\0';

      char *start = buffer;
      do {
         char *eol = strpbrk(start, "\n");

         if (!eol) {
            if (start != buffer)
               break;

            do {
               nread = read(fd, buffer, sizeof(buffer) - 1);
               if (nread >= 0)
                  buffer[nread] = '\0';
            } while(nread > 0 && NULL == ((eol = strpbrk(buffer, "\n"))));

            if (nread <= 0) {
               nread = 0;
               buffill = 0;
               break;
            }

            buffill = (size_t)nread;
            start = eol + 1;
            continue;
         }

         *eol = '\0';

         if (String_startsWith(start, "Pss:")) {
            process->m_pss += strtol(start + 4, NULL, 10);
         } else if (String_startsWith(start, "Swap:")) {
            process->m_swap += strtol(start + 5, NULL, 10);
         } else if (String_startsWith(start, "SwapPss:")) {
            process->m_psswp += strtol(start + 8, NULL, 10);
         }

         start = eol + 1;
      } while (eol);

      buffill -= start - buffer;
      memmove(start, buffer, buffill);
   } while (nread > 0);

   close(fd);
   return true;
}
```

cgzones: ok, nvm

linux/LinuxProcessList.c
```diff
@@ -479,54 +479,43 @@ static bool LinuxProcessList_readSmapsFile(LinuxProcess* process, const char* di
...
-            {
-              process->m_psswp += strtol(start+9, &start, 10);
-              start += 3;
-            }else
-              break;
-
-        }while(1);
-   }//while read
-   close(fd);
+   while (!fgets(buffer, sizeof(buffer), f)) {
```
cgzones, line 498: why `**!**fgets(...` ?

BenBE: It became too late yesterday …

cgzones:
Applied, thanks.

PR #289, opened 2020-10-29
Add HTOP_$platform defines to config.h header

cgzones:
Can be used to conditionally compile platform specific code.


BenBE (approved):
LGTM.

PR #290, opened 2020-10-29
Use integer type for item count instead of char

cgzones:


Meter.c
```diff
@@ -195,8 +195,8 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
 
    // First draw in the bar[] buffer...
    int offset = 0;
-   int items = this->curItems;
-   for (int i = 0; i < items; i++) {
+   uint8_t items = this->curItems;
```
BenBE, line 198: No need to have this put as an extra variable. Only re-used in one further loop, should use `this->curItems` too.

Meter.c
```diff
@@ -296,8 +296,8 @@ static void GraphMeterMode_draw(Meter* this, int x, int y, int w) {
       Meter_updateValues(this, buffer, nValues - 1);
 
       double value = 0.0;
-      int items = this->curItems;
-      for (int i = 0; i < items; i++)
+      uint8_t items = this->curItems;
+      for (uint8_t i = 0; i < items; i++)
```
BenBE, line 300: dito.

CPUMeter.c
```diff
@@ -51,8 +51,8 @@ static void CPUMeter_updateValues(Meter* this, char* buffer, int size) {
    int cpu = this->param;
    if (cpu > this->pl->cpuCount) {
       xSnprintf(buffer, size, "absent");
-      int items = this->curItems;
-      for (int i = 0; i < items; i++)
+      uint8_t items = this->curItems;
+      for (uint8_t i = 0; i < items; i++)
```
BenBE, line 55: dito.

BenBE (changes requested):

Meter.c
cgzones: don't we save (theoretically) re-fetching `this->curItems` on every loop condition test this way?

BenBE: As this value can be shown by the compiler to be constant it will optimized to be read once anyway (if using build optimizations).

In contrast, having the extra variable adds additional indirection for the reader of the source code, which doesn't add any benefit over what a good compiler would do anyway.

BenBE (approved):

PR #293, opened 2020-10-30
Fix column sorting

eersoy93:
Fix segmentation fault when column name is NULL when sorting argument is invalid column (ex: ./htop -s --help).

So, some columns (ex: SECATTR) can be sortable now.

NOTE: This is my first PR for htop-dev/htop.

### DIFF
--- a/ColumnsPanel.c
+++ b/ColumnsPanel.c
@@ -135,15 +135,6 @@ ColumnsPanel* ColumnsPanel_new(Settings* settings) {
    return this;
 }
 
-int ColumnsPanel_fieldNameToIndex(const char* name) {
-   for (int j = 1; j <= Platform_numberOfFields; j++) {
-      if (String_eq(name, Process_fields[j].name)) {
-         return j;
-      }
-   }
-   return -1;
-}
-
 void ColumnsPanel_update(Panel* super) {
    ColumnsPanel* this = (ColumnsPanel*) super;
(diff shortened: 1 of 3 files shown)


cgzones:
Thanks for reporting, this is indeed a bug.

Maybe just edit the function `ColumnsPanel_fieldNameToIndex`, which only purpose is this search (it's only used here).

```c
diff --git a/ColumnsPanel.c b/ColumnsPanel.c
index f825019..d616495 100644
--- a/ColumnsPanel.c
+++ b/ColumnsPanel.c
@@ -137,7 +137,7 @@ ColumnsPanel* ColumnsPanel_new(Settings* settings) {
 
 int ColumnsPanel_fieldNameToIndex(const char* name) {
    for (int j = 1; j <= Platform_numberOfFields; j++) {
-      if (String_eq(name, Process_fields[j].name)) {
+      if (Process_fields[j].name && String_eq(name, Process_fields[j].name)) {
          return j;
       }
    }
```

eersoy93:
The ColumnsPanel_fieldNameToIndex() function doesn't iterate all columns for Linux processes. I have tried fixing this bug on Linux (Debian 10).

cgzones:
Thanks, applied via https://github.com/htop-dev/htop/commit/59ef15b2ad6037f40d7fe4207b2b59dd11b14b8b

(with some minor style adjustments)

PR #297, opened 2020-10-31
Assert Vector_get returns an object

cgzones:
It is generally assumed Vector_get returns a non-NULL object.
Use a generic assert in Vector_get instead of in callers.

Spotted by @BenBE

### DIFF
--- a/Panel.c
+++ b/Panel.c
@@ -261,7 +261,6 @@ void Panel_draw(Panel* this, bool focus) {
       int line = 0;
       for(int i = first; line < h && i < upTo; i++) {
          Object* itemObj = Vector_get(this->items, i);
-         assert(itemObj); if(!itemObj) continue;
          RichString_begin(item);
          Object_display(itemObj, &item);
          int itemLen = RichString_sizeVal(item);
@@ -288,7 +287,6 @@ void Panel_draw(Panel* this, bool focus) {
 
    } else {
       Object* oldObj = Vector_get(this->items, this->oldSelected);
-      assert(oldObj);
       RichString_begin(old);
       Object_display(oldObj, &old);
       int oldLen = RichString_sizeVal(old);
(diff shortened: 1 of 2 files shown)


BenBE:
Small objection or at least one point that needs a bit of discussion:

I think the NULL checks after the call are generally okay. The reason that I mentioned the initial oddity (assert+ NULL check) was that I wasn't sure, what the error behaviour of the caller should be in case of a NULL return, as `assert(obj);` (error out) and `if(!obj) continue;` (ignore) somewhat contradict themselves.

Also: Having the assert in `Vector_get` is a hidden API change in that previously you could legitimately store NULL pointers in a vector (just had to deal with it yourself). With this change NULL is forbidden. As far as my overview of the source goes this shouldn't affect any existing code, but nonetheless it's something to be aware of.

cgzones:
> Also: Having the assert in Vector_get is a hidden API change in that previously you could legitimately store NULL pointers in a vector (just had to deal with it yourself). With this change NULL is forbidden. As far as my overview of the source goes this shouldn't affect any existing code, but nonetheless it's something to be aware of.

No, you can not (in debug mode) put a NULL-pointer into a Vector:

https://github.com/htop-dev/htop/blob/15652e7b8102e86b3405254405d8ee5d2a239004/Vector.c#L191-L194

https://github.com/htop-dev/htop/blob/15652e7b8102e86b3405254405d8ee5d2a239004/Vector.c#L254-L257

https://github.com/htop-dev/htop/blob/15652e7b8102e86b3405254405d8ee5d2a239004/Vector.c#L289-L291

resulting in 

https://github.com/htop-dev/htop/blob/15652e7b8102e86b3405254405d8ee5d2a239004/Object.c#L20-L22

BenBE:
Ah, k. If that check already was there, I'm fine with this change.

PR #298, opened 2020-11-01
Macro cleanup

BenBE:
This PR consists of two parts:
1. The first 4 commits fix some macro issues I noticed
2. The next 4 patches fix spacing around keywords
3. The remainder fixes code formatting to be more consistent

When reviewing I recommend ignoring whitespace changes.


Action.c
```diff
@@ -59,14 +59,15 @@ Object* Action_pickFromVector(State* st, Panel* list, int x, bool followProcess)
...
-   Panel_resize(panel, COLS, LINES-y-1);
+   Panel_resize(panel, COLS, LINES - y - 1);
    if (panelFocus == list && ch == 13) {
       if (followProcess) {
          Process* selected = (Process*)Panel_getSelected(panel);
-         if (selected && selected->pid == pid)
+         if (selected && selected->pid == pid) {
             return Panel_getSelected(list);
-         else
+         } else {
```
cgzones, line 68: If we touch this code, maybe remove the else branch? (else-after-return)

RichString.h
```diff
@@ -15,21 +15,21 @@ in the source distribution for its full text.
 #define RichString_size(this) ((this)->chlen)
 #define RichString_sizeVal(this) ((this).chlen)
 
-#define RichString_begin(this) RichString (this); RichString_beginAllocated(this);
-#define RichString_beginAllocated(this) memset(&(this), 0, sizeof(RichString)); (this).chptr = (this).chstr;
-#define RichString_end(this) RichString_prune(&(this));
+#define RichString_begin(this) RichString (this); RichString_beginAllocated(this)
+#define RichString_beginAllocated(this) memset(&(this), 0, sizeof(RichString)); (this).chptr = (this).chstr
```
cgzones, line 19: `RichString_beginAllocated` now becomes *unsafe* cause one can append it, e.g. `RichString_beginAllocated(out) + 2;`. (I rather have an unnecessary semicolon than an unsafe macro.)
Maybe convert into an `static inline` function?

TraceScreen.c
```diff
@@ -141,11 +144,14 @@ void TraceScreen_updateTrace(InfoScreen* super) {
    FD_SET(fd_strace, &fds);
 
    struct timeval tv;
-   tv.tv_sec = 0; tv.tv_usec = 500;
-   int ready = select(fd_strace+1, &fds, NULL, NULL, &tv);
+   tv.tv_sec = 0;
```
cgzones, line 147: initialize with `struct timeval tv = { .tv_sec = 0, .tv_usec = 500 };`

dragonflybsd/Platform.c
```diff
@@ -169,15 +169,17 @@ double Platform_setCPUValues(Meter* this, int cpu) {
...
    } else {
       v[2] = cpuData->systemAllPercent;
       this->curItems = 3;
-      percent = v[0]+v[1]+v[2];
+      percent = v[0] + v[1] + v[2];
    }
 
    percent = CLAMP(percent, 0.0, 100.0);
-   if (isnan(percent)) percent = 0.0;
+   if (isnan(percent)) {
```
cgzones, line 180: strictly speaking this is dead code as `CLAMP(NAN, $min, $max) == $min`

linux/LinuxProcessList.c
```diff
@@ -777,24 +808,25 @@ static int handleNetlinkMsg(struct nl_msg *nlmsg, void *linuxProcess) {
       memcpy(&stats, nla_data(nla_next(nla_data(nlattr), &rem)), sizeof(stats));
       assert(lp->super.pid == (pid_t)stats.ac_pid);
 
-      timeDelta = (stats.ac_etime*1000 - lp->delay_read_time);
-      #define BOUNDS(x) isnan(x) ? 0.0 : ((x) > 100) ? 100.0 : (x);
-      #define DELTAPERC(x,y) BOUNDS((float) ((x) - (y)) / timeDelta * 100);
+      timeDelta = (stats.ac_etime * 1000 - lp->delay_read_time);
```
cgzones, line 811: remove braces

linux/LinuxProcessList.c
```diff
@@ -976,41 +1021,47 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, const char*
...
-      if (settings->flags & PROCESS_FLAG_LINUX_IOPRIO)
+
+      if (settings->flags & PROCESS_FLAG_LINUX_IOPRIO) {
          LinuxProcess_updateIOPriority(lp);
+      }
+
       float percent_cpu = (lp->utime + lp->stime - lasttimes) / period * 100.0;
       proc->percent_cpu = CLAMP(percent_cpu, 0.0, cpus * 100.0);
-      if (isnan(proc->percent_cpu)) proc->percent_cpu = 0.0;
+      if (isnan(proc->percent_cpu))
```
cgzones, line 1059: see above

linux/LinuxProcessList.c
```diff
@@ -1111,7 +1169,11 @@ static inline void LinuxProcessList_scanMemoryInfo(ProcessList* this) {
    char buffer[128];
    while (fgets(buffer, 128, file)) {
 
-      #define tryRead(label, variable) do { if (String_startsWith(buffer, label) && sscanf(buffer + strlen(label), " %32llu kB", variable)) { break; } } while(0)
+      #define tryRead(label, variable) \
```
cgzones, line 1172: We should probably exit the case (*break*), if just the string was found (not necessarily data was parsed):

```c
#define tryRead(label, variable) \
         if (String_startsWith(buffer, label) { \
            sscanf(buffer + strlen(label), " %32llu kB", variable)) \
            break; \
         }

Also I like to align the trailing back-slashes

File: linux/LinuxProcessList.c [hunk 8]
```diff
@@ -1208,8 +1270,15 @@ static inline void LinuxProcessList_scanZfsArcstats(LinuxProcessList* lpl) {
    }
    char buffer[128];
    while (fgets(buffer, 128, file)) {
-      #define tryRead(label, variable) do { if (String_startsWith(buffer, label) && sscanf(buffer + strlen(label), " %*2u %32llu", variable)) { break; } } while(0)
-      #define tryReadFlag(label, variable, flag) do { if (String_startsWith(buffer, label) && sscanf(buffer + strlen(label), " %*2u %32llu", variable)) { (flag) = 1; break; } else { (flag) = 0; } } while(0)
+      #define tryRead(label, variable) \
```
cgzones, line 1273: dito

solaris/Platform.c
```diff
@@ -187,15 +193,17 @@ double Platform_setCPUValues(Meter* this, int cpu) {
...
    } else {
       v[2] = cpuData->systemAllPercent;
       this->curItems = 3;
-      percent = v[0]+v[1]+v[2];
+      percent = v[0] + v[1] + v[2];
    }
 
    percent = CLAMP(percent, 0.0, 100.0);
-   if (isnan(percent)) percent = 0.0;
+   if (isnan(percent)) {
```
cgzones, line 204: dito

dragonflybsd/Platform.c
BenBE: I'd prefer to still have it explicit, as the implementation of CLAMP might change and you don't want to rely on its implementation. Thus shortened to: `percent = isnan(percent) ? 0.0 : CLAMP(percent, 0.0, 100.0);`

linux/LinuxProcessList.c
BenBE: Actually, the original macro has another bug: It should have been `!sscanf(...)` all along.

linux/LinuxProcessList.c
BenBE: Same bug here too.

linux/LinuxProcessList.c
BenBE: Should be fixed in the new revision.

Issue #299, opened 2020-11-02
Disk IO meter crashes when USB storage device unplugged


lethosor:
- htop: `3.0.3-dev` (seen on both f8bd5ac and 15652e7, so likely not caused by #272 - 3.0.2 does not have the disk I/O meter)
- OS: Ubuntu 20.10 (`Linux 5.8.0-26-generic #27-Ubuntu SMP Wed Oct 21 22:29:16 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux`)

To reproduce:

1. Connect a USB storage device (I've tried with both a FAT32 and EXT4-formatted thumb drive)
2. Unmount the device
3. Physically remove the device (crashes with SIGABRT)

Stack trace from gdb (built with `make debug` from 15652e7b8102e86b3405254405d8ee5d2a239004):
```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:49
#1  0x00007ffff7c2e864 in __GI_abort () at abort.c:79
#2  0x000055555557bb9c in fail () at XUtils.c:23
#3  0x000055555557c422 in xSnprintf (buf=0x7fffffffad90 "117271100277873", len=16, 
    fmt=0x5555555884b3 "%.1f%%") at XUtils.c:192
#4  0x0000555555569554 in DIskIOMeter_display (cast=0x55555559eb30, out=0x7fffffffae20)
    at DiskIOMeter.c:85
#5  0x000055555556e9d7 in Meter_displayBuffer (this=0x55555559eb30, 
    buffer=0x7fffffffd4a0 "16.0EB 16.0EB 1172711002778737920.0%", out=0x7fffffffae20)
    at Meter.c:101
#6  0x000055555556ede0 in TextMeterMode_draw (this=0x55555559eb30, x=11, y=8, w=44)
    at Meter.c:160
#7  0x000055555556b655 in Header_draw (this=0x55555559e5b0) at Header.c:176
#8  0x0000555555577068 in checkRecalculation (this=0x55555559dc20, oldTime=0x7fffffffd6d0, 
    sortTimeout=0x7fffffffd6ac, redraw=0x7fffffffd6a9, rescan=0x7fffffffd6aa, 
    timedOut=0x7fffffffd6a8) at ScreenManager.c:119
#9  0x0000555555577280 in ScreenManager_run (this=0x55555559dc20, lastFocus=0x0, lastKey=0x0)
    at ScreenManager.c:162
[...]

PR #300, opened 2020-11-02
Handle data wraparounds in IO Meters

cgzones:
If the current data is smaller than the previous one, either by a retrieve error
or a device removal or a original data wraparound, sanitize the value to zero.

Fixes: #299


BenBE (approved):
Nit picking mode: "unit in kilo" -> kilo-what??? KiB would be better.

lethosor:
Confirmed that this fixes #299 on my end - thanks!

Issue #301, opened 2020-11-03
Show total resources of multi process applications


Nudin:
Computersystems nowadays have more and more CPU-Cores, due due to this more and more applications use multiprocessing to make better use of the hardware resources. This is increasingly getting an issue with htop (and similar tools).
On my desktop I currently have 35 processes by chromium, 11 by Firefox and 8 by my RSS reader. In university it is common to have machines with 128 CPU cores that are shared by several users who all run computations that may or may not use multiprocessing.
In both these cases it isn't possible anymore to use htop to figure out which applications are using most CPU/memory/….
What would be desirable is to see sum of recourse usage of all the processes that belong to one "application". This is of course not trivial since from the systems point of view all those processes are independent.

How others approach this issue:
- atop allows you to display the resources joined by user or processes name
- KDE/gnome is currently working on using cgroups with systemd ([[https://lwn.net/Articles/834329/|article on lwn]])

The first has the limitation that it cannot distinguish two instances of a multiprocessing application running in parallel but joins them to one application.
The second has the problem that 1. it's specific to Linux, 2. it needs systemd or an equivalent daemons to put the applications in cgroups and 3. it only works if the processes have been configured and tagged by the DE.

My proposal would be to use a better heuristic:
- Two Processes P1 and P2 are joined if:
	- P2 is the child of P1 *and*
	- P1 and P2 have the same executable *and*
	- P1 and P2 have the same user *and*
	- P1 and P2 have the same session

Therefore this process tree:
- 1 init
	- 2 foo
	- 3 bar
		- 4 bar
		- 5 bar
			- 6 bar
	- 7 baz
	- 8 bar
		- 9 bar

Would be merged to:
- 1 init
	- 2 foo
	- … bar
	- 7 baz
	- … bar

When enabled the rows off an 'application' that was created by merging 2 or more Processes should show:
- An indicator that it is not a process but an application instead of the pid in the PID column
- in columns like USER, PRI, NI, S:
	- if the values of all processes are the same, display it
	- otherwise display an indicator signalling that there are multiple values
- For fields, that show resource usage, show the combined resource usage
	- For some columns this is the sum of the values (example: cpu usage)
	- For others this might be harder to compute (example: shared memory)

This feature should be an opt-in (default off) in the options.

I know this is a big feature, hard to implement and the heuristic is kinda ugly, but as outlined this problem is already an issue and with the ongoing rise of numbers of cpu cores and containerisation it's likely getting bigger and bigger.

[...]

PR #303, opened 2020-11-04
Add option to merge processes of same application

cgzones:
Experimental - Linux only !!

Closes: #301

Based on #302


Process.h
```diff
@@ -115,6 +117,11 @@ typedef struct ProcessFieldData_ {
 // Implemented in platform-specific code:
 void Process_writeField(const Process* this, RichString* str, ProcessField field);
 long Process_compare(const void* v1, const void* v2);
+/* returns 1 on match and if v1 is the master process,
+ *         2 on match and if v2 is the master process,
+ *      NULL else */
```
BenBE, line 122: Should be 0; NULL should be reserved for the context of pointers.

PR #305, opened 2020-11-08
Improved command display

BenBE:
Replaces #42

State:
- Rebased onto master
- Adds command line space heuristic

Open issues:
- Only show magenta for things that changed in merged view

Note for `faccessat`: This currently only implements the specific case required for this PR.
Assumptions are documented in the function implementation.
If someone feels there should be a full compat layer for this function I'd appreciate a small patch.

For other caveats see the original PR discussion.


fasterit:
Very nice work!

Can you please replace "?" with "no perm" as in the DISK I/O meters for visual consistency.
For that reason also use their color (PROCESS_SHADOW). There is no reason for a failed read to be an alerting red. This is expected when running htop as a user.

marxin:
Thank you guys, it's a nice improvement!
Kudos!

PR #313, opened 2020-11-16
Coverity Findings

cgzones:


BenBE:
Can you somehow see why Coverity complains about `linebuf` being tained only after those many iterations of that loop? This looks illogical.

The other to patches LGTM.

cgzones:
```
** CID 215352:  Insecure data handling  (TAINTED_SCALAR)
/linux/LinuxProcessList.c: 602 in LinuxProcessList_readOpenVZData()


________________________________________________________________________________________________________
*** CID 215352:  Insecure data handling  (TAINTED_SCALAR)
/linux/LinuxProcessList.c: 602 in LinuxProcessList_readOpenVZData()
596           do {
597              name_value_sep++;
598           } while (*name_value_sep != '\0' && *name_value_sep <= 32);
599     
600           char* value_end = name_value_sep;
601     
>>>     CID 215352:  Insecure data handling  (TAINTED_SCALAR)
>>>     Using tainted variable "*value_end" as a loop boundary.
602           while(*value_end > 32) {
603              value_end++;
604           }
605     
606           if (name_value_sep == value_end) {
607              continue;
```


```
79. tainted_data_argument: Calling function fgets taints argument linebuf.
82. tainted_data_transitive: Call to function strchr with tainted argument linebuf returns tainted data.
83. var_assign: Assigning: name_value_sep = strchr, which taints name_value_sep.
89. var_assign_var: Assigning: value_end = name_value_sep. Both are now tainted.
CID 215352 (#1 of 1): Untrusted loop bound (TAINTED_SCALAR)90. tainted_data: Using tainted variable *value_end as a loop boundary.
```

BenBE:
I looked up the coverity report earlier. My question about it was more about:
What would you have to do in order to convince Coverity that `value_end` had been sanitized/untainted?

I mean apart from "ignoring the issue by adding a meta comment" to ignore the issue.

Cause the issue (IMHO that's a false-positive from Coverity) doesn't magically go away by writing the same statement, just with different operators …

cgzones:
Applied the first two patches.
The remaining commit is more a hide than a fix. I am fine with declaring it a false-positive at Covertity.

Issue #314, opened 2020-11-16
Performance with --enable-debug unbearable


BenBE:
When creating a build with `--enable-debug` active the resulting binary is so slow, that startup already takes several seconds and navigating around in the program is unbearably slow.

A quick look at `cachegrind` suggests the main culprits are the Vector and Hashtable consistency checks.

While I agree that these checks should be available in the code, I'd suggest having them split into a separate `enable-consistency-debug` switch for `configure`.

As it is now, I get 77.3% of all runtime spent in `Vector_isConsistent` and another 21.4% spent in `Object_isA`. One cycle takes about 6m45s, thus making debugging anything borderline impossible (unless you're willing to watch yourself aging in the process).

cgzones:
From my experience this slow-down is happening because of https://github.com/htop-dev/htop/blob/15eab2012d2100e1ddd20c186db23a8172b5858d/ProcessList.c#L176 and https://github.com/htop-dev/htop/blob/15eab2012d2100e1ddd20c186db23a8172b5858d/ProcessList.c#L131 in **tree**-mode. In **sorted**-mode the performance is quite ok.

cgzones:
@BenBE  can this be closed since #315 got merged?

BenBE:
Performance still isn't great overall even with this patch applied since `Vector_isConsistent` and `Hashtable_isConsistent` both are quite slow, but I think, for now we can close this as we are at least down into the "seconds per redraw" range (instead of minutes) …

Having the consistency stuff as a separate debug feature might be worth a thought.

PR #277, opened 2020-10-22
Hashtable update + dynamic growth

cgzones:

PR #315, opened 2020-11-17
Avoid calling Object_isA from inside Vector_isConsistent

BenBE:
This makes usage of `--enable-debug` somewhat bearable again.

Addresses #314 (probably still incomplete).
Basically drops `Object_isA` check from `Vector_isConsistent` as this gives O(n²) behaviour for most callers. Removes full consistency check from `Vector_get` to avoid O(n²) runtime there, but introduces `Object_isA` check on return value, so that all inserts and all fetches from the `Vector` perform storage type checking on accessed objects.

This still lacks the counterpart fixes for `Hashtable_isConsistent`, but given #277 is still pending similar precautions might better be addressed there instead.

### DIFF
--- a/Object.c
+++ b/Object.c
@@ -21,13 +21,10 @@ bool Object_isA(const Object* o, const ObjectClass* klass) {
    if (!o)
       return false;
 
-   const ObjectClass* type = o->klass;
-   while (type) {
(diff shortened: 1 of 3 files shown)


natoscott:
@BenBE LGTM.

PR #318, opened 2020-11-17
DarwinProcessList: retry getting list of all processes on ENOMEM

cgzones:
The process count might change between the two sysctl() calls getting
the size and getting the data.

Retry (3 times) in case the data-retrieval sysctl() call fails with ENOMEM.

see http://mirror.informatimago.com/next/developer.apple.com/qa/qa2001/qa1123.html

Related: #118

### DIFF
--- a/darwin/DarwinProcessList.c
+++ b/darwin/DarwinProcessList.c
@@ -8,6 +8,7 @@ in the source distribution for its full text.
 #include "DarwinProcessList.h"
 
 #include <err.h>
+#include <errno.h>
 #include <libproc.h>
 #include <stdbool.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


darwin/DarwinProcessList.c
```diff
@@ -93,39 +95,40 @@ unsigned ProcessList_allocateCPULoadInfo(processor_cpu_load_info_t* p) {
...
-      CRT_fatalError("Unable to get size of kproc_infos");
-   }
+   int retries = 3;
 
-   processes = xMalloc(*count);
-   if (processes == NULL) {
-      CRT_fatalError("Out of memory for kproc_infos");
-   }
+   while (retries > 0) {
+      retries--;
```
BenBE, line 113: Use for loop? OR at least inline the decrement?

darwin/DarwinProcessList.c
```diff
@@ -93,39 +95,40 @@ unsigned ProcessList_allocateCPULoadInfo(processor_cpu_load_info_t* p) {
...
 
-   if (sysctl(mib, 4, processes, count, NULL, 0) < 0) {
-      CRT_fatalError("Unable to get kinfo_procs");
-   }
+      size_t size = 0;
+      if (sysctl(mib, 4, NULL, &size, NULL, 0) < 0) {
+         CRT_fatalError("Unable to get size of kproc_infos");
+      }
+
+      processes = xRealloc(processes, size);
```
BenBE, line 120: Maybe over-allocate by a few entries?

BenBE: Also: size should never be zero by this point.

darwin/DarwinProcessList.c
```diff
@@ -93,39 +95,40 @@ unsigned ProcessList_allocateCPULoadInfo(processor_cpu_load_info_t* p) {
...
+         *count = size / sizeof(struct kinfo_proc);
+         return processes;
+      }
+
+      if (errno != ENOMEM)
+         break;
+   }
 
-   return processes;
+   CRT_fatalError("Unable to get kinfo_procs");
```
BenBE, line 131: Forward information about `errno`.

cgzones: `CRT_fatalError()` does print `strerror(errno)`

darwin/DarwinProcessList.c
cgzones: switched to for loop

darwin/DarwinProcessList.c
cgzones: added a check for `size == 0`

Do we need to allocate more? It *works* for couple hours already and the sysctl documentation says:
> The size of the available data can be determined by calling sysctl() with a NULL parameter for oldp. The size of the available data will be returned in the location pointed to by oldlenp. For some operations, the amount of space may change often. For these operations, the system attempts to round up so that the returned size is large enough for a call to return the data shortly thereafter.

BenBE (approved):

PR #321, opened 2020-11-17
LinuxProcessList: use openat instead of building path strings

cgzones:
openat() is available since Linux 2.6.16

### DIFF
--- a/Compat.h
+++ b/Compat.h
@@ -7,7 +7,11 @@ Released under the GNU GPLv2, see the COPYING file
 in the source distribution for its full text.
 */
 
+#include "config.h" // IWYU pragma: keep
+
+#include <fcntl.h>
 #include <stddef.h>
+#include <unistd.h>
 #include <sys/stat.h>
 
 
@@ -21,6 +25,30 @@ int Compat_fstatat(int dirfd,
                    const char* pathname,
                    struct stat* statbuf,
                    int flags);
+
+#ifdef HAVE_OPENAT
+
+typedef int openat_arg_t;
+
+static inline void Compat_openatArgClose(openat_arg_t dirfd) {
+   close(dirfd);
+}
+
+static inline int Compat_openat(openat_arg_t dirfd, const char* pathname, int flags) {
+   return openat(dirfd, pathname, flags);
+}
+
+#else /* HAVE_OPENAT */
+
+typedef const char* openat_arg_t;
+
+static inline void Compat_openatArgClose(openat_arg_t dirpath) {
+   (void)dirpath;
+}
+
+int Compat_openat(openat_arg_t dirpath, const char* pathname, int flags);
+
+#endif /* HAVE_OPENAT */
 
 int Compat_readlinkat(int dirfd,
                    const char* dirpath,
--- a/configure.ac
+++ b/configure.ac
@@ -92,7 +92,7 @@ AC_TYPE_UINT64_T
 # ----------------------------------------------------------------------
 AC_FUNC_CLOSEDIR_VOID
 AC_FUNC_STAT
-AC_CHECK_FUNCS([faccessat fstatat memmove readlinkat strdup strncasecmp strstr])
+AC_CHECK_FUNCS([faccessat fstatat memmove openat readlinkat strdup strncasecmp strstr])
 
 AC_SEARCH_LIBS([dlopen], [dl dld])
 
--- a/configure.ac
+++ b/configure.ac
@@ -92,7 +92,12 @@ AC_TYPE_UINT64_T
 # ----------------------------------------------------------------------
 AC_FUNC_CLOSEDIR_VOID
 AC_FUNC_STAT
-AC_CHECK_FUNCS([faccessat fstatat memmove openat readlinkat strdup strncasecmp strstr])
+AC_CHECK_FUNCS([\
+   faccessat\
+   fstatat\
+   openat\
+   readlinkat\
+])
 
 AC_SEARCH_LIBS([dlopen], [dl dld])
 
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -81,6 +81,20 @@ static ssize_t xread(int fd, void* buf, size_t count) {
    }
 }
 
+static FILE* fopenat(int dirfd, const char* pathname, const char* mode) {
+   assert(String_eq(mode, "r")); /* only currently supported mode */
+
+   int fd = openat(dirfd, pathname, O_RDONLY);
+   if (fd < 0)
+      return NULL;
+
+   FILE* stream = fdopen(fd, mode);
+   if (!stream)
+      close(fd);
+
+   return stream;
+}
+
 static int sortTtyDrivers(const void* va, const void* vb) {
    const TtyDriver* a = (const TtyDriver*) va;
    const TtyDriver* b = (const TtyDriver*) vb;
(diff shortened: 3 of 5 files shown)


BenBE:
`readlinkat` and `fstatat` have been available since then too and yet we put them in `Compat.h` …

linux/LinuxProcessList.c
```diff
@@ -1151,12 +1150,18 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, const char*
...
 
       // Exception handler.
 
 errorReadingProcess:
       {
+         if (procFd >= 0)
+            close(procFd);
+
          if (preExisting) {
             ProcessList_remove(pl, proc);
```
BenBE, line 1166: Shouldn't this trigger the diff highlighting logic?

linux/LinuxProcessList.c
```diff
@@ -1222,49 +1227,65 @@ static inline void LinuxProcessList_scanMemoryInfo(ProcessList* this) {
...
       unsigned long long int compr_data_size = 0;
 
       if (!fscanf(disksize_file, "%llu\n", &size) ||
           !fscanf(mm_stat_file, "    %llu       %llu", &orig_data_size, &compr_data_size)) {
-         fclose(disksize_file);
          fclose(mm_stat_file);
-         break;
+         fclose(disksize_file);
+         close(zramFd);
+         continue;
```
BenBE, line 1274: Why not just force `size`, `orig_data_size` and `compr_data_size` all to zero here and fall through?

Or invert this condition and handle the three additions from below in here.

linux/LinuxProcessList.c
```diff
@@ -1605,5 +1626,13 @@ void ProcessList_goThroughEntries(ProcessList* super, bool pauseProcessUpdate) {
...
-   LinuxProcessList_recurseProcTree(this, PROCDIR, NULL, period, tv);
+   unsigned long long now = tv.tv_sec * 1000ULL + tv.tv_usec / 1000ULL;
+
+   int procDirfd = open(PROCDIR, O_RDONLY | O_DIRECTORY | O_NOFOLLOW);
+   if (procDirfd < 0) {
+      CRT_fatalError("Can not open the root process directory " PROCDIR);
+   }
+
+   /* LinuxProcessList_recurseProcTree() closes procDirfd */
+   LinuxProcessList_recurseProcTree(this, procDirfd, NULL, period, now);
```
BenBE, line 1637: Not quite happy with this design. IMHO a resource provided by the caller should be freed by the caller. Makes for a few more LOC, but is cleaner from a resource management PoV IMO.

linux/LinuxProcessList.c
cgzones: Diff highlighting is fixed now

linux/LinuxProcessList.c
cgzones: dropped the zram changes, no real gain

Action.c
```diff
@@ -216,6 +216,16 @@ static Htop_Reaction actionToggleKernelThreads(State* st) {
...
 static Htop_Reaction actionToggleUserlandThreads(State* st) {
    st->settings->hideUserlandThreads = !st->settings->hideUserlandThreads;
+
+   if (st->settings->hideUserlandThreads) {
+      /* Do not show userland threads as tombed */
+      ProcessList_noTombedUserlandThreads(st->pl);
+   } else {
+      /* Do not show userland threads as new */
+      ProcessList_scan(st->pl, false);
+      ProcessList_noNewUserlandThreads(st->pl);
```
BenBE, line 226: This discards legitimately new userland threads.

configure.ac
```diff
@@ -88,7 +88,7 @@ AC_TYPE_UID_T
 # ----------------------------------------------------------------------
 AC_FUNC_CLOSEDIR_VOID
 AC_FUNC_STAT
-AC_CHECK_FUNCS([fstatat memmove readlinkat strdup strncasecmp strstr])
+AC_CHECK_FUNCS([fstatat memmove openat readlinkat strdup strncasecmp strstr])
```
BenBE, line 91: What about splitting this list one function per line?

Action.c
cgzones: With this patch userland threads are not parsed at all when disabled. So the are legitimately new on first scan after toggle-ling the setting.
It will only affect userland threads started **exactly** at the same time as the setting is switched, as there is one scan before (line 225), and one scan directly after https://github.com/htop-dev/htop/blob/be568b1153206a067fd09c0f7d2dcea3d7dbaa68/ScreenManager.c#L115 (cause we return `HTOP_RECALCULATE`)

BenBE: What about tracking their PIDs (with `show==false`), but not filling in any other details while hiding userland threads?

cgzones: moved the thread skipping to #334

linux/LinuxProcessList.c
cgzones: reworked

configure.ac
```diff
@@ -88,7 +88,7 @@ AC_TYPE_UID_T
 # ----------------------------------------------------------------------
 AC_FUNC_CLOSEDIR_VOID
 AC_FUNC_STAT
-AC_CHECK_FUNCS([faccessat fstatat memmove readlinkat strdup strncasecmp strstr])
+AC_CHECK_FUNCS([faccessat fstatat openat readlinkat])
```
BenBE, line 91:
```suggestion
AC_CHECK_FUNCS([\
   faccessat\
   fstatat\
   openat\
   readlinkat\
])
```

cgzones: changed

configure.ac
cgzones: done

BenBE:
Can you summarize what is missing for this PR?

cgzones:
Tested (normal and compat version).

*Should* work.

natoscott (approved):
LGTM.

Compat.h
```diff
@@ -22,6 +26,34 @@ int Compat_fstatat(int dirfd,
...
+
+typedef int openat_arg_t;
+
+static inline void Compat_openatArgClose(openat_arg_t dirfd) {
+   close(dirfd);
+}
+
+static inline int Compat_openat(openat_arg_t dirfd,
+                                const char* pathname,
+                                int flags) {
```
BenBE, line 39: No need to wrap this argument list …

Compat.h
```diff
@@ -22,6 +26,34 @@ int Compat_fstatat(int dirfd,
...
+
+typedef const char* openat_arg_t;
+
+static inline void Compat_openatArgClose(openat_arg_t dirpath) {
+   (void)dirpath;
+}
+
+int Compat_openat(openat_arg_t dirpath,
+                  const char* pathname,
+                  int flags);
```
BenBE, line 53: Dito.

linux/LinuxProcessList.c
```diff
@@ -1193,9 +1175,16 @@ static bool LinuxProcessList_readCmdlineFile(Process* process, const char* dirna
...
    /* execve could change /proc/[pid]/exe, so procExe should be updated */
-   xSnprintf(command, sizeof(command), "%s/%s/exe", dirname, name);
-   if ((amtRead = readlink(command, filename, sizeof(filename) - 1)) > 0) {
+#if defined(HAVE_READLINKAT) && defined(HAVE_OPENAT)
+   if ((amtRead = readlinkat(procFd, "exe", filename, sizeof(filename) - 1)) > 0) {
+#else
+   char path[4096];
+   xSnprintf(path, sizeof(path), "%s/exe", procFd);
+   if ((amtRead = readlink(path, filename, sizeof(filename) - 1)) > 0) {
+#endif
```
BenBE, line 1187: I'd prefer if we could avoid starting a block within conditionally compiled sections of code.
```suggestion
#if defined(HAVE_READLINKAT) && defined(HAVE_OPENAT)
   amtRead = readlinkat(procFd, "exe", filename, sizeof(filename) - 1);
#else
   char path[4096];
   xSnprintf(path, sizeof(path), "%s/exe", procFd);
   amtRead = readlink(path, filename, sizeof(filename) - 1);
#endif
   if (amtRead > 0) {
```

linux/LinuxProcessList.c
```diff
@@ -1484,12 +1492,20 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, const char*
...
       proc->updated = true;
+
+      Compat_openatArgClose(procFd);
+
       continue;
 
       // Exception handler.
 
 errorReadingProcess:
       {
```
BenBE, line 1503: Any reason for this extra code block for grouping? Doesn't it do without?

BenBE (commented):
Minor stylistic notes. Apart from this LGTM.

PR #323, opened 2020-11-18
IWYU update

cgzones:


Compat.c
```diff
@@ -7,13 +7,16 @@ in the source distribution for its full text.
...
 
-#include <fcntl.h>
+#include "Compat.h"
+
 #include <unistd.h>
-#include <sys/types.h>
 #include <sys/stat.h>
 
-#include "Compat.h"
+#ifndef HAVE_FSTATAT
```
BenBE, line 15: Not sure if we really have to include those headers conditionally. All for it if it were headers that might be missing, but for normal system headers this is probably a bit overkill.

BenBE (commented):
LGTM. Two minor suggestions.

PR #324, opened 2020-11-19
Minor cleanups to platform-specific init and done

natoscott:
Move platform-specific code out of the htop.c main function
and into the platform sub-directories - primarily this is
the Linux procfs path check and sensors setup/teardown; not
needed on any other platforms.  No functional changes here.

### DIFF
--- a/freebsd/Platform.h
+++ b/freebsd/Platform.h
@@ -26,9 +26,13 @@ extern const SignalItem Platform_signals[];
 
 extern const unsigned int Platform_numberOfSignals;
 
+extern const MeterClass* const Platform_meterTypes[];
+
 void Platform_setBindings(Htop_Action* keys);
 
-extern const MeterClass* const Platform_meterTypes[];
+void Platform_init(void);
+
+void Platform_done(void);
 
 int Platform_getUptime(void);
 
--- a/freebsd/Platform.h
+++ b/freebsd/Platform.h
@@ -28,11 +28,11 @@ extern const unsigned int Platform_numberOfSignals;
 
 extern const MeterClass* const Platform_meterTypes[];
 
-void Platform_setBindings(Htop_Action* keys);
-
 void Platform_init(void);
 
 void Platform_done(void);
+
+void Platform_setBindings(Htop_Action* keys);
 
 int Platform_getUptime(void);
 
--- a/htop.c
+++ b/htop.c
@@ -35,18 +35,12 @@ in the source distribution for its full text.
 #include "UsersTable.h"
 #include "XUtils.h"
 
-
-#ifdef HAVE_LIBSENSORS
-#include <sensors/sensors.h>
-#endif
-
-
 static void printVersionFlag(void) {
-   fputs("htop " VERSION "\n", stdout);
+   fputs(PACKAGE " " VERSION "\n", stdout);
 }
 
 static void printHelpFlag(void) {
-   fputs("htop " VERSION "\n"
+   fputs(PACKAGE " " VERSION "\n"
          COPYRIGHT "\n"
          "Released under the GNU GPLv2.\n\n"
          "-C --no-color                   Use a monochrome color scheme\n"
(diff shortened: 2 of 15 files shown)


freebsd/Platform.h
```diff
@@ -26,9 +26,13 @@ extern const SignalItem Platform_signals[];
 
 extern const unsigned int Platform_numberOfSignals;
 
+extern const MeterClass* const Platform_meterTypes[];
+
 void Platform_setBindings(Htop_Action* keys);
```
BenBE, line 31: Shouldn't this go after the `Platform_done` function for consistent order?

htop.c
```diff
@@ -270,22 +264,16 @@ static void setCommFilter(State* state, char** commFilter) {
...
-   if (lc_ctype != NULL) {
+   if (lc_ctype != NULL)
       setlocale(LC_CTYPE, lc_ctype);
-   } else if ((lc_ctype = getenv("LC_ALL"))) {
+   else if ((lc_ctype = getenv("LC_ALL")))
       setlocale(LC_CTYPE, lc_ctype);
-   } else {
+   else
       setlocale(LC_CTYPE, "");
-   }
```
BenBE, line 279: `if` statements with an else clause should keep braces (cf. styleguide).

linux/Platform.c
```diff
@@ -107,6 +108,22 @@ static time_t Platform_Battery_cacheTime;
 static double Platform_Battery_cacheLevel = NAN;
 static ACPresence Platform_Battery_cacheIsOnAC;
 
+void Platform_init(void) {
+   if (access(PROCDIR, R_OK) != 0) {
+      fprintf(stderr, "Error: could not read procfs (compiled to look in %s).\n", PROCDIR);
+      exit(1);
+   }
+#ifdef HAVE_LIBSENSORS
```
BenBE, line 116: Add a blank line for better visual structure.

linux/Platform.c
```diff
@@ -107,6 +108,22 @@ static time_t Platform_Battery_cacheTime;
 static double Platform_Battery_cacheLevel = NAN;
 static ACPresence Platform_Battery_cacheIsOnAC;
 
+void Platform_init(void) {
+   if (access(PROCDIR, R_OK) != 0) {
+      fprintf(stderr, "Error: could not read procfs (compiled to look in %s).\n", PROCDIR);
+      exit(1);
+   }
```
BenBE, line 115: Isn't this check common to all platforms? Also maybe have `Platform_init` return `bool` and keep exitting to `htop.c`.

BenBE (changes requested):

BenBE:
Maybe also rephrase the title as "Introduce interface for platform-specific initialization".

freebsd/Platform.h
natoscott: Heh.  This was a of case of a variable being intermingled with function declarations, while all other variables were declared together near the start of the header, so I shifted this variable declaration together with the others for consistent ordering :P

htop.c
natoscott: Hmm - guess I've missed this part of the styleguide before.  I disagree with that being in the styleguide - its not consistent with plenty of existing code (which *is* pretty inconsistent though TBH) and IMO, in this simple case, it introduces syntactic noise that clutters the code and reduces readability.  Let's discuss that elsewhere though - I'll send a follow up PR to the styleguide for wider discussion.

linux/Platform.c
natoscott: +1

linux/Platform.c
natoscott: Nah, PROCDIR is only used on Linux.

natoscott: Oh, also, other functions in htop.c (like arg parsing) exit at the point of failure, so I'd rather leave it as-is for now.

freebsd/Platform.h
BenBE: I was referring to the `Platform_setBindings` declaration, because on all other platforms this one comes after the `Platform_done` declaration.

BenBE (approved):
No "blocking" changes left. Fixing the one remaining issue with declaration order appreciated though.

freebsd/Platform.h
natoscott: @BenBE ah I see - yep, will tweak that to be more consistent.

htop.c
natoscott: https://github.com/htop-dev/htop/pull/336

PR #327, opened 2020-11-20
Add compiler hints on memory allocating utility functions

cgzones:
Might benefit some optimizations or some warnings.

### DIFF
--- a/XUtils.h
+++ b/XUtils.h
@@ -21,15 +21,15 @@ in the source distribution for its full text.
 
 void fail(void) ATTR_NORETURN;
 
-void* xMalloc(size_t size);
+void* xMalloc(size_t size) ATTR_ALLOC_SIZE1(1) ATTR_MALLOC;
 
-void* xMallocArray(size_t nmemb, size_t size);
+void* xMallocArray(size_t nmemb, size_t size) ATTR_ALLOC_SIZE2(1, 2) ATTR_MALLOC;
 
-void* xCalloc(size_t nmemb, size_t size);
+void* xCalloc(size_t nmemb, size_t size) ATTR_ALLOC_SIZE2(1, 2) ATTR_MALLOC;
 
-void* xRealloc(void* ptr, size_t size);
+void* xRealloc(void* ptr, size_t size) ATTR_ALLOC_SIZE1(2);
 
-void* xReallocArray(void* ptr, size_t nmemb, size_t size);
+void* xReallocArray(void* ptr, size_t nmemb, size_t size) ATTR_ALLOC_SIZE2(2, 3);
 
 /*
  * String_startsWith gives better performance if strlen(match) can be computed
@@ -47,17 +47,17 @@ static inline bool String_eq(const char* s1, const char* s2) {
    return strcmp(s1, s2) == 0;
 }
 
-char* String_cat(const char* s1, const char* s2);
+char* String_cat(const char* s1, const char* s2) ATTR_MALLOC;
 
-char* String_trim(const char* in);
+char* String_trim(const char* in) ATTR_MALLOC;
 
 char** String_split(const char* s, char sep, size_t* n);
 
 void String_freeArray(char** s);
 
-char* String_getToken(const char* line, unsigned short int numMatch);
+char* String_getToken(const char* line, unsigned short int numMatch) ATTR_MALLOC;
 
-char* String_readLine(FILE* fd);
+char* String_readLine(FILE* fd) ATTR_MALLOC;
 
 /* Always null-terminates dest. Caller must pass a strictly positive size. */
 size_t String_safeStrncpy(char *restrict dest, const char *restrict src, size_t size);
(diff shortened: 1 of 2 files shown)


XUtils.h
```diff
@@ -19,11 +19,11 @@ in the source distribution for its full text.
...
 void fail(void) ATTR_NORETURN;
 
-void* xMalloc(size_t size);
+void* xMalloc(size_t size) ATTR_ALLOC_SIZE1(1) ATTR_MALLOC;
 
-void* xCalloc(size_t nmemb, size_t size);
+void* xCalloc(size_t nmemb, size_t size) ATTR_ALLOC_SIZE2(1, 2) ATTR_MALLOC;
 
-void* xRealloc(void* ptr, size_t size);
+void* xRealloc(void* ptr, size_t size) ATTR_ALLOC_SIZE1(2);
```
BenBE, line 26: Quoting from [GCC documentation](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html) (emphasis mine):

> This tells the compiler that a function is `malloc`-like, i.e., that **the pointer `P` returned by the function cannot alias any other pointer valid when the function returns**, and moreover **no pointers to valid objects occur in any storage addressed by `P`**.
>
> Using this attribute can improve optimization. Compiler predicts that a function with the attribute returns non-null in most cases. Functions like `malloc` and `calloc` have this property because they return a pointer to uninitialized or zeroed-out storage. **However, functions like `realloc` do not have this property, as they can return a pointer to storage containing pointers.**

Maybe also include `returns_nonnull` for the flags we use? As we force `abort` on error, there is no way the function may return `NULL`.

XUtils.h
```diff
@@ -41,26 +41,26 @@ static inline bool String_eq(const char* s1, const char* s2) {
...
 }
 
-char* String_cat(const char* s1, const char* s2);
+char* String_cat(const char* s1, const char* s2) ATTR_MALLOC;
 
-char* String_trim(const char* in);
+char* String_trim(const char* in) ATTR_MALLOC;
 
-char** String_split(const char* s, char sep, size_t* n);
+char** String_split(const char* s, char sep, size_t* n) ATTR_MALLOC;
```
BenBE, line 48: Violates no-pointers requirement.

XUtils.h
```diff
@@ -41,26 +41,26 @@ static inline bool String_eq(const char* s1, const char* s2) {
...
 
 void String_freeArray(char** s);
 
-char* String_getToken(const char* line, unsigned short int numMatch);
+char* String_getToken(const char* line, unsigned short int numMatch) ATTR_MALLOC;
```
BenBE, line 52: Violates no-aliasing requirement.

XUtils.h
```diff
@@ -41,26 +41,26 @@ static inline bool String_eq(const char* s1, const char* s2) {
...
 int xAsprintf(char** strp, const char* fmt, ...);
 
 ATTR_FORMAT(printf, 3, 4)
 int xSnprintf(char* buf, int len, const char* fmt, ...);
 
-char* xStrdup(const char* str) ATTR_NONNULL;
+char* xStrdup(const char* str) ATTR_NONNULL ATTR_MALLOC;
 
-char* xStrndup(const char* str, size_t len) ATTR_NONNULL;
+char* xStrndup(const char* str, size_t len) ATTR_NONNULL ATTR_MALLOC;
```
BenBE, line 64: Maybe `ALLOC_SIZE(2)`?

Vector.h
```diff
@@ -25,7 +25,7 @@ typedef struct Vector_ {
    bool owner;
 } Vector;
 
-Vector* Vector_new(const ObjectClass* type, bool owner, int size);
+Vector* Vector_new(const ObjectClass* type, bool owner, int size) ATTR_MALLOC;
```
BenBE, line 28: Violates no-pointers requirement.

UsersTable.h
```diff
@@ -13,7 +13,7 @@ typedef struct UsersTable_ {
    Hashtable* users;
 } UsersTable;
 
-UsersTable* UsersTable_new(void);
+UsersTable* UsersTable_new(void) ATTR_MALLOC;
```
BenBE, line 16: Violates no-pointers requirement.

BenBE (commented):
Not completely sure how strict the "no valid pointers in returned storage" is to be taken.

XUtils.h
cgzones: dropped

XUtils.h
cgzones: no, we are allocating `size + 1`

Vector.h
cgzones: dropped

UsersTable.h
cgzones: dropped

XUtils.h
cgzones: It does return a new copy
https://github.com/htop-dev/htop/blob/be568b1153206a067fd09c0f7d2dcea3d7dbaa68/XUtils.c#L143

XUtils.h
cgzones: Maybe at some point..

XUtils.h
BenBE: Would have been to nice, if you could provide a delta to the compiler too …

BenBE (approved):

PR #330, opened 2020-11-21
Revamp keyboard input

BenBE:
This cleans up the keyboard input stuff somewhat and allows for using 'q' to exit any of the info screens.

The downside is, that this interferes with the inline search, thus searching for anything that starts with 'q' is no longer possible.


BenBE:
The issue with searching for anything starting in `q` has been resolved by introducing a vim-like `/` to trigger the search (acting to escape the first character).

PR #332, opened 2020-11-22
Various issues noted by cppcheck

BenBE:


cgzones:
Isn't the remaining commit the opposite of what cppcheck suggests?

Maybe move the variables into the `if (passedTimeInMs > 500) {` block?

BenBE:
> Isn't the remaining commit the opposite of what cppcheck suggests?

Yes and no.

From a storage PoV no. From the visibility PoV yes.

> Maybe move the variables into the `if (passedTimeInMs > 500) {` block?

Will do.

I moved them out to group them with the existing variables that are also used in this context, to make their combined use more obvious.

cgzones:
I prefer it this way, cause now the compiler guarantees the value is not changed behind our back in an other function.

cgzones (approved):

PR #333, opened 2020-11-22
Add support to change numeric options in settings screen

cgzones:
Like `delay` or `highlightDelaySecs`


OptionItem.c
```diff
@@ -0,0 +1,181 @@
...
+   this->max = max;
+   return this;
+}
+
+NumberItem* NumberItem_newByVal(const char* text, int value, int min, int max) {
+   assert(min <= max);
+
+   NumberItem* this = AllocThis(NumberItem);
+   this->text = xStrdup(text);
+   this->value = value;
```
BenBE, line 138: Force min/max clamping on ctor?

OptionItem.h
```diff
@@ -0,0 +1,68 @@
...
+#define OptionItem_kind(this_)              As_OptionItem(this_)->kind
+
+typedef struct OptionItem_ {
+   Object super;
+} OptionItem;
+
+typedef struct CheckItem_ {
+   OptionItem super;
+
+   char* text;
```
BenBE, line 36: Put `text` in `OptionItem` as common field?

OptionItem.c
```diff
@@ -0,0 +1,181 @@
...
+   free(this);
+}
+
+static void NumberItem_display(const Object* cast, RichString* out) {
+   const NumberItem* this = (const NumberItem*)cast;
+   assert (this != NULL);
+
+   char buffer[12];
+   RichString_write(out, CRT_colors[CHECK_BOX], "[");
+   xSnprintf(buffer, sizeof(buffer), "%d", NumberItem_get(this));
```
BenBE, line 53: Allow for sclaed display?

E.g. for the refresh delay it stores 1 internally, but shows 0.1 using a provided scale factor?

BenBE (commented):
Looks good so far.

OptionItem.c
cgzones: added

OptionItem.h
cgzones: changed

OptionItem.c
cgzones: then it gets complicated, e.g. how to display `0.1` as it is no exact floating point number.

BenBE: Two options:
1. Just use a fixed point display (number of digits behind decimal point provided), display with zero-padding at least `fixed-point+1` digits, insert decimal point at correct place.
2. `xSnprintf("%.*f", decimal_places, (double)number/scale);` (cf. https://stackoverflow.com/a/16413665)
3. Allow for custom display handler to be provided (probably overkill)

Option 2 is probably fully sufficient and easiest to implement.

cgzones: Support for scaling in both directions (`*10`, `/10`) and scaling with non-multiples of 10 will also complicate things.
d
After all is this really needed or can we expect user who want to change the scan interval what deci-seconds are?

BenBE: I think it's more intuitive, if the user sees fractional seconds. I don't think deciseconds is common enough that the average user will understand without mentally stumpbling over it first.

Scaling with just `*10` and `/10` (or powers thereof) should suffice (and works as suggested above).

cgzones:
Note: when merging, check for new introduced configuration settings and remove redundant `xstrdup` calls in first arguments of `CheckItem_newByRef/CheckItem_newByValue`. (e.g. #305)

BenBE (approved):
Space with `-H10000` one too little, but still acceptable.

PR #336, opened 2020-11-23
Tweak style guide wording around single code statements

natoscott:
There was wording about brace-enclosing single code statements
being 'strongly encouraged' - this isn't consistently used and
IMO it introduces unnecessary noise in otherwise neat, concise
code.

I've reworded (dropped) this section and also fixed a handful
of minor typos while reading this doc a little more carefully.

### DIFF
--- a/docs/styleguide.md
+++ b/docs/styleguide.md
@@ -78,7 +78,7 @@ Symbol Exports
 
 Exports of symbols should be used sparingly.
 Thus unless a function you write is intended to become public API of a module you should mark it as `static`.
-If a function should be public API an appropriate declaration for that function has to be placed in the accompaying header file.
+If a function should be public API an appropriate declaration for that function has to be placed in the accompanying header file.
 
 Please avoid function-like macros, in particular when exporting them in a header file.
 They have several downsides (re-evaluation of arguments, syntactic escapes, weak typing) for which usually a better alternative like an actual function exists.
@@ -107,7 +107,7 @@ Styling the code
 ----------------
 
 Now for the style details that can mostly be automated: Indentation, spacing and bracing.
-While there is no definitve code style we use, a set of rules loosely enforced has evolved.
+While there is no definitive code style we use, a set of rules loosely enforced has evolved.
 
 Indentation in the code is done by three (3) spaces. No tabs are used. Ever.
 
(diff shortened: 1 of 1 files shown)


BenBE:
The wording was deliberately "strongly encouraged" and not "required".
The reason why I prefer them to be present is twofold:
1. Simplifies extensibility when another statement needs to be added
2. Makes introducing a `goto fail`-style bug a lot harder.

While the first one is a mostly stylistic, the second one is a security and defense-in-depth related issue.

In code-bases I usually work with at work, braces are even enforced by code style.

cgzones:
From my personal preference I like to skip braces around single-line branches, but I am willing to follow the majority.

Maybe we can suggest adding an empty line after a single-line branch, to avoid coding errors?

Also modern compilers ([gcc](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wno-misleading-indentation), [clang](https://clang.llvm.org/docs/DiagnosticsReference.html#wmisleading-indentation)) warn on
```c
if (x)
    bar();
    foo();
```

BenBE:
I'm fine with skipping the braces for single `if` (no `else`) any time. A blank before/after for visual grouping is appreciated.

Where the disagreement mostly comes from is with `if`/`else` constructs with both parts (or even more complex ones with `else if`. My preference, mostly originating from the code I see daily, is to include optional braces for blocks, even if not strictly required.

IMHO `;` closes a statement, thus having the next line still continue that basic structure kinda trips me off when reading. With braces it's `if (cond) { statement; } else { /* … */ }` thus making the continuation much clearer.

natoscott:
> From my personal preference I like to skip braces around single-line branches, but I am willing to follow the majority.
> 
> Maybe we can suggest adding an empty line after a single-line branch, to avoid coding errors?
> 
> Also modern compilers ([gcc](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wno-misleading-indentation), [clang](https://clang.llvm.org/docs/DiagnosticsReference.html#wmisleading-indentation)) warn on
> 
> ```c
> if (x)
>     bar();
>     foo();
> ```

+1 - my preference is also to skip braces where possible.  I suspect coding styles which mandate (or 'strongly recommend') the additional braces pre-date the toolchain warnings about the above class of problem - used to be a big problem but that's no longer the case.

natoscott:
> Where the disagreement mostly comes from is with `if`/`else` constructs with both parts (or even more complex ones with `else if`.

Yep, this is the only aspect of contention I think.

> IMHO `;` closes a statement,

I think this statement is kinda wishful thinking though - in the case of an if/else construct, the ';' doesn't close the statement - we're only part-way done.  The 'else' cannot exist without the 'if', its part of the one single construct.

BenBE:
> > IMHO `;` closes a statement,
> 
> I think this statement is kinda wishful thinking though - in the case of an if/else construct, the ';' doesn't close the statement - we're only part-way done. The 'else' cannot exist without the 'if', its part of the one single construct.

Syntactically sure. But visually it's kinda confusing nonetheless …

natoscott:
> [...]
> Syntactically sure. But visually it's kinda confusing nonetheless …

OK - it seems like we've got different ideas on this one and the existing code base uses both forms, so there's no single clear option.  Luckily its a really minor syntax thing :) - I think our best bet will be to allow both, and not discourage either form.  Original author (of new code) chooses what they want to use in terms of single-line bracing, and we stick with that as best we can.

@BenBE @cgzones how does that sound?

fasterit:
>  I think our best bet will be to allow both, and not discourage either form. Original author (of new code) chooses what they want to use in terms of single-line bracing, and we stick with that as best we can.

I think our code base is small enough that we can seek consistency.
It is ugly and confusing to have no consistent style in a code base.

I'm fine with curly braces or without, but let's agree on one style and use that.

docs/styleguide.md
```diff
@@ -128,41 +128,80 @@ if (very_long_condition &&
...
 ```
 
-While this code style isn't fully consistent in the existing code base it is strongly recommended that new code follows these rules.
+While the existing code base isn't fully consistent with this code style isn't fully consistent yet it is strongly recommended that new code follows these rules.
 Adapting surrounding code near places you need to touch is encouraged.
 Try to separate such changes into a single, clean-up only commit to reduce noise while reviewing your changes.
 
+When writing your code consistency with the surrounding codebase is favoured.
+
+Don't shy away from leaving (sinlge) blank lines to separate different groups of related statements.
```
natoscott, line 174: single

docs/styleguide.md
```diff
@@ -128,41 +128,80 @@ if (very_long_condition &&
...
-   // something else
+   metric = -1;
 }
+
+if (fd >= 0)
+   close(fd);
 ```
 
-While this code style isn't fully consistent in the existing code base it is strongly recommended that new code follows these rules.
+While the existing code base isn't fully consistent with this code style isn't fully consistent yet it is strongly recommended that new code follows these rules.
```
natoscott, line 168: One too many "isn't fully consistent" phrases here?

docs/styleguide.md
```diff
@@ -128,41 +128,80 @@ if (very_long_condition &&
...
+Control flow statements and the instruction making up their body should not be put on a single line,
+i.e. after the condition of an if statement a new line should be inserted and the body indented accordingly.
+
+```c
+if (answer)
+   return 42;
+else if (again)
+   continue;
+else
+   goto bowl_of_petunias;
```
natoscott, line 149: Not sure about suggesting "goto" in an example ;) ... could use 'break' perhaps, but its a trivial thing - up to you entirely.

natoscott (commented):
Looks good @BenBE - couple minor suggestions, but ACK from me in terms of merging.

docs/styleguide.md
BenBE: The name of the label and the goto itself is intentional (references to the Hitchhiker's Guide to the Galaxy, chapter 18).

docs/styleguide.md
BenBE:
```suggestion
While the existing code base isn't fully consistent with this code style yet it is strongly recommended that new code follows these rules.
```

PR #337, opened 2020-11-23
Hide process selection on ESC

cgzones:
Do not highlight the current process line after pressing ESC in the main
screen.
Restore after pressing any key.

Normal:
pre

After ESC:
post


BenBE:
Can you please explain this a bit more in detail? Maybe I'm missing something …

Issue #329, opened 2020-11-20
visual glitches with cpu temperature


eworm-de:
Enabling cpu temperature in cpu meter results in visual glitches:
![Screenshot_2020-11-20_17-33-47](https://user-images.githubusercontent.com/970853/99826233-5e0da100-2b58-11eb-9736-7a712e31f4fb.png)
* The last character is never filled. This is because the degree character is encoded multi-byte: It counts two, but displays just one.
* Adding color codes in between the degree bytes breaks the display even more.

PR #338, opened 2020-11-23
Allow non-ascii characters in Meter-Bar

cgzones:
E.g. degree sign for temperature.

Closes: #329


eworm-de:
Works for me, thanks a lot!

PR #339, opened 2020-11-23
LinuxProcess_adjustTime: simplify by not using double

cgzones:
Does not work with -ffast-math else.

    linux/LinuxProcessList.c:306:11: runtime error: nan is outside the range of representable values of type 'unsigned long long'
    SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior linux/LinuxProcessList.c:306:11 in 
    linux/LinuxProcessList.c:1089:50: runtime error: unsigned integer overflow: 426 - 140733456409269 cannot be represented in type 'unsigned long long'
    SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior linux/LinuxProcessList.c:1089:50 in

### DIFF
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -292,18 +292,16 @@ void ProcessList_delete(ProcessList* pl) {
 }
 
 static inline unsigned long long LinuxProcess_adjustTime(unsigned long long t) {
-   static double jiffy = NAN;
-   if (isnan(jiffy)) {
+   static long jiffy = -1;
+   if (jiffy == -1) {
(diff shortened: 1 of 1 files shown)


cgzones:
Note:
We are using NAN as a special value for data that is not available, e.g. CPU frequency/temperature.
So *-ffast-math* causes to ignore `isnan()` checks and garbage will be printed (which is somewhat fine if one wants to use that flag).
But `LinuxProcess_adjustTime()` is essential for basic functionality like the `CPU%` and `TIME+` columns.

PR #341, opened 2020-11-25
Add xReadfile wrapper for reading small to medium size files

cgzones:
Inspired by proposed Linux syscall

Avoid file descriptor leaks like 4af8c63f

### DIFF
--- a/XUtils.c
+++ b/XUtils.c
@@ -10,6 +10,7 @@ in the source distribution for its full text.
 #include "XUtils.h"
 
 #include <assert.h>
+#include <errno.h>
 #include <stdarg.h>
 #include <stdlib.h>
 #include <string.h>
@@ -209,4 +210,53 @@ char* xStrndup(const char* str, size_t len) {
       fail();
    }
    return data;
+}
+
+static ssize_t readfd_internal(int fd, void* buffer, size_t count) {
+   if (!count) {
+      close(fd);
+      return -EINVAL;
+   }
+
+   ssize_t alreadyRead = 0;
+   count--; // reserve one for null-terminator
+
+   for (;;) {
+      ssize_t res = read(fd, buffer, count);
+      if (res == -1) {
+         if (errno == EINTR)
+            continue;
+
+         close(fd);
+         return -errno;
+      }
+
+      if (res > 0) {
+         buffer = ((char*)buffer) + res;
+         count -= (size_t)res;
+         alreadyRead += res;
+      }
+
+      if (count == 0 || res == 0) {
+         close(fd);
+         *((char*)buffer) = '\0';
+         return alreadyRead;
+      }
+   }
+}
+
+ssize_t xReadfile(const char* pathname, void* buffer, size_t count) {
+   int fd = open(pathname, O_RDONLY);
+   if (fd < 0)
+      return -errno;
+
+   return readfd_internal(fd, buffer, count);
+}
+
+ssize_t xReadfileat(openat_arg_t dirfd, const char* pathname, void* buffer, size_t count) {
+   int fd = Compat_openat(dirfd, pathname, O_RDONLY);
+   if (fd < 0)
+      return -errno;
+
+   return readfd_internal(fd, buffer, count);
 }
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -58,29 +58,6 @@ in the source distribution for its full text.
 #endif
 
 
-static ssize_t xread(int fd, void* buf, size_t count) {
-   // Read some bytes. Retry on EINTR and when we don't get as many bytes as we requested.
-   size_t alreadyRead = 0;
-   for (;;) {
-      ssize_t res = read(fd, buf, count);
-      if (res == -1) {
-         if (errno == EINTR)
-            continue;
-         return -1;
-      }
-
-      if (res > 0) {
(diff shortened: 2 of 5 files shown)


BenBE:
Is it intentional that this does not allow to detect the file being too large?

What about `openat` support?

cgzones:
> Is it intentional that this does not allow to detect the file being too large?

Yes.
One can check if the return value equals the passed `count`, then the file is probably bigger (except it has unlikely the size `count`).
But the current code this wrapper replaces does not check for this case and for procfs files we assess what maximum size they have under normal circumstances.

> What about openat support?

Will come if #321 gets accepted.

Compat.c
```diff
@@ -117,3 +117,68 @@ int Compat_readlinkat(int dirfd,
...
+         buffer = ((char*)buffer) + res;
+         count -= (size_t)res;
+         alreadyRead += res;
+      }
+
+      if (count == 0 || res == 0) {
+         close(fd);
+         return alreadyRead;
+      }
+   }
```
BenBE, line 147: Extract to `xReadFD` for reuse in `xReadfileat` …

linux/LinuxProcessList.c
```diff
@@ -108,28 +85,14 @@ static int sortTtyDrivers(const void* va, const void* vb) {
...
-   }
-   if (bufLen == 0) {
-      free(buf);
+   char buf[16384];
+   ssize_t r = xReadfile(PROCTTYDRIVERSFILE, buf, sizeof(buf) - 1);
+   if (r < 0)
       return;
-   }
+
+   buf[r] = '\0';
```
BenBE, line 94: What about making the NUL termination part of the xReadfile API? I.e. internally ensured before terminating that the byte following the last index written in buffer is forced to `\0`?

XUtils.c
```diff
@@ -210,3 +211,46 @@ char* xStrndup(const char* str, size_t len) {
...
+
+      if (res > 0) {
+         buffer = ((char*)buffer) + res;
+         count -= (size_t)res;
+         alreadyRead += res;
+      }
+
+      if (count == 0 || res == 0) {
+         close(fd);
+         ((char*)buffer)[alreadyRead] = '\0';
```
BenBE, line 236: Avoid buffer overflow:
```suggestion
         if (count)
            ((char*)buffer)[alreadyRead] = '\0';
```

BenBE (changes requested):

XUtils.c
cgzones: there is previously the line `count--; // reserve one for null-terminator`.

Or should we check
```c
ssize_t xReadfile(const char* pathname, void* buffer, size_t count) {
   if (count == 0)
      return -EINVAL;
   ...
```
?

BenBE: Unconditionally decrementing causes `MAX_INT` to be read, which is likely undesired as well. But decrementing the `count` on entry is quite a good solution. So return `-EINVAL` on `!count` before decrementing it. Didn't notice the decrement when I initially read through the source.

cgzones:
Got a crash:

```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==34090==ERROR: AddressSanitizer: SEGV on unknown address 0x03e80000852a (pc 0x7ff3dd3bcfe1 bp 0x7ffe38245970 sp 0x7ffe38244f40 T0)
==34090==The signal is caused by a READ memory access.
    #0 0x7ff3dd3bcfe1 in raise nptl/../sysdeps/unix/sysv/linux/raise.c:51:1
    #1 0x4ed427 in CRT_handleSIGSEGV /home/christian/Coding/workspaces/htop/CRT.c:894:4
    #2 0x7ff3dd3bd13f  (/lib/x86_64-linux-gnu/libpthread.so.0+0x1413f)
    #3 0x7ff3dd254082  string/../sysdeps/x86_64/multiarch/../strchr.S:31
    #4 0x43cab8 in strchr (/home/christian/Coding/workspaces/htop/htop+0x43cab8)
    #5 0x57632b in LinuxProcessList_readStatFile /home/christian/Coding/workspaces/htop/linux/LinuxProcessList.c:351:18
    #6 0x56f201 in LinuxProcessList_recurseProcTree /home/christian/Coding/workspaces/htop/linux/LinuxProcessList.c:1330:13
    #7 0x567c6a in ProcessList_goThroughEntries /home/christian/Coding/workspaces/htop/linux/LinuxProcessList.c:1905:4
    #8 0x535c05 in ProcessList_scan /home/christian/Coding/workspaces/htop/ProcessList.c:346:4
    #9 0x53e803 in checkRecalculation /home/christian/Coding/workspaces/htop/ScreenManager.c:108:7
    #10 0x53bfc8 in ScreenManager_run /home/christian/Coding/workspaces/htop/ScreenManager.c:159:10
    #11 0x4fefe7 in main /home/christian/Coding/workspaces/htop/htop.c:336:4
    #12 0x7ff3dd1ddcc9 in __libc_start_main csu/../csu/libc-start.c:308:16
    #13 0x4278f9 in _start (/home/christian/Coding/workspaces/htop/htop+0x4278f9)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV nptl/../sysdeps/unix/sysv/linux/raise.c:51:1 in raise
==34090==ABORTING
```

Checking on master...

BenBE (approved):

PR #342, opened 2020-11-26
Drop unneeded parameters to the ScreenManager constructor

natoscott:
All calls to ScreenManager_new always pass the same first
five values, the orientation is always HORIZONTAL and the
y1 parameter is always the height of the passed-in header
struct pointer.  I think its safe to assert at this point
that no VERTICAL orientation will arrive (if it does, its
no harm in re-adding this then) - so we can remove unused
conditionals (and TODOs) based on orientation too.

### DIFF
--- a/Action.c
+++ b/Action.c
@@ -46,7 +46,7 @@ Object* Action_pickFromVector(State* st, Panel* list, int x, bool followProcess)
    Settings* settings = st->settings;
 
    int y = panel->y;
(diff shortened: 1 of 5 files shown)


fasterit:
Discussed in team meeting -> go for it

BenBE:
How does this relate to the signal selection / other list selections on the left of the screen?

natoscott:
> How does this relate to the signal selection / other list selections on the left of the screen?

Not at all - this simply removes dead code and unnecessary parameter passing.

Issue #344, opened 2020-11-26
pwd column would be useful for many server cases


beppe9000:
hishamhm/htop#288

fasterit:
Would be a useful info in #183.

cgzones:
For a cwd column see #345.

In the Environment-Screen (select a process and press e), there is an entry `pwd`.
I am not sure if it exists for all processes though.

PR #345, opened 2020-11-26
Add Linux cwd process column

cgzones:
Related: #344

### DIFF
--- a/linux/LinuxProcess.h
+++ b/linux/LinuxProcess.h
@@ -28,6 +28,7 @@ in the source distribution for its full text.
 #define PROCESS_FLAG_LINUX_CTXT     0x00004000
 #define PROCESS_FLAG_LINUX_SECATTR  0x00008000
 #define PROCESS_FLAG_LINUX_LRS_FIX  0x00010000
+#define PROCESS_FLAG_LINUX_CWD      0x00020000
 
 typedef enum UnsupportedProcessFields {
    FLAGS = 9,
@@ -94,7 +95,8 @@ typedef enum LinuxProcessFields {
    SECATTR = 123,
    PROC_COMM = 124,
    PROC_EXE = 125,
-   LAST_PROCESSFIELD = 126,
+   CWD = 126,
+   LAST_PROCESSFIELD = 127,
 } LinuxProcessField;
 
 /* LinuxProcessMergedCommand is populated by LinuxProcess_makeCommandStr: It
@@ -181,6 +183,7 @@ typedef struct LinuxProcess_ {
    unsigned long ctxt_diff;
    char* secattr;
    unsigned long long int last_mlrs_calctime;
+   char* cwd;
 } LinuxProcess;
 
 #define Process_isKernelThread(_process) (((const LinuxProcess*)(_process))->isKernelThread)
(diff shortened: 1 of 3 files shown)


linux/LinuxProcess.h
```diff
@@ -19,14 +19,15 @@ in the source distribution for its full text.
...
-#define PROCESS_FLAG_LINUX_SECATTR  0x8000
+#define PROCESS_FLAG_LINUX_IOPRIO   0x00100
+#define PROCESS_FLAG_LINUX_OPENVZ   0x00200
+#define PROCESS_FLAG_LINUX_VSERVER  0x00400
+#define PROCESS_FLAG_LINUX_CGROUP   0x00800
+#define PROCESS_FLAG_LINUX_OOM      0x01000
+#define PROCESS_FLAG_LINUX_SMAPS    0x02000
+#define PROCESS_FLAG_LINUX_CTXT     0x04000
+#define PROCESS_FLAG_LINUX_SECATTR  0x08000
+#define PROCESS_FLAG_LINUX_CWD      0x10000
```
BenBE, line 30: Cf. #238.

linux/LinuxProcessList.c
```diff
@@ -785,6 +785,27 @@ static void LinuxProcessList_readSecattrData(LinuxProcess* process, const char*
    process->secattr = xStrdup(buffer);
 }
 
+static void LinuxProcessList_readCwd(LinuxProcess* process, const char* dirname, const char* name) {
+   char filename[MAX_NAME + 1];
+   xSnprintf(filename, sizeof(filename), "%s/%s/cwd", dirname, name);
+
+   char pathBuffer[PATH_MAX + 1];
+   ssize_t r = readlink(filename, pathBuffer, sizeof(pathBuffer) - 1);
```
BenBE, line 793: Use `readlinkat` via #321.

BenBE (changes requested):
What's the reasoning for renaming `flags` to `processFlags` for this PR?

BenBE:
Checking if all processes in `/proc` have a `cwd` symlink:
```bash
bash -c 'for a in /proc/*/; do [ -e $a/cwd ] || echo $a; done' # as root
```
Returns only non-PID directories for me.

IF you can read the symlink you should always get a non-empty string back. Thus `NULL` might act as a `no perm` marker. Haven't checked for processes running in different FS namespaces though.

BenBE:
Three issues I noticed while testing:

1. Overflows if path is too long (causes mis-alignment).
2. Regularly just shows `/proc/<pid>/fdinfo (deleted)` as cwd
3. `N/A` should be shadowed.

cgzones:
> * Overflows if path is too long (causes mis-alignment).

ideas?

> * Regularly just shows `/proc/<pid>/fdinfo (deleted)` as cwd

resolved

> * `N/A` should be shadowed.

resolved

BenBE:
> > * Overflows if path is too long (causes mis-alignment).
> 
> ideas?
> 
What about `xSnprintf(buffer, n, "%-25.25s", lp->cwd);` ???

BenBE (approved):

PR #349, opened 2020-11-27
Install app icon in hicolor icon theme

pinotree:
Switch the install location of the application icon from the legacy pixmaps location to the global hicolor icon theme (following the size of the icon).

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -7,8 +7,8 @@ EXTRA_DIST = $(dist_man_MANS) htop.desktop htop.png \
 install-sh autogen.sh missing
 applicationsdir = $(datadir)/applications
 applications_DATA = htop.desktop
-pixmapdir = $(datadir)/pixmaps
-pixmap_DATA = htop.png
+icon128dir = $(datadir)/icons/hicolor/128x128/apps
+icon128_DATA = htop.png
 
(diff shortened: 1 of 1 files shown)


BenBE:
Wouldn't it be best to provide icons for different sizes too?

pinotree:
> Wouldn't it be best to provide icons for different sizes too?

Possibly, although that can be done later on IMHO (and it will require creating a directory structure for the application icons).

fasterit:
@pinotree What is the benefit of moving the icon to that different directory?

pinotree:
The `pixmaps` directory is considered a legacy location for icons, and historically used mostly for XPM icons.

> it looks like pixmaps is better as DEs can then override the icon with their choice of an icon "theme".

That happens already, in case an icon theme provides an "htop" icon. So there is no difference from this POV.

If an icon theme provides its own version of some application icon, usually it is to provide one fitting with the icon theme. An example is a monochrome icon theme. If the icon theme version of an application icon "destroys the brand", I'd say it is a bug in the icon theme.

fasterit:
> `pixmaps` [..] considered a legacy location

[citation needed] i.e. where does it say that? I tried to find something like that in the "desktop standards" (i.e. the docs from freedesktop.org) but failed to do so.

fasterit:
We provide an .svg icon since 8029e9af04e6cd812383c392b791e6c4fd50b26d, that should work for common Linux DEs. Closing.

PR #353, opened 2020-11-29
Linux: avoid float division by 0 after system sleep

cgzones:
linux/LinuxProcessList.c:1403:63: runtime error: division by zero
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior linux/LinuxProcessList.c:1403:63 in

### DIFF
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -1362,8 +1362,9 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
          LinuxProcess_updateIOPriority(lp);
       }
 
-      float percent_cpu = (lp->utime + lp->stime - lasttimes) / period * 100.0;
-      proc->percent_cpu = isnan(percent_cpu) ? 0.0 : CLAMP(percent_cpu, 0.0, cpus * 100.0);
+      /* period might be 0 after system sleep */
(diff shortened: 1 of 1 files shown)


XUtils.h
```diff
@@ -63,4 +63,11 @@ char* xStrdup(const char* str) ATTR_NONNULL;
 
 char* xStrndup(const char* str, size_t len) ATTR_NONNULL;
 
+static inline bool double_equal(double a, double b) {
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wfloat-equal"
+   return a == b;
+#pragma GCC diagnostic pop
```
BenBE, line 70:
```suggestion
   return fabs(a-b) < 1E-6;
```

BenBE (changes requested):

XUtils.h
cgzones: I really want to compare to literal zero (cause the value is used as divisor), so the hack just to silence the compiler.

BenBE: But given the context where this is used, it doesn't really make sense to allow for other arbitrarily small intervals to pass through either. As formulated with this suggestion you could resolve times as low as 1µs, which already is far below the time we need for one cycle through our code. Filtering small differences also avoids "exploding numbers" issues down the way.

BenBE (approved):

Issue #350, opened 2020-11-28
Pressure stall info counters labeling is indistinctive


kkm000:
The new Pressure stall counters look valuable, thanks for adding them. They have display labeling issues, however:
1. Their names are indistinguishable in the Left/Right columns menus in the configuration screen: all 5 of them read as `Pressure Stall Infor`.
2. The labels are shortened to `Som` and `Ful` on the main screen when readouts are set to bar mode.
```
  Avg[||                                                           1.3%]   
  Mem[||||||||||||||||||||||||||||||||||||||||||||||||||||||10.2G/11.7G]   
  Som[|||                                        xxxx 0.43% 0.50% 0.46%]   
  Ful[                                           xxxx 0.00% 0.00% 0.00%]   
  Ful[                                           xxxx 0.00% 0.00% 0.00%]

Setup            Left column          Right column         Available meters
Meters           CPU [Bar]            Task counter [Text]  Clock
Display options  Memory [Bar]         Load average [Text]  Date
Colors           Pressure Stall Infor Uptime [Text]        Date and Time
Columns          Pressure Stall Infor Disk IO [Text]       Load averages: 1
                 Pressure Stall Infor                      Load: average of
                                                           Memory
```
(2) is not an issue in text readout mode:
```
  Avg[||                                                   
  Mem[|||||||||||||||||||||||||||||||||||||||||||||||||||||
  Some CPU pressure: 0.44% 0.49% 0.49%                     
  Full IO  pressure: 0.00% 0.00% 0.00%                     
  Full Mem pressure: 0.00% 0.00% 0.00%

Setup            Left column          Right column         
Meters           CPU [Bar]            Task counter [Text]  
Display options  Memory [Bar]         Load average [Text]  
Colors           Pressure Stall Infor Uptime [Text]        
Columns          Pressure Stall Infor Disk IO [Text]       
                 Pressure Stall Infor                      
```
I kinda used to the labeling of CPU load as `Avg`, but having bars labeled `Som`, `Som` and, you knew it was coming, `Som` is a different degree of inconvenience. :)

---
htop 3.0.3-dev built from the current tip of the master branch, 19b5141.
Configured as `./configure CC=clang-8 CFLAGS='-Oz'` on a Debian 10 machine with a bpo kernel 5.8.
```
$ uname -srvmo
Linux 5.8.0-0.bpo.2-amd64 #1 SMP Debian 5.8.10-1~bpo10+1 (2020-09-26) x86_64 GNU/Linux
```

[...]

PR #354, opened 2020-11-29
PressureStallMeter: improve display strings

cgzones:
Related: #350


BenBE:
I needed to actually read the docs to see the difference between the some vs full variants of this. Would be great, if the description could hint on this (some=partial stall on resource, full=all processes stalled on resources).

Could we keep CPU and IO uppercased (as abbreviations)?

PR #357, opened 2020-12-01
Dynamically load libsensors at runtime

cgzones:
Suggested by @eworm-de

### DIFF
--- a/linux/LibSensors.c
+++ b/linux/LibSensors.c
@@ -1,0 +1,105 @@
+#include "LibSensors.h"
+
+#ifdef HAVE_SENSORS_SENSORS_H
+
+#include <dlfcn.h>
+#include <errno.h>
+#include <limits.h>
+#include <sensors/sensors.h>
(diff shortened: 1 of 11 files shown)


eworm-de:
Oh, that was fast. 😊
About to test this...

eworm-de:
Works for me.

However I would not mind having libsensors and its development files as a build dependency. That should save some extra code.

linux/LibSensors.c
```diff
@@ -0,0 +1,105 @@
...
+dlfailure:
+   if (dlopenHandle) {
+      dlclose(dlopenHandle);
+      dlopenHandle = NULL;
+   }
+   return -1;
+}
+
+void LibSensors_cleanup(void) {
+    if (dlopenHandle) {
```
BenBE, line 60: Indentation

cgzones: thanks, fixed

PR #360, opened 2020-12-02
Set locale only once and do not override it later

cgzones:


CRT.c
```diff
@@ -725,9 +725,6 @@ void CRT_init(const int* delay, int colorScheme, bool allowUnicode) {
       CRT_colorScheme = COLORSCHEME_MONOCHROME;
    CRT_setColors(CRT_colorScheme);
 
-   /* initialize locale */
```
BenBE, line 728: Move comment to remaining instance of locale management code … ;-)

BenBE (approved):

PR #362, opened 2020-12-03
GitHub CI updates

cgzones:


BenBE (changes requested):
NB: [to plenk](https://en.wikipedia.org/wiki/Plenken)

BenBE (approved):

PR #367, opened 2020-12-05
OpenBSD update

cgzones:
- compilation failures like `return &this->this;` -> `return &this->super;`
- iwyu update
- misc cleanup

### DIFF
--- a/openbsd/OpenBSDProcess.c
+++ b/openbsd/OpenBSDProcess.c
@@ -6,26 +6,15 @@ Released under the GNU GPLv2, see the COPYING file
 in the source distribution for its full text.
 */
 
-#include "Process.h"
-#include "ProcessList.h"
 #include "OpenBSDProcess.h"
-#include "Platform.h"
-#include "CRT.h"
 
 #include <stdlib.h>
-#include <unistd.h>
-#include <sys/syscall.h>
 
+#include "CRT.h"
+#include "Process.h"
+#include "RichString.h"
+#include "XUtils.h"
 
-const ProcessClass OpenBSDProcess_class = {
-   .super = {
-      .extends = Class(Process),
-      .display = Process_display,
-      .delete = Process_delete,
-      .compare = OpenBSDProcess_compare
-   },
-   .writeField = OpenBSDProcess_writeField,
-};
 
 ProcessFieldData Process_fields[] = {
    [0] = {
@@ -200,7 +189,7 @@ Process* OpenBSDProcess_new(const Settings* settings) {
    OpenBSDProcess* this = xCalloc(sizeof(OpenBSDProcess), 1);
    Object_setClass(this, Class(OpenBSDProcess));
    Process_init(&this->super, settings);
-   return &this->this;
+   return &this->super;
 }
 
 void Process_delete(Object* cast) {
@@ -209,8 +198,8 @@ void Process_delete(Object* cast) {
    free(this);
 }
 
-void OpenBSDProcess_writeField(const Process* this, RichString* str, ProcessField field) {
-   //OpenBSDProcess* fp = (OpenBSDProcess*) this;
+static void OpenBSDProcess_writeField(const Process* this, RichString* str, ProcessField field) {
+   //const OpenBSDProcess* op = (const OpenBSDProcess*) this;
    char buffer[256]; buffer[255] = '\0';
    int attr = CRT_colors[DEFAULT_COLOR];
    //int n = sizeof(buffer) - 1;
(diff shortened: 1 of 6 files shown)


openbsd/OpenBSDProcess.c
```diff
@@ -223,17 +212,17 @@ void OpenBSDProcess_writeField(const Process* this, RichString* str, ProcessFiel
...
 
-   if (settings->direction == 1) {
+   /*if (settings->direction == 1) {
       p1 = (const OpenBSDProcess*)v1;
       p2 = (const OpenBSDProcess*)v2;
    } else {
       p2 = (const OpenBSDProcess*)v1;
       p1 = (const OpenBSDProcess*)v2;
-   }
+   }*/
```
BenBE, line 225: Better mark the variables as unused, than commenting out code

```c
   if (settings->direction == 1) {
      p1 = (const OpenBSDProcess*)v1;
      p2 = (const OpenBSDProcess*)v2;
   } else {
      p2 = (const OpenBSDProcess*)v1;
      p1 = (const OpenBSDProcess*)v2;
   }

   (void)p1;
   (void)p2;
```

openbsd/OpenBSDProcessList.c
```diff
@@ -176,30 +173,25 @@ static double getpcpu(const struct kinfo_proc* kp) {
    if (fscale == 0)
```
BenBE, line 173: Float equal comparison, use `fscale < 1E-6` instead.

openbsd/Platform.c
```diff
@@ -216,45 +211,10 @@ void Platform_setMemoryValues(Meter* this) {
    this->values[2] = cachedMem;
 }
 
-/*
- * Copyright (c) 1994 Thorsten Lockert <tholo@sigmasoft.com>
- * All rights reserved.
- *
- * Taken almost directly from OpenBSD's top(1)
- */
```
BenBE, line 224: How's the copyright situation of this? `All rights reserved.` usually doesn't imply and license grant to use the code?

fasterit: That seems to be a BSD-license import from https://github.com/openbsd/src/blob/master/usr.bin/top/machine.c#L818

I suggest to keep this header but amend:

```
* Originally released under a BSD-3 license
* Modified through htop developers applying GPL-2
```

The code has been significantly modified through htop developers, so it is licensed GPL-2 now. But it does (in my non-professional assessment) not constitute a separate work as the original structure and logic is retained. So we should keep the original copyright claim.
Read: Somebody would likely be free to copy these pieces of code again and claim their use as fair under BSD-3.

BenBE: Okay, than this updated copyright notice should go to the new location of this piece of code, as this is just the location where the code was removed from.

openbsd/OpenBSDProcessList.c
cgzones: `fscale` is of type `long`

openbsd/Platform.c
cgzones: The *new* code is in OpenBSDProcessList.c:
```c
   int nswap = swapctl(SWAP_NSWAP, 0, 0);
   if (nswap > 0) {
      struct swapent swdev[nswap];
      int rnswap = swapctl(SWAP_STATS, swdev, nswap);

      /* Total things up */
      unsigned long long int total = 0, used = 0;
      for (int i = 0; i < rnswap; i++) {
         if (swdev[i].se_flags & SWF_ENABLE) {
            used += (swdev[i].se_inuse / (1024 / DEV_BSIZE));
            total += (swdev[i].se_nblks / (1024 / DEV_BSIZE));
         }
      }

      pl->totalSwap = total;
      pl->usedSwap = used;
   } else {
      pl->totalSwap = pl->usedSwap = 0;
   }
```

In my opinion (IANAL) one can not claim copyright on trivial code or on code than can be almost only written in one way.
How would one else write code for OpenBSD to get swap statistics?  see https://man.openbsd.org/man2/swapctl.2

BenBE: What's so bad about just adding in the two lines suggested by @fasterit to give attribution of where that source originally came from?

I know of a case (in the US) where someone got some DMCA over 5 assembly instructions for swapping two registers, because they coincided with the source code of the original manufacturer and being placed the exact same location as programs where expecting this function at that location …

So: Doesn't hurt to include some attribution here, even if the original code has been heavily modified by us already.

cgzones: Added copyright note

BenBE (approved):

PR #373, opened 2020-12-08
Support clock_gettime() on OSX El Capitan and earlier

cgzones:
Reported-by: @nunobett

### DIFF
--- a/Compat.c
+++ b/Compat.c
@@ -11,11 +11,17 @@ in the source distribution for its full text.
 
 #include <errno.h>
 #include <fcntl.h> // IWYU pragma: keep
+#include <time.h>
 #include <unistd.h>
 #include <sys/stat.h>
 #include <sys/types.h> // IWYU pragma: keep
 
 #include "XUtils.h" // IWYU pragma: keep
+
+#ifdef HAVE_HOST_GET_CLOCK_SERVICE
+#include <mach/clock.h>
+#include <mach/mach.h>
+#endif
 
 
 int Compat_faccessat(int dirfd,
@@ -116,4 +122,32 @@ ssize_t Compat_readlinkat(int dirfd,
    return readlink(path, buf, bufsize);
 
 #endif
+}
+
+int Compat_clock_monotonic_gettime(struct timespec *tp) {
+
+#if defined(HAVE_CLOCK_GETTIME)
+
+   return clock_gettime(CLOCK_MONOTONIC, tp);
+
+#elif defined(HAVE_HOST_GET_CLOCK_SERVICE)
+
+   clock_serv_t cclock;
+   mach_timespec_t mts;
+
+   host_get_clock_service(mach_host_self(), SYSTEM_CLOCK, &cclock);
+   clock_get_time(cclock, &mts);
+   mach_port_deallocate(mach_task_self(), cclock);
+
+   tp->tv_sec = mts.tv_sec;
+   tp->tv_nsec = mts.tv_nsec;
+
+   return 0;
+
+#else
+
+#error No Compat_clock_monotonic_gettime() implementation!
+
+#endif
+
 }
--- a/ProcessList.c
+++ b/ProcessList.c
@@ -539,7 +539,9 @@ void ProcessList_scan(ProcessList* this, bool pauseProcessUpdate) {
       this->scanTs = 0;
       firstScanDone = true;
    } else if (Compat_clock_monotonic_gettime(&now) == 0) {
-      this->scanTs = now.tv_sec + now.tv_nsec / 1000000000;
+      // save time in millisecond, so with a delay in deciseconds
+      // there are no irregularities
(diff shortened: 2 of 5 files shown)


nunobett:
I was reading more about this issue and I don't think this per se will suffice, because no clock_gettime() implementation existed before osx 10.12.

Check out https://github.com/zeromq/libzmq/issues/2175.

ProcessList.c
```diff
@@ -541,8 +539,8 @@ void ProcessList_scan(ProcessList* this, bool pauseProcessUpdate) {
    if (!firstScanDone) {
       this->scanTs = 0;
       firstScanDone = true;
-   } else if (clock_gettime(CLOCK_MONOTONIC, &now) == 0) {
-      this->scanTs = now.tv_sec;
+   } else {
+      this->scanTs = time(NULL);
```
BenBE, line 543: `time(2)` is not monotonic …

BenBE (commented):
LGTM. What about the `CLOCK_MONOTONIC` constant of `clock_gettime`? I.e. when the function is supported, but that particular constant is not? Or is it guaranteed that that constant is available if the function is?

adsr (approved):

Compat.c
```diff
@@ -117,3 +123,31 @@ ssize_t Compat_readlinkat(int dirfd,
 
 #endif
 }
+
+int Compat_clock_monotonic_gettime(struct timespec *tp) {
+
+#ifdef HAVE_CLOCK_GETTIME
```
BenBE, line 129: What about the following for consistency?
```suggestion
#if defined(HAVE_CLOCK_GETTIME)
```

ProcessList.c
```diff
@@ -538,8 +538,10 @@ void ProcessList_scan(ProcessList* this, bool pauseProcessUpdate) {
    if (!firstScanDone) {
       this->scanTs = 0;
       firstScanDone = true;
-   } else if (clock_gettime(CLOCK_MONOTONIC, &now) == 0) {
-      this->scanTs = now.tv_sec;
+   } else if (Compat_clock_monotonic_gettime(&now) == 0) {
+      // save time in hundredth of a second, so with a delay in deciseconds
+      // there are no irregularities
+      this->scanTs = 100 * now.tv_sec + now.tv_nsec / 10000000;
```
BenBE, line 544: Why not use milliseconds?

BenBE (approved):

lutzmad:
Hello,
based on htop 3.1.1 I add the follwoing code to get htop up and running with older MacOS with no monotonic clock available.
```
--- generic/gettime.c.orig	2021-10-14 23:42:05.000000000 +0200
+++ generic/gettime.c	2021-11-18 21:25:18.000000000 +0100
@@ -51,7 +51,14 @@

 #else

-# error "No monotonic clock available"
+// # error "No monotonic clock available"
+#warning "No monotonic clock available, fall back to gettimeofday"
+
+   struct timeval tv;
+   if (gettimeofday(&tv, NULL) == 0)
+      *msec = ((uint64_t)tv.tv_sec * 1000) + ((uint64_t)tv.tv_usec / 1000);
+   else
+      *msec = 0;

 #endif
 }
```
Feel free to add the code to the master if the fix seems to be useful for you,
on the other hand, I can send a pull request also.

With regards,
Lutz

BenBE:
> Hello, based on htop 3.1.1 I add the following code to get htop up and running with older MacOS with no monotonic clock available.
> 
> […]

The patch itself looks fine, except for a few things:

1. Leaving old code commented out in the code base is usually bad style. Thus IMHO it's better to fully remove that commented out line.
2. AFAIK `#warning` is [not portable](https://stackoverflow.com/a/279669) and thus should not be used in code that is expected to be used on several platforms.
3. As the `generic` folder is intended as a baseline for code common to multiple platforms all branches should be runnable on all platforms that use this code. While this the case with `gettimeofday` I'm not quite comfortable with that addition there. This is probably open for debate.

> Feel free to add the code to the master if the fix seems to be useful for you, on the other hand, I can send a pull request also.
> 
> With regards, Lutz

PR appreciated for easier handling.

natoscott:
It may result in cleaner code to add a runtime fall back to CLOCK_REALTIME using the (existing) clock_gettime API rather than gettimeofday (or perhaps in addition to?).  According to the clock_gettime man page all implementations support CLOCK_REALTIME.

Agree with @BenBE that it'd be best to get rid of \#warning (and \#error too) if we can.

lutzmad:
Hello BenBE, natoscott,  
thanks for your hints, I will cleanup my code changes and add a pull request for easier handling.

@BenBE, @natoscott,  
you are right, the #warning and #error messages are ugly, but helpful on the other hand also. I do not know a more easy way to add some hints, additional information, to the code.

I will be back with a pull request as soon as possiible,
with regards,
Lutz

PR #378, opened 2020-12-09
Fix pause mode ("Z") in tree view

fasterit:


ScreenManager.c
```diff
@@ -106,7 +106,7 @@ static void checkRecalculation(ScreenManager* this, double* oldTime, int* sortTi
    if (*rescan) {
       *oldTime = newTime;
       ProcessList_scan(pl, this->state->pauseProcessUpdate);
-      if (*sortTimeout == 0 || this->settings->treeView) {
+      if (!(this->state->pauseProcessUpdate) && (*sortTimeout == 0 || this->settings->treeView)) {
```
BenBE, line 109: No need for parens …

```suggestion
      if (!this->state->pauseProcessUpdate && (*sortTimeout == 0 || this->settings->treeView)) {
```

fasterit: ok, I didn't find a guidance on this in the styleguide. I actually checked :)

PR #379, opened 2020-12-10
Cull the definitions of pageSize and pageSizeKB from CRT.c

natoscott:
By storing the per-process m_resident and m_virt values in the form
htop wants to display them in (KB, not pages), we no longer need to
have definitions of pageSize and pageSizeKB in the common CRT code.

These variables were never really CRT (i.e. display) related in the
first place.  It turns out the darwin platform code doesn't need to
use these at all (the process values are extracted from the kernel
in bytes not pages) and the other platforms can each use their own
local pagesize variables, in more appropriate locations.

Some platforms were actually already doing this, so this change is
removing duplication of logic and variables there.

### DIFF
--- a/solaris/SolarisProcessList.c
+++ b/solaris/SolarisProcessList.c
@@ -25,8 +25,10 @@ in the source distribution for its full text.
 
 #include "CRT.h"
 
-
 #define MAXCMDLINE 255
+
+static int pageSize;
+static int pageSizeKB;
 
 char* SolarisProcessList_readZoneName(kstat_ctl_t* kd, SolarisProcess* sproc) {
    char* zname;
@@ -50,13 +52,18 @@ ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* pidMatchList, ui
 
    spl->kd = kstat_open();
 
-   pl->cpuCount = sysconf(_SC_NPROCESSORS_ONLN);
+   pageSize = sysconf(_SC_PAGESIZE);
+   if (pageSize == -1)
+      CRT_fatalError("Cannot get pagesize by sysconf(_SC_PAGESIZE)");
+   pageSizeKB = pageSize / 1024;
 
-   if (pl->cpuCount == 1 ) {
+   pl->cpuCount = sysconf(_SC_NPROCESSORS_ONLN);
+   if (pl->cpuCount == -1)
+      CRT_fatalError("Cannot get CPU count by sysconf(_SC_NPROCESSORS_ONLN)");
+   else if (pl->cpuCount == 1)
       spl->cpus = xRealloc(spl->cpus, sizeof(CPUData));
-   } else {
+   else
       spl->cpus = xRealloc(spl->cpus, (pl->cpuCount + 1) * sizeof(CPUData));
-   }
 
    return pl;
 }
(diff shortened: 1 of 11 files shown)


solaris/SolarisProcessList.c
```diff
@@ -50,13 +52,18 @@ ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* pidMatchList, ui
...
 
-   if (pl->cpuCount == 1 ) {
+   pl->cpuCount = sysconf(_SC_NPROCESSORS_ONLN);
+   if (pl->cpuCount == -1)
+      CRT_fatalError("Cannot get CPU count by sysconf(_SC_NPROCESSORS_ONLN)");
+   else if (pl->cpuCount == 1)
       spl->cpus = xRealloc(spl->cpus, sizeof(CPUData));
-   } else {
+   else
       spl->cpus = xRealloc(spl->cpus, (pl->cpuCount + 1) * sizeof(CPUData));
```
BenBE, line 66: Why special-case this allocation based on CPU count?

solaris/SolarisProcessList.c
```diff
@@ -50,13 +52,18 @@ ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* pidMatchList, ui
...
 
-   pl->cpuCount = sysconf(_SC_NPROCESSORS_ONLN);
+   pageSize = sysconf(_SC_PAGESIZE);
+   if (pageSize == -1)
+      CRT_fatalError("Cannot get pagesize by sysconf(_SC_PAGESIZE)");
+   pageSizeKB = pageSize / 1024;
 
-   if (pl->cpuCount == 1 ) {
+   pl->cpuCount = sysconf(_SC_NPROCESSORS_ONLN);
+   if (pl->cpuCount == -1)
```
BenBE, line 61: Why not check `pl->cpuCount < 1`? Zero CPUs ain't that helpful. ;-)

solaris/SolarisProcessList.c
natoscott: AIUI, the reason the code is this way is that when we have two or more CPUs on a system we always allocate an extra row in the array for the aggregated values (summed across all CPUs).

This behaviour is not changed here - I just added some missing error handling.

solaris/SolarisProcessList.c
natoscott: Heh, indeed.  It's a matter of following the docs exactly I guess (I'm just keeping the existing code convention going there) - the syscall is documented to return -1 on failure.  If zero was returned the syscall would be broken of course - by definition, at least one CPU is always online: the one currently executing htop. ;)  Worse, if we treated that as an "error" the call to CRT_fatalError uses errno which presumably would not have been set since zero is not an error (even though semantically it is an error of course).

solaris/SolarisProcessList.c
BenBE: I figured as much. Just wondered if this special-casing is really necessary or if we just say to allocate this extra entry regardless of number of CPUs …

natoscott: Yep, the special casing could certainly be removed and we'd waste that little bit of memory - *shrug*.  Since the code already caters for it, I'd be inclined to leave it as-is though.

Issue #368, opened 2020-12-06
CPU usage incorrect on macOS Big Sur with M1 chip


sfzhu93:
I wrote a non-stop loop in Python. On the right, `top` correctly shows the cpu usage as 100% but on the left `htop`
<img width="1434" alt="Screen Shot 2020-12-06 at 3 27 04 AM" src="https://user-images.githubusercontent.com/42506672/101275384-37b04e00-3773-11eb-8173-c62c62f46f71.png">
 shows aroung 2.4%.

BenBE:
The screenshots you provided do not provide enough information to exclude the possibility you that you are watching two different instances of Python. Especially as the runtime of the process doesn't match up in the slightest (1.9 seconds vs. 1 minute, 19 seconds). Please provide a better illustration of the issue to avoid the possibility that the issue is just related to oversight on your side.

sfzhu93:
@BenBE  Hello, I double checked that again. Every second htop refreshes, the time of the python command only increases about 0.03-0.04. I guess it is the CPU time, not the time span since the process being created. Because the CPU usage is incorrect, the CPU time is also incorrect.

sfzhu93:
In addition, this is the source code of my deadloop.py
```
while True:
    pass
```
and a more complete screenshot of it on htop
<img width="1370" alt="Screen Shot 2020-12-06 at 4 06 49 PM" src="https://user-images.githubusercontent.com/42506672/101292407-459ab900-37dd-11eb-9b88-c22e3d83e80f.png">

BenBE:
Thank you for reporting this issue.

Can you maybe run some checks with the current master (3.0.3)? Also, when compiling, try to enable debugging symbols and have a look at the data reported for the affected process in `darwin/DarwinProcessList.c` around line 204. The information of interest are the contents of `ps[i]` and `proc` for one process. Depending on your proficiency with `gdb` (tip: [pwndbg](https://github.com/pwndbg/pwndbg)) you can also try to insert some statements like `printf("%d    \r\n", value);` at that location, which will appear in the upper left corner; although a full, consistent dump of the two structures is preferred.

cfstras:
Still happens on master (ded9c5d36332401bb8d5e33b6f72de188a42560d):
The pid matches python, and any process with an infinite loop gets 2.4% CPU for me.

<img width="961" alt="image" src="https://user-images.githubusercontent.com/1156864/101607922-65acc100-3a05-11eb-849c-adea5a088213.png">

[...]

Issue #306, opened 2020-11-10
Restructure platform-dependent interface into central structure


BenBE:
At the moment the various platform implementations have quite different file structures and organizations where different functions are implemented. While some platforms implement certain features within their `platform.c` others do similar things in their `processlist.c` implementation. Yet other features are available on one platform, but not on another; but you wouldn't really know until you searched that exact platform for the place that feature is implemented or not.

My suggestion thus would be to transform the platform interface into an class structure that registers these platform specific implementations in a more or less platform-central place (like `platform.c` and provides `NULL` pointers for each feature not implemented on that platform. The core source then can translate such calls in a way that the error behaviour for platforms that do not implement a certain feature is uniform; as code duplication in that situation is reduced.

Reason for this ticket: In #239 I face exactly this issue: If moving the code to gether these information to each of the platforms I can isolate the actual lock listing quite well, but none-the-less it would produce code duplication for each platform, that does not support this feature (which at first will be the majority). Having a central platform structure wher I can simply register existing implementations where they exist, I can avoid duplication of the not-implemented case. It also would make much more obvious, which features each platform actually supports as these implementations would be linked together in a much more central place.

BenBE:
While investigating:
It seems like the contents of `Platform.h` for each platform should actually be identical. That being the case we should have one central `Platform.h` in the root and each platform only implements its versions of these functions in `Platform.c` (or any number of files that platform is split up into).

BenBE:
One option I noticed we have to reduce code duplication AND have a sane baseline could even be having the `unsupported` platform implementation living in `/Platform.c` with all functions marked weak. That way each platform just implements the functions it really supports and unsupported features don't cause any source to be written. Disadvantage: We'd need to have some logic in `configure.ac` to provide compiler-specific means to mark symbols as weak exports (when compiling with anything else but some GCC or clang).

natoscott:
@BenBE I like your idea here - I think this will end up with a neater code base.  Not sure about using weak symbols as it may not be easily portable to all platforms, but certainly there's other approaches that could achieve the same thing.  Perhaps a cpp-based HAVE macro kind of thing, or a set of function pointers with not-supported functions as defaults, and overridden at startup in each platforms initialisation ... we'll definitely find some approach I'm sure.

PR #381, opened 2020-12-11
Unhardcode tick-to-ms conversion

amomchilov:
Division by 100000.0  worked because `sysconf(_SC_CLK_TCK)` happened to be 100.

By unhardcoding:

1) It becomes more clear what this `10000.0` figure comes from.
2) It protects against bugs in the case `sysconf(_SC_CLK_TCK)` ever changes.

### DIFF
--- a/darwin/DarwinProcess.c
+++ b/darwin/DarwinProcess.c
@@ -301,23 +301,28 @@ void DarwinProcess_setFromKInfoProc(Process* proc, const struct kinfo_proc* ps, 
    proc->updated = true;
 }
 
+static double ticksToNanoseconds(const double ticks) {
+   const double nanos_per_sec = 1e9;
+   return ticks / (double) Platform_clockTicksPerSec * Platform_timebaseToNS * nanos_per_sec;
+}
+
 void DarwinProcess_setFromLibprocPidinfo(DarwinProcess* proc, DarwinProcessList* dpl) {
    struct proc_taskinfo pti;
 
    if (sizeof(pti) == proc_pidinfo(proc->super.pid, PROC_PIDTASKINFO, 0, &pti, sizeof(pti))) {
-      if (0 != proc->utime || 0 != proc->stime) {
-         uint64_t diff = (pti.pti_total_system - proc->stime)
-                       + (pti.pti_total_user - proc->utime);
+      double time_interval = ticksToNanoseconds(dpl->global_diff) / (double) dpl->super.cpuCount;
 
-         proc->super.percent_cpu = (double)diff * (double)dpl->super.cpuCount * Platform_timebaseToNS
-                                 / ((double)dpl->global_diff * 100000.0);
+      uint64_t total_existing_time = proc->stime + proc->utime;
+      uint64_t total_current_time = pti.pti_total_system + pti.pti_total_user;
 
-//       fprintf(stderr, "%f %llu %llu %llu %llu %llu\n", proc->super.percent_cpu,
-//               proc->stime, proc->utime, pti.pti_total_system, pti.pti_total_user, dpl->global_diff);
-//       exit(7);
+      if (total_existing_time && 1e-6 < time_interval) {
+         uint64_t total_time_diff = total_current_time - total_existing_time;
+         proc->super.percent_cpu = ((double)total_time_diff / time_interval) * 100.0;
+      } else {
+         proc->super.percent_cpu = 0.0;
       }
 
-      proc->super.time = (pti.pti_total_system + pti.pti_total_user) / 10000000;
+      proc->super.time = total_current_time / 10000000;
       proc->super.nlwp = pti.pti_threadnum;
       proc->super.m_virt = pti.pti_virtual_size / ONE_K;
       proc->super.m_resident = pti.pti_resident_size / ONE_K;
--- a/darwin/Platform.c
+++ b/darwin/Platform.c
@@ -10,16 +10,18 @@ in the source distribution for its full text.
 
 #include "Platform.h"
 
+#include <errno.h>
 #include <math.h>
 #include <stdlib.h>
-
-#include <CoreFoundation/CoreFoundation.h>
+#include <unistd.h>
 #include <CoreFoundation/CFString.h>
+#include <CoreFoundation/CoreFoundation.h>
 #include <IOKit/ps/IOPowerSources.h>
 #include <IOKit/ps/IOPSKeys.h>
 
 #include "ClockMeter.h"
 #include "CPUMeter.h"
+#include "CRT.h"
 #include "DarwinProcessList.h"
 #include "DateMeter.h"
 #include "DateTimeMeter.h"
@@ -114,6 +116,8 @@ int Platform_numberOfFields = LAST_PROCESSFIELD;
 
 double Platform_timebaseToNS = 1.0;
 
+long Platform_clockTicksPerSec = -1;
+
 void Platform_init(void) {
    // Check if we can determine the timebase used on this system.
    // If the API is unavailable assume we get our timebase in nanoseconds.
@@ -124,6 +128,14 @@ void Platform_init(void) {
 #else
    Platform_timebaseToNS = 1.0;
 #endif
+
+   // Determine the number of clock ticks per second
+   errno = 0;
+   Platform_clockTicksPerSec = sysconf(_SC_CLK_TCK);
+
+   if (errno || Platform_clockTicksPerSec < 1) {
+      CRT_fatalError("Unable to retrieve clock tick rate.\n");
+   }
 }
 
 void Platform_done(void) {
--- a/darwin/Platform.h
+++ b/darwin/Platform.h
@@ -27,6 +27,8 @@ extern int Platform_numberOfFields;
 
 extern double Platform_timebaseToNS;
 
+extern long Platform_clockTicksPerSec;
+
 extern const SignalItem Platform_signals[];
 
 extern const unsigned int Platform_numberOfSignals;


darwin/DarwinProcess.c
```diff
@@ -240,6 +240,12 @@ void DarwinProcess_setFromKInfoProc(Process* proc, const struct kinfo_proc* ps,
    proc->updated = true;
 }
 
+double ticks_to_nanoseconds(const double ticks) {
+   const double secs_per_tick = 1.0 / (double)sysconf(_SC_CLK_TCK);
```
natoscott, line 244: Probably best to cache the results of this sysconf call so that we're not recalculating it on every sample?  (perhaps a global 'hertz' var set from either the darwin Platform_init or ProcessList_new routine)

sysconf can also fail, so doing this call once on startup will make error checking / handling easier.

amomchilov: Agreed! Though I'm not a fan of `herz`, I'd rather something like `clock_ticks_per_s`. Thoughts?

I just feeling out if this PR is a move in the right direction.

I noticed that a *lot* of this code-base is unclear about units, exact derivations, etc., so I'm not sure if this kind of change would be desired or not.

natoscott: > Agreed! Though I'm not a fan of `herz`, I'd rather something like `clock_ticks_per_s`. Thoughts?

Sure, doesn't have to be hertz - if its multi-words like that though camel case is more commonly used in htop (see darwin/Platform.c for some examples).

> I just feeling out if this PR is a move in the right direction.

Looks like its moving in a less broken direction - so two thumbs up from me.

amomchilov: Ah yes, I didn't notice. Looks like local vars are snake case, and functions/types are camel case. I'll change my identifiers to match.

Some other questions:

1. Should I add an explicit but technically redundant `#include <unistd.h>`? It's what defines `sysconf`, but it's also already indirectly included (probably by CoreFoundation, but I haven't checked.

2. The return type of `sysconf` is `long`. Should the cache variable match that and be a `long`, or use `uint64_t` like the rest? Or should I cache it after conversion to `double`?

3. I saw that `sysconf(_SC_CLK_TCK)` is also called in `LinuxProcess_adjustTime`. They called the cache var `jiffy`. It appears that they use a lazy-initialization pattern, that makes `sysconf(_SC_CLK_TCK)` called upon the first call to `LinuxProcess_adjustTime`.

    They fail silently if the `errno` is set, by just assuming 100 Hz. That seems... like the wrong choice to me. Wouldn't it be better to fail loudly than to assume a potentially wrong value? I feel like I'm pulling a thread here haha.

natoscott: >     1. Should I add an explicit but technically redundant `#include <unistd.h>`? It's what defines `sysconf`, but it's also already indirectly included (probably by CoreFoundation, but I haven't checked.

I'd not add it explicitly unless its needed to compile - as you say, chances are its included indirectly already.

>     2. The return type of `sysconf` is `long`. Should the cache variable match that and be a `long`, or use `uint64_t` like the rest? Or should I cache it after conversion to `double`?

I recommend you do as much work up front as possible, at startup, and as little work as possible during each sample.
Thus, I think its best to store it as a double (like you had it in the original patch).

>     3. I saw that `sysconf(_SC_CLK_TCK)` is also called in `LinuxProcess_adjustTime`. They called the cache var `jiffy`. It appears that they use a lazy-initialization pattern, that makes `sysconf(_SC_CLK_TCK)` called upon the first call to `LinuxProcess_adjustTime`.

There's a good chance this code can be improved also.

>        They fail silently if the `errno` is set, by just assuming 100 Hz. That seems... like the wrong choice to me. Wouldn't it be better to fail loudly than to assume a potentially wrong value? I feel like I'm pulling a thread here haha.

Heh - yep.  It definitely would be better to fail htop at startup if this call fails - on Linux too, IMO.

BenBE:
Might help resolve #368 …

darwin/DarwinProcess.c
amomchilov: > I recommend you do as much work up front as possible, at startup, and as little work as possible during each sample.
Thus, I think its best to store it as a double (like you had it in the original patch).

I had a change of mind on this. Though `double` suits my usecase, some other feature might end up re-using it and converting it to `uint64_t`. It's just one `cvttsd2si` instruction anyway, so I was leaning towards keeping it as its original `long`.


>Heh - yep. It definitely would be better to fail htop at startup if this call fails - on Linux too, IMO.

Opened #384 to track this.

darwin/Platform.c
```diff
@@ -133,8 +136,17 @@ const MeterClass* const Platform_meterTypes[] = {
 
 int Platform_numberOfFields = 100;
 
+static void setClockTicksPerSec() {
+   errno = 0;
+   Platform_clockTicksPerSec = sysconf(_SC_CLK_TCK);
+
+   if (errno || -1 == Platform_clockTicksPerSec) {
```
BenBE, line 143: `Platform_clockTicksPerSec < 1` doesn't make much sense either; unless time runs backwards, in which case only `Platform_clockTicksPerSec == 0`would remain meaningless, but `Platform_clockTicksPerSec == -1` became ambiguous …

amomchilov: Are you suggesting that I initialize it to a value of `0`, and reserve `-1` for detecting errors?

BenBE: No, more like just checking for `< 1` and treat that as an error … ;-)

darwin/DarwinProcess.c
```diff
@@ -240,6 +241,12 @@ void DarwinProcess_setFromKInfoProc(Process* proc, const struct kinfo_proc* ps,
    proc->updated = true;
 }
 
+static double ticksToNanoseconds(const double ticks) {
+   const double secs_per_tick = 1.0 / (double) Platform_clockTicksPerSec;
+   const double nanos_per_sec = 1e9;
+   return ticks * secs_per_tick * nanos_per_sec;
```
BenBE, line 247: For better precision avoid multiplications with small numbers; instead prefer `ticks * nanos_per_sec / Platform_clockTicksPerSec`. Overall saves on one multiplication and thus reduces the round-off error introduced in these operations.

darwin/Platform.c
```diff
@@ -22,6 +22,7 @@ in the source distribution for its full text.
 #include "zfs/ZfsArcMeter.h"
 #include "zfs/ZfsCompressedArcMeter.h"
 #include "DarwinProcessList.h"
+#include "CRT.h"
```
BenBE, line 25: Header includes are sorted alphabetical, files in parent directories preceeding those from subdirectories.

Please refer to the styleguide for the exact order (seems this file has still been missed when this was updated everywhere).

darwin/Platform.h
```diff
@@ -31,6 +31,8 @@ extern const unsigned int Platform_numberOfSignals;
 
 extern const MeterClass* const Platform_meterTypes[];
 
+extern long Platform_clockTicksPerSec;
```
BenBE, line 34: Probably something for another PR, but if multiple platforms have this (Darwin + Linux, but basically *BSD too), this should most likely go into a global platform header (cf. #306) for some central interface for such universally needed information.

BenBE (changes requested):
Not sure if all the explicit `double` casts are technically required.

darwin/Platform.h
amomchilov: Agreed. See also: https://github.com/htop-dev/htop/issues/384

darwin/DarwinProcess.c
amomchilov: My $0.02, I'm more concerned about correctness/clarity than saving a few hundred multiplications.

Precision isn't much of a concern here, IMO:
1) `1.0 / 100.0 * 1e9` is losslessly representable/computable with `double`. It could even be done with `float` actually, and still wouldn't lose precision when `Platform_clockTicksPerSec == 100`
2) Large numbers are also an issue for precision. Doubles start truncating integer representations to even numbers after `9,007,199,254,740,993 (253 + 1).`. The computation of `ticks * nanos_per_sec` could become a pretty large value itself, which threatens precision. The issue would kick in after `9,007,199` ticks. At 100 ticks/s, that's 104 days of uptime with 1 CPU core, 52 days on a dual core, 26 days on a quad core, ... and so on. A maximally precise computation would be alternating multiplications and divisions in a way that stays in the more precise regions of `double`. So something like  `ticks / Platform_clockTicksPerSec * nanos_per_sec`
3) And more fundamentally, this computation is inescapably imprecise, since the process' times are updated non-atomically, separately from the elapsed CPU ticks. There will be a variable time between the measurements, that introduces inaccuracy.
4) None of this really matters anyway, since the result is only presented to 3 significant digits.

I don't mind changing it to whatever your suggest, but I would advise prioritizing clarity.

darwin/DarwinProcess.c
amomchilov: @natoscott NVM, the style guide suggests to add such headers: https://github.com/htop-dev/htop/blob/master/docs/styleguide.md#import-and-use-of-headers

darwin/DarwinProcess.c
BenBE: The suggestion from 2 seems fine with me.

FWIW: Neither `0.1` nor `0.01` are representable exactly in neither `float` nor `double`. Thus if there are simple ways to avoid predictable imprecision you oughta go them.

darwin/Platform.h
BenBE: This comment was mostly about this declaration itself, which is probably better suited to have in some platform-independent header and not with every platform.

darwin/DarwinProcess.c
```diff
@@ -6,6 +6,7 @@ in the source distribution for its full text.
 */
 
 #include "DarwinProcess.h"
+#include "Platform.h"
```
BenBE, line 9: Cf. styleguide on include order.

darwin/DarwinProcess.c
```diff
@@ -248,8 +255,9 @@ void DarwinProcess_setFromLibprocPidinfo(DarwinProcess* proc, DarwinProcessList*
          uint64_t diff = (pti.pti_total_system - proc->stime)
                        + (pti.pti_total_user - proc->utime);
 
-         proc->super.percent_cpu = (double)diff * (double)dpl->super.cpuCount
-                                 / ((double)dpl->global_diff * 100000.0);
+         double time_interval = ticksToNanoseconds(dpl->global_diff) / (double) dpl->super.cpuCount;
+
+         proc->super.percent_cpu = ((double)diff / time_interval) * 100.0;
```
BenBE, line 260: Should include some check for `time_interval` to not be too small. Epsilon check for `interval > 1e-6` should be fine.

darwin/Platform.c
```diff
@@ -6,28 +6,30 @@ Released under the GNU GPLv2, see the COPYING file
 in the source distribution for its full text.
 */
 
-#include "Platform.h"
```
BenBE, line 9: Include order and spacing after includes (cf. styleguide).

BenBE (changes requested):

darwin/DarwinProcess.c
amomchilov: What's the meaning if interval is so small? Two back-to-back updates were done in very short succession?

darwin/DarwinProcess.c
amomchilov: I think I got it right this time, let me know.

darwin/DarwinProcess.c
BenBE: For example. Another possibility is, when the notebook was in standby for a certain time. @cgzones fixed a similar issue for Linux recently, cf. #353 …

BenBE:
I rebased you PR onto #392 to include the changes from there.
Resolved the inevitable merge conflicts in the process.

BenBE (approved):

darwin/DarwinProcess.c
```diff
@@ -240,23 +241,28 @@ void DarwinProcess_setFromKInfoProc(Process* proc, const struct kinfo_proc* ps,
    proc->updated = true;
 }
 
+static double ticksToNanoseconds(const double ticks) {
+   const double nanos_per_sec = 1e9;
+   return ticks / (double) Platform_clockTicksPerSec * Platform_timebaseToNS * nanos_per_sec;
```
cgzones, line 246: `Platform_clockTicksPerSec * Platform_timebaseToNS * nanos_per_sec` is a runtime constant.
Maybe pre-compute it

darwin/DarwinProcess.c
```diff
@@ -240,23 +241,28 @@ void DarwinProcess_setFromKInfoProc(Process* proc, const struct kinfo_proc* ps,
...
+}
+
 void DarwinProcess_setFromLibprocPidinfo(DarwinProcess* proc, DarwinProcessList* dpl) {
    struct proc_taskinfo pti;
 
    if (sizeof(pti) == proc_pidinfo(proc->super.pid, PROC_PIDTASKINFO, 0, &pti, sizeof(pti))) {
-      if (0 != proc->utime || 0 != proc->stime) {
-         uint64_t diff = (pti.pti_total_system - proc->stime)
-                       + (pti.pti_total_user - proc->utime);
+      double time_interval = ticksToNanoseconds(dpl->global_diff) / (double) dpl->super.cpuCount;
```
cgzones, line 253: `double time_interval = ticksToNanoseconds(dpl->global_diff) / (double) dpl->super.cpuCount;` is constant for all processes of a single update cycle.
Maybe pre-compute it in `DarwinProcessList.c::ProcessList_goThroughEntries()`

darwin/DarwinProcess.c
BenBE: You mean something like?
```c
   static const double tickToNanosecondsScale = (double) Platform_clockTicksPerSec * Platform_timebaseToNS * 1e9;
   return ticks / tickToNanosecondsScale;
```

cgzones: I think this would not work, as `static` variables are initialized before `main` (and `Platform_clockTicksPerSec` and `Platform_timebaseToNS` are not yet set at this point).

Maybe it's OK to compute once every cycle when caching `time_interval` (see below).

BenBE: k, had to recheck on the actual initialization semantics here. Would take a second (`bool`) variable to act as lazy initialization guard to calculate this constant on first entry.

But as both values are set in `Platform_init` we can just initialize that scaling factor there.

darwin/DarwinProcess.c
BenBE: Then this should be passed in as parameter to `DarwinProcess_setFromLibprocPidinfo` to avoid depending on global state.

darwin/Platform.c
```diff
@@ -124,6 +128,14 @@ void Platform_init(void) {
 #else
    Platform_timebaseToNS = 1.0;
 #endif
+
+   // Determine the number of clock ticks per second
+   errno = 0;
+   Platform_clockTicksPerSec = sysconf(_SC_CLK_TCK);
+
+   if (errno || Platform_clockTicksPerSec < 1) {
```
natoscott, line 136: Don't set/test errno - just test for the return value being (exactly) -1 to detect an error here (as documented).

darwin/Platform.c
```diff
@@ -114,6 +116,8 @@ int Platform_numberOfFields = LAST_PROCESSFIELD;
 
 double Platform_timebaseToNS = 1.0;
 
+long Platform_clockTicksPerSec = -1;
```
natoscott, line 119: There's no need to initialize this to -1.

darwin/Platform.c
BenBE: My documentation on `sysconf` reads a bit different in this section:

> RETURN VALUE
> The return value of sysconf() is one of the following:
>
> *  On error, -1 is returned **and errno is set to indicate the cause of the error** (for example, EINVAL, indicating that name is invalid).
>
> *  If name corresponds to a maximum or minimum limit, and that limit is indeterminate, -1 is returned and errno is not changed. **(To distinguish an indeterminate limit from an error, set errno to zero before the call, and then check whether errno is nonzero when -1 is returned.)**
>
> *  If name corresponds to an option, a positive value is returned if the option is supported, and -1 is returned if the option is not supported.
>
> *  Otherwise, the current value of the option or limit is returned. This value will not be more restrictive than the corresponding value that was described to the application in <unistd.h> or <limits.h> when the application was compiled.

(Emphasis mine)

What would be of discussion would be handling the case that this setting was indeterminate is correct, but given you'd have no value their either, the current check should yield sensible behaviour in all cases.

darwin/Platform.c
BenBE: This value is used in case someone inadvertedly calls `DarwinProcess_ticksToNanoseconds` before `Platform_init` …

natoscott: Hmm.  Platform_init is called quite early during startup, there's no chance of that IMO - i.e. we haven't initialized the entire platform but we're already making per-process timing calculations?  (IMO *that* would be the bug to fix, not initializing the variable to -1 instead of the default zero).

But, not a big deal either way, -1 is OK too.

darwin/Platform.c
natoscott: | My documentation on sysconf [...]

Aha, I did not know that - thanks for the pointer.  Further complicating things, on Mac OS X this is the documentation:

https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/sysconf.3.html

I think its fine either way - personally, I'd stick with the simpler just-explicitly-test-minus-one like we do pretty much everywhere else, but I'm OK with the code as is here too.

amomchilov: I'll do whatever you guys settle on. Trying to implement this little change (which involved things I understood) is bringing me to the edge of my knowledge/expertise :P

BenBE: As both on MacOS and Linux they agree on set `errno` on error and the check as given above is quite broad, I'd suggest to stick with this version. It both catches out-of-range values AND when the function reports any error; and does so consistently even when the sysconf implementation is inconsistent in this regard.

natoscott: @BenBE +1

darwin/Platform.c
BenBE: Agreed, calling in that weird order would definitively a bug worth fixing.
But so is potentially causing a div-by-zero … ;-)
Let's go with it as is for now.

cgzones:
Applied via 67ccd6b909d28ab84c77acecdfee927337489cc2

Thanks for the contribution!

PR #385, opened 2020-12-11
Setting to alternate used memory counting for Linux

### LINKED ISSUE #281: Use MemTotal: - MemAvailable: for usedMem calculation on Linux 3.14+
htop currently parses /proc/meminfo and uses the following formula to calculate used memory under Linux:

`usedMem = ((MemTotal:) + (MemFree:)) - ((Buffers:) + ((Cached:) + (SReclaimable:) - (Shmem:))`

However, this does not properly display actual used memory and there is a much simpler and accurate way to do it.

Since Linux 3.14 was released in March 2014 the best way to report this number is to just do the following instead:

`usedMem = ((MemTotal:) - (MemAvailable:))`

[/proc/meminfo: provide estimated available memory git commit for Linux 3.14-rc1](https://github.com/torvalds/linux/commit/34e431b0ae398fc54ea69ff85ec700722c9da773)

> 
> **/proc/meminfo: provide estimated available memory** 
> Many load balancing and workload placing programs check /proc/meminfo to
> estimate how much free memory is available.  They generally do this by
> adding up "free" and "cached", which was fine ten years ago, but is
> pretty much guaranteed to be wrong today.
> 
> It is wrong because Cached includes memory that is not freeable as page
> cache, for example shared memory segments, tmpfs, and ramfs, and it does
> not include reclaimable slab memory, which can take up a large fraction
> of system memory on mostly idle systems with lots of files.
> 
> Currently, the amount of memory that is available for a new workload,
> without pushing the system into swap, can be estimated from MemFree,
> Active(file), Inactive(file), and SReclaimable, as well as the "low"
> watermarks from /proc/zoneinfo.
> 
> However, this may change in the future, and user space really should not
> be expected to know kernel internals to come up with an estimate for the
> amount of free memory.
> 
> It is more convenient to provide such an estimate in /proc/meminfo.  If
> things change in the future, we only have to change it in one place.
> 
> Signed-off-by: Rik van Riel <riel@redhat.com>
> Reported-by: Erik Mouw <erik.mouw_2@nxp.com>
> Acked-by: Johannes Weiner <hannes@cmpxchg.org>
> Signed-off-by: Andrew Morton <akpm@linux-foundation.org>
> Signed-off-by: Linus Torvalds <torvalds@linux-foundation.org>

tWido:
Closes #281

### DIFF
--- a/DisplayOptionsPanel.c
+++ b/DisplayOptionsPanel.c
@@ -119,6 +119,9 @@ DisplayOptionsPanel* DisplayOptionsPanel_new(Settings* settings, ScreenManager* 
    Panel_add(super, (Object*) CheckItem_newByRef("Also show CPU temperature (requires libsensors)", &(settings->showCPUTemperature)));
    Panel_add(super, (Object*) CheckItem_newByRef("- Show temperature in degree Fahrenheit instead of Celsius", &(settings->degreeFahrenheit)));
    #endif
+   #ifdef HTOP_LINUX
+   Panel_add(super, (Object*) CheckItem_newByRef("Use simpler formula to count used memory", &(settings->simpleUsedMemForm)));
+   #endif
    Panel_add(super, (Object*) CheckItem_newByRef("Enable the mouse", &(settings->enableMouse)));
    Panel_add(super, (Object*) NumberItem_newByRef("Update interval (in seconds)", &(settings->delay), -1, 1, 255));
    Panel_add(super, (Object*) CheckItem_newByRef("Highlight new and old processes", &(settings->highlightChanges)));
--- a/Settings.c
+++ b/Settings.c
@@ -193,6 +193,10 @@ static bool Settings_read(Settings* this, const char* fileName, int initialCpuCo
          this->updateProcessNames = atoi(option[1]);
       } else if (String_eq(option[0], "account_guest_in_cpu_meter")) {
          this->accountGuestInCPUMeter = atoi(option[1]);
+      #ifdef HTOP_LINUX
+      } else if (String_eq(option[0], "simple_used_mem_form")) {
+         this->simpleUsedMemForm = atoi(option[1]);
+      #endif
       } else if (String_eq(option[0], "delay")) {
          this->delay = CLAMP(atoi(option[1]), 1, 255);
       } else if (String_eq(option[0], "color_scheme")) {
@@ -298,6 +302,9 @@ bool Settings_write(Settings* this) {
    #endif
    fprintf(fd, "update_process_names=%d\n", (int) this->updateProcessNames);
    fprintf(fd, "account_guest_in_cpu_meter=%d\n", (int) this->accountGuestInCPUMeter);
+   #ifdef HTOP_LINUX
+   fprintf(fd, "simple_used_mem_form=%d\n", (int) this->simpleUsedMemForm);
+   #endif
    fprintf(fd, "color_scheme=%d\n", (int) this->colorScheme);
    fprintf(fd, "enable_mouse=%d\n", (int) this->enableMouse);
    fprintf(fd, "delay=%d\n", (int) this->delay);
--- a/Settings.h
+++ b/Settings.h
@@ -61,6 +61,9 @@ typedef struct Settings_ {
    bool accountGuestInCPUMeter;
    bool headerMargin;
    bool enableMouse;
+   #ifdef HTOP_LINUX
+   bool simpleUsedMemForm;
+   #endif
    #ifdef HAVE_LIBHWLOC
    bool topologyAffinity;
    #endif
--- a/linux/Platform.c
+++ b/linux/Platform.c
@@ -287,8 +287,14 @@ void Platform_setMemoryValues(Meter* this) {
    long int usedMem = pl->usedMem;
    long int buffersMem = pl->buffersMem;
    long int cachedMem = pl->cachedMem;
-   usedMem -= buffersMem + cachedMem;
    this->total = pl->totalMem;
+
+   if (pl->settings->simpleUsedMemForm) {
+      usedMem = pl->totalMem - lpl->availableMem;
+   } else {
+      usedMem -= buffersMem + cachedMem;
+   }
+
    this->values[0] = usedMem;
    this->values[1] = buffersMem;
    this->values[2] = cachedMem;
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -1514,6 +1514,7 @@ static inline void LinuxProcessList_scanMemoryInfo(ProcessList* this) {
       case 'M':
          tryRead("MemTotal:", &this->totalMem);
          tryRead("MemFree:", &freeMem);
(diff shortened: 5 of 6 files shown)


linux/Platform.c
```diff
@@ -287,8 +287,15 @@ void Platform_setMemoryValues(Meter* this) {
    long int usedMem = pl->usedMem;
    long int buffersMem = pl->buffersMem;
    long int cachedMem = pl->cachedMem;
-   usedMem -= buffersMem + cachedMem;
    this->total = pl->totalMem;
+
+   if (Settings_getSettings()->usedMemOnly) {
```
BenBE, line 292: The settings are already part of the value passed in `pl` by looking up `pl->settings`. Please don't introduce any unnecessary globals …

Settings.c
```diff
@@ -433,3 +441,7 @@ void Settings_invertSortOrder(Settings* this) {
       this->direction = 1;
    }
 }
+
+const Settings* Settings_getSettings(void) {
+   return settings;
+}
```
BenBE, line 447: No need for this new API, as the value is already handed everywhere where it's possibly needed.

Settings.c
```diff
@@ -19,6 +19,8 @@ in the source distribution for its full text.
 #include "XUtils.h"
 
 
+static const Settings* settings;
+
```
BenBE, line 23: No need for this global …

Settings.c
```diff
@@ -423,6 +428,9 @@ Settings* Settings_new(int initialCpuCount) {
       this->showMergedCommand = false;
       this->headerMargin = true;
    }
+
+   settings = this;
+
```
BenBE, line 433: if we were to store this globally, there should be some proper reuse strategy, as done in this way may cause hard-to-debug issues. Routines should not have non-obvious globally visible side-effects.

Settings.h
```diff
@@ -78,4 +79,6 @@ Settings* Settings_new(int initialCpuCount);
 
 void Settings_invertSortOrder(Settings* this);
 
+const Settings* Settings_getSettings(void);
+
```
BenBE, line 83: This is already handled differently in the rest of the code. Cf. first field in `ProcessList` structure ;-)

BenBE (changes requested):
Lacks actual UI control to change this setting.

linux/Platform.c
tWido: Yep, that pointer is exactly what I've been looking for, thanks for hint.

tWido:
> Lacks actual UI control to change this setting.

Added setting to UI. For now it's standalone commit, I will squash it after everything is approved.

BenBE (changes requested):
Only offer this settings on Linux OR implement it on all platforms proper.

DisplayOptionsPanel.c
```diff
@@ -119,6 +119,9 @@ DisplayOptionsPanel* DisplayOptionsPanel_new(Settings* settings, ScreenManager*
    Panel_add(super, (Object*) CheckItem_newByRef("Also show CPU temperature (requires libsensors)", &(settings->showCPUTemperature)));
    Panel_add(super, (Object*) CheckItem_newByRef("- Show temperature in degree Fahrenheit instead of Celsius", &(settings->degreeFahrenheit)));
    #endif
+   #ifdef __linux__
```
BenBE, line 122: There's a macro `HTOP_LINUX` for this.

ProcessList.h
```diff
@@ -63,6 +63,7 @@ typedef struct ProcessList_ {
 
    unsigned long long int totalMem;
    unsigned long long int usedMem;
+   unsigned long long int availableMem;
```
BenBE, line 66: Platform-dependent fields normally should also go into the platform-dependent struct (in this case `LinuxProcessList`).

BenBE (changes requested):

BenBE:
Quite unhappy with having this basically Linux-only, when other OS (like FreeBSD) should provide similar information too.

Thus the open question: Is it okay, to have the display of available memory in this way be Linux-specific, or should we aim for having this consistent across platforms?

fasterit:
> Quite unhappy with having this basically Linux-only, when other OS (like FreeBSD) should provide similar information too.
> 
> Thus the open question: Is it okay, to have the display of available memory in this way be Linux-specific, or should we aim for having this consistent across platforms?

If we can have this on BSD and / or Darwin, too, it would be much preferred. And we could use the global struct then.

DisplayOptionsPanel.c
```diff
@@ -119,6 +119,7 @@ DisplayOptionsPanel* DisplayOptionsPanel_new(Settings* settings, ScreenManager*
    Panel_add(super, (Object*) CheckItem_newByRef("Also show CPU temperature (requires libsensors)", &(settings->showCPUTemperature)));
    Panel_add(super, (Object*) CheckItem_newByRef("- Show temperature in degree Fahrenheit instead of Celsius", &(settings->degreeFahrenheit)));
    #endif
+   Panel_add(super, (Object*) CheckItem_newByRef("Show only memory that can't be freed", &(settings->usedMemOnly)));
```
fasterit, line 123: Suggest to reword:
Count only memory that cannot be freed as used

tWido:
Now i wonder, issue mentions only used memory. So are other counters okay and should be displayed or are they affiliated with unfreeable memory? I will ask the requester for that info.

For BSD we lack information even more. It would be great, if some BSD kernel dev provided some knowledge, otherwise i will try to look it up in documentation.

linux/Platform.c
```diff
@@ -287,8 +287,17 @@ void Platform_setMemoryValues(Meter* this) {
...
    long int buffersMem = pl->buffersMem;
    long int cachedMem = pl->cachedMem;
-   usedMem -= buffersMem + cachedMem;
    this->total = pl->totalMem;
+
+   if (pl->settings->usedMemOnly) {
+      usedMem = pl->totalMem - lpl->availableMem;
+      this->values[0] = usedMem;
+      this->values[1] = 0;
+      this->values[2] = 0;
```
fasterit, line 296: Why not show buffersMem and cachedMem in the meter even with the changed usedMem calculation?

tWido: Changed calculation resulted in higher memory consumption, so i thought that part of buffersMem and/or cachedMem is counted in usedMem. Such overlap might end up in showing more memory being used, than there actually is. But I might be just wrong. I asked for more info in linked issue.

tWido:
I've tried a little test. Ran a script to take up whole RAM and see differences in output. With proposed calculation formula, used went up to  cca 15.5  out fo 15.6GB. Actual formula showed cca 15.2, so proposed one is a bit more precise. 
Screenshot_20201216_160240
I also looked at code of linked git commit and MemAvailable is surely best guess at available and thus used memory.

So, hopefully last question is, should this new formula be:
- default, without option to switch back to old one
- default, with option to switch back
- just an option

BenBE:
> * just an option

fasterit:
Merged #452. Many thanks to @tWido and @ChrisCheney for reporting the issue and an initial implementation.

PR #386, opened 2020-12-11
Drop redundant sys/dirent.h include

ginggs:
sys/dirent.h is included by dirent.h in FreeBSD, and does not exist in Debian GNU/kFreeBSD


BenBE (approved):
LGTM.

BenBE (approved):

PR #387, opened 2020-12-12
Fix building against a CentOS 6 kernel

chrisburr:
This PR contains two minor fixes that were needed in conda-forge for htop 3.0.3 to be buildable against the CentOS 6 kernel headers.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -142,7 +142,7 @@ linux_platform_headers = \
 	zfs/ZfsCompressedArcMeter.h
 
 if HTOP_LINUX
-AM_LDFLAGS += -rdynamic
+AM_LDFLAGS += -rdynamic -lrt
 myhtopplatsources = \
 	linux/IOPriorityPanel.c \
 	linux/LibSensors.c \
--- a/Makefile.am
+++ b/Makefile.am
@@ -142,7 +142,7 @@ linux_platform_headers = \
 	zfs/ZfsCompressedArcMeter.h
 
 if HTOP_LINUX
-AM_LDFLAGS += -rdynamic -lrt
+AM_LDFLAGS += -rdynamic
 myhtopplatsources = \
 	linux/IOPriorityPanel.c \
 	linux/LibSensors.c \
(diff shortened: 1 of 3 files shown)


Makefile.am
```diff
@@ -142,7 +142,7 @@ linux_platform_headers = \
 	zfs/ZfsCompressedArcMeter.h
 
 if HTOP_LINUX
-AM_LDFLAGS += -rdynamic
+AM_LDFLAGS += -rdynamic -lrt
```
BenBE, line 145: Shouldn't this be detected via `configure.ac` instead? Similar to what is done for `dlopen` …

linux/LinuxProcessList.c
```diff
@@ -56,6 +56,11 @@ in the source distribution for its full text.
 #include "LibSensors.h"
 #endif
 
+// CentOS 6's kernel doesn't provide a definition of O_PATH
+#ifndef O_PATH
+# define O_PATH 010000000
```
BenBE, line 61: Is this octal number intentional? Or should it read `0x10000000` instead?
```suggestion
# define O_PATH 0x10000000
```

BenBE (changes requested):

linux/LinuxProcessList.c
chrisburr: Yes, this matches the way it's defined in Linux:

https://github.com/torvalds/linux/blob/6f0d349d922ba44e4348a17a78ea51b7135965b1/include/uapi/asm-generic/fcntl.h#L84-L86

Makefile.am
chrisburr: Probably, I'll give it a try when I'm back in front of a computer

linux/LinuxProcessList.c
BenBE: Can you add this reference somehow for documentation?

Like: `// based on definition taken from uapi/asm-generic/fcnth.h in Linux kernel tree`

Just so that people have a reference where this comes from.

BenBE:
Merged with minor rebase/squashing.

PR #388, opened 2020-12-12
Refactor makeCommandStr

BenBE:
This is some refactoring work for `makeCommandStr` to make this suitable for platform independent use.

Changes include:
* Move `cmdline`, `comm` and `exe` to platform-independent `Process` structure.
* Move `ProcessMergeCommand` to platform-independent `Process` structure.
* Move `procExeDeleted` to platform-independent `Process` structure.
* Allow for individual activation/deactivation of highlighted components (basename, comm, deleted, separator, …)

The current state is mostly some PoC and still WIP.


cgzones:
After switching from *merged* to *standard* mode:
```
linux/LinuxProcess.c:432:10: runtime error: implicit conversion from type 'int' of value -38 (32-bit, signed) to type 'size_t' (aka 'unsigned long') changed the value to 18446744073709551578 (64-bit, unsigned)
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior linux/LinuxProcess.c:432:10 in
```

BenBE:
Moving this item back a few releases, as we'd run into several issues I noticed:

1. To move `PROC_COMM`/`PROC_EXE` to platform-independent grounds I'd have to change user-facing enums, thus breaking existing configurations using these columns. Requires a proper upgrade path, which is out-of-scope of this issue.
2. Some internal APIs (`Process_isKernelThread`/`Process_isUserlandThread`) are named as Platform-Independent, but actually are not (platform-implemented everywhere except MacOS). Repairing this is another issue I don't want to address right now, as this is a general code issue.
3. To implement some changes I noticed I'd actually need some updates still pending in open PRs (e.g. #405, #411).

Thus even though I'd really like to bring this refactoring in ASAP it's causing too much noise to justify it right now.

cgzones:
Maybe you could refactor everything process command related into a C++ like class:
A struct with all the relevant data fields:
```c
stuct ProcessCommand {
   char* fullCommand;
   char* comm;   
   char* exe;

   int startIndexOfExeInFullCommand;
   int endIndexOfExeInFullCommand;
   int startIndexOfCommInFullCommand;
   int endIndexOfCommInFullCommand;
   ...
```

and then have public methods
```c
void ProcessCommand_updateExe(ProcessCommand* pc, const char* exe);
void ProcessCommand_updateComm(ProcessCommand* pc, const char* comm);
void ProcessCommand_updateFullCommand(ProcessCommand* pc, const char* fullCommand);
void ProcessCommand_format(const ProcessCommand* pc, RichString* str);
```

and **no** code outside of these public functions (and maybe some private static ones) reads/writes any of the data fields of the ProcessCommand struct directly.
That way it might be easier to ensure/review/maintain all invariants are always guaranteed (indices are either -1 or point to valid memory, indices are -1 if the corresponding data is not-available/NULL, ...).

Another thought: does it make sense to colour common command-path prefixes as gray (like `/usr/bin/`, `/usr/sbin/`, `/sbin/`, `/bin/`, `/lib/systemd/`, `/usr/libexec/`)?

BenBE:
The first step I'm working on right now is bringing all the required fields into the common code base, i.e. the following fields:
* comm (cmdline, the process command line)
* procComm (procComm, command title)
* procExe (procExe, running executable)

* basenameOffset (cmdlineBasenameOffset, actually end of first command line token)
* procCmdlineBasenameOffset (cmdlineBasenameStart, start of command file basename)
* procCmdlineBasenameEnd -> Removed (overlaps with cmdlineBasenameOffset)
* procExeBasenameOffset (procExeBasenameStart)
* procExeLen -> Removed (implicit by `strlen(procExe)`)
* procExeDeleted (procExeDeleted, true if ProcExe was modified/deleted in the filesystem)
* …

After this I planned to make the common code paths (like extracting the basename from the command line, more homogeneous. Abstracting them away in a pseudo-class might be worth a look, but most implementations will likely use one of two implementations anyway: Extract from NUL-terminated vs. extract from blank-separated, where the first may trigger the second one internally via heuristic (e.g. for Chrome, wine). When setting the command line I more or less already know most of the fields, thus a per-OS split of the "calculate-internal-fields" routine doesn't make much sense. It's extracting this routine and bring it to a common code-base that's currently the hardest part. But to facilitate this I need to have all required fields in the common code-base first, which is ATM still WIP.

Process.c
```diff
@@ -179,35 +179,359 @@ void Process_fillStarttimeBuffer(Process* this) {
...
-            basename = i + 1;
-         } else if (comm[i] == ':') {
-            len = i + 1;
-            break;
+/*
+ * TASK_COMM_LEN is defined to be 16 for /proc/[pid]/comm in man proc(5), but it is
+ * not available in an userspace header - so define it.
+ *
+ * Note: This is taken from LINUX headers, but implicitly taken for other platforms
+ * for sake of bravity.
```
natoscott, line 187: brevity?

Process.c
```diff
@@ -179,35 +179,359 @@ void Process_fillStarttimeBuffer(Process* this) {
...
+    * Its content is based on the fields cmdline, comm, and exe. */
+   if (
+       mc->prevMergeSet == showMergedCommand &&
+       mc->prevPathSet == showProgramPath &&
+       mc->prevCommSet == searchCommInCmdline &&
+       mc->prevCmdlineSet == stripExeFromCmdline &&
+       !mc->cmdlineChanged &&
+       !mc->commChanged &&
+       !mc->exeChanged)
+   {
```
natoscott, line 329: Opening brace should be on previous line?

Process.h
```diff
@@ -55,6 +55,37 @@ typedef enum ProcessField_ {
 
 struct Settings_;
 
+/* Holds information about regions of the cmdline that should be
+ * highlighted (e.g. program basename, delimiter, comm). */
+typedef struct ProcessCmdlineHighlight_
+{
```
natoscott, line 61: Opening brace on previous line.

Process.h
```diff
@@ -55,6 +55,37 @@ typedef enum ProcessField_ {
...
+   int attr;      /* The attributes used to highlight */
+   int flags;     /* Special flags used for selective highlighting, zero for always */
+} ProcessCmdlineHighlight;
+
+/* ProcessMergedCommand is populated by Process_makeCommandStr: It
+ * contains the merged Command string, and the information needed by
+ * Process_writeCommand to color the string. str will be NULL for kernel
+ * threads and zombies */
+typedef struct ProcessMergedCommand_
+{
```
natoscott, line 73: Ditto.

Process.h
```diff
@@ -84,6 +112,12 @@ typedef struct Process_ {
    /* Foreground group identifier of the controlling terminal */
    int tpgid;
 
+   /* This is a kernel (helper) task */
+   bool isKernelThread;
+
+   /* This is a userland thread / LWP */
+   bool isUserThread;
+
    /*
```
natoscott, line 121: Could these two and the procExeDeleted field be single-bit bitfields?  (as done in ProcessMergedCommand)

linux/LinuxProcess.h
```diff
@@ -30,42 +30,8 @@ in the source distribution for its full text.
...
-   bool unmatchedExe;   /* whether exe matched with cmdline */
-   bool cmdlineChanged; /* whether cmdline changed */
-   bool exeChanged;     /* whether exe changed */
-   bool commChanged;    /* whether comm changed */
-   bool prevMergeSet;   /* whether showMergedCommand was set */
-   bool prevPathSet;    /* whether showProgramPath was set */
-   bool prevCommSet;    /* whether findCommInCmdline was set */
-   bool prevCmdlineSet; /* whether findCommInCmdline was set */
-} LinuxProcessMergedCommand;
-
```
natoscott, line 57: Awesome!

linux/LinuxProcessList.c
```diff
@@ -1165,32 +1164,37 @@ static bool LinuxProcessList_readCmdlineFile(Process* process, openat_arg_t proc
...
-            lp->procExeDeleted = String_eq(lp->procExe + strlen(lp->procExe) - strlen(deletedMarker), deletedMarker);
+         if (strlen(process->procExe) > strlen(deletedMarker)) {
+            process->procExeDeleted = String_eq(process->procExe + strlen(process->procExe) - strlen(deletedMarker), deletedMarker);
+
+            if (process->procExeDeleted) {
+               process->procExe[strlen(process->procExe) - strlen(deletedMarker)] = '\0';
 
-            if (lp->procExeDeleted && strlen(lp->procExe) - strlen(deletedMarker) == 1 && lp->procExe[0] == '/') {
-               lp->procExeBasenameOffset = 0;
+               if (String_eq(process->procExe, "/"))
```
natoscott, line 1187: Is this right?  String_eq will be comparing the full procExe string not just the first character.

natoscott (changes requested):
@BenBE few small thing, but this is looking good!  A big step forward for all non-Linux platforms.

linux/LinuxProcessList.c
BenBE: Yes. The full string is `/ (deleted)` when read from procfs and the ` (deleted)` is trimmed off the end of that string in the previous line. Thus only `/` remains at this point.

Process.h
BenBE: Sure they could. Maybe also combine with the fields `updated`, `tag`, `show`, `wasShown`, and `showChildren` below.

Process.c
BenBE: Pondered about this, but I find it visually clearer to have the opening brace on a separate line. Otherwise this kinda looks strange.

natoscott:
@BenBE there seems to be a regression relating to the MergedCommand 'str' field and its sometimes-null property - here's my notes from IRC overnight:
```
<nathans> in linux/LinuxProcess.c LinuxProcess_writeField handling of case COMM
<nathans> there's this conditional: if ((Process_isUserlandThread(this) && this->settings->showThreadNames) || !this->mergedCommand.str)
<nathans> if true, we then call Process_writeField
<nathans> the problem is the final branch of the condition - the !null mergedCommand.str check - if that is null, we get into trouble
<nathans> because Process_writeField unconditional dereferences that pointer
<nathans> the exact path there is: Process_writeField calls Process_writeCommand (line 635) which calls RichString_appendWide (line 527) which gets passed the null mergedCommand.str
```

Process.h
```diff
@@ -84,6 +112,12 @@ typedef struct Process_ {
    /* Foreground group identifier of the controlling terminal */
    int tpgid;
 
+   /* This is a kernel (helper) task */
+   bool isKernelThread;
+
+   /* This is a userland thread / LWP */
+   bool isUserThread;
```
natoscott, line 119: In many (if not all) other places this field was 'isUserlandThread' including in the comment above - could we stick with that convention?  There are functions named isUserlandThread() and fields in structures in ProcessList.h and Settings.h that are also named with the 'userland' convention.

linux/LinuxProcess.c
```diff
@@ -106,11 +103,11 @@ const ProcessFieldData Process_fields[LAST_PROCESSFIELD] = {
  * happens on what is displayed - whether comm, full path, basename, etc.. So
  * this follows LinuxProcess_writeField(COMM) and LinuxProcess_writeCommand */
```
natoscott, line 104: This comment is out-of-date now - the reference to LinuxProcess_writeCommand should be Process_writeCommand now.

Process.h
BenBE: Will update the patch series accordingly to use `isUserlandThread`.

linux/LinuxProcess.c
BenBE: Okay, oversight. Already fixed another instance and hoped I caught 'em all …

BenBE:
> @BenBE there seems to be a regression relating to the MergedCommand 'str' field and its sometimes-null property - here's my notes from IRC overnight:
> 
> ```
> <nathans> in linux/LinuxProcess.c LinuxProcess_writeField handling of case COMM
> <nathans> there's this conditional: if ((Process_isUserlandThread(this) && this->settings->showThreadNames) || !this->mergedCommand.str)
> <nathans> if true, we then call Process_writeField
> <nathans> the problem is the final branch of the condition - the !null mergedCommand.str check - if that is null, we get into trouble
> <nathans> because Process_writeField unconditional dereferences that pointer
> <nathans> the exact path there is: Process_writeField calls Process_writeCommand (line 635) which calls RichString_appendWide (line 527) which gets passed the null mergedCommand.str
> ```

Should be fixed now. The regression was introduced by accident as `COMM` was calling `Process_getCommand` which should have called `Process_getCommandStr` instead of reading the base field. Basically the same situation we also have with `Process_writeField` and `Process_compareByKey` which should fall back to a sane `*_Base` implemenetation instead of macro-handling it.

natoscott:
@BenBE hmm, I'm still seeing a sigsegv in the same location...?

```
Error information:
------------------
A signal 11 (Segmentation fault) was received.

Backtrace information:
----------------------
The following function calls were active when the issue was detected:
---
./htop(CRT_handleSIGSEGV+0xb3)[0x40d8c3]
/lib64/libc.so.6(+0x3da60)[0x7f1c7ac2ea60]
/lib64/libc.so.6(+0x162605)[0x7f1c7ad53605]
./htop(RichString_appendWide+0x22)[0x418f92]
./htop(Process_writeCommand+0x35)[0x416645]
./htop(Process_writeField+0x6d4)[0x416d84]
./htop[0x41e907]
./htop(Process_display+0x31)[0x414ee1]
./htop(Panel_draw+0x1db)[0x41460b]
./htop(ScreenManager_run+0x182)[0x419882]
./htop(CommandLine_run+0x5d8)[0x40c588]
/lib64/libc.so.6(__libc_start_main+0xf2)[0x7f1c7ac191e2]
./htop(_start+0x2e)[0x408ace]
---
```

BenBE:
@natoscott Nope, this looks to be another one. Any special steps to reproduce this one? Branch seems to run fine on my systems.

natoscott:
> @natoscott Nope, this looks to be another one. Any special steps to reproduce this one? Branch seems to run fine on my systems.

Looks like the process triggering it here is the selinux 'setroubleshootd' daemon - here's details:

```
(gdb) p *this
$4 = {super = {klass = 0x42ec00 <LinuxProcess_class>}, processList = 0x447af0, 
  settings = 0x448300, pid = 189783, ppid = 1, tgid = 189781, pgrp = 189781, 
  session = 189781, tpgid = -1, isKernelThread = false, 
  isUserlandThread = true, tty_nr = 0, tty_name = 0x0, st_uid = 989, 
  user = 0x64c990 "setroubleshoot", time = 179, 
  cmdline = 0x64c9b0 "/usr/bin/python3\n-Es\n/usr/sbin/setroubleshootd\n-f", 
  cmdlineBasenameEnd = 16, cmdlineBasenameStart = 9, 
  procComm = 0x64c9f0 "setroubleshootd", procExe = 0x0, 
  procExeBasenameOffset = 0, procExeDeleted = false, processor = 7, 
  percent_cpu = 100, percent_mem = 0.835037589, priority = 20, nice = 0, 
  nlwp = 4, starttime_ctime = 1619421558, starttime_show = "17:19 \000", 
  m_virt = 595880, m_resident = 133664, minflt = 7306, majflt = 0, 
  state = 82 'R', updated = false, tag = false, show = true, wasShown = true, 
  showChildren = true, seenStampMs = 0, tombStampMs = 0, indent = 3, 
  tree_left = 105, tree_right = 106, tree_depth = 2, tree_index = 53, 
  mergedCommand = {str = 0x0, maxLen = 64, highlightCount = 0, highlights = {{
        offset = 0, length = 0, attr = 0, flags = 0}, {offset = 0, length = 0, 
        attr = 0, flags = 0}, {offset = 0, length = 0, attr = 0, flags = 0}, {
        offset = 0, length = 0, attr = 0, flags = 0}, {offset = 0, length = 0, 
        attr = 0, flags = 0}, {offset = 0, length = 0, attr = 0, flags = 0}, {
        offset = 0, length = 0, attr = 0, flags = 0}, {offset = 0, length = 0, 
        attr = 0, flags = 0}}, separateComm = false, unmatchedExe = false, 
    cmdlineChanged = true, exeChanged = false, commChanged = true, 
    prevMergeSet = false, prevPathSet = false, prevCommSet = false, 
    prevCmdlineSet = false}}
```

and the stack trace looks like this (mergedCommand.str field is null, similar to before):

```
(gdb) bt
#0  0x00007ffff7d60605 in __strlen_avx2 () from /lib64/libc.so.6
#1  0x0000000000418f92 in RichString_appendWide (
    this=this@entry=0x7fffffffaee0, attrs=attrs@entry=10240, data=0x0)
    at RichString.c:200
#2  0x0000000000416645 in Process_writeCommand (this=this@entry=0x64c690, 
    attr=attr@entry=10240, baseAttr=baseAttr@entry=2107392, 
    str=str@entry=0x7fffffffaee0) at Process.c:527
#3  0x0000000000416d84 in Process_writeField (this=this@entry=0x64c690, 
    str=str@entry=0x7fffffffaee0, field=field@entry=COMM) at Process.c:635
#4  0x000000000041e907 in LinuxProcess_writeField (this=0x64c690, 
    str=0x7fffffffaee0, field=<optimized out>) at linux/LinuxProcess.c:307
#5  0x0000000000414ee1 in Process_display (cast=0x64c690, out=0x7fffffffaee0)
    at Process.c:776
#6  0x000000000041460b in Panel_draw (this=0x493300, 
    force_redraw=<optimized out>, focus=<optimized out>, 
    highlightSelected=true, hideFunctionBar=<optimized out>) at Panel.c:274
#7  0x0000000000419882 in ScreenManager_drawPanels (
    force_redraw=<optimized out>, focus=0, this=0x451760)
    at ScreenManager.c:129
#8  ScreenManager_run (this=this@entry=0x451760, 
    lastFocus=lastFocus@entry=0x0, lastKey=lastKey@entry=0x0)
    at ScreenManager.c:162
#9  0x000000000040c588 in CommandLine_run (name=<optimized out>, 
    argc=<optimized out>, argv=<optimized out>) at CommandLine.c:352
#10 0x00007ffff7c261e2 in __libc_start_main () from /lib64/libc.so.6
#11 0x0000000000408ace in _start ()
```

BenBE:
> Looks like the process triggering it here is the selinux 'setroubleshootd' daemon - here's details:
> 
> ```
> (gdb) p *this
> $4 = {super = {klass = 0x42ec00 <LinuxProcess_class>}, processList = 0x447af0, 
>   settings = 0x448300, pid = 189783, ppid = 1, tgid = 189781, pgrp = 189781, 
>   session = 189781, tpgid = -1, isKernelThread = false, 
>   isUserlandThread = true, tty_nr = 0, tty_name = 0x0, st_uid = 989, 
>   user = 0x64c990 "setroubleshoot", time = 179, 
>   cmdline = 0x64c9b0 "/usr/bin/python3\n-Es\n/usr/sbin/setroubleshootd\n-f", 
>   cmdlineBasenameEnd = 16, cmdlineBasenameStart = 9, 
>   procComm = 0x64c9f0 "setroubleshootd", procExe = 0x0, 
>   procExeBasenameOffset = 0, procExeDeleted = false, processor = 7, 
>   percent_cpu = 100, percent_mem = 0.835037589, priority = 20, nice = 0, 
>   nlwp = 4, starttime_ctime = 1619421558, starttime_show = "17:19 \000", 
>   m_virt = 595880, m_resident = 133664, minflt = 7306, majflt = 0, 
>   state = 82 'R', updated = false, tag = false, show = true, wasShown = true, 
>   showChildren = true, seenStampMs = 0, tombStampMs = 0, indent = 3, 
>   tree_left = 105, tree_right = 106, tree_depth = 2, tree_index = 53, 
>   mergedCommand = {str = 0x0, maxLen = 64, highlightCount = 0, highlights = {{
>         offset = 0, length = 0, attr = 0, flags = 0}, {offset = 0, length = 0, 
>         attr = 0, flags = 0}, {offset = 0, length = 0, attr = 0, flags = 0}, {
>         offset = 0, length = 0, attr = 0, flags = 0}, {offset = 0, length = 0, 
>         attr = 0, flags = 0}, {offset = 0, length = 0, attr = 0, flags = 0}, {
>         offset = 0, length = 0, attr = 0, flags = 0}, {offset = 0, length = 0, 
>         attr = 0, flags = 0}}, separateComm = false, unmatchedExe = false, 
>     cmdlineChanged = true, exeChanged = false, commChanged = true, 
>     prevMergeSet = false, prevPathSet = false, prevCommSet = false, 
>     prevCmdlineSet = false}}
> ```

Something is odd:
```
(gdb) p *this
$4 = {super = {klass = 0x42ec00 <LinuxProcess_class>}, processList = 0x447af0, 
  settings = 0x448300, pid = 189783, ppid = 1, tgid = 189781, pgrp = 189781, 
  session = 189781, tpgid = -1, isKernelThread = false, 
  isUserlandThread = true, …, user = 0x64c990 "setroubleshoot", …,
  cmdline = 0x64c9b0 "/usr/bin/python3\n-Es\n/usr/sbin/setroubleshootd\n-f", 
  cmdlineBasenameEnd = 16, cmdlineBasenameStart = 9, 
  procComm = 0x64c9f0 "setroubleshootd", procExe = 0x0, 
  procExeBasenameOffset = 0, procExeDeleted = false, …,
  state = 82 'R', updated = false, tag = false, show = true, wasShown = true, 
  …, seenStampMs = 0, tombStampMs = 0, …,
  mergedCommand = {str = 0x0, maxLen = 64, highlightCount = 0,
    highlights = {…}, separateComm = false, unmatchedExe = false,
    cmdlineChanged = true, exeChanged = false, commChanged = true,
    …
    }}
```

What we learn from this:
* The process is definitively a userland thread (`tgid != pid`)
* The thread exited (`!updated` and `wasShown`)
* The process's command string hasn't previously been created or `makeCommandStr` failed
* There are changes to `cmdline` and `procComm` that haven't yet been processed
* The thread hasn't been marked as zombie, otherwise `cmdlineBasenameStart` and `cmdlineBasenameEnd` would be `0`/`-1`
* The process was pre-existing 

IIRC in the latest release there was special treatment when processing thread names.

@natoscott Can you confirm, that you are displaying userland threads?

natoscott:
> @natoscott Can you confirm, that you are displaying userland threads?

Yep ...

```
$ grep threads ~/.config/htop/htoprc
hide_kernel_threads=1
hide_userland_threads=0
highlight_threads=1
```

natoscott (approved):

eworm-de:
This breaks to show userland threads: Basename is no longer highlighted, instead question marks appear.
Screenshot_2021-04-28_08-22-23

Some revisions crashed with segmentation fault, so `git bisect` found the first bad commit could be any of:
5b3553fd697e042fc5fd59cb55e4a4ad7f33bd25
5fdabf8d76e0427e572ca635469ad41a9c18ebd8
ccd3d731b47c89a07fa669afc948ff6e69cb1524
f5e56fee4fa85300262a4dd617c8536d24633bcd

Process.c
```diff
@@ -36,6 +36,9 @@ in the source distribution for its full text.
 #endif
 
 
+/* Used to identify kernel threads in Comm and Exe columns */
+static const char *const kthreadID = "KTHREAD";
```
cgzones, line 40: Most of the new code aligns the pointer specifier to the type on the left.
While this is generally also my preference (e.g. to avoid `char* a, b;`), most of the existing code uses right-/name-alignment and the style-guide also mentions `astyle ... -k1` (= `--align-pointer=type`).

Should we adjust the style guide to allow both types of alignment (with or without a preference)?

Process.c
```diff
@@ -253,35 +256,374 @@ void Process_fillStarttimeBuffer(Process* this) {
...
+      if (*token) {
+         do {
+            ++token;
+         } while (*token && '\n' == *token);
+      }
+   }
+   return false;
+}
+
+static int matchCmdlinePrefixWithExeSuffix(const char *cmdline, int cmdlineBaseOffset, const char *exe, int exeBaseOffset, int exeBaseLen) {
```
cgzones, line 307: Maybe one sentence of comment what this function does and what it returns.

Process.c
```diff
@@ -253,35 +256,374 @@ void Process_fillStarttimeBuffer(Process* this) {
...
+   /* Display cmdline if it hasn't been consumed by procExe */
+   if (*cmdline)
+      (void)stpcpyWithNewlineConversion(str, cmdline);
+
+   #undef WRITE_SEPARATOR
+   #undef WRITE_HIGHLIGHT
+}
+
+void Process_writeCommand(const Process *this, int attr, int baseAttr, RichString *str) {
+   (void)baseAttr;
```
cgzones, line 591: Why not remove the function parameter `baseAttr`?

Process.c
```diff
@@ -381,6 +723,33 @@ void Process_writeField(const Process* this, RichString* str, ProcessField field
...
+      const char* procComm;
+      if (this->procComm) {
+         attr = CRT_colors[Process_isUserlandThread(this) ? PROCESS_THREAD_COMM : PROCESS_COMM];
+         procComm = this->procComm;
+      } else {
+         attr = CRT_colors[PROCESS_SHADOW];
+         procComm = Process_isKernelThread(this) ? kthreadID : "N/A";
+      }
+      /* 15 being (TASK_COMM_LEN - 1) */
+      Process_printLeftAlignedField(str, attr, procComm, 15);
```
cgzones, line 736: maybe just use `TASK_COMM_LEN - 1` instead of `15` and remove the comment?
Every compiler should compute it at compile time.

Process.c
```diff
@@ -381,6 +723,33 @@ void Process_writeField(const Process* this, RichString* str, ProcessField field
...
+      if (this->procExe) {
+         attr = CRT_colors[Process_isUserlandThread(this) ? PROCESS_THREAD_BASENAME : PROCESS_BASENAME];
+         if (this->procExeDeleted)
+            attr = CRT_colors[FAILED_READ];
+         procExe = this->procExe + this->procExeBasenameOffset;
+      } else {
+         attr = CRT_colors[PROCESS_SHADOW];
+         procExe = Process_isKernelThread(this) ? kthreadID : "N/A";
+      }
+      Process_printLeftAlignedField(str, attr, procExe, 15);
```
cgzones, line 750: dito

Process.h
```diff
@@ -241,6 +297,18 @@ static inline bool Process_isChildOf(const Process* this, pid_t pid) {
...
+#define Process_isUserlandThread(_process) ((_process)->isUserlandThread)
+
+static inline bool Process_isThread(const Process *this) {
+   return Process_isUserlandThread(this) || Process_isKernelThread(this);
+}
+
+#define CMDLINE_HIGHLIGHT_FLAG_SEPARATOR  0x00000001
+#define CMDLINE_HIGHLIGHT_FLAG_BASENAME   0x00000002
+#define CMDLINE_HIGHLIGHT_FLAG_COMM       0x00000004
+#define CMDLINE_HIGHLIGHT_FLAG_DELETED    0x00000008
```
cgzones, line 310: These macros seem to be only used in `Process.c`, maybe also define them there?

Process.h
```diff
@@ -241,6 +297,18 @@ static inline bool Process_isChildOf(const Process* this, pid_t pid) {
    return pid == Process_getParentPid(this);
 }
 
+#define Process_isKernelThread(_process) ((_process)->isKernelThread)
```
cgzones, line 300: I'd prefer a type safe function
```c
static inline bool Process_isKernelThread(const Process *this) {
   return this->isKernelThread;
}
```

ProcessList.c
```diff
@@ -624,8 +626,6 @@ void ProcessList_scan(ProcessList* this, bool pauseProcessUpdate) {
             // immediately remove
             ProcessList_remove(this, p);
          }
-      } else {
-         p->updated = false;
```
cgzones, line 628: Why is this no longer needed?
I do not see where else it might be set, so the previous condition `} else if (p->updated == false) {` can trigger on no-longer existing entries.

dragonflybsd/DragonFlyBSDProcessList.c
```diff
@@ -385,12 +385,13 @@ void ProcessList_goThroughEntries(ProcessList* super, bool pauseProcessUpdate) {
...
             proc->pid = (pid_t)kproc->kp_ktaddr;
-            dfp->kernel = 1;
+            proc->isKernelThread = true;
          } else {
             proc->pid = kproc->kp_pid;		// process ID
-            dfp->kernel = 0;
+            proc->isKernelThread = false;
          }
-         proc->ppid = kproc->kp_ppid;		// parent process id
+         proc->isUserlandThread = kproc->kp_nthreads > 1;
```
cgzones, line 393: This seems to be wrong, cause also the parent process can have `kp_nthreads` bigger than 1.

linux/LinuxProcessList.c
```diff
@@ -1455,19 +1451,17 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
...
-         proc->basenameOffset = -1;
-         free_and_xStrdup(&proc->comm, command);
-         lp->procCmdlineBasenameOffset = 0;
-         lp->procCmdlineBasenameEnd = 0;
-         lp->mergedCommand.commChanged = true;
+      if (proc->state == 'Z' && (proc->cmdlineBasenameEnd == 0)) {
+         proc->cmdlineBasenameEnd = -1;
+         free_and_xStrdup(&proc->cmdline, command);
+         proc->cmdlineBasenameStart = 0;
+         proc->mergedCommand.commChanged = true;
```
cgzones, line 1458: why set `proc->mergedCommand.commChanged` (instead of `cmdlineChanged`), when we have changed `proc->cmdline`?

linux/LinuxProcessList.c
```diff
@@ -1455,19 +1451,17 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
...
          if (settings->showThreadNames || Process_isKernelThread(proc)) {
-            proc->basenameOffset = -1;
-            free_and_xStrdup(&proc->comm, command);
-            lp->procCmdlineBasenameOffset = 0;
-            lp->procCmdlineBasenameEnd = 0;
-            lp->mergedCommand.commChanged = true;
+            proc->cmdlineBasenameEnd = -1;
+            free_and_xStrdup(&proc->cmdline, command);
+            proc->cmdlineBasenameStart = 0;
+            proc->mergedCommand.commChanged = true;
```
cgzones, line 1464: dito

Process.c
BenBE: This has been taken 1:1 from the current dev branch, just moved over from `linux/LinuxProcess.c` …

I'm indifferent on that one.

Process.c
BenBE: Mostly because I'm trying to stick with just moving the code into the right places for now. There's a second set of changes I noticed where some more cleanup of this could be done (but which would delay this PR even further).

Process.h
BenBE: As the MergeCommand structure is public interface, so should the constants used in that publically accessible interface.

ProcessList.c
BenBE: This is set before each run of `Process_goThroughProcesses` as an initialization step.
https://github.com/BenBE/htop/blob/command_color_instr/ProcessList.c#L589

dragonflybsd/DragonFlyBSDProcessList.c
BenBE: Just ported over what originally was in https://github.com/BenBE/htop/blob/master/dragonflybsd/DragonFlyBSDProcess.h#L18

linux/LinuxProcessList.c
BenBE: It's the same in the dev branch:
https://github.com/BenBE/htop/blob/master/linux/LinuxProcessList.c#L1458-L1471

`proc->comm` is `proc->cmdline` and still `proc->mergedCommand->commChanged` refers to the process command name, not the command line.

linux/LinuxProcessList.c
BenBE: Same issue as above.

BenBE:
@eworm-de Could you check again?

I know one of the following two commits still changes the behaviour of the (merged) command column, but after some heavy commit reordering I'm out of concentration for now:

* 12aeca8b Move Process_makeCommandStr to global Process implementation
* 2560b377 Call makeCommandStr on all platforms

Unfortunately 12aeca8b on its own does not work (not entirely sure yet why) and 2560b377 already shows the changed command column behaviour.

Could someone please check that
```c
   /* Nothing to do to (Re)Generate the Command string, if the process is:
    * - a kernel thread, or
    * - a zombie from before being under htop's watch, or
    * - a user thread and showThreadNames is not set */
   if (
       Process_isKernelThread(this) ||
       (this->state == 'Z' && !this->mergedCommand.str) ||
       (Process_isUserlandThread(this) && !settings->showThreadNames))
      return;
```
is the inversion of
```c
      /* (Re)Generate the Command string, but only if the process is:
       * - not a kernel thread, and
       * - not a zombie or it became zombie under htop's watch, and
       * - not a user thread or if showThreadNames is not set */
      if (!Process_isKernelThread(proc) &&
          (proc->state != 'Z' || proc->mergedCommand.str) &&
          (!Process_isUserlandThread(proc) || !settings->showThreadNames)) {
         Process_makeCommandStr(proc);
      }
```

The first block is inside `Process_makeCommandStr` and should exit whenever the second block (Linux `ProcessList_scan` logic) would not have called it in the first place.

Process.c
cgzones: Added to meeting agenda

ProcessList.c
cgzones: true

cgzones:
> Could someone please check that
> 
> ```c
>    /* Nothing to do to (Re)Generate the Command string, if the process is:
>     * - a kernel thread, or
>     * - a zombie from before being under htop's watch, or
>     * - a user thread and showThreadNames is not set */
>    if (
>        Process_isKernelThread(this) ||
>        (this->state == 'Z' && !this->mergedCommand.str) ||
>        (Process_isUserlandThread(this) && !settings->showThreadNames))
>       return;
> ```
> 
> is the inversion of
> 
> ```c
>       /* (Re)Generate the Command string, but only if the process is:
>        * - not a kernel thread, and
>        * - not a zombie or it became zombie under htop's watch, and
>        * - not a user thread or if showThreadNames is not set */
>       if (!Process_isKernelThread(proc) &&
>           (proc->state != 'Z' || proc->mergedCommand.str) &&
>           (!Process_isUserlandThread(proc) || !settings->showThreadNames)) {
>          Process_makeCommandStr(proc);
>       }
> ```

Shouldn't it be
```
...
(!Process_isUserlandThread(proc) || settings->showThreadNames)) {
...
```
?

BenBE:
Will take a look, but this looks likely to be the culprit.

eworm-de:
Still the same, now `git bisect` gives the first bad commit could be any of:
12aeca8bab303e2e3d951f2a81144940c2233628
2560b377fb746d7d609d5da48384da896648c81d

BenBE:
@eworm-de Please see [this comment](https://github.com/htop-dev/htop/pull/388#issuecomment-830599678). As far as my investigations have gone, it seems there's some problem left with https://github.com/htop-dev/htop/commit/12aeca8bab303e2e3d951f2a81144940c2233628 that causes it to change the behaviour (in particular on Linux) compared to master. While I managed to look through the large code block that was moved from `linux/LinuxProcess.c` to `Process.c` there's still some accompanying changes I didn't manage to double-check yet. The reduction of the number of crashing commits is mostly due to some re-ordering of commits I did affecting the handling of `Process_isKernelThread` and `Process_isUserlandThread`.

As I understand it right now we have a situation where the commit moving `Process_makeCommandStr` introduces some oddity, which gets partially corrected through the second commit which introduces calls to `Process_makeCommandStr` to some more central places in the code. Aim should be finding the oddity introduced by the first of those two commits.

I also noticed that calling `Process_makeCommandStr` with a kernel thread causes a crash (when skipping the short-circuit check at the beginning). As we already mark when something important has changed, we might opt to skip those checks (see comment above) and update `Process_makeCommandStr` to handle all cases properly.

Also @cgzones noticed some strange ASAN-related issues.

BenBE:
Going through the changes in 12aeca8bab303e2e3d951f2a81144940c2233628 I think I nailed down the issue to where it's caused.

Until now the `COMM` field on Linux was printed by some special casing inside `LinuxProcess_writeField` like this:
```c
   case COMM: {
      if ((Process_isUserlandThread(this) && this->settings->showThreadNames) || !this->mergedCommand.str) {
         Process_writeField(this, str, field);
      } else {
         LinuxProcess_writeCommandField(this, str, buffer, n, attr);
      }
      return;
   }
```

Up to that patch `Process_writeField` calls `Process_writeCommand` which unconditionally uses `this->cmdline` to display the command column (and in cases of threads this contains the thread name). On the other hand `LinuxProcess_writeCommandField` is basically a copy of the code in `Process_writeField` for the `COMM` field (as used for all other OS), but instead of `Process_writeCommand` the Linux-specific version calls `LinuxProcess_writeCommandStr` instead, which unconditionally uses `this->mergeCommand.str` alongside the other information prepared by a call to `LinuxProcess_makeCommandStr`.

With the patch, `Process_writeCommand` has been removed and `LinuxProcess_writeCommandStr` has taken its place, thus in the above code sample making those two code paths equal like this:

```c
   case COMM: {
      if ((Process_isUserlandThread(this) && this->settings->showThreadNames) || !this->mergedCommand.str) {
         LinuxProcess_writeCommandField(this, str, buffer, n, attr);
      } else {
         LinuxProcess_writeCommandField(this, str, buffer, n, attr);
      }
      return;
   }
```

I think that makes the issue introduced by this patch much more obvious, I guess. ;-)

So, the change of behaviour is two-fold in this case:
1. `Process_writeCommand` and `LinuxProcess_writeCommandStr` worked slightly different. Thus unifying these code paths caused the rendering of userland threads to be handled by the wrong routine that lacked the set of prepared information for proper operation. This information would only be provided by the introduction of the call to `Process_makeCommandStr` in the second commit (2560b377fb746d7d609d5da48384da896648c81d) and that way covering up the actual issue from the wrong code path being used.
2. The commit 12aeca8bab303e2e3d951f2a81144940c2233628 makes an issue visible, that has slept in this implementation for quite some time already, i.e. that for some items a shortcut had been chosen and hoping things were in sync enough to cover up that not all situations properly generated information for both methods to work, should they be selected due to some mis-alignment.

The proposed (short-term) fix thus is back-porting the behaviour of the old `Process_writeCommand` function and integrate it into the implementation of `LinuxProcess_writeCommandStr`. This can be done without much effort, as the major check can just be added quite early in the function implementation and re-inserting the code of the old implementation without much need for change. The longer-term fix is hinted in commit 2560b377fb746d7d609d5da48384da896648c81d by having all platform implementations always call `Process_makeCommandStr` to prepare the required meta-information for rendering.

Okay, so remains the question with the question marks: These appear, when the `this->cmdline` field is displayed by `RichString_appendWide` without it being processed by `stpcpyWithNewlineConversion` first. `LinuxProcess_makeCommandStr` performs this operation, so `LinuxProcess_writeCommandStr` does not need to; while in contrast `Process_writeCommand` does take the unmodified `this->cmdline` field without performing this conversion. This is fine for thread names, as they generally don't contain newline characters, but breaks down for entire command lines, when arguments are split by `NUL` characters, which are internally converted to newline characters.

I hope this somehow explains this strange bug and also where we stand on fixing it. I hope to get around whipping up some updated patch series that resolves the mentioned issues soon.

BenBE:
@eworm-de I updated the patch set which hopefully fixes the crash you noticed.

Also I noticed the command merging is applied to threads now to, which gives e.g. `chrome|HangWatcher` in the Command column (if merged view is active).

This is a change from the old behaviour, which can be restored by removing the negation from the `this->settings->showThreadNames` check in `Process.c:403`. But I think this is up for discussion, as showing both is IMHO more consistent with how the merged Command view functions, but showing only the thread name would be more consistent with previous (buggy IMHO) behaviour. IMO userland threads should be just as well be handled by `Process_makeCommandStr`, not by the current fallback code. We may adapt the produced view for those though …

Open for input here.

eworm-de:
Highlighting of thread basename works now, but question marks are still there. `git bisect` pinpoints commit b180cd619b0bac7baff4888b265c6a466bf11597 to be the first bad one. I did not see any crashes.

BenBE:
Off-site notes from @cgzones about an ASAN issue he observed:

```
==46657==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x606000003d12 at pc 0x000000537c43 bp 0x7ffd2b7a5320 sp 0x7ffd2b7a5318
WRITE of size 1 at 0x606000003d12 thread T0
    #0 0x537c42 in stpcpyWithNewlineConversion /home/christian/Coding/workspaces/htop/Process.c:376:12
    #1 0x53297e in Process_makeCommandStr /home/christian/Coding/workspaces/htop/Process.c:585:13
    #2 0x549dfd in ProcessList_scan /home/christian/Coding/workspaces/htop/ProcessList.c:613:7
    #3 0x5576f6 in checkRecalculation /home/christian/Coding/workspaces/htop/ScreenManager.c:109:7
    #4 0x555181 in ScreenManager_run /home/christian/Coding/workspaces/htop/ScreenManager.c:158:10
    #5 0x4e757d in CommandLine_run /home/christian/Coding/workspaces/htop/CommandLine.c:352:4
    #6 0x504dd2 in main /home/christian/Coding/workspaces/htop/htop.c:15:11
    #7 0x7f901f6efd09 in __libc_start_main csu/../csu/libc-start.c:308:16
    #8 0x427a69 in _start (/home/christian/Coding/workspaces/htop/htop+0x427a69)

0x606000003d12 is located 0 bytes to the right of 50-byte region [0x606000003ce0,0x606000003d12)
allocated by thread T0 here:
    #0 0x4a2e42 in calloc (/home/christian/Coding/workspaces/htop/htop+0x4a2e42)
    #1 0x570b5a in xCalloc /home/christian/Coding/workspaces/htop/XUtils.c:55:17
    #2 0x52f692 in Process_makeCommandStr /home/christian/Coding/workspaces/htop/Process.c:428:17
    #3 0x549dfd in ProcessList_scan /home/christian/Coding/workspaces/htop/ProcessList.c:613:7
    #4 0x5576f6 in checkRecalculation /home/christian/Coding/workspaces/htop/ScreenManager.c:109:7
    #5 0x555181 in ScreenManager_run /home/christian/Coding/workspaces/htop/ScreenManager.c:158:10
    #6 0x4e757d in CommandLine_run /home/christian/Coding/workspaces/htop/CommandLine.c:352:4
    #7 0x504dd2 in main /home/christian/Coding/workspaces/htop/htop.c:15:11
    #8 0x7f901f6efd09 in __libc_start_main csu/../csu/libc-start.c:308:16

SUMMARY: AddressSanitizer: heap-buffer-overflow /home/christian/Coding/workspaces/htop/Process.c:376:12 in stpcpyWithNewlineConversion
Shadow bytes around the buggy address:
  0x0c0c7fff8750: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c0c7fff8760: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c0c7fff8770: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c0c7fff8780: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c0c7fff8790: fa fa fa fa fa fa fa fa fa fa fa fa 00 00 00 00
=>0x0c0c7fff87a0: 00 00[02]fa fa fa fa fa 00 00 00 00 00 00 07 fa
  0x0c0c7fff87b0: fa fa fa fa 00 00 00 00 00 00 00 fa fa fa fa fa
  0x0c0c7fff87c0: 00 00 00 00 00 00 05 fa fa fa fa fa 00 00 00 00
  0x0c0c7fff87d0: 00 00 07 fa fa fa fa fa 00 00 00 00 00 00 07 fa
  0x0c0c7fff87e0: fa fa fa fa 00 00 00 00 00 00 07 fa fa fa fa fa
  0x0c0c7fff87f0: 00 00 00 00 00 00 07 fa fa fa fa fa 00 00 00 00
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
  Shadow gap:              cc
==46657==ABORTING

<<<<<<<<<< stderr output <<<<<<<<<<


FATAL PROGRAM ERROR DETECTED
============================

- Your htop version (htop --version)
- Likely steps to reproduce (How did it happened?)

Error information:
------------------
A signal 6 (Aborted) was received.

Backtrace information:
----------------------
The following function calls were active when the issue was detected:
---
./htop(backtrace+0x5b)[0x45ee0b]
./htop(CRT_handleSIGSEGV+0x1d0)[0x4f2f40]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x14140)[0x7f901f8ce140]
/lib/x86_64-linux-gnu/libc.so.6(gsignal+0x141)[0x7f901f704ce1]
/lib/x86_64-linux-gnu/libc.so.6(abort+0x123)[0x7f901f6ee537]
./htop[0x4c1147]
./htop[0x4bf671]
./htop[0x4a6bb4]
./htop[0x4a877a]
./htop(__asan_report_store1+0x2b)[0x4a913b]
./htop[0x537c43]
./htop(Process_makeCommandStr+0x388f)[0x53297f]
./htop(ProcessList_scan+0x4fe)[0x549dfe]
./htop[0x5576f7]
./htop(ScreenManager_run+0x382)[0x555182]
./htop(CommandLine_run+0x79e)[0x4e757e]
./htop(main+0x13)[0x504dd3]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xea)[0x7f901f6efd0a]
./htop(_start+0x2a)[0x427a6a]
---

To make the above information more practical to work with,
you should provide a disassembly of your binary.
This can usually be done by running the following command:

htop 3.0.6-dev aborting.

Aborted
```

Also he proposed th following patch:
```diff
diff --git a/Process.c b/Process.c
index fd95981..5dbf17c 100644
--- a/Process.c
+++ b/Process.c
@@ -500,10 +500,14 @@ void Process_makeCommandStr(Process *this) {
       }
 
       if (showProgramPath) {
-         WRITE_HIGHLIGHT(cmdlineBasenameStart, cmdlineBasenameEnd - cmdlineBasenameStart, baseAttr, CMDLINE_HIGHLIGHT_FLAG_BASENAME);
+         if (cmdlineBasenameEnd > cmdlineBasenameStart) {
+            WRITE_HIGHLIGHT(cmdlineBasenameStart, cmdlineBasenameEnd - cmdlineBasenameStart, baseAttr, CMDLINE_HIGHLIGHT_FLAG_BASENAME);
+         }
          (void)stpcpyWithNewlineConversion(str, cmdline);
       } else {
-         WRITE_HIGHLIGHT(0, cmdlineBasenameEnd - cmdlineBasenameStart, baseAttr, CMDLINE_HIGHLIGHT_FLAG_BASENAME);
+         if (cmdlineBasenameEnd > cmdlineBasenameStart) {
+            WRITE_HIGHLIGHT(0, cmdlineBasenameEnd - cmdlineBasenameStart, baseAttr, CMDLINE_HIGHLIGHT_FLAG_BASENAME);
+         }
          (void)stpcpyWithNewlineConversion(str, cmdline + cmdlineBasenameStart);
       }
```

BenBE:
@eworm-de The most recent commit of this PR now should give proper results with the question mark issue resolved. Please test.

cgzones (approved):

fasterit:
Linux:  ☑

I think this needs a bit of testing on MacOS:

Mac

fasterit:
Also seems to need some more testing on FreeBSD:

FreeBSD

cgzones:
FreeBSD crash:

```
==2542==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x602000002537 at pc 0x00000035cf54 bp 0x7fffffffe650 sp 0x7fffffffe648                                                                                                                                      
WRITE of size 1 at 0x602000002537 thread T0                                                                                                                                                                                                                                   
    #0 0x35cf53 in stpcpyWithNewlineConversion /usr/home/christian/workspace/htop/htop/Process.c:374:17                                                                                                                                                                       
    #1 0x3588b5 in Process_makeCommandStr /usr/home/christian/workspace/htop/htop/Process.c:505:13                                                                                                                                                                            
    #2 0x370509 in ProcessList_scan /usr/home/christian/workspace/htop/htop/ProcessList.c:613:7                                                                                                                                                                               
    #3 0x3055a0 in CommandLine_run /usr/home/christian/workspace/htop/htop/CommandLine.c:345:4                                                                                                                                                                                
    #4 0x325ad2 in main /usr/home/christian/workspace/htop/htop/htop.c:15:11                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                              
0x602000002537 is located 0 bytes to the right of 7-byte region [0x602000002530,0x602000002537)                                                                                                                                                                               
allocated by thread T0 here:                                                                                                                                                                                                                                                  
    #0 0x2c7af2 in calloc /usr/src/contrib/llvm-project/compiler-rt/lib/asan/asan_malloc_linux.cpp:154:3                                                                                                                                                                      
    #1 0x3992aa in xCalloc /usr/home/christian/workspace/htop/htop/XUtils.c:55:17                                                                                                                                                                                             
    #2 0x3550ee in Process_makeCommandStr /usr/home/christian/workspace/htop/htop/Process.c:429:17                                                                                                                                                                            
    #3 0x370509 in ProcessList_scan /usr/home/christian/workspace/htop/htop/ProcessList.c:613:7                                                                                                                                                                               
    #4 0x3055a0 in CommandLine_run /usr/home/christian/workspace/htop/htop/CommandLine.c:345:4                                                                                                                                                                                
    #5 0x325ad2 in main /usr/home/christian/workspace/htop/htop/htop.c:15:11                                                                                                                                                                                                  
    #6 0x27095f in _start /usr/src/lib/csu/amd64/crt1_c.c:75:7                                                                                                                                                                                                                
    #7 0x8003d9007  (<unknown module>)
```

BenBE:
Can you chack if `mc->maxLen` is set for those crashes? I noticed that the `xCalloc` HBO is likely caused my a out-of-date value there. Will likely correct this to be calculated on-the-fly instead of trusting the `mc->maxLen` field.

Process.c
```diff
@@ -253,35 +256,396 @@ void Process_fillStarttimeBuffer(Process* this) {
...
+   /* The field separtor "│" has been chosen such that it will not match any
+    * valid string used for searching or filtering */
+   const char *SEPARATOR = CRT_treeStr[TREE_STR_VERT];
+   const int SEPARATOR_LEN = strlen(SEPARATOR);
+
+   /* Check for any changed fields since we last built this string */
+   if (mc->cmdlineChanged || mc->commChanged || mc->exeChanged) {
+      free(mc->str);
+      /* Accommodate the column text, two field separators and terminating NUL */
+      mc->str = xCalloc(1, mc->maxLen + 2 * SEPARATOR_LEN + 1);
```
BenBE, line 429: This causes those ASAN overflow issues. Please check if the `mc->maxLen` value is up-to-date.

eworm-de:
> @eworm-de The most recent commit of this PR now should give proper results with the question mark issue resolved. Please test.

Running eb23edc255ecce377ed267bfb6e11a043bcf31e8 now, looks fine. Thanks!

cgzones:
With the following patch I am getting nice merged command lines on FreeBSD:

```diff
diff --git a/Process.h b/Process.h
index 0b2f198..e573207 100644
--- a/Process.h
+++ b/Process.h
@@ -46,8 +46,8 @@ typedef enum ProcessField_ {
    NLWP = 51,
    TGID = 52,
    PERCENT_NORM_CPU = 53,
-   PROC_COMM = 124,
-   PROC_EXE = 125,
+   PROC_COMM = 55,
+   PROC_EXE = 56,
 
    /* Platform specific fields, defined in ${platform}/ProcessField.h */
    PLATFORM_PROCESS_FIELDS
diff --git a/freebsd/FreeBSDProcess.c b/freebsd/FreeBSDProcess.c
index 2370b9f..4fd02a9 100644
--- a/freebsd/FreeBSDProcess.c
+++ b/freebsd/FreeBSDProcess.c
@@ -44,6 +44,8 @@ const ProcessFieldData Process_fields[LAST_PROCESSFIELD] = {
    [TIME] = { .name = "TIME", .title = "  TIME+  ", .description = "Total time the process has spent in user and system time", .flags = 0, .defaultSortDesc = true, },
    [NLWP] = { .name = "NLWP", .title = "NLWP ", .description = "Number of threads in the process", .flags = 0, .defaultSortDesc = true, },
    [TGID] = { .name = "TGID", .title = "TGID", .description = "Thread group ID (i.e. process ID)", .flags = 0, .pidColumn = true, },
+   [PROC_COMM] = { .name = "COMM", .title = "COMM            ", .description = "comm string of the process", .flags = 0, },
+   [PROC_EXE] = { .name = "EXE", .title = "EXE             ", .description = "Basename of exe of the process", .flags = 0, },
    [JID] = { .name = "JID", .title = "JID", .description = "Jail prison ID", .flags = 0, .pidColumn = true, },
    [JAIL] = { .name = "JAIL", .title = "JAIL        ", .description = "Jail prison name", .flags = 0, },
 };
diff --git a/freebsd/FreeBSDProcessList.c b/freebsd/FreeBSDProcessList.c
index ec58cec..2738e0d 100644
--- a/freebsd/FreeBSDProcessList.c
+++ b/freebsd/FreeBSDProcessList.c
@@ -377,29 +377,70 @@ static inline void FreeBSDProcessList_scanMemoryInfo(ProcessList* pl) {
    pl->usedSwap *= pageSizeKb;
 }
 
-static char* FreeBSDProcessList_readProcessName(kvm_t* kd, const struct kinfo_proc* kproc, int* basenameEnd) {
+static char* FreeBSDProcessList_getExe(const struct kinfo_proc* kproc) {
+   const int mib[] = { CTL_KERN, KERN_PROC, KERN_PROC_PATHNAME, kproc->ki_pid };
+   char buffer[2048];
+   size_t size = sizeof(buffer);
+   if (sysctl(mib, 4, buffer, &size, NULL, 0) != 0)
+      return NULL;
+
+   /* Kernel threads return an empty buffer */
+   if (buffer[0] == '\0')
+      return NULL;
+
+   return xStrdup(buffer);
+}
+
+static void FreeBSDProcessList_updateProcessName(kvm_t* kd, const struct kinfo_proc* kproc, Process* proc) {
+   if (!proc->procComm || !String_eq(proc->procComm, kproc->ki_comm)) {
+      free_and_xStrdup(&proc->procComm, kproc->ki_comm);
+      proc->mergedCommand.commChanged = true;
+   }
+
+   if (proc->procExe) {
+      const char* begin = strstr(proc->procExe, kproc->ki_comm);
+      proc->procExeBasenameOffset = begin ? (begin - proc->procExe) : 0;
+   } else {
+      proc->procExeBasenameOffset = 0;
+   }
+
    char** argv = kvm_getargv(kd, kproc, 0);
-   if (!argv) {
-      return xStrdup(kproc->ki_comm);
+   if (!argv || !argv[0]) {
+      if (!proc->cmdline || !String_eq(proc->cmdline, kproc->ki_comm)) {
+   if (!argv || !argv[0]) {
+      if (!proc->cmdline || !String_eq(proc->cmdline, kproc->ki_comm)) {
+         free_and_xStrdup(&proc->cmdline, kproc->ki_comm);
+         proc->mergedCommand.cmdlineChanged = true;
+         proc->cmdlineBasenameStart = 0;
+         proc->cmdlineBasenameEnd = strlen(kproc->ki_comm);
+      }
+      return;
    }
-   int len = 0;
+
+   size_t len = 0;
    for (int i = 0; argv[i]; i++) {
       len += strlen(argv[i]) + 1;
    }
-   char* comm = xMalloc(len);
-   char* at = comm;
-   *basenameEnd = 0;
+
+   char* cmdline = xMalloc(len);
+   char* at = cmdline;
    for (int i = 0; argv[i]; i++) {
       at = stpcpy(at, argv[i]);
-      if (!*basenameEnd) {
-         *basenameEnd = at - comm;
-      }
       *at = ' ';
       at++;
    }
    at--;
    *at = '\0';
-   return comm;
+
+   const char* begin = strstr(cmdline, kproc->ki_comm);
+   if (begin) {
+      proc->cmdlineBasenameStart = begin - cmdline;
+      proc->cmdlineBasenameEnd   = begin - cmdline + strlen(proc->procComm);
+   } else {
+      proc->cmdlineBasenameStart = 0;
+      proc->cmdlineBasenameEnd = -1;
+   }
+   free(proc->cmdline);
+   proc->cmdline = cmdline;
+   proc->mergedCommand.cmdlineChanged = true;
 }
 
 static char* FreeBSDProcessList_readJailName(const struct kinfo_proc* kproc) {
@@ -469,8 +510,8 @@ void ProcessList_goThroughEntries(ProcessList* super, bool pauseProcessUpdate) {
          proc->user = UsersTable_getRef(super->usersTable, proc->st_uid);
          ProcessList_add(super, proc);
 
-         proc->cmdline = FreeBSDProcessList_readProcessName(fpl->kd, kproc, &proc->cmdlineBasenameEnd);
-         proc->mergedCommand.cmdlineChanged = true;
+         proc->procExe = FreeBSDProcessList_getExe(kproc);
+         FreeBSDProcessList_updateProcessName(fpl->kd, kproc, proc);
 
          fp->jname = FreeBSDProcessList_readJailName(kproc);
 
@@ -497,9 +538,7 @@ void ProcessList_goThroughEntries(ProcessList* super, bool pauseProcessUpdate) {
             proc->user = UsersTable_getRef(super->usersTable, proc->st_uid);
          }
          if (settings->updateProcessNames) {
-            free(proc->cmdline);
-            proc->cmdline = FreeBSDProcessList_readProcessName(fpl->kd, kproc, &proc->cmdlineBasenameEnd);
-            proc->mergedCommand.cmdlineChanged = true;
+            FreeBSDProcessList_updateProcessName(fpl->kd, kproc, proc);
          }
       }
```

Two notes:
  * The change of the internal number of `PROC_COMM` and `PROC_EXE` is needed, cause else `LAST_PROCESSFIELD` is not getting assigned a number bigger than all process fields:
  
    ```
    freebsd/FreeBSDProcess.c:47:5: error: array designator index (124) exceeds array bounds (102)
       [PROC_COMM] = { .name = "COMM", .title = "COMM            ", .description = "comm string of the process", .flags = 0, },
        ^~~~~~~~~
    freebsd/FreeBSDProcess.c:48:5: error: array designator index (125) exceeds array bounds (102)
       [PROC_EXE] = { .name = "EXE", .title = "EXE             ", .description = "Basename of exe of the process", .flags = 0, },
        ^~~~~~~~
    ```

  * `FreeBSDProcessList_updateProcessName()` probably should be generalized to be used on all BSD platforms, maybe even on Linux.

BenBE:
Don't change the `PROC_COMM` and `PROC_EXE` IDs, as this might otherwise cause compat issues …
Maybe define a dummy field on FreeBSD for now?

cgzones:
Could you take a look at https://github.com/cgzones/htop/tree/test_388, current appended commit:
https://github.com/cgzones/htop/commit/7d4171ae6d40a7002b98cd938aa48ee1e0697196

BenBE:
@cgzones See https://github.com/BenBE/htop/pull/1

fasterit (approved):
Tested the latest command_color_instr:HEAD on MacOS again. Seems to work fine!
Good to commit to master.

PR #390, opened 2020-12-12
macOS -  add column to indicate whether the process is running under translation

bakaid:
Can be used to determine which processes are running under Rosetta 2.

Tested on arm64 Big Sur (11.0.1) with M1 processor.

### DIFF
--- a/darwin/DarwinProcess.c
+++ b/darwin/DarwinProcess.c
@@ -27,6 +27,35 @@ const ProcessClass DarwinProcess_class = {
    .writeField = Process_writeField,
 };
 
+ProcessFieldData Process_fields[] = {
+   [0] = { .name = "", .title = NULL, .description = NULL, .flags = 0, },
+   [PID] = { .name = "PID", .title = "    PID ", .description = "Process/thread ID", .flags = 0, },
+   [COMM] = { .name = "Command", .title = "Command ", .description = "Command line", .flags = 0, },
+   [STATE] = { .name = "STATE", .title = "S ", .description = "Process state (S sleeping, R running, D disk, Z zombie, T traced, W paging)", .flags = 0, },
+   [PPID] = { .name = "PPID", .title = "   PPID ", .description = "Parent process ID", .flags = 0, },
+   [PGRP] = { .name = "PGRP", .title = "   PGRP ", .description = "Process group ID", .flags = 0, },
+   [SESSION] = { .name = "SESSION", .title = "    SID ", .description = "Process's session ID", .flags = 0, },
+   [TTY_NR] = { .name = "TTY_NR", .title = "    TTY ", .description = "Controlling terminal", .flags = 0, },
+   [TPGID] = { .name = "TPGID", .title = "  TPGID ", .description = "Process ID of the fg process group of the controlling terminal", .flags = 0, },
+   [MINFLT] = { .name = "MINFLT", .title = "     MINFLT ", .description = "Number of minor faults which have not required loading a memory page from disk", .flags = 0, },
+   [MAJFLT] = { .name = "MAJFLT", .title = "     MAJFLT ", .description = "Number of major faults which have required loading a memory page from disk", .flags = 0, },
+   [PRIORITY] = { .name = "PRIORITY", .title = "PRI ", .description = "Kernel's internal priority for the process", .flags = 0, },
+   [NICE] = { .name = "NICE", .title = " NI ", .description = "Nice value (the higher the value, the more it lets other processes take priority)", .flags = 0, },
+   [STARTTIME] = { .name = "STARTTIME", .title = "START ", .description = "Time the process was started", .flags = 0, },
+
+   [PROCESSOR] = { .name = "PROCESSOR", .title = "CPU ", .description = "Id of the CPU the process last executed on", .flags = 0, },
+   [M_VIRT] = { .name = "M_VIRT", .title = " VIRT ", .description = "Total program size in virtual memory", .flags = 0, },
+   [M_RESIDENT] = { .name = "M_RESIDENT", .title = "  RES ", .description = "Resident set size, size of the text and data sections, plus stack usage", .flags = 0, },
+   [ST_UID] = { .name = "ST_UID", .title = "  UID ", .description = "User ID of the process owner", .flags = 0, },
+   [PERCENT_CPU] = { .name = "PERCENT_CPU", .title = "CPU% ", .description = "Percentage of the CPU time the process used in the last sampling", .flags = 0, },
+   [PERCENT_MEM] = { .name = "PERCENT_MEM", .title = "MEM% ", .description = "Percentage of the memory the process is using, based on resident memory size", .flags = 0, },
+   [USER] = { .name = "USER", .title = "USER      ", .description = "Username of the process owner (or user ID if name cannot be determined)", .flags = 0, },
+   [TIME] = { .name = "TIME", .title = "  TIME+  ", .description = "Total time the process has spent in user and system time", .flags = 0, },
+   [NLWP] = { .name = "NLWP", .title = "NLWP ", .description = "Number of threads in the process", .flags = 0, },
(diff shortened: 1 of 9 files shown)


Process.c
```diff
@@ -349,6 +349,7 @@ void Process_writeField(const Process* this, RichString* str, ProcessField field
       break;
    }
    case PROCESSOR: xSnprintf(buffer, n, "%3d ", Settings_cpuId(this->settings, this->processor)); break;
+   case TRANSLATED: xSnprintf(buffer, n, "%c ", this->translated ? 'T' : ' '); break;
```
BenBE, line 352: Maybe use `N` for native?

darwin/DarwinProcess.c
```diff
@@ -223,6 +223,7 @@ void DarwinProcess_setFromKInfoProc(Process* proc, const struct kinfo_proc* ps,
       /* e_tdev = (major << 24) | (minor & 0xffffff) */
       /* e_tdev == -1 for "no device" */
       proc->tty_nr = ps->kp_eproc.e_tdev & 0xff; /* TODO tty_nr is unsigned */
+      proc->translated = ps->kp_proc.p_flag & P_TRANSLATED;
```
BenBE, line 226: When was the `P_TRANSLATED` flag introduced in the headers?

bakaid: It is present since at least Mac OS X 10.5: https://opensource.apple.com/source/xnu/xnu-1228/bsd/sys/proc.h.auto.html (I suppose it was created for the first PPC/x86 Rosetta).

Process.c
bakaid: Makes sense, pushed new commit.

BenBE (approved):

Process.h
```diff
@@ -49,6 +49,7 @@ typedef enum ProcessFields {
    NLWP = 51,
    TGID = 52,
    PERCENT_NORM_CPU = 53,
+   TRANSLATED = 54,
```
cgzones, line 52: This should probably not go into the cross-platform `Process` struct, but in the darwin specific `DarwinProcess` one

Process.h
```diff
@@ -87,6 +88,7 @@ typedef struct Process_ {
    uid_t st_uid;
    unsigned long int flags;
    int processor;
+   bool translated;
```
cgzones, line 91: dito

cgzones (changes requested):

Process.h
bakaid: Since DarwinProcess was not yet made ready to writeField/compare based on DarwinProcess-specific data, this required a bit more extensive changes. Pushed.

darwin/DarwinProcess.c
```diff
@@ -194,7 +261,7 @@ static char* DarwinProcess_getCmdLine(const struct kinfo_proc* k, int* basenameO
    return retval;
 }
 
-void DarwinProcess_setFromKInfoProc(Process* proc, const struct kinfo_proc* ps, bool exists) {
+void DarwinProcess_setFromKInfoProc(DarwinProcess* proc, const struct kinfo_proc* ps, bool exists) {
```
BenBE, line 264: Canonically keep `Process* proc` and add a local variable `DarwinProcess* dp = (DarwinProcess*)proc;` directly at the start. Avoids having to indirect all the fields unnecessarily.

darwin/DarwinProcess.c
```diff
@@ -213,32 +280,32 @@ void DarwinProcess_setFromKInfoProc(Process* proc, const struct kinfo_proc* ps,
...
-      proc->tpgid = ps->kp_eproc.e_tpgid;
-      proc->tgid = proc->pid;
-      proc->st_uid = ps->kp_eproc.e_ucred.cr_uid;
+      proc->super.pid = ep->p_pid;
+      proc->super.ppid = ps->kp_eproc.e_ppid;
+      proc->super.pgrp = ps->kp_eproc.e_pgid;
+      proc->super.session = 0; /* TODO Get the session id */
+      proc->super.tpgid = ps->kp_eproc.e_tpgid;
+      proc->super.tgid = proc->super.pid;
+      proc->super.st_uid = ps->kp_eproc.e_ucred.cr_uid;
```
BenBE, line 289: See comment above.

darwin/DarwinProcess.c
```diff
@@ -22,9 +22,39 @@ const ProcessClass DarwinProcess_class = {
...
+   [PERCENT_CPU] = { .name = "PERCENT_CPU", .title = "CPU% ", .description = "Percentage of the CPU time the process used in the last sampling", .flags = 0, },
+   [PERCENT_MEM] = { .name = "PERCENT_MEM", .title = "MEM% ", .description = "Percentage of the memory the process is using, based on resident memory size", .flags = 0, },
+   [USER] = { .name = "USER", .title = "USER      ", .description = "Username of the process owner (or user ID if name cannot be determined)", .flags = 0, },
+   [TIME] = { .name = "TIME", .title = "  TIME+  ", .description = "Total time the process has spent in user and system time", .flags = 0, },
+   [NLWP] = { .name = "NLWP", .title = "NLWP ", .description = "Number of threads in the process", .flags = 0, },
+   [TGID] = { .name = "TGID", .title = "   TGID ", .description = "Thread group ID (i.e. process ID)", .flags = 0, },
+   [TRANSLATED] = { .name = "TRANSLATED", .title = "T ", .description = "Translation info (T translated, N native)", .flags = 0, },
+   [LAST_PROCESSFIELD] = { .name = "*** report bug! ***", .title = NULL, .description = NULL, .flags = 0, },
 };
```
BenBE, line 58: Why move this definition from `darwin/Platform.h`?

BenBE (changes requested):

darwin/DarwinProcess.c
bakaid: Because:
 - I need the `DarwinProcessField` enum values (in this case only `TRANSLATED`, because that is the only one so far) in `DarwinProcess.c` for implementing `DarwinProcess_writeField` and `DarwinProcess_compare`, and we do not include `Platform.h` from `DarwinProcess.c` or `DarwinProcess.h`
 - `Process_fields` is declared and defined in the respective *Process.{h,c} files in linux, freebsd and dragonflybsd. Darwin seemed to be an outlier by doing it in `Platform.h`

darwin/DarwinProcess.c
bakaid: Done.

darwin/DarwinProcess.c
BenBE: Ack. Just rechecked with the other platforms. The only other platform that has it in Platform.c is on `unsupported`. Could you split off moving the `Process_fields` variable into its own commit and add another similar change for the `unsupported` platform too (own commit preferred).

NB: Including additional headers inside the `.c` files usually isn't that complicated. Thus with the first point alone I'd preferred it without moving everything around. ;-)

bakaid: Obviously, I decided on this course based on both of the points.
PR split into 3 commits.

darwin/DarwinProcess.h
```diff
@@ -29,6 +35,10 @@ Process* DarwinProcess_new(const Settings* settings);
 
 void Process_delete(Object* cast);
 
+void DarwinProcess_writeField(const Process* this, RichString* str, ProcessField field);
+
+long DarwinProcess_compare(const void* v1, const void* v2);
+
```
BenBE, line 41: No need to have those two declarations externally visible.

You can just move the initialization of the class info structure to the bottom of `darwin/DarwinProcess.c` (compare with other platforms like `linux` in `linux/LinuxProces.c`).

bakaid: Done.

cgzones (approved):
LGTM

p.s.: thanks for all the internal code re-orderings

BenBE (approved):
LGTM.

PR #391, opened 2020-12-12
Fix dlopen issue for libsensors5 in Debian Buster, Bullseye

fasterit:
libsensors.so is provided only by the -dev package, so search for
libsensors.so.5 (installed from the libsensors5 package) explicitly

see: dpkg-query -S libsensors.so

### DIFF
--- a/linux/LibSensors.c
+++ b/linux/LibSensors.c
@@ -21,9 +21,13 @@ static void* dlopenHandle = NULL;
 
 int LibSensors_init(FILE* input) {
    if (!dlopenHandle) {
-      dlopenHandle = dlopen("libsensors.so", RTLD_LAZY);
-      if (!dlopenHandle)
-         goto dlfailure;
+      dlopenHandle = dlopen("libsensors.so.", RTLD_LAZY);
+      if (!dlopenHandle) {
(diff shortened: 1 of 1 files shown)


linux/LibSensors.c
```diff
@@ -21,9 +21,13 @@ static void* dlopenHandle = NULL;
...
-      dlopenHandle = dlopen("libsensors.so", RTLD_LAZY);
-      if (!dlopenHandle)
-         goto dlfailure;
+      dlopenHandle = dlopen("libsensors.so.", RTLD_LAZY);
+      if (!dlopenHandle) {
+      /* Debian contains no unversioned .so in libsensors5, only in the -dev package, so work around that: */
+         dlopenHandle = dlopen("libsensors.so.5", RTLD_LAZY);
+         if (!dlopenHandle)
+            goto dlfailure;
+      }
```
BenBE, line 30:
```suggestion
      /* Debian: workaround for unversioned .so for libsensors5 being present only in -dev package */
      dlopenHandle = dlopen("libsensors.so.", RTLD_LAZY);
      if (!dlopenHandle)
         dlopenHandle = dlopen("libsensors.so.5", RTLD_LAZY);
      if (!dlopenHandle)
         goto dlfailure;
```

cgzones (approved):
LGTM

Issue #384, opened 2020-12-11
Linux: read sysconf(_SC_CLK_TCK) on startup and fail loudly


amomchilov:
Currently it's lazily loaded and fails quietly by assuming a value of 100.

https://github.com/htop-dev/htop/blob/2700d9906959a462e753c01c42b9251b7b3b0b8b/linux/LinuxProcessList.c#L276-L287

If the read fails, something is already off. There's no reason to think 100 would be a reasonable default in such a context. Instead, we should crash and fail loudly.

Follow up from: https://github.com/htop-dev/htop/pull/381#discussion_r540682849

PR #396, opened 2020-12-14
Cleanup initialization of jiffies on the Linux platform

natoscott:
Small cleanups - add error handling, remove a local static
variable and refactor LinuxProcess_adjustTime (also rename
it, as its in LinuxProcessList.c not LinuxProcess.c) - and
while there, move the related 'btime' global variable into
LinuxProcessList.c so it can be made static.

Resolves https://github.com/htop-dev/htop/issues/384

### DIFF
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -63,6 +63,11 @@ in the source distribution for its full text.
 # define O_PATH 010000000
 #endif
 
+static long long btime;
+
+static long jiffy;
+
+
 static FILE* fopenat(openat_arg_t openatArg, const char* pathname, const char* mode) {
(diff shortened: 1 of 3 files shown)


linux/LinuxProcessList.c
```diff
@@ -63,6 +63,11 @@ in the source distribution for its full text.
 # define O_PATH 010000000
 #endif
 
+static long long btime;
+
+static long jiffy;
+
+
 static FILE* fopenat(openat_arg_t openatArg, const char* pathname, const char* mode) {
```
BenBE, line 71: No need for second blank here …

BenBE (approved):

cgzones:
Applied via a3db2da4

Issue #402, opened 2020-12-15
Setup screen does not handle KEY_RESIZE (changed window size)


fasterit:
Reproducer:

- Press F2
- Change Window size

--> content lost / footer (only fixed after next key press but not up  / down in any of the select lists)

PR #403, opened 2020-12-15
ScreenManager: redraw current FunctionBar on resize

cgzones:
Closes: #402

### DIFF
--- a/CommandScreen.c
+++ b/CommandScreen.c
@@ -17,9 +17,10 @@ static void CommandScreen_scan(InfoScreen* this) {
    Panel_prune(panel);
 
    const char* p = Process_getCommand(this->process);
-   char* line = xMalloc(COLS + 1);
+   char line[COLS + 1];
    int line_offset = 0, last_spc = -1, len;
    for (; *p != '\0'; p++, line_offset++) {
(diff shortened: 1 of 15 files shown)


BenBE (approved):

BenBE:
Resizing in the DisplayOptions panel fixed with this patch applied.

cgzones:
Currently this breaks the FunctionBar when using Filter or Search mode.
I'll rethink the whole FunctionBar drawing logic...

CommandScreen.c
```diff
@@ -17,7 +17,7 @@ static void CommandScreen_scan(InfoScreen* this) {
    Panel_prune(panel);
 
    const char* p = Process_getCommand(this->process);
-   char* line = xMalloc(COLS + 1);
+   char line[COLS + 1];
    int line_offset = 0, last_spc = -1, len;
    for (; *p != '\0'; p++, line_offset++) {
       line[line_offset] = *p;
```
BenBE, line 23: Add extra `assert(line_offset < COLS + 1)` as cautionary measure when converting this to a stack buffer.

BenBE:
Drawn this way the function bar looks kinda weird:
image
Also with large fonts the two options for Del and F10 are pushed out of the display area.

PR #404, opened 2020-12-15
Meter: fix artifacts with very tiny width

cgzones:
- The Bar Meter might override the right border
- The TextMeter might wrap-around into the next line

### DIFF
--- a/Meter.c
+++ b/Meter.c
@@ -156,16 +156,20 @@ ListItem* Meter_toListItem(Meter* this, bool moving) {
 static void TextMeterMode_draw(Meter* this, int x, int y, int w) {
    char buffer[METER_BUFFER_LEN];
    Meter_updateValues(this, buffer, sizeof(buffer));
-   (void) w;
 
    attrset(CRT_colors[METER_TEXT]);
-   mvaddstr(y, x, this->caption);
+   mvaddnstr(y, x, this->caption, w - 1);
+   attrset(CRT_colors[RESET_COLOR]);
+
    int captionLen = strlen(this->caption);
    x += captionLen;
-   attrset(CRT_colors[RESET_COLOR]);
+   w -= captionLen;
+   if (w <= 0)
+      return;
+
    RichString_begin(out);
    Meter_displayBuffer(this, buffer, &out);
-   RichString_printVal(out, y, x);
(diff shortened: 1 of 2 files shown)


BenBE:
When zooming in … really far.
image
image

While zooming this flickers into view briefly:
image

Minor detail: The first column seems to be kept unused inside the meter:
image

BenBE:
Redraw issue still exists in DisplayOption panel:
image
Just zoom in and one step out again to get this beauty of screen artifact … ;-)

cgzones:
At a second thought I am not sure if this is of any value:
The main points are: htop does not crash and there are no artefacts after the zoom/font-size is back to normal (which currently is already fulfilled).

BenBE:
Agree. Just documenting those really extreme cases. No real need to fix them (apart from the initial observation of text overflowing other stuff.

That being said, the only things I think we should care about (apart from not crashing) is things being layed out in proper order (thus `  ][` should not happen (instead it should probably just skip to `  0-`. Similar the `  0[|25]` case might be better of utilizing the first column of the meter too (`  0[25.]`, or even `  0[25%]`). Furthermore it might be an option to skip certain parts of the string if things leave to little space (`  0[|||25.5% 800MHz 42°C]` --> `  0[||25.5% 800MHz]` --> `  0[||||25.5%]`).

We could even argue to skip drawing the header if we have too little space and just display the process list if below (40x10). Or some static error text ("too little space") …

Remains the question, where the `Mq` comes from when zooming in?

BenBE:
The padding between columns is still overwritten by text meters:
image

Meter.c
```diff
@@ -156,16 +156,20 @@ ListItem* Meter_toListItem(Meter* this, bool moving) {
 static void TextMeterMode_draw(Meter* this, int x, int y, int w) {
    char buffer[METER_BUFFER_LEN];
    Meter_updateValues(this, buffer, sizeof(buffer));
-   (void) w;
 
    attrset(CRT_colors[METER_TEXT]);
-   mvaddstr(y, x, this->caption);
+   mvaddnstr(y, x, this->caption, w);
```
BenBE, line 161: To address the issue in the screenshot probably also needs:
```suggestion
   mvaddnstr(y, x, this->caption, w - 1);
```

BenBE (approved):
LGTM.

PR #413, opened 2020-12-19
Added keybind 'N' for sorting by PID

jakem72360:
`N` key was selected, as it seems to be free and is backwards compatible with `top`.


Action.c
```diff
@@ -450,22 +454,22 @@ static const struct {
...
+   { .key = "       m: ", .info = "toggle merged command" },
+   { .key = "       Z: ", .info = "pause/resume process updates" },
+   { .key = "       u: ", .info = "show processes of a single user" },
+   { .key = "       H: ", .info = "hide/show user process threads" },
+   { .key = "       K: ", .info = "hide/show kernel threads" },
+   { .key = "       F: ", .info = "cursor follows process" },
+   { .key = "  F6 + -: ", .info = "expand/collapse tree" },
+   { .key = " N P M T: ", .info = "sort by PID, CPU%, MEM% or TIME" },
+   { .key = "       I: ", .info = "invert sort order" },
+   { .key = "  F6 > .: ", .info = "select sort column" },
```
BenBE, line 472: Resizing the key column causes too much unnecessary noise in the patch.

Action.c
```diff
@@ -564,24 +568,24 @@ static Htop_Reaction actionHelp(State* st) {
...
-      if (String_eq(helpLeft[item].key, "      H: ")) {
+      if (String_eq(helpLeft[item].key, "       H: ")) {
          attrset(CRT_colors[PROCESS_THREAD]);
-         mvaddstr(line + item, 32, "threads");
-      } else if (String_eq(helpLeft[item].key, "      K: ")) {
+         mvaddstr(line + item, 33, "threads");
+      } else if (String_eq(helpLeft[item].key, "       K: ")) {
          attrset(CRT_colors[PROCESS_THREAD]);
-         mvaddstr(line + item, 26, "threads");
+         mvaddstr(line + item, 27, "threads");
```
BenBE, line 579: Further resize-related noise.

Action.c
```diff
@@ -564,24 +568,24 @@ static Htop_Reaction actionHelp(State* st) {
...
    }
    int leftHelpItems = item;
 
    for (item = 0; helpRight[item].key; item++) {
       attrset(CRT_colors[HELP_BOLD]);
-      mvaddstr(line + item, 40, helpRight[item].key);
+      mvaddstr(line + item, 41, helpRight[item].key);
       attrset(CRT_colors[DEFAULT_COLOR]);
-      mvaddstr(line + item, 49, helpRight[item].info);
+      mvaddstr(line + item, 50, helpRight[item].info);
```
BenBE, line 588: dito.

Action.c
```diff
@@ -450,22 +454,22 @@ static const struct {
...
+   { .key = "    F4 \\: ",.info = "incremental name filtering" },
+   { .key = "    F5 t: ", .info = "tree view" },
+   { .key = "       p: ", .info = "toggle program path" },
+   { .key = "       m: ", .info = "toggle merged command" },
+   { .key = "       Z: ", .info = "pause/resume process updates" },
+   { .key = "       u: ", .info = "show processes of a single user" },
+   { .key = "       H: ", .info = "hide/show user process threads" },
+   { .key = "       K: ", .info = "hide/show kernel threads" },
+   { .key = "       F: ", .info = "cursor follows process" },
+   { .key = "  F6 + -: ", .info = "expand/collapse tree" },
```
BenBE, line 469: <kbd>F6</kbd> is no more.

BenBE (changes requested):

jakem72360:
What would your suggestion be? Adding the keybind to `key` in `helpLeft` would violate the leading space. Adding more padding causes the info section of the line to be cut off. Hardcoding coordinates is flimsy in general, and this seemed like the most logical solution.

I'm happy to change the patch, and am open to suggestions...

BenBE:
Instead of adding spaces everywhere I think it'd be less noise, if just the horizontal offset of everything was shifted to the right by one. The entry you need to actually update has two spaces at the front, thus you could fit it in there.

An alternative was to split the PMT line entirely (One function per line) and shuffle things from the left to the right (or vice versa) and add the N shortcut in a second commit.

jakem72360:
I've updated it with changes only to the hardcoded X positions. I had assumed the leading spaces were desired, as they were present in both arrays.

Regarding splitting the sort functions onto multiple lines, I'll leave that decision to you. I think it looks better being one character wider, than a column being three lines longer. But like I said, it's up to you.

fasterit:
see `Action.c` a bit below your inserted line:
```
   keys['n'] = actionIncNext;
   keys['N'] = actionIncPrev;

```
Either we remove these bindings or we need to find something else to map to PID sorting.

jakem72360:
I just noticed that. What do you think is best? If we remove the original bindings, we can preserve some compatibility with `top`. If we keep those bindings, I'd suggest the `D` key as it's the only letter in PID not already in use. For now, I'll update the PR with the `D` binding.

fasterit:
> I just noticed that. What do you think is best?
I think we should remove the old n / N bindings and the stale code they call.
I asked on IRC what fellow devs think about this. That would free the "N" key then for PID sorting.
o/ DLange

jakem72360:
> I think we should remove the old n / N bindings and the stale code they call.
> I asked on IRC what fellow devs think about this. That would free the "N" key then for PID sorting.
> o/ DLange

Created #425 with proposal for code removal

fasterit:
Merged the patch (fixed back to using the N key) in the series up to f1463fd.

Thank you very much!

Issue #361, opened 2020-12-02
Crash in LinuxProcess_makeCommandStr()


cgzones:
Crash on a Raspberry PI 3 with Debian sid:

```
FATAL PROGRAM ERROR DETECTED
============================
Please check at https://htop.dev/issues whether this issue has already been reported.
If no similar issue has been reported before, please create a new issue with the following information:

- Your htop version (htop --version)
- Your OS and kernel version (uname -a)
- Your distribution and release (lsb_release -a)
- Likely steps to reproduce (How did it happened?)
- Backtrace of the issue (see below)

Error information:
------------------
A signal 11 (Segmentation fault) was received.

Backtrace information:
----------------------
The following function calls were active when the issue was detected:
---
/root/workspace/htop/htop_rpi3(CRT_handleSIGSEGV+0x98)[0xaaaad806dcc8]
linux-vdso.so.1(__kernel_rt_sigreturn+0x0)[0xffffa2bfb7bc]
/root/workspace/htop/htop_rpi3(LinuxProcess_makeCommandStr+0x518)[0xaaaad808ae98]
/root/workspace/htop/htop_rpi3(+0x33064)[0xaaaad8090064]
/root/workspace/htop/htop_rpi3(ProcessList_goThroughEntries+0x7ac)[0xaaaad808e050]
/root/workspace/htop/htop_rpi3(+0x258cc)[0xaaaad80828cc]
/root/workspace/htop/htop_rpi3(ScreenManager_run+0x100)[0xaaaad8082bb4]
/root/workspace/htop/htop_rpi3(main+0x610)[0xaaaad806bb60]
/lib/aarch64-linux-gnu/libc.so.6(__libc_start_main+0xe8)[0xffffa2915198]
/root/workspace/htop/htop_rpi3(+0xed20)[0xaaaad806bd20]
---

To make the above information more practical to work with,
you should provide a disassembly of your binary.
This can usually be done by running the following command:

   objdump -d -S -w `which htop` > ~/htop.objdump

Please include the generated file in your report.

Running this program with debug symbols or inside a debugger may provide further insights.

Thank you for helping to improve htop!

htop 3.0.3-dev aborting.

!!! Chained handler did not exit. Forcing exit.
```

Compiled with `-O3 -flto=4 -march=native -ffast-math`

Will try to reproduce with `-g`.

[htop.objdump.txt](https://github.com/htop-dev/htop/files/5630919/htop.objdump.txt)

legendtang:
[...]

PR #417, opened 2020-12-21
Harden makeCommandStr

BenBE:
This should resolve the NULL pointer dereference issues seen on Raspberry Pi as noted in #361.

The second patch also adds 4 additional in-depth checks to ensure we are always pointing inside the strings we process. These are not strictly necessary, but are defense-in-depth in case something else went wrong when reading these information.


natoscott (approved):
@BenBE LGTM.

Issue #421, opened 2020-12-23
CJK characters break column alignment


Serentty:
My Linux username is 世蓮. This shows up properly (as expected, since it's the terminal's job to render the characters), but the columns don't align properly. I'm guessing this easier has to do with not accounting for two-column characters when aligning columns, or counting columns as bytes (since my username is six bytes long).
![Screenshot_20201222_200502](https://user-images.githubusercontent.com/614548/102948566-2dd95b00-4494-11eb-967a-0660cb5089b0.png)

Serentty:
Thank you!

PR #422, opened 2020-12-23
Process: use correct column aligning on wide username

cgzones:
Closes: #421


fasterit (changes requested):
Breaks non-unicode build:
image

cgzones:
> Breaks non-unicode build

Thanks for testing; fixed.

fasterit (approved):

PR #423, opened 2020-12-23
fix argument type following prototype change in "Invert Process_compa…

hadfl:
…re resolution so that superclass matches run first"

function prototype has been changed recently:
https://github.com/htop-dev/htop/commit/3d1703f16faf5bd3c73976909e1b6e03061a7f72#diff-b44874a4f5ef90eadea51feae4fe62ec855f1280d4e4e20106ecc719fc1dd6ddL59-R59


cgzones:
Thank you for the contribution.

PR #427, opened 2020-12-23
Enable going back to previous search matches (Shift-F3)

fasterit:


BenBE (approved):
LGTM. Maybe use S-F3 to keep key labels short?

fasterit:
> LGTM. Maybe use S-F3 to keep key labels short?

Yeah, I thought about that as well. Not that we don't have enough space but because of the aesthetics.

Issue #32, opened 2020-08-29
System load/Pressure meters do not display values correctly.


pthfdr-42:
If I set the PSI or System load meters' style to "bar", they show the 3 values stacked together, which is ridiculous.
Also, all 3 PSI stats are labelled "Som" or "Ful".

My recommendations:

For Load average, show a green bar from 0 to 1, a yellow bar over it from 1 to physical CPU count, and finally a red bar over it when the load is between 1x to 2x of physical CPU count.
Mockup:
![w](https://user-images.githubusercontent.com/42509223/91639041-484f7a80-ea46-11ea-8a51-9f312f5c3824.png)

For PSI labelling:
S-C (Some CPU), S-M (Some Memory), S-I (Some I/O), F-M (Full Memory) and F-I (Full I/O).

cgzones:
The pressure stall meter got improved with https://github.com/htop-dev/htop/commit/1d8192c10b4e149c7f9126f00534f6a8488d10c4 and https://github.com/htop-dev/htop/commit/ded9c5d36332401bb8d5e33b6f72de188a42560d.

PR #428, opened 2020-12-23
LoadMeter: dynamically adjust color and total of Bar 

cgzones:
Change the color and total based on the actual 1min load value:

    < 1         : green and total of 1.0
    < cpu-count : yellow and total of cpu-count
    else        : read and total of 2*cpu-count

Closes: #32


pthfdr-42:
This should be used on graph as well.

BenBE:
Just a minor typo fix in the commit message.

BenBE (approved):
LGTM.

eworm-de:
This is kind of inconsistent IMHO.
Probably depends on the type of machine (desktop/notebook vs. server), but for green I would use `< cpu-count/2`.

BenBE:
IMHO this is fine this way, as `cpu-count` is always `>1` which isn't guaranteed for `cpu-count/2`.

Semantically the current limits mean, that for green there will always be some time where the whole CPU could be sleeping/idling. With yellow, there is potentially always one task for execution (thus likely no CPU idle time) and with read you're likely to start experiencing symptoms of a busy system.

PR #431, opened 2020-12-24
add support to display CPU frequencies on Solarish platforms

hadfl:
tested on OmniOS r151036 (illumos based distribution) and Solaris 11.4

### DIFF
--- a/solaris/SolarisProcessList.c
+++ b/solaris/SolarisProcessList.c
@@ -72,11 +72,11 @@ static inline void SolarisProcessList_scanCPUTime(ProcessList* pl) {
    const SolarisProcessList* spl = (SolarisProcessList*) pl;
    int cpus = pl->cpuCount;
    kstat_t* cpuinfo = NULL;
-   int kchain = 0;
    kstat_named_t* idletime = NULL;
    kstat_named_t* intrtime = NULL;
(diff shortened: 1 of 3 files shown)


solaris/SolarisProcessList.c
```diff
@@ -71,12 +71,13 @@ ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* pidMatchList, ui
 static inline void SolarisProcessList_scanCPUTime(ProcessList* pl) {
    const SolarisProcessList* spl = (SolarisProcessList*) pl;
    int cpus = pl->cpuCount;
-   kstat_t* cpuinfo = NULL;
-   int kchain = 0;
+   kstat_t* cpu = NULL;
```
fasterit, line 74: Does this need distinct cpu and cpu_info or can it be folded into one?
(They look essentially the same in the stat_lookup calls below.)

hadfl: the same variable can certainly be used; the aim to use distinct ones was to have the variable name match the kstat module.

changed the variable name back to `cpuinfo` and using it for both kstat lookups.

fasterit: Great, thank you very much.
Merry Christmas, may the Sun always shine :-)

Issue #243, opened 2020-10-11
Preserve colors/highlighting for highlighted lines


BenBE:
While reviewing #240 it became quite visible, that when highlighting lines (e.g. with space, or with the newly added highlighting for new/dead processes) the markup of the line is dropped. With only a few lines affected with the search or manually selected processes this isn't a big issue, but when highlighting of lines becomes more visible, so becomes the dropped formatting for these lines.

Thus we should build some mechanism to add highlighting for a line AND keep it's formatting apart from the added highlighting.

BenBE:
Idea for a patch:

```diff
diff --git a/Panel.c b/Panel.c
index fd9de2b..831452f 100644
--- a/Panel.c
+++ b/Panel.c
@@ -278,7 +278,7 @@ void Panel_draw(Panel* this, bool focus) {
          }
          if (item.highlightAttr) {
             attrset(item.highlightAttr);
-            RichString_setAttr(&item, item.highlightAttr);
+            RichString_setAttr_preserveBold(&item, item.highlightAttr);
             this->selectedLen = itemLen;
          }
          mvhline(y + line, x, ' ', this->w);
diff --git a/RichString.c b/RichString.c
index 904b44b..1a0d6ce 100644
--- a/RichString.c
+++ b/RichString.c
@@ -69,6 +69,15 @@ inline void RichString_setAttrn(RichString* this, int attrs, int start, int fini
    }
 }
 
+inline void RichString_setAttrn_preserveBold(RichString* this, int attrs, int start, int finish) {
+   cchar_t* ch = this->chptr + start;
+   finish = CLAMP(finish, 0, this->chlen - 1);
+   for (int i = start; i <= finish; i++) {
+      ch->attr = (ch->attr & A_BOLD) ? (attrs | A_BOLD) : attrs;
+      ch++;
+   }
+}
+
 int RichString_findChar(RichString* this, char c, int start) {
    wchar_t wc = btowc(c);
    cchar_t* ch = this->chptr + start;
@@ -100,6 +109,15 @@ void RichString_setAttrn(RichString* this, int attrs, int start, int finish) {
    }
 }
 
+void RichString_setAttrn_preserveBold(RichString* this, int attrs, int start, int finish) {
+   chtype* ch = this->chptr + start;
+   finish = CLAMP(finish, 0, this->chlen - 1);
+   for (int i = start; i <= finish; i++) {
+      *ch = (*ch & 0xff) | attrs | (*ch & A_BOLD);
[...]

PR #434, opened 2020-12-25
Keep the bold attribute when highlighting selections

BenBE:
Based on patch from #243.


cgzones:
The colour change from black to some high-intensity one is unfortunate (I thought it would stay raw black).

PR #437, opened 2020-12-28
Show arrow indicating order of sorted process column

cgzones:


CRT.c
```diff
@@ -48,6 +48,8 @@ static const char* const CRT_treeStrAscii[LAST_TREE_STR] = {
    [TREE_STR_TEND] = ",",
    [TREE_STR_OPEN] = "+",
    [TREE_STR_SHUT] = "-",
+   [TREE_STR_ASC]  = " ",
+   [TREE_STR_DESC] = " ",
```
BenBE, line 52: Use `+` / `-` here ?

CRT.c
```diff
@@ -61,6 +63,8 @@ static const char* const CRT_treeStrUtf8[LAST_TREE_STR] = {
                                      // WITH VERTICAL STROKE' (U+1FBAF, "\xf0\x9f\xae\xaf") when
                                      // Unicode 13 is common
    [TREE_STR_SHUT] = "\xe2\x94\x80", // ─
+   [TREE_STR_ASC]  = "\xf0\x9f\xa2\x91", // 🢑
+   [TREE_STR_DESC] = "\xf0\x9f\xa2\x93", // 🢓
```
BenBE, line 67: Use proper arrows instead of just the triangles?

cgzones:
My intention was to only add a small hint for the sorting order, that's why I didn't add an ascii variant (`+`/`-` or `^`/`v` or ...) and used the tiny triangles instead of any big arrows.

fasterit:
> My intention was to only add a small hint for the sorting order, that's why I didn't add an ascii variant (`+`/`-` or `^`/`v` or ...) and used the tiny triangles instead of any big arrows.

Still degrading gracefully (think -U case) is much better than using blanks, I'd be for "+ / -" as that is common.

cgzones:
Added `+`/`-` for non-unicode mode.
What about the unicode characters?
Personally I prefer the small triangle [`🢑`](https://unicode-table.com/de/1F891/), or does the majority like a taller arrow, like [`🠑` ](https://unicode-table.com/de/1F811/)?

fasterit:
> Personally I prefer the small triangle [`🢑`](https://unicode-table.com/de/1F891/), or does the majority like a taller arrow, like [`🠑` ](https://unicode-table.com/de/1F811/)?

Same for me, the indication for sorting is typically + /- or up / down triangles, not arrows.
There's many more triangles, may be you want to consider △ and ▽ which are more visually distinct than your fly stain :-). Cf. https://www.fileformat.info/info/unicode/block/geometric_shapes/list.htm

cgzones:
Now using [`△`](https://unicode-table.com/de/25B3/) and [`▽`](https://unicode-table.com/de/25BD/).

BenBE (approved):

ProcessList.c
```diff
@@ -112,6 +112,13 @@ void ProcessList_printHeader(ProcessList* this, RichString* header) {
       }
 
       RichString_appendWide(header, color, alignedProcessFieldTitle(fields[i]));
+      if (key == fields[i] && RichString_getCharVal(*header, RichString_size(header) - 1) == ' ') {
+         header->chlen--;  // rewind to override space
+         RichString_appendnWide(header,
+                                CRT_colors[PANEL_SELECTION_FOCUS],
+                                CRT_treeStr[settings->direction == 1 ? TREE_STR_DESC : TREE_STR_ASC],
```
fasterit, line 119:
```suggestion
                                CRT_treeStr[Settings_getActiveDirection(this->settings) == 1 ? TREE_STR_DESC : TREE_STR_ASC],
```
Won't work otherwise (press I to confirm)

fasterit (changes requested):
The sort indicator won't update on I as is, pls. see suggested code change.
Otherwise looks good.

ProcessList.c
fasterit: Committed in a61f0ce

fasterit (approved):
Good to get merged now

PR #438, opened 2020-12-28
Only initialize and gather delay accounting data if a related column is enabled

cgzones:
Avoid creating and communicating over a netlink socket by default, which
triggers cap_net_admin checks as root.


BenBE (approved):
LGTM.

Issue #168, opened 2020-09-22
Feature request: Add a display option to hide header and footer


mskar:
Please add an option to hide the header and footer.

What I see:

![image](https://user-images.githubusercontent.com/13444106/93837245-5e221b00-fc53-11ea-9867-5f27b6aa8dbe.png)

What I want to see (header and footer hidden):
![image](https://user-images.githubusercontent.com/13444106/93837278-772acc00-fc53-11ea-9646-603b02278f6e.png)

fasterit:
Remove the meters from the left and right columns and disable "Leave a margin around header".
You'll have the top empty as you want it and just have to live with the footer for now.

BenBE:
If I see it correctly we could go about this by not adding the header and footer to the set of blocks to render in the main display routine. I'll take a look at this if I can build a patch for this. Shouldn't be too hard.

A quick&dirty change to get rid of the header is this patch:
```diff
diff --git a/Header.c b/Header.c
index d2f3b88..3e3b6a1 100644
--- a/Header.c
+++ b/Header.c
@@ -156,6 +156,8 @@ void Header_reinit(Header* this) {
 }
 
 void Header_draw(const Header* this) {
+   return;
+
    int height = this->height;
    int pad = this->pad;
    attrset(CRT_colors[RESET_COLOR]);
@@ -177,6 +179,8 @@ void Header_draw(const Header* this) {
 }
 
 int Header_calculateHeight(Header* this) {
+   return 0;
+
    int pad = this->settings->headerMargin ? 2 : 0;
    int maxHeight = pad;
 
```

This doesn't handle the case of the function bar yet, as the handling of the function bar is a bit more distributed across the code base.

While looking at the code base I noticed that vertical alignment of panels is not yet implemented; a change that would make implementation of this change a bit easier as this really would just be hiding the function bar under some circumstances. Unfortunately hiding the function bar makes the overall screen layout sensitive to keyboard input as entering incremental search mode would have to display the function bar and thus cause a resize of all panels to be triggered.

fasterit:
I think we can just leave the footer as is. No need to be able to configure everything and add that much complexity as a result.

BenBE:
Which leaves deciding on a keyboard shortcut for it (no need for CLI option nor permanent setting IMHO).

fasterit:
not either
esp. as we disable / enable UI elements via settings and not via keyboard shortcuts
consistency++

BenBE:
[...]

PR #440, opened 2021-01-01
Add possibility to hide function bar in MainPanel

cgzones:
Also fix the selection-cursor currently getting hidden by the FunctionBar when scrolling down.

### DIFF
--- a/Panel.c
+++ b/Panel.c
@@ -208,7 +208,7 @@ void Panel_splice(Panel* this, Vector* from) {
    this->needsRedraw = true;
 }
 
-void Panel_draw(Panel* this, bool force_redraw, bool focus, bool highlightSelected) {
+void Panel_draw(Panel* this, bool force_redraw, bool focus, bool highlightSelected, bool hideFunctionBar) {
    assert (this != NULL);
 
    int size = Vector_size(this->items);
@@ -216,6 +216,9 @@ void Panel_draw(Panel* this, bool force_redraw, bool focus, bool highlightSelect
    int y = this->y;
    int x = this->x;
    int h = this->h;
+
+   if (hideFunctionBar)
+      h++;
 
    const int header_attr = focus
                          ? CRT_colors[PANEL_HEADER_FOCUS]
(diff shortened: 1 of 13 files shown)


BenBE:
Typo in commit message:
hight -> height

Panel.c
```diff
@@ -336,10 +339,14 @@ void Panel_draw(Panel* this, bool force_redraw, bool focus, bool highlightSelect
...
+   const int size = Vector_size(this->items);
 
-   #define CLAMP_INDEX(var, delta, min, max) \
-      CLAMP((var) + (delta), (min), MAXIMUM(0, (max)))
+   #define PANEL_SCROLL(amount)                                                                                                                                      \
+   do {                                                                                                                                                              \
+      this->selected += (amount);                                                                                                                                    \
+         this->scrollV = CLAMP(this->scrollV + (amount), 0, MAXIMUM(0, (size - (this->h + this->hightModifier) - (RichString_sizeVal(this->header) > 0 ? 1 : 0))));  \
+         this->needsRedraw = true;                                                                                                                                   \
+      } while (0)
```
BenBE, line 349: indentation

BenBE (changes requested):
Overall typo: `hight` -> `height`

fasterit:
Please make "Esc" toggle the function bar (and probably also the selection cursor bar).
Otherwise this is odd from a usability perspective.

People will very soon ask for a setting to hide the Function bar permanently if this PR gets merged, cf.  #168.
So you should probably implement that at the same time.

I'd personally prefer a simple setting to this Esc logic which may annoy the people that just want to get rid of the selection bar but keep those unsatisfied that want to always get rid of the Function bar.

BenBE:
I'd suggest not joining the ESC hiding of the cursor and the overall function bar hiding. IMO these are different things and should be kept separate. Based on demand in #168 I'd even argue that warrants a config option for the function bar hiding; though the ESC behaviour should be kept unimpeded by that setting.

BenBE (approved):
LGTM.

PR #442, opened 2021-01-01
Hashtable: misc tweaks

cgzones:
- use more distinct typename for key type
- introduce typename for size type
- fail hard on too big size request
- hide implementation of Hashtable and HashtableItem

### DIFF
--- a/Hashtable.c
+++ b/Hashtable.c
@@ -21,14 +21,14 @@ in the source distribution for its full text.
 
 typedef struct HashtableItem_ {
    ht_key_t key;
-   unsigned int probe;
+   size_t probe;
    void* value;
 } HashtableItem;
 
 struct Hashtable_ {
-   unsigned int size;
+   size_t size;
    HashtableItem* buckets;
-   unsigned int items;
+   size_t items;
    bool owner;
 };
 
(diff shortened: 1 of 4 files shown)


BenBE (commented):
LGTM so far.

Any particular reason for not using `uint64_t` for `ht_size_t`?

cgzones:
> Any particular reason for not using uint64_t for ht_size_t?

 I didn't want to risk a slow-down on 32bit architectures; switched to `size_t`.

Hashtable.c
```diff
@@ -80,18 +94,16 @@ static const uint64_t OEISprimes[] = {
    34359738337, 68719476731, 137438953447
 };
 
-static uint64_t nextPrime(unsigned int n) {
-   assert(n <= OEISprimes[ARRAYSIZE(OEISprimes) - 1]);
```
BenBE, line 84: Why not keep this assert?

Hashtable.c
```diff
@@ -194,15 +206,19 @@ void Hashtable_setSize(Hashtable* this, unsigned int size) {
...
+void Hashtable_put(Hashtable* this, ht_key_t key, void* value) {
 
    assert(Hashtable_isConsistent(this));
    assert(this->size > 0);
    assert(value);
 
    /* grow on load-factor > 0.7 */
-   if (10 * this->items > 7 * this->size)
+   if (10 * this->items > 7 * this->size) {
+      if (((ht_size_t)-1) / 2 < this->size)
```
BenBE, line 217: Actually would also need to check for last index of OEIS prime array.

BenBE (commented):
Splitting pure type changes and actual code tweaks would be easier for review.

cgzones:
Split the changes into multiple commits and reworked the logic for reaching to big primes.

BenBE (approved):
LGTM.

PR #444, opened 2021-01-01
Object: return int on comparison

cgzones:
Comparisons do, due to the new introduced shaceship-comparisons,
only return -1, 0, 1 or the result of strcmp().

### DIFF
--- a/ListItem.c
+++ b/ListItem.c
@@ -57,7 +57,7 @@ void ListItem_append(ListItem* this, const char* text) {
    this->value[newLen] = '\0';
 }
 
-static long ListItem_compare(const void* cast1, const void* cast2) {
+static int ListItem_compare(const void* cast1, const void* cast2) {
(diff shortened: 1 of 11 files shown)


BenBE (commented):
Do we have any cases where the result of `strcmp` is returned and where this can not be replaced to only return `-1`, `0`, `1` instead (or where doing so would be diametral to performance)?

cgzones:
The only current calls of the Object compare functions are in Vector.c:
https://github.com/htop-dev/htop/blob/293c16e22da7fbda98eae1836f11e642fcbd699b/Vector.c#L116
https://github.com/htop-dev/htop/blob/293c16e22da7fbda98eae1836f11e642fcbd699b/Vector.c#L164
https://github.com/htop-dev/htop/blob/293c16e22da7fbda98eae1836f11e642fcbd699b/Vector.c#L322

So we only care about `== 0`, `> 0` and `< 0` (and combinations of those), and not exact values.

PR #446, opened 2021-01-02
Fix CPU percentage on M1 silicon Macs

### LINKED ISSUE #368: CPU usage incorrect on macOS Big Sur with M1 chip

dghost:
Explanation in [this post](https://github.com/htop-dev/htop/issues/368#issuecomment-753439724), but TL;DR: CPU percentage was being divided by Platform_timebaseToNS instead of multiplied.

Resolves #368

### DIFF
--- a/darwin/DarwinProcessList.c
+++ b/darwin/DarwinProcessList.c
@@ -161,7 +161,7 @@ void ProcessList_delete(ProcessList* this) {
 
 static double ticksToNanoseconds(const double ticks) {
    const double nanos_per_sec = 1e9;
-   return ticks / (double) Platform_clockTicksPerSec * Platform_timebaseToNS * nanos_per_sec;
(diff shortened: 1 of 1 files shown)


BenBE (approved):

cgzones:
Thanks @dghost for your analysis and the patch. Applied.

sharewax:
Sorted by CPU, but it doesn't shows it correctly which process is taken CPU:
выява
I've builded rust from mac ports. 
выява

BenBE:
Given this PR only touched one formula I'd assume this breakage resulted from somewhere else. Please continue discussion on this process state breakage over in the issue I split out from this. TIA.

PR #449, opened 2021-01-05
Linux: report free huge pages as free memory

davezarzycki:
Reserved huge pages are taken out of the pool of normal memory and therefore are not "free" or "available" to normal applications. This gives the mistaken illusion that all reserved huge pages are in use when they might not be.


davezarzycki:
This can be closed if #450 is merged (and is more feature complete / preferable).

cgzones:
Closing in favour of #450

Issue #447, opened 2021-01-05
Enhancement request: huge page support


davezarzycki:
From the perspective of `htop`, it seems like reserved huge pages are reported as "used" regular memory, regardless of whether the huge pages are actually used or not. This makes the "memory" meter fairly useless on my system.

There are multiple ways that this can be fixed. The best solution in my opinion would be to add huge page meters (per huge page size) and subtract the reserved huge pages from the normal memory total.

What follows is `/proc/meminfo` but that isn't the full picture because that file predates the hardware supporting multiple page sizes. To get the full picture, one must also iterate /sys/kernel/mm/hugepages. See the example at the end.

```
$ cat /proc/meminfo
MemTotal:       593026380 kB
MemFree:        22093964 kB
MemAvailable:   34258672 kB
Buffers:          333600 kB
Cached:         13545432 kB
SwapCached:            0 kB
Active:          7230172 kB
Inactive:        6951556 kB
Active(anon):      20240 kB
Inactive(anon):   312488 kB
Active(file):    7209932 kB
Inactive(file):  6639068 kB
Unevictable:        1868 kB
Mlocked:            1868 kB
SwapTotal:      500106236 kB
SwapFree:       500106236 kB
Dirty:               284 kB
Writeback:             0 kB
AnonPages:        311036 kB
Mapped:           309464 kB
Shmem:             28432 kB
KReclaimable:     782220 kB
Slab:            1172164 kB
SReclaimable:     782220 kB
SUnreclaim:       389944 kB
KernelStack:       21232 kB
PageTables:         7404 kB
NFS_Unstable:          0 kB
Bounce:                0 kB
WritebackTmp:          0 kB
CommitLimit:    519795360 kB
Committed_AS:    2849184 kB
VmallocTotal:   34359738367 kB
VmallocUsed:      628072 kB
VmallocChunk:          0 kB
Percpu:            60288 kB
HardwareCorrupted:     0 kB
AnonHugePages:         0 kB
ShmemHugePages:        0 kB
ShmemPmdMapped:        0 kB
FileHugePages:         0 kB
FilePmdMapped:         0 kB
CmaTotal:              0 kB
CmaFree:               0 kB
HugePages_Total:    8192
HugePages_Free:     8138
HugePages_Rsvd:        4
HugePages_Surp:        0
Hugepagesize:       2048 kB
Hugetlb:        553648128 kB
DirectMap4k:      801512 kB
DirectMap2M:    22953984 kB
[...]

PR #450, opened 2021-01-05
Linux: handle hugepages

cgzones:
Subtract hugepages from normal memory.
Add a HugePageMeter.

Closes: #447

/cc @davezarzycki


davezarzycki:
Thanks for the super fast turnaround. This much better than my patch: #449 .

linux/LinuxProcessList.c
```diff
@@ -1554,6 +1554,57 @@ static inline void LinuxProcessList_scanMemoryInfo(ProcessList* this) {
...
+      unsigned long int hugePageSize = strtoul(name + strlen("hugepages-"), &endptr, 10);
+      if (!endptr || *endptr != 'k')
+         continue;
+
+      char content[64];
+      char hugePagePath[128];
+      ssize_t r;
+
+      xSnprintf(hugePagePath, sizeof(hugePagePath), "/sys/kernel/mm/hugepages/%s/nr_hugepages", name);
+      r = xReadfile(hugePagePath, content, sizeof(content));
```
BenBE, line 1586: Maybe add `xReadfileat`?

linux/HugePageMeter.c
```diff
@@ -0,0 +1,63 @@
...
+      .display = HugePageMeter_display,
+   },
+   .updateValues = HugePageMeter_updateValues,
+   .defaultMode = BAR_METERMODE,
+   .maxItems = 1,
+   .total = 100.0,
+   .attributes = HugePageMeter_attributes,
+   .name = "HugePages",
+   .uiName = "HugePages",
+   .caption = "HuP"
```
BenBE, line 62: Just `HP ` maybe?

linux/LinuxProcessList.c
cgzones: For `xReadfileat()` we need a file-descriptor of the parent directory; so we'd have to open `/sys/kernel/mm/hugepages/`, then open the distinct pagesize directory (e.g. `hugepages-2048kb`) and then pass that to `xReadfileat()`.

As `LinuxProcessList_scanHugePages()` is only called once every update cycle I avoided that.

linux/HugePageMeter.c
davezarzycki: For whatever it may be worth, I still think that people that use huge pages will want a meter per page size. So for example, x86-64 supports 2 MiB and 1 GiB huge pages. On other architectures, the available huge page sizes aren't friendly to two or three character abbreviations. For example, on arm64, a 512 MiB huge page size is possible: https://www.kernel.org/doc/html/latest/arm64/hugetlbpage.html

cgzones: Sadly Meters do not (yet) support options, so currently we would have to create (at compile time) Meters for all possible/supported page sizes.
As long we support *only* 2, 512 and 1024 MiB that might be a workaround.

davezarzycki: I was thinking about this some more. We could just have different values for "the" huge page meter, so rather than "used" versus "total", it becomes "used-2m", "used-1g", "total". And the most number of huge page sizes I've seen supported at the same time seems to be four. What do you think?

BenBE: We already support for meters to show more than just one value. So this probably won't be the main issue to show them. What might be somewhat harder, when not falling back to hardcoding the actual sizes is to transmit the actual labels. If the size of hugepages are more or less known, you could also just alloc an array with say 24 entries (1MiB to 8TiB) and just put the proper values in the right slot or `NAN` if unsupported. On the Meter side, coloring would than just count the used slots with the first used getting color 1, the second color 2, and so on. Sure some more LOC, but wouldn't limit extensibility too much and avoids hardcoding the labels (HugePages that ain't Power of Two make little sense).

davezarzycki: So think I've more or less implemented this model. Please see the following commit that builds on this pull request: https://github.com/davezarzycki/htop/commit/410c24341deaceb1b539fc8cad3873b0fb813eef

davezarzycki: Thanks for the feedback. Here is my updated branch that is derived from this pull request: https://github.com/davezarzycki/htop/tree/per_hugepage_values

davezarzycki: The above branch now has a fix for the hasty `nextUsed` change I just made.

davezarzycki: I think I've addressed all of the feedback to date, including the import order change that I overlooked yesterday.

I've also updated the branch after two additional tests where zero or one huge page size is available (rather than the two sizes on my system). Now zero or one available huge page size will print " used:" instead of the page size, and zero available huge page sizes are now treated as if one huge page size exists (with zero used and zero total).

BenBE (changes requested):

davezarzycki:
@cgzones – I've rebased and updated my `per_hugepage_values` branch based on this pull request. I think I've addressed all but one the feedback notes @BenBE has provided. I don't feel any emotional attachment to my patch that builds on your pull request, so if you want to just "squash" it into your pull request, that's okay with me. If credit matters to you, then I can just keep updating my branch based on your pull request.

@BenBE – I'm not sure what you mean by [this comment](https://github.com/htop-dev/htop/pull/450#discussion_r558847538). That `total` field is effectively `const`, is it not? In any case, the update function does update the total and I haven't figured out a way yet where that can go awry.

BenBE (approved):

cgzones:
Thanks @davezarzycki for contribution.

PR #453, opened 2021-01-07
Linux: Add SwapCached to the swap meter

davezarzycki:
According to the Linux kernel documentation, "SwapCached" tracks "memory that once was swapped out, is swapped back in but still also is in the swapfile (if memory is needed it doesn't need to be swapped out AGAIN because it is already in the swapfile. This saves I/O)."


Action.c
```diff
@@ -530,8 +530,9 @@ static Htop_Reaction actionHelp(State* st) {
    attrset(CRT_colors[DEFAULT_COLOR]);
    mvaddstr(line++, 0, "Swap bar:      ");
    addattrstr(CRT_colors[BAR_BORDER], "[");
-   addattrstr(CRT_colors[SWAP], "used");
-   addattrstr(CRT_colors[BAR_SHADOW], "                                          used/total");
+   addattrstr(CRT_colors[SWAP], "used"); addstr("/");
+   addattrstr(CRT_colors[MEMORY_CACHE], "cache");
```
cgzones, line 534: `SWAP_CACHE` instead of `MEMORY_CACHE` ?

SwapMeter.c
```diff
@@ -38,6 +39,9 @@ static void SwapMeter_display(const Object* cast, RichString* out) {
    Meter_humanUnit(buffer, this->values[0], sizeof(buffer));
    RichString_appendAscii(out, CRT_colors[METER_TEXT], " used:");
    RichString_appendAscii(out, CRT_colors[METER_VALUE], buffer);
+   Meter_humanUnit(buffer, this->values[1], sizeof(buffer));
```
cgzones, line 42: Please a blank line.

Also this block should only be compiled on Linux (`#ifdef HTOP_LINUX`), as on other platforms it will always show `0K`.

cgzones (changes requested):

davezarzycki:
I believe I've addressed the given feedback.

cgzones (approved):
LGTM

Action.c
```diff
@@ -530,8 +530,9 @@ static Htop_Reaction actionHelp(State* st) {
    attrset(CRT_colors[DEFAULT_COLOR]);
    mvaddstr(line++, 0, "Swap bar:      ");
    addattrstr(CRT_colors[BAR_BORDER], "[");
-   addattrstr(CRT_colors[SWAP], "used");
-   addattrstr(CRT_colors[BAR_SHADOW], "                                          used/total");
+   addattrstr(CRT_colors[SWAP], "used"); addstr("/");
```
BenBE, line 533: Shouldn't the `/` be marked in `BAR_SHADOW` for consistency?

BenBE (changes requested):
I'm not sure if `this->values[1]` in `Platform_setSwapValues` should be assigned `NAN` when that value is unavailable. What about skipping the `cached` display if it's not available?

Also: Shouldn't this attribute also be assigned for non-Linux platforms? Is this (or some similar value) available on other platforms too? If so, could you try to add it?

Action.c
cgzones: This fulfils consistency (I am not saying the old behaviour should persist) :
https://github.com/htop-dev/htop/blob/bfcb8ca0196eef942e6363e2fd7faa80eddec644/Action.c#L505-L519

BenBE: Ah, k.

davezarzycki: I agree that the `/` color ought to be `BAR_SHADOW` but I didn't want to then fix the non-swap meter help output in what should just be a "swap cache" pull request. Can I do a followup PR instead that focuses on fixing all of the help output?

davezarzycki:
Hi @BenBE – I'm new to contributing to `htop` so I'm not sure of what exactly you are requesting. And I don't have access to other platforms. Are you just asking that either `NAN` or zero be assigned to `this->values[1]` inside of `Platform_setSwapValues` on all supported platforms?

BenBE:
> And I don't have access to other platforms.

For this basically some research would be fine to get a feel if this needs counterparts for other platforms. E.g. for FreeBSD it seems like [nothing comparable exists](https://wiki.freebsd.org/Memory).

> Are you just asking that either `NAN` or zero be assigned to `this->values[1]` inside of `Platform_setSwapValues` on all supported platforms?

I was suggesting to assign `NAN` on all unsupported platforms as well on supported platforms if the value couldn't be retrieved there.

davezarzycki:
I don't see any thing to suggest that the other platforms have an exported statistic like Linux's `SwapCached`.

I believe I've addressed the rest of the feedback to date.

BenBE:
Minor addition: FreeBSD 12 has something comparable (used by e.g. PostgreSQL), but that's deprecated and removed in FreeBSD 13. Thus probably not worth the effort to implement.

BenBE (approved):

PR #454, opened 2021-01-07
Drop usage of formatted error messages from <err.h>

cgzones:
They do not clean up the ncurses environment, leaving the terminal in a
broken state.

Also drop bare usage of exit(3).


BenBE (approved):
LGTM.

PR #455, opened 2021-01-08
InfoScreen: fix uncontrolled format string

deviant:
mvwprintw takes a format string as its fourth argument, and title is
user-controlled. This results in e.g. crashing when trying to trace a
process with a format specifier in its command line.


cgzones:
This is bad.

Thanks for the fix.

PR #458, opened 2021-01-08
CI: enable format attributes in ncurses headers

cgzones:
Avoid format string issues like bfcb8ca by helping compilers spot such
bogus usages.

Also use LTO and O3 in the full-featured gcc job, which might trigger
additional warnings on advanced inlining, like 3695cbd and ad3acfc

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,11 @@ name: CI
 
 on: [ push, pull_request ]
 
+env:
+  # Enable format attributes in ncurses headers
+  # Enable fortified memory/string handling
+  CPPFLAGS: -DGCC_PRINTF -DGCC_SCANF -D_FORTIFY_SOURCE=2
+
 jobs:
   build-ubuntu-latest-minimal-gcc:
     runs-on: ubuntu-latest
@@ -47,6 +52,10 @@ jobs:
 
   build-ubuntu-latest-full-featured-gcc:
     runs-on: ubuntu-latest
+    # Enable LTO, might trigger additional warnings on advanced inlining
+    env:
+      CFLAGS: -O3 -g -flto
+      LDFLAGS: -O3 -g -flto
     steps:
     - uses: actions/checkout@v2
     - name: Install Dependencies
--- a/CRT.c
+++ b/CRT.c
@@ -818,9 +818,13 @@ void CRT_handleSIGSEGV(int signal) {
       "- Your OS and kernel version (uname -a)\n"
       "- Your distribution and release (lsb_release -a)\n"
       "- Likely steps to reproduce (How did it happened?)\n"
+   );
+
 #ifdef HAVE_EXECINFO_H
-      "- Backtrace of the issue (see below)\n"
+   fprintf(stderr, "- Backtrace of the issue (see below)\n");
 #endif
+
+   fprintf(stderr,
       "\n"
    );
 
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -47,6 +52,10 @@ jobs:
 
   build-ubuntu-latest-full-featured-gcc:
     runs-on: ubuntu-latest
+    # Enable LTO, might trigger additional warnings on advanced inlining
+    env:
+      CFLAGS: -O3 -g -flto
+      LDFLAGS: -flto
```
BenBE, line 58: With LTO you should repeat the flags for `-O3 -g` to ensure they appear in the final linking command.

cgzones: added

BenBE (approved):
LGTM.

fasterit:
splitting the fprintfs wrapped in IFDEFs would make the error logic more readable and elegant

BenBE:
> splitting the fprintfs wrapped in IFDEFs would make the error logic more readable and elegant

Having them as one large `fprintf` each was done intentional as IMO this reads much nicer in the code and causes less noise when looking at the textual (multi-line) message.

Or are we talking about some other `fprintf`s I'm missing right now?

fasterit:
> Or are we talking about some other `fprintf`s I'm missing right now?

I was talking about the ones that have a macro expansion indirection now that you need to lookup 10 or so lines above to understand what is printed on different platforms.

BenBE:
> > Or are we talking about some other `fprintf`s I'm missing right now?
> 
> I was talking about the ones that have a macro expansion indirection now that you need to lookup 10 or so lines above to understand what is printed on different platforms.

Ah, k. I can live with both versions, personally prefer the old variant, but can understand why this patch was made to split them.

cgzones:
@fasterit so you prefer:
```c
#ifdef HTOP_DARWIN
fprintf(stderr,
      "---\n"
      "\n"
      "To make the above information more practical to work with,\n"
      "you should provide a disassembly of your binary.\n"
      "This can usually be done by running the following command:\n"
      "\n"
      "   otool -tvV `which htop` > ~/htop.otool\n"
      "\n"
      "Please include the generated file in your report.\n"
      "\n"
   );
#else
fprintf(stderr,
      "---\n"
      "\n"
      "To make the above information more practical to work with,\n"
      "you should provide a disassembly of your binary.\n"
      "This can usually be done by running the following command:\n"
      "\n"
      "   objdump -d -S -w `which htop` > ~/htop.objdump\n"
      "\n"
      "Please include the generated file in your report.\n"
      "\n"
   );
#endif
```
?

BenBE:
@cgzones Why not stick with the existing version?

cgzones:
> Why not stick with the existing version?

and overriding the clang warning with `-Wno-embedded-directive`?

fasterit:
> @fasterit so you prefer [..]:

I was more thinking of

```c
fprintf(stderr,
      "---\n"
      "\n"
      "To make the above information more practical to work with,\n"
      "you should provide a disassembly of your binary.\n"
      "This can usually be done by running the following command:\n"
      "\n"
   );

#ifdef HTOP_DARWIN
      printf(sterr, "   otool -tvV `which htop` > ~/htop.otool\n");
#else
      printf(sterr, "   objdump -d -S -w `which htop` > ~/htop.objdump\n");
#endif

fprintf(stderr,
      "\n"
      "Please include the generated file in your report.\n"
      "\n"
   );
```
And similarly for the other location in the source.

fasterit:
awesome, thanks

Issue #459, opened 2021-01-09
htop, run strace (key s) -> assert fails / segfaults


fasterit:
htop HEAD at a076488 (current at time of submission)

htop: RichString.c:78: RichString_setAttrn: Assertion `(0) <= (this->chlen - 1)' failed.

```
=== snip ===

Error information:
------------------
A signal 6 (Aborted) was received.

Backtrace information:
----------------------
The following function calls were active when the issue was detected:
---
./htop(CRT_handleSIGSEGV+0xc1)[0x559224385881]
/lib/x86_64-linux-gnu/libc.so.6(+0x37840)[0x7efe816d5840]
/lib/x86_64-linux-gnu/libc.so.6(gsignal+0x10b)[0x7efe816d57bb]
/lib/x86_64-linux-gnu/libc.so.6(abort+0x121)[0x7efe816c0535]
/lib/x86_64-linux-gnu/libc.so.6(+0x2240f)[0x7efe816c040f]
/lib/x86_64-linux-gnu/libc.so.6(+0x30102)[0x7efe816ce102]
./htop(+0x1e9a3)[0x5592243909a3]
./htop(+0x1eba2)[0x559224390ba2]
./htop(Panel_draw+0x74a)[0x55922438d2fa]
./htop(InfoScreen_drawTitled+0x15a)[0x559224388a0a]
./htop(InfoScreen_run+0x2c)[0x559224388b8c]
./htop(+0xf608)[0x559224381608]
./htop(+0x176ee)[0x5592243896ee]
./htop(ScreenManager_run+0x279)[0x5592243915b9]
./htop(main+0x604)[0x559224380ff4]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xeb)[0x7efe816c209b]
./htop(_start+0x2a)[0x55922438128a]

=== snip ===

bt full:

#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
        set = {__val = {0, 94086165386816, 115, 139631558485111, 485258788864, 94086165386816, 94086165386816, 94086165386816, 
            94086165386816, 94086165386905, 94086165386916, 94086165386816, 94086165386916, 0, 0, 0}}
        pid = <optimized out>
        tid = <optimized out>
        ret = <optimized out>
#1  0x00007efe816c0535 in __GI_abort () at abort.c:79
        save_stage = 1
[...]

PR #460, opened 2021-01-10
RichString_setAttrn: refactor to take a length instead of a stop index

cgzones:
Fixes: #459

### DIFF
--- a/Process.c
+++ b/Process.c
@@ -183,7 +183,8 @@ void Process_fillStarttimeBuffer(Process* this) {
 }
 
 static inline void Process_writeCommand(const Process* this, int attr, int baseattr, RichString* str) {
-   int start = RichString_size(str), finish = 0;
+   int start = RichString_size(str);
+   int len = 0;
    const char* comm = this->comm;
 
    if (this->settings->highlightBaseName || !this->settings->showProgramPath) {
@@ -192,25 +193,24 @@ static inline void Process_writeCommand(const Process* this, int attr, int basea
          if (comm[i] == '/') {
             basename = i + 1;
          } else if (comm[i] == ':') {
-            finish = i + 1;
+            len = i + 1;
             break;
          }
       }
-      if (!finish) {
+      if (len == 0) {
          if (this->settings->showProgramPath) {
             start += basename;
          } else {
             comm += basename;
          }
-         finish = this->basenameOffset - basename;
+         len = this->basenameOffset - basename;
       }
-      finish += start - 1;
    }
 
    RichString_appendWide(str, attr, comm);
 
    if (this->settings->highlightBaseName) {
-      RichString_setAttrn(str, baseattr, start, finish);
+      RichString_setAttrn(str, baseattr, start, len);
    }
 }
 
--- a/RichString.c
+++ b/RichString.c
@@ -73,12 +73,10 @@ static inline int RichString_writeFromAscii(RichString* this, int attrs, const c
    return len;
 }
 
-inline void RichString_setAttrn(RichString* this, int attrs, int start, int finish) {
-   cchar_t* ch = this->chptr + start;
-   finish = CLAMP(finish, 0, this->chlen - 1);
-   for (int i = start; i <= finish; i++) {
-      ch->attr = attrs;
-      ch++;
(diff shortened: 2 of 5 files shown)


Process.c
```diff
@@ -183,7 +183,7 @@ void Process_fillStarttimeBuffer(Process* this) {
 }
 
 static inline void Process_writeCommand(const Process* this, int attr, int baseattr, RichString* str) {
-   int start = RichString_size(str), finish = 0;
+   int start = RichString_size(str), len = 0;
```
BenBE, line 186: Refactor for one var per line?

Process.c
```diff
@@ -192,25 +192,24 @@ static inline void Process_writeCommand(const Process* this, int attr, int basea
...
+      if (len == 0) {
          if (this->settings->showProgramPath) {
             start += basename;
          } else {
             comm += basename;
          }
-         finish = this->basenameOffset - basename;
+         len = this->basenameOffset - basename;
       }
-      finish += start - 1;
```
BenBE, line 207: Not sure about this change. Can you explain?

RichString.c
```diff
@@ -73,10 +73,10 @@ static inline int RichString_writeFromAscii(RichString* this, int attrs, const c
    return len;
 }
 
-inline void RichString_setAttrn(RichString* this, int attrs, int start, int finish) {
+inline void RichString_setAttrn(RichString* this, int attrs, int start, int len) {
```
BenBE, line 76: len is a size entity, thus what about `size_t`?

Maybe also indicate if `len` specifies `bytes` or `cells`/`code points`/`characters`.

RichString.c
```diff
@@ -110,10 +110,10 @@ static inline int RichString_writeFromAscii(RichString* this, int attrs, const c
    return RichString_writeFromWide(this, attrs, data_c, from, len);
 }
 
-void RichString_setAttrn(RichString* this, int attrs, int start, int finish) {
+void RichString_setAttrn(RichString* this, int attrs, int start, int len) {
```
BenBE, line 113: dito

BenBE (commented):
LGTM. Some minor optional points for consideration.

RichString.c
cgzones: renamed to `charcount`

Process.c
cgzones: done

Process.c
cgzones: Previously we needed the finish index for `RichString_setAttrn()`. So the length had to be adjusted by the starting point (`+ start`) and since the finish index meant the last character index to change (and not 1 past the last one) `1` had to be subtracted.
Now that's obsolete.

BenBE (approved):

RichString.c
```diff
@@ -73,10 +73,10 @@ static inline int RichString_writeFromAscii(RichString* this, int attrs, const c
...
 
-inline void RichString_setAttrn(RichString* this, int attrs, int start, int finish) {
+inline void RichString_setAttrn(RichString* this, int attrs, int start, int charcount) {
    cchar_t* ch = this->chptr + start;
-   finish = CLAMP(finish, 0, this->chlen - 1);
-   for (int i = start; i <= finish; i++) {
+   charcount = CLAMP(charcount, 0, this->chlen);
+   for (int i = 0; i < charcount; i++) {
       ch->attr = attrs;
       ch++;
```
BenBE, line 81: Optional: Keep `ch` const and use `ch[i]->attr` for access.

RichString.c
```diff
@@ -110,10 +110,10 @@ static inline int RichString_writeFromAscii(RichString* this, int attrs, const c
...
 
-void RichString_setAttrn(RichString* this, int attrs, int start, int finish) {
+void RichString_setAttrn(RichString* this, int attrs, int start, int charcount) {
    chtype* ch = this->chptr + start;
-   finish = CLAMP(finish, 0, this->chlen - 1);
-   for (int i = start; i <= finish; i++) {
+   charcount = CLAMP(charcount, 0, this->chlen);
+   for (int i = 0; i < charcount; i++) {
       *ch = (*ch & 0xff) | attrs;
       ch++;
```
BenBE, line 118: Optional: dito.

BenBE (approved):

BenBE (approved):

PR #461, opened 2021-01-10
Linux: use correct column alignment for wide process commands

cgzones:


linux/LinuxProcess.c
```diff
@@ -707,15 +707,18 @@ static void LinuxProcess_writeField(const Process* this, RichString* str, Proces
...
-         xSnprintf(buffer, n, "%-15.15s ", Process_isKernelThread(lp) ? kthreadID : "N/A");
+         procComm = Process_isKernelThread(lp) ? kthreadID : "N/A";
       }
-      break;
+      /* 15 being (TASK_COMM_LEN - 1) */
+      int c = RichString_appendnWide(str, attr, procComm, MINIMUM(15, strlen(procComm)));
+      RichString_appendChr(str, ' ', 16 - c);
+      return;
    }
    case PROC_EXE: {
```
BenBE, line 723: `PROC_EXE` is affected from this issue just the same. as is the `CWD` column.

Similar may also apply for the DragonFlyBSD/FreeBSD `JAIL` column, and Solaris `ZONE` column.

BenBE (changes requested):

BenBE (approved):

freebsd/FreeBSDProcessList.c
```diff
@@ -411,7 +411,7 @@ static char* FreeBSDProcessList_readProcessName(kvm_t* kd, const struct kinfo_pr
 }
 
 static char* FreeBSDProcessList_readJailName(const struct kinfo_proc* kproc) {
-   char*  jname = NULL;
+   char*  jname;
```
BenBE, line 414: Keep initializion to `NULL` and only assign `"N/A""`if still at `NULL` upon trying to return.

BenBE (changes requested):
Addressing the issue remarked upon by TravisCI.

BenBE (approved):

Process.c
```diff
@@ -244,6 +244,11 @@ void Process_outputRate(RichString* str, char* buffer, size_t n, double rate, in
    }
 }
 
+void Process_printLeftAlignedField(RichString* str, int attr, const char* content, unsigned int width) {
+   int c = RichString_appendnWide(str, attr, content, MINIMUM(width, strlen(content)));
```
fasterit, line 248: strlen will not account for wide chars (it essentially counts bytes)
So testcase:
```bash
echo -e '#!/bin/bash\nsleep 600' > 你好吗 && chmod +x 你好吗 && ./你好吗
```
and run htop with COMM as a visible column

BenBE: Ref: https://github.com/htop-dev/htop/blob/master/RichString.c#L53

PR #464, opened 2021-01-11
Fix palette

deviant:
This fixes a bug where the Light Terminal colour scheme renders white elements in the default foreground colour.

If the last commit message seems a little sparse, it's because I had actually written up a 5-paragraph explanation for it *twice*, lost the first one due to power outage, and the second one because git didn't use the commit message and I only noticed after I had done a further rebase that wiped out .git/COMMIT_EDITMSG. I'm only dying inside very slightly right now


fasterit:
This needs thorough testing on the Light themes of Gnome, KDE and XFCE4 as @deviant uses a very specific terminal theme.
It is a substantial refactoring with renames in ~820 lines, so makes backporting harder.

fasterit:
Hey @deviant, we discussed this PR in the core team meeting yesterday. While we understand the rationale for your changes we consider the noise / benefit ratio to not be viable enough to merge it. Please consider getting some feedback on ideas / plans before attempting larger refactorings in the future. The underlying issue you wanted to fix has been merged with your #465 PR.

PR #469, opened 2021-01-11
Separate data-update and drawing of header

cgzones:
Currently on `Header_draw()` all Meters run `_updateValues()`.
Separate into `Header_draw()` and `Header_updateData()`.

### DIFF
--- a/CPUMeter.c
+++ b/CPUMeter.c
@@ -187,6 +187,15 @@ static void AllCPUsMeter_getRange(const Meter* this, int* start, int* count) {
    }
 }
 
+static void AllCPUsMeter_updateValues(Meter* this) {
+   CPUMeterData* data = this->meterData;
+   Meter** meters = data->meters;
+   int start, count;
+   AllCPUsMeter_getRange(this, &start, &count);
+   for (int i = 0; i < count; i++)
+      Meter_updateValues(meters[i]);
+}
+
 static void CPUMeterCommonInit(Meter* this, int ncol) {
    int cpus = this->pl->cpuCount;
    CPUMeterData* data = this->meterData;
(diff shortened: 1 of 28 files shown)


CPUMeter.c
```diff
@@ -187,6 +187,15 @@ static void AllCPUsMeter_getRange(const Meter* this, int* start, int* count) {
    }
 }
 
+static void AllCPUsMeter_updateValues(Meter* this) {
+   CPUMeterData* data = this->meterData;
+   Meter** meters = data->meters;
+   int start, count;
+   AllCPUsMeter_getRange(this, &start, &count);
+   for (int i = 0; i < count; i++)
```
natoscott, line 195: Should this loop start from 'start' and not zero?

ScreenManager.c
```diff
@@ -105,6 +105,8 @@ static void checkRecalculation(ScreenManager* this, double* oldTime, int* sortTi
 
    if (*rescan) {
       *oldTime = newTime;
+      // alawys update header, especially to avoid gaps in graph meters
```
natoscott, line 108: Little typo on 'alawys' there

natoscott:
@cgzones it would be good to merge this once you feel its ready to go - I have some changes incoming that remove the repeated calls to gettimeofday from Meters like DiskIO and NetworkIO (passing the initial sample timestamp through for Meters to use if they need that), as well as the repeated calculations of sample interval in these Meters.  But those changes introduce parameters which will conflict with your parameter changes here, so it would be good if this one can go in first.

ScreenManager.c
cgzones: thanks, fixed

CPUMeter.c
cgzones: `meters` holds single core meters in the range [0..count-1] with the ids [start..start+count-1]

natoscott (approved):
@cgzones LGTM.

BenBE (approved):

PR #443, opened 2021-01-01
Cache stderr to be able to print assert messages

cgzones:

PR #470, opened 2021-01-11
Introduce assert(3) wrapper

cgzones:
In case of an assert failure, call CRT_done() to cleanup up the ncurses
environment resulting in the assert message to be visible in the
terminal.

Related to #443

### DIFF
--- a/Macros.h
+++ b/Macros.h
@@ -3,6 +3,7 @@
 
 #include <assert.h> // IWYU pragma: keep
 
+
 #ifndef MINIMUM
 #define MINIMUM(a, b)             ((a) < (b) ? (a) : (b))
 #endif
@@ -12,7 +13,7 @@
 #endif
 
 #ifndef CLAMP
-#define CLAMP(x, low, high)       (assert((low) <= (high)), ((x) > (high)) ? (high) : MAXIMUM(x, low))
+#define CLAMP(x, low, high)       (HTOP_ASSERT((low) <= (high)), ((x) > (high)) ? (high) : MAXIMUM(x, low))
 #endif
 
 #ifndef ARRAYSIZE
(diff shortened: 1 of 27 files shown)


Macros.h
```diff
@@ -27,6 +26,14 @@
 #define SPACESHIP_NULLSTR(a, b)   strcmp((a) ? (a) : "", (b) ? (b) : "")
 #endif
 
+#ifndef HTOP_ASSERT
+#ifndef NDEBUG
+#define HTOP_ASSERT(expr)         ((expr) ? ((void)0) : (CRT_done(), assert(expr)))
```
BenBE, line 31: This may re-evaluate `expr` twice …

According to documentation `assert(3)` calls `abort(3)` which `raise(3)`s some `SIGABRT`. Maybe catch things from that end?

Macros.h
```diff
@@ -1,7 +1,6 @@
 #ifndef HEADER_Macros
 #define HEADER_Macros
 
-#include <assert.h> // IWYU pragma: keep
```
BenBE, line 4: Please keep this as using `HTOP_ASSERT` requires this header to be included. Removing this line scatters this all over the place unnecessarily.

cgzones: dropped

Macros.h
cgzones: One should not pass an expression to `assert(3)`, cause it might (depending on `NDEBUG` being defined or not) evaluated once or not at all.
So currently the expression gets evaluated 0 time in normal mode and 1 time in debug mode.
With the new wrapper it would get evaluated 0 time in normal mode, 1 time in debug mode if true and 2 times in debug mode if false.
The main point of using `assert(3)` instead of `raise(SIGABRT)` is the message formatting of `assert(3)` (which we otherwise would need to replicate).

cgzones:
Closing in favour of #443

PR #472, opened 2021-01-12
Do not try to set not owned capabilities

cgzones:
If the process has already less capabilities than we are trying to keep,
do not try to set them.


maxiberta:
Fwiw, this prevents [the snap](https://snapcraft.io/htop) (on the `edge` channel) from aborting with `Error: can not set process capabilities: Operation not permitted` when passing `--drop-capabilities`.

BenBE (approved):
LGTM.

PR #477, opened 2021-01-14
Define PATH_MAX for GNU/hurd

fasterit:
Otherwise fails with
"> linux/LinuxProcessList.c:889:20: error: ‘PATH_MAX’ undeclared (first use in this function)"

### DIFF
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -42,6 +42,7 @@ in the source distribution for its full text.
 #include "LinuxProcess.h"
 #include "Macros.h"
 #include "Object.h"
+#include "Platform.h" // needed for GNU/hurd to get PATH_MAX
 #include "Process.h"
 #include "Settings.h"
 #include "XUtils.h"
--- a/linux/Platform.c
+++ b/linux/Platform.c
@@ -14,7 +14,6 @@ in the source distribution for its full text.
 #include <dirent.h>
 #include <fcntl.h>
 #include <inttypes.h>
-#include <limits.h>
 #include <math.h>
 #include <stdint.h>
 #include <stdio.h>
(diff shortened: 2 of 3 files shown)


BenBE (approved):

cgzones (approved):

BenBE (approved):

jrtc27:
This is a bit of a hacky fix; it'll work in practice but really you need to handle ENAMETOOLONG and reallocate a bigger buffer if you want to do it properly (and then you don't need a fake PATH_MAX, just a default starting value that's good enough almost all of the time).

BenBE:
> This is a bit of a hacky fix; it'll work in practice but really you need to handle ENAMETOOLONG and reallocate a bigger buffer if you want to do it properly (and then you don't need a fake PATH_MAX, just a default starting value that's good enough almost all of the time).

A proper patch will be happily accepted. ;-)

Yet FWIW, there aren't too many situations where htop actually needs to handle long paths: procfs is at most 128 chars (even in extreme cases) and sysfs hasn't too many long filenames either. The only place where long filenames could occure is when calling `readlink`/`readlinkat` on the `/proc/<pid>/exe` symlink to determine the EXE column information.

jrtc27:
> > This is a bit of a hacky fix; it'll work in practice but really you need to handle ENAMETOOLONG and reallocate a bigger buffer if you want to do it properly (and then you don't need a fake PATH_MAX, just a default starting value that's good enough almost all of the time).
> 
> A proper patch will be happily accepted. ;-)
> 
> Yet FWIW, there aren't too many situations where htop actually needs to handle long paths: procfs is at most 128 chars (even in extreme cases) and sysfs hasn't too many long filenames either. The only place where long filenames could occure is when calling `readlink`/`readlinkat` on the `/proc/<pid>/exe` symlink to determine the EXE column information.

Yes, that was the one that came to mind (and similarly `/proc/<pid>/fd/<n>` if it's used somewhere). Also technically there's no reason `/proc` can't be a symlink to `/really/long/path` where the actual procfs is mounted, even if that's a stupid thing to do...

Issue #471, opened 2021-01-12
3.0.2 -> 3.0.3 bug: slow cpufreq on AMD EPYC 7502P (32 cores + SMT)


KonstantinPlotnikov:
On Arch Linux server with CPU AMD EPYC 7502P (32 cores + SMT):
Usage of the option "Also show CPU frequency" makes htop very slow to launch and very laggy since version 3.0.3.
Using strace -T shows this:
```
openat(AT_FDCWD, "/sys/devices/system/cpu/cpu8/cpufreq/scaling_cur_freq", O_RDONLY) = 4 <0.000025>
fstat(4, {st_mode=S_IFREG|0444, st_size=4096, ...}) = 0 <0.000011>
read(4, "3341788\n", 4096)              = 8 <0.016539>
close(4)                                = 0 <0.000023>
```

Downgrading to 3.0.2 or disabling option "Also show CPU frequency" solves issue.

BenBE:
Can you re-test with the current 3.0.5 release? Cf. also, #451 and #467.

KonstantinPlotnikov:
Re-tested on 3.0.5: bug still present.
#467 looks very much alike.

From brief code lookup the key change is https://github.com/htop-dev/htop/commit/edf1b10d2c3fa94c23a23cfc947490cf1ebf66d1

And the reason is that on this machine reading from /sys/devices/system/cpu/cpufreq/policy....../scaling_cur_freq is slow (14.5 ms per core).
My desktop has intel cpu and there are no such problem. I will test it on Ryzen 1600 later if needed (maybe it is Zen-related problem)

[AMD EPYC 7502P.txt](https://github.com/htop-dev/htop/files/5801599/AMD.EPYC.7502P.txt)
[Intell Core i5 8400.txt](https://github.com/htop-dev/htop/files/5801600/Intell.Core.i5.8400.txt)

cgzones:
Some data:

Intel(R) Core(TM) i7-5500U
```
openat(AT_FDCWD, "/sys/devices/system/cpu/cpu0/cpufreq/scaling_cur_freq", O_RDONLY) = 4 <0.000051>
fstat(4, {st_mode=S_IFREG|0444, st_size=4096, ...}) = 0 <0.000026>
read(4, "936367\n", 4096)               = 7 <0.000039>
close(4)                                = 0 <0.000029>
[...]

PR #478, opened 2021-01-14
Linux: fall back to cpuinfo on slow scaling_cur_freq read

cgzones:
On some AMD CPUs read()ing scaling_cur_freq is quite slow (> 1ms).
This delay accumulates for every cpu.
If the read on CPU 0 takes longer than 500us bail out and fall back to
reading the frequencies from /proc/cpuinfo.
Once the condition has been met, bail out early for the next couple of
scans.

Closes: #471

### DIFF
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -20,6 +20,7 @@ in the source distribution for its full text.
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <time.h>
 #include <unistd.h>
 #include <sys/stat.h>
 #include <sys/time.h>
@@ -1727,9 +1728,27 @@ static int scanCPUFreqencyFromSysCPUFreq(LinuxProcessList* this) {
    int numCPUsWithFrequency = 0;
    unsigned long totalFrequency = 0;
 
+   /*
+    * On some AMD and Intel CPUs read()ing scaling_cur_freq is quite slow (> 1ms). This delay
+    * accumulates for every core. For details see issue#471.
+    * If the read on CPU 0 takes longer than 500us bail out and fall back to reading the
+    * frequencies from /proc/cpuinfo.
+    * Once the condition has been met, bail out early for the next couple of scans.
+    */
+   static int timeout = 0;
+
+   if (timeout > 0) {
+      timeout--;
+      return -1;
+   }
+
    for (int i = 0; i < cpus; ++i) {
       char pathBuffer[64];
       xSnprintf(pathBuffer, sizeof(pathBuffer), "/sys/devices/system/cpu/cpu%d/cpufreq/scaling_cur_freq", i);
+
+      struct timespec start;
+      if (i == 0)
+         clock_gettime(CLOCK_MONOTONIC, &start);
 
       FILE* file = fopen(pathBuffer, "r");
       if (!file)
@@ -1745,6 +1764,17 @@ static int scanCPUFreqencyFromSysCPUFreq(LinuxProcessList* this) {
       }
 
       fclose(file);
+
+      if (i == 0) {
+         struct timespec end;
+         clock_gettime(CLOCK_MONOTONIC, &end);
+         const time_t timeTakenUs = (end.tv_sec - start.tv_sec) * 1000000 + (end.tv_nsec - start.tv_nsec) / 1000;
+         if (timeTakenUs > 500) {
+            timeout = 30;
+            return -1;
+         }
+      }
+
    }
 
    if (numCPUsWithFrequency > 0)


linux/LinuxProcessList.c
```diff
@@ -1727,10 +1728,28 @@ static int scanCPUFreqencyFromSysCPUFreq(LinuxProcessList* this) {
    int numCPUsWithFrequency = 0;
    unsigned long totalFrequency = 0;
 
+   /*
+    * On some AMD CPUs read()ing scaling_cur_freq is quite slow (> 1ms).
```
KonstantinPlotnikov, line 1732: Just to be accurate.
As you could see in https://github.com/htop-dev/htop/issues/471#issuecomment-759378878 this is not AMD specific issue. At least some of the Intel Xeon CPUs also share such behavior.

KonstantinPlotnikov:
This fix completely solves issue #471 for me. Thank you!

BenBE (commented):
LGTM. Maybe update the comment to include this seems to be the case whenever a non-`intel_pstate` driver for CPU frequency is used (affects AMD and some Intel [Xeon] CPUs).

fasterit:
@BenBE: intel_pstate was also observed slow in #471 (Intel Core i7-7700 case).

I'd still be interested _why_ we see the difference between 15ms and ~0ms. It does not seem to be dependent on the driver. Is it consistent for each system?
May be we are just seeing the kernel context switching and scheduling another application while we are suspended for the `read` syscall. In that case testing the first read may not be much better than guessing.

Related:
`readfile()` will probably be useful in the future cf. `https://git.kernel.org/pub/scm/linux/kernel/git/gregkh/driver-core.git/commit/?h=readfile&id=3317335407261fd5f525bd5fd63ade139fed6ccb`

BenBE:
@fasterit:
> Related:
> `readfile()` will probably be useful in the future cf. `https://git.kernel.org/pub/scm/linux/kernel/git/gregkh/driver-core.git/commit/?h=readfile&id=3317335407261fd5f525bd5fd63ade139fed6ccb`

Which kernel release does this correspond to? Looks like it might be worth to check for this syscall in `xReadfile`.

fasterit:
@BenBE: None yet, Linus still needs to merge that one. Greg has had it out since ~07/2020 and it seems rather ready, so may be as early as  5.12.

clain-kp:
> I'd still be interested _why_ we see the difference between 15ms and ~0ms. It does not seem to be dependent on the driver. Is it consistent for each system?
> May be we are just seeing the kernel context switching and scheduling another application while we are suspended for the read syscall. In that case testing the first read may not be much better than guessing.

I have 11 AMD EPYC ( 1 Zen + 10 Zen2) systems each one of those has this issue. And those systems are idle most of the time. Reboot does not affect this.
I did not run specific automatic test to collect statistics on single machine, but every manual check gives pretty consistent result ( 0 of ~15 ms).

As for the reason of this: I do not know where to look. If you have any idea I could collect almost any data from "slow" CPUs.

KonstantinPlotnikov:
> I'd still be interested why we see the difference between 15ms and ~0ms. It does not seem to be dependent on the driver. Is it consistent for each system?
> May be we are just seeing the kernel context switching and scheduling another application while we are suspended for the read syscall. In that case testing the first read may not be much better than guessing.

I have 11 AMD EPYC ( 1 Zen + 10 Zen2) systems each one of those has this issue. And those systems are idle most of the time. Reboot does not affect this.
I did not run specific automatic test to collect statistics on single machine, but every manual check gives pretty consistent result ( 0 of ~15 ms).

As for the reason of this: I do not know where to look. If you have any idea I could collect almost any data from "slow" CPUs.

cgzones:
Once the `readfile(2)` syscall gets merged I am looking forward to use it in our `xReadfile()` wrapper conditionally (and maybe convert some frequently used `fopen()` + `fread()/fscanf()` + `fclose()` pairs to it).
But I fear that has nothing to do with this slowdown, as the data shows the `open` and `close` calls on the `scaling_cur_freq` inodes are fast; it's only the `read` call that's slow, probably because the driver takes that time to gather the data from the actual hardware (maybe the frequency is not stored/cached and can only be retrieved after a context switch).

BenBE:
> As for the reason of this: I do not know where to look. If you have any idea I could collect almost any data from "slow" CPUs.

I think the most interesting data here are likely flame graphs[1] of the kernel stack while executing this particular read syscall. To make that more obvious you'd most likely would write a small program that just polls that value in a loop and measure it with `perf` or some kernel tracing to get the flame graph data[2]. While this will likely give results pointing to the culprit, it's unlikely we can do anything about it in userspace; thus the results are likely to be reported to The Kernel People™ …

[1] http://www.brendangregg.com/flamegraphs.html
[2] https://www.gabriel.urdhr.fr/2014/05/23/flamegraph/#stacksampling

PR #412, opened 2020-12-18
[WIP] Build rework and static build and musl support

cgzones:

PR #482, opened 2021-01-15
configure overhaul

cgzones:
* overhaul option handling
  - switch Linux capabilities default from "no" to "check"
  - document default settings
  - use more readable formatting
* reformat for improved reabability
* misc modernizations
  - require autoconf version 2.69
    was released in 2012 and one still can configure and build on older
    systems (just not generate the configure script)
  - use modern C99 compiler check
  - drop obsolete checks: AC_C_CONST, AC_FUNC_CLOSEDIR_VOID, AC_FUNC_STAT
  - drop AC_HEADER_STDBOOL in favor of C99 compatibility
* fail immediately on missing requirement
* add option to build a static htop binary (adopted from #412)

### DIFF
--- a/README
+++ b/README
@@ -44,6 +44,59 @@ install the header files for `ncurses` (libncursesw*-dev) and other required dev
 
 By default `make install` will install into `/usr/local`, for changing the path use `./configure --prefix=/some/path`.
 
+### Build Options
+
+`htop` has several build-time options to enable/disable additional features.
+
+#### Generic
+
+  * `--enable-unicode`:
+    enable Unicode support
+    dependency: *libncursesw*
+    default: *yes*
+  * `--enable-hwloc`:
+    enable hwloc support for CPU affinity; disables Linux affinity
+    dependency: *libhwloc*
+    default: *no*
+  * `--enable-setuid`:
+    enable setuid support for privilege dropping
+    default: *no*
+  * `--enable-debug`:
+    Enable asserts and internal sanity checks; implies a performance penalty
+    default: *no*
+
+#### Linux
+
+  * `--enable-sensors`:
+    enable libsensors(3) support for reading temperature data
+    dependencies: *libsensors-dev*(build-time), at runtime *libsensors* is loaded via `dlopen(3)` if available
+    default: *check*
+  * `--enable-capabilities`:
+    enable Linux capabilities support
+    dependency: *libcap*
+    default: *check*
+  * `--with-proc`:
+    location of a Linux-compatible proc filesystem
+    default: */proc*
+  * `--enable-openvz`:
+    enable OpenVZ support
+    default: *no*
+  * `--enable-vserver`:
+    enable VServer support
+    default: *no*
+  * `--enable-ancient-vserver`:
+    enable ancient VServer support (implies `--enable-vserver`)
+    default: *no*
+   * `--enable-linux-affinity`:
+    enable Linux `sched_setaffinity(2)` and `sched_getaffinity(2)` for affinity support; conflicts with hwloc
+    default: *check*
+  * `--enable-delayacct`:
+    enable Linux delay accounting support
+    dependencies: *pkg-config*(build-time), *libnl-3* and *libnl-genl-3*
+    default: *check*
+
+## Usage
+
 See the manual page (`man htop`) or the on-line help ('F1' or 'h' inside `htop`) for a list of supported key commands.
 
 ## Support
--- a/README
+++ b/README
@@ -61,6 +61,9 @@ By default `make install` will install into `/usr/local`, for changing the path 
   * `--enable-setuid`:
     enable setuid support for privilege dropping
     default: *no*
+  *`--enable-static`:
+    build a static htop binary; hwloc and delay accounting are not supported
+    default: *no*
   * `--enable-debug`:
     Enable asserts and internal sanity checks; implies a performance penalty
     default: *no*
--- a/configure.ac
+++ b/configure.ac
@@ -84,7 +84,7 @@ AC_CHECK_HEADERS([ \
     sys/param.h \
     sys/time.h \
     unistd.h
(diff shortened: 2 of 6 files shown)


README
```diff
@@ -44,6 +44,59 @@ install the header files for `ncurses` (libncursesw*-dev) and other required dev
...
+    default: *yes*
+  * `--enable-hwloc`:
+    enable hwloc support for CPU affinity; disables Linux affinity
+    build dependency: *libhwloc*
+    default: *no*
+  * `--enable-setuid`:
+    enable setuid support for privilege dropping
+    default: *no*
+  * `--enable-debug`:
+    Enable asserts and internal sanity checks; implies a performance cut
```
fasterit, line 65: performance penalty

fasterit (approved):
configure --enable-static works like a charm, nice!

Please add static to the configure report

configure.ac
```diff
@@ -1,152 +1,213 @@
...
 AC_CONFIG_SRCDIR([htop.c])
-AC_CONFIG_AUX_DIR([.])
+AC_CONFIG_AUX_DIR([build-aux])
 AC_CONFIG_HEADERS([config.h])
 
-# Required by hwloc scripts
 AC_CANONICAL_TARGET
+AM_INIT_AUTOMAKE([-Wall std-options subdir-objects])
 
-AM_INIT_AUTOMAKE([1.11])
```
BenBE, line 14: What's the differnce between those two invocations? What do they accomplish?

configure.ac
```diff
@@ -1,152 +1,213 @@
...
+
+
+# ----------------------------------------------------------------------
+# Checks for compiler.
+# ----------------------------------------------------------------------
+
+AC_PROG_CC
+AM_PROG_CC_C_O
+AC_PROG_CC_C99
+AS_IF([test "x$ac_cv_prog_cc_c99" = xno], [AC_MSG_ERROR([htop is written in C99. A newer compiler is required])])
```
BenBE, line 69: Missing trailing dot in error message?

configure.ac
```diff
@@ -1,152 +1,213 @@
...
+   AC_CHECK_HEADERS([mach/mach_time.h])
+fi
+
+# ----------------------------------------------------------------------
+
+
+# ----------------------------------------------------------------------
 # Checks for typedefs, structures, and compiler characteristics.
 # ----------------------------------------------------------------------
-AC_HEADER_STDBOOL
```
BenBE, line 82: Why is this check for `stdbool.h` removed, when several of our source files use it (and AFAIK `stdbool.h` is defined in the C99 standard?

configure.ac
```diff
@@ -182,97 +242,140 @@ m4_define([HTOP_CHECK_SCRIPT],
...
-   AC_CHECK_HEADERS([hwloc.h],[:], [missing_headers="$missing_headers $ac_header"])
+
+AC_ARG_ENABLE([ancient_vserver],
+              [AS_HELP_STRING([--enable-ancient-vserver],
+                              [enable ancient VServer support (implies --enable-vserver) @<:@default=no@:>@])],
+              [],
+              [enable_ancient_vserver=no])
+if test "x$enable_ancient_vserver" = xyes; then
+    AC_DEFINE([HAVE_VSERVER], [1], [Define if vserver support enabled.])
+    AC_DEFINE([HAVE_ANCIENT_VSERVER], [1], [Define if ancient vserver support enabled.])
```
BenBE, line 370: VServer for consistency with the lines above?

configure.ac
```diff
@@ -357,43 +528,69 @@ AM_CFLAGS="\
...
+
 AX_CHECK_COMPILE_FLAG([-Wnull-dereference], [AM_CFLAGS="$AM_CFLAGS -Wnull-dereference"], , [-Werror])
 
-AC_ARG_ENABLE([werror], [AS_HELP_STRING([--enable-werror], [Treat warnings as errors (default: warnings are not errors)])], [enable_werror="$enableval"], [enable_werror=no])
-AS_IF([test "x$enable_werror" = "xyes"], [AM_CFLAGS="$AM_CFLAGS -Werror"])
+AC_ARG_ENABLE([werror],
+              [AS_HELP_STRING([--enable-werror],
+                              [Treat warnings as errors @<:@default=no@:>@])],
+              [],
+              [enable_werror=no])
```
BenBE, line 554: What about -Werror default yes on debug builds?

linux/SystemdMeter.c
```diff
@@ -21,28 +21,55 @@ in the source distribution for its full text.
...
 static unsigned int nJobs = INVALID_VALUE;
-static void* dlopenHandle = NULL;
-static sd_bus* bus = NULL;
 
 static void SystemdMeter_done(ATTR_UNUSED Meter* this) {
    free(systemState);
    systemState = NULL;
 
+#ifdef BUILD_STATIC
+# ifdef HAVE_LIBSYSTEMD
```
BenBE, line 66: Indentation mismatch with closing `#endif`

BenBE (changes requested):

configure.ac
cgzones: See https://www.gnu.org/software/automake/manual/html_node/Public-Macros.html for details:
Passing a version is deprecated.
See https://www.gnu.org/software/automake/manual/html_node/List-of-Automake-options.html for details on the options:
 - `-Wall`: enable automake warnings
 - `std-options`: check generated executables support `--help` and `--version`
 - `subdir-objects`: generate object files in subdirs, e.g. *linux/LinuxPorcess.o* instead of *./LinuxProcess.o*

configure.ac
cgzones: `AC_HEADER_STDBOOL` only performs checks and defines `HAVE_STDBOOL_H` and `HAVE__BOOL`, but we never actually depend or use these definitions.

BenBE (approved):

Issue #484, opened 2021-01-17
Let Memory [Text] span both columns


edrozenberg:
Recent changes have caused the `Memory [Text]` meter to no longer be able to span both columns when the second column has empty space. As a result I'm no longer able to see the cache value in a standard 80 column terminal. `Memory [LED]` can span both columns but I prefer to use `Memory [Text]`.

<img width="479" alt="Mem Text Does Not Fit" src="https://user-images.githubusercontent.com/2648417/104857046-b946e000-58ca-11eb-9e86-4798533ba14c.png">

Sug: let the `Memory [Text]` span the entire width when the second column is empty.

BenBE:
This might be addressed by #445 …

edrozenberg:
Thanks will give it a try when 3.0.6 releases.

PR #485, opened 2021-01-18
Allow meters in text mode to span empty neighbors to the right

cgzones:
Closes: #484

### DIFF
--- a/Header.c
+++ b/Header.c
@@ -150,35 +150,71 @@ void Header_reinit(Header* this) {
 }
 
 void Header_draw(const Header* this) {
-   int height = this->height;
-   int pad = this->pad;
+   const int height = this->height;
+   const int pad = this->pad;
    attrset(CRT_colors[RESET_COLOR]);
    for (int y = 0; y < height; y++) {
       mvhline(y, 0, ' ', COLS);
    }
-   int width = COLS / this->nrColumns - (pad * this->nrColumns - 1) - 1;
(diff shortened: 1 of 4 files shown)


natoscott (approved):
LGTM - one tiny possible nit: in some of the pointer casts are using whitespace before the asterix where previously/typically there was/is none - e.g.  "(const Object \*)" vs "(const Object\*)"

Header.c
```diff
@@ -150,35 +150,71 @@ void Header_reinit(Header* this) {
...
+      int height = pad;
+      for (int j = 0; j < Vector_size(meters); j++) {
+         const Meter* meter = (const Meter *) Vector_get(meters, j);
+
+         if (height >= curHeight + curMeter->h)
+            break;
+
+         height += meter->h;
+         if (height <= curHeight)
+            continue;
```
BenBE, line 198: I'm not sure this successfully catches, if two meters are offset by a few rows.

```
+---+---+
|   | 3 |
+ 1 +---+
|   |   |
+---+ 4 +
| 2 |   |
+---+---+
```

Question related to when checking 1 if non-blank 4 is caught?
Also what of 2 if 4 were even one line higher?

cgzones:
> in some of the pointer casts are using whitespace before the asterix where previously/typically there was/is none - e.g. "(const Object *)" vs "(const Object*)"

Dropped the blanks

Header.c
cgzones: I tested several cases and all seems to work.

PR #486, opened 2021-01-19
Find the correct library for clock_gettime before trying to use it

abbeyj:
Otherwise if clock_gettime is librt then this code will incorrectly believe
that the function does not exist at all.

This fixes compatibility with RHEL 6.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -92,6 +92,10 @@ AC_TYPE_UINT64_T
 # ----------------------------------------------------------------------
 AC_FUNC_CLOSEDIR_VOID
 AC_FUNC_STAT
+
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -102,7 +104,6 @@ AC_CHECK_FUNCS([\
 ])
 
 AC_SEARCH_LIBS([dlopen], [dl dld])
-AC_SEARCH_LIBS([clock_gettime], [rt])
```
BenBE, line 105: Maybe move the whole block together …

BenBE (changes requested):

BenBE (approved):

PR #489, opened 2021-01-21
Sort out the mess around column sorting that has accumulated over time

fasterit:


cgzones:
With this the tie-breaker takes the order of current column. I prefer sorting the tie-breaker always in an ascending order.

```c
// Implement tie-breaker (needed to make tree mode more stable)
if (!result)
   return SPACESHIP_NUMBER(((const Process*)v1)->pid, ((const Process*)v2)->pid);
```

Should the following (Linux-)process fields also have a default descending sorting order: `MAJFLT`, `CMAJFLT`, `UTIME`, `STIME`, `CUTIME`, `CSTIME`, `RCHAR`, `WCHAR`, `SYSCR`, `SYSCW`, `RBYTES`, `WBYTES`, `CNCLWB`, `IO_READ_RATE`, `IO_WRITE_RATE`, `IO_RATE`, `OOM`, `PERCENT_CPU_DELAY`, `PERCENT_IO_DELAY`, `PERCENT_SWAP_DELAY`, `M_PSS`, `M_SWAP`, `M_PSSWP` and `CTXT` ?

Maybe introduce macro identifiers for the special values `-1` and `1`.

fasterit:
> With this the tie-breaker takes the order of current column. I prefer sorting the tie-breaker always in an ascending order.
> 
> ```c
> // Implement tie-breaker (needed to make tree mode more stable)
> if (!result)
>    return SPACESHIP_NUMBER(((const Process*)v1)->pid, ((const Process*)v2)->pid);
> ```
> 
> Should the following (Linux-)process fields also have a default descending sorting order: `MAJFLT`, `CMAJFLT`, `UTIME`, `STIME`, `CUTIME`, `CSTIME`, `RCHAR`, `WCHAR`, `SYSCR`, `SYSCW`, `RBYTES`, `WBYTES`, `CNCLWB`, `IO_READ_RATE`, `IO_WRITE_RATE`, `IO_RATE`, `OOM`, `PERCENT_CPU_DELAY`, `PERCENT_IO_DELAY`, `PERCENT_SWAP_DELAY`, `M_PSS`, `M_SWAP`, `M_PSSWP` and `CTXT` ?

Implemented in https://github.com/htop-dev/htop/pull/489/commits/55f700eed41f67a6bee86e1823bdbb4e06c1ba3e
Thanks for the review!

BenBE:
What about?

```diff
diff --git a/Process.c b/Process.c
index 5d95ba1..c76f47a 100644
--- a/Process.c
+++ b/Process.c
@@ -489,24 +489,21 @@ int Process_pidCompare(const void* v1, const void* v2) {
 }
 
 int Process_compare(const void* v1, const void* v2) {
-   const Process *p1, *p2;
-   const Settings *settings = ((const Process*)v1)->settings;
+   const Process *p1 = (const Process*)v1;
+   const Process *p2 = (const Process*)v2;
 
-   if (Settings_getActiveDirection(settings) == 1) {
-      p1 = (const Process*)v1;
-      p2 = (const Process*)v2;
-   } else {
-      p2 = (const Process*)v1;
-      p1 = (const Process*)v2;
-   }
+   const Settings *settings = p1->settings;
 
    ProcessField key = Settings_getActiveSortKey(settings);
 
    int result = Process_compareByKey(p1, p2, key);
 
+   if (Settings_getActiveDirection(settings) != 1)
+      result = -result;
+
    // Implement tie-breaker (needed to make tree mode more stable)
    if (!result)
-      return SPACESHIP_NUMBER(((const Process*)v1)->pid, ((const Process*)v2)->pid);
+      return SPACESHIP_NUMBER(p1->pid, p2->pid);
 
    return result;
 }
```

Much more straight forward and avoids multiple of the pointer casts …

fasterit:
> Much more straight forward and avoids multiple of the pointer casts …

That looks more beautiful. Not really related to the problem we are fixing here but merged into this PR regardless.
Nicer code is always better :-).

Process.c
```diff
@@ -489,24 +489,21 @@ int Process_pidCompare(const void* v1, const void* v2) {
...
-      p2 = (const Process*)v1;
-      p1 = (const Process*)v2;
-   }
+   const Settings *settings = p1->settings;
 
    ProcessField key = Settings_getActiveSortKey(settings);
 
    int result = Process_compareByKey(p1, p2, key);
 
+   if (Settings_getActiveDirection(settings) != 1)
```
cgzones, line 501: Maybe insert after the tie-breaker, cause `-0` should be `0`.

cgzones (approved):

Process.c
fasterit: That would make the tie breaker depend on sort order again which @BenBE objected against.

BenBE: @cgzones -0 == 0, thus this works as intended. Just reordered the code to make it remove several pointer casts that @fasterit introduced.

@fasterit I didn't object to anything. *g* Just put it in that order as there was some comment to avoid making the tie-breaker direction dependent for added stability of the sorting.

BenBE (approved):
LGTM.

Just a thing I noticed while testing: This PR badly interacts with existing configurations causing the existing configuration to be interpreted as inverted (e.g. with `M_RESIDENT`). Possibly worth a note in the changelog.

PR #491, opened 2021-01-22
GitHub CI: Add job for building on s390x

jonathan-albrecht-ibm:
Previously the Travis CI was building on linux on s390x but there is not the same out of the box support under GitHub Actions. It is fairly straightforward to use the https://github.com/marketplace/actions/run-on-architecture action to build on s390x and other architectures.

This PR adds a job to run a full featured gcc build on s390x. Other architectures could be added to the matrix as well.

The run-on-architecture builds are slower than the GitHub runner native builds probably because they are running on an emulator. It could be made to only run on pull_request and maybe the make distcheck step is not needed which would reduce the build time.

Thoughts? I'm happy for any suggestions on making this better.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,6 +91,37 @@ jobs:
     - name: Distcheck
       run: make distcheck DISTCHECK_CONFIGURE_FLAGS='--enable-werror --enable-openvz --enable-vserver --enable-ancient-vserver --enable-unicode --enable-hwloc --enable-setuid --enable-delayacct --with-sensors --with-capabilities'
 
+  build-ubuntu-arch-matrix-full-featured-gcc:
+    runs-on: ubuntu-20.04
+    name: Build on ${{ matrix.distro }} ${{ matrix.arch }}
+    strategy:
+      matrix:
+        include:
+          - arch: s390x
+            distro: ubuntu20.04
+    steps:
+    - uses: actions/checkout@v2
+    - uses: uraimo/run-on-arch-action@v2.0.8
+      name: Configure and Build
+      id: arch-matrix-gcc
+      with:
+        arch: ${{ matrix.arch }}
+        distro: ${{ matrix.distro }}
+        githubToken: ${{ github.token }}
+        shell: /bin/bash
+        install: |
+          apt-get -q -y update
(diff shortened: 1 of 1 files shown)


cgzones:
Does this actually increase the portability of htop?
Probably htop does have some issues on non x86 architectures regarding gathering CPU frequencies/temperatures, SMP, ..., but these are runtime related. Build results on such architectures should no deviate from x86, cause we are not (yet) using any conditional architecture code (https://sourceforge.net/p/predef/wiki/Architectures/).
So I think it is not worth running a build on s390x for every commit/pull-request; we may however create an extensive CI roster, which runs every Saturday? and gets triggered before every release.

fasterit:
@cgzones: I don't think this adds value. It is an emulated environment so we don't find any real issues related to the hardware / platform. This is basically just burning CPU cycles for little gain. We could just as well build on emulated ARM or RISC CPUs to mostly find issues with these emulation environments or the compiler suite, not with htop. Not our turf, really.
So -1 from me. Even for Saturday jobs, not worth it for us, not worth it for the environment.

BenBE:
Only worth if the hardware quirks were emulated too, which are probably not. I.e. misaligned pointer reads on ARM cause quite funny results. But the compiler warnings should already cover those.

jonathan-albrecht-ibm:
Thanks for the comments everyone. I agree with all the points about not being able to catch runtime related issues. I was more thinking it could catch issues at compile time. Even though htop doesn't have architecture dependent code, sometimes the headers of libs it compiles against do and it could help catch portability problems. For example, if someone misses using a typedef and the raw type triggers a compile failure on some platforms.

It doesn't sound like this has been an issue in the past but I think there might still be value running a matrix of these other architectures at a much lower frequency. I think @cgzones idea of once a week and/or before a release would be really good.

fasterit:
We discussed this during the core team meeting today. We do not think building on an emulated hardware will add much benefit for htop. We would likely just find qemu or gcc oddities and we want to leave these opportunities to Debian and Red Hat.
Both build for S390 as release architectures and have access to real hardware so they can cross-check much better.

jonathan-albrecht-ibm:
Understood, thanks for taking the time to consider it.

Issue #480, opened 2021-01-15
Architecture + uname + LSB meter


BenBE:
I find myself regularly in need of knowing the information provided in `uname -a` in particular the running kernel's version and architecture for a system. As we already have similar "static meters" for the current hostname it would be fitting to have a similar meter to show the mentioned information of running kernel type/version, LSB release, and architecture.

An example meter may look like:
`System: GNU/Linux 5.10.7-1-generic [amd64] / Ubuntu 20.04 [focal]`

If the LSB part can't be read, just skip it:
`System: GNU/Linux 5.10.7-1-generic [amd64]`

ahgamut:
I'd like to try this using the [`utsname`](https://man7.org/linux/man-pages/man2/uname.2.html) struct.

Can someone assign this issue to me?

fasterit:
@ahgamut done. Good luck.

ahgamut:
Thanks! I've submitted a PR (#493).

PR #493, opened 2021-01-27
[#480] SysarchMeter to view kernel version/arch info

ahgamut:
`SysarchMeter` calls the [`uname`](https://man7.org/linux/man-pages/man2/uname.2.html) function to obtain the kernel version and
architecture in a `utsname` variable. If the `version` data of the `utsname` is not empty, the distro version is also printed.

`SysarchMeter` follows the style of 
* `HostnameMeter` (uses the same attributes) and 
* `DiskIOMeter` (uses `snprintf` to print to buffer).

Used `astyle -r -xb -s3 -p -xg -c -k1 -W1 \*.c \*.h` to format the two newly added files.

Sample Screenshot:
Screenshot from 2021-01-27 15-27-00

### DIFF
--- a/SysArchMeter.c
+++ b/SysArchMeter.c
@@ -1,0 +1,83 @@
+/*
+htop - SysArchMeter.c
+(C) 2021 htop dev team
+Released under the GNU GPLv2, see the COPYING file
+in the source distribution for its full text.
+*/
+#include "config.h"  // IWYU pragma: keep
+
+#include "SysArchMeter.h"
+
+#include <stdio.h>
+#include <string.h>
+#include <sys/utsname.h>
+
+#include "XUtils.h"
+
+
+static const int SysArchMeter_attributes[] = {HOSTNAME};
+
+static void SysArchMeter_updateValues(Meter* this, char* buffer, size_t size) {
+   static struct utsname uname_info;
+   static int uname_result;
+   static char distro[3][64] = { {'\0'}, {'\0'}, {'\0'} };
+   static bool loaded_data = false;
+
+   (void)this;
+
+   if(!loaded_data) {
+      uname_result = uname(&uname_info);
+      FILE* fp = popen("lsb_release --id --release --codename", "r");
+      if(fp) {
+         char line[96] = {'\0'};
+         size_t n = 0;
+
+         while(fgets(line, sizeof(line), fp)) {
+            n = strcspn(line, ":");
+            if(n > 0 && (n + 1) < strlen(line)) {
+               char* value = String_trim(&line[n + 1]);
+               line[n] = '\0';
+
+               if(String_eq(line, "Distributor ID"))
+                  snprintf(distro[0], sizeof(distro[0]), "%s", value);
+               else if(String_eq(line, "Release"))
+                  snprintf(distro[1], sizeof(distro[1]), "%s", value);
+               else if(String_eq(line, "Codename"))
+                  snprintf(distro[2], sizeof(distro[2]), "%s", value);
+
+               free(value);
+            }
+         }
+         if(!distro[0][0])
+            snprintf(distro[0], sizeof(distro[0]), "Unknown");
+         pclose(fp);
+      }
+      loaded_data = true;
+   }
+
+   if(uname_result == 0) {
+      if (distro[1][0] && distro[2][0])
+         snprintf(buffer, size, "%s %s [%s] / %s %s (%s)", uname_info.sysname, uname_info.release, uname_info.machine, distro[0], distro[1], distro[2]);
+      else if(distro[1][0])
+         snprintf(buffer, size, "%s %s [%s] / %s %s", uname_info.sysname, uname_info.release, uname_info.machine, distro[0], distro[1]);
+      else
+         snprintf(buffer, size, "%s %s [%s]", uname_info.sysname, uname_info.release, uname_info.machine);
+   } else {
+      snprintf(buffer, size, "Unknown");
+   }
+}
+
+const MeterClass SysArchMeter_class = {
+   .super = {
+      .extends = Class(Meter),
+      .delete = Meter_delete
+   },
+   .updateValues = SysArchMeter_updateValues,
+   .defaultMode = TEXT_METERMODE,
+   .maxItems = 0,
+   .total = 100.0,
+   .attributes = SysArchMeter_attributes,
+   .name = "System",
+   .uiName = "System",
+   .caption = "System: ",
+};
--- a/openbsd/Platform.c
+++ b/openbsd/Platform.c
@@ -37,6 +37,7 @@ in the source distribution for its full text.
 #include "Settings.h"
 #include "SignalsPanel.h"
 #include "SwapMeter.h"
+#include "SysArchMeter.h"
 #include "TasksMeter.h"
 #include "UptimeMeter.h"
 #include "XUtils.h"
@@ -99,6 +100,7 @@ const MeterClass* const Platform_meterTypes[] = {
    &UptimeMeter_class,
    &BatteryMeter_class,
    &HostnameMeter_class,
+   &SysArchMeter_class,
    &AllCPUsMeter_class,
    &AllCPUs2Meter_class,
    &AllCPUs4Meter_class,
--- a/Makefile.am
+++ b/Makefile.am
@@ -67,6 +67,7 @@ myhtopsources = \
 	Settings.c \
 	SignalsPanel.c \
 	SwapMeter.c \
+	SysArchMeter.c \
 	TasksMeter.c \
 	TraceScreen.c \
 	UptimeMeter.c \
@@ -121,6 +122,7 @@ myhtopheaders = \
 	Settings.h \
 	SignalsPanel.h \
 	SwapMeter.h \
+	SysArchMeter.h \
 	TasksMeter.h \
 	TraceScreen.h \
 	UptimeMeter.h \
--- a/SysArchMeter.h
+++ b/SysArchMeter.h
@@ -1,0 +1,14 @@
+#ifndef HEADER_SysArchMeter
+#define HEADER_SysArchMeter
+/*
+htop - SysArchMeter.h
+(C) 2021 htop dev team
+Released under the GNU GPLv2, see the COPYING file
+in the source distribution for its full text.
+*/
+#include "Meter.h"
+
+
+extern const MeterClass SysArchMeter_class;
+
+#endif
--- a/darwin/Platform.c
+++ b/darwin/Platform.c
@@ -31,6 +31,7 @@ in the source distribution for its full text.
 #include "MemoryMeter.h"
 #include "ProcessLocksScreen.h"
 #include "SwapMeter.h"
+#include "SysArchMeter.h"
 #include "TasksMeter.h"
 #include "UptimeMeter.h"
 #include "zfs/ZfsArcMeter.h"
@@ -93,6 +94,7 @@ const MeterClass* const Platform_meterTypes[] = {
    &TasksMeter_class,
    &BatteryMeter_class,
    &HostnameMeter_class,
+   &SysArchMeter_class,
    &UptimeMeter_class,
    &AllCPUsMeter_class,
    &AllCPUs2Meter_class,
--- a/dragonflybsd/Platform.c
+++ b/dragonflybsd/Platform.c
@@ -19,6 +19,7 @@ in the source distribution for its full text.
 #include "DateMeter.h"
 #include "DateTimeMeter.h"
 #include "HostnameMeter.h"
+#include "SysArchMeter.h"
 #include "DragonFlyBSDProcess.h"
 #include "DragonFlyBSDProcessList.h"
 
@@ -85,6 +86,7 @@ const MeterClass* const Platform_meterTypes[] = {
    &UptimeMeter_class,
    &BatteryMeter_class,
    &HostnameMeter_class,
+   &SysArchMeter_class,
    &AllCPUsMeter_class,
    &AllCPUs2Meter_class,
    &AllCPUs4Meter_class,
--- a/freebsd/Platform.c
+++ b/freebsd/Platform.c
@@ -42,6 +42,7 @@ in the source distribution for its full text.
 #include "ProcessList.h"
 #include "Settings.h"
 #include "SwapMeter.h"
+#include "SysArchMeter.h"
 #include "TasksMeter.h"
 #include "UptimeMeter.h"
 #include "XUtils.h"
@@ -103,6 +104,7 @@ const MeterClass* const Platform_meterTypes[] = {
    &UptimeMeter_class,
    &BatteryMeter_class,
    &HostnameMeter_class,
+   &SysArchMeter_class,
    &AllCPUsMeter_class,
    &AllCPUs2Meter_class,
    &AllCPUs4Meter_class,
--- a/linux/Platform.c
+++ b/linux/Platform.c
@@ -49,6 +49,7 @@ in the source distribution for its full text.
 #include "SELinuxMeter.h"
 #include "Settings.h"
 #include "SwapMeter.h"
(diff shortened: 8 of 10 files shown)


ahgamut:
I've updated all the `Platform.c` files except for `unsupported/Platform.c`. Should I change it for that as well?

The [`man` page for `uname`](https://man7.org/linux/man-pages/man2/uname.2.html) says that `uname` conforms to:

```
POSIX.1-2001, POSIX.1-2008, SVr4.  There is no uname() call in 4.3BSD.
```

Does that affect any of the `Platform.c` files updated in these commits?

BenBE:
Thank you for writing this PR. On first glance the code looks good and reasonable with a few details to still polish. As I'm a bit short on time right now, just a rough overview of what I saw; will take another look later.

- The copyright header has the wrong year. ;-) Even if you could argue, that we somehow entered an infinite 2020 at some point in March … :P
- After the last `#include` there are usually two blank lines before the actual code part of the file.
- In your output of the meter have a closer look at the original example again. For your system the output of the meter should probably end in `/ Debian 10 [buster]`. Cf. [lsb_release](https://refspecs.linuxfoundation.org/LSB_3.0.0/LSB-Core-generic/LSB-Core-generic.html#LSBRELEASE) for more details. I hope this also explains the note "skip this part if not available" in the original description.
- I wonder if it should be `SysArchMeter`? (up-case A)

Feel free to squash and amend your commits on your branch as needed. Having things in as few independent commits as reasonable, without loosing logical separation of steps is much preferred.

BenBE:
> I've updated all the `Platform.c` files except for `unsupported/Platform.c`. Should I change it for that as well?

Sure.

Just feel free to implement the meter for the "unsupported" platform by hard-coding something like `Unknown Platform` with a comment near it reminding the developer to replace this with an implementation based on e.g. `uname`/`lsb_release` (or similar for that platform) as required for the actual platform. The "Unknown" platform is more or less a template for people who want to port `htop` to some new architecture/OS.

> The [`man` page for `uname`](https://man7.org/linux/man-pages/man2/uname.2.html) says that `uname` conforms to:
> 
> ```
> POSIX.1-2001, POSIX.1-2008, SVr4.  There is no uname() call in 4.3BSD.
> ```
> 
> Does that affect any of the `Platform.c` files updated in these commits?

Nope. For all practical purposes we can assume `uname` exists on all sensible, supported platforms.

FWIW: I'll happily review a PR with a minimal patch to make this work by the first person to demonstrate via video how he got the code compiling on plain 4.3BSD with an C99 compliant compiler … ;-)

ahgamut:
>     The copyright header has the wrong year. ;-)   
>     Even if you could argue, that we somehow entered an infinite 2020 at some point in March … :P

I'm still stuck in 2020 xD, but I changed the files to have 2021.
 
>     After the last `#include` there are usually two blank lines before the actual code part of the file.

`HostnameMeter.h` has only one line after the last `#include`, so I just followed that.

>    In your output of the meter have a closer look at the original example again. For your system the output of the meter should probably end in `/ Debian 10 [buster]`. Cf. [lsb_release](https://refspecs.linuxfoundation.org/LSB_3.0.0/LSB-Core-generic/LSB-Core-generic.html#LSBRELEASE) for more details. I hope this also explains the note "skip this part if not available" in the original description.

Now `lsb_release` is called to obtain the distro, instead of using `utsname.version`.

Screenshot from 2021-01-27 18-07-06


>    I wonder if it should be `SysArchMeter`? (up-case A)

Changed to `SysArchMeter`.

fasterit:
None of this will change during the runtime of htop. So get the data once and be good with it?

ahgamut:
Okay, that means I will need to use `static` variables to store the obtained info, resulting in something like:

```c
static const int SysArchMeter_attributes[] = {HOSTNAME};
// use has_data to check if other variables have been loaded
static struct utsname uname_info;
static char distro_info[256];
static int uname_result;
static bool has_data = false;
```

Is there a way to do this without `static` variables?

openbsd/Platform.c
```diff
@@ -38,6 +38,7 @@ in the source distribution for its full text.
 #include "SignalsPanel.h"
 #include "SwapMeter.h"
 #include "TasksMeter.h"
+#include "SysArchMeter.h"
```
BenBE, line 41: Include sort order

SysArchMeter.c
```diff
@@ -0,0 +1,70 @@
...
+                  uname_info.machine);
+      } else {
+         snprintf(buffer, size, "%s %s [%s] / %s", uname_info.sysname, uname_info.release,
+                  uname_info.machine, distro);
+      }
+   }
+   else
+   {
+      snprintf(buffer, size, "Unknown (error)");
+   }
```
BenBE, line 54: Inconsistent use of braces. Cf. style guide for details. Usually for single line statements braces are left out, if they are put they go on the same line.

BenBE (changes requested):

BenBE:
> None of this will change during the runtime of htop. So get the data once and be good with it?

The LSB information /may/ change*. but even then it would suffice fetching it once every 5 minutes.

*When the package `lsb-release` is updated.

SysArchMeter.c
```diff
@@ -0,0 +1,64 @@
...
+static const int SysArchMeter_attributes[] = {HOSTNAME};
+
+static void SysArchMeter_updateValues(Meter* this, char* buffer, size_t size) {
+   (void)this;
+
+   struct utsname name;
+   char distro_info[256];
+
+   int res = uname(&name);
+   FILE* fp = popen("lsb_release -d -s", "r");
```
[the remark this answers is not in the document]
ahgamut: Output of `lsb_release -ds`:

```
Debian GNU/Linux 10 (buster)
```

Output of `lsb_release -irc`

```
Distributor ID:	Debian
Release:	10
Codename:	buster
```

Single line was easier to parse (I just removed the newline).
If `lsb_release -irc` is better for most distros, we can change it to that.

SysArchMeter.c
```diff
@@ -0,0 +1,64 @@
...
+      pclose(fp);
+   }
+
+   if(res == 0)
+   {
+      if (strlen(distro_info) == 0) {
+         snprintf(buffer, size, "%s-%s [%s]", name.sysname, name.release,
+                  name.machine);
+      } else {
+         snprintf(buffer, size, "%s-%s [%s] / %s", name.sysname, name.release,
```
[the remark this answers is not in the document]
ahgamut: So just each output separated by a space. Got it.

SysArchMeter.c
ahgamut: Is there an(other) automatic way to check if the code conforms to the style guide?
I was using `astyle -r -xb -s3 -p -xg -c -k1 -W1 \*.c \*.h` as provided in the [guide](https://github.com/htop-dev/htop/blob/master/docs/styleguide.md),  but it appears that alone is insufficient.

ahgamut:
Thank you for your guidance @BenBE!
I'm summarizing the requested changes:

- [ ] `lsb_release -ds` (single line) or `lsb_release -ircs` (requires `String_split`)
- [x] used `sizeof` instead of a hardcoded number
- [x] changed dashes to spaces when printing kernel and version
- [x] sort order and braces usage (conforming to style guide)
- [ ] LSB information may change -- read just once, or check every 5 minutes?

I'll push from my local system after confirming preferences on the above.

fasterit:
>     * LSB information may change -- read just once, or check every 5 minutes?

Just once. If Ubuntu fixes a typo in their LSB package we can live with htop needing a restart.
If real LSB info changes (release version) people need a restart anyways which means htop will also (need to be) restarted.

SysArchMeter.c
```diff
@@ -0,0 +1,70 @@
...
+         snprintf(buffer, size, "%s %s [%s]", uname_info.sysname, uname_info.release,
+                  uname_info.machine);
+      } else {
+         snprintf(buffer, size, "%s %s [%s] / %s", uname_info.sysname, uname_info.release,
+                  uname_info.machine, distro);
+      }
+   }
+   else
+   {
+      snprintf(buffer, size, "Unknown (error)");
```
fasterit, line 53: I'd remove the " (error)" as it is quite expected on non-Linux / non-BSD to not be able to identify the system with utsname / lsb_release.

ahgamut: Got it.

fasterit (approved):
Looks, good. Thank you!

BenBE:
@fasterit, @ginggs, @natoscott: Maybe add `Suggests: lsb-release` to debian/control when packaging.

SysArchMeter.c
```diff
@@ -0,0 +1,74 @@
...
+   static char distro[256] = {'\0'};
+   static bool loaded_data = false;
+
+   (void)this;
+
+   if(!loaded_data)
+   {
+      uname_result = uname(&uname_info);
+      FILE* fp = popen("lsb_release --id --release --codename --short", "r");
+      if(fp != NULL)
```
BenBE, line 32: No explicit `!= NULL` necessary in check condition.

SysArchMeter.c
```diff
@@ -0,0 +1,74 @@
...
+      {
```
BenBE, line 33: opening braces should go on the line with the if statement.

SysArchMeter.c
```diff
@@ -0,0 +1,74 @@
...
+         if(fread(distro, sizeof(*distro), sizeof(distro), fp))
+         {
+            size_t num_lines = 0;
+            char** d_split = String_split(distro, '\n', &num_lines);
+            distro[0] = '\0';
+            if(num_lines == 3) // lsb_release returns exactly 3 lines
+               snprintf(distro, sizeof(distro), "%s %s (%s)", d_split[0], d_split[1], d_split[2]);
```
BenBE, line 40: Not the most robust way to do it, even though it gets the job done.

More conservative approaches would be e.g. (a) using the non-short output and parsing the information based on the prefix (susceptible to changes to those strings) or (b) calling the version with just the short flag and one option each, which makes sure we could detect missing information.

Another issue may arise with the `String_split` and the final `\n`in the process output causing actually 4 strings to be returned by splitting. Thus the check should probably be for 3 lines or more with just caring for the content of the first 3 lines of output.

SysArchMeter.c
```diff
@@ -0,0 +1,74 @@
...
+            String_freeArray(d_split);
+         }
+         pclose(fp);
+      }
+      loaded_data = true;
+   }
+
+   if(uname_result == 0)
+   {
```
BenBE, line 49: Opening brace on same line as `if` statement.
```suggestion
   if(uname_result == 0) {
```

SysArchMeter.c
```diff
@@ -0,0 +1,74 @@
...
+      if (strlen(distro) == 0)
+         snprintf(buffer, size, "%s %s [%s]", uname_info.sysname, uname_info.release,
+                  uname_info.machine);
+      else
+         snprintf(buffer, size, "%s %s [%s] / %s", uname_info.sysname, uname_info.release,
+                  uname_info.machine, distro);
```
BenBE, line 55: No need to line-wrap here. Better have one longer line, than arbitrarily word-wrap inside the argument list.
```suggestion
         snprintf(buffer, size, "%s %s [%s]", uname_info.sysname, uname_info.release, uname_info.machine);
      else
         snprintf(buffer, size, "%s %s [%s] / %s", uname_info.sysname, uname_info.release, uname_info.machine, distro);
```

BenBE (changes requested):

BenBE:
Small note on the output of `lsb_release`: The order of the information returned is not guaranteed. Both `lsb_release -irc` and `lsb_release -cri` give the same output.

SysArchMeter.c
ahgamut: > More conservative approaches would be e.g. (a) using the non-short output and parsing the information based on the prefix (susceptible to changes to those strings) or (b) calling the version with just the short flag and one option each, which makes sure we could detect missing information.

For (a), the [reference page](https://refspecs.linuxfoundation.org/LSB_3.0.0/LSB-Core-generic/LSB-Core-generic.html#LSBRELEASE) gives the format for non-short output (`prefix:\t info`). It doesn't say anything about the prefix strings changing, though.

Calling `lsb_release` thrice as in (b) would avoid any ambiguities regarding the order of the output.

> Another issue may arise with the String_split and the final `\n` in the process output causing actually 4 strings to be returned by splitting. Thus the check should probably be for 3 lines or more with just caring for the content of the first 3 lines of output.

Fixed.

SysArchMeter.c
ahgamut: > No need to line-wrap here. Better have one longer line, than arbitrarily word-wrap inside the argument list.

That results in lines of 152 and 106 characters respectively:

```c
   if(uname_result == 0) {
      if (lsb_valid)
         snprintf(buffer, size, "%s %s [%s] / %s %s (%s)", uname_info.sysname, uname_info.release, uname_info.machine, distro[0], distro[1], distro[2]);
      else
         snprintf(buffer, size, "%s %s [%s]", uname_info.sysname, uname_info.release, uname_info.machine);
   }
```

Would it be better to line-wrap in groups of 3 instead?

```c
   if(uname_result == 0) {
      if (lsb_valid)
         snprintf(buffer, size, "%s %s [%s] / %s %s (%s)", 
                 uname_info.sysname, uname_info.release, uname_info.machine, 
                 distro[0], distro[1], distro[2]);
      else
         snprintf(buffer, size, "%s %s [%s]",
                 uname_info.sysname, uname_info.release, uname_info.machine);
   }
```

ahgamut:
> Small note on the output of `lsb_release`: The order of the information returned is not guaranteed. Both `lsb_release -irc` and `lsb_release -cri` give the same output.

`lsb_release` is now called thrice to avoid any ambiguity in the output order.

fasterit:
> `lsb_release` is now called thrice to avoid any ambiguity in the output order.

Please no. Parse the prefixes as we do in htop in virtually anything the kernel exposes in /proc.
You can assume the prefixes to stay. At least as much as the command line switches.

SysArchMeter.c
BenBE: There are several other lines, that are similarly long. Your idea works too.

SysArchMeter.c
```diff
@@ -0,0 +1,84 @@
...
+
+         lsb_valid = (num_lines >= 3);
+         pclose(fp);
+      }
+      loaded_data = true;
+   }
+
+   if(uname_result == 0) {
+      if (lsb_valid)
+         snprintf(buffer, size, "%s %s [%s] / %s %s (%s)", uname_info.sysname, uname_info.release, uname_info.machine, distro[0], distro[1], distro[2]);
```
BenBE, line 63: Better skip the `lsb_valid` check in favor of just checking for both DistID & ReleaseVer vs DistID, ReleaseVer & codename populated.
```suggestion
      if (distro[1][0] && distro[2][0])
         snprintf(buffer, size, "%s %s [%s] / %s %s (%s)",
            uname_info.sysname, uname_info.release, uname_info.machine,
            distro[0][0] ? distro[0] : "Unknown", distro[1], distro[2]);
      else if (distro[1][0])
         snprintf(buffer, size, "%s %s [%s] / %s %s",
            uname_info.sysname, uname_info.release, uname_info.machine,
            distro[0][0] ? distro[0] : "Unknown", distro[1]);
```

Accounts for most cases with the distribution or its release being unknown, as long as there's a version available.

BenBE (changes requested):
I think we can do without the `lsb_valid` check and also keep the flexibility, that not every distribution release may have a codename assigned.

BenBE (approved):
LGTM.

Thank you very much for your work on this feature.

ahgamut:
It's awesome that I got to contribute to `htop`.
Thanks @BenBE and @fasterit for your patience!

BenBE:
Always glad to see new people with their contributions and ideas. :)

PR #495, opened 2021-01-27
Process field overhaul and FreeBSD cleanup

cgzones:


BenBE (approved):
This PR is a mess given the amount of changes packed together. Splitting this in two would actually been much more helpful for reviewing.

This being said. the overall changes LGTM. As this would otherwise conflict with #388 I'll ACK this for now. Also having multiple patches touching similar things (`String_eq` conversions, idle time handling, kernel thread handling, …) doesn't make it more easy to review. Please keep upcoming PRs smaller and more clean.

PR #445, opened 2021-01-01
Add ability to change the header format

cgzones:
Instead of just two equals sized columns, add the ability to select from predefined formats.

**NOTE**: Please backup your htop configuration (*~/.config/htop/htoprc*) before testing, cause after running this modified htop and reverting back to master, the configuration file becomes invalid!

PR #497, opened 2021-01-28
Leave less right margin next to long (text) meters

fasterit:
Esp. visible with "Leave a margin around header" setting and a long text meter in the right column, e.g. the system (arch) meter from #493

### DIFF
--- a/Header.c
+++ b/Header.c
@@ -156,7 +156,7 @@ void Header_draw(const Header* this) {
    for (int y = 0; y < height; y++) {
       mvhline(y, 0, ' ', COLS);
    }
-   const int width = COLS / this->nrColumns - (pad * this->nrColumns - 1) - 1;
+   const int width = COLS / this->nrColumns - pad;
    int x = pad;
 
    Header_forEachColumn(this, col) {


fasterit:
for @BenBE :-) ... problem visual:
grafik

Seems the patch will make the blank line between header and the pane space go away on some window resizes. So this may need a bit more tweaking.

Header.c
```diff
@@ -156,7 +156,7 @@ void Header_draw(const Header* this) {
    for (int y = 0; y < height; y++) {
       mvhline(y, 0, ' ', COLS);
    }
-   const int width = COLS / this->nrColumns - (pad * this->nrColumns - 1) - 1;
+   const int width = COLS / this->nrColumns - pad;
```
BenBE, line 159: Isn't the padding around the header applied left & right (thus `-(2*pad)`?

fasterit: I think it only "applies" to the right margin as the left one is set by offsetting the x location on write.
`actualWidth` a few lines below accounts for both margins, too, I think.

We can also merge #445 instead. I'd still do

```diff
diff --git a/Header.c b/Header.c
index ad73586..15e255a 100644
--- a/Header.c
+++ b/Header.c
@@ -206,7 +206,7 @@ void Header_draw(const Header* this) {
    for (int y = 0; y < height; y++) {
       mvhline(y, 0, ' ', COLS);
    }
-   const int width = COLS - 2 * pad;
+   const int width = COLS - pad;
    int x = pad;
    float roundingLoss = 0.0f;
```
there as well. Better to cut into the right margin than miss a character printed for two blank spaces of margin on the right.

fasterit: Just FTR: We don't do two blanks of margin on the right of the processes panel either. So we should not be overly strict on it for the header.

BenBE:
I'd prefer #445

PR #498, opened 2021-01-28
Autotools follow-up

cgzones:


BenBE (approved):
Looks as sane as Autotools can get …

PR #500, opened 2021-01-29
Quote SYSCONFDIR definition

cgzones:
Also drop related TODO, cause we indeed using the correct way of getting
$sysconfdir from autoconf

### DIFF
--- a/Settings.c
+++ b/Settings.c
@@ -424,10 +424,7 @@ Settings* Settings_new(int initialCpuCount) {
    }
    if (!ok) {
       this->changed = true;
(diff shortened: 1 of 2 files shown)


Settings.c
```diff
@@ -424,7 +424,6 @@ Settings* Settings_new(int initialCpuCount) {
    }
    if (!ok) {
       this->changed = true;
-      // TODO: how to get SYSCONFDIR correctly through Autoconf?
       char* systemSettings = String_cat(SYSCONFDIR, "/htoprc");
```
BenBE, line 427: if this is a string literal, we can skip this `String_cat` …

PR #518, opened 2021-02-07
Remove force sort order to ASC when returning to tree mode

fasterit:
Bug found by @BenBE via IRC

Quote from IRC:
> Setup: Go to tree view, sort DESC by RES, hit F5, sort by DESC
> Test: Press F5 repeatedly to change between list view and tree view.
> Expectation: Should stay in DESC order for both views
> Actual: Switches to ASC when re-entering tree view mode.

### DIFF
--- a/Action.c
+++ b/Action.c
@@ -226,9 +226,6 @@ static Htop_Reaction actionToggleMergedCommand(State* st) {
 
 static Htop_Reaction actionToggleTreeView(State* st) {
    st->settings->treeView = !st->settings->treeView;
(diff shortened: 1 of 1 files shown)


BenBE:
I'm not sure right now, but doesn't this re-introduce another issue with the sorting we recently fixed? Aim/intention for the test was to check if sort orders for list/tree mode are handled as separate state.

Having tree mode return to ASC would be absolutely correct, if it was previously active in tree mode, regardless of the list mode setting.

Issue #104, opened 2020-09-07
Feature request: Account for zswap in memory usage


SharkWipf:
This may be out of scope for htop, but currently htop has no way of accounting for zswap usage.
Unlike zram, zswap does not create a virtual, dynamically resizing swap device, nor does it report the compressed size to the OS through the normal channels.  
Instead, it reports the _uncompressed_ size to the OS, and keeps the real stats to itself in `/sys/kernel/debug/zswap/`.
As a result, the uncompressed swap usage can exceed the real swap device size, the OS will report the actual swap device as full (despite it possibly not being touched by swap at all), but will continue filling up zswap in the background with no way to see it in `htop` (or `free`, for that matter):

![image](https://user-images.githubusercontent.com/751205/92417780-44b5a500-f164-11ea-9de5-a9c71d7f4538.png)

It would be more accurate, albeit potentially out of scope, to account for zswap usage in the swap usage, showing the compressed swap rather than the uncompressed swap.

Skirmisher:
It's unfortunate that there's no way to access zswap statistics without requiring users to have debugfs mounted and readable without root. IMO, even if swap usage is not corrected, a zswap meter (similar to the zram one) would be most immediately useful for assessing the status of system memory/swap on a zswap-enabled system. Perhaps implementing such a thing would make sense even with the extra setup requirement, for the benefit of those who require it. (I'm moving my Fedora system over to zswap from zram, since I have a large swap partition that's being suboptimally used due to zram being full, and came here while looking for a replacement solution for htop's zram meter.)

PR #521, opened 2021-02-09
RFC: Linux: Do not include zram in normal swap

cgzones:
Zram uses the actual RAM as storage, unlike standard swap partitions or
files. It is not an additional memory resource, but a transformation of
an already existing one.

### DIFF
--- a/linux/Platform.c
+++ b/linux/Platform.c
@@ -297,8 +297,10 @@ void Platform_setMemoryValues(Meter* this) {
 
 void Platform_setSwapValues(Meter* this) {
    const ProcessList* pl = this->pl;
-   this->total = pl->totalSwap;
-   this->values[0] = pl->usedSwap;
+   const LinuxProcessList* lpl = (const LinuxProcessList*) pl;
+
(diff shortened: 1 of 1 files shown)


fasterit:
Hm, we have a zram meter.
There may be value in seeing the total swap as what it is, regardless of the backing store.
If somebody puts a swap file onto a tmpfs we don't special account for that either.

BenBE:
One option to differentiate between classical swap and the ZRAM stuff would be having the compressed ZRAM size as an additional value in the RAM/memory meter and the uncompressed ZRAM size as an extra value of the Swap meter.

Like:
```
Mem:23.9G used:13.8G buffers:1.24G cache:6.3G zram-c:23M
Swp:0K used:0K cache:0K zram-x:42M
```

And as that's kinda silly and just confusing users, report both values in the normal Mem meter …

```
Mem:23.9G used:13.8G buffers:1.24G cache:6.3G zram:42M->23M
Swp:0K used:0K cache:0K
```

fasterit:
Related: #104

cgzones:
Closing according to our team discussion.

Issue #527, opened 2021-02-15
not working with btime 0 (uninitialized time)


mndza:
Hello,

Since commit https://github.com/htop-dev/htop/commit/b7836515e8db8ff7a7cb99247472ac10992a4518 `htop` does not work on my embedded system if I did not set the time first (used to work fine). 

The problem is caused by these lines:
https://github.com/htop-dev/htop/blob/b7836515e8db8ff7a7cb99247472ac10992a4518/linux/LinuxProcessList.c#L246-L247

I think the main mistake here is confusing the 0 with an uninitialized value, while the error catching before was done with the return codes of the read calls (`sscanf`, `fgets`, ...).

natoscott:
@mndza so AIUI your /proc/stat file contains a 'btime 0' line, is that correct?  Thanks.

mndza:
Yes, exactly.

natoscott:
OK, thanks, I'll tweak the code to better handle this case.

PR #528, opened 2021-02-15
Correctly detect failure to initialize boottime

natoscott:
A zero value for btime (boottime) in /proc/stat is a
real situation that happens, so deal with this case.

Resolves https://github.com/htop-dev/htop/issues/527

### DIFF
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -65,7 +65,7 @@ in the source distribution for its full text.
 # define O_PATH 010000000
 #endif
 
-static long long btime;
+static long long btime = -1;
 
 static long jiffy;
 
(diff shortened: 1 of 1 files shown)


linux/LinuxProcessList.c
```diff
@@ -241,7 +241,7 @@ ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* pidMatchList, ui
       }
       fclose(statfile);
 
-      if (!btime)
+      if (btime == -1)
```
BenBE, line 244: Why not use some local variable to record ´btime´ presence and check on that variable instead. Using a magic value here is just prone to break another time when some unlucky guy has a kernel reporting -1 as boot tiime when cosmic rays just hit the wrong way …

fasterit: We can introduce that when solar activity raises significantly :-)

natoscott: It's gonna take more than just gamma rays ...   :)
```
$ grep btime linux/fs/proc/stat.c 
		"btime %llu\n"
```

PR #531, opened 2021-02-16
Enable making with -ggdb3 on configure --enable-debug

fasterit:


configure.ac
```diff
@@ -567,11 +567,13 @@ fi
...
 AC_ARG_ENABLE([debug],
               [AS_HELP_STRING([--enable-debug],
-                              [Enable asserts and internal sanity checks @<:@default=no@:>@])],
+                              [Enable compiling with maximum debug info, asserts and internal sanity checks @<:@default=no@:>@])],
               [],
               [enable_debug=no])
 if test "x$enable_debug" != xyes; then
    AM_CPPFLAGS="$AM_CPPFLAGS -DNDEBUG"
+else
+   AM_CPPFLAGS="$AM_CPPFLAGS -ggdb3"
```
BenBE, line 576: Please also include `-Og` (Optimize for debugging).

Supported by both [GCC](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html) and [LLVM](https://github.com/llvm-mirror/clang/commit/14bfc9e99e6e6903b09480a22c153032be77ae4e).

fasterit: > Please also include `-Og` (Optimize for debugging).

That would be overridden by the config from autotools later in the GCC line (typically -O2), "If you use multiple -O options, with or without level numbers, the last such option is the one that is effective." from https://gcc.gnu.org/onlinedocs/gcc-10.2.0/gcc/Optimize-Options.html

fasterit: I've added it for `make debug` in 40b5156. That way it actually makes sense to have both `./configure --enable-debug` (same O level as prod build) and `make debug` (-Og as another debugging aid).

Makefile.am
```diff
@@ -296,10 +296,10 @@ profile:
 	$(MAKE) all AM_CPPFLAGS="-pg -O2 -DNDEBUG"
 
 debug:
-	$(MAKE) all AM_CPPFLAGS="-ggdb -DDEBUG"
+	$(MAKE) all AM_CPPFLAGS="-ggdb -Og" CFLAGS="$(patsubst -O%,,$(CFLAGS))"
```
cgzones, line 299: Should we use `-ggdb3` here too?

fasterit: yes, makes total sense

BenBE (approved):

PR #533, opened 2021-02-16
SysArchMeter: read os-release instead of running lsb-release

cgzones:
`os-release` is available on FreeBSD by default.
Also avoid executing a third-party program.

Examples:
```
  Linux 5.10.0-3-amd64 [x86_64] @ Debian GNU/Linux bullseye/sid
  FreeBSD 12.2-RELEASE-p3 [amd64]
```

Closes: #516

### DIFF
--- a/SysArchMeter.c
+++ b/SysArchMeter.c
@@ -9,7 +9,6 @@ in the source distribution for its full text.
 #include "SysArchMeter.h"
 
 #include <stdio.h>
-#include <string.h>
 #include <sys/utsname.h>
 
 #include "XUtils.h"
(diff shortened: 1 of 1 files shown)


SysArchMeter.c
```diff
@@ -9,67 +9,75 @@ in the source distribution for its full text.
...
+   }
 
-   (void)this;
+   char name[64] = {'\0'};
+   char version[64] = {'\0'};
+   char lineBuffer[256];
+   while (fgets(lineBuffer, sizeof(lineBuffer), stream)) {
+      if (String_startsWith(lineBuffer, "PRETTY_NAME=\"")) {
+         size_t start = strlen("PRETTY_NAME=\"");
+         size_t len = strrchr(lineBuffer, '"') - lineBuffer - start;
```
BenBE, line 35: This seems off when no second double quote character is in that string.

SysArchMeter.c
```diff
@@ -9,67 +9,75 @@ in the source distribution for its full text.
...
+         String_safeStrncpy(buffer, lineBuffer + start, MINIMUM(bufferLen, len + 1));
+         fclose(stream);
+         return;
+      }
+      if (String_startsWith(lineBuffer, "NAME=\"")) {
+         size_t start = strlen("NAME=\"");
+         size_t len = strrchr(lineBuffer, '"') - lineBuffer - start;
```
BenBE, line 42: dito.

SysArchMeter.c
```diff
@@ -9,67 +9,75 @@ in the source distribution for its full text.
...
+         String_safeStrncpy(name, lineBuffer + start, MINIMUM(sizeof(name), len + 1));
+         continue;
+      }
+      if (String_startsWith(lineBuffer, "VERSION=\"")) {
+         size_t start = strlen("VERSION=\"");
+         size_t len = strrchr(lineBuffer, '"') - lineBuffer - start;
```
BenBE, line 48: dito.

BenBE (changes requested):
`strrchr` may return `NULL` which makes `len` more or less undefined.

SysArchMeter.c
cgzones: Since we matched `PRETTY_NAME=\"`, which contains a `"`, `strrchr()` can not return `NULL` and len might be at least `0`.

BenBE: In which case it still would return `len=-1` …

BenBE (approved):

natoscott:
@cgzones @BenBE LGTM too.  As a followup (I'm happy to do this), it would be good to have this Meter separate its "sampling" logic (value extraction) from its "display" logic - other Meters use Platform_getBlah routines for this.  SysArchMeter does not though and it definitely should since its sampling logic doesn't work as-is on platforms like Darwin or Solaris (nor PCP or Windows).

BenBE:
> @cgzones @BenBE LGTM too.

So, let's merge this as is.

> As a followup (I'm happy to do this), it would be good to have this Meter separate its "sampling" logic (value extraction) from its "display" logic - other Meters use Platform_getBlah routines for this. SysArchMeter does not though and it definitely should since its sampling logic doesn't work as-is on platforms like Darwin or Solaris (nor PCP or Windows).

Sure, go ahead.

PR #536, opened 2021-02-17
Add a platform for Performance Co-Pilot (PCP) metrics

natoscott:
This introduces a htop 'platform' for extracting metrics using the [PCP](https://pcp.io) performance metrics API - [PMAPI(3)](https://man7.org/linux/man-pages/man3/PMAPI.3.html).  It can be used via the --enable-pcp=yes configure option.

So far I've added support for live localhost metrics only, and only using pre-defined metrics already found in htop.  If available, all sampling is performed by [pmcd(1)](https://man7.org/linux/man-pages/man1/pmcd.1.html) - else, we fallback to htop doing the metric sampling itself (all handled below the PMAPI interfaces used here).  When pmcd is used, it's usually run children with elevated privileges so this htop platform does not ever need to be setuid, nor does htop itself need to run as root (authentication with pmcd is available).

Additionally, the PMAPI allows us to support archives for historical analysis and for automated regression tests in htop.  We'll need platform-specific command line argument additions, which isn't yet feasible in htop (not difficult to add though).

The goal of this first version is minimal impact in terms of modifying the htop code base, to introduce key ideas in PCP (metric namespace, metadata, APIs and so on) and give us something to discuss, experiment with and build on.


configure.ac
```diff
@@ -95,6 +95,31 @@ esac
...
+   no)
+      ;;
+   yes)
+      AC_CHECK_HEADERS([pcp/pmapi.h], [my_htop_platform=pcp],
+		       [AC_MSG_ERROR([can not find PCP header file])])
+      AC_SEARCH_LIBS([pmNewContext], [pcp], [], [AC_MSG_ERROR([can not find pcp library])])
+      AC_DEFINE([HTOP_PCP], [1], [Define if building pcp htop binary.])
+      ;;
+   *)
+      AC_MSG_ERROR([bad value '$enable_static' for --enable-static option])
```
fasterit, line 117: `$enable_pcp` and `--enable-pcp` (c&p mistake)

pcp/Platform.h
```diff
@@ -0,0 +1,257 @@
...
+#include <ctype.h>
+#include <stdbool.h>
+#include <pcp/pmapi.h>
+
+#undef PACKAGE_URL
+#undef PACKAGE_NAME
+#undef PACKAGE_STRING
+#undef PACKAGE_TARNAME
+#undef PACKAGE_VERSION
+#undef PACKAGE_BUGREPORT
```
BenBE, line 21: Why `#undef` all those defines? Should be commented/documented. Especially as this is in the middle of all the includes.

pcp/Platform.h
```diff
@@ -0,0 +1,257 @@
...
+   PCP_PROC_MEM_LIBRS,		/* proc.memory.librss */
+   PCP_PROC_MEM_DATRS,		/* proc.memory.datrss */
+   PCP_PROC_MEM_DIRTY,		/* proc.memory.dirty */
+
+   PCP_PROC_SMAPS_PSS,		/* proc.smaps.pss */
+   PCP_PROC_SMAPS_SWAP,		/* proc.smaps.swap */
+   PCP_PROC_SMAPS_SWAPPSS,	/* proc.smaps.swappss */
+
+   PCP_METRIC_COUNT		/* total metric count */
+} Metric;
```
BenBE, line 233: Uniform indentation, maybe avoid tabs?

pcp/PCPProcessList.h
```diff
@@ -0,0 +1,70 @@
...
+
+#include <stdbool.h>
+#include <sys/types.h>
+
+#include "Hashtable.h"
+#include "ProcessList.h"
+#include "pcp/Platform.h"
+#include "UsersTable.h"
+#include "zfs/ZfsArcStats.h"
+
```
BenBE, line 20: 2 blank lines.

pcp/PCPProcessList.h
```diff
@@ -0,0 +1,70 @@
+#ifndef HEADER_PCPProcessList
+#define HEADER_PCPProcessList
+/*
+htop - PCPProcessList.h
+(C) 2014 Hisham H. Muhammad
+Released under the GNU GPLv2, see the COPYING file
+in the source distribution for its full text.
+*/
+
+#include "config.h"
```
BenBE, line 10: Possibly add `// IWYU pragma: keep`

pcp/PCPProcessList.c
```diff
@@ -0,0 +1,816 @@
...
+
+   if (Metric_instance(PCP_PROC_MEM_DATRS, pid, offset, &value, PM_TYPE_U32))
+      process->m_drs = value.ul;
+   else
+      process->m_drs = 0;
+
+   if (Metric_instance(PCP_PROC_MEM_DIRTY, pid, offset, &value, PM_TYPE_U32))
+      process->m_dt = value.ul;
+   else
+      process->m_dt = 0;
```
BenBE, line 299: Use some minimal macro to DRY this source portion?

pcp/Platform.c
```diff
@@ -0,0 +1,874 @@
+/*
+htop - linux/Platform.c
+(C) 2014 Hisham H. Muhammad
+(C) 2020-2021 htop dev team
+(C) 2020-2021 Red Hat, Inc.  All Rights Reserved.
+Released under the GNU GPLv2, see the COPYING file
+in the source distribution for its full text.
+*/
+
+#include "config.h"
```
BenBE, line 10: Include `// IWYU pragma: keep`.

pcp/Platform.c
```diff
@@ -0,0 +1,874 @@
...
+#include "ProvideCurses.h"
+#include "Settings.h"
+#include "SwapMeter.h"
+#include "SysArchMeter.h"
+#include "TasksMeter.h"
+#include "UptimeMeter.h"
+#include "XUtils.h"
+#include "linux/PressureStallMeter.h"
+#include "linux/ZramMeter.h"
+#include "linux/ZramStats.h"
```
BenBE, line 43: This should not be included unconditionally, unless we're really compiling PCP on Linux.

pcp/Platform.c
```diff
@@ -0,0 +1,874 @@
...
+#include "Macros.h"
+#include "MainPanel.h"
+#include "MemoryMeter.h"
+#include "Meter.h"
+#include "NetworkIOMeter.h"
+#include "Object.h"
+#include "Panel.h"
+#include "PCPProcess.h"
+#include "PCPProcessList.h"
+#include "Platform.h"
```
BenBE, line 32: Prefer to have relative paths based on main directory, thus including the platform directory.

pcp/Platform.c
```diff
@@ -0,0 +1,874 @@
...
+
+
+typedef struct Platform_ {
+   int context;			/* PMAPI(3) context identifier */
+   unsigned int total;		/* total number of all metrics */
+   const char** names;		/* name array indexed by Metric */
+   pmID* pmids;			/* all known metric identifiers */
+   pmID* fetch;			/* enabled identifiers for sampling */
+   pmDesc* descs;		/* metric desc array indexed by Metric */
+   pmResult* result;		/* sample values result indexed by Metric */
```
BenBE, line 56: IWYU violation: which header provides these types? Should be included as part of this file (unless part of public interface, in which case the public interface header should include these).

pcp/Platform.c
BenBE, line 32: `pcp/Platform.h` should be first include after `config.h`. Cf. style guide.

BenBE (changes requested):

pcp/Platform.h
natoscott: We have a conflict with macros of the same name from PCP config.h - I'll add a comment to that effect.

pcp/Platform.c
natoscott: I think its OK - we could be looking at a remote Linux host, or an archive from Linux, on some other platform.

pcp/Platform.c
natoscott: Hmm?  That's not what other platforms seem to do... e.g. grep ProcessList.h darwin/*c linux/*c

pcp/Platform.c
natoscott: Its <pcp/pmapi.h> - which arrives via pcp/Platform.h ... seems strange to have to include it in the pcp/Platform.c as well...?

pcp/Platform.c
natoscott: Yep re first header - we seem to be a bit inconsistent about the relative path though, e.g. linux/Platform.c does not include the linux/ prefix.

pcp/Platform.c
BenBE: If it's included via `pcp/Platform.h` it's okay, to drop it in the `.c`.

pcp/Platform.c
BenBE: That inconsistency should have been fixed as part of #616 recently.

pcp/Platform.c
BenBE: Mostly cautious because foreign platform headers may include system headers that might not be available for cross-compilation. Like when compiling PCP on FreeBSD. That way this include may reference system headers not available on the actual platform being compiled for (FreeBSD).

natoscott: @BenBE it's a good point.  Fortunately the individual Meters tend to include only Meter.h and then related "core" platform headers for their implementation - and call out to platform-specific code via Platform.h - this is just what we want.  If a platform-specific Meter didn't do this (and couldn't be refactored for some reason), we'd not be able to use that specific Meter for the PCP platform.  That's OK too.

pcp/Platform.c
natoscott: +1 ... I'll update the PCP code shortly to follow that convention.

natoscott: Fixing this in next update - thanks @BenBE

configure.ac
```diff
@@ -95,6 +95,31 @@ esac
...
+              [],
+              [enable_pcp=no])
+case "$enable_pcp" in
+   no)
+      ;;
+   yes)
+      AC_CHECK_HEADERS([pcp/pmapi.h], [my_htop_platform=pcp],
+		       [AC_MSG_ERROR([can not find PCP header file])])
+      AC_SEARCH_LIBS([pmNewContext], [pcp], [], [AC_MSG_ERROR([can not find pcp library])])
+      AC_DEFINE([HTOP_PCP], [1], [Define if building pcp htop binary.])
```
BenBE, line 114: (In)Consistent casing

pcp/PCPProcessList.c
```diff
@@ -0,0 +1,786 @@
...
+
+   if (Metric_instance(PCP_PROC_STARTTIME, pid, offset, &value, PM_TYPE_U64))
+      process->starttime_ctime = PCPProcessList_adjustTime(value.ull);
+   else
+      process->starttime_ctime = 0;
+
+   if (Metric_instance(PCP_PROC_PROCESSOR, pid, offset, &value, PM_TYPE_U32))
+      process->processor = value.ul;
+   else
+      process->processor = 0;
```
BenBE, line 206: This calls for some small macro to reduce the repeated code.

pcp/PCPProcessList.c
```diff
@@ -0,0 +1,786 @@
...
+
+   if (Metric_instance(PCP_PROC_IO_SYSCW, pid, offset, &value, PM_TYPE_U64))
+      process->io_syscw = value.ull;
+   else
+      process->io_syscw = ULLONG_MAX;
+
+   if (Metric_instance(PCP_PROC_IO_CANCELLED, pid, offset, &value, PM_TYPE_U64))
+      process->io_cancelled_write_bytes = value.ull / ONE_K;
+   else
+      process->io_cancelled_write_bytes = ULLONG_MAX;
```
BenBE, line 237: dito

BenBE (changes requested):

pcp/PCPProcessList.c
```diff
@@ -86,29 +86,55 @@ void ProcessList_delete(ProcessList* pl) {
...
-      process->ppid = 1;
+static inline unsigned long long Metric_instance_time(int metric, int pid, int offset) {
+   pmAtomValue value;
+   if (Metric_instance(metric, pid, offset, &value, PM_TYPE_U64))
+      return value.ull / 10;
+   return 0;
+}
+
+static inline unsigned long long Metric_instance_ONE_K(int metric, int pid, int offset) {
+    pmAtomValue value;
```
BenBE, line 118: Indentation

pcp/PCPProcessList.c
```diff
@@ -86,29 +86,55 @@ void ProcessList_delete(ProcessList* pl) {
...
-      process->state = 'X';
+      return uchar;
    }
+   return fallback;
+}
+
+static void PCPProcessList_updateID(Process* process, int pid, int offset) {
+   process->tgid = Metric_instance_u32(PCP_PROC_TGID, pid, offset, 1);
+   process->ppid = Metric_instance_u32(PCP_PROC_PPID, pid, offset, 1);
+   process->state = Metric_instance_char(PCP_PROC_STATE, pid, offset, 'X');
```
BenBE, line 137: Is there some consistency for undefined states across other platforms? I've seen `'?'` being used on Darwin, though. Let's try not to introduce another such set … ;-)

natoscott: @BenBE heh, sure thing - '?' works too, will update shortly.

BenBE (approved):

natoscott:
Merged.

PR #537, opened 2021-02-17
Use unsigned types for CPU counts and associated variables

cgzones:


Affinity.h
```diff
@@ -27,16 +27,16 @@ in the source distribution for its full text.
 
 typedef struct Affinity_ {
    ProcessList* pl;
-   int size;
-   int used;
-   int* cpus;
+   unsigned int size;
+   unsigned int used;
```
BenBE, line 31: Why not even use `size_t` here?

AffinityPanel.c
```diff
@@ -422,12 +422,12 @@ Affinity* AffinityPanel_getAffinity(Panel* super, ProcessList* pl) {
    Affinity* affinity = Affinity_new(pl);
 
    #ifdef HAVE_LIBHWLOC
-   int i;
+   unsigned int i;
```
BenBE, line 425: dito.

AvailableMetersPanel.c
```diff
@@ -113,10 +113,10 @@ AvailableMetersPanel* AvailableMetersPanel_new(Settings* settings, Header* heade
    }
    // Handle (&CPUMeter_class)
    const MeterClass* type = &CPUMeter_class;
-   int cpus = pl->cpuCount;
+   unsigned int cpus = pl->cpuCount;
```
BenBE, line 116: dito.

AvailableMetersPanel.c
```diff
@@ -113,10 +113,10 @@ AvailableMetersPanel* AvailableMetersPanel_new(Settings* settings, Header* heade
...
    if (cpus > 1) {
       Panel_add(super, (Object*) ListItem_new("CPU average", 0));
-      for (int i = 1; i <= cpus; i++) {
+      for (unsigned int i = 1; i <= cpus; i++) {
```
BenBE, line 119: dito.

BenBE (changes requested):

Affinity.h
natoscott: To my eye the 32 bit value is correct in these instances, as opposed to size_t which IIRC is 64 bit on 32 bit platforms.  Certainly in the Linux kernel (and I'll punt all other kernels too) 32 bit integers are used for CPU numbering - seems noone is planning a >4billion CPU count machine anytime soon.

In my (weird) brain at least, size_t is used primarily when dealing with bytes - i.e. sizes of buffers/memory, number of bytes read/written, etc - and not arbitrary entity counts that don't need the larger size.

BTW, not sure when it was added (I missed the PR when it flew past earlier I guess), but with the recent 'memory_t' addition I woulda voted for size_t there rather than a new 64 bit type.  *shrug* - no big deal of course.

natoscott: Oh, forgot to mention - rest of the changes LGTM too - liking the removal of the second pass, and updating of running tasks once more.  Cheers!

BenBE: The addition of `memory_t` was done by @cgzones in 3d497a37. The intention there was to allow for consistent typing of memory/stats counter quantities above 4GiB, which wouldn't be possible with `size_t` on 32-bit platforms. Thus while unintuitive its the correct way to go there.

`size_t` is intended for counting things, where you have an array or other sort of list as backing storage and where this value can/is used to index into that storage, which is the case with the CPUs.

I use `size_t` usually whenever there is a quantity of things where it's unlikely to exceed values of 2^32, or is supported through the platform if it does.

natoscott: Ah OK - yep, fair enough re 32 bit platforms, from googling there's certainly alot of confusion around `size_t` (myself included).

This definition of `size_t` ("intended for counting things") is likely going to match hundreds of other places throughout htop FWIW - far beyond CPUs - that would now be using `int` or `unsigned int`.  And then the next logical question is - should all places "counting things" that can be negative start using `ssize_t`?  This becomes quite a long piece of string to start pulling on and I'm not sure its going to be an overall win for the code base?  (lotsa code churn)

Thanks for the `memory_t` note too, good to know.  Last thing I'll say there is its not really clear then (to me at least) why `uint64_t` wasn't used like in many other places in htop already.  But its no big deal - I'm not advocating strongly for change there now ... though I'd ideally not want to see a proliferation of types like `cpu_t`, and so on for counting/sizing various aspects of other things.  From my POV, I like to be able to look at code and say "I know exactly what that type is" - things like `memory_t` etc add just a little obfuscation I could live without.  (*shrug*)

cgzones:
Applied the first commit (/proc/stat parsing related) via https://github.com/htop-dev/htop/commit/0cfc9b09805679cf16bec30b0dc7a769a269d628

BenBE:
@cgzones What's missing for this PR to go life?

natoscott (approved):
LGTM as-is.

PR #543, opened 2021-02-26
Export symbols for --enable-debug

BenBE:
Please carefully review, the following patches, as I might have modified some mentions of `static` that should have been kept untouched.


configure.ac
```diff
@@ -562,9 +562,14 @@ AC_ARG_ENABLE([debug],
               [enable_debug=no])
 if test "x$enable_debug" != xyes; then
    AM_CPPFLAGS="$AM_CPPFLAGS -DNDEBUG"
+   AC_DEFINE([SYM_PRIVATE], [static], [Control visibility of symbols to be used only inside a compilation unit.])
+   AX_CHECK_COMPILE_FLAG([-Wno-error=missing-prototypes], [AM_CFLAGS="$AM_CFLAGS -Wno-error=missing-prototypes"], , [-Werror])
+   AX_CHECK_COMPILE_FLAG([-Wno-missing-prototypes], [AM_CFLAGS="$AM_CFLAGS -Wno-missing-prototypes"], , [-Werror])
```
cgzones, line 567: These compiler flag additions are in the wrong branch, see the CI failure.

cgzones (commented):
I am not a big fan of these changes (wrapping standard C keywords into macros).

Does anyone know if libunwind produces readable backtraces with entries for static functions?

configure.ac
BenBE: Thanks. Fixed. I wondered what went wrong, but couldn't see it …

BenBE:
> I am not a big fan of these changes (wrapping standard C keywords into macros).
> 
> Does anyone know if libunwind produces readable backtraces with entries for static functions?

Problem is, that as quoted in the manpages for `backtrace(3)` (emphasis mine):
> The symbol names may be unavailable without the use of special linker options. For systems using the GNU linker, it is necessary to use the `-rdynamic` linker option. **Note that names of `static` functions are not exposed, and won't be available in the backtrace.**

Thus, the problem is that these symbols aren't exported if `static` is used. I couldn't find if this also affects `libunwind` so this needs to be tested.

@cgzones If you provide a patch to use `libunwind` when available this might be a way around exporting all symbols in debug mode. But this needs to be tested AND will need a solid fallback if `libunwind` is not available.

cgzones:
On FreeBSD `backtrace_symbols_fd(3)` does not work, probably because it is not safe to use in signal handlers (see https://stackoverflow.com/questions/77005/how-to-automatically-generate-a-stacktrace-when-my-program-crashes).
Going to test libunwind...

Issue #408, opened 2020-12-16
graphical mode for network meter


eworm-de:
Currently the network meter does not show anything in graphical mode. Please make graphical mode available with reasonable scaling on y-axis, possibly dynamic depending on recent spikes.

Nudin:
I currently give this a try, seems to be easy to fix.

PR #544, opened 2021-02-26
Implement graph and bar for NetworkIOMeter & rescale graphs when total changed

Nudin:
NetworkIOMeter currently only works in text and LED mode, bar and graph mode were not supported meaning they stayed empty (#408). This PR changes that. Graph and bar are plotted as they are for other meters. The total is set to 100kib/s initially and continuously changed to the maximum value seen so far.
While implementing this, I noticed an unexpected behavior of the graph display method: When the total is changed the new values will be displayed in respect to that, but the old values will continue to be displayed relative to the old total value. Through this the graph is rendered incorrectly. This bug does not only occur to the new graph of NetworkIOMeter, but currently already affects  LoadAverageMeter, DiskIOMeter, UptimeMeter & TasksMeter. The first patch in this PR fixes this issue for all graphs.

### DIFF
--- a/Meter.c
+++ b/Meter.c
@@ -334,7 +334,6 @@ static void GraphMeterMode_draw(Meter* this, int x, int y, int w) {
       double value = 0.0;
       for (uint8_t i = 0; i < this->curItems; i++)
          value += this->values[i];
-      value /= this->total;
       data->values[nValues - 1] = value;
    }
 
@@ -345,8 +344,10 @@ static void GraphMeterMode_draw(Meter* this, int x, int y, int w) {
    }
    for (; i < nValues - 1; i += 2, k++) {
       int pix = GraphMeterMode_pixPerRow * GRAPH_HEIGHT;
-      int v1 = CLAMP((int) lround(data->values[i] * pix), 1, pix);
-      int v2 = CLAMP((int) lround(data->values[i + 1] * pix), 1, pix);
+      if (this->total < 1)
+         this->total = 1;
+      int v1 = CLAMP((int) lround(data->values[i] / this->total * pix), 1, pix);
+      int v2 = CLAMP((int) lround(data->values[i + 1] / this->total * pix), 1, pix);
 
       int colorIdx = GRAPH_1;
       for (int line = 0; line < GRAPH_HEIGHT; line++) {
--- a/NetworkIOMeter.c
+++ b/NetworkIOMeter.c
@@ -24,7 +24,7 @@ static unsigned long int cached_rxp_diff = 0;
 static unsigned long int cached_txb_diff = 0;
 static unsigned long int cached_txp_diff = 0;
 
-static void NetworkIOMeter_updateValues(ATTR_UNUSED Meter* this, char* buffer, size_t len) {
+static void NetworkIOMeter_updateValues(Meter* this, char* buffer, size_t len) {
    static unsigned long long int cached_last_update = 0;
 
    struct timeval tv;
@@ -80,6 +80,12 @@ static void NetworkIOMeter_updateValues(ATTR_UNUSED Meter* this, char* buffer, s
       cached_txp_total = packetsTransmitted;
    }
 
+   this->values[0] = cached_rxb_diff;
+   this->values[1] = cached_txb_diff;
+   if (cached_rxb_diff + cached_txb_diff > this->total) {
+      this->total = cached_rxb_diff + cached_txb_diff;
+   }
+
    char bufferBytesReceived[12], bufferBytesTransmitted[12];
    Meter_humanUnit(bufferBytesReceived, cached_rxb_diff, sizeof(bufferBytesReceived));
    Meter_humanUnit(bufferBytesTransmitted, cached_txb_diff, sizeof(bufferBytesTransmitted));
@@ -116,7 +122,7 @@ const MeterClass NetworkIOMeter_class = {
    },
    .updateValues = NetworkIOMeter_updateValues,
    .defaultMode = TEXT_METERMODE,
-   .maxItems = 0,
+   .maxItems = 2,
    .total = 100.0,
    .attributes = NetworkIOMeter_attributes,
    .name = "NetworkIO",


BenBE (approved):

Meter.c
```diff
@@ -345,8 +344,8 @@ static void GraphMeterMode_draw(Meter* this, int x, int y, int w) {
    }
    for (; i < nValues - 1; i += 2, k++) {
       int pix = GraphMeterMode_pixPerRow * GRAPH_HEIGHT;
-      int v1 = CLAMP((int) lround(data->values[i] * pix), 1, pix);
-      int v2 = CLAMP((int) lround(data->values[i + 1] * pix), 1, pix);
+      int v1 = CLAMP((int) lround(data->values[i] / this->total * pix), 1, pix);
```
fasterit, line 347: This should probably get protection against division-by-zero as meters update "total" dynamically.
Of course meters should not set a total to zero but rounding happens and better safe than sorry.
Same for the `v2` line below.

NetworkIOMeter.c
```diff
@@ -80,6 +80,12 @@ static void NetworkIOMeter_updateValues(ATTR_UNUSED Meter* this, char* buffer, s
       cached_txp_total = packetsTransmitted;
    }
 
+   this->values[0] = cached_rxb_diff;
+   this->values[1] = cached_txb_diff;
+   if( cached_rxb_diff + cached_txb_diff > this->total ) {
```
fasterit, line 85: Styleguide
```suggestion
   if (cached_rxb_diff + cached_txb_diff > this->total) {
```

fasterit (commented):
thank you very much for the PR, please review the change requests

Meter.c
```diff
@@ -345,8 +344,11 @@ static void GraphMeterMode_draw(Meter* this, int x, int y, int w) {
    }
    for (; i < nValues - 1; i += 2, k++) {
       int pix = GraphMeterMode_pixPerRow * GRAPH_HEIGHT;
-      int v1 = CLAMP((int) lround(data->values[i] * pix), 1, pix);
-      int v2 = CLAMP((int) lround(data->values[i + 1] * pix), 1, pix);
+      if (this->total == 0) {
```
fasterit, line 347:
```suggestion
      if (this->total < 1) {
```
Rationale:
* (nearly) never compare float values to zero
* we had the issue with very small denominators scaling the iometers very high before

fasterit (commented):
Otherwise looks fine to me

Meter.c
```diff
@@ -345,8 +345,10 @@ static void GraphMeterMode_draw(Meter* this, int x, int y, int w) {
    }
    for (; i < nValues - 1; i += 2, k++) {
       int pix = GraphMeterMode_pixPerRow * GRAPH_HEIGHT;
-      int v1 = CLAMP((int) lround(data->values[i] * pix), 1, pix);
-      int v2 = CLAMP((int) lround(data->values[i + 1] * pix), 1, pix);
+      if (this->total < FLT_EPSILON)
```
fasterit, line 348: Nope.
https://bitbashing.io/comparing-floats.html
"[F]loat values aren’t equidistant! For values smaller than 1, FLT_EPSILON quickly becomes too large to be useful."
Also creates discontinuity at 1 as you clamp to 1 in the next line (meter would jump from huge value to 1).

fasterit (commented):
`== 0` is bad, `< FLT_EPSILON` is worse, see link in review comment

BenBE:
Minor typo in the second commit message.

Meter.c
Nudin: So what is the correct way? FLT_MIN does not work either. Should I use an arbitrary small constant like 0.0001?

Meter.c
Nudin: But why would we limit the code to total being bigger than 1? I totally see use cases where we might want to display something that moves on the scale of 0 to 0.1, when clamping it to 1 the graph will become useless in such cases?

fasterit: Well, <ins>you</ins> made the zero case jump to one:
```C
if (this->total < FLT_EPSILON)
   this->total = 1;
```

I'd be fine with 1 as currently there is no meter that has a use for a total < 1.
But you can clamp to "an arbitrary small constant like 0.0001" as you suggest below as well. Doesn't matter.
Just not use a C constant wrongly and please make the function continuous, so if `total < ε` then also clamp total to ε.

Meter.c
BenBE: What about going with the suggestion from @fasterit from above and using `1`. To understand why this is quite reasonable we need to have a look at the usual values that these meters are displaying. As those values are usually some integer source and rarely ever scaled down we can assume that the smallest granularity we are likely to see is `1 unit` where unit is typically `Bytes/s` or `KiB/s`. With this in mind there'd be rarely any reason to go for anything below `1/10000` AKA `<1E-4`. And even if your smallest value was something like `1 B/s` you'd probably still don't want to make this the maximum for your scale. Instead for all intents and purposes is's more reasonable to start your scale at `1KiB/s` if the common value you'll be seeing is much larger than this; and thus err on the topic of discarding extremely small/unlikely values.

Nudin: Ok, I set it to 1. If we'll add meters for things that are smaller than 1, it can still be changed again.

NetworkIOMeter.c
```diff
@@ -80,6 +80,12 @@ static void NetworkIOMeter_updateValues(ATTR_UNUSED Meter* this, char* buffer, s
       cached_txp_total = packetsTransmitted;
    }
 
+   this->values[0] = cached_rxb_diff;
+   this->values[1] = cached_txb_diff;
+   if (cached_rxb_diff + cached_txb_diff > this->total) {
+      this->total = cached_rxb_diff;
```
fasterit, line 86: Why is the total not `cached_rxb_diff + cached_txb_diff`?

Nudin: fixed it.

Issue #542, opened 2021-02-25
No threads on OSX


Karsten1987:
This is essentially a cross-post from this original issue here: https://github.com/hishamhm/htop/issues/552

The described behavior is still present (at least on my setup) where toggling `H` or `K` does not have any effect, nor do I see any threads visualized. 

```
 ➭ htop --version
htop 3.0.4
```

The ticket's been opened quite a while ago and I haven't found any issue/post around that topic on this new repo. Please feel free to point me to the right location in case this ticket is a duplicate.

cgzones:
I rebased https://github.com/hishamhm/htop/pull/848 in #546. Can someone give it a try?

polluks:
Works fine:
````
➭ htop -V
htop 3.0.5
➭ uname -a
... 20.3.0 Darwin Kernel Version 20.3.0: Thu Jan 21 00:06:51 PST 2021; root:xnu-7195.81.3~1/RELEASE_ARM64_T8101 arm64
````

BenBE:
Unless someone with a Mac contributes a patch it's unlikely, as we currently do not have a test system for this platform.

emadpres:
I don't have the knowledge to patch the issue, but you can count on me as a tester

hashok:
> I rebased [hishamhm/htop#848](https://github.com/hishamhm/htop/pull/848) in #546. Can someone give it a try?

FYI: https://github.com/htop-dev/htop/pull/546#issuecomment-1698907153
It seems to be working fine for me under sudo.

emadpres:
I can also confirm the same, as I reported [here](https://github.com/htop-dev/htop/pull/546#issuecomment-1699178996).

![Screenshot 2023-08-30 at 3 24 29 PM](https://github.com/htop-dev/htop/assets/2047884/43035311-1221-4e2f-aec6-3acac705eb69)

barracuda156:
This has introduced a breakage, see: https://github.com/htop-dev/htop/issues/1686

PR #546, opened 2021-02-28
Darwin: scan thread information

cgzones:
Origin: https://github.com/hishamhm/htop/pull/848

Closes: #542

/cc @robaho @Karsten1987 @fasterit

### DIFF
--- a/darwin/DarwinProcess.c
+++ b/darwin/DarwinProcess.c
@@ -401,12 +401,28 @@ void DarwinProcess_setFromLibprocPidinfo(DarwinProcess* proc, DarwinProcessTable
    }
 }
 
+static ProcessState stateToChar(int run_state) {
+   switch (run_state) {
+   case TH_STATE_RUNNING:
+      return RUNNING;
+   case TH_STATE_STOPPED:
+      return STOPPED;
+   case TH_STATE_WAITING:
+      return WAITING;
+   case TH_STATE_UNINTERRUPTIBLE:
+      return UNINTERRUPTIBLE_WAIT;
+   case TH_STATE_HALTED:
+      return BLOCKED;
+   }
+   return UNKNOWN;
+}
+
 /*
  * Scan threads for process state information.
  * Based on: http://stackoverflow.com/questions/6788274/ios-mac-cpu-usage-for-thread
  * and       https://github.com/max-horvath/htop-osx/blob/e86692e869e30b0bc7264b3675d2a4014866ef46/ProcessList.c
  */
-void DarwinProcess_scanThreads(DarwinProcess* dp) {
+void DarwinProcess_scanThreads(DarwinProcess* dp, DarwinProcessTable* dpt) {
    Process* proc = (Process*) dp;
    kern_return_t ret;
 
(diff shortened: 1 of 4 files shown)


natoscott:
@cgzones looks good to me.  Might be worth setting 'isThread = false;' in DarwinProcess_new instead of relying on the calloc there, for code clarity.  If that's not preferred and calloc is OK - which is fine by me - there's some other initializations-to-zero in that function that could be removed.

Also I wonder about the taskAccess field (not from your new code, but related to threads) - it seems like this field might cause incorrect behaviour if htop samples a process before it creates any threads - if it only creates threads later in its life, those threads would never be counted because this flag was set first time through.  If thats the case, I think it could be safely removed - it appears to be an optimization only, AFAICS.

darwin/DarwinProcess.c
```diff
@@ -368,31 +383,61 @@ void DarwinProcess_scanThreads(DarwinProcess* dp) {
    integer_t run_state = 999;
    for (unsigned int i = 0; i < thread_count; i++) {
       thread_info_data_t thinfo;
-      mach_msg_type_number_t thread_info_count = THREAD_BASIC_INFO_COUNT;
-      ret = thread_info(thread_list[i], THREAD_BASIC_INFO, (thread_info_t)thinfo, &thread_info_count);
+      mach_msg_type_number_t thread_info_count = THREAD_EXTENDED_INFO_COUNT;
+      ret = thread_info(thread_list[i], THREAD_EXTENDED_INFO_COUNT, (thread_info_t)thinfo, &thread_info_count);
```
Karsten1987, line 387: are you sure that the `thinfo` struct is passed by copy and not by pointer-to? I've seen a couple of SO posts passing it in via `(thread_info_t) &thinfo`: https://stackoverflow.com/a/33414855/4583130

the same actually applies to the similar logic above for obtaining the `TASK_BASIC_INFO`.

Karsten1987 (commented):
I honestly don't really see a difference with this change.

I've added a quick screenshot of my local output on my macbook, where the executable `a.out` contains two concurrently running pthreads. However, these are not visualized in any way.
Screen Shot 2021-03-01 at 1 37 21 PM

For comparison, the same application built for Linux clearly shows the two running threads (highlighted in green):
Screen Shot 2021-03-01 at 1 36 33 PM

BenBE:
Just a small tip: Press <kbd>Esc</kbd> once to hide the process list cursor. Also Tree Mode might be of help to group the threads optically below their parent.

darwin/DarwinProcess.h
```diff
@@ -20,6 +20,7 @@ typedef struct DarwinProcess_ {
    uint64_t stime;
    bool taskAccess;
    bool translated;
+   bool isThread;
```
BenBE, line 23: Should be handled via global `Process::isUSerlandThread` field instead of introducing this new one.

darwin/DarwinProcess.c
```diff
@@ -96,8 +96,7 @@ static int DarwinProcess_compareByKey(const Process* v1, const Process* v2, Proc
 }
 
 bool Process_isThread(const Process* this) {
-   (void) this;
-   return false;
+   return ((const DarwinProcess*)this)->isThread;
 }
```
BenBE, line 100: Obsoleted by changes in #388.

BenBE (changes requested):

darwin/DarwinProcess.c
```diff
@@ -338,12 +338,28 @@ void DarwinProcess_setFromLibprocPidinfo(DarwinProcess* proc, DarwinProcessList*
...
+   return '?';
+}
+
 /*
  * Scan threads for process state information.
  * Based on: http://stackoverflow.com/questions/6788274/ios-mac-cpu-usage-for-thread
  * and       https://github.com/max-horvath/htop-osx/blob/e86692e869e30b0bc7264b3675d2a4014866ef46/ProcessList.c
  */
-void DarwinProcess_scanThreads(DarwinProcess* dp) {
+void DarwinProcess_scanThreads(DarwinProcess* dp, DarwinProcessList *dpl) {
```
BenBE, line 362: Consistent pointer alignment …

darwin/DarwinProcess.h
```diff
@@ -38,6 +38,6 @@ void DarwinProcess_setFromLibprocPidinfo(DarwinProcess* proc, DarwinProcessList*
  * Based on: http://stackoverflow.com/questions/6788274/ios-mac-cpu-usage-for-thread
  * and       https://github.com/max-horvath/htop-osx/blob/e86692e869e30b0bc7264b3675d2a4014866ef46/ProcessList.c
  */
-void DarwinProcess_scanThreads(DarwinProcess* dp);
+void DarwinProcess_scanThreads(DarwinProcess* dp, DarwinProcessList *dpl);
```
BenBE, line 41: Pointer alignment …

BenBE:
Please check if this also fixes #622.

Karrq:
I checked out this lcoally and built it but I still don't see the threads, even running the tool with sudo.

On MBP M2 Ventura 13.3

BenBE:
Have you checked that your configuration has display of userland threads activated.

Karrq:
Yep I had unchecked "Hide userland process threads" and also tried toggling it on and off with `H` just to make sure and I don't see any change in my output, and I also tried tree/list mode...
I was trying to look at the threads of a specific process that I'm interested in as Activity Monitor only reports the number of threads (8), so I'm sure there are threads for that process.

(btw v3.2.2 also doesn't show threads per process, but I can see the total number of user threads under tasks)

BenBE:
Okay, strange. I can confirm this is somewhat strange behaviour. @cgzones Can you check your patch again?

cgzones:
Recent version of MacOS limit the access to foreign threads via `thread_info()`. I am not aware of an alternative API.
It might work though with a signed binary running as root (not tested).

hashok:
I've compiled `htop` from this branch on MacBook Pro Max M2 running macOS Ventura 13.5.1 (all updates installed as of 30-Aug-2023) and I *do see user land threads* 🎉 , but only when running `sudo htop`. I have _not_ signed the `htop` binary indeed. So it's definitely an improvement over the current `htop` version 3.2.2 from Homebrew.

emadpres:
I also just compiled [`cgzones:darwin_threads`](https://github.com/cgzones/htop/tree/darwin_threads) on M2 Max (McaOS 13.5.1), and I can confirm the threads are shown (only with `sudo`).

Not sure how relevant, the followings are broken:
1. refreshing
2. Tree structure (see `|`s)
3. scrolling


here is a side-by-side comparison of `htop` v3.2.2 with the compiled branch:
Screenshot 2023-08-30 at 3 24 29 PM

emadpres:
> Works on macOS 14.0 @emadpres Your PID is really high, too high for the tree.

What do you mean it's too high? That's what I get.

Screenshot 2023-08-31 at 1 34 10 PM

BenBE:
@cgzones Can you rebase? What'S missing to push this along?

polluks:
@emadpres I was talking about the width: Your PID have 7 digits but htop is fixed for 5 digits.
This seems to be a bug. However your system should not use 3 million processes in 4 days...

cgzones:
> What's missing to push this along?

Testing: check all threads are displayed, no memory is leaked, tree glitches.

(Only compile tested the latest rebase.)

BenBE (approved):
LGTM. No obvious memleaks.

Minor nitpick: Neither `identifer_info_count` nor `extended_info_count` are used or checked after their assignment through the call to `thread_info`, but access to at least the first item of the returned arrays is assumed.

cgzones:
> LGTM. No obvious memleaks.
> 
> Minor nitpick: Neither `identifer_info_count` nor `extended_info_count` are used or checked after their assignment through the call to `thread_info`, but access to at least the first item of the returned arrays is assumed.

The output parameters `identifer_info` and `extended_info` are not arrays, see https://web.mit.edu/darwin/src/modules/xnu/osfmk/man/thread_info.html for API details.

BenBE:
> > LGTM. No obvious memleaks.
> > Minor nitpick: Neither `identifer_info_count` nor `extended_info_count` are used or checked after their assignment through the call to `thread_info`, but access to at least the first item of the returned arrays is assumed.
> 
> The output parameters `identifer_info` and `extended_info` are not arrays, see https://web.mit.edu/darwin/src/modules/xnu/osfmk/man/thread_info.html for API details.

Ah, I see. So basically buffer sizes. In that case, at least checking for "a whole item was returned" should suffice. But as mentioned, that is somewhat nitpicky.

BenBE:
@cgzones Can you have a look at the PID auto-sizing issue? Likely missing the width update call for that column. Bugfix for that should get its own PR.

@fasterit Can you take a look at this PR, so we can push it along?

dundee:
Any update on this? Would be great to see threads on Mac!

fasterit:
@dundee: Did you test the [cgzones:darwin_threads](https://github.com/cgzones/htop/tree/darwin_threads) branch?

dundee:
@fasterit Yes, it works fine for me (using sudo).

findepi:
@cgzones @BenBE thanks for this PR!
Small question, which version this is/will be available in?
latest version from brew is 3.3.0 and doesn't show threads (even with sido). Current main branch built from source shows threads on my Mac (when run with sudo).

BenBE:
If your from-source build reports `3.3.0` you should re-run `autogen.sh`.

That aside: from the commit mentioned in the merge commit you can determine exactly when this landed, but I'd guess it's highly likely this was after 3.3.0 was released; thus this is probably not included in any release yet. Have a look at the milestone of this PR to check for which release it is planned.

findepi:
@BenBE apologies. i checked version wrong. my from-source build is `3.4.0-dev-3.3.0-144-gdb73229`.

> Have a look at the milestone of this PR to check for which release it is planned.

oh, i didn't notice that, but that's so nice.
looking forward to 3.4.0 release then. 
thanks!

Issue #538, opened 2021-02-18
Provide a way to select which sensor to use for CPU temperature


hrw:
At the moment htop has hardcoded list of x86-64 cpu sensors. So it does not read them on my computers ;(

```
[hrw@jagular htop]$ sensors
acpitz-acpi-0
Adapter: ACPI interface
temp1:        +61.0°C  (crit = +95.0°C)
temp2:        +60.0°C  (crit = +95.0°C)
temp3:        +61.0°C  (crit = +95.0°C)
temp4:        +60.0°C  (crit = +95.0°C)
temp5:        +60.0°C  (crit = +95.0°C)
temp6:        +60.0°C  (crit = +95.0°C)
temp7:        +60.0°C  (crit = +95.0°C)
temp8:        +59.0°C  (crit = +95.0°C)
temp9:        +60.0°C  (crit = +95.0°C)
```

For now I added 'acpitz' to hardcoded list but that's not a good solution.

BenBE:
Related to #529 …

fasterit:
Just FTR acpitz = ACPI Thermal Zone

hardhub:
Why is it closed?

It is not possible to choose lm_sensors value.

I need to choose motherboard sensor, because only that one is accurate.

BenBE:
Because we're not going to implement sensor readout from sysfs directly and there's #200 will likely bring the flexibility needed to choose the sensor to use for the meters … apart from that: [read that fine manual](https://linux.die.net/man/5/sensors.conf) for libsensors if you need to re-assign things.

Issue #529, opened 2021-02-15
Temperature not showing on Via Eden and AMD Phenom CPUs


aleemont:
Temperature is showing as `N/A` when enabled on a Via Eden x2 u4200 even though `libsensors` is actually installed.
O.S.: Arch Linux 
Device: HP t510 Thin Client

I tried on my main pc (which is running Arch, with an intel core i5 cpu) and temperature is working fine. Thank you.

cgzones:
What does `sensors -u` show on your system?

aleemont:
`
$ sensors -u

via_cputemp-isa-0000
Adapter: ISA adapter
Core 0:
    temp1_input: 34.000

via_cputemp-isa-0001
Adapter: ISA adapter
Core 1:
    temp1_input: 37.000
`

riveravaldez:
Hi, I'm having similar issue in similar conditions but in a different machine with updated Debian (testing).

```
$ sensors -u
drivetemp-scsi-1-0
Adapter: SCSI adapter
temp1:
  temp1_input: 38.000
  temp1_max: 60.000
  temp1_min: 0.000
  temp1_crit: 85.000
  temp1_lcrit: -41.000
  temp1_lowest: 23.000
  temp1_highest: 41.000

it8718-isa-0290
Adapter: ISA adapter
in0:
  in0_input: 1.056
  in0_min: 0.000
  in0_max: 4.080
  in0_alarm: 0.000
  in0_beep: 0.000
in1:
  in1_input: 1.904
  in1_min: 0.000
  in1_max: 4.080
  in1_alarm: 0.000
  in1_beep: 0.000
in2:
  in2_input: 3.136
  in2_min: 0.000
  in2_max: 4.080
  in2_alarm: 0.000
  in2_beep: 0.000
in3:
  in3_input: 4.080
  in3_min: 0.000
  in3_max: 4.080
  in3_alarm: 1.000
  in3_beep: 0.000
in4:
  in4_input: 2.928
  in4_min: 0.000
  in4_max: 4.080
  in4_alarm: 0.000
  in4_beep: 0.000
in5:
  in5_input: 2.144
  in5_min: 0.000
  in5_max: 4.080
  in5_alarm: 0.000
  in5_beep: 0.000
in6:
  in6_input: 2.144
  in6_min: 0.000
  in6_max: 4.080
  in6_alarm: 0.000
  in6_beep: 0.000
in7:
  in7_input: 3.136
  in7_min: 0.000
[...]

PR #547, opened 2021-02-28
Linux: Rework libsensors parsing

cgzones:
Do not read driver depended labels, just count the number of
temperatures given:

- on #CPU:
  + platform temp = max cpu temp
  + CPU temps = first to last
- on #CPU + 1:
  + platform temp = first temp
  + CPU temps = second to last
- on #CPU / 2:
  + platform temp = max cpu temp
  + CPU temps = first to last concat first to last
      (with SMT core x + cpu count is the logical core of the physical
      core x)
- on #CPU / 2 + 1:
  + platform temp = first temp
  + CPU temps = second to last concat second to last
      (with SMT core x + cpu count is the logical core of the physical
      core x)

Closes: #529
Closes: #538

### DIFF
--- a/linux/LibSensors.c
+++ b/linux/LibSensors.c
@@ -15,22 +15,18 @@
 #define sym_sensors_init sensors_init
 #define sym_sensors_cleanup sensors_cleanup
 #define sym_sensors_get_detected_chips sensors_get_detected_chips
-#define sym_sensors_snprintf_chip_name sensors_snprintf_chip_name
 #define sym_sensors_get_features sensors_get_features
 #define sym_sensors_get_subfeature sensors_get_subfeature
 #define sym_sensors_get_value sensors_get_value
-#define sym_sensors_get_label sensors_get_label
 
 #else
 
 static int (*sym_sensors_init)(FILE*);
 static void (*sym_sensors_cleanup)(void);
 static const sensors_chip_name* (*sym_sensors_get_detected_chips)(const sensors_chip_name*, int*);
-static int (*sym_sensors_snprintf_chip_name)(char*, size_t, const sensors_chip_name*);
 static const sensors_feature* (*sym_sensors_get_features)(const sensors_chip_name*, int*);
 static const sensors_subfeature* (*sym_sensors_get_subfeature)(const sensors_chip_name*, const sensors_feature*, sensors_subfeature_type);
 static int (*sym_sensors_get_value)(const sensors_chip_name*, int, double*);
-static char* (*sym_sensors_get_label)(const sensors_chip_name*, const sensors_feature*);
 
 static void* dlopenHandle = NULL;
 
@@ -66,11 +62,9 @@ int LibSensors_init(FILE* input) {
       resolve(sensors_init);
       resolve(sensors_cleanup);
       resolve(sensors_get_detected_chips);
-      resolve(sensors_snprintf_chip_name);
       resolve(sensors_get_features);
       resolve(sensors_get_subfeature);
       resolve(sensors_get_value);
-      resolve(sensors_get_label);
 
       #undef resolve
    }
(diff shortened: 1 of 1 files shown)


linux/LibSensors.c
```diff
@@ -105,99 +99,142 @@ void LibSensors_cleanup(void) {
...
+      const char* prefix;
+      int priority;
+   } tempDrivers[] =  {
+      { "coretemp",    1, },
+      { "via_cputemp", 1, },
+      { "cpu_thermal", 1, },
+      { "k10temp",     1, },
+      { "zenpower",    1, },
+      /* Low priority drivers */
+      { "acpitz",      0, },
```
BenBE, line 113: No need for the last comma inside the structure initializer

linux/LibSensors.c
```diff
@@ -105,99 +99,142 @@ void LibSensors_cleanup(void) {
...
+
 void LibSensors_getCPUTemperatures(CPUData* cpus, unsigned int cpuCount) {
-   for (unsigned int i = 0; i <= cpuCount; i++)
-      cpus[i].temperature = NAN;
+   double data[cpuCount + 1];
+#ifdef __clang_analyzer__
+   memset(data, -1, ARRAYSIZE(data)); /* please clang analyzer */
+#endif
+   for (size_t i = 0; i < ARRAYSIZE(data); i++)
+      data[i] = NAN;
```
BenBE, line 129: Why not splitting this `NAN` assignment into a separate loop? Should make the Clang Analyzer just as happy …

linux/LibSensors.c
```diff
@@ -105,99 +99,142 @@ void LibSensors_cleanup(void) {
...
-          !String_startsWith(buffer, "cpu_thermal") &&
-          !String_startsWith(buffer, "k10temp") &&
-          !String_startsWith(buffer, "zenpower"))
+   for (const sensors_chip_name* chip = sym_sensors_get_detected_chips(NULL, &n); chip; chip = sym_sensors_get_detected_chips(NULL, &n)) {
+      const int priority = tempDriverPriority(chip);
+      if (priority < 0)
          continue;
 
+      if (priority < maxPriority)
+         continue;
```
BenBE, line 146: This loop construction might lead to may lead to values from a higher-prioritized chip to be ignored in favor of values read from a lower-prioritized one (priority inversion issue), which may have irritating consequences when `acpitz` reports 8 values, but `coretemp` only reports 5.

linux/LibSensors.c
```diff
@@ -105,99 +99,142 @@ void LibSensors_cleanup(void) {
...
-            /* Raspberry Pi has only temp1 */
-            tempId = 0;
-         } else if (String_startsWith(label, "Tdie")) {
-            tempId = 0;
-         } else if (String_startsWith(label, "Core ")) {
-            tempId = 1 + atoi(label + strlen("Core "));
-         } else {
-            tempId = UINT_MAX;
-         }
+         int tempID = atoi(feature->name + strlen("temp"));
```
BenBE, line 158: Why not use `strtoul` instead and save on the cast below?

linux/LibSensors.c
```diff
@@ -105,99 +99,142 @@ void LibSensors_cleanup(void) {
...
+         if (isnan(data[i]))
             continue;
 
-         maxTemp = MAXIMUM(maxTemp, coreTemp);
+         maxTemp = MAXIMUM(maxTemp, data[i]);
       }
 
-      cpus[0].temperature = maxTemp;
+      data[0] = maxTemp;
+   }
```
BenBE, line 215: Document intended fall-thru

BenBE (changes requested):

linux/LibSensors.c
cgzones: What do you mean by separate loop?
The analyzer is just dump and thinks `0 < ARRAYSIZE(data)` might be `false`.

linux/LibSensors.c
cgzones: In case `coretemp` reports 5 values, any data from `acpitz` should be ignored, cause it might not be a CPU sensor.

linux/LibSensors.c
BenBE: Due to the VLA `data` this actually might be zero, when overflowing with `cpuCount = MAX_INT` due to the implicit conversion from `unsigned int` to `signed int` for the VLA length argument.

linux/LibSensors.c
```diff
@@ -105,99 +99,146 @@ void LibSensors_cleanup(void) {
...
-          !String_startsWith(buffer, "zenpower"))
+   for (const sensors_chip_name* chip = sym_sensors_get_detected_chips(NULL, &n); chip; chip = sym_sensors_get_detected_chips(NULL, &n)) {
+      const int priority = tempDriverPriority(chip);
+      if (priority < 0)
+         continue;
+
+      if (priority > topPriority)
          continue;
 
+      topPriority = priority;
```
BenBE, line 148: may still mix values from different priorities depending on reported order of sensors.

cgzones: Should be fixed now

linux/LibSensors.c
cgzones: added an `assert()` and used `cpuCount + 1` as upper bound to please clang analyzer

BenBE (approved):

PR #548, opened 2021-02-28
fbinfer

cgzones:
Some findings from the Facebook Infer analyzer.

### DIFF
--- a/DateTimeMeter.c
+++ b/DateTimeMeter.c
@@ -26,8 +26,7 @@ static void DateTimeMeter_updateValues(Meter* this, char* buffer, size_t size) {
    int year = lt->tm_year + 1900;
    if (((year % 4 == 0) && (year % 100 != 0)) || (year % 400 == 0)) {
       this->total = 366;
-   }
-   else {
+   } else {
       this->total = 365;
    }
    this->values[0] = lt->tm_yday;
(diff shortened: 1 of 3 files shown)


DateTimeMeter.c
```diff
@@ -22,16 +22,16 @@ static const int DateTimeMeter_attributes[] = {
...
    time_t t = time(NULL);
    struct tm result;
-   const struct tm* lt = localtime_r(&t, &result);
-   int year = lt->tm_year + 1900;
+   localtime_r(&t, &result);
+   int year = result.tm_year + 1900;
    if (((year % 4 == 0) && (year % 100 != 0)) || (year % 400 == 0)) {
       this->total = 366;
    }
    else {
```
BenBE, line 30: No need for line break; maybe even no bracesß

BenBE (approved):

BenBE (approved):

natoscott:
@cgzones this seems like code churn for little/no benefit :( ... am I missing something?

Other than the one-line dead assignment change (+1 for that), nothing is improving here AFAICS ... mainly we replace some potential null assignments with use of uninitialized data instead (if fbinfer says this code is OK now, it's giving a false sense of security - let's not use it IMO).  As for adding asserts on null pointers right before dereferencing them, that's code noise and useless in practice.

If we're concerned about the localtime_r failure cases now, we should instead check for failure (null return) and do something appropriate, not pass bad data from the stack.  Since these functions only use a time_t value from time(2) on the previous line, I'm pretty sure they cannot actually fail in the ways we're using 'em.

cgzones:
Dropped adding the `assert()`'s and the `localtime_r()` changes.

PR #549, opened 2021-03-01
Fix integer sizing issues in the NetworkIO Meter

natoscott:
On Linux kernels the size of the values exported for network
device bytes and packets has used a 64 bit integer for quite
some time (2.6+ IIRC).  Make the procfs value extraction use
correct types and change internal types used to rate convert
these counters (within the NetworkIO Meter) 64 bit integers,
where appropriate.

### DIFF
--- a/freebsd/Platform.c
+++ b/freebsd/Platform.c
@@ -321,8 +321,7 @@ bool Platform_getNetworkIO(NetworkIOData* data) {
    size_t countLen = sizeof(count);
    const int countMib[] = { CTL_NET, PF_LINK, NETLINK_GENERIC, IFMIB_SYSTEM, IFMIB_IFCOUNT };
 
-   int r;
-   r = sysctl(countMib, ARRAYSIZE(countMib), &count, &countLen, NULL, 0);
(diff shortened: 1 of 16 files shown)


freebsd/Platform.c
```diff
@@ -315,24 +315,18 @@ bool Platform_getDiskIO(DiskIOData* data) {
...
-   int r;
-
+bool Platform_getNetworkIO(NetworkIOData* data) {
    // get number of interfaces
    int count;
    size_t countLen = sizeof(count);
    const int countMib[] = { CTL_NET, PF_LINK, NETLINK_GENERIC, IFMIB_SYSTEM, IFMIB_IFCOUNT };
 
+   int r;
    r = sysctl(countMib, ARRAYSIZE(countMib), &count, &countLen, NULL, 0);
```
BenBE, line 325: Why not in one line?

BenBE (commented):
Apart from this the same comments as with #550 …

freebsd/Platform.c
natoscott: Yep, will clean that up - thanks.

PR #550, opened 2021-03-01
Fix integer sizing issues in the DiskIO Meter

natoscott:
On Linux kernels the size of the values exported for block
device bytes has used a 64 bit integer for quite some time
(2.6+ IIRC).  Make the procfs value extraction use correct
types and change internal types used to rate convert these
counters (within the DiskIO Meter) 64 bit integers, where
appropriate.

### DIFF
--- a/DiskIOMeter.c
+++ b/DiskIOMeter.c
@@ -26,12 +26,12 @@ static const int DiskIOMeter_attributes[] = {
 };
 
 static bool hasData = false;
-static unsigned long int cached_read_diff = 0;
-static unsigned long int cached_write_diff = 0;
-static double cached_utilisation_diff = 0.0;
+static unsigned long int cached_read_diff;
+static unsigned long int cached_write_diff;
+static double cached_utilisation_diff;
 
 static void DiskIOMeter_updateValues(Meter* this, char* buffer, size_t len) {
-   static unsigned long long int cached_last_update = 0;
+   static unsigned long long int cached_last_update;
 
    struct timeval tv;
    gettimeofday(&tv, NULL);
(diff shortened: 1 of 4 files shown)


DiskIOMeter.c
```diff
@@ -26,12 +26,12 @@ static const int DiskIOMeter_attributes[] = {
 };
 
 static bool hasData = false;
-static unsigned long int cached_read_diff = 0;
-static unsigned long int cached_write_diff = 0;
-static double cached_utilisation_diff = 0.0;
+static unsigned long int cached_read_diff;
+static unsigned long int cached_write_diff;
```
BenBE, line 30: Why not go for `uint32_t`/`uint64_t` directly?

DiskIOMeter.c
```diff
@@ -56,21 +57,26 @@ static void DiskIOMeter_updateValues(Meter* this, char* buffer, size_t len) {
       }
 
       if (data.totalBytesRead > cached_read_total) {
-         cached_read_diff = (data.totalBytesRead - cached_read_total) / 1024; /* Meter_humanUnit() expects unit in kilo */
+         diff = data.totalBytesRead - cached_read_total;
+         diff /= 1024; /* Meter_humanUnit() expects unit in kilo */
+         cached_read_diff = (unsigned int)diff;
       } else {
-         cached_read_diff = 0;
+         cached_read_diff = 0UL;
```
BenBE, line 64: Why this explicit markup as `unsigned long`?

DiskIOMeter.c
```diff
@@ -56,21 +57,26 @@ static void DiskIOMeter_updateValues(Meter* this, char* buffer, size_t len) {
...
       }
       cached_read_total = data.totalBytesRead;
 
       if (data.totalBytesWritten > cached_write_total) {
-         cached_write_diff = (data.totalBytesWritten - cached_write_total) / 1024; /* Meter_humanUnit() expects unit in kilo */
+         diff = data.totalBytesWritten - cached_write_total;
+         diff /= 1024; /* Meter_humanUnit() expects unit in kilo */
+         cached_write_diff = (unsigned int)diff;
```
BenBE, line 71: Type truncation: `unsigned int` casted, but `cached_write_diff` is `unsigned long int`…

PR #553, opened 2021-03-02
Separate display from sampling in SysArch and Hostname Meters

natoscott:
Several of our newer meters have merged coding concerns in terms
of extracting values and displaying those values.  This commit
rectifies that for the SysArch and Hostname meters, allowing use
of this code with alternative front/back ends.  The SysArch code
is also refined to detect whether the platform has an os-release
file at all and/or the sys/utsname.h header via configure.ac.

### DIFF
--- a/Generic.c
+++ b/Generic.c
@@ -1,0 +1,102 @@
+/*
+htop - Generic.c
+(C) 2021 htop dev team
+Released under the GNU GPLv2, see the COPYING file
+in the source distribution for its full text.
+*/
+#include "config.h"  // IWYU pragma: keep
+
+#include "Generic.h"
+
+#include <stdio.h>
+#ifdef HAVE_SYS_UTSNAME_H
+#include <sys/utsname.h>
+#endif
+#include <unistd.h>
+
+#include "XUtils.h"
+
+void Generic_Hostname(char* buffer, size_t size) {
+   gethostname(buffer, size - 1);
+}
+
+#ifdef HAVE_SYS_UTSNAME_H
+
+#ifndef OSRELEASEFILE
+#define OSRELEASEFILE "/etc/os-release"
+#endif
+
+static void parseOSRelease(char* buffer, size_t bufferLen) {
+   FILE* stream = fopen(OSRELEASEFILE, "r");
+   if (!stream) {
+      xSnprintf(buffer, bufferLen, "No OS Release");
+      return;
+   }
+
+   char name[64] = {'\0'};
+   char version[64] = {'\0'};
+   char lineBuffer[256];
+   while (fgets(lineBuffer, sizeof(lineBuffer), stream)) {
+      if (String_startsWith(lineBuffer, "PRETTY_NAME=\"")) {
+         const char* start = lineBuffer + strlen("PRETTY_NAME=\"");
+         const char* stop = strrchr(lineBuffer, '"');
+         if (!stop || stop <= start)
+            continue;
+         String_safeStrncpy(buffer, start, MINIMUM(bufferLen, (size_t)(stop - start + 1)));
+         fclose(stream);
+         return;
+      }
+      if (String_startsWith(lineBuffer, "NAME=\"")) {
+         const char* start = lineBuffer + strlen("NAME=\"");
+         const char* stop = strrchr(lineBuffer, '"');
+         if (!stop || stop <= start)
+            continue;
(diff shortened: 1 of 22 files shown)


Generic.c
```diff
@@ -0,0 +1,102 @@
...
+in the source distribution for its full text.
+*/
+#include "config.h"  // IWYU pragma: keep
+
+#include "Generic.h"
+
+#include <stdio.h>
+#ifdef HAVE_SYS_UTSNAME_H
+#include <sys/utsname.h>
+#endif
```
BenBE, line 14: Maybe put this include last, cf. styleguide. Also two blank lines after includes.

Generic.c
```diff
@@ -0,0 +1,102 @@
...
+         String_safeStrncpy(version, start, MINIMUM(sizeof(version), (size_t)(stop - start + 1)));
+         continue;
+      }
+   }
+   fclose(stream);
+
+   snprintf(buffer, bufferLen, "%s%s%s", name[0] ? name : "", name[0] && version[0] ? " " : "", version);
+}
+
+char* Generic_OSRelease(void) {
```
BenBE, line 71: Will have to check with WinNT, but might be good to allow the actual platform to override/provide the implementation of this.

configure.ac
```diff
@@ -608,6 +621,7 @@ AC_MSG_RESULT([
   ${PACKAGE_NAME} ${VERSION}
 
   platform:                  $my_htop_platform
+  os-release file:           $with_os_release
   (Linux) proc directory:    $with_proc
```
BenBE, line 625: The label for proc is used mainly on Linux, but checked on all platforms (except WinNT). Maybe drop platform marker?

Makefile.am
```diff
@@ -42,6 +42,7 @@ myhtopsources = \
 	DisplayOptionsPanel.c \
 	EnvScreen.c \
 	FunctionBar.c \
+	Generic.c \
```
BenBE, line 45: Should only be added in for platforms actually forwarding here.

Makefile.am
```diff
@@ -96,6 +97,7 @@ myhtopheaders = \
 	DisplayOptionsPanel.h \
 	EnvScreen.h \
 	FunctionBar.h \
+	Generic.h \
```
BenBE, line 100: dito.

BenBE (changes requested):

Generic.c
natoscott: @BenBE yeah, agreed.  This PR had an initial incantation where I created a generic/ directory, for optional sharing of implementations of useful routines (zfs/openzfs_sysctl.\* is in this category too) and platforms pick and choose which files to use.  That was slightly more invasive though, so I backed off and went with the simpler approach - probably I should have stuck with it.  Also, currently all conditionally compiled .c files are below sub-directories - and changing that looked a bit odd in Makefile.am.  To my mind, it is kinda nice to know any code you look at in the main src directory is always compiled in, and the conditional code is all in subdirs.  *shrug*

Given those notes, whats your preference?  I don't really mind either way, but I guess I slightly favour generic/\*.[c|h] for these shared implementations.  Then should openzfs_sysctl.* move to generic/ if we go that way, to indicate its a shared implementation used by some platforms only?

natoscott: > Will have to check with WinNT, but might be good to allow the actual platform to override/provide the implementation of this.

https://stackoverflow.com/questions/5773281/compilation-fails-due-to-missing-file-sys-utsname-h

BenBE: I think I prefer a `generic/` or `common/` subfolder for these kinds of stuff. Just like you also suggested for the openzfs stuff.

Re WinNT: Yeah, that's also what I found for my research: Building with MSys/Cygwin would provide them, but they aren't native headers for WinNT. So second platform after the unsupported one that implements these functions differently.

darwin/DarwinProcessList.c
```diff
@@ -20,9 +20,9 @@ in the source distribution for its full text.
 
 #include "CRT.h"
 #include "DarwinProcess.h"
+#include "generic/openzfs_sysctl.h"
```
BenBE, line 23: Include order ...

BenBE (commented):
Include order has subfolder includes after parent dirs.

BenBE (approved):

fasterit:
Merged as 5b50ae3 +3

PR #554, opened 2021-03-02
Darwin: add normalized CPU percentage column

cgzones:
Missed in 15eab2012d21


BenBE (changes requested):
Also missing from the unsupported port …

fasterit:
Extended to the Unsupported platform as requested by @BenBE (0bdceb8) and merged in 4f9cf14

PR #555, opened 2021-03-02
Misc

cgzones:


TraceScreen.c
```diff
@@ -93,8 +93,7 @@ bool TraceScreen_forkTracer(TraceScreen* this) {
 
       // Should never reach here, unless execlp fails ...
       const char* message = "Could not execute 'strace'. Please make sure it is available in your $PATH.";
-      ssize_t written = write(STDERR_FILENO, message, strlen(message));
-      (void) written;
+      (void)! write(STDERR_FILENO, message, strlen(message));
```
BenBE, line 96: Why negate the result from `write`?

BenBE (changes requested):

HostnameMeter.c
```diff
@@ -19,8 +19,7 @@ static const int HostnameMeter_attributes[] = {
    HOSTNAME
 };
 
-static void HostnameMeter_updateValues(Meter* this, char* buffer, size_t size) {
-   (void) this;
+static void HostnameMeter_updateValues(ATTR_UNUSED Meter* this, char* buffer, size_t size) {
```
natoscott, line 22: @cgzones this is fixed in #553 FWIW.

TraceScreen.c
cgzones: `write(2)` is annotated with `__attribute__((warn_unused_result)` so compilers warn on ignored return value.
Casting the result to `(void)` only pleases clang but not gcc, see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66425.
See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66425#c34 for the negation hack.

BenBE: Urgs, k.

BenBE (approved):

Issue #556, opened 2021-03-03
Add sum of tmpfs mounts as additional item in mem meter


birdie-github:
A recurring issue for lots of Linux users nowadays is that they run out of RAM without understanding where the RAM has gone.

It's due to modern Linux distros (at the very least Fedora and Arch) mounting `tmpfs` to `/tmp` `/run` and `/dev/shm`. Other distros are also slowly moving in this direction.

Please include all the `tmpfs` mount points usage into the `Used RAM` report/value.

Right now we have:

```
Mem:15.6G used:7.40G buffers:1.11G cache:3.59G 
```
It could be changed to:
```
Mem:15.6G used:8.40G buffers:1.11G cache:3.59G tmpfs: 1.0G
````

A much bigger issue is that the Linux kernel doesn't treat `tmpfs` as used RAM and Linux kernel developers seemingly don't care: https://bugzilla.kernel.org/show_bug.cgi?id=201675

fasterit:
Since 3d497a3760626d398fffc8f4594e8b9658f67d55 tmpfs usage is contained in the `used` memory item.

birdie-github:
It's great it's included but I'd love to see this information as a separate column/field to avoid confusion.

cgzones:
Something like #558?

BenBE:
@birdie-github Please add comments instead of editing your initial issue description as this is much easier to follow up upon.

PR #558, opened 2021-03-03
Linux: individually show shared memory

cgzones:
Shmem: Total memory used by shared memory (shmem) and tmpfs

Source: https://www.kernel.org/doc/Documentation/filesystems/proc.txt

Closes: #556


natoscott (approved):
@cgzones LGTM

BenBE:
Will have to test this with my setup if it gives reasonable results, but the general patch looks already good. More comments after I got around to testing it.

BenBE (approved):
Values seem reasonable even with large RAM disks partially in swap.

Issue #557, opened 2021-03-03
Cursor doesn’t follow item when pressing H


gonzaarcr:
I would be nice that the cursor still follows the same item when pressing H (show/hide user process) or K (show/hide kernel) if it’s posible. Really handy when using tree mode.

I don’t know if it’s a bug or a feature request.

Htop version: 3.0.5

PR #559, opened 2021-03-03
Follow followed process when switching thread visibilities

cgzones:
Do not stop following a process when switching the visibility of
userland or kernel threads.

Related: #557


natoscott (approved):
@cgzones LGTM

BenBE (approved):

BenBE:
Depending on complexity it might be worth selecting the parent process when a thread is selected and this thread will be hidden after the keypress …

fasterit:
> Depending on complexity it might be worth selecting the parent process when a thread is selected and this thread will be hidden after the keypress …

Good idea but quite complex, can be implemented as another PR.

PR #561, opened 2021-03-12
RichString: do not unnecessarily clean whole buffer

cgzones:
The local stack buffer does not need to be cleaned to zeros when
  - just initialized, cause the length is set to 0 and the first
    character is set to '\0', so all printing functions will safely stop
  - no further used, i.e. the variable goes out of scope

### DIFF
--- a/ProcessList.c
+++ b/ProcessList.c
@@ -94,7 +94,7 @@ static const char* alignedProcessFieldTitle(ProcessField field) {
 }
 
 void ProcessList_printHeader(const ProcessList* this, RichString* header) {
-   RichString_prune(header);
+   RichString_rewind(header, RichString_size(header));
 
    const Settings* settings = this->settings;
    const ProcessField* fields = settings->fields;
--- a/RichString.c
+++ b/RichString.c
@@ -185,11 +185,11 @@ int RichString_findChar(const RichString* this, char c, int start) {
 
 #endif /* HAVE_LIBNCURSESW */
 
-void RichString_prune(RichString* this) {
-   if (this->chlen > RICHSTRING_MAXLEN)
+void RichString_delete(RichString* this) {
+   if (this->chlen > RICHSTRING_MAXLEN) {
       free(this->chptr);
-   memset(this, 0, sizeof(RichString));
-   this->chptr = this->chstr;
+      this->chptr = this->chstr;
+   }
 }
 
 void RichString_setAttr(RichString* this, int attrs) {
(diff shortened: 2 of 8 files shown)


RichString.c
```diff
@@ -185,11 +185,9 @@ int RichString_findChar(const RichString* this, char c, int start) {
 
 #endif /* HAVE_LIBNCURSESW */
 
-void RichString_prune(RichString* this) {
+void RichString_delete(RichString* this) {
    if (this->chlen > RICHSTRING_MAXLEN)
       free(this->chptr);
-   memset(this, 0, sizeof(RichString));
-   this->chptr = this->chstr;
```
BenBE, line 192: Don#t drop this, because this leaves a dangling pointer to potentially unallocated memory in ´this-<chptr´

ProcessList.c
```diff
@@ -94,7 +94,7 @@ static const char* alignedProcessFieldTitle(ProcessField field) {
 }
 
 void ProcessList_printHeader(const ProcessList* this, RichString* header) {
-   RichString_prune(header);
+   RichString_rewind(header, RichString_size(header));
```
BenBE, line 97: Why not call `RichString_setLen(header, 0)` instead?

BenBE (changes requested):
This assumes implicitly that `*_display` API is called with empty `RichString` on entry. This is a new assumption about the API that didn't exist previously (although it should have been a given already).

cgzones:
The polymorphic `Object` `display()` function is called at https://github.com/htop-dev/htop/blob/d9f2eacbc5b3fccf63b24944ce9a30d762baea3c/Meter.c#L101 (where the RichStirng is initialzed at https://github.com/htop-dev/htop/blob/d9f2eacbc5b3fccf63b24944ce9a30d762baea3c/Meter.c#L167 and https://github.com/htop-dev/htop/blob/d9f2eacbc5b3fccf63b24944ce9a30d762baea3c/Meter.c#L389) 

and at https://github.com/htop-dev/htop/blob/d9f2eacbc5b3fccf63b24944ce9a30d762baea3c/Panel.c#L273-L274 https://github.com/htop-dev/htop/blob/d9f2eacbc5b3fccf63b24944ce9a30d762baea3c/Panel.c#L300-L301 https://github.com/htop-dev/htop/blob/d9f2eacbc5b3fccf63b24944ce9a30d762baea3c/Panel.c#L304-L305

So the `RichString` is empty on `display()` calls.

ProcessList.c
cgzones: cause `RichString_setLen()` is currently a private function, which I'd like to keep

BenBE (approved):

PR #564, opened 2021-03-17
Move libcap use to (Linux) platform-specific code

natoscott:
The libcap code is Linux-specific so move it all below
the linux/ platform subdirectory.  As this feature has
custom command-line long options I provide a mechanism
whereby each platform can add custom long options that
augment the main htop options.  We'll make use this of
this with the pcp/ platform in due course to implement
the --host and --archive options there.

Related to https://github.com/htop-dev/htop/pull/536

### DIFF
--- a/htop.c
+++ b/htop.c
@@ -36,20 +36,12 @@ in the source distribution for its full text.
 #include "XUtils.h"
 
 
-#ifndef PLATFORM_LONG_OPTIONS
-#define PLATFORM_LONG_OPTIONS
-#endif
-
-#ifndef PLATFORM_LONG_OPTIONS_USAGE
-#define PLATFORM_LONG_OPTIONS_USAGE
-#endif
-
 static void printVersionFlag(const char* name) {
-   fprintf(stdout, "%s " VERSION "\n", name);
+   printf("%s " VERSION "\n", name);
 }
 
 static void printHelpFlag(const char* name) {
-   fprintf(stdout, "%s " VERSION "\n"
+   printf("%s " VERSION "\n"
          COPYRIGHT "\n"
          "Released under the GNU GPLv2.\n\n"
          "-C --no-color                   Use a monochrome color scheme\n"
--- a/htop.c
+++ b/htop.c
@@ -48,7 +48,6 @@ static void printHelpFlag(const char* name) {
          "-d --delay=DELAY                Set the delay between updates, in tenths of seconds\n"
          "-F --filter=FILTER              Show only the commands matching the given filter\n"
          "-h --help                       Print this help screen\n"
-         PLATFORM_LONG_OPTIONS_USAGE
          "-H --highlight-changes[=DELAY]  Highlight new and old processes\n"
          "-M --no-mouse                   Disable the mouse\n"
          "-p --pid=PID[,PID,PID...]       Show only the given PIDs\n"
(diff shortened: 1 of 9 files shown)


htop.c
```diff
@@ -35,37 +35,28 @@ in the source distribution for its full text.
 #include "UsersTable.h"
 #include "XUtils.h"
 
-#ifdef HAVE_LIBCAP
-#include <sys/capability.h>
-#endif
 
+#ifndef PLATFORM_LONG_OPTIONS
```
cgzones, line 39: My preference would be to not default-define them here, but define them in any `${platform}/platform.h` file.
That way we would fail hard on a missing definition or a messed up include order.

htop.c
```diff
@@ -35,37 +35,28 @@ in the source distribution for its full text.
...
+#define PLATFORM_LONG_OPTIONS
+#endif
 
-#ifdef HAVE_LIBCAP
-enum CapMode {
-   CAP_MODE_NONE,
-   CAP_MODE_BASIC,
-   CAP_MODE_STRICT
-};
+#ifndef PLATFORM_LONG_OPTIONS_USAGE
```
cgzones, line 43: dito

htop.c
```diff
@@ -35,37 +35,28 @@ in the source distribution for its full text.
...
+#define PLATFORM_LONG_OPTIONS_USAGE
 #endif
 
-static void printVersionFlag(void) {
-   fputs(PACKAGE " " VERSION "\n", stdout);
+static void printVersionFlag(const char* name) {
+   fprintf(stdout, "%s " VERSION "\n", name);
```
cgzones, line 48: Maybe use `printf(...` instead of `fprintf(stdout, ...` ?

htop.c
```diff
@@ -35,37 +35,28 @@ in the source distribution for its full text.
...
 }
 
-static void printHelpFlag(void) {
-   fputs(PACKAGE " " VERSION "\n"
+static void printHelpFlag(const char* name) {
+   fprintf(stdout, "%s " VERSION "\n"
```
cgzones, line 52: dito

linux/Platform.h
```diff
@@ -27,6 +27,19 @@ in the source distribution for its full text.
...
 #endif
 
+#ifdef HAVE_LIBCAP
+   #define PLATFORM_LONG_OPTIONS \
+      {"drop-capabilities", optional_argument, 0, 128},
+   #define PLATFORM_LONG_OPTIONS_USAGE \
+         "   --drop-capabilities[=none|basic|strict] Drop Linux capabilities when running as root\n" \
+         "                                none - do not drop any capabilities\n" \
+         "                                basic (default) - drop all capabilities not needed by htop\n" \
+         "                                strict - drop all capabilities except those needed for core functionality\n"
```
BenBE, line 37: Prone to format string issues.

htop.c
```diff
@@ -76,9 +59,9 @@ static void printHelpFlag(void) {
          "-V --version                    Print version info\n"
          "\n"
          "Long options may be passed with a single dash.\n\n"
-         "Press F1 inside " PACKAGE " for online help.\n"
-         "See 'man " PACKAGE "' for more information.\n",
-         stdout);
+         "Press F1 inside %s for online help.\n"
+         "See 'man %s' for more information.\n",
+         name, name, name);
```
BenBE, line 64: Why not use positional arguments here? Allows for re-use of the same argument.

natoscott: Ah, good idea!  Trying it out though, I get...

```
htop.c: In function ‘printHelpFlag’:
htop.c:64:10: warning: ISO C does not support %n$ operand number formats [-Wformat=]
   64 |          name);
      |          ^~~~
```

linux/Platform.h
natoscott: I guess.  Any better alternative you can suggest?

cgzones: One could split the `printf()` call in `htop.c:printHelpFlag()` into 
```c
printf("%s " VERSION "\n"
         COPYRIGHT "\n"
         "Released under the GNU GPLv2.\n\n"
         "-C --no-color                   Use a monochrome color scheme\n"
         ...
         "-V --version                    Print version info\n",
         name);

Platform_printLongOptionsUsage(name);

printf("\n"
         "Long options may be passed with a single dash.\n\n"
         "Press F1 inside %s for online help.\n"
         "See 'man %s' for more information.\n",
         name, name);
```

but I am also fine with the current implementation, relying on `-Wformat` to alert on format string issues.

cgzones (approved):

linux/Platform.h
natoscott: @cgzones nice - I like that, it's cleaner and more flexible.  Will implement shortly and then merge, I think that's the last thing here.

PR #566, opened 2021-03-17
MemoryMeter: show shared memory before cached

cgzones:
Shared memory is less free-able than cached memory.

Show it beforehand.

Follow up of: #558


BenBE (approved):

natoscott (approved):
LGTM too - one thing maybe for the future, it would be nice to change those hard coded values array offsets to something symbolic (enum) ... then the code would describe itself, and the new comments here (and commented-out code lines) wouldn't be needed here.

Obviously, doing this affects alot of meters though.

PR #572, opened 2021-03-21
Rework TTY column

cgzones:
* Rename internal identifier from `TTY_NR` to just `TTY`
* Unify column header on platforms
* Use `devname(3)` on BSD derivate to show the actual terminal,
  simplifies current FreeBSD implementation.
* Use `unsigned long int` as id type, to fit `dev_t` on Linux.

Only on Solaris the terminal path is not yet resolved.

Plus some *BSD cosmetics.

### DIFF
--- a/Process.c
+++ b/Process.c
@@ -374,17 +374,15 @@ void Process_writeField(const Process* this, RichString* str, ProcessField field
       xSnprintf(buffer, n, "%*d ", Process_pidDigits, this->tgid);
       break;
    case TPGID: xSnprintf(buffer, n, "%*d ", Process_pidDigits, this->tpgid); break;
-   case TTY_NR: {
-      unsigned int major = major(this->tty_nr);
-      unsigned int minor = minor(this->tty_nr);
-      if (major == 0 && minor == 0) {
+   case TTY:
+      if (!this->tty_name) {
          attr = CRT_colors[PROCESS_SHADOW];
-         xSnprintf(buffer, n, "(none)   ");
+         xSnprintf(buffer, n, "(no tty) ");
       } else {
-         xSnprintf(buffer, n, "%3u:%3u  ", major, minor);
+         const char* name = String_startsWith(this->tty_name, "/dev/") ? (this->tty_name + strlen("/dev/")) : this->tty_name;
+         xSnprintf(buffer, n, "%-8s ", name);
       }
       break;
-   }
    case USER:
       if (Process_getuid != this->st_uid)
          attr = CRT_colors[PROCESS_SHADOW];
@@ -431,6 +429,7 @@ void Process_display(const Object* cast, RichString* out) {
 void Process_done(Process* this) {
    assert (this != NULL);
    free(this->comm);
+   free(this->tty_name);
 }
 
 static const char* Process_getCommandStr(const Process* p) {
@@ -605,8 +604,9 @@ int Process_compareByKey_Base(const Process* p1, const Process* p2, ProcessField
       return SPACESHIP_NUMBER(p1->tgid, p2->tgid);
    case TPGID:
       return SPACESHIP_NUMBER(p1->tpgid, p2->tpgid);
-   case TTY_NR:
-      return SPACESHIP_NUMBER(p1->tty_nr, p2->tty_nr);
+   case TTY:
+      /* Order no tty last */
+      return SPACESHIP_DEFAULTSTR(p1->tty_name, p2->tty_name, "\x7F");
    case USER:
       return SPACESHIP_NULLSTR(p1->user, p2->user);
    default:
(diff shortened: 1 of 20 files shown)


darwin/DarwinProcess.c
```diff
@@ -272,9 +272,16 @@ void DarwinProcess_setFromKInfoProc(Process* proc, const struct kinfo_proc* ps,
       proc->tpgid = ps->kp_eproc.e_tpgid;
       proc->tgid = proc->pid;
       proc->st_uid = ps->kp_eproc.e_ucred.cr_uid;
-      /* e_tdev = (major << 24) | (minor & 0xffffff) */
-      /* e_tdev == -1 for "no device" */
-      proc->tty_nr = ps->kp_eproc.e_tdev & 0xff; /* TODO tty_nr is unsigned */
+      {
```
natoscott, line 275: do we need the extra scoping/braces here?

dragonflybsd/DragonFlyBSDProcessList.c
```diff
@@ -396,7 +396,16 @@ void ProcessList_goThroughEntries(ProcessList* super, bool pauseProcessUpdate) {
          proc->tgid = kproc->kp_pid;		// thread group id
          proc->pgrp = kproc->kp_pgid;		// process group id
          proc->session = kproc->kp_sid;
-         proc->tty_nr = kproc->kp_tdev;		// control terminal device number
+         {
```
natoscott, line 399: ditto

freebsd/FreeBSDProcessList.c
```diff
@@ -543,6 +469,16 @@ void ProcessList_goThroughEntries(ProcessList* super, bool pauseProcessUpdate) {
          ProcessList_add(super, proc);
          proc->comm = FreeBSDProcessList_readProcessName(fpl->kd, kproc, &proc->basenameOffset);
          fp->jname = FreeBSDProcessList_readJailName(kproc);
+         {
```
natoscott, line 472: ditto

openbsd/OpenBSDProcessList.c
```diff
@@ -225,7 +225,16 @@ static void OpenBSDProcessList_scanProcs(OpenBSDProcessList* this) {
          proc->tpgid = kproc->p_tpgid;
          proc->tgid = kproc->p_pid;
          proc->session = kproc->p_sid;
-         proc->tty_nr = kproc->p_tdev;
+         {
```
natoscott, line 228: ditto

natoscott (approved):
minor optional change suggested, LGTM

Process.c
```diff
@@ -374,17 +374,15 @@ void Process_writeField(const Process* this, RichString* str, ProcessField field
...
    case TPGID: xSnprintf(buffer, n, "%*d ", Process_pidDigits, this->tpgid); break;
-   case TTY_NR: {
-      unsigned int major = major(this->tty_nr);
-      unsigned int minor = minor(this->tty_nr);
-      if (major == 0 && minor == 0) {
+   case TTY:
+      if (!this->tty_name) {
          attr = CRT_colors[PROCESS_SHADOW];
-         xSnprintf(buffer, n, "(none)   ");
+         xSnprintf(buffer, n, "(no tty) ");
```
BenBE, line 380: Shouldn't this be `N/A` to be consistent with other columns when data is unavailable?

Process.c
```diff
@@ -605,8 +604,11 @@ int Process_compareByKey_Base(const Process* p1, const Process* p2, ProcessField
       return SPACESHIP_NUMBER(p1->tgid, p2->tgid);
    case TPGID:
       return SPACESHIP_NUMBER(p1->tpgid, p2->tpgid);
-   case TTY_NR:
-      return SPACESHIP_NUMBER(p1->tty_nr, p2->tty_nr);
+   case TTY:
+      /* Order no tty last */
+      if (!p1->tty_name || !p2->tty_name)
+         return -SPACESHIP_NUMBER(p1->tty_name, p2->tty_name);
+      return strcmp(p1->tty_name, p2->tty_name);
```
BenBE, line 611: Use `SPACESHIP_STRING`?

darwin/DarwinProcess.c
BenBE: Not strictly needed, but IMHO they make things more readable. Otherwise you could reorder this after the translated flag and separate it by some blank line.

BenBE (commented):
LGTM. Some small stuff left …

darwin/DarwinProcess.c
cgzones: Dropped the extra blocks.

Process.c
cgzones: Now using `SPACESHIP_DEFAULTSTR(p1->tty_name, p2->tty_name, "\x7F")`

Process.c
cgzones: In my understanding 'N/A' means some error happened or htop does not have sufficient privileges to gather the data.
But (most) processes just do not have a controlling terminal, so `(no tty)` is a very common case.

BenBE (approved):

PR #574, opened 2021-03-22
Merging all the points related to calculating time in one place

smalinux:
The end goal is to consolidate all the points in htop that can only work in
live-only mode today, so that will be able to inject PCP archive mode and have
a chance at it working.
The biggest problem we've got at this moment is all the places that are
independently asking the kernel to 'give me the time right now'.
Each of those needs to be audited and ultimately changed to allow platforms to
manage their own idea of time.
So, all the calls to gettimeofday(2) and time(2) are potential problems.
Ultimately I want to get these down to just one or two.

### DIFF
--- a/Compat.c
+++ b/Compat.c
@@ -11,17 +11,11 @@ in the source distribution for its full text.
 
 #include <errno.h>
 #include <fcntl.h> // IWYU pragma: keep
-#include <time.h>
 #include <unistd.h>
 #include <sys/stat.h>
 #include <sys/types.h> // IWYU pragma: keep
 
 #include "XUtils.h" // IWYU pragma: keep
-
-#ifdef HAVE_HOST_GET_CLOCK_SERVICE
-#include <mach/clock.h>
-#include <mach/mach.h>
-#endif
 
 
 int Compat_faccessat(int dirfd,
@@ -122,32 +116,4 @@ ssize_t Compat_readlinkat(int dirfd,
    return readlink(path, buf, bufsize);
 
 #endif
-}
-
-int Compat_clock_monotonic_gettime(struct timespec *tp) {
-
-#if defined(HAVE_CLOCK_GETTIME)
-
-   return clock_gettime(CLOCK_MONOTONIC, tp);
-
-#elif defined(HAVE_HOST_GET_CLOCK_SERVICE)
-
-   clock_serv_t cclock;
-   mach_timespec_t mts;
-
-   host_get_clock_service(mach_host_self(), SYSTEM_CLOCK, &cclock);
-   clock_get_time(cclock, &mts);
-   mach_port_deallocate(mach_task_self(), cclock);
-
-   tp->tv_sec = mts.tv_sec;
-   tp->tv_nsec = mts.tv_nsec;
-
-   return 0;
-
-#else
-
-#error No Compat_clock_monotonic_gettime() implementation!
-
-#endif
-
 }
--- a/Process.h
+++ b/Process.h
@@ -174,8 +174,8 @@ typedef struct Process_ {
    /*
     * Internal time counts for showing new and exited processes.
     */
-   time_t seenTs;
-   time_t tombTs;
+   uint64_t seenStampMs;
+   uint64_t tombStampMs;
 
    /*
     * Internal state for tree-mode.
--- a/ProcessList.h
+++ b/ProcessList.h
@@ -48,6 +48,9 @@ typedef struct ProcessList_ {
    Hashtable* displayTreeSet;
    Hashtable* draftingTreeSet;
 
+   struct timeval timestamp;  /* time of the current sample */
+   uint64_t timestampMs;      /* current time in milliseconds */
+
    Panel* panel;
    int following;
    uid_t userId;
--- a/ProcessList.h
+++ b/ProcessList.h
@@ -48,8 +48,9 @@ typedef struct ProcessList_ {
    Hashtable* displayTreeSet;
    Hashtable* draftingTreeSet;
 
-   struct timeval timestamp;  /* time of the current sample */
-   uint64_t timestampMs;      /* current time in milliseconds */
+   struct timeval realtime;   /* time of the current sample */
+   uint64_t realtimeMs;       /* current time in milliseconds */
+   uint64_t monotonicMs;      /* same, but from monotonic clock */
 
    Panel* panel;
    int following;
@@ -79,8 +80,6 @@ typedef struct ProcessList_ {
    memory_t cachedSwap;
 
    unsigned int cpuCount;
-
-   time_t scanTs;
 } ProcessList;
 
 ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* pidMatchList, uid_t userId);
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -189,6 +189,13 @@ static void LinuxProcessList_updateCPUcount(ProcessList* super, FILE* stream) {
    }
 }
 
+static void LinuxProcessList_updateTime(LinuxProcessList* this) {
+   ProcessList* pl = &(this->super);
+
+   gettimeofday(&pl->timestamp, NULL);
+   pl->timestampMs = (uint64_t)&pl->timestamp.tv_sec * 1000 + (uint64_t)pl->timestamp.tv_usec / 1000;
+}
+
 ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* pidMatchList, uid_t userId) {
    LinuxProcessList* this = xCalloc(1, sizeof(LinuxProcessList));
    ProcessList* pl = &(this->super);
@@ -1968,6 +1975,7 @@ void ProcessList_goThroughEntries(ProcessList* super, bool pauseProcessUpdate) {
    LinuxProcessList* this = (LinuxProcessList*) super;
    const Settings* settings = super->settings;
 
+   LinuxProcessList_updateTime(this);
    LinuxProcessList_scanMemoryInfo(super);
    LinuxProcessList_scanHugePages(this);
    LinuxProcessList_scanZfsArcstats(this);
@@ -1989,10 +1997,6 @@ void ProcessList_goThroughEntries(ProcessList* super, bool pauseProcessUpdate) {
       return;
    }
 
-   struct timeval tv;
-   gettimeofday(&tv, NULL);
-   unsigned long long now = tv.tv_sec * 1000ULL + tv.tv_usec / 1000ULL;
-
    /* PROCDIR is an absolute path */
    assert(PROCDIR[0] == '/');
 #ifdef HAVE_OPENAT
@@ -2001,5 +2005,5 @@ void ProcessList_goThroughEntries(ProcessList* super, bool pauseProcessUpdate) {
    openat_arg_t rootFd = "";
 #endif
 
-   LinuxProcessList_recurseProcTree(this, rootFd, PROCDIR, NULL, period, now);
+   LinuxProcessList_recurseProcTree(this, rootFd, PROCDIR, NULL, period, super->timestampMs);
 }
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -189,13 +189,6 @@ static void LinuxProcessList_updateCPUcount(ProcessList* super, FILE* stream) {
    }
 }
 
-static void LinuxProcessList_updateTime(LinuxProcessList* this) {
-   ProcessList* pl = &(this->super);
-
-   gettimeofday(&pl->timestamp, NULL);
-   pl->timestampMs = (uint64_t)&pl->timestamp.tv_sec * 1000 + (uint64_t)pl->timestamp.tv_usec / 1000;
-}
-
 ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* pidMatchList, uid_t userId) {
    LinuxProcessList* this = xCalloc(1, sizeof(LinuxProcessList));
    ProcessList* pl = &(this->super);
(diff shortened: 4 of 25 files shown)


Compat.c
```diff
@@ -128,7 +128,7 @@ int Compat_clock_monotonic_gettime(struct timespec *tp) {
 
 #if defined(HAVE_CLOCK_GETTIME)
 
-   return clock_gettime(CLOCK_MONOTONIC, tp);
```
a3f, line 131: This change should be in a separate commit.

XUtils.c
```diff
@@ -303,3 +304,8 @@ ssize_t xReadfileat(openat_arg_t dirfd, const char* pathname, void* buffer, size
 
    return readfd_internal(fd, buffer, count);
 }
+
+int xClock_gettime (clockid_t __clock_id, struct timespec *__tp) {
+   return clock_gettime (__clock_id, __tp);
```
a3f, line 309: What's the added value of having this wrapper? Also `__` prefixed identifiers are reserved for the C implementation and/or denote internal stuff. You could just call it `clock_id` and `tp` directly here.

linux/Platform.h
```diff
@@ -72,7 +72,7 @@ bool Platform_getDiskIO(DiskIOData* data);
 
 bool Platform_getNetworkIO(NetworkIOData* data);
 
-void Platform_getBattery(double *percent, ACPresence *isOnAC);
+void Platform_getBattery(Meter* this, double* percent, ACPresence* isOnAC);
```
a3f, line 75: It's arguably better to have `T *t` instead of `T* t`, because when you define multiple variables of the same type in one go you would write `T *t1, *t2`, not `T* t1, t2`. Regardless, such changes are out-of-place for this commit.

natoscott:
@smalinux looks like there's an inconsistency between a header and associated function call (LinuxProcessList_updateCPUcount?) - possibly merge fallout.  Its causing CI issues currently.

linux/LinuxProcessList.c
```diff
@@ -189,6 +189,20 @@ static void LinuxProcessList_updateCPUcount(ProcessList* super, FILE* stream) {
    }
 }
 
+static void LinuxProcessList_updateTime(LinuxProcessList* this) {
```
a3f, line 192: How often would this function be called?

linux/LinuxProcessList.c
```diff
@@ -1971,6 +1985,7 @@ void ProcessList_goThroughEntries(ProcessList* super, bool pauseProcessUpdate) {
    LinuxProcessList_scanMemoryInfo(super);
    LinuxProcessList_scanHugePages(this);
    LinuxProcessList_scanZfsArcstats(this);
+   LinuxProcessList_updateCPUcount(this);
```
a3f, line 1988: copy-paste error? Shouldn't this be `LinuxProcessList_updateTime`?

BenBE:
Also AFAIR there is a timestamp passed around for each update already. Please check if there are any functions that don't currently already use that timestamp for their timing needs.

linux/LinuxProcessList.c
```diff
@@ -1989,9 +2004,7 @@ void ProcessList_goThroughEntries(ProcessList* super, bool pauseProcessUpdate) {
       return;
    }
 
-   struct timeval tv;
-   gettimeofday(&tv, NULL);
-   unsigned long long now = tv.tv_sec * 1000ULL + tv.tv_usec / 1000ULL;
+   unsigned long long now = super->timestamp.tv_sec * 1000ULL + super->timestamp.tv_usec / 1000ULL;
```
a3f, line 2007: Given that it's repeated quite a few times, a prerequisite patch adding a `tv_to_msec()` might be useful.

a3f (changes requested):
Having to list multiple bullet points is a good indicator a commit should be split up into a number of commits where each message just describes a single issue.

You can still give an overview in the cover-letter (here, PR caption), but the individual commits should focus on the backstory of the changeset at hand.

PR also seems to introduce a regression.

linux/Platform.h
BenBE: Please note the general coding style, which seems to prefer `T* t`.

a3f: I see. Please dismiss this then. :)

linux/LinuxProcessList.c
BenBE: Usually once before each process scan. In normal operation on bare OS at most about 10 times a second (minimum interval being `.1` seconds) …

linux/LinuxProcessList.c
smalinux: It exists in two different places, each one of them has its own calculation, the rest of the duplicate places are removed.

smalinux:
> Also AFAIR there is a timestamp passed around for each update already. Please check if there are any functions that don't currently already use that timestamp for their timing needs.

I used `$ git grep "gettimeofday"`, `$ git grep "time.h"` and `$ git grep "timestamp"` so I 'think' all places that used timestamp exists in one place (my changes)

smalinux:
I retreated all 'xClock' related changes, I think it will not affect PCP-related work, so I left clock_gettime(2) as-is at this stage.

Also, I retreated from removing `#include <time.h>` although did not use it, because other Platforms may depend on this, I think this may be cause problems.

BenBE:
The CI still reports one issue:

```
	clang -DHAVE_CONFIG_H -I.  -DNDEBUG -DGCC_PRINTF -DGCC_SCANF -D_FORTIFY_SOURCE=2 -Wall -Wcast-align -Wcast-qual -Wextra -Wfloat-equal -Wformat=2 -Winit-self -Wmissing-format-attribute -Wmissing-noreturn -Wmissing-prototypes -Wpointer-arith -Wshadow -Wstrict-prototypes -Wundef -Wunused -Wwrite-strings -Wnull-dereference -Werror -pedantic -std=c99 -D_XOPEN_SOURCE_EXTENDED -DSYSCONFDIR="\"/usr/local/etc\"" -I"./darwin" -g -O2 -MT BatteryMeter.o -MD -MP -MF $depbase.Tpo -c -o BatteryMeter.o BatteryMeter.c &&\
BatteryMeter.c:28:40: error: too many arguments to function call, expected 2, have 3
   Platform_getBattery(this, &percent, &isOnAC);
   ~~~~~~~~~~~~~~~~~~~                 ^~~~~~~
./darwin/Platform.h:70:1: note: 'Platform_getBattery' declared here
void Platform_getBattery(double *percent, ACPresence *isOnAC);
^
1 error generated.
make[1]: *** [BatteryMeter.o] Error 1
```

linux/LinuxProcessList.c
a3f: I can't follow. I still see it here and in `LinuxProcessList_updateTime`.

ProcessList.h
```diff
@@ -48,6 +48,12 @@ typedef struct ProcessList_ {
    Hashtable* displayTreeSet;
    Hashtable* draftingTreeSet;
 
+   struct timeval timestamp;  /* time of the current sample */
+   struct timeval previous;   /* time of the previous sample */
+   uint64_t timestampMs;   /* current time in milliseconds */
+   uint64_t previousMs;		/* previous time in milliseconds */
+   uint64_t intervalMs;		/* milliseconds between 2 samples */
```
cgzones, line 55: The fields `previous`, `previousMs` and `intervalMs` are currently only assigned, but never used.
I suppose they are getting used in future additions?

Also, to please my eyes, please align the comments.

linux/LinuxProcessList.c
```diff
@@ -189,6 +189,20 @@ static void LinuxProcessList_updateCPUcount(ProcessList* super, FILE* stream) {
...
 
+static void LinuxProcessList_updateTime(LinuxProcessList* this) {
+   ProcessList* pl = &(this->super);
+
+   struct timeval tv;
+   static uint64_t cached_last_update;
+   gettimeofday(&tv, NULL);
+
+   pl->timestamp     = tv;
+   pl->timestampMs   = (uint64_t)tv.tv_sec * 1000 + (uint64_t)tv.tv_usec / 1000;
```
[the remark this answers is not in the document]
cgzones: I think it is OK to compute the millisecond value here once every cycle.

linux/LinuxProcessList.c
```diff
@@ -189,6 +189,20 @@ static void LinuxProcessList_updateCPUcount(ProcessList* super, FILE* stream) {
...
+   pl->intervalMs    = pl->timestampMs - pl->previousMs;
+   pl->previousMs    = cached_last_update;
+   cached_last_update = pl->timestampMs;
```
cgzones, line 203: Please align all the assignments

linux/LinuxProcessList.c
```diff
@@ -1971,6 +1985,7 @@ void ProcessList_goThroughEntries(ProcessList* super, bool pauseProcessUpdate) {
    LinuxProcessList_scanMemoryInfo(super);
    LinuxProcessList_scanHugePages(this);
    LinuxProcessList_scanZfsArcstats(this);
+   LinuxProcessList_updateTime(this);
```
cgzones, line 1988: Maybe insert this function call as first action, so every function can assume the data is up-to-date.

linux/LinuxProcessList.c
cgzones: Can't `now` be replaced by `pl->timestampMs`?

ProcessList.h
smalinux: > The fields previous, previousMs and intervalMs are currently only assigned, but never used.
I suppose they are getting used in future additions?

Yes, should I leave them as-is? or remove them for now?

linux/LinuxProcessList.c
smalinux: If `unsigned long long now` == `uint64_t pl->timestampMs`? Yes :) (and I 'think' it-is)

cgzones:
I think we need to reorder the updates in `Screenmanager.c`:
```diff
diff --git a/ScreenManager.c b/ScreenManager.c
index c4fbfee..23a635f 100644
--- a/ScreenManager.c
+++ b/ScreenManager.c
@@ -105,13 +105,13 @@ static void checkRecalculation(ScreenManager* this, double* oldTime, int* sortTi
 
    if (*rescan) {
       *oldTime = newTime;
-      // always update header, especially to avoid gaps in graph meters
-      Header_updateData(this->header);
       ProcessList_scan(pl, this->state->pauseProcessUpdate);
       if (!this->state->pauseProcessUpdate && (*sortTimeout == 0 || this->settings->treeView)) {
          ProcessList_sort(pl);
          *sortTimeout = 1;
       }
+      // always update header, especially to avoid gaps in graph meters
+      Header_updateData(this->header);
       *redraw = true;
    }
    if (*redraw) {
```

else the NetworkIO and DiskIO meters do not see the updated timestamps.

BenBE:
That's already wrong with the current code flow, as the process update updates some fields for the task counts displayed in the header …

natoscott:
> I retreated all 'xClock' related changes, I think it will not affect PCP-related work, so I left clock_gettime(2) as-is at this stage.

We'll need to come back to it later - but for now this PR is looking like a good standalone improvement (independent of the needs with historical reporting via PCP).

linux/LinuxProcessList.c
natoscott: There's still a separate gettimeofday call in checkRecalculation ScreenManager.c too - isn't that the very first place we could capture the 'current time of sampling' (perhaps via a new routine named like Platform_getCurrentTime?)  Then we'd not need a later LinuxProcessList_updateTime call.

smalinux: @natoscott Excellent, I can't call Platform_getCurrentTime directly from ScreenManager.c (this causes problems with CI and other OSs)
So, I made ProcessList_getCurrentTime.
I think this's optimal because our time-related variables exist within ProcessList struct (ProcessList.h)  and my new function in ProcessList.c. I could call it directly from there also.

natoscott: @smalinux I'll send some code through shortly - we should be able to call platform code from ScreenManager.c if we need to.  I'll need to reproduce that CI failure though to see the exact issue.

natoscott:
@smalinux I've created a branch with a series of updates: https://github.com/natoscott/htop/tree/smalinux-CtrTime

Could you review and merge with your branch if it looks OK?   This adds code to handle both realtime (gettimeofday) and monotonic clock accesses, with a single realtime clock access at the start of sampling (i.e. in ScreenManager.c).

Process.h
```diff
@@ -174,8 +174,8 @@ typedef struct Process_ {
    /*
     * Internal time counts for showing new and exited processes.
     */
-   time_t seenTs;
-   time_t tombTs;
+   uint64_t seenMs;
+   uint64_t tombMs;
```
BenBE, line 178: Why rename these fields?

BenBE (commented):
Mostly LGTM. Not yet happy with the rename of `seenTs` and `tombTs`.

Process.h
natoscott: > [...] let's change it back if its confusing.

@smalinux are you happy to do the rename?  I'm about to go on holidays for a few days.  :)

ProcessList.h
natoscott: > That's already wrong with the current code flow, as the process update updates some fields for the task counts displayed in the header …

I've opened https://github.com/htop-dev/htop/pull/587 to ensure this bug fix goes in as a standalone fix, as discussed on IRC.

Process.h
natoscott: @BenBE oh, also, time_t is typically a time-in-seconds (since the epoch) - hence the change in type (i.e. that part should stay as-is IMO).

smalinux: > @smalinux are you happy to do the rename?

@natoscott Yes, certainly, happy holiday :)

BenBE: I'm both fine with the type change (that should stay) and also with the `Ms` marker at the end. It's just a bit unfortunate to drop the mid-part `Ts` altogether. Maybe `seenStampMs`/`tombStampMs`?

natoscott: @BenBE @smalinux that sounds good to me

linux/LinuxProcessList.c
```diff
@@ -189,6 +189,7 @@ static void LinuxProcessList_updateCPUcount(ProcessList* super, FILE* stream) {
    }
 }
 
+
```
BenBE, line 192: No need for this blank line.

BenBE (approved):
LGTM.

PR #575, opened 2021-03-23
Abstract htop main function to allow for a platform binary

natoscott:
One review request relating to the PCP platform is to have
a clearly separate binary from the regular htop so that we
have no confusion as to what is being requested to run, to
aid debugging, and a bunch of other good reasons.

This commit renames htop.c to CommandLine.c and provides a
minimal htop main function for 'native' platforms to use.
The PCP version of this will setup libpcp.so and then call
the same CommandLine_run function as regular htop.

Related to https://github.com/htop-dev/htop/pull/536

### DIFF
--- a/CommandLine.c
+++ b/CommandLine.c
@@ -1,0 +1,372 @@
+/*
+htop - CommandLine.c
+(C) 2004-2011 Hisham H. Muhammad
+(C) 2020-2021 htop dev team
+Released under the GNU GPLv2, see the COPYING file
+in the source distribution for its full text.
+*/
+
+#include "config.h" // IWYU pragma: keep
+
+#include "CommandLine.h"
+
+#include <assert.h>
+#include <getopt.h>
+#include <locale.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+#include <unistd.h>
+
+#include "Action.h"
+#include "CRT.h"
+#include "Hashtable.h"
+#include "Header.h"
+#include "IncSet.h"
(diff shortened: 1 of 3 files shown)


CommandLine.c
```diff
@@ -0,0 +1,371 @@
...
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+#include <unistd.h>
+
+#include "Action.h"
+#include "CRT.h"
+#include "CommandLine.h"
```
BenBE, line 23: The `#include "CommandLine.h"` should go right after `#include "config.h"`.

BenBE (changes requested):
LGTM. Although the noise in `Makefile.am` is kinda a bit overkill (at least for now).

CommandLine.c
natoscott: > LGTM. Although the noise in `Makefile.am` is kinda a bit overkill (at least for now).

Yep - makes more sense once we have the PCP front-end of course.

BenBE: ACK. That's what I meant with the parens …

BenBE (approved):

CommandLine.h
```diff
@@ -0,0 +1,7 @@
+#ifndef HEADER_CommandLine
+#define HEADER_CommandLine
+
```
cgzones, line 3: Do you like to add a Copyright header?

cgzones (approved):

CommandLine.h
natoscott: Done.

PR #580, opened 2021-03-27
don't include offline CPUs in summary for OpenBSD

sthen:
By default, OpenBSD disables SMT (hyperthreading) cpu pseudo-cores.
This can be changed at runtime by setting the hw.smt sysctl so they
may become active later, therefore they are still present in cpu
stat structures but are marked as offline.

As done with native top(1), this drops them from the cpu summary
graphs.

### DIFF
--- a/openbsd/OpenBSDProcessList.c
+++ b/openbsd/OpenBSDProcessList.c
@@ -37,11 +37,15 @@ static int pageSize;
 static int pageSizeKB;
 
 ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* pidMatchList, uid_t userId) {
-   const int mib[] = { CTL_HW, HW_NCPU };
+   const int nmib[] = { CTL_HW, HW_NCPU };
+   const int mib[] = { CTL_HW, HW_NCPUONLINE };
    const int fmib[] = { CTL_KERN, KERN_FSCALE };
+   int ncmib[] = { CTL_KERN, KERN_CPUSTATS, 0 };
    int r;
+   unsigned int cpu_index_c = 0, ncpu;
    size_t size;
    char errbuf[_POSIX2_LINE_MAX];
+   struct cpustats cpu_stats;
 
    OpenBSDProcessList* opl = xCalloc(1, sizeof(OpenBSDProcessList));
    ProcessList* pl = (ProcessList*) opl;
@@ -53,6 +57,12 @@ ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* pidMatchList, ui
       pl->cpuCount = 1;
    }
    opl->cpus = xCalloc(pl->cpuCount + 1, sizeof(CPUData));
+
+   size = sizeof(int);
+   r = sysctl(nmib, 2, &ncpu, &size, NULL, 0);
+   if (r < 0) {
+      ncpu = pl->cpuCount;
+   }
 
    size = sizeof(fscale);
    if (sysctl(fmib, 2, &fscale, &size, NULL, 0) < 0) {
@@ -75,6 +85,21 @@ ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* pidMatchList, ui
    }
 
    opl->cpuSpeed = -1;
+
+   size = sizeof(cpu_stats);
+   for (unsigned int i = 0; i < ncpu; i++) {
+      ncmib[2] = i;
+      if (sysctl(ncmib, 3, &cpu_stats, &size, NULL, 0) < 0) {
+         CRT_fatalError("ncmib sysctl call failed");
+      }
+      if (cpu_stats.cs_flags & CPUSTATS_ONLINE) {
+         opl->cpus[cpu_index_c].cpuIndex = i;
+         cpu_index_c++;
+      }
+
+      if (cpu_index_c == pl->cpuCount)
+         break;
+   }
 
    return pl;
 }
@@ -340,7 +365,7 @@ static void OpenBSDProcessList_scanCPUTime(OpenBSDProcessList* this) {
    u_int64_t avg[CPUSTATES] = {0};
 
    for (unsigned int i = 0; i < this->super.cpuCount; i++) {
-      getKernelCPUTimes(i, kernelTimes);
+      getKernelCPUTimes(this->cpus[i].cpuIndex, kernelTimes);
       CPUData* cpu = this->cpus + i + 1;
       kernelCPUTimesToHtop(kernelTimes, cpu);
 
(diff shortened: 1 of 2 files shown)


openbsd/OpenBSDProcessList.h
```diff
@@ -43,7 +43,7 @@ typedef struct OpenBSDProcessList_ {
 
    CPUData* cpus;
    int cpuSpeed;
-
+   int *cpuIndex;
```
cgzones, line 46: What is the point of this array?
It is only written to, but never read from.

sthen: Good point, it's a useless remnant of an earlier attempt at skipping offline cpus, I've pushed a commit to drop that.

BenBE:
Can you squash your commits? Makes the history cleaner and easier to follow.

sthen:
On 2021/03/28 13:37, BenBE wrote:
> Can you squash your commits? Makes the history cleaner and easier to follow.
> 
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub, or unsubscribe.*
> 

If you can tell me what to type to do that, yes. (I don't normally use git).

BenBE:
When on your branch:
```bash
# Fetch latest changes
git fetch --all

# Apply latest changes
git pull --rebase your-repo your-branch

# Base local changes on upstream master
git rebase -i htop-upstream/master

# Push the modified branch to your repo to update this PR.
git push --force-with-lease your-repo your-branch
```
In the editor that opens change the second line from `pick` to `fixup`. See also the notes given in the descriptive text in that editor.

openbsd/OpenBSDProcessList.c
```diff
@@ -76,6 +86,17 @@ ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* pidMatchList, ui
...
+   size = sizeof(cpu_stats);
+   for (unsigned int i = 0; i < ncpu; i++) {
+      ncmib[2] = i;
+      sysctl(ncmib, 3, &cpu_stats, &size, NULL, 0);
+      if (cpu_stats.cs_flags & CPUSTATS_ONLINE)
+         cpu_index_c++;
+
+      if (cpu_index_c == pl->cpuCount)
+         break;
+   }
```
cgzones, line 98: What is this loop about?
`cpu_index_c` is being increased and the loop is aborted when it reaches a specific value, but there are no further implications.

sthen: Hmm, looks like I missed a bit when I was merging the patch which was originally done against 3.0.1. It will work on Intel CPUs without that loop (the skipped cpus are the higher numbered ones) but they're interspersed on AMD (skipped would be 1,3,5,7,..) I just pushed an updated version that I think should do the trick

sthen: I asked someone to test on AMD, this version is looking ok.

openbsd/OpenBSDProcessList.c
```diff
@@ -53,6 +57,13 @@ ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* pidMatchList, ui
       pl->cpuCount = 1;
    }
    opl->cpus = xCalloc(pl->cpuCount + 1, sizeof(CPUData));
+   opl->cpuIndex = xRealloc(opl->cpuIndex, pl->cpuCount * sizeof(int));
```
cgzones, line 60: `*opl` was initialized in this function, no need to realloc (just malloc (or calloc, see later comment))

openbsd/OpenBSDProcessList.c
```diff
@@ -76,6 +87,19 @@ ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* pidMatchList, ui
 
    opl->cpuSpeed = -1;
 
+   size = sizeof(cpu_stats);
+   for (unsigned int i = 0; i < ncpu; i++) {
+      ncmib[2] = i;
+      sysctl(ncmib, 3, &cpu_stats, &size, NULL, 0);
```
cgzones, line 93: please check for a sysctl failure

openbsd/OpenBSDProcessList.h
```diff
@@ -43,7 +43,7 @@ typedef struct OpenBSDProcessList_ {
 
    CPUData* cpus;
    int cpuSpeed;
-
+   int* cpuIndex;
```
cgzones, line 46: Maybe include the index in the `CPUData` struct (the one unused `int` of the average entry is negligible)

cgzones (changes requested):

openbsd/OpenBSDProcessList.c
```diff
@@ -76,6 +86,21 @@ ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* pidMatchList, ui
 
    opl->cpuSpeed = -1;
 
+   size = sizeof(cpu_stats);
+   for (unsigned int i = 0; i < ncpu; i++) {
+      ncmib[2] = i;
```
BenBE, line 91: Move actual initialization of `ncmib` here and make it const.
```suggestion
      const int ncmib[] = { CTL_KERN, KERN_CPUSTATS, i };
```

openbsd/OpenBSDProcessList.c
```diff
@@ -37,11 +37,15 @@ static int pageSize;
 static int pageSizeKB;
 
 ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* pidMatchList, uid_t userId) {
-   const int mib[] = { CTL_HW, HW_NCPU };
+   const int nmib[] = { CTL_HW, HW_NCPU };
+   const int mib[] = { CTL_HW, HW_NCPUONLINE };
    const int fmib[] = { CTL_KERN, KERN_FSCALE };
+   int ncmib[] = { CTL_KERN, KERN_CPUSTATS, 0 };
    int r;
+   unsigned int cpu_index_c = 0, ncpu;
```
BenBE, line 45: Single variable per line …

openbsd/OpenBSDProcessList.c
```diff
@@ -76,6 +86,21 @@ ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* pidMatchList, ui
 
    opl->cpuSpeed = -1;
 
+   size = sizeof(cpu_stats);
+   for (unsigned int i = 0; i < ncpu; i++) {
+      ncmib[2] = i;
+      if (sysctl(ncmib, 3, &cpu_stats, &size, NULL, 0) < 0) {
```
BenBE, line 92: Shouldn't `size` be re-initialized inside the loop?

BenBE (changes requested):

sthen:
Since you have a pretty clear idea of what changes you want, could one of you fix it up please? Most of my development work is with OpenBSD and we don't use git so what is probably a 2 minute job for you takes me about half an hour each time to figure out how to rebase the commit history how you want it. Thanks!

cgzones:
Applied with adjustments via feec16cbb53dabc6a52ef2f69a6a13798be82617.
Thanks for your contribution.

sthen:
Many thanks.

PR #583, opened 2021-03-28
RFC: TaskMeter: always show number of threads

cgzones:
Always show the number of kernel and userland threads, even when they
are disabled to not be shown in the process list.

The data is already available and might improve understanding the system
utilization.

Use a shadow color in case the kind of thread is hidden, else the normal
meter one.


TasksMeter.c
```diff
@@ -43,28 +39,20 @@ static void TasksMeter_display(const Object* cast, RichString* out) {
...
-      RichString_appendAscii(out, threadValueColor, buffer);
-      RichString_appendAscii(out, threadCaptionColor, " kthr");
-   }
+
+   RichString_appendAscii(out, settings->hideUserlandThreads ? CRT_colors[METER_SHADOW] : CRT_colors[METER_TEXT], ", ");
+   xSnprintf(buffer, sizeof(buffer), "%d", (int)this->values[1]);
+   RichString_appendAscii(out, settings->hideUserlandThreads ? CRT_colors[METER_SHADOW] : CRT_colors[TASKS_RUNNING], buffer);
+   RichString_appendAscii(out, settings->hideUserlandThreads ? CRT_colors[METER_SHADOW] : CRT_colors[METER_TEXT], " thr");
+
+
```
BenBE, line 50: No need for two blank lines.

BenBE (approved):
LGTM.

natoscott (approved):
LGTM

TasksMeter.c
cgzones: dropped

PR #585, opened 2021-03-30
FreeBSD: improve actual processor calculation logic

bland74:
Found that current htop(1) and system's top(1) calculations may be off. The reason is that FreeBSD has a bit more complicated logic which tracks actual cpu where a process is running on.


cgzones (approved):

cgzones:
Thanks, applied.

PR #414, opened 2020-12-19
Make highlighting of deleted executables configurable

BenBE:
Addresses #383. Also updates documentation of the highlighting.

Should go in AFTER #388 due to code dependencies.

PR #595, opened 2021-04-09
Linux: check processes for using deleted shared libraries

cgzones:
Shared libraries can be replaced by an upgrade, highlight processes
using deleted shared libraries.

Link with highlightDeletedExe setting, enabled by default.

Currently only checked on Linux.

### DIFF
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -517,18 +517,24 @@ static void LinuxProcessList_calcLibSize_helper(ATTR_UNUSED ht_key_t key, void* 
    *d += v->size;
 }
 
-static uint64_t LinuxProcessList_calcLibSize(openat_arg_t procFd) {
+static void LinuxProcessList_readMaps(LinuxProcess* process, openat_arg_t procFd, bool calcSize, bool checkDeletedLib) {
+   Process* proc = (Process*)process;
+
+   proc->usesDeletedLib = false;
+
    FILE* mapsfile = fopenat(procFd, "maps", "r");
    if (!mapsfile)
-      return 0;
+      return;
 
-   Hashtable* ht = Hashtable_new(64, true);
+   Hashtable* ht = NULL;
+   if (calcSize)
+      ht = Hashtable_new(64, true);
 
    char buffer[1024];
    while (fgets(buffer, sizeof(buffer), mapsfile)) {
       uint64_t map_start;
       uint64_t map_end;
-      char map_perm[5];
+      bool map_execute;
       unsigned int map_devmaj;
       unsigned int map_devmin;
       uint64_t map_inode;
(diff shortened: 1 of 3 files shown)


BenBE:
What about combining this with #414?

BenBE (commented):
Patch itself LGTM, but I still wonder if we really should split the two features, especially as #414 more or less implies this flag is also triggered.

cgzones:
Postpone after #414 to merge enable-logic with deleted executables.

alicektx:
...really really thank you for doing / implementing this ability into htop! had been using [this](https://github.com/gsauthof/utility#oldprocs) so far to check such...

BenBE:
@cgzones Could you update this now that #414 is merged?
Please note this earlier [review comment](ttps://github.com/htop-dev/htop/pull/595#pullrequestreview-632706242).

linux/LinuxProcessList.c
```diff
@@ -575,61 +580,65 @@ static uint64_t LinuxProcessList_calcLibSize(openat_arg_t procFd) {
...
+            readptr++;
+
+         if (*readptr != '/')
+            continue;
+
+         if (String_startsWith(readptr, "/memfd:"))
+            continue;
+
+         if (strstr(readptr, " (deleted)\n")) {
+            proc->usesDeletedLib = true;
```
BenBE, line 605: Mind this bugfix:
https://github.com/htop-dev/htop/commit/51ecc62d92d78d8ac5a3367e0ca8f4865880319d

You need to track the state transition to mark for `makeCommandStr`.

BenBE (changes requested):

linux/LinuxProcessList.c
```diff
@@ -1340,9 +1347,28 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
       if (settings->flags & PROCESS_FLAG_IO)
          LinuxProcessList_readIoFile(lp, procFd, pl->realtimeMs);
 
-      if (!LinuxProcessList_readStatmFile(lp, procFd, !!(settings->flags & PROCESS_FLAG_LINUX_LRS_FIX), pl->realtimeMs))
+      if (!LinuxProcessList_readStatmFile(lp, procFd))
          goto errorReadingProcess;
 
+      {
+         bool prev = proc->usesDeletedLib;
+         proc->usesDeletedLib = false;
```
BenBE, line 1355: Only reset this flag, once we actually re-calculate it. Maybe put directly before call to `LinuxProcessList_readMaps`?

BenBE (changes requested):

linux/LinuxProcessList.c
cgzones: true, thanks for catching

BenBE: I think that's why we do those code reviews ;-) Less stuff we need to fix later. :P

BenBE (approved):

PR #596, opened 2021-04-10
Convert process time to days if applicable

davezarzycki:
With big multicore machines, it's easy to accumulate process time.

### DIFF
--- a/Process.c
+++ b/Process.c
@@ -150,11 +150,15 @@ void Process_printTime(RichString* str, unsigned long long totalHundredths) {
    unsigned long long totalSeconds = totalHundredths / 100;
 
    unsigned long long hours = totalSeconds / 3600;
+   unsigned long long days = totalSeconds / 86400;
    int minutes = (totalSeconds / 60) % 60;
    int seconds = totalSeconds % 60;
    int hundredths = totalHundredths - (totalSeconds * 100);
    char buffer[10];
-   if (hours >= 100) {
+   if (days >= 100) {
+      xSnprintf(buffer, sizeof(buffer), "%7llud ", days);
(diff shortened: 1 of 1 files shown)


Process.c
```diff
@@ -150,11 +150,15 @@ void Process_printTime(RichString* str, unsigned long long totalHundredths) {
    unsigned long long totalSeconds = totalHundredths / 100;
 
    unsigned long long hours = totalSeconds / 3600;
+   unsigned long long days = totalSeconds / (24 * 3600);
```
BenBE, line 153: Prefer explicit `86400` and avoid parens.

davezarzycki: Okay. I'll fix that but may I ask why? Constant folding is something compilers are really good at.

BenBE: Two reasons: First is, that for consistency you'd otherwise also have to have the `3600` be written as `60 * 60`. Second is, that by using the `86400` as a single constant you avoid the additional parens that make the code at that place look unnecessarily complex when that value is a quite commonly used value.

Thus even though constant folding in compilers has become really good in recent years™, there's no additional value gained from it in this case. You could BTW also have used the value from `HOURS` in your calculation instead, if the emphasis really would have needed to be on the day having 24 hours.

cgzones:
@davezarzycki Is 100 days a good cut off point? This means hours are counted up to 2376.

davezarzycki:
I don't have any strong opinion here. If I were to "go nuts" micro-optimizing this, then I'd probably either have "hours" and "days" output support a couple of decimal places or I'd support output like "40d15h". I could see arguments either way.

Honestly, the "TIME" output is kind of silly once one has a lot of cores. A compute bound program of mine just finished that had a couple *years* of compute time despite only running about a week of real "wall clock" time.

Issue #591, opened 2021-04-07
Memory meter behaves strangely


BenBE:
Let's have a look at the following screenshot:
![image](https://user-images.githubusercontent.com/111196/113844190-8529df80-9794-11eb-977f-76dd4b64f060.png)

The memory meter seems totally inconsistent (also the used memory value), as this would indicate that Chromium uses less than 1GiB of memory, which is unlikely given the memory stats below. There's one large `tmpfs` that accounts for about 14.2G (14299228K), thus the value shown for shared memory is kinda reasonable; but the used memory value is not.

```
$ cat /proc/meminfo
MemTotal:       49194612 kB
MemFree:          693196 kB
MemAvailable:   16044012 kB
Buffers:         1715840 kB
Cached:         29206036 kB
SwapCached:            0 kB
Active:         32616500 kB
Inactive:       12612180 kB
Active(anon):   28372980 kB
Inactive(anon):  1847288 kB
Active(file):    4243520 kB
Inactive(file): 10764892 kB
Unevictable:         128 kB
Mlocked:             128 kB
SwapTotal:             0 kB
SwapFree:              0 kB
Dirty:               608 kB
Writeback:             0 kB
AnonPages:      14307064 kB
Mapped:          2191528 kB
Shmem:          16185716 kB
KReclaimable:     939988 kB
Slab:            1354232 kB
SReclaimable:     939988 kB
SUnreclaim:       414244 kB
KernelStack:       37536 kB
PageTables:       159536 kB
NFS_Unstable:          0 kB
Bounce:                0 kB
WritebackTmp:          0 kB
CommitLimit:    24597304 kB
Committed_AS:   51851836 kB
VmallocTotal:   34359738367 kB
VmallocUsed:       71560 kB
VmallocChunk:          0 kB
Percpu:            22656 kB
HardwareCorrupted:     0 kB
AnonHugePages:    169984 kB
ShmemHugePages:        0 kB
ShmemPmdMapped:        0 kB                                                                                                                                                                  
FileHugePages:         0 kB                                                                                                                                                                  
FilePmdMapped:         0 kB                                                                                                                                                                  
CmaTotal:              0 kB                                                                                                                                                                  
CmaFree:               0 kB                                                                                                                                                                  
HugePages_Total:       0                                                                                                                                                                     
HugePages_Free:        0                                                                                                                                                                     
HugePages_Rsvd:        0                                                                                                                                                                     
HugePages_Surp:        0                                                                                                                                                                     
[...]

PR #597, opened 2021-04-14
Linux: handle Shmem being part of Cached

cgzones:
See https://lore.kernel.org/patchwork/patch/648763/

Do not show twice by subtracting from Cached.

Closes: #591


BenBE (approved):
Patch LGTM and runtime values presented are sensible in the context of both massive `tmpfs` usage and partially swapped `tmpfs` mounts.

BenBE:
What I didn't test right now is the behaviour for partially swapped `tmpfs` when the usage exceeds available RAM and thus forces things to be at least partially swapped to fit (and yes, that's possible).

PR #598, opened 2021-04-14
Process print rework

cgzones:
Make functions formatting data for a process field column less error
prone, unify interfaces and improve some internals.

* Process_printBytes
  - rename from Process_humanNumber
  - take number in bytes, not kilobytes
  - handle petabytes
  - increase buffer to avoid crashes when the passed value is
    ~ ULLONG_MAX

* Process_printKBytes
  - add wrapper for Process_printBytes taking kilobytes keeping -1 as
  special value

* Process_printCount
  - rename from Process_colorNumber

* Process_printTime
  - add coloring parameter as other print functions
  - improve coloring and formatting for larger times

* Process_printRate
  - rename from Process_outputRate
  - use local buffer instead of passed one; this function prints to the
    RichString after all

 Linux: update IO fields

- fix header width of IO_READ_RATE

- save data in bytes (not kilobytes) to better compute rate

- fix rate data: multiply with 1000 to compensate time difference in
  milliseconds

- rename unit less variable now into realtimeMs

- use Process_printBytes(..., data * pageSize, ...) instead of
  Process_printKBytes(..., data * pageSizeKB, ...) to avoid wrapper

### DIFF
--- a/Process.c
+++ b/Process.c
@@ -49,8 +49,8 @@ void Process_setupColumnWidths() {
    assert(Process_pidDigits <= PROCESS_MAX_PID_DIGITS);
 }
 
-void Process_humanNumber(RichString* str, unsigned long long number, bool coloring) {
-   char buffer[10];
+void Process_printBytes(RichString* str, unsigned long long number, bool coloring) {
+   char buffer[16];
    int len;
 
    int largeNumberColor = coloring ? CRT_colors[LARGE_NUMBER] : CRT_colors[PROCESS];
@@ -62,7 +62,12 @@ void Process_humanNumber(RichString* str, unsigned long long number, bool colori
    if (number == ULLONG_MAX) {
       //Invalid number
       RichString_appendAscii(str, shadowColor, "  N/A ");
-   } else if (number < 1000) {
+      return;
+   }
+
+   number /= ONE_K;
+
+   if (number < 1000) {
       //Plain number, no markings
       len = xSnprintf(buffer, sizeof(buffer), "%5llu ", number);
       RichString_appendnAscii(str, processColor, buffer, len);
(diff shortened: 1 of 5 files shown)


Process.c
```diff
@@ -108,14 +113,43 @@ void Process_humanNumber(RichString* str, unsigned long long number, bool colori
...
+      len = xSnprintf(buffer, sizeof(buffer), "%4lluT ", number);
+      RichString_appendnAscii(str, largeNumberColor, buffer, len);
+   } else if (number < 10000ULL * ONE_G) {
+      //1 digit PB, 3 digit TB
+      number /= ONE_G;
+      len = xSnprintf(buffer, sizeof(buffer), "%1llu", number/1000);
+      RichString_appendnAscii(str, largeNumberColor, buffer, len);
+      number %= 1000;
+      len = xSnprintf(buffer, sizeof(buffer), "%03lluP ", number);
+      RichString_appendnAscii(str, largeNumberColor, buffer, len);
```
BenBE, line 137: Why bother splitting this print when printing both parts the same color regardless?

Process.c
```diff
@@ -211,32 +293,41 @@ static inline void Process_writeCommand(const Process* this, int attr, int basea
...
    } else if (rate < ONE_T) {
-      int len = snprintf(buffer, n, "%7.2f G/s ", rate / ONE_G);
+      int len = snprintf(buffer, sizeof(buffer), "%7.2f G/s ", rate / ONE_G);
+      RichString_appendnAscii(str, largeNumberColor, buffer, len);
+   } else if (rate < ONE_P) {
+      int len = snprintf(buffer, sizeof(buffer), "%7.2f T/s ", rate / ONE_T);
       RichString_appendnAscii(str, largeNumberColor, buffer, len);
    } else {
-      int len = snprintf(buffer, n, "%7.2f T/s ", rate / ONE_T);
+      int len = snprintf(buffer, sizeof(buffer), "%7.2f P/s ", rate / ONE_P);
```
BenBE, line 330: Actually it'd have to be `KB/s`, `MB/s`, `GB/s`, `TB/s`, and `PB/s`.
And as we're actually using binary prefixes that'd be even `KiB/s`, `MiB/s`, `GiB/s`, `TiB/s`, and `PiB/s`.

/PedanticMode

Process.c
cgzones: thanks, simplified

Process.c
cgzones: Do we want to spend two columns for a bit more readability?

BenBE: No.

*importantly points to `/PedanticMode`*

PR #599, opened 2021-04-14
Optimize RichString usage

cgzones:
* Use RichString_appendnAscii where possible

  `RichString_appendnAscii()` avoids a `strlen(3)` call over `RichString_appendAscii()`.
Use the former where the length is available from a previous checked `snprintf(3)` call.

 * Process: print default buffer in ascii

   `RichString_appendWide()` is more expensive than `RichString_appendAscii()` due to the calls to `mbstowcs(3)` and `iswprint(3)`.

   Use the latter to print the process field buffer by default.

   For the following fields this theoretically can corrupt the output:
    - TTY

* LinuxProcess: print default buffer in ascii

  `RichString_appendWide()` is more expensive than `RichString_appendAscii()` due to the calls to `mbstowcs(3)` and
`iswprint(3)`.

  Use the latter to print the process field buffer by default.

  For the following fields this theoretically can corrupt the output:
    - SECATTR
    - CGROUP
    - CTID

### DIFF
--- a/CPUMeter.c
+++ b/CPUMeter.c
@@ -101,51 +101,53 @@ static void CPUMeter_updateValues(Meter* this) {
 
 static void CPUMeter_display(const Object* cast, RichString* out) {
    char buffer[50];
+   int len;
    const Meter* this = (const Meter*)cast;
+
    if (this->param > this->pl->cpuCount) {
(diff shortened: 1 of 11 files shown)


BenBE (approved):
LGTM.

PR #603, opened 2021-04-17
Add NetBSD support in htop(1)

fraggerfox:
This implementation makes NetBSD use htop(1) without the need of [mount_procfs(8)](https://man.netbsd.org/mount_procfs.8).

The implementation has been copied over and modified from the OpenBSD implementation in htop(1).

Make NetBSD no longer [masquerade as Linux](http://cvsweb.netbsd.org/bsdweb.cgi/~checkout~/pkgsrc/sysutils/htop/patches/patch-configure.ac).

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -236,6 +236,30 @@ myhtopplatheaders = $(dragonflybsd_platform_headers)
 myhtopplatsources = $(dragonflybsd_platform_sources)
 endif
 
+# NetBSD
+# -------
+
+netbsd_platform_headers = \
+	netbsd/Platform.h \
+	netbsd/ProcessField.h \
+	netbsd/NetBSDProcessList.h \
+	netbsd/NetBSDProcess.h \
+	generic/hostname.h \
+	generic/uname.h
+
+netbsd_platform_sources = \
+	netbsd/Platform.c \
+	netbsd/NetBSDProcessList.c \
+	netbsd/NetBSDProcess.c \
+	generic/hostname.c \
+	generic/uname.c
+
+if HTOP_NETBSD
+AM_LDFLAGS += -lkvm
+myhtopplatheaders = $(netbsd_platform_headers)
+myhtopplatsources = $(netbsd_platform_sources)
+endif
+
 # OpenBSD
 # -------
 
--- a/Makefile.am
+++ b/Makefile.am
@@ -240,22 +240,24 @@ endif
 # -------
 
 netbsd_platform_headers = \
+	generic/gettime.h \
+	generic/hostname.h \
+	generic/uname.h \
 	netbsd/Platform.h \
 	netbsd/ProcessField.h \
-	netbsd/NetBSDProcessList.h \
 	netbsd/NetBSDProcess.h \
-	generic/hostname.h \
-	generic/uname.h
+	netbsd/NetBSDProcessList.h
 
 netbsd_platform_sources = \
-	netbsd/Platform.c \
-	netbsd/NetBSDProcessList.c \
-	netbsd/NetBSDProcess.c \
+	generic/gettime.c \
 	generic/hostname.c \
-	generic/uname.c
+	generic/uname.c \
+	netbsd/Platform.c \
+	netbsd/NetBSDProcess.c \
+	netbsd/NetBSDProcessList.c
 
 if HTOP_NETBSD
-AM_LDFLAGS += -lkvm
+myhtopplatprogram = htop
 myhtopplatheaders = $(netbsd_platform_headers)
 myhtopplatsources = $(netbsd_platform_sources)
 endif
--- a/netbsd/NetBSDProcess.c
+++ b/netbsd/NetBSDProcess.c
@@ -183,6 +183,12 @@ const ProcessFieldData Process_fields[LAST_PROCESSFIELD] = {
       .flags = 0,
       .pidColumn = true,
    },
+   [PROC_COMM] = {
+      .name = "COMM",
+      .title = "COMM            ",
+      .description = "comm string of the process",
+      .flags = 0,
+   },
 };
 
 Process* NetBSDProcess_new(const Settings* settings) {
@@ -235,7 +241,3 @@ const ProcessClass NetBSDProcess_class = {
    .writeField = NetBSDProcess_writeField,
    .compareByKey = NetBSDProcess_compareByKey
 };
-
-bool Process_isThread(const Process* this) {
-   return Process_isKernelThread(this) || Process_isUserlandThread(this);
-}
--- a/netbsd/NetBSDProcess.c
+++ b/netbsd/NetBSDProcess.c
@@ -1,0 +1,240 @@
+/*
+htop - NetBSDProcess.c
+(C) 2015 Hisham H. Muhammad
+(C) 2021 Santhosh Raju
+Released under the GNU GPLv2, see the COPYING file
+in the source distribution for its full text.
+*/
+
+#include "NetBSDProcess.h"
+
+#include <stdlib.h>
+
+#include "CRT.h"
+#include "Process.h"
+#include "RichString.h"
+#include "XUtils.h"
+
+
+const ProcessFieldData Process_fields[LAST_PROCESSFIELD] = {
+   [0] = {
+      .name = "",
+      .title = NULL,
+      .description = NULL,
+      .flags = 0,
+   },
+   [PID] = {
+      .name = "PID",
+      .title = "PID",
+      .description = "Process/thread ID",
+      .flags = 0,
+      .pidColumn = true,
+   },
+   [COMM] = {
+      .name = "Command",
+      .title = "Command ",
+      .description = "Command line",
+      .flags = 0,
+   },
+   [STATE] = {
+      .name = "STATE",
+      .title = "S ",
+      .description = "Process state (S sleeping, R running, D disk, Z zombie, T traced, W paging)",
+      .flags = 0,
+   },
+   [PPID] = {
+      .name = "PPID",
+      .title = "PPID",
+      .description = "Parent process ID",
+      .flags = 0,
+      .pidColumn = true,
+   },
+   [PGRP] = {
+      .name = "PGRP",
+      .title = "PGRP",
+      .description = "Process group ID",
+      .flags = 0,
+      .pidColumn = true,
+   },
+   [SESSION] = {
+      .name = "SESSION",
+      .title = "SESN",
+      .description = "Process's session ID",
+      .flags = 0,
+      .pidColumn = true,
+   },
+   [TTY] = {
+      .name = "TTY",
+      .title = "TTY      ",
+      .description = "Controlling terminal",
+      .flags = 0,
+   },
+   [TPGID] = {
+      .name = "TPGID",
+      .title = "TPGID",
+      .description = "Process ID of the fg process group of the controlling terminal",
+      .flags = 0,
+      .pidColumn = true,
+   },
+   [MINFLT] = {
+      .name = "MINFLT",
+      .title = "     MINFLT ",
+      .description = "Number of minor faults which have not required loading a memory page from disk",
+      .flags = 0,
+      .defaultSortDesc = true,
+   },
+   [MAJFLT] = {
+      .name = "MAJFLT",
+      .title = "     MAJFLT ",
+      .description = "Number of major faults which have required loading a memory page from disk",
+      .flags = 0,
+      .defaultSortDesc = true,
+   },
+   [PRIORITY] = {
+      .name = "PRIORITY",
+      .title = "PRI ",
+      .description = "Kernel's internal priority for the process",
+      .flags = 0,
+   },
+   [NICE] = {
+      .name = "NICE",
+      .title = " NI ",
+      .description = "Nice value (the higher the value, the more it lets other processes take priority)",
+      .flags = 0,
+   },
+   [STARTTIME] = {
+      .name = "STARTTIME",
+      .title = "START ",
+      .description = "Time the process was started",
+      .flags = 0,
+   },
+   [PROCESSOR] = {
+      .name = "PROCESSOR",
+      .title = "CPU ",
+      .description = "Id of the CPU the process last executed on",
+      .flags = 0,
+   },
+   [M_VIRT] = {
+      .name = "M_VIRT",
+      .title = " VIRT ",
+      .description = "Total program size in virtual memory",
+      .flags = 0,
+      .defaultSortDesc = true,
+   },
+   [M_RESIDENT] = {
+      .name = "M_RESIDENT",
+      .title = "  RES ",
+      .description = "Resident set size, size of the text and data sections, plus stack usage",
+      .flags = 0,
+      .defaultSortDesc = true,
+   },
+   [ST_UID] = {
+      .name = "ST_UID",
+      .title = "  UID ",
+      .description = "User ID of the process owner",
+      .flags = 0,
+   },
+   [PERCENT_CPU] = {
+      .name = "PERCENT_CPU",
+      .title = "CPU% ",
+      .description = "Percentage of the CPU time the process used in the last sampling",
+      .flags = 0,
+      .defaultSortDesc = true,
+   },
+   [PERCENT_NORM_CPU] = {
+      .name = "PERCENT_NORM_CPU",
+      .title = "NCPU%",
+      .description = "Normalized percentage of the CPU time the process used in the last sampling (normalized by cpu count)",
+      .flags = 0,
+      .defaultSortDesc = true,
+   },
+   [PERCENT_MEM] = {
+      .name = "PERCENT_MEM",
+      .title = "MEM% ",
+      .description = "Percentage of the memory the process is using, based on resident memory size",
+      .flags = 0,
+      .defaultSortDesc = true,
+   },
+   [USER] = {
+      .name = "USER",
+      .title = "USER      ",
+      .description = "Username of the process owner (or user ID if name cannot be determined)",
+      .flags = 0,
+   },
+   [TIME] = {
+      .name = "TIME",
+      .title = "  TIME+  ",
+      .description = "Total time the process has spent in user and system time",
+      .flags = 0,
+      .defaultSortDesc = true,
+   },
+   [NLWP] = {
+      .name = "NLWP",
+      .title = "NLWP ",
+      .description = "Number of threads in the process",
+      .flags = 0,
+   },
+   [TGID] = {
+      .name = "TGID",
+      .title = "TGID",
+      .description = "Thread group ID (i.e. process ID)",
+      .flags = 0,
+      .pidColumn = true,
+   },
+};
+
+Process* NetBSDProcess_new(const Settings* settings) {
+   NetBSDProcess* this = xCalloc(sizeof(NetBSDProcess), 1);
+   Object_setClass(this, Class(NetBSDProcess));
+   Process_init(&this->super, settings);
+   return &this->super;
+}
+
+void Process_delete(Object* cast) {
+   NetBSDProcess* this = (NetBSDProcess*) cast;
+   Process_done((Process*)cast);
+   free(this);
+}
+
+static void NetBSDProcess_writeField(const Process* this, RichString* str, ProcessField field) {
+   //const NetBSDProcess* op = (const NetBSDProcess*) this;
+   char buffer[256]; buffer[255] = '\0';
+   int attr = CRT_colors[DEFAULT_COLOR];
+   //int n = sizeof(buffer) - 1;
+   switch (field) {
+   // add NetBSD-specific fields here
+   default:
+      Process_writeField(this, str, field);
+      return;
+   }
+   RichString_appendWide(str, attr, buffer);
+}
+
+static int NetBSDProcess_compareByKey(const Process* v1, const Process* v2, ProcessField key) {
+   const NetBSDProcess* p1 = (const NetBSDProcess*)v1;
+   const NetBSDProcess* p2 = (const NetBSDProcess*)v2;
+
+   // remove if actually used
+   (void)p1; (void)p2;
+
+   switch (key) {
+   // add NetBSD-specific fields here
+   default:
+      return Process_compareByKey_Base(v1, v2, key);
+   }
+}
+
+const ProcessClass NetBSDProcess_class = {
+   .super = {
+      .extends = Class(Process),
+      .display = Process_display,
+      .delete = Process_delete,
+      .compare = Process_compare
+   },
+   .writeField = NetBSDProcess_writeField,
+   .compareByKey = NetBSDProcess_compareByKey
+};
+
+bool Process_isThread(const Process* this) {
+   return Process_isKernelThread(this) || Process_isUserlandThread(this);
+}
--- a/netbsd/NetBSDProcess.c
+++ b/netbsd/NetBSDProcess.c
@@ -195,6 +195,19 @@ const ProcessFieldData Process_fields[LAST_PROCESSFIELD] = {
       .description = "comm string of the process",
       .flags = 0,
    },
+   [PROC_EXE] = {
+      .name = "EXE",
+      .title = "EXE             ",
+      .description = "Basename of exe of the process",
+      .flags = 0,
+   },
+   [CWD] = {
+      .name = "CWD",
+      .title = "CWD                       ",
+      .description = "The current working directory of the process",
+      .flags = PROCESS_FLAG_CWD,
+   },
+
 };
 
 Process* NetBSDProcess_new(const Settings* settings) {
--- a/netbsd/NetBSDProcess.c
+++ b/netbsd/NetBSDProcess.c
@@ -8,7 +8,7 @@ Released under the GNU GPLv2, see the COPYING file
 in the source distribution for its full text.
 */
 
-#include "NetBSDProcess.h"
+#include "netbsd/NetBSDProcess.h"
 
 #include <stdlib.h>
 
--- a/netbsd/NetBSDProcess.c
+++ b/netbsd/NetBSDProcess.c
@@ -199,10 +199,9 @@ void Process_delete(Object* cast) {
 }
 
 static void NetBSDProcess_writeField(const Process* this, RichString* str, ProcessField field) {
-   //const NetBSDProcess* op = (const NetBSDProcess*) this;
    char buffer[256]; buffer[255] = '\0';
    int attr = CRT_colors[DEFAULT_COLOR];
-   //int n = sizeof(buffer) - 1;
+
    switch (field) {
    // add NetBSD-specific fields here
    default:
--- a/netbsd/NetBSDProcess.c
+++ b/netbsd/NetBSDProcess.c
@@ -1,7 +1,9 @@
 /*
 htop - NetBSDProcess.c
 (C) 2015 Hisham H. Muhammad
+(C) 2015 Michael McConville
 (C) 2021 Santhosh Raju
+(C) 2021 htop dev team
 Released under the GNU GPLv2, see the COPYING file
 in the source distribution for its full text.
 */
--- a/netbsd/NetBSDProcess.c
+++ b/netbsd/NetBSDProcess.c
@@ -110,6 +110,12 @@ const ProcessFieldData Process_fields[LAST_PROCESSFIELD] = {
       .description = "Time the process was started",
       .flags = 0,
    },
+   [ELAPSED] = {
+      .name = "ELAPSED",
+      .title = "ELAPSED  ",
+      .description = "Time since the process was started",
+      .flags = 0,
+   },
    [PROCESSOR] = {
       .name = "PROCESSOR",
       .title = "CPU ",
--- a/netbsd/NetBSDProcess.h
+++ b/netbsd/NetBSDProcess.h
@@ -21,10 +21,6 @@ typedef struct NetBSDProcess_ {
    Process super;
 } NetBSDProcess;
 
-#define Process_isKernelThread(_process) (_process->pgrp == 0)
-
-#define Process_isUserlandThread(_process) (_process->pid != _process->tgid)
-
 extern const ProcessClass NetBSDProcess_class;
 
 extern const ProcessFieldData Process_fields[LAST_PROCESSFIELD];
@@ -32,7 +28,5 @@ extern const ProcessFieldData Process_fields[LAST_PROCESSFIELD];
 Process* NetBSDProcess_new(const Settings* settings);
 
 void Process_delete(Object* cast);
-
-bool Process_isThread(const Process* this);
 
 #endif
--- a/netbsd/NetBSDProcess.h
+++ b/netbsd/NetBSDProcess.h
@@ -4,6 +4,8 @@
 htop - NetBSDProcess.h
 (C) 2015 Hisham H. Muhammad
 (C) 2015 Michael McConville
+(C) 2021 Santhosh Raju
+(C) 2021 htop dev team
 Released under the GNU GPLv2, see the COPYING file
 in the source distribution for its full text.
 */
--- a/netbsd/NetBSDProcess.h
+++ b/netbsd/NetBSDProcess.h
@@ -1,0 +1,36 @@
+#ifndef HEADER_NetBSDProcess
+#define HEADER_NetBSDProcess
+/*
+htop - NetBSDProcess.h
+(C) 2015 Hisham H. Muhammad
+(C) 2015 Michael McConville
+Released under the GNU GPLv2, see the COPYING file
+in the source distribution for its full text.
+*/
+
+#include <stdbool.h>
+
+#include "Object.h"
+#include "Process.h"
+#include "Settings.h"
+
+
+typedef struct NetBSDProcess_ {
+   Process super;
+} NetBSDProcess;
+
+#define Process_isKernelThread(_process) (_process->pgrp == 0)
+
+#define Process_isUserlandThread(_process) (_process->pid != _process->tgid)
+
+extern const ProcessClass NetBSDProcess_class;
+
+extern const ProcessFieldData Process_fields[LAST_PROCESSFIELD];
+
+Process* NetBSDProcess_new(const Settings* settings);
+
+void Process_delete(Object* cast);
+
+bool Process_isThread(const Process* this);
+
+#endif
--- a/netbsd/NetBSDProcessList.c
+++ b/netbsd/NetBSDProcessList.c
@@ -8,7 +8,7 @@ Released under the GNU GPLv2, see the COPYING file
 in the source distribution for its full text.
 */
 
-#include "NetBSDProcessList.h"
+#include "netbsd/NetBSDProcessList.h"
 
 #include <kvm.h>
 #include <limits.h>
@@ -27,11 +27,12 @@ in the source distribution for its full text.
 #include "CRT.h"
 #include "Macros.h"
 #include "Object.h"
-#include "NetBSDProcess.h"
+#include "netbsd/NetBSDProcess.h"
 #include "Process.h"
 #include "ProcessList.h"
 #include "Settings.h"
 #include "XUtils.h"
+
 
 static long fscale;
 static int pageSize;
--- a/netbsd/NetBSDProcessList.c
+++ b/netbsd/NetBSDProcessList.c
@@ -114,6 +114,44 @@ static void NetBSDProcessList_scanMemoryInfo(ProcessList* pl) {
    pl->usedSwap = uvmexp.swpginuse * pageSizeKB;
 }
 
+static void NetBSDProcessList_updateExe(const struct kinfo_proc2* kproc, Process* proc) {
+   const int mib[] = { CTL_KERN, KERN_PROC_ARGS, kproc->p_pid, KERN_PROC_PATHNAME };
+   char buffer[2048];
+   size_t size = sizeof(buffer);
+   if (sysctl(mib, 4, buffer, &size, NULL, 0) != 0) {
+      Process_updateExe(proc, NULL);
+      return;
+   }
+   printf("%s\n", buffer);
+   /* Kernel threads return an empty buffer */
+   if (buffer[0] == '\0') {
+      Process_updateExe(proc, NULL);
+      return;
+   }
+
+   Process_updateExe(proc, buffer);
+}
+
+static void NetBSDProcessList_updateCwd(const struct kinfo_proc2* kproc, Process* proc) {
+   const int mib[] = { CTL_KERN, KERN_PROC_ARGS, kproc->p_pid, KERN_PROC_CWD };
+   char buffer[2048];
+   size_t size = sizeof(buffer);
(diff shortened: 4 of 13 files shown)


netbsd/Platform.c
```diff
@@ -0,0 +1,324 @@
...
+      return;
+   }
+   *one     = (double) loadAverage.ldavg[0] / loadAverage.fscale;
+   *five    = (double) loadAverage.ldavg[1] / loadAverage.fscale;
+   *fifteen = (double) loadAverage.ldavg[2] / loadAverage.fscale;
+}
+
+int Platform_getMaxPid() {
+   // this is hard-coded in sys/sys/proc.h - no sysctl exists
+   return 30000;
```
niacat, line 195: 30000 is only the limit if there are <16K processes. Likely should be sizeof(int32_t), since that's the limit of the pid_t data type?

netbsd/Platform.c
```diff
@@ -0,0 +1,324 @@
...
+   env = xMalloc(capacity);
+   for (char** p = ptr; *p; p++) {
+      size_t len = strlen(*p) + 1;
+
+      if (size + len > capacity) {
+         capacity *= 2;
+         env = xRealloc(env, capacity);
+      }
+
+      strlcpy(env + size, *p, len);
```
BenBE, line 282: What about using `String_safeStrncpy` instead?

netbsd/NetBSDProcess.c
```diff
@@ -0,0 +1,239 @@
...
+   Process_done((Process*)cast);
+   free(this);
+}
+
+static void NetBSDProcess_writeField(const Process* this, RichString* str, ProcessField field) {
+   char buffer[256]; buffer[255] = '\0';
+   int attr = CRT_colors[DEFAULT_COLOR];
+
+   switch (field) {
+   // add NetBSD-specific fields here
```
BenBE, line 204: Are there really only platform-independent fields?

netbsd/NetBSDProcess.c
```diff
@@ -0,0 +1,239 @@
...
+      .delete = Process_delete,
+      .compare = Process_compare
+   },
+   .writeField = NetBSDProcess_writeField,
+   .compareByKey = NetBSDProcess_compareByKey
+};
+
+bool Process_isThread(const Process* this) {
+   return Process_isKernelThread(this) || Process_isUserlandThread(this);
+}
```
BenBE, line 239: Likely to be superseeded by #388 that changes thread handling to be performed globally. No longer used when #388 lands.

netbsd/NetBSDProcess.h
```diff
@@ -0,0 +1,36 @@
...
+#include "Settings.h"
+
+
+typedef struct NetBSDProcess_ {
+   Process super;
+} NetBSDProcess;
+
+#define Process_isKernelThread(_process) (_process->pgrp == 0)
+
+#define Process_isUserlandThread(_process) (_process->pid != _process->tgid)
```
BenBE, line 24: Cf. note on thread handling. These are also obsolete. Implementations should use the `isUserThread` and `isKernelThread` fields from the global `Process` structure instead. Cf. #388 for details.

netbsd/NetBSDProcess.h
```diff
@@ -0,0 +1,36 @@
...
+
+extern const ProcessClass NetBSDProcess_class;
+
+extern const ProcessFieldData Process_fields[LAST_PROCESSFIELD];
+
+Process* NetBSDProcess_new(const Settings* settings);
+
+void Process_delete(Object* cast);
+
+bool Process_isThread(const Process* this);
```
BenBE, line 34: dito.

netbsd/NetBSDProcess.c
```diff
@@ -0,0 +1,239 @@
+/*
+htop - NetBSDProcess.c
+(C) 2015 Hisham H. Muhammad
+(C) 2021 Santhosh Raju
+Released under the GNU GPLv2, see the COPYING file
+in the source distribution for its full text.
+*/
+
+#include "NetBSDProcess.h"
```
BenBE, line 9: Should include platform path (relative to main directory path) -> `netbsd/NetBSDProcess.h`

netbsd/NetBSDProcessList.c
```diff
@@ -0,0 +1,314 @@
+/*
+htop - NetBSDProcessList.c
+(C) 2014 Hisham H. Muhammad
+(C) 2021 Santhosh Raju
+Released under the GNU GPLv2, see the COPYING file
+in the source distribution for its full text.
+*/
+
+#include "NetBSDProcessList.h"
```
BenBE, line 9: File include path.

netbsd/NetBSDProcessList.c
```diff
@@ -0,0 +1,314 @@
...
+#include <sys/sched.h>
+#include <sys/swap.h>
+#include <sys/sysctl.h>
+#include <sys/types.h>
+#include <uvm/uvm_extern.h>
+
+#include "CRT.h"
+#include "Macros.h"
+#include "Object.h"
+#include "NetBSDProcess.h"
```
BenBE, line 28: dito.

netbsd/NetBSDProcessList.c
```diff
@@ -0,0 +1,314 @@
...
+#include "Process.h"
+#include "ProcessList.h"
+#include "Settings.h"
+#include "XUtils.h"
+
```
BenBE, line 33: 2 blank lines

netbsd/NetBSDProcessList.c
```diff
@@ -0,0 +1,314 @@
...
+   opl->kd = kvm_openfiles(NULL, NULL, NULL, KVM_NO_FILES, errbuf);
+   if (opl->kd == NULL) {
+      CRT_fatalError("kvm_openfiles() failed");
+   }
+
+   return pl;
+}
+
+void ProcessList_delete(ProcessList* this) {
+   NetBSDProcessList* opl = (NetBSDProcessList*) this;
```
BenBE, line 80: Consistently this would be called `npl` instead of `opl`.

netbsd/NetBSDProcessList.c
```diff
@@ -0,0 +1,314 @@
...
+static int pageSizeKB;
+
+ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* pidMatchList, uid_t userId) {
+   const int mib[] = { CTL_HW, HW_NCPU };
+   const int fmib[] = { CTL_KERN, KERN_FSCALE };
+   int r;
+   size_t size;
+   char errbuf[_POSIX2_LINE_MAX];
+
+   NetBSDProcessList* opl = xCalloc(1, sizeof(NetBSDProcessList));
```
BenBE, line 45: `npl` instead of `opl`

netbsd/NetBSDProcessList.c
```diff
@@ -0,0 +1,314 @@
...
+
+   for (int i = 0; i < CPUSTATES; i++) {
+      avg[i] /= this->super.cpuCount;
+   }
+
+   kernelCPUTimesToHtop(avg, this->cpus);
+}
+
+void ProcessList_goThroughEntries(ProcessList* super, bool pauseProcessUpdate) {
+   NetBSDProcessList* opl = (NetBSDProcessList*) super;
```
BenBE, line 303: `npl` instead of `opl`.

netbsd/Platform.c
```diff
@@ -0,0 +1,324 @@
+/*
+htop - netbsd/Platform.c
+(C) 2014 Hisham H. Muhammad
+(C) 2021 Santhosh Raju
+Released under the GNU GPLv2, see the COPYING file
+in the source distribution for its full text.
+*/
+
+#include "Platform.h"
```
BenBE, line 9: Include path relative to main directory.

netbsd/Platform.c
```diff
@@ -0,0 +1,324 @@
...
+#include "ClockMeter.h"
+#include "DateMeter.h"
+#include "DateTimeMeter.h"
+#include "HostnameMeter.h"
+#include "LoadAverageMeter.h"
+#include "Macros.h"
+#include "MemoryMeter.h"
+#include "Meter.h"
+#include "NetBSDProcess.h"
+#include "NetBSDProcessList.h"
```
BenBE, line 33: dito.

netbsd/Platform.c
```diff
@@ -0,0 +1,324 @@
...
+   *fifteen = (double) loadAverage.ldavg[2] / loadAverage.fscale;
+}
+
+int Platform_getMaxPid() {
+   // this is hard-coded in sys/sys/proc.h - no sysctl exists
+   return 30000;
+}
+
+double Platform_setCPUValues(Meter* this, int cpu) {
+   const NetBSDProcessList* pl = (const NetBSDProcessList*) this->pl;
```
BenBE, line 199: `npl` instead of just `pl`.

netbsd/NetBSDProcessList.c
```diff
@@ -0,0 +1,314 @@
...
+   int count = 0;
+   int nlwps = 0;
+
+   const struct kinfo_proc2* kprocs = kvm_getproc2(this->kd, KERN_PROC_ALL, 0, sizeof(struct kinfo_proc2), &count);
+
+   for (int i = 0; i < count; i++) {
+      const struct kinfo_proc2* kproc = &kprocs[i];
+
+      bool preExisting = false;
+      Process* proc = ProcessList_getProcess(&this->super, kproc->p_pid, &preExisting, NetBSDProcess_new);
```
BenBE, line 174: Several fields in the global `Process` structure will be renamed following #388. This mostly affects the `cmdline`, `procComm` (Custom Thread/command name) and `procExe` (main executable file reported by the kernel). When any of these fields is updated a call to `Process_makeCommandStr` needs to be done to update some internal process data. Also note that the thread handling is now handled in the global `Process` structure via the `isUserlandThread` and `isKernelThread` fields. Implementations should set those fields accordingly (`isUserlandThread` should be `false` for the main process)

BenBE (changes requested):
Thank you very much for your implementation to support NetBSD natively for htop.

As there's one bigger change upcoming that affects all current platforms (and thus also all upcoming ones, I took the liberty to remark on the locations where this potentially requires changes to your PR.

netbsd/Platform.c
fraggerfox: It seem you are right, I shall look into this and get back.

fraggerfox:
@BenBE Thank you for going through the code and giving a detailed feedback, let me go through each of them and try to resolve it.

BenBE:
@fraggerfox Another point that came up in discussions in the team was the issue of long-term support of the platform. Would you be willing to help mid-/long-term to keep the NetBSD port up-to-date and aid in debugging issues specific to this platform? Or do you know someone who would be willing to take up on this task?

fraggerfox:
> 
> 
> @fraggerfox Another point that came up in discussions in the team was the issue of long-term support of the platform. Would you be willing to help mid-/long-term to keep the NetBSD port up-to-date and aid in debugging issues specific to this platform? Or do you know someone who would be willing to take up on this task?

Yes, I do not mind putting in my time mid-/long-term to support the NetBSD port. Hopefully, I should have more time to commit to maintenance in the coming months. In case of issues / interruptions I can notify the team ahead of time.

niacat:
In general I think there's lots of interest in this, and I'm also able to help. We're also in `#netbsd-code` on freenode and can answer questions about APIs and offer testing.

netbsd/NetBSDProcess.c
fraggerfox: To the best of my knowledge yes, NetBSD does not have any platform-dependent fields like how FreeBSD has Jails and Jail ID (jid).

netbsd/Platform.c
fraggerfox: This should be fixed in f789aa5

netbsd/Platform.c
fraggerfox: Should be fixed in ed2f0dd

netbsd/NetBSDProcess.c
fraggerfox: Should be fixed in f20c703

netbsd/NetBSDProcessList.c
fraggerfox: Should be fixed in f20c703

netbsd/NetBSDProcessList.c
fraggerfox: Should be fixed in f20c703

netbsd/NetBSDProcessList.c
fraggerfox: Should be fixed in f20c703

netbsd/NetBSDProcessList.c
fraggerfox: Should be fixed in 50ae2bd

netbsd/NetBSDProcessList.c
fraggerfox: Should be fixed in 50ae2bd

netbsd/NetBSDProcessList.c
fraggerfox: Should be fixed in 50ae2bd

netbsd/Platform.c
fraggerfox: Should be fixed in d52c724

netbsd/Platform.c
fraggerfox: Should be fixed in d52c724

netbsd/Platform.c
fraggerfox: Should be fixed in d52c724

netbsd/Platform.c
```diff
@@ -191,13 +191,14 @@ void Platform_getLoadAverage(double* one, double* five, double* fifteen) {
...
+   // https://nxr.netbsd.org/xref/src/sys/sys/ansi.h#__pid_t
+   // pid is assigned as a 32bit Integer.
+   return INT32_MAX;
 }
 
 double Platform_setCPUValues(Meter* this, int cpu) {
-   const NetBSDProcessList* pl = (const NetBSDProcessList*) this->pl;
-   const CPUData* cpuData = &(pl->cpus[cpu]);
+   const NetBSDProcessList* npl = (const NetBSDProcessList*) this->pl;
+   const CPUData* cpuData = &(npl->cpus[cpu]);
```
BenBE, line 201: No need for those extra parens.

netbsd/NetBSDProcessList.c
```diff
@@ -0,0 +1,315 @@
...
+         this->super.runningTasks++;
+      }
+      proc->updated = true;
+   }
+}
+
+static unsigned long long saturatingSub(unsigned long long a, unsigned long long b) {
+   return a > b ? a - b : 0;
+}
+
```
BenBE, line 246: If this is used in multiple places, we may consider to extract this into `XUtils.[ch]` for generic re-use. This function may be of interest for other platforms too. Please extract as separate commit and update other platforms that may profit from this.

BenBE (commented):
Changes LGTM.

netbsd/Platform.c
fraggerfox: Should be fixed in b7c3b51

netbsd/NetBSDProcessList.c
fraggerfox: This is also used by OpenBSD, I guess I should coordinate with the person who is responsible for the OpenBSD parts to do the refactor. Should I do it as a part of another ticket?

BenBE: If you do it as a separate commit that only refactors that aspect I'd even accept it as part of this series. Refactoring changes just shouldn't be mixed in with other stuff in the same commit. If the changes become too large we can still split it off into a separate PR.

fraggerfox: Understood, let me look into this then. Thank you for the advice.

Makefile.am
```diff
@@ -241,22 +241,24 @@ endif
...
+	generic/gettime.h \
+	generic/hostname.h \
+	generic/uname.h \
 	netbsd/Platform.h \
 	netbsd/ProcessField.h \
 	netbsd/NetBSDProcessList.h \
-	netbsd/NetBSDProcess.h \
-	generic/hostname.h \
-	generic/uname.h
+	netbsd/NetBSDProcess.h
```
BenBE, line 250: Should sort before `NetBSDProcessList.h`

Makefile.am
```diff
@@ -241,22 +241,24 @@ endif
...
 netbsd_platform_sources = \
+	generic/gettime.c \
+	generic/hostname.c \
+	generic/uname.c \
 	netbsd/Platform.c \
 	netbsd/NetBSDProcessList.c \
-	netbsd/NetBSDProcess.c \
-	generic/hostname.c \
-	generic/uname.c
+	netbsd/NetBSDProcess.c
```
BenBE, line 258: dito.

netbsd/Platform.h
```diff
@@ -70,4 +71,20 @@ static inline void Platform_getRelease(char** string) {
    *string = Generic_uname();
 }
 
+#define PLATFORM_LONG_OPTIONS
```
BenBE, line 74: Should maybe include a note there are none.

netbsd/NetBSDProcessList.c
fraggerfox: Should be refactored in d43cd5c

Makefile.am
fraggerfox: Should be fixed in 0fd67f9

netbsd/Platform.h
fraggerfox: Should be fixed in 0fd67f9

Makefile.am
fraggerfox: Should be fixed in 0fd67f9

netbsd/Platform.h
fraggerfox: Added a note in 4331107

XUtils.h
```diff
@@ -76,4 +76,7 @@ char* xStrndup(const char* str, size_t len) ATTR_NONNULL ATTR_MALLOC;
 ssize_t xReadfile(const char* pathname, void* buffer, size_t count);
 ssize_t xReadfileat(openat_arg_t dirfd, const char* pathname, void* buffer, size_t count);
 
+/* This subtraction is used by NetBSD / OpenBSD for calculation of CPU usage items. */
+unsigned long long saturatingSub(unsigned long long a, unsigned long long b);
```
cgzones, line 80: Since this is such a trivial function, it's probably worth defining it `inline`.
Another option might be to implement it as macro in `Macros.h`, to be type generic.
Thoughts @BenBE ?

p.s.: it's used also at https://github.com/htop-dev/htop/blob/323d7e73aa1cc13b3a0870530e8fbd27102259ec/linux/LinuxProcessList.c#L1809

netbsd/NetBSDProcessList.c
```diff
@@ -0,0 +1,311 @@
...
+      proc->priority = kproc->p_priority - PZERO;
+
+      struct kinfo_lwp* klwps = kvm_getlwps(this->kd, kproc->p_pid, kproc->p_paddr, sizeof(struct kinfo_lwp), &nlwps);
+
+      switch (kproc->p_realstat) {
+      case SIDL:     proc->state = 'I'; break;
+      case SACTIVE:
+	// We only consider the first LWP with a one of the below states.
+        for (int j = 0; j < nlwps; j++) {
+          if (klwps) {
```
cgzones, line 215: in case `klwps` is `NULL`, `proc->state` is not being set

XUtils.h
BenBE: I think we should prefer a type-safe implementation over the macro one, especially as this automagically does CSE and avoids side-effects due to re-evaluation of arguments as the macro variant would do. The only ever implementations we probably need to have are for `int32_t`, `int64_t` and `double`. And while `XUtils.[ch]` primarily focuses on memory/string stuff, I wonder if the fit into `Macros.h` would be much better. Nah, well, make it `static inline` and put it in `Macros.h` ;-)

fraggerfox: Do you want me to shift the function definition to `Macros.h` and make it `static inline`?

netbsd/NetBSDProcessList.c
fraggerfox: Should be fixed in 896dea7

netbsd/NetBSDProcessList.c
```diff
@@ -222,6 +222,9 @@ static void NetBSDProcessList_scanProcs(NetBSDProcessList* this) {
             }
             if (proc->state != '?')
             break;
+	  } else {
```
BenBE, line 225: Indentation seems broken

BenBE (changes requested):

netbsd/NetBSDProcessList.c
fraggerfox: Seems that entire section had issues with indentation, should be fixed in ee27a7e

XUtils.h
fraggerfox: Should be done in c68bab1

netbsd/NetBSDProcess.h
```diff
@@ -3,7 +3,7 @@
 /*
 htop - NetBSDProcess.h
 (C) 2015 Hisham H. Muhammad
-(C) 2015 Michael McConville
+(C) 2021 Santhosh Raju
```
BenBE: Can you explain this change of the copyright header?

fraggerfox: That was the author of the OpenBSD's `OpenBSDProcess.h`, do you want me to retain the original author of the file?

BenBE: I'd prefer that. Also due to the many addtions you may also add the `htop dev team`.

fraggerfox: Sure, I shall update the headers.

fraggerfox: This should be done in ababf2e

netbsd/NetBSDProcess.c
fraggerfox: Should be updated in 237e66c

netbsd/NetBSDProcess.h
fraggerfox: Should be updated in 237e66c

netbsd/NetBSDProcess.h
fraggerfox: Should be updated in 237e66c

netbsd/NetBSDProcessList.c
fraggerfox: Should be updated in 237e66c

netbsd/NetBSDProcessList.c
```diff
@@ -0,0 +1,328 @@
...
+      case SZOMB:    proc->state = 'Z'; break;
+      case SDEAD:    proc->state = 'D'; break;
+      default:       proc->state = '?';
+      }
+
+      this->super.totalTasks++;
+      // SRUN ('R') means runnable, not running
+      if (proc->state == 'P') {
+         this->super.runningTasks++;
+      }
```
BenBE, line 255: This likely also should count the running kernel threads here.

netbsd/Platform.c
```diff
@@ -0,0 +1,327 @@
...
+   }
+
+   totalPercent = CLAMP(totalPercent, 0.0, 100.0);
+
+   v[CPU_METER_TEMPERATURE] = NAN;
+
+   return totalPercent;
+}
+
+void Platform_setMemoryValues(Meter* this) {
```
BenBE, line 234: Looking at the help screen this normally should be 4 values: `used`/`buffers`/`shared`/`cached`. Could you please re-check the memory meter really just want's those three values / could you add the `shared` value? Note: `shared` includes stuff like tmpfs on Linux for usability reasons, that are normally accounted as `cached`.

netbsd/Platform.h
```diff
@@ -0,0 +1,94 @@
...
+
+FileLocks_ProcessData* Platform_getProcessLocks(pid_t pid);
+
+bool Platform_getDiskIO(DiskIOData* data);
+
+bool Platform_getNetworkIO(NetworkIOData* data);
+
+void Platform_getBattery(double* percent, ACPresence* isOnAC);
+
+static inline void Platform_getHostname(char* buffer, size_t size) {
```
BenBE, line 72: Missing `#include <stddef.h>` for `size_t`.

netbsd/Platform.h
```diff
@@ -0,0 +1,94 @@
...
+   *string = Generic_uname();
+}
+
+static inline void Platform_longOptionsUsage(ATTR_UNUSED const char* name) { }
+
+static inline bool Platform_getLongOption(ATTR_UNUSED int opt, ATTR_UNUSED int argc, ATTR_UNUSED char** argv) {
+   return false;
+}
+
+static inline void Platform_gettime_realtime(struct timeval* tv, uint64_t* msec) {
```
BenBE, line 86: Missing `#include <stdint.h>` for `uint64_t`.
Also include for `struct timeval` missing.

netbsd/README.md
```diff
@@ -0,0 +1,34 @@
...
+* Memory being split into used/buffers/cache
+* Basic features of htop(1) like kill / nice / sort / search works as expected
+* If the meter displays are working as expected
+* If the process states are displayed correctly
+* If Thread information and count are displayed correctly
+
+TODO
+---
+
+* Get the implementation reviewed.
```
BenBE, line 34: I think at least this section can be dropped.

For the other parts of this file I think those are better of in the PR discussion so they are more visible. If necessary additional issues for specific tasks should be created for those instead.

BenBE (changes requested):
Some details still need refinement.

Also may benefit to include `ELAPSED` column from #627.

Furthermore `PROC_EXE` (via `KERN_PROC_PATHNAME`) and `CWD` (via `KERN_PROC_CWD`) might be nice to have.

netbsd/Platform.c
```diff
@@ -29,8 +29,8 @@ in the source distribution for its full text.
 #include "Macros.h"
 #include "MemoryMeter.h"
 #include "Meter.h"
-#include "NetBSDProcess.h"
-#include "NetBSDProcessList.h"
+#include "netbsd/NetBSDProcess.h"
+#include "netbsd/NetBSDProcessList.h"
```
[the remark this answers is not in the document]
fraggerfox: Should be fixed in 3cc9561

netbsd/NetBSDProcessList.c
```diff
@@ -0,0 +1,328 @@
...
+#include <sys/sched.h>
+#include <sys/swap.h>
+#include <sys/sysctl.h>
+#include <sys/types.h>
+#include <uvm/uvm_extern.h>
+
+#include "CRT.h"
+#include "Macros.h"
+#include "Object.h"
+#include "netbsd/NetBSDProcess.h"
```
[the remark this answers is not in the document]
fraggerfox: Should be fixed in 3cc9561

netbsd/NetBSDProcessList.h
```diff
@@ -0,0 +1,57 @@
...
+   unsigned long long int intrPeriod;
+   unsigned long long int idlePeriod;
+} CPUData;
+
+typedef struct NetBSDProcessList_ {
+   ProcessList super;
+   kvm_t* kd;
+
+   CPUData* cpus;
+
```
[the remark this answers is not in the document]
fraggerfox: The trailing blank space should be fixed in 3cc9561

Note: The other couple of BSDs also tend to follow this pattern, which is why I put it here.

netbsd/Platform.h
fraggerfox: Should be fixed in 3cc9561

netbsd/Platform.h
fraggerfox: Should be fixed in 3cc9561

netbsd/README.md
fraggerfox: Should be fixed in adf13a2

netbsd/NetBSDProcess.c
```diff
@@ -0,0 +1,243 @@
...
+      .flags = 0,
+      .pidColumn = true,
+   },
+   [PROC_COMM] = {
+      .name = "COMM",
+      .title = "COMM            ",
+      .description = "comm string of the process",
+      .flags = 0,
+   },
+};
```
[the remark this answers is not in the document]
fraggerfox: I will need to dig a bit deeper into this, I shall get back once I have more time to look into this.

netbsd/Platform.c
fraggerfox: This was one of the areas I spent quite sometime trying to figure out how things work, the current implementation is based on the values exposed by [uvm(9)'s sysctl interface](https://man.netbsd.org/uvm.9#SYSCTL), uvm is the virtual memory system of NetBSD. I am not sure how `shared` would be defined in this situation, I shall ask around a bit in the NetBSD developers circle and get back to you on this.

netbsd/NetBSDProcessList.c
fraggerfox: We set the `proc->state` to `'P'` based on the status of the running thread, this is done a few lines above. Does that answer your question about counting the running kernel threads?

XUtils.h
BenBE: Please also consolidate with the location/macro that @cgzones mentioned, if not yet done. Else feel free to mark this one here as resolved.

netbsd/NetBSDProcessList.c
BenBE: I'm missing a reference to `this->super.this->super.runningTasks++;` for all processes marked as kernel threads (helper threads of the kernel, not necessarily running).

netbsd/Platform.c
BenBE: TIA. Otherwise assume `0` for now.

XUtils.h
fraggerfox: You mean the one for `LinuxProcessList.c` too?

BenBE: Yes.

fraggerfox: This should be done in 1e279c6

netbsd/NetBSDProcess.c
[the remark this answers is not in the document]
fraggerfox: Even though I added support for `PROC_EXE` and `CWD` using the current syscall parameters it shows junk values in the columns need to figure out what is happening here.

This is done in commit a107535

netbsd/NetBSDProcessList.c
```diff
@@ -203,13 +241,18 @@ static void NetBSDProcessList_scanProcs(NetBSDProcessList* this) {
          proc->user = UsersTable_getRef(this->super.usersTable, proc->st_uid);
          ProcessList_add(&this->super, proc);
 
+	 NetBSDProcessList_updateExe(kproc, proc);
```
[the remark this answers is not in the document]
fraggerfox: Should be done in 88133cd

netbsd/NetBSDProcess.c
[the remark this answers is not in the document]
niacat: Looks like the order in the mib is wrong. It should be:

`CTL_KERN, KERN_PROC_ARGS, pid, KERN_PROC_PATHNAME`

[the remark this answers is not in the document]
fraggerfox: Doh! Let me try that.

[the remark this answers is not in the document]
fraggerfox: Thank you @niacat , should be fixed in cd9005a

netbsd/NetBSDProcessList.c
fraggerfox: I looked up FreeBSD / OpenBSD code and added a section above this to count the Kernel / User threads. Let me know if that works.

Should be available in 68109ea

netbsd/Platform.c
fraggerfox: I looked up the other BSD implementation and followed the same over here now. They do not set the `shared` value. 

Should be done in a3140cd

niacat:
@fraggerfox 

if you're interested in adding support for the "network stats" or disk I/O feature, maybe take a look at sysstat(1):

https://github.com/NetBSD/src/blob/trunk/usr.bin/systat/ifstat.c

however, for other sysstat features, i'd be careful to avoid introducing a dependency on /dev/kmem.

[envsys(4)](https://man.netbsd.org/envsys.4) and proplib can be used to implement the battery feature.

of course, i'm sure this can all be done at a later date - i can look into it after this PR is done, if you like.

BenBE:
> @fraggerfox
> 
> if you're interested in adding support for the "network stats" or disk I/O feature, maybe take a look at sysstat(1):
> 
> https://github.com/NetBSD/src/blob/trunk/usr.bin/systat/ifstat.c
> 
> however, for other sysstat features,

I think this can be done in a later PR. No need to rush things here …

> i'd be careful to avoid introducing a dependency on /dev/kmem. since actually using libkvm to read kernel memory is harshly frowned upon these days, i'd also check for a function other than kvm_open.

Can you elaborate? Because as I've seen it with other BSDs there is some common dependency on libkvm for most things that htop needs to gather.

What are you referring to regarding `kvm_open`? In `configure.ac` that's only to check if the lib is present and what paths are needed to use it. No need to check for much else.

> [envsys(4)](https://man.netbsd.org/envsys.4) and proplib can be used to implement the battery feature.

Can/should be done in a later PR.

> of course, i'm sure this can all be done at a later date - i can look into it after this PR is done, if you like.

Sure.

fraggerfox:
Yup I am interested @niacat , waiting for this PR to go in, then fine tune and add more features. Will most probably need your help with it.

niacat:
> Can you elaborate? Because as I've seen it with other BSDs there is some common dependency on libkvm for most things that htop needs to gather.

Unfortunately it's confusing, because libkvm is used for two purposes:

- reading directly from kernel memory through `kvm_open()` (unsafe)
- providing a friendly interface to sysctl through functions like `kvm_getprocs()` (safe)

This is mostly a historical wart, because these functions _used_ to rely on direct kernel memory access, but were updated to use the safer sysctl interface and not moved to another library.

There's a few bits of functionality that sysstat needs direct memory access to accomplish, but upstream in NetBSD we're moving everything to use sysctl-backed interfaces.

BenBE:
@fraggerfox Before doing the review, do you mind squashing in the most recent fixes into the commits they belong to? Those changes LGTM.

Also: Are there any open issues from your side that you think should be addressed as part of this PR?

fraggerfox:
@BenBE Let me squish the `CWD` and `PROC_EXE` commits into that single one.

At the moment no, I think most of the issues have been addressed. Will need to look into #656 once this goes in.

BenBE:
@fraggerfox Did some commit mashing ;-) No functional changes.

fraggerfox:
> 
> 
> @fraggerfox Did some commit mashing ;-) No functional changes.

Looks good.

BenBE (approved):
Source-wise it LGTM. Haven't tested to run this though.

niacat:
fraggerfox asked me to test his code, but unfortunately htop fails to configure with NetBSD libcurses.

~~~
checking for ncursesw/curses.h... no
checking for ncurses/ncurses.h... no
checking for ncurses/curses.h... no
checking for ncurses.h... no
configure: error: can not find required ncurses header file
~~~

this should be unnecessary, as NetBSD libcurses should be good enough.

so, I re-ran configure with `export CFLAGS="-I/usr/pkg/include"` and `export LDFLAGS="-L/usr/pkg/lib -Wl,-R/usr/pkg/lib"` so it would find a copy of ncurses I installed.

after this it appears to work fine, although the output is still less useful than `top(1)` - it groups the file cache together with active memory used by applications, and it does not show kernel threads,

BenBE:
> @fraggerfox asked me to test his code, but unfortunately `htop` fails to configure with NetBSD `libcurses`.
> 
> ```
> checking for ncursesw/curses.h... no
> checking for ncurses/ncurses.h... no
> checking for ncurses/curses.h... no
> checking for ncurses.h... no
> configure: error: can not find required ncurses header file
> ```
> 
> this should be unnecessary, as NetBSD libcurses should be good enough.
> 
> so, I re-ran configure with `export CFLAGS="-I/usr/pkg/include"` and `export LDFLAGS="-L/usr/pkg/lib -Wl,-R/usr/pkg/lib"` so it would find a copy of ncurses I installed.

Looking through `configure.ac` we only ever look for the ncurses headers, but not for any headers only called `curses.h`. Can you have a shot at updating `configure.ac` to also look for `curses.h` as a last resort? Also have a look at `ProvideCurses.h` to see if this needs any updates. That's the only place in `htop` that directly includes any of the `ncurses`/`curses` headers.

> after this it appears to work fine, although the output is still less useful than `top(1)` - it groups the file cache together with active memory used by applications,

This looks like when reading the memory stats some details are still missing. Actually there's a second open point (regarding shared memory) … @niacat Could you have a look at that?

> and it does not show kernel threads. It is definitely not a regression from previous htop on netbsd, though

Can you see if the check for kernel threads in `netbsd/NetBSDProcessList.c` ([this line](https://github.com/fraggerfox/htop-dev/blob/a176972c897a2d528a88537a635372d8a6ebc53f/netbsd/NetBSDProcessList.c#L237)) is done properly?

fraggerfox:
> 
> 
> fraggerfox asked me to test his code, but unfortunately htop fails to configure with NetBSD libcurses.
> 
> ```
> checking for ncursesw/curses.h... no
> checking for ncurses/ncurses.h... no
> checking for ncurses/curses.h... no
> checking for ncurses.h... no
> configure: error: can not find required ncurses header file
> ```
> 
> this should be unnecessary, as NetBSD libcurses should be good enough.
> 
> so, I re-ran configure with `export CFLAGS="-I/usr/pkg/include"` and `export LDFLAGS="-L/usr/pkg/lib -Wl,-R/usr/pkg/lib"` so it would find a copy of ncurses I installed.
> 

When running the build via the `pkgsrc` infrastructure these flags are passed in, I guess this is because of `USE_NCURSES=    yes` . When I was building it outside `pkgsrc` I passed in the flags manually as you have described above.
 
> after this it appears to work fine, although the output is still less useful than `top(1)` - it groups the file cache together with active memory used by applications, and it does not show kernel threads. it is definitely not a regression from previous htop on netbsd, though

Regarding the memory usage I mostly referred to `top(1)`'s code and [uvm((9)' sysctl interface](https://man.netbsd.org/uvm.9#SYSCTL) to get the [necessary values](https://github.com/htop-dev/htop/pull/603/files#diff-5a8ba703fe6150b28ee7dddfb0ec82795560f6c9e5aaa93cbab03dae90ba7119R104).

For the kernel threads, yes I agree, it pretty much does what the current `htop(1)` does in NetBSD, is there a better way to process this information?

alarixbsd:
We should probably get this merged and improve things further later.

In particular getting it to build with X/Open Curses seems difficult - it is peeking into ncurses-specific structs.

BenBE:
Could you open issues for the remaining items on your list? What I see is:

* Kernel thread display
* Active / file cache memory display
* X/Open Curses support (#660)

fraggerfox:
I opened a couple more tickets for the issues raised.

* Kernel thread display - #665 
* Active / file cache memory display - #666

PR #604, opened 2021-04-17
Darwin: obtain CPU data using Intel Power Gadget

danchr:
The [Intel Power Gadget](https://software.intel.com/content/www/us/en/develop/articles/intel-power-gadget.html) is a proprietary framework and driver from Intel, which allows user-space processes to query the CPU.  This PR adds support for using it for obtaining CPU frequencies and temperatures using it. Since the library might not be available at runtime, I've made the build system link against it weakly.

I've never contributed to `htop` before, and I'm not sure what your standards are, but I thought I'd post it here.

### DIFF
--- a/DisplayOptionsPanel.c
+++ b/DisplayOptionsPanel.c
@@ -117,11 +117,17 @@ DisplayOptionsPanel* DisplayOptionsPanel_new(Settings* settings, ScreenManager* 
    Panel_add(super, (Object*) CheckItem_newByRef("Update process names on every refresh", &(settings->updateProcessNames)));
    Panel_add(super, (Object*) CheckItem_newByRef("Add guest time in CPU meter percentage", &(settings->accountGuestInCPUMeter)));
    Panel_add(super, (Object*) CheckItem_newByRef("Also show CPU percentage numerically", &(settings->showCPUUsage)));
+   #if !defined(HTOP_DARWIN)
    Panel_add(super, (Object*) CheckItem_newByRef("Also show CPU frequency", &(settings->showCPUFrequency)));
+   #elif defined(HAVE_POWER_GADGET)
+   Panel_add(super, (Object*) CheckItem_newByRef("Also show CPU frequency (requires Intel Power Gadget)", &(settings->showCPUFrequency)));
+   #endif
    #ifdef BUILD_WITH_CPU_TEMP
    Panel_add(super, (Object*) CheckItem_newByRef(
    #if defined(HTOP_LINUX)
                                                  "Also show CPU temperature (requires libsensors)",
+   #elif defined(HTOP_DARWIN) && defined(HAVE_POWER_GADGET)
+                                                 "Also show CPU temperature (requires Intel Power Gadget)",
    #elif defined(HTOP_FREEBSD)
                                                  "Also show CPU temperature",
    #else
--- a/configure.ac
+++ b/configure.ac
@@ -299,6 +299,33 @@ if test "$enable_static" = yes; then
    AC_SEARCH_LIBS([Gpm_GetEvent], [gpm])
 fi
 
+AC_ARG_ENABLE([power-gadget],
+              [AS_HELP_STRING([--enable-power-gadget],
+                              [enable Intel Power Gadget support for obtaining CPU frequencies @<:@default=no@:>@])],
+              [],
+              [enable_power_gadget=no])
+case "$enable_power_gadget" in
+   no)
+      ;;
+   yes)
+      if test "$my_htop_platform" != darwin; then
+         AC_MSG_ERROR([the Intel Power Gadget is only supported on macOS])
+      fi
+
+      AC_COMPILE_IFELSE([
+         AC_LANG_PROGRAM(
+         [[#include <IntelPowerGadget/PowerGadgetLib.h>]],
+         [[PG_Initialize();]])],
+         [
+         AC_DEFINE([HAVE_POWER_GADGET], [1], [Using Intel Power Gadget for CPU frequencies])
+         LIBS="$LIBS -weak_framework IntelPowerGadget"
+         ],
+         [AC_MSG_ERROR([can not use IntelPowerGadget])], [-framework IntelPowerGadget])
+      ;;
+   *)
+      AC_MSG_ERROR([bad value '$enable_power_gadget' for --enable-power-gadget])
+      ;;
+esac
 
 AC_ARG_ENABLE([hwloc],
               [AS_HELP_STRING([--enable-hwloc],
@@ -509,7 +536,7 @@ case "$enable_sensors" in
       AC_MSG_ERROR([bad value '$enable_sensors' for --enable-sensors])
       ;;
 esac
-if test "$enable_sensors" = yes || test "$my_htop_platform" = freebsd; then
+if test "$enable_sensors" = yes || test "$my_htop_platform" = freebsd || test "$enable_power_gadget" = yes; then
    AC_DEFINE([BUILD_WITH_CPU_TEMP], [1], [Define if CPU temperature option should be enabled.])
 fi
 
--- a/darwin/DarwinProcessList.c
+++ b/darwin/DarwinProcessList.c
@@ -9,6 +9,7 @@ in the source distribution for its full text.
 
 #include <errno.h>
 #include <libproc.h>
+#include <math.h>
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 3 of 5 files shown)


DisplayOptionsPanel.c
```diff
@@ -122,6 +122,8 @@ DisplayOptionsPanel* DisplayOptionsPanel_new(Settings* settings, ScreenManager*
    Panel_add(super, (Object*) CheckItem_newByRef(
    #if defined(HTOP_LINUX)
                                                  "Also show CPU temperature (requires libsensors)",
+   #elif defined(HAVE_POWER_GADGET)
```
BenBE, line 125:
```suggestion
   #elif defined(HTOP_DARWIN) && defined(HAVE_POWER_GADGET)
```

configure.ac
```diff
@@ -299,6 +299,33 @@ if test "$enable_static" = yes; then
...
+        [
+        AC_DEFINE([HAVE_POWER_GADGET], [1], [Using Intel Power Gadget for CPU frequencies])
+        LIBS="$LIBS -weak_framework IntelPowerGadget"
+        ],
+        [AC_MSG_ERROR([can not use IntelPowerGadget])], [-framework IntelPowerGadget])
+      ;;
+   *)
+      AC_MSG_ERROR([bad value '$enable_power_gadget' for --enable-power-gadget])
+      ;;
+esac
```
BenBE, line 328: Check indentation. Should be 3 spaces per level.

darwin/DarwinProcessList.c
```diff
@@ -153,9 +226,26 @@ ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* pidMatchList, ui
...
+   }
+
+   free(this->cpu_freqs);
+   free(this->cpu_temps);
+#endif
+
+   ProcessList_done(super);
+
    free(this);
+
```
BenBE, line 248: No need for this blank line.

darwin/DarwinProcessList.c
```diff
@@ -128,13 +133,81 @@ static struct kinfo_proc* ProcessList_getKInfoProcs(size_t* count) {
...
+                  }
+
+                  PGSample_Release(this->cpu_sample);
+               }
+
+               this->cpu_sample = new_sample;
+            }
+         }
+      }
+   }
```
BenBE, line 191: Try to reduce the nesting by e.g. inverting conditions and possibly early exiting.

darwin/DarwinProcessList.c
```diff
@@ -128,13 +133,81 @@ static struct kinfo_proc* ProcessList_getKInfoProcs(size_t* count) {
...
 
    ProcessList_init(&this->super, Class(DarwinProcess), usersTable, pidMatchList, userId);
 
    /* Initialize the CPU information */
    this->super.cpuCount = ProcessList_allocateCPULoadInfo(&this->prev_load);
+
+#ifdef HAVE_POWER_GADGET
+   this->power_gadget_is_available = false;
+   this->cpu_count = 0;
+   this->cpu_sample = 0;
```
BenBE, line 206: Implicit by `xCalloc` call from constructor of `this`.

darwin/DarwinProcessList.c
```diff
@@ -153,9 +226,26 @@ ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* pidMatchList, ui
...
+   DarwinProcessList* this = (DarwinProcessList*) super;
+
+#ifdef HAVE_POWER_GADGET
+   if (this->cpu_sample) {
+      PGSample_Release(this->cpu_sample);
+   }
+
+   if (this->power_gadget_is_available) {
+      PG_Shutdown();
+   }
```
BenBE, line 239: May skip those braces. Cf. `STYLEGUIDE`.

BenBE (changes requested):

danchr:
There, I think I got most of your comments?

darwin/DarwinProcessList.c
```diff
@@ -128,13 +133,94 @@ static struct kinfo_proc* ProcessList_getKInfoProcs(size_t* count) {
...
+      }
+
+      // note the check for cpu_count below; we definitely don't want
+      // to corrupt memory should the core/cpu count be inconsistent
+      for (int c = 0; c < core_count && cpu_idx < cpu_count; c++) {
+         double mean, min, max;
+
+         if (show_freq && old_sample)
+            if (PGSample_GetIACoreFrequency(old_sample, new_sample, c,
+                                            &mean, &min, &max))
```
BenBE, line 190: No need to wrap this IMHO.

darwin/DarwinProcessList.c
```diff
@@ -128,13 +133,94 @@ static struct kinfo_proc* ProcessList_getKInfoProcs(size_t* count) {
...
+
+   if (!this->initialized_power_gadget && PG_Initialize != NULL)
+      this->initialized_power_gadget = PG_Initialize();
+
+   if (!this->initialized_power_gadget)
+      return;
+
+   if (!PG_GetNumPackages(&package_count)) {
+      return;
+   }
```
BenBE, line 168: No need for those braces.

darwin/DarwinProcessList.c
```diff
@@ -153,8 +239,26 @@ ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* pidMatchList, ui
...
-   ProcessList_done(this);
+void ProcessList_delete(ProcessList* super) {
+   DarwinProcessList* this = (DarwinProcessList*) super;
+
+#ifdef HAVE_POWER_GADGET
+   if (this->initialized_power_gadget) {
+      for (int i = 0; i < this->cpu_count; i++) {
+         if (this->cpu_samples[i])
+            PGSample_Release(this->cpu_samples[i]);
+      }
```
BenBE, line 250: Inconsistent braces.

BenBE (commented):
Mostly LGTM.

danchr:
There, I think I got most of your comment. I also went through the code again to make sure that it handled the various failure cases correctly.

DisplayOptionsPanel.c
```diff
@@ -117,11 +117,17 @@ DisplayOptionsPanel* DisplayOptionsPanel_new(Settings* settings, ScreenManager*
    Panel_add(super, (Object*) CheckItem_newByRef("Update process names on every refresh", &(settings->updateProcessNames)));
    Panel_add(super, (Object*) CheckItem_newByRef("Add guest time in CPU meter percentage", &(settings->accountGuestInCPUMeter)));
    Panel_add(super, (Object*) CheckItem_newByRef("Also show CPU percentage numerically", &(settings->showCPUUsage)));
+   #ifndef HTOP_DARWIN
```
BenBE, line 120: Should use `defined` due to `#elif defined()`
```suggestion
   #if !defined(HTOP_DARWIN)
```

darwin/DarwinProcessList.h
```diff
@@ -13,7 +13,6 @@ in the source distribution for its full text.
 #include "ProcessList.h"
 #include "zfs/ZfsArcStats.h"
 
-
```
BenBE, line 16: Two blanks after list of headers

BenBE (changes requested):

BenBE (approved):
LGTM. Didn't test due to lack of Apple™ hardware …

fasterit:
As I read it the EULA of the "Intel Power Gadget" is incompatible with the GPLv2, e.g.
~~~text
"LICENSE GRANT:
Subject to all of the terms and conditions of this Agreement, Intel Corporation ("Intel") grants to you a non-exclusive, non-transferable copyright license to use the Materials internally only.

and

LICENSE RESTRICTIONS:
You may not distribute, sublicense or transfer any components of the Materials and derivatives thereof to any third party."
~~~

I don't think we can commit this and make homebrew et al distribute it.

Probably better to write a frequency measurement loop as per https://stackoverflow.com/a/39951453 which points at https://github.com/cyring/CoreFreq/blob/master/corefreqd.c which is GPLv2. The Intel Software mentioned in this PR also disallows re-engineering per their EULA. So not a legal source of info, either.

BenBE:
Due to the licensing issues @fasterit mentioned, I'd suggest to close this issue. Risking license issues for downstream distributions would be a sad result of adding a feature and thus the opposite of what is intended.

While there are ways to work around static linking restrictions in GPLv2 I won't recommend making this PR a precedent. Thus I'll go ahead and close this PR. If the copyright and license situation for the used interfaces improves another go on its inclusion may well follow.

danchr:
Fair enough — I actually hadn't noticed that `htop` is GPL.

PR #605, opened 2021-04-18
Misc

cgzones:


ProcessList.c
```diff
@@ -564,7 +564,7 @@ void ProcessList_rebuildPanel(ProcessList* this) {
 
 Process* ProcessList_getProcess(ProcessList* this, pid_t pid, bool* preExisting, Process_New constructor) {
    Process* proc = (Process*) Hashtable_get(this->processTable, pid);
-   *preExisting = proc;
+   *preExisting = (proc != NULL);
```
BenBE, line 567: Unnecessary parens …

Process.c
```diff
@@ -617,6 +617,7 @@ int Process_compareByKey_Base(const Process* p1, const Process* p2, ProcessField
    case USER:
       return SPACESHIP_NULLSTR(p1->user, p2->user);
    default:
+      assert(0); /* should never be reached */
```
BenBE, line 620: Maybe use comma operator to provide some meaningful message in case of assertion failure.

ProcessList.c
cgzones: removed

Process.c
cgzones: added

Issue #602, opened 2021-04-16
Is the CPU% 4.3  right on FreeBSD?


fundon:
```
FreeBSD: 13.0
htop: 3.0.4
```

<img width="959" alt="Screen Shot 2021-04-16 at 17 27 56" src="https://user-images.githubusercontent.com/27926/115004269-3806e580-9ed9-11eb-9d14-15930b5366ff.png">

cgzones:
Could you try out the patch in #606?

Seems like all other tools on FreeBSD scale the CPU usage on the process swap time.

fundon:
<img width="1909" alt="Screen Shot 2021-04-23 at 14 57 26" src="https://user-images.githubusercontent.com/27926/115831534-4fe3e980-a444-11eb-86ff-25c29676dcce.png">

gaod:
> Could you try out the patch in #606?
> 
> Seems like all other tools on FreeBSD scale the CPU usage on the process swap time.
> 
> In my virtual machine `kern.ccpu` is `0`, so the scale factor is just `1.0`.

The kern.ccpu is set to 0 in ULE scheduler.

fasterit:
Testing report: htop on FreeBSD seems to be showing CPU=0 for every task, both in v3.1.2 as well as current main. This draft PR does not have any impact on this.
The FreeBSD 13.0-RELEASE-p10 VM I test in gets kernelFScale=2048 from the sysctl. That seems right but somewhere down the line the CPU calc seems to be broken.
/DLange

grahamperrin:
For recent `3.3.0-dev`: 

* https://github.com/htop-dev/htop/issues/1208

fasterit:
We're closing stale / probably solved issues while updating the release schedule. This one seems to be in that batch.
Please feel free to reopen with new information if the issue is still showing in the current git main head.

PR #606, opened 2021-04-18
FreeBSD: scale CPU% usage on swap time

cgzones:
See https://forums.freebsd.org/threads/how-does-cpu-utilization-is-calculates.12108/
and https://ess.cs.tu-dortmund.de/lockdoc-elixir/freebsd-lockdoc/lockdoc-v13.0-0.6/source/bin/ps/print.c#L650

Closes: #602

### DIFF
--- a/freebsd/FreeBSDProcessList.c
+++ b/freebsd/FreeBSDProcessList.c
@@ -442,6 +442,13 @@ void ProcessList_goThroughEntries(ProcessList* super, bool pauseProcessUpdate) {
       return;
    }
 
+   int ccpu;
+   size_t size = sizeof(ccpu);
+   if (sysctlbyname("kern.ccpu", &ccpu, &size, NULL, 0) == -1) {
+      ccpu = 0;
+   }
+   const double decayfactor = log(ccpu / kernelFScale);
+
    int count = 0;
    const struct kinfo_proc* kprocs = kvm_getprocs(fpl->kd, KERN_PROC_PROC, 0, &count);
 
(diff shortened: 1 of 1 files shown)


freebsd/FreeBSDProcessList.c
```diff
@@ -442,6 +442,13 @@ void ProcessList_goThroughEntries(ProcessList* super, bool pauseProcessUpdate) {
       return;
    }
 
+   int ccpu;
+   size_t size = sizeof(ccpu);
+   if (sysctlbyname("kern.ccpu", &ccpu, &size, NULL, 0) == -1) {
+      ccpu = 0;
+   }
+   const double decayfactor = log(ccpu / kernelFScale);
```
BenBE, line 450: `kernelFScale` might be zero. Not ensured non-zero by the code querying this from the syscall.

cgzones: https://github.com/htop-dev/htop/blob/feec16cbb53dabc6a52ef2f69a6a13798be82617/freebsd/FreeBSDProcessList.c#L138-L142

You mean `kern.fscale` might be `0`? Is that possible?

BenBE: Highly unlikely, given that there's other code already that does a similar division and I'm not aware of any bug reports regarding this. Mostly DiD note.

gaod: FYI, https://is.cuni.cz/webapps/zzp/download/120118301

BenBE: Actually just looking at the remarked upon code part: `ccpu` is set to `0`, which will cause a math error with the following calculation for `decayfactor`. I don't think that's right?

Explorer09: I've made pull request #1275 that would reject zero `kern.fscale` value.

BenBE:
@cgzones Anything left for this PR to go forward?

fasterit:
@cgzones ping. Does this need to go into main or does this need some testing or ...?

PR #607, opened 2021-04-18
RFC: Drop setting accountGuestInCPUMeter

cgzones:
Introduced in 8ace29c267ba.

This setting affects only Linux.
In the source data /proc/stat from the kernel, the Guest and Guest-Nice
parts are already included in User and Nice.
Htop accounts for that by subtracting them accordingly:
https://github.com/htop-dev/htop/blob/feec16cbb53dabc6a52ef2f69a6a13798be82617/linux/LinuxProcessList.c#L1787-L1789

With the setting accountGuestInCPUMeter disabled these Guest and Guest-
Nice and the Steal CPU share percentage utilizations are not counted
towards the total CPU utilization, but only if detailedCPUTime is
simultaneously enabled.
This is kind of unintuitive and only has the limited affect on the
summed total CPU utilization in a CPU Bar Meter:
[||||||||| X.X%]

Drop this setting and always count the hypervisor related shares Steal
and Guest to the total percentage sum.


BenBE (approved):
Patch itself looks okay.

You can argue for both views: On the one hand hypervisor time is used CPU time, but you could also argue that hypervisor time is not spent by the current system. With the first view you'd at least never underestimate the load on your system (as was an issue with the memory meter recently).

ginggs:
Closing as 'wontfix' after team discussion.  This change would hide necessary information.

PR #627, opened 2021-05-16
Add ELAPSED process column

cgzones:
Add process columns showing the elapsed time since the process was
started.
Similar to STARTTIME, but shows the time passed since the process start
instead of the fixed start time of the process.

Closes https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=782636

### DIFF
--- a/Process.c
+++ b/Process.c
@@ -777,6 +777,7 @@ void Process_writeField(const Process* this, RichString* str, ProcessField field
       Process_printLeftAlignedField(str, attr, procExe, TASK_COMM_LEN - 1);
       return;
    }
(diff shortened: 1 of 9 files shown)


BenBE (commented):
Sure ID 54 isn't taken elsewhere? IIRC on Linux we had global IDs in the Linux column definitions …

Otherwise LGTM.

cgzones:
You mean one of these: https://github.com/htop-dev/htop/commit/d872e363081a892d65dede6a90721d3a2e8b0ee6 ?

BenBE:
Yes. Just wanted to be sure, it's not accidentally conflicting with existing IDs.

Issue #618, opened 2021-05-03
Include pregenerated configure file on in Github tarball


noocsharp:
The release tarball on Bintray used to include a pregenerated configure file. However now that that seems to be gone we have to use the Github tarball, which doesn't.

This lack of a configure file also causes the instructions for building here: https://htop.dev/downloads.html#sources to be incorrect.

fasterit:
@natoscott commented in https://github.com/htop-dev/htop/issues/620#issuecomment-832396011 that we'll not be providing configure scripts inside tarballs anymore. So this issue should probably be sorted by fixing the build instructions to match.

noocsharp:
Is there any way for not providing configure scripts to be reconsidered? My understanding that release tarballs are supposed to include configure scripts to allow the program to be installed without worrying about autotools. It also seems trivial to generate a release tarball by just using `make dist`.

fasterit:
Do you have any rationale for having make, gcc and friends but not autotools on a platform?

We discussed that in the core team but found no real reasoning any more in 2021. Of course it was different when people were mailing around floppys or the lucky ones downloading from Simtel, but those days are gone. Typically it is just one `apt get autoconf` away.

We'd be happy to reconsider the decision if there are good arguments to do so.

noocsharp:
make, a c compiler (tcc also builds htop just fine) and friends are the lowest common denominator. They can be found on any standard system, and don't have any dependencies. If you force the user to have autotools installed, you also force them to have perl installed. So whether intentional or not, you now need perl to build htop.

I think the question should be why _should_ autotools be required for the user to build a program when it is so easy to avoid it? I'm not saying y'all should ditch autotools as your build system, all I ask for is a tarball with a prebuilt configure file. Autotools is designed to be used like this, and nearly all other projects that use autotools as their build system ship their tarballs with a configure file.

BenBE:
I've built `htop` on systems like YoctoLinux and various other platforms where you basically first have to build your own toolchain before even getting to building things like `htop`. And by the time you could build `htop` on those platforms you already had a bunch of other tools built that essentially suffice to also build `autotools`. So even while having `autotools` as a dependency for builds might seem like adding in much trouble, it's not actually much of a new dependency: You can't essentially have e.g. a Debian system without `perl`. And even with things like OpenWRT there's several places where at least for the host (build) environment you'll at some point invoke some perl script or another to build some stuff.

noocsharp:
[...]

PR #639, opened 2021-05-24
README updates/formating

Eideen:
To make it easy to build

### DIFF
--- a/README
+++ b/README
@@ -24,24 +24,47 @@ For more information and details on how to contribute to `htop` visit [htop.dev]
 
 ## Build instructions
 
-This program is distributed as a standard GNU autotools-based package.
+### Prerequisite
+List of build-time dependencies:
+*  `build-essential` standard GNU autotools-based
+*  `autoconf`
+*  `autotools`
+*  `ncurses` (libncursesw*-dev)
 
-Compiling `htop` requires the header files for `ncurses` (libncursesw*-dev). Install these and other required packages for C development from your package manager.
+**Note about `ncurses`:**
+> htop requires ncurses 6.0, the appropriate package is often enough still called libncurses5 (+-dev for headers). Also ncurses usually comes in two flavours:
+>* With Unicode support.
+>* Without Unicode support.
+>
+> This is also something that is usually reflected in the package name (the additional w).
 
-Then, when compiling from a [release tarball](https://github.com/htop-dev/htop/releases), run:
+List of additional build-time dependencies (based on feature flags):
+*  `sensors`
+*  `hwloc`
+*  `libcap`
 
+Compiling `htop` requires the header files for `ncurses` . Install these and other required packages for C development from your package manager.
+
+**Debian/ubuntu**
 ~~~ shell
-./configure && make
+sudo apt install libncursesw5-dev autotools-dev autoconf
 ~~~
 
-Alternatively, for compiling sources downloaded from the Git repository (`git clone` or downloads from [Github releases](https://github.com/htop-dev/htop/releases/)),
-install the header files for `ncurses` (libncursesw*-dev) and other required development packages from your distribution's package manager. Then run:
+List of additional build-time dependencies (based on feature flags):
+*  `sensors`
+*  `hwloc`
+*  `ncurses`
+*  `libcap`
+
+### Compiling from source:
+To compiling from sources downloaded from the Git repository (`git clone` or downloads from [Github releases](https://github.com/htop-dev/htop/releases/)), then run:
 
 ~~~ shell
 ./autogen.sh && ./configure && make
 ~~~
 
-By default `make install` will install into `/usr/local`, for changing the path use `./configure --prefix=/some/path`.
+### Install
+To install to system run `make install`. By default `make install` installs into `/usr/local`. To change this path use `./configure --prefix=/some/path`.
 
 ### Build Options
 
@@ -94,8 +117,25 @@ By default `make install` will install into `/usr/local`, for changing the path 
     dependencies: *pkg-config*(build-time), *libnl-3* and *libnl-genl-3*
     default: *check*
 
-## Usage
 
+## Runtime dependencies:
+`htop` has a set of fixed minimum runtime dependencies, which is kept as minimal as possible:
+* `ncurses` libraries for terminal handling (wide character support).
+
+### Runtime optional dependencies:
+`htop` has a set of fixed optional dependencies, depending on build/configure option used:
+* `libdl`, if not building static and support for some of the optional libraries is enabled, is always required when support for to optionally load dependencies (i.e. `libsensors`, `systemd`) is present.
+* `libcap`, user-space interfaces to the POSIX 1003.1e, is always required when `--enable-capabilities` was used to configure `htop`.
+* `libsensors`, readout of temperatures and CPU speeds, is optional even when `--enable-sensors` was used to configure `htop`.
+* `systemd` is optional when `--enable-static` was not used to configure `htop` (Linux only). If building statically and `libsystemd` is not found by `configure` support for the SystemD meter is disabled entirely.
+
+`htop` checks for the availability of the actual runtime lib as `htop` runs.
+
+**BSD**
+On most *BSD systems you also have `kvm` as a static requirement to read all the kernel information.
+
+More information on required and optional dependencies can be found in [configure.ac](configure.ac).
+## Usage
 See the manual page (`man htop`) or the on-line help ('F1' or 'h' inside `htop`) for a list of supported key commands.
 
 ## Support


README
```diff
@@ -25,24 +25,33 @@ For more information and details on how to contribute to `htop` visit [htop.dev]
 
 ## Build instructions
 
+### prerequisite
```
BenBE, line 28: Headlines should follow existing ones, and thus start with an uppercase letter.

README
```diff
@@ -25,24 +25,33 @@ For more information and details on how to contribute to `htop` visit [htop.dev]
...
 This program is distributed as a standard GNU autotools-based package.
 
 Compiling `htop` requires the header files for `ncurses` (libncursesw*-dev). Install these and other required packages for C development from your package manager.
 
-Then, when compiling from a [release tarball](https://bintray.com/htop/source/htop), run:
+**Debian/ubuntu**
```
BenBE, line 33: If you want to go to the lengths to moronify dependency installation, you probably should do it completely for all supported platforms. A better alternative would be to just mention the required (and optional) libraries required to build. Note that these change for each platform due to different features. `htop` does not only run on Debian/*buntu, but on several other platforms too.

README
```diff
@@ -25,24 +25,33 @@ For more information and details on how to contribute to `htop` visit [htop.dev]
...
 ~~~
 
-Alternatively, for compiling sources downloaded from the Git repository (`git clone` or downloads from [Github releases](https://github.com/htop-dev/htop/releases/)),
-install the header files for `ncurses` (libncursesw*-dev) and other required development packages from your distribution's package manager. Then run:
+### compiling from source:
+Alternatively, for compiling sources downloaded from the Git repository (`git clone` or downloads from [Github releases](https://github.com/htop-dev/htop/releases/)), then run:
 
 ~~~ shell
 ./autogen.sh && ./configure && make
 ~~~
```
BenBE, line 51: This is no longer up-to-date, as the source tarballs will likely not ship the `configure` script anymore (cf. #618), thus the full cycle is necessary. Also you might want to include the call to `make` to actually compile the software. ;-)

BenBE (changes requested):

README
```diff
@@ -25,23 +25,33 @@ For more information and details on how to contribute to `htop` visit [htop.dev]
 
 ## Build instructions
 
+### prerequisite
 This program is distributed as a standard GNU autotools-based package.
 
 Compiling `htop` requires the header files for `ncurses` (libncursesw*-dev). Install these and other required packages for C development from your package manager.
 
+**Debian/ubuntu**
+~~~ Bash
```
BenBE, line 34: Inconsistent markup option with other sections …

BenBE:
Please rebase your branch an squash fixup commits like typo fixes to your own commits.

README
Eideen: I understand what you mean but only listed the required (and optional) libraries. And i think it is imported do to so. 

From a users perspective, that rarely build from source, it is easier to work with some examples what to install. There is 
>   Running `htop` requires `ncurses` libraries (typically named libncursesw*).

*Does that mean i need to install the base library or the  `libncursesw*-dev`?*
*Does all packages managers use the same name for the  `libncursesw*-dev`?*

I currently only run Debian based machines, so it would be wrong of me to state: 
````md

**Redhat/Fedora/CentOS**
~~~ Bash
sudo dnf install libncursesw5-dev autotools-dev autoconf
~~~

````
Without testing it.

Comment (BenBE) [reply README:33]: With the package dependencies there are usually two cases to distinguish: build time and compile time. With `htop` the build time dependencies include all the development headers (on Debian-based distros usually `-dev`), some compiler (`build-essential`) and the actual build system (`autoconf`). If you actually want to run your compiled `htop` binary afterwards, you also need the runtime dependencies, which depend on the features you `configure`.

If you look into the file `configure.ac` you'll notice that especially ncurses is one of those libraries where every system does it's own. In particular: Although `htop` requires ncurses 6.0, the appropriate package is often enough still called `libncurses5` (+`-dev` for headers). Also ncurses usually comes in two flavours: with and without Unicode support. This is also something that is usually reflected in the package name (the additional `w`).

Back to runtime dependencies: `htop` has a set of fixed runtime dependencies, which is kept as minimal as possible. Currently this only includes `ncurses` as well as `libcap` (if capability support is enabled) and `libdl` (i not building static and support for some of the optional libraries is enabled). On most *BSD systems you also have `kvm` as a static requirement to read all the kernel information.

Optional features include things like the `libsensors` readout of temperatures and CPU speeds, or the `systemd` meter, where the build system only needs the appropriate headers at compile time and checks for the availability of the actual runtime lib as `htop` runs.

The full g(l)ory details can be found in `configure.ac` as well in all those files calling `dlopen`.

Which brings us to "What packages do I need?": Usually packages are called similar to the library. Some systems include the `lib` part, others do not. But for all platforms there's a good way to determine the actual name, by looking at the package overview for those systems and looking up, which files they provide.

Comment (Eideen) [reply README:51]: I think it is a great write up, you have done @BenBE. I think it would be great to add a section about dependencies.

For packages theres it to many options like[ Debian/buster](https://packages.debian.org/source/buster/ncurses), 20+ packages, how can a user not familiar with `ncurses`, easily know which to install?

I think 4 lines in a README.md is a well worth it, to bring more users in that can help build from source, for testing.

Comment (BenBE) [reply README:51]: Do you mind adding a proper section regarding the situation with dependencies?

File: README [hunk 5]
```diff
@@ -116,6 +116,23 @@ To install to system run `make install`. By default `make install` will install
     dependencies: *pkg-config*(build-time), *libnl-3* and *libnl-genl-3*
     default: *check*
 
+## runtime dependencies:
+`htop` has a set of fixed runtime dependencies, which is kept as minimal as possible:
+* `ncurses`, libraries for terminal handling (wide character support)
+* `libcap`, *Optional* user-space interfaces to the POSIX 1003.1e
+* `libdl`, if not building static and support for some of the optional libraries is enabled
+* `libsensors`, readout of temperatures and CPU speeds.
```
BenBE, line 124: This one is also optional.

README
```diff
@@ -116,6 +116,23 @@ To install to system run `make install`. By default `make install` will install
...
+* `systemd`
```
BenBE, line 125: This one is optional. TBH, this was actually the first library we explicitly included in a way so we can run without it.

README
```diff
@@ -116,6 +116,23 @@ To install to system run `make install`. By default `make install` will install
...
+
+`htop` checks for the availability of the actual runtime lib as `htop` runs.
+
+**BSD**    
+On most *BSD systems you also have `kvm` as a static requirement to read all the kernel information.
+
+
+For more info [configure.ac](configure.ac)
```
BenBE, line 133:
```suggestion
More information on required and optional dependencies can be found in [configure.ac](configure.ac).
```

BenBE (changes requested):

README
Eideen: So all of the runtime dependencies are optional?

> ## Runtime optional libraries:
> `htop` has a set of fixed runtime optional libraries for more features, which is kept as minimal as possible:
> 
> * `ncurses`, libraries for terminal handling (wide character support)
> * `libcap`, user-space interfaces to the POSIX 1003.1e
> * `libdl`, if not building static and support for some of the optional libraries is enabled
> * `libsensors`, readout of temperatures and CPU speeds.
> * `systemd`

BenBE: * `ncurses` is always required.
* `libcap` is always required when `--enable-capabilities` was used to configure `htop`.
* `libsensors` is optional even when `--enable-sensors` was used to configure `htop`.
* `systemd` is optional when `--enable-static` was not used to configure `htop` (Linux only). If building statically and `libsystemd` is not found by `configure` support for the SystemD meter is disabled entirely.
* `libdl` is always required when support for to optionally load dependencies (i.e. `libsensors`, `systemd`) is present.

Eideen:
I add the changes as a fixup.

README
```diff
@@ -25,24 +25,45 @@ For more information and details on how to contribute to `htop` visit [htop.dev]
...
 
-This program is distributed as a standard GNU autotools-based package.
-
-Compiling `htop` requires the header files for `ncurses` (libncursesw*-dev). Install these and other required packages for C development from your package manager.
-
-Then, when compiling from a [release tarball](https://bintray.com/htop/source/htop), run:
-
+### Prerequisite
+
+List of Build time dependency's:
```
BenBE, line 30:
```suggestion
List of build-time dependencies:
```

README
```diff
@@ -25,24 +25,45 @@ For more information and details on how to contribute to `htop` visit [htop.dev]
...
+*  `ncurses` (libncursesw*-dev)
+
+**Note about `ncurses`:**
+> htop requires ncurses 6.0, the appropriate package is often enough still called libncurses5 (+-dev for headers). Also ncurses usually comes in two flavours:
+* With Unicode support.
+* Without Unicode support.
+>
+> This is also something that is usually reflected in the package name (the additional w).
+
+List of Build time optional:
```
BenBE, line 43:
```suggestion
List of additional build-time dependencies (based on feature flags):
```

README
```diff
@@ -25,24 +25,45 @@ For more information and details on how to contribute to `htop` visit [htop.dev]
...
+*  `sensors`
+*  `hwloc`
+*  `ncurses`
```
BenBE, line 46: Isn't `ncurses` duplicate in this list?

README
```diff
@@ -25,24 +25,45 @@ For more information and details on how to contribute to `htop` visit [htop.dev]
...
+*  `libcap`
+
+Compiling `htop` requires the header files for `ncurses` . Install these and other required packages for C development from your package manager.
+
+**Debian/ubuntu**
```
BenBE, line 51:
```suggestion
**Debian/Ubuntu**
```

README
```diff
@@ -25,24 +25,45 @@ For more information and details on how to contribute to `htop` visit [htop.dev]
...
+### Compiling from source or Github tarball:
+When compiling from a [release tarball](https://bintray.com/htop/source/htop), or compiling sources downloaded from the Git repository (`git clone` or downloads from [Github releases](https://github.com/htop-dev/htop/releases/)), then run:
 
 ~~~ shell
 ./autogen.sh && ./configure && make
 ~~~
 
-By default `make install` will install into `/usr/local`, for changing the path use `./configure --prefix=/some/path`.
+### Install
+To install to system run `make install`. By default `make install` will install into `/usr/local`, for changing the path use `./configure --prefix=/some/path`.
```
BenBE, line 66:
```suggestion
To install to system run `make install`. By default `make install` installs into `/usr/local`. To change this path use `./configure --prefix=/some/path`.
```

README
```diff
@@ -95,6 +116,28 @@ By default `make install` will install into `/usr/local`, for changing the path
...
+## Runtime dependencies:
+`htop` has a set of fixed minimum runtime dependencies, which is kept as minimal as possible:
+* `ncurses` libraries for terminal handling (wide character support).
+
+### Runtime optional dependencies:
+`htop` has a set of fixed optional dependencies, depending on build/configure option used:
+* `libdl`, if not building static and support for some of the optional libraries is enabled, is always required when support for to optionally load dependencies (i.e. `libsensors`, `systemd`) is present.
+* `libcap`, user-space interfaces to the POSIX 1003.1e, is always required when `--enable-capabilities` was used to configure `htop`.
+* `libsensors`, readout of temperatures and CPU speeds, is optional even when `--enable-sensors` was used to configure `htop`.
+* `systemd` is optional when `--enable-static` was not used to configure `htop` (Linux only). If building statically and `libsystemd` is not found by `configure` support for the SystemD meter is disabled entirely.
```
BenBE, line 128: Please check the grammar in this line.

README
```diff
@@ -95,6 +116,28 @@ By default `make install` will install into `/usr/local`, for changing the path
...
+
+
+
```
BenBE, line 131: No need for so much blank space …

BenBE (changes requested):

README
Eideen: I have looked at it, the line is a copy from [you](https://github.com/htop-dev/htop/pull/639#discussion_r650382302),

> * `systemd` is optional when `--enable-static` was not used to configure `htop` (Linux only). If building statically and `libsystemd` is not found by `configure` support for the SystemD meter is disabled entirely.

It looks good to me.

BenBE: k. Got tripped by the support after `configure` being part of the action, not the condition in that sentence. So while grammatically correct it's probably not the best way to phrase it as it you can get trip easily while trying to parse that sentence.

BenBE (approved):
Please prefer to rebase your commits onto the target branch instead of merging.

PR #641, opened 2021-05-25
Handle CWD as a globally available field

BenBE:


BenBE:
The code for `*BSDProcessList_updateCwd` is identical on DragonFlyBSD, FreeBSD and also NetBSD (AFAIRTFM, plz check @fraggerfox). On OpenBSD it's essentially identical with the only difference being the MIB passed in to the `sysctl`. So basically we could generate the full function body via a macro in `generic/` that takes the name of the field to write, the MIB initializer and be done with it. The field to write is also interesting, as for at least DragonFly, FreeBSD and NetBSD the `*BSDProcessList_updateExe` function is essentially the same, also just differing in the MIB being passed.

solaris/SolarisProcessList.c
```diff
@@ -309,6 +309,19 @@ static void SolarisProcessList_updateExe(pid_t pid, Process* proc) {
...
+   char path[32];
+   xSnprintf(path, sizeof(path), "/proc/%d/cwd", pid);
+
+   char target[PATH_MAX];
+   ssize_t ret = readlink(path, target, sizeof(target) - 1);
+   if (ret <= 0)
+      return;
+
+   target[ret] = '\0';
+   Process_updateExe(proc, target);
```
cgzones, line 322: ?

BenBE: ECONCENTRATION …

natoscott (approved):
@BenBE LGTM too

PR #649, opened 2021-06-09
Linux: drop O_PATH usage and update process uid on change

### LINKED ISSUE #534: [Regression] Htop >= 3.0.3 does not list any processes
On GoboLinux, since version 3.0.3 htop's process list is empty for me. I have also tried the latest version 3.0.5.

Any release prior to that works as expected. Bug reference: https://github.com/gobolinux/Recipes/issues/111 (you will find some logs there)

(Ncurses v6.2, Linux 5.6.10)

cgzones:
* O_PATH is available since Linux 2.6.39, but we are using fstat(2) on the returned file descriptor in LinuxProcessList_statProcessDir(), which is only supported since Linux 3.6.

  Fixes #534

* Always check if the user of a process changed, e.g. by using setuid(2).

### DIFF
--- a/Process.c
+++ b/Process.c
@@ -963,6 +963,7 @@ void Process_init(Process* this, const Settings* settings) {
    this->show = true;
    this->updated = false;
    this->cmdlineBasenameEnd = -1;
+   this->st_uid = (uid_t)-1;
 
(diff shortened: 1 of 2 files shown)


BenBE (approved):
Patch LGTM.

Issue #648, opened 2021-06-06
Linux: incorrect CPU frequencies when some CPUs are offline


mwahlroos:
The CPU frequency figures may show incorrect (or unnecessarily "N/A") values for CPU cores if some CPUs or cores have been taken logically offline.

The way I read it, the code currently sums the number of CPUs from /proc/stat, and then proceeds to read the frequencies for CPUs 0..N-1 (for N total CPUs as determined from the stat file) either from CPUfreq information in sysfs, or by parsing /proc/cpuinfo if frequency information from CPUfreq is unavailable or too slow to read.

This works correctly if all CPUs from 0 to N-1 are online.

However, if, say, on a quad-core system (CPUs 0..3), cores 1 and 3 have been taken logically offline, the total number of CPUs that appear in /proc/stat is 2, and the code attempts to read the frequencies for cores 0..1 rather than 0 and 2, although CPU 1 is offline with no valid frequency information available, and the second online core is actually number 2.

On my system, the frequency read from CPUfreq for CPU 1 would then appear to be some fixed value that's incorrect for the actual second online core. If the frequencies are read from /proc/cpuinfo instead, the frequency reported in htop for the second core would appear to be "N/A" instead.

This may be a bit of a weird corner case, but I've seen it when taking every second core offline to soft-disable SMT, so I thought I'd report it. I don't know how the entries in /sys/devices/system/cpu/ change if CPUs are disabled in a system with actual CPU hotplug, but I suppose this might affect some such systems as well.

I've been working on this a bit but I don't have a fix yet.

BenBE:
Seeing you mention `procfs` I'll assume you're noticing this issue on Linux. Also given this code hasn't been touched since the last release AFAIR, this ain't be a very recent breakage, but something that happened to break when reading the sensor information was last touched.

mwahlroos:
Yes, this is on Linux, Fedora 34 to be precise.

And sorry I didn't remember to mention the htop version. I'm seeing this both on 3.0.5 and git master. It doesn't necessarily look like a regression to me but rather just something that might not have occurred as a possibility to whoever wrote the frequency reading code. The logic of the bug/omission seems pretty clear once you've seen it and read the code, so I haven't even tried bisecting its history.

cgzones:
@mwahlroos has this be solved by #656?

mwahlroos:
> @mwahlroos has this be solved by #656?

@cgzones yes, that does resolve this. Thanks for your work on this!

PR #650, opened 2021-06-09
Linux: handle offline CPUs when getting frequencies

mwahlroos:
Fixes: #648 

When getting frequencies from CPUFreq, the CPU nodes in sysfs are iterated from 0 to M-1 where M is the number of configured CPUs in the system and which should include offline ones as well, thus allowing for iterating through the full number of possible CPUs. The ones that are offline according to sysfs are ignored, and frequencies are fetched for up to N others where N is the number for which CPUData entries have been allocated.

When getting frequencies from `/proc/cpuinfo`, the CPU nodes and their frequencies are just taken in the order in which they appear. Offline CPUs do not seem to appear in `cpuinfo`, so they're automatically skipped. The CPU IDs from `cpuinfo` are ignored, apart from checking the output line format.

Possible issues that I can think of:

- The sysconf variable `_SC_NPROCESSORS_CONF` used for getting the total number of CPUs here (in the CPUFreq code path) isn't standard, but it seems to have been mentioned on the Linux man page `sysconf(3)` since circa 2011.
- It might not be the best idea to check the online status of every CPU node separately from sysfs. I don't know how fast or slow reading that information is, or if it has performance implications if there's a large number of CPUs or cores. It might be possible to parse e.g. `/sys/devices/system/cpu/online` for a list of online CPUs instead, which would require a bit of work to do properly but would allow for a single read.
- The `/proc/cpuinfo` code path in its form suggested here assumes that offline CPUs do not appear and that the online nodes appear in the order of their CPU IDs. The author of the previous code didn't seem to want to make that assumption, and I'm not sure if it's valid.

Any feedback is welcome!

### DIFF
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -1824,8 +1824,33 @@ static inline double LinuxProcessList_scanCPUTime(ProcessList* super) {
    return period;
 }
 
+static bool cpuIsOnline(unsigned int cpuId) {
+      /* if there's no sysfs file for the CPU node indicating online/offline status,
+       * the CPU is probably online, so assume online by default
+       */
+      bool online = 1;
+      char pathBuffer[64];
+
+      xSnprintf(pathBuffer, sizeof(pathBuffer), "/sys/devices/system/cpu/cpu%u/online", cpuId);
+
+      FILE* file = fopen(pathBuffer, "r");
+      unsigned int readval;
+      if (file) {
+         if (fscanf(file, "%u", &readval) == 1) {
+            online = readval;
+         }
+         fclose(file);
+      }
+
+      return online;
+}
+
+static int getConfiguredCPUCount(void) {
+   return (int) sysconf(_SC_NPROCESSORS_CONF);
+}
+
 static int scanCPUFreqencyFromSysCPUFreq(LinuxProcessList* this) {
-   unsigned int cpus = this->super.cpuCount;
+   int configuredCPUs = getConfiguredCPUCount();
    int numCPUsWithFrequency = 0;
    unsigned long totalFrequency = 0;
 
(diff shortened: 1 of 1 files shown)


linux/LinuxProcessList.c
```diff
@@ -1824,8 +1824,33 @@ static inline double LinuxProcessList_scanCPUTime(ProcessList* super) {
...
+static bool cpuIsOnline(unsigned int cpuId) {
+      /* if there's no sysfs file for the CPU node indicating online/offline status,
+       * the CPU is probably online, so assume online by default
+       */
+      bool online = 1;
+      char pathBuffer[64];
+
+      xSnprintf(pathBuffer, sizeof(pathBuffer), "/sys/devices/system/cpu/cpu%u/online", cpuId);
+
+      FILE* file = fopen(pathBuffer, "r");
```
BenBE, line 1836: Please use the wrapper function for `openat` or `readfile` to reduce the number of FS level operations.

linux/LinuxProcessList.c
```diff
@@ -1824,8 +1824,33 @@ static inline double LinuxProcessList_scanCPUTime(ProcessList* super) {
...
+         }
+         fclose(file);
+      }
+
+      return online;
+}
+
+static int getConfiguredCPUCount(void) {
+   return (int) sysconf(_SC_NPROCESSORS_CONF);
+}
```
BenBE, line 1850: This should be available globally already IIRC.

linux/LinuxProcessList.c
```diff
@@ -1824,8 +1824,33 @@ static inline double LinuxProcessList_scanCPUTime(ProcessList* super) {
...
+
 static int scanCPUFreqencyFromSysCPUFreq(LinuxProcessList* this) {
-   unsigned int cpus = this->super.cpuCount;
+   int configuredCPUs = getConfiguredCPUCount();
```
BenBE, line 1853: Try to cache that value and avoid requesting it on every run; occasional refresh is no problem though.

linux/LinuxProcessList.c
```diff
@@ -1843,12 +1868,25 @@ static int scanCPUFreqencyFromSysCPUFreq(LinuxProcessList* this) {
...
+         continue;
+
       xSnprintf(pathBuffer, sizeof(pathBuffer), "/sys/devices/system/cpu/cpu%u/cpufreq/scaling_cur_freq", i);
 
       struct timespec start;
-      if (i == 0)
+      if (onlineCPUIndex == 0)
          clock_gettime(CLOCK_MONOTONIC, &start);
 
       FILE* file = fopen(pathBuffer, "r");
```
BenBE, line 1892: Avoid `fopen` in favour of `openat`; `opendir` the path `/sys/devices/system/cpu` and only scan from there. May combine this with the above and opendir `cpu%d` additionally.

linux/LinuxProcessList.c
```diff
@@ -1859,14 +1897,14 @@ static int scanCPUFreqencyFromSysCPUFreq(LinuxProcessList* this) {
       if (fscanf(file, "%lu", &frequency) == 1) {
```
BenBE, line 1897: Use `readfile` API

BenBE (changes requested):
Please follow the style guide regarding indentation.

linux/LinuxProcessList.c
mwahlroos: I tried looking for that but couldn't find it.

ProcessList has [cpuCount](https://github.com/htop-dev/htop/blob/8154125d4bf26fff20241eebe17954254409a522/ProcessList.h#L84), which is updated on Linux by [LinuxProcessList_updateCPUcount](https://github.com/htop-dev/htop/blob/8154125d4bf26fff20241eebe17954254409a522/linux/LinuxProcessList.c#L167), but that number only includes the number of active CPUs, and using it as the upper limit for the CPU id leads to precisely the problem that I'm trying to solve.

Any pointers to where this would be available globally?

cgzones:
@mwahlroos can this be closed as #656 has been merged?

mwahlroos:
Closing this PR as #656 does this better.

PR #651, opened 2021-06-10
Minor cleanup to the PCP platform from further valgrind testing

natoscott:


pcp/Platform.c
```diff
@@ -370,6 +370,10 @@ void Metric_enableThreads(void) {
 }
 
 bool Metric_fetch(struct timeval *timestamp) {
+   if (pcp->result) {
+      pmFreeResult(pcp->result);
+      pcp->result = NULL;
+   }
    int sts = pmFetch(pcp->total, pcp->fetch, &pcp->result);
    if (sts < 0) {
       if (pmDebugOptions.appl0)
```
BenBE, line 379: Unnecessary extra closing paren in error message …

pcp/Platform.c
```diff
@@ -370,6 +370,10 @@ void Metric_enableThreads(void) {
 }
 
 bool Metric_fetch(struct timeval *timestamp) {
+   if (pcp->result) {
+      pmFreeResult(pcp->result);
```
BenBE, line 374: If `pmFreeResult` is `NULL`-safe, why check for `NULL` here? If it's not, then `Platform_done` should have that check too.

pcp/Platform.c
```diff
@@ -502,6 +506,8 @@ void Platform_init(void) {
 
 void Platform_done(void) {
    pmDestroyContext(pcp->context);
+   pmFreeResult(pcp->result);
```
BenBE, line 509: Shouldn't the result be freed before freeing the context? Which resource is dependent on which?

BenBE (changes requested):

pcp/Platform.c
natoscott: Yep, needs a NULL check in done too - thanks!

pcp/Platform.c
natoscott: There's no dependency there - doesn't matter which happens first.

cgzones (approved):

BenBE (approved):

PR #655, opened 2021-06-12
Limit deleted library check

cgzones:
Reading and parsing `/proc/<pid>/maps` is quite expensive.

Do not check for deleted libraries if the main binary has been deleted;
in this case the deleted binary takes precedence.

Do not check in threads.  The check is void for kernel threads and user-
land threads can just inherit the state from the main process structure.

### DIFF
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -1354,7 +1354,8 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
       {
          bool prev = proc->usesDeletedLib;
 
-         if ((lp->m_lrs == 0 && (settings->flags & PROCESS_FLAG_LINUX_LRS_FIX)) || settings->highlightDeletedExe) {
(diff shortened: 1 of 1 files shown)


BenBE (approved):

linux/LinuxProcessList.c
```diff
@@ -1365,8 +1366,8 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
                LinuxProcessList_readMaps(lp, procFd, settings->flags & PROCESS_FLAG_LINUX_LRS_FIX, settings->highlightDeletedExe);
             }
          } else {
-            /* reset if setting got disabled */
-            proc->usesDeletedLib = false;
+            /* Copy from process structure in threads and reset if setting got disabled */
+            proc->usesDeletedLib = proc->isUserlandThread ? parent->usesDeletedLib : false;
```
BenBE, line 1370:
```suggestion
            proc->usesDeletedLib = (proc->isUserlandThread && parent) ? parent->usesDeletedLib : false;
```

PR #656, opened 2021-06-12
Rework CPU counting

cgzones:
Currently htop does not support offline CPUs and hot-swapping, e.g. via
    `echo 0 > /sys/devices/system/cpu/cpu2/online`

Split the current single `cpuCount` variable into `activeCPUs` and `existingCPUs`.

Supersedes: #650
Related: #580

/cc @mwahlroos @sthen 

@natoscott please take a look at the PCP code

TODO:

- [x] Check on Darwin
- [x] Check on DragonFlyBSD
- [x] Check on FreeBSD
- [x] Check on NetBSD (#603)
- [x] Check on OpenBSD
- [x] Check on Solaris
- [x] Check with PCP
- [x] Hide offline CPUs in affinity panel

### DIFF
--- a/ProcessList.h
+++ b/ProcessList.h
@@ -87,9 +87,11 @@ typedef struct ProcessList_ {
    unsigned int existingCPUs;
 } ProcessList;
 
+/* Implemented by platforms */
 ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* dynamicMeters, Hashtable* pidMatchList, uid_t userId);
 void ProcessList_delete(ProcessList* pl);
 void ProcessList_goThroughEntries(ProcessList* super, bool pauseProcessUpdate);
+bool ProcessList_isCPUonline(const ProcessList* super, unsigned int id);
 
 
 ProcessList* ProcessList_init(ProcessList* this, const ObjectClass* klass, UsersTable* usersTable, Hashtable* dynamicMeters, Hashtable* pidMatchList, uid_t userId);
--- a/ProcessList.h
+++ b/ProcessList.h
@@ -83,7 +83,8 @@ typedef struct ProcessList_ {
    memory_t usedSwap;
    memory_t cachedSwap;
 
-   unsigned int cpuCount;
+   unsigned int activeCPUs;
+   unsigned int existingCPUs;
 } ProcessList;
 
 ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* dynamicMeters, Hashtable* pidMatchList, uid_t userId);
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -158,30 +158,85 @@ static void LinuxProcessList_initNetlinkSocket(LinuxProcessList* this) {
 
 #endif
 
-static void LinuxProcessList_updateCPUcount(ProcessList* super, FILE* stream) {
+static void LinuxProcessList_updateCPUcount(ProcessList* super) {
+   /* Similiar to get_nprocs_conf(3) / _SC_NPROCESSORS_CONF
+    * https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/getsysstats.c;hb=HEAD
+    */
+
    LinuxProcessList* this = (LinuxProcessList*) super;
+   unsigned int existing = 0, active = 0;
 
-   unsigned int cpus = 0;
-   char buffer[PROC_LINE_LENGTH + 1];
-   while (fgets(buffer, sizeof(buffer), stream)) {
-      if (String_startsWith(buffer, "cpu")) {
-         cpus++;
+   DIR* dir = opendir("/sys/devices/system/cpu");
+   if (!dir) {
+      super->activeCPUs = 1;
+      super->existingCPUs = 1;
+      this->cpuData = xReallocArray(this->cpuData, 2, sizeof(CPUData));
+      this->cpuData[0].online = true;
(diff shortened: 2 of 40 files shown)


solaris/SolarisProcessList.c
```diff
@@ -46,32 +46,70 @@ static char* SolarisProcessList_readZoneName(kstat_ctl_t* kd, SolarisProcess* sp
...
 }
 
+static void SolarisProcessList_updateCPUcount(ProcessList* super) {
+   SolarisProcessList* spl = (SolarisProcessList*) super;
+   long int s;
+   bool change = false;
+
+   s = sysconf(_SC_NPROCESSORS_CONF);
+   if (s < 1)
+      CRT_fatalError("Cannot get exisitng CPU count by sysconf(_SC_NPROCESSORS_CONF)");
```
natoscott, line 56: typo on "existing".

ProcessList.h
```diff
@@ -81,12 +81,15 @@ typedef struct ProcessList_ {
    memory_t usedSwap;
    memory_t cachedSwap;
 
-   unsigned int cpuCount;
+   unsigned int activeCPUs;
+   unsigned int existingCPUs;
```
natoscott, line 85: No big deal but I find the naming slightly unclear - AIUI 'active' means 'online' and 'existing' means the CPU exists but is either online or offline?  An alternative might be to use 'onlineCPUs' and something like 'actualCPUs' or 'realCPUs' here?

linux/LinuxProcessList.c
```diff
@@ -158,30 +158,85 @@ static void LinuxProcessList_initNetlinkSocket(LinuxProcessList* this) {
 
 #endif
 
-static void LinuxProcessList_updateCPUcount(ProcessList* super, FILE* stream) {
+static void LinuxProcessList_updateCPUcount(ProcessList* super) {
```
natoscott, line 161: Hmm.  This sysfs-walk looks like a relatively expensive approach - going from what is currently a handful of syscalls at most, per sample, to potentially many thousands of additional syscalls with each sample on large CPU count machines.

Does an offline CPU appear at all in /proc/stat output?  If not, we should be able to do this operation using a combination of a/ one-time call to sysconf(_SC_NPROCESSORS_CONF) at startup and, b/ CPU state-management based on presence (or zero values?) in /proc/stat (while reading stats) at sample time... i.e. if cpu22 of 32 is missing in /proc/stat mark it as !online ... and update activeCPU count accordingly - would that work?

linux/LinuxProcessList.c
```diff
@@ -1771,33 +1824,50 @@ static inline void LinuxProcessList_scanZfsArcstats(LinuxProcessList* lpl) {
...
+      // The rest will remain at zero.
+      unsigned int adjCpuId;
       if (i == 0) {
          (void) sscanf(buffer,   "cpu  %16llu %16llu %16llu %16llu %16llu %16llu %16llu %16llu %16llu %16llu",         &usertime, &nicetime, &systemtime, &idletime, &ioWait, &irq, &softIrq, &steal, &guest, &guestnice);
+         adjCpuId = 0;
       } else {
          unsigned int cpuid;
          (void) sscanf(buffer, "cpu%4u %16llu %16llu %16llu %16llu %16llu %16llu %16llu %16llu %16llu %16llu", &cpuid, &usertime, &nicetime, &systemtime, &idletime, &ioWait, &irq, &softIrq, &steal, &guest, &guestnice);
-         assert(cpuid == i - 1);
+         adjCpuId = cpuid + 1;
```
natoscott, line 1859: Ah, this is what I was looking for in that previous comment.  :)  So an alternate approach to finding offline CPUs might be to mark all offline at start of this function, and then mark each individually as 'online' as we parse the file and find a CPU (and update counts as we go if we need total online vs actual).

ProcessList.h
BenBE: I'd prefer `actualCPUs` over `realCPUs` to not clash with virtualization stuff …

linux/LinuxProcessList.c
mwahlroos: Reporter of the issue that probably got this started here. I have no other authority on the matter. :)

Another alternative in case of Linux would be to parse the list of online CPU nodes at `/sys/devices/system/cpu/online` on update. That would be single read from sysfs rather than one read per CPU.

The sysfs files are listed in the [Linux sysfs documentation](https://github.com/torvalds/linux/blob/master/Documentation/ABI/testing/sysfs-devices-system-cpu). The format for the CPU-listing files seems to be e.g. "0-3,5,7", which in case of the aggregate `online` file would mean that CPU nodes 0 to 3, and additionally 5 and 7 are online.

That would give explicit and authoritative information on which CPUs are currently being scheduled. It might of course be simpler to just assume that the ones not appearing in `/proc/stat` are offline. That seems to be the case on my machine but I have no broader knowledge on how the file lists the CPUs.

fasterit:
Works on MacOSX 11.3.1 (Intel CPU)
https://discussions.apple.com/thread/250559322 on how to offline CPUs / Hypertreading on MacOS

linux/LinuxProcessList.c
cgzones: The original proposal used `sysconf(_SC_NPROCESSORS_CONF)`, which is implemented as a sysfs-walk, see https://code.woboq.org/userspace/glibc/sysdeps/unix/sysv/linux/getsysstats.c.html#236.
By doing the walk on our own we can search specific for the data we need.
Maybe at RC stage we can take a more serious look on performance.

Issue #526, opened 2021-02-13
Custom Meters from external source


BenBE:
Based on several requests (cf. #406, [#899<sup>historic</sup>](https://github.com/hishamhm/htop/issues/899)) and as the general case would be quite interesting, it might be good to have a generic "Custom Script Meter".

The basic idea for such a meter would be, that `htop` could allow for up to 4 external "meter scripts" that it would execute and run in the background. The output of the program would then be parsed into a meter as follows:

```
bar_range:0 1000
bar_data:666 123 42 23
bar_color:4 6 2 1
bar_label:789/1000
text:Custom meter showing 666 foo per 1000 bar. 42 baz happened.
led:666 foo / 123 bar / 42 baz

```

The lines have the following meaning:
* `bar_range`: Specifies the min and max limits of the bar meter. Values in `bar_data` are scaled accordingly
* `bar_data`: Up to seven integer values separated by space giving the data to show in the bar graph.
* `bar_color`: The colors for up to seven channels denoted by up to 7 integers separated by spaces (1=blue, 2=green, 3=cyan, 4=red, 5=magenta, 6=yellow/orange, 7=grey, 8=shadowed). Missing colors are defaulted to 7 (grey).
* `bar_label`: The text to display right-aligned in the meter. Text left-truncated at spaces if too long. Merges color with displayed bar values.
* `text`: Text to display in the text meter style.
* `led`: Text for the "large digits" display. Digits are automatically displayed as large retro 7-segment LED characters. Otherwise identical to text. If not present, caption from `text` is used.

The option value starts directly after the colon (`:`). For options giving plain numbers, these need to be separated by exactly one space, no decimal separator or thousands separator may be given. Invalid values are dropped and the whole record is replaced by its default values. Settings for `bar_range` and `bar_color` are remembered (thus only need to be given once), other values must be given at 1Hz or higher. Parsing for the values of one second ends when encountering a blank line. Last record parsed is displayed.

fasterit:
Thanks for writing the idea up. I think this would make a lot of sense.

Just a few random thoughts from reading through your initial proposal:

variable=value is much more common than variable:value, I'd stick with "="

htop could just listen to a named pipe (FIFO) and programs sending "meter info" there would get bars or column info as requested displayed. Records need to be smaller than `PIPE_BUF`(512 Unix, 4096 on Linux) to allow sequenced concurrency.

[...]

PR #669, opened 2021-06-23
Introduce the DynamicMeter class for runtime extension

natoscott:
This commit is based on exploratory work by Sohaib Mohamed.

The end goal is two-fold - to support addition of Meters we
build via configuration files for both the PCP platform and
for scripts ( https://github.com/htop-dev/htop/issues/526 )

Here, we focus on generic code and the PCP support.  A new
class DynamicMeter is introduced - its use the special case
'param' field handling that previously was used only by the
CPUMeter, such that every runtime-configured Meter is given
a unique identifier.  Unlike with the CPUMeter this is used
internally only.  When reading/writing to htoprc instead of
CPU(N) - where N is an integer param (CPU number) - we use
the string name for each meter.  For example, if we have a
configuration for a DynamicMeter for some Redis metrics, we
might read and write "Dynamic(redis)".  This identifier is
subsequently matched (back) up to the configuration file so
we're able to re-create arbitrary user configurations.

The PCP platform configuration file format is fairly simple.
We expand configs from several directories, including the
users homedir alongside htoprc (below htop/meters/) and also
/etc/pcp/htop/meters.  The format will be described via a
new pcp-htop(5) man page, but its basically ini-style and
each Meter has one or more metric expressions associated, as
well as specifications for labels, color and so on via a dot
separated notation for individual metrics within the Meter.

A few initial sample configuration files are provided below
./pcp/meters that give the general idea.  The PCP "derived"
metric specification - see pmRegisterDerived(3) - is used
as the syntax for specifying metrics in PCP DynamicMeters.


AvailableMetersPanel.c
```diff
@@ -91,6 +92,47 @@ const PanelClass AvailableMetersPanel_class = {
...
+   Panel* super;
+   unsigned int id;
+   unsigned int offset;
+} DynamicIterator;
+
+static void AvailableMetersPanel_addDynamicMeter(ATTR_UNUSED ht_key_t key, void* value, void* data) {
+   const DynamicMeter* meter = (const DynamicMeter*)value;
+   DynamicIterator* iter = (DynamicIterator*)data;
+   unsigned int identifier = (iter->offset << 16) | iter->id;
+   const char* label = meter->description? meter->description : meter->name;
```
BenBE, line 119:
```suggestion
   const char* label = meter->description ? meter->description : meter->name;
```

DynamicMeter.c
```diff
@@ -0,0 +1,92 @@
...
+
+#include "DynamicMeter.h"
+
+#include "CRT.h"
+#include "Object.h"
+#include "Platform.h"
+#include "ProcessList.h"
+#include "RichString.h"
+#include "XUtils.h"
+
```
BenBE, line 18:
```suggestion


```

DynamicMeter.c
```diff
@@ -0,0 +1,92 @@
...
+static const int DynamicMeter_attributes[] = {
+   DYNAMIC_GRAY,
```
BenBE, line 20: Why not also offer dark gray too?

DynamicMeter.c
```diff
@@ -0,0 +1,92 @@
...
+
+unsigned int DynamicMeter_search(const ProcessList* pl, const char* name) {
+   DynamicIterator iter = { .key = 0, .name = name };
+   Hashtable_foreach(pl->dynamicMeters, DynamicMeter_compare, &iter);
+   return iter.key;
+}
+
+const char* DynamicMeter_lookup(const ProcessList* pl, unsigned int key) {
+   const DynamicMeter* meter = Hashtable_get(pl->dynamicMeters, key);
+   return meter->name;
```
BenBE, line 54: Missing NULL check for keys that don't exist?

DynamicMeter.c
```diff
@@ -0,0 +1,92 @@
...
+
+static void DynamicMeter_display(const Object* cast, RichString* out) {
+   const Meter* meter = (const Meter*)cast;
+   Platform_dynamicMeterDisplay(meter, out);
+}
+
+static void DynamicMeter_getUiName(const Meter* this, char* name, size_t length) {
+   const ProcessList* pl = this->pl;
+   const DynamicMeter* meter = Hashtable_get(pl->dynamicMeters, this->param);
+   xSnprintf(name, length, "%s", meter->name);
```
BenBE, line 73: Possible NULL deref on `meter`

DynamicMeter.h
```diff
@@ -0,0 +1,26 @@
...
+   char* description;
+   unsigned int type;
+   double maximum;
+
+   void* dynamicData;  /* platform-specific meter data */
+} DynamicMeter;
+
+Hashtable* DynamicMeters_new(void);
+
+const char* DynamicMeter_lookup(const ProcessList* pl, unsigned int);
```
BenBE, line 20: Missing argument name for second parameter.

AvailableMetersPanel.c
```diff
@@ -91,6 +92,47 @@ const PanelClass AvailableMetersPanel_class = {
...
+   Panel_add(iter->super, (Object*) ListItem_new(label, identifier));
+   iter->id++;
+}
+
+// Handle (&DynamicMeter_class) entries in the AvailableMetersPanel
+static void AvailableMetersPanel_addDynamicMeters(Panel* super, const ProcessList *pl, unsigned int offset) {
+   DynamicIterator iter = { .super = super, .id = 1, .offset = offset };
+   Hashtable_foreach(pl->dynamicMeters, AvailableMetersPanel_addDynamicMeter, &iter);
```
BenBE, line 127: NULL deref on platforms w/o dynamic meters (when `DynamicMeters_new` returns NULL).

BenBE (changes requested):

AvailableMetersPanel.c
natoscott: Fixed, thanks.

DynamicMeter.c
natoscott: Sure.

DynamicMeter.c
natoscott: OK, will add a guard.

AvailableMetersPanel.c
natoscott: Hmm - that would require the platform to have a DynamicMeter_class entry in its Platform_meterTypes array though, which would be a monumental botch.  So, I don't think this one warrants extra special case handling.

BenBE: Then this at least should get a proper `assert` (if not here, then at least in the `Hashtable_foreach` to check things are non-NULL …

pcp/PCPDynamicMeter.c
```diff
@@ -0,0 +1,368 @@
...
+         CRT_fatalError(note);
+      }
+   } else {
+      /* this is a property of a dynamic metric - the metric expression */
+      /* may not have been observed yet - i.e. we allow for any ordering */
+      metric = PCPDynamicMeter_lookupMetric(meters, meter, key);
+      if (String_eq(p, "color")) {
+         if (String_eq(value, "gray"))
+             metric->color = DYNAMIC_GRAY;
+	 else if (String_eq(value, "darkgray"))
```
BenBE, line 83: Indentation

natoscott: ah, accidental tab - fixed thanks.

DynamicMeter.c
natoscott: Shouldn't ever happen but added this one to be on the safe side.

DynamicMeter.h
```diff
@@ -0,0 +1,25 @@
+#ifndef HEADER_DynamicMeter
+#define HEADER_DynamicMeter
+
+#include "Meter.h"
```
BenBE, line 4: Missing headers for `Hashtable` and `ProcessList`

pcp/PCPDynamicMeter.c
```diff
@@ -0,0 +1,368 @@
...
+         if (!isalnum(*p) && *p != '_')
+            break;
+      }
+      p++;
+   }
+   if (*p != '\0') { /* badness */
+      char note[1024];
+      xSnprintf(note, sizeof(note),
+                "Invalid meter name at %s line %u\n\"%s\"",
+                path, line, key);
```
BenBE, line 137: May crash htop on overlong meter names (from config file?)

pcp/PCPDynamicMeter.h
```diff
@@ -0,0 +1,36 @@
+#ifndef HEADER_PCPDynamicMeter
+#define HEADER_PCPDynamicMeter
+
+#include "CRT.h"
+#include "DynamicMeter.h"
```
BenBE, line 5: Missing `Hashtable.h` and `RichString.h`

BenBE (changes requested):
Local changes pending upload; have fixes for some of those locally already.

pcp/PCPDynamicMeter.c
```diff
@@ -0,0 +1,370 @@
...
+      if (!Metric_values(metric->id, &atom, 1, desc->type))
+         continue;
+      nodata = 0;
+
+      if (i > 0)
+         RichString_appendnAscii(out, CRT_colors[metric->color], " ", 1);
+
+      if (metric->label) {
+         len = xSnprintf(buffer, sizeof(buffer), "%s ", metric->label);
+         RichString_appendnAscii(out, CRT_colors[metric->color], buffer, len);
```
smalinux, line 334: I think the color here should be fixed to match the rest of the htop theme, I suggest this instead:
```
RichString_appendnAscii(out, CRT_colors[METER_TEXT], buffer, len);
```

pcp/PCPDynamicMeter.c
natoscott: It'd have to be either the file path that was super long I guess, or the libpcp-supplied error message.  We're about to call CRT_fatalError here, so its mainly a matter of how we fail.  Looking into the xSnprintf internals, we're not going to return from that if we would overrun the buffer - its going to call _exit internally - so there's probably not anything meaningful we can do here.

pcp/PCPDynamicMeter.c
natoscott: Yep, agreed - will fix up, thanks @smalinux

pcp/PCPDynamicMeter.c
BenBE: We could use `xAsprintf` here and avoid the whole buffer length issue. Just need to make sure the returned string is freed subsequently.

natoscott: OK, will make that change.  There is no 'subsequently' though - the string is passed to CRT_fatalError.

pcp/PCPDynamicMeter.c
```diff
@@ -97,7 +99,9 @@ static void PCPDynamicMeter_parseMetric(PCPDynamicMeters* meters, PCPDynamicMete
          else if (String_eq(value, "white"))
              metric->color = DYNAMIC_WHITE;
       } else if (String_eq(p, "label")) {
-         free_and_xStrdup(&metric->label, value);
+         char* label = String_cat(value, ": ");
+         free_and_xStrdup(&metric->label, label);
+	 free(label);
```
BenBE, line 104: Indentation

pcp/PCPDynamicMeter.c
```diff
@@ -65,13 +66,14 @@ static void PCPDynamicMeter_parseMetric(PCPDynamicMeters* meters, PCPDynamicMete
       /* use derived metrics in dynamic meters for simplicity */
       char* error;
       if (pmRegisterDerivedMetric(metric->name, value, &error) < 0) {
-         char note[1024];
-         xSnprintf(note, sizeof(note),
+         char* note;
+         xAsprintf(&note,
                    "failed to parse expression in %s at line %u\n%s\n%s",
                    path, line, error, pmGetProgname());
```
BenBE: Customarily the program name is prefixed in an error message. So what about this?
```suggestion
                   "%s: Failed to parse expression in %s at line %u\n%s\n",
                   pmGetProgname(), path, line, error);
```

pcp/PCPDynamicMeter.c
```diff
@@ -182,7 +188,9 @@ static void PCPDynamicMeter_parseFile(PCPDynamicMeters* meters, const char* path
          PCPDynamicMeter_validateMeterName(key+1, path, lineno);
          meter = PCPDynamicMeter_new(meters, key+1);
       } else if (value && String_eq(key, "caption")) {
-         free_and_xStrdup(&meter->super.caption, value);
+         char* caption = String_cat(value, ": ");
+         free_and_xStrdup(&meter->super.caption, caption);
+	 free(caption);
```
BenBE, line 193: Indentation

BenBE (approved):

Issue #667, opened 2021-06-22
Darwin 9.8.0: Missing AT_FDCWD and AT_SYMLINK_NOFOLLOW


polluks:
````
make  all-am
depbase=`echo Compat.o | sed 's|[^/]*$|.deps/&|;s|\.o$||'`;\
        gcc -std=gnu99 -DHAVE_CONFIG_H -I.  -DNDEBUG  -Wall -Wcast-align -Wcast-qual -Wextra -Wfloat-equal -Wformat=2 -Winit-self -Wmissing-format-attribute -Wmissing-noreturn -Wmissing-prototypes -Wpointer-arith -Wshadow -Wstrict-prototypes -Wundef -Wunused -Wwrite-strings -pedantic -std=c99 -D_XOPEN_SOURCE_EXTENDED -DSYSCONFDIR="\"/usr/local/etc\"" -I"./darwin" -g -O2 -MT Compat.o -MD -MP -MF $depbase.Tpo -c -o Compat.o Compat.c &&\
        mv -f $depbase.Tpo $depbase.Po
In file included from Compat.c:18:
XUtils.h:24: warning: 'alloc_size' attribute directive ignored
XUtils.h:26: warning: 'alloc_size' attribute directive ignored
XUtils.h:28: warning: 'alloc_size' attribute directive ignored
XUtils.h:30: warning: 'alloc_size' attribute directive ignored
XUtils.h:32: warning: 'alloc_size' attribute directive ignored
Compat.c: In function 'Compat_faccessat':
Compat.c:40: error: 'AT_FDCWD' undeclared (first use in this function)
Compat.c:40: error: (Each undeclared identifier is reported only once
Compat.c:40: error: for each function it appears in.)
Compat.c: In function 'Compat_fstatat':
Compat.c:75: error: 'AT_SYMLINK_NOFOLLOW' undeclared (first use in this function)
make[1]: *** [Compat.o] Error 1
````

fasterit:
Current master/head compiles on recent Darwin/MacOS.
Darwin 9.8.0 is the version that supports MacOS 10.5.8 and that is ~5 years old, so not supported.

PR #670, opened 2021-06-27
Check for alloc_size attribute in configure.ac

### LINKED ISSUE #668: `alloc_size` requires GCC 4.3
As referenced in #667, you get a warning about an unknown attribute `alloc_size` when compiling `Compat.c`, because the function attributes for the allocators are only defined for GCC 4.3 and above:
https://github.com/htop-dev/htop/blob/07170aee4cb3bbe0efadd6c60c50d421583a9b6d/Macros.h#L40-L41

BenBE:
Fixes #668

### DIFF
--- a/Macros.h
+++ b/Macros.h
@@ -37,8 +37,6 @@
 #define ATTR_NONNULL                    __attribute__((nonnull))
 #define ATTR_NORETURN                   __attribute__((noreturn))
 #define ATTR_UNUSED                     __attribute__((unused))
-#define ATTR_ALLOC_SIZE1(a)             __attribute__((alloc_size (a)))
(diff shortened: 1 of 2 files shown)


BenBE:
@polluks Can you please have a look if those warnings you saw in #667 (besides the undefined symbols errors) when compiling `Compat.c` are still present or if they are fixed by this patch?

polluks:
@BenBE Still the same. I suppose you have to check the compiler version.

BenBE:
Can you provide the `config.log`? Really wondering why the `configure` isn't picking up that warning (there's a compiler check in configure that should catch this case).

polluks:
Oops, I have to do first a `git pull origin pull/670/head`.

PR #679, opened 2021-07-07
Use PATH lookup for systemctl in systemd meter

solson:
Before this change, the systemd meter was mysteriously broken on NixOS. I figured out the problem by strace'ing htop which revealved the hardcoded `/bin/systemctl` execve. NixOS has systemctl in PATH, but not at `/bin/systemctl`, so this change makes it work on all my NixOS machines.


BenBE (approved):
LGTM.

PR #680, opened 2021-07-08
Tweak README as we no longer provide release tarball downloads containing a configure script.

natoscott:


BenBE:
LGTM.

What about #639?

natoscott:
> LGTM.
> 
> What about #639?

Oh sure - that's more detailed than this PR so let's go with that then.

PR #688, opened 2021-07-13
netbsd: Support building with libcurses

### LINKED ISSUE #660: NetBSD: Allow for building with X/Open Curses
In particular getting it to build with X/Open Curses seems difficult - it is peeking into ncurses-specific structs.

_Originally posted by @alarixbsd in https://github.com/htop-dev/htop/issues/603#issuecomment-862979241_

AlarixBSD:
Right now Unicode support must be disabled, because htop peeks
into the ncurses cchar_t struct with Unicode enabled. NetBSD's cchar_t
has different contents.

Partially fixes #660

### DIFF
--- a/CRT.c
+++ b/CRT.c
@@ -885,6 +885,7 @@ void CRT_init(const Settings* settings, bool allowUnicode) {
    sigaction (SIGSYS, &act, &old_sig_handler[SIGSYS]);
    sigaction (SIGABRT, &act, &old_sig_handler[SIGABRT]);
 
+   signal(SIGINT, CRT_handleSIGTERM);
    signal(SIGTERM, CRT_handleSIGTERM);
    signal(SIGQUIT, CRT_handleSIGTERM);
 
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -329,8 +329,9 @@ else
     HTOP_CHECK_SCRIPT([ncurses], [wnoutrefresh], [HAVE_LIBNCURSES], [ncurses5-config],
      HTOP_CHECK_LIB([ncurses6],  [doupdate], [HAVE_LIBNCURSES],
       HTOP_CHECK_LIB([ncurses],  [doupdate], [HAVE_LIBNCURSES],
-       AC_MSG_ERROR([can not find required library libncurses])
-   ))))
+       HTOP_CHECK_LIB([curses],  [doupdate], [HAVE_LIBNCURSES],
+        AC_MSG_ERROR([can not find required curses library])
```
BenBE, line 333:
```suggestion
        AC_MSG_ERROR([can not find required curses/ncurses library])
```

CRT.c
```diff
@@ -885,6 +885,7 @@ void CRT_init(const Settings* settings, bool allowUnicode) {
    sigaction (SIGSYS, &act, &old_sig_handler[SIGSYS]);
    sigaction (SIGABRT, &act, &old_sig_handler[SIGABRT]);
 
+   signal(SIGINT, CRT_handleSIGTERM);
```
BenBE, line 888: Why this additional handler?

AlarixBSD: ncurses installs its own signal handler for SIGINT, more traditional curses implementations don't.

BenBE:
I split the SIGINT handler into its own commit, as that one affects all platforms. Actually the behavior on Ctrl+C still isn't optimal (maybe best should perform a regular quit, but without saving anything), but that's best left for another PR.

PR #692, opened 2021-07-14
netbsd: Support curses libraries without ncurses mouse support

AlarixBSD:
This adds a configure check for the ncurses getmouse() function
and disables mouse-related code paths when mouse support is
not present in the curses library.

This is necessary for stable versions of NetBSD's libcurses, the
development version has stub mouse functions for compatibility
with ncurses.

### DIFF
--- a/CRT.c
+++ b/CRT.c
@@ -836,7 +836,9 @@ void CRT_init(const Settings* settings, bool allowUnicode) {
    nonl();
    intrflush(stdscr, false);
    keypad(stdscr, true);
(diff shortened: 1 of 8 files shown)


BenBE:
IMHO this also should disable the mouse support related setting and related code for that setting.

BenBE (commented):
Isn't that still missing the field `enableMouse` in the `Settings` structure?

BenBE (approved):

PR #693, opened 2021-07-14
netbsd: Support display of CPU frequency

AlarixBSD:


netbsd/NetBSDProcessList.c
```diff
@@ -38,6 +39,22 @@ static long fscale;
...
+#if defined(__powerpc__)
+   "machdep.intrepid.frequency.current",
+#endif
+#if defined(__mips__)
+   "machdep.loongson.frequency.current",
+#endif
+#if defined(__i386__) || defined(__x86_64__)
+   "machdep.est.frequency.current",
+   "machdep.powernow.frequency.current",
+#endif
```
BenBE, line 52: Any particular reason to exclude those on non-matching platforms?

BenBE:
Feel free to squash those two commits.

netbsd/NetBSDProcessList.c
AlarixBSD: Basically a micro-optimization to avoid a couple of pointless system calls, otherwise the #ifdefs just provide context to what those variables are for.

BenBE: As those are only fallbacks for when the newer `sysctl`s can't be found, I don't see much value in keeping those. In particular as you might want to save that you skipped the newer entry point and thus also just could save the index of the `sysctl` which finally worked out.

netbsd/NetBSDProcessList.c
```diff
@@ -361,12 +378,58 @@ static void NetBSDProcessList_scanCPUTime(NetBSDProcessList* this) {
...
+   int freq = 0;
+   size_t freqSize = sizeof(freq);
+
+   for (unsigned int i = 0; i <= cpus; i++) {
+         this->cpus[i].frequency = NAN;
+   }
+
+   /* newer hardware supports per-core frequency, for e.g. ARM big.LITTLE */
+   for (unsigned int i = 0; i <= cpus; i++) {
+      snprintf(name, sizeof(name), "machdep.cpufreq.cpu%u.current", i);
```
BenBE, line 394: `xSnprintf` for proper error checking.

netbsd/NetBSDProcessList.c
```diff
@@ -361,12 +378,58 @@ static void NetBSDProcessList_scanCPUTime(NetBSDProcessList* this) {
...
+      if (sysctlbyname(name, &freq, &freqSize, NULL, 0) != -1) {
```
BenBE, line 395: Doesn't need `freqsize` a re-initialization before each call?

netbsd/NetBSDProcessList.c
```diff
@@ -361,12 +378,58 @@ static void NetBSDProcessList_scanCPUTime(NetBSDProcessList* this) {
...
+   if (match) {
+      return;
+   }
+
+   /*
+    * Iterate through legacy sysctl nodes for single-core frequency until
+    * we find a match...
+    */
+   for (const char** s = freqSysctls; *s != NULL; ++s) {
+      if (sysctlbyname(*s, &freq, &freqSize, NULL, 0) != -1) {
```
BenBE, line 410: dito.

netbsd/NetBSDProcessList.c
AlarixBSD: The code needs to be kept because the newer variables are only used on ARM currently. I'll tidy it up though.

netbsd/NetBSDProcessList.c
AlarixBSD: Hmm, maybe, I doubt the kernel will ever return a different size than the initial oldlenp. But we can change it just to be safe.

netbsd/NetBSDProcessList.c
```diff
@@ -361,12 +372,60 @@ static void NetBSDProcessList_scanCPUTime(NetBSDProcessList* this) {
...
 
+static void NetBSDProcessList_scanCPUFrequency(NetBSDProcessList* this) {
+   unsigned int cpus = this->super.cpuCount;
+   bool match = false;
+   char name[64];
+   int freq = 0;
+   size_t freqSize;
+
+   for (unsigned int i = 0; i <= cpus; i++) {
+         this->cpus[i].frequency = NAN;
```
BenBE, line 383: Indentation

netbsd/NetBSDProcessList.c
BenBE: I didn't mean to "drop those `sysctl`s", but to drop the `#ifdef`s.

Another idea was to "remember" which you last used, so you won't have to search for it every time.

BenBE (approved):

PR #697, opened 2021-07-16
Introduce versioned config files and config_reader_min_version

fasterit:


Settings.c
```diff
@@ -144,7 +144,15 @@ static bool Settings_read(Settings* this, const char* fileName, unsigned int ini
          String_freeArray(option);
          continue;
       }
-      if (String_eq(option[0], "fields")) {
+      if (String_eq(option[0], "config_reader_min_version")) {
+         this->config_version = atoi(option[1]);
+         if (this->config_version > CONFIG_READER_MIN_VERSION) {
+             // the version of the config file on disk is newer than what we can read
+             fprintf(stderr, "WARNING: The config file at %s is a newer format (v.%d) than what this version of htop can read (v.%d).\n", fileName, this->config_version, CONFIG_READER_MIN_VERSION);
```
BenBE, line 151:
```suggestion
             fprintf(stderr, "WARNING: The config file %s requires support for a newer format (v%d) than what this version of htop can read (v%d).\n", fileName, this->config_version, CONFIG_READER_MIN_VERSION);
```

fasterit: reads better, updated, thanks

Explorer09:
I'm not sure if it is a good idea to embed a version number in the config file. Did anyone think of a good use case for this? IMHO, the config file should not be versioned, but htop should just ignore any unsupported fields.

BenBE:
I don't think the exact htop version is of interest, but the parser version that produced this file is.

Explorer09:
May I ask why do we ever need to check the parser version? It suggests a backward incompatibility to the htop config format but I didn't see any backward incompatibility exist. And it wouldn't make sense to introduce a backward incompatibility to a config file anyway. This version checking PR (pull request) seems to be a solution in search of a problem.

BenBE:
This PR is based on a problem we already face:

One example of this is for example the active meters and columns in the configuration file. At the moment those are stored by referencing internal IDs that have to be kept equal at all costs for new versions of htop to be able to read old configuration files. This leads e.g. to the situation that some columns like CWD, EXE and COMM have IDs from the Linux-/platform-specific range for those IDs as those IDs have originally been implemented for Linux only and have recently been generalized (by PR #641 et.al.) to be platform-agnostic. The better way would be to have string identifiers for those and have htop map those to the correct (internal) IDs.

This might be a minor inconvenience now, but with the introduction of dynamic meters and dynamic columns in PRs #707 et.al. this only worsens the issue: The internal ID now depends on the order that the config files for those columns are read on start-up. So the only way here is to actually store some name that can be resolved by htop when processing the config file.

Also the old config format does not really allow to provide parameters to meters and columns, which is e.g. necessary for some meters like the Network I/O interface meter from #234.

The idea behind this PR is not to magically make these new features work on old versions of htop (which wont work), but to allow old versions to detect this configuration downgrade. Also having this feature allows for updating the configuration in situations where field meanings subtly changed. For example the column IDs are stored off-by-one in current configuration files. With the version field available we could at some point decide that versions conforming to the new version would store those IDs as-as without that additional addition.

That's also why I remarked thatI remarked that the config file should not only include the minimum version needed to process the configuration file, but also some current version to denote possibly non-breaking changes (like additional fields, that some older-version parser could ignore without too much hassle (except for dropping their values).

Explorer09:
@BenBE  Thanks for information, but I'm personally not convinced. For most of the problem you mentioned, I think something like "feature flags" or "tags" would be a better approach. My idea is to let forks of htop benefit too - so that they can make local, incompatible changes to the config file without messing up with the "versioning".

As an example, for the "column ID" off-by-one problem you mentioned, how about: storing the new IDs in the key name called `fields_v2` in contrast to the old name `fields`? That way the old htop can detect the breaking changes but we won't need to "version" the entire config file.

fasterit:
Too complicated, @Explorer09 . htop will continue to ignore fields it doesn't know about.
So if you run a local patched version, feel free to add `X-explorer9-` or so prefixes to your config lines, we won't confidentially hit these.

fasterit:
> I don't think the exact htop version is of interest, but the parser version that produced this file is.

The reason why I put the htop version in is exactly that, we know parser or logic changes in between our versions.
It doesn't help if the parser writing is v2 but somebody changed the interpretation of a flag elsewhere in the code.
So recording the htop version is more easy **and** more useful at the same time.

BenBE (approved):

Settings.h
```diff
@@ -26,6 +28,7 @@ typedef struct {
 
 typedef struct Settings_ {
    char* filename;
+   int config_version;
```
cgzones, line 31: This variable is never read from.

cgzones (approved):
LGTM

Settings.h
fasterit: It is set and made available when we have the next incompatible change to fence that one.

Settings.c
```diff
@@ -166,7 +166,15 @@ static bool Settings_read(Settings* this, const char* fileName, unsigned int ini
          String_freeArray(option);
          continue;
       }
-      if (String_eq(option[0], "fields")) {
+      if (String_eq(option[0], "config_reader_min_version")) {
+         this->config_version = atoi(option[1]);
+         if (this->config_version > CONFIG_READER_MIN_VERSION) {
+             // the version of the config file on disk is newer than what we can read
+             fprintf(stderr, "WARNING: The config file %s requires support for a newer format (v%d) than what this version of htop can read (v%d).\n", fileName, this->config_version, CONFIG_READER_MIN_VERSION);
+             fprintf(stderr, "         It will be overwritten when this version of htop exits.\n");
```
natoscott, line 174: Please use PACKAGE here instead of "htop".  The WARNING line is longer than 80 chars too - might be worth wrapping it earlier, and spreading this note across three lines instead of two?
Also it may make the message slightly more clear perhaps using this phrasing ... ?

```
fprintf(stderr, "WARNING: %s specifies configuration format version v%d, but this %s binary supports up to v%d.", fileName, this->config_version, PACKAGE, CONFIG_READER_MIN_VERSION);
fprintf(stderr, "         The configuration version will be downgraded to v%d when %s exits.\n", CONFIG_READER_MIN_VERSION, PACKAGE);
```

Settings.h
natoscott, line 31: "unsigned int" here?  or size_t?  :)

natoscott (approved):
Minor suggestions only, LGTM either way.

Settings.c
fasterit: Committed in 35d94a5 . Thank you very much for the improvement.

Settings.h
fasterit: No config file reader is worth it, if it cannot handle -1 :-)

Settings.c
```diff
@@ -169,10 +169,10 @@ static bool Settings_read(Settings* this, const char* fileName, unsigned int ini
       if (String_eq(option[0], "config_reader_min_version")) {
          this->config_version = atoi(option[1]);
          if (this->config_version > CONFIG_READER_MIN_VERSION) {
-             // the version of the config file on disk is newer than what we can read
-             fprintf(stderr, "WARNING: The config file %s requires support for a newer format (v%d) than what this version of htop can read (v%d).\n", fileName, this->config_version, CONFIG_READER_MIN_VERSION);
-             fprintf(stderr, "         It will be overwritten when this version of htop exits.\n");
-             return false;
+            // the version of the config file on disk is newer than what we can read
+            fprintf(stderr, "WARNING: %s specifies configuration format version v%d, but this %s binary supports up to v%d.", fileName, this->config_version, PACKAGE, CONFIG_READER_MIN_VERSION);
+            fprintf(stderr, "         The configuration version will be downgraded to v%d when %s exits.\n", CONFIG_READER_MIN_VERSION, PACKAGE);
```
Explorer09, line 174: Sorry for commenting too late, but I think it might be better if htop can stop (and quit) when the config version is found out to be too new.
Or make a backup (with filename like htoprc.backup) before writing to the file.
My idea is not to destroy the user's config file as it might be working for him. And he might not be aware that the config file is incompatible before we alert him.

Explorer09: P.S. Remember that users can't see the stderr message unless htop halts or crashes, so it is more likely that config file gets downgraded/overwritten before user can see the warning and do something with it.

fasterit: Yes, and that is fine. If people run downgraded versions of htop in parallel they will be able to re-create a config file from backup or by just manually setting what they wanted again.

If you find a .backup file, do you overwrite this? Do you do .backup.YYMMDDHHSSmm and fill up the hard disk? This is a rabbit hole. This file is for consumption by `htop`. The fact that it can be edited by hand is just a convenience to the developers and power users that don't complain :-).

Explorer09: @fasterit You know that's one of the reasons I was skeptical of this PR. And no, I don't think people can re-create a config file from scratch that easily.

Just to tell my personal story: I had a painful experience when Firefox (the web browser) kills off my session (with all opened tabs) when it performs an update. I have a habit of leaving many tabs open in order to maintain the information I needed for work. Although Firefox does backup sessions every time it starts (which is good), it's update system is not always reliable of keeping my sessions anyway, and occasionally I lost them. It takes the hell a lot of time to lookup my browsing history and recover which web pages I had opened before. Now I always manually backup my session before a Firefox update, just to get rid of the unreliable update process.

> If you find a .backup file, do you overwrite this? Do you do .backup.YYMMDDHHSSmm and fill up the hard disk?

Don't overwrite the backup. That's what the backup is for. And Firefox does include a date for it's session backups (for Firefox upgrades; it looks like "upgrade.jsonlz4-YYYYMMDDhhmmss").

fasterit: @Explorer09 **Firefox**: Right-click on any of the tabs, choose "Select All Tabs" from the context menu, then right-click on any tab again and select "Bookmark Tabs". This will create a folder of bookmarks for you that you can restore at any time.

**htop**: `cp ~/.config/htop/htoprc{,_$(date +%y%m%d%H%M%S)}`
Put it into your `~/.bashrc` if you want to.

Explorer09: @fasterit For Firefox, No, bookmarks and tabs are different use cases and shouldn't be mixed. I consider bookmarks for long term references and tabs for short term. I may not leave a tab open forever, but I need it open when my job is half done and have to continue it tomorrow. This is what I mean by "short term" and saving the tabs as bookmarks won't fit my workflow.

PR #698, opened 2021-07-17
Respect "Show custom thread names" setting

### LINKED ISSUE #615: Option `Show custom thread names` inconsistent
Starting with  `Show custom thread names` disabled:
    `sd-resolve|/lib/systemd/systemd-timesyncd`
Enable `Show custom thread names`:
    `sd-resolve`
Disable `Show custom thread names`:
    `sd-resolve`

Might be fixed together with #388?

BenBE:
Fixes #615


cgzones:
This patch broke highlighting for *real* processes, as the setting `showThreadNames` should only affect userland threads.

PR #685, opened 2021-07-12
Dynamic meter updates

natoscott:

PR #707, opened 2021-07-23
PCP: Dynamic columns

smalinux:
Implements support for arbitrary Performance Co-Pilot
metrics with per-process instance domains to form new
htop columns.  The column-to-metric mappings are setup
using configuration files which will be documented via
man pages as part of a follow-up commit.

We provide an initial set of column configurations so
as to provide new capabilities to pcp-htop: including
configs for containers, open fd counts, scheduler run
queue time, tcp/udp bytes/calls sent/recv, delay acct,
virtual machine guests, detailed virtual memory, swap.

Note there is a change to the configuration file path
resolution algorithm introduced for 'dynamic meters'.
First, look in any custom PCP_HTOP_DIR location.  Then
iterate, in priority order, users home directory, then
local sysadmins files in /etc/pcp/htop, then readonly
configuration files below /usr/share/pcp/htop.  This
final location becomes the preferred place for our own
shipped meter and column files.

The Settings file (htoprc) writing code is updated to
not using the numeric identifier for dynamic columns.
The same strategy used for dynamic meters is used here
where we write Dynamic(name) so the name can be setup
once more at start.  Regular (static) columns writing
to htoprc - i.e. numerically indexed - is unchanged.


AvailableColumnsPanel.c
```diff
@@ -29,6 +38,15 @@ static void AvailableColumnsPanel_delete(Object* object) {
...
 
+
+/*
+void AvailableColumnsPanel_insertDynamicColumnName(AvailableColumnsPanel* this, int at, int key, const ProcessList* pl) {
+   int param = abs(key - LAST_STATIC_PROCESSFIELD);
+   const DynamicColumn* column = Hashtable_get(pl->dynamicColumns, param);
+   //Panel_insert(this->columns, at, (Object*) ListItem_new("lol", 1));
+}
+*/
+
```
BenBE, line 49: Avoid dead/commented-out code

ColumnsPanel.c
```diff
@@ -115,7 +116,20 @@ const PanelClass ColumnsPanel_class = {
...
+typedef struct {
+   Panel* super;
+   unsigned int id;
+   unsigned int offset;
+} DynamicIterator;
+
+static void ColumnsPanel_addDynamicColumnCaption(Panel* super, int param, const ProcessList* pl) {
+   int field = abs(param - LAST_STATIC_PROCESSFIELD);
+   const DynamicColumn* column = Hashtable_get(pl->dynamicColumns, field);
+   if(column)
```
BenBE, line 128:
```suggestion
   if (column)
```

Makefile.am
```diff
@@ -359,25 +361,27 @@ endif
 # --------------------------
 
 pcp_platform_headers = \
+	linux/PressureStallMeter.h \
+	linux/ZramMeter.h \
+	linux/ZramStats.h \
```
BenBE, line 366: Still not quite happy including Linux stuff on PCP explicitly …

Makefile.am
```diff
@@ -359,25 +361,27 @@ endif
...
-	linux/PressureStallMeter.h \
-	linux/ZramMeter.h \
-	linux/ZramStats.h \
 	zfs/ZfsArcMeter.h \
 	zfs/ZfsArcStats.h \
 	zfs/ZfsCompressedArcMeter.h
 
 pcp_platform_sources = \
+	linux/PressureStallMeter.c \
+	linux/ZramMeter.c \
```
BenBE, line 379: dito.

Process.c
```diff
@@ -712,203 +713,208 @@ void Process_writeField(const Process* this, RichString* str, ProcessField field
...
+            return;
+         }
 
-      char* buf = buffer;
-      int maxIndent = 0;
-      bool lastItem = (this->indent < 0);
-      int indent = (this->indent < 0 ? -this->indent : this->indent);
+         char* buf = buffer;
+         int maxIndent = 0;
+         bool lastItem = (this->indent < 0);
```
BenBE, line 731: No need for extra parens …

Process.c
```diff
@@ -712,203 +713,208 @@ void Process_writeField(const Process* this, RichString* str, ProcessField field
...
+         int indent = (this->indent < 0 ? -this->indent : this->indent);
```
BenBE, line 732: dito.

pcp/PCPDynamicColumn.c
```diff
@@ -0,0 +1,336 @@
...
+#include "config.h" // IWYU pragma: keep
+
+#include "pcp/PCPDynamicColumn.h"
+
+#include <math.h>
+
+#include "CRT.h"
+#include "Macros.h"
+#include "Object.h"
+#include "PCPProcess.h"
```
BenBE, line 17:
```suggestion
#include "pcp/PCPProcess.h"
```

Plus proper sorting below

pcp/PCPDynamicColumn.c
```diff
@@ -0,0 +1,336 @@
...
+      free(note);
+   }
+
+   while (*p) {
+      if (p == key) {
+         if (!isalpha(*p) && *p != '_')
+            break;
+      } else {
+         if (!isalnum(*p) && *p != '_')
+            break;
```
BenBE, line 135: Maybe also allow ``-``?

pcp/PCPDynamicColumn.c
```diff
@@ -0,0 +1,336 @@
...
+         break;
+      case PM_TYPE_DOUBLE:
+         xSnprintf(buffer, size, "%*.2f", width, pp->dc[index].d);
+         RichString_appendAscii(str, attr, buffer);
+         break;
+      default:
+         RichString_appendAscii(str, CRT_colors[METER_VALUE_ERROR], "no data");
+         break;
+   }
+ }
```
BenBE, line 311: Indentation

pcp/PCPProcess.c
```diff
@@ -186,6 +192,7 @@ static int PCPProcess_compareByKey(const Process* v1, const Process* v2, Process
    const PCPProcess* p1 = (const PCPProcess*)v1;
    const PCPProcess* p2 = (const PCPProcess*)v2;
 
+   if(key < LAST_STATIC_PROCESSFIELD) {
```
BenBE, line 195: Indentation of this whole block …

pcp/PCPProcess.c
```diff
@@ -248,6 +255,8 @@ static int PCPProcess_compareByKey(const Process* v1, const Process* v2, Process
    default:
       return Process_compareByKey_Base(v1, v2, key);
    }
+   } else
+      return PCPDynamicColumn_compareByKey(p1, p2, key);
```
BenBE, line 259: Also note the rule on consistent embracing of blocks …

pcp/PCPProcess.h
```diff
@@ -11,6 +11,7 @@ in the source distribution for its full text.
 
 #include "config.h" // IWYU pragma: keep
 
+#include <pcp/pmapi.h>
```
BenBE, line 14: Sort order …

BenBE (changes requested):
Please fix all those fixme/todo comments.

Also what about having the Custom columns be handled as a sub-call to each platform's WriteColumn function? That way the overall public interface stays minimal.

smalinux:
Just for recording, my friend @a3f gave me a code review [here](https://github.com/smalinux/htop/commit/25af7bad7c8293a930262bc1fb8f1521dfab0002). thank you, @a3f

ProcessList.c
```diff
@@ -82,7 +84,16 @@ void ProcessList_setPanel(ProcessList* this, Panel* panel) {
    this->panel = panel;
 }
 
-static const char* alignedProcessFieldTitle(ProcessField field) {
+static void alignedDynamicColumnTitle(const ProcessList* this, int field, char buffer[], int size) {
+   int key = (int)field - LAST_STATIC_PROCESSFIELD;
+   const DynamicColumn* column = Hashtable_get(this->dynamicColumns, key);
+   if (!column)
+      return;
```
BenBE, line 91: Buffer uninitialized on non-existing columns.

ProcessList.c
```diff
@@ -576,6 +591,7 @@ Process* ProcessList_getProcess(ProcessList* this, pid_t pid, bool* preExisting,
    return proc;
 }
 
+// LOOKATME
```
BenBE, line 594: Please do so and cleanup before final revision … ;-)

ProcessList.h
```diff
@@ -51,6 +51,7 @@ typedef struct ProcessList_ {
    Hashtable* draftingTreeSet;
 
    Hashtable* dynamicMeters;  /* runtime-discovered meters */
+   Hashtable* dynamicColumns;  /* runtime-discovered Columns */
```
BenBE, line 54: Alignment of those comments seems odd.

netbsd/NetBSDProcessList.c
```diff
@@ -40,10 +40,16 @@ static int pageSize;
...
+#if defined(__powerpc__)
    "machdep.intrepid.frequency.current",
+#endif
+#if defined(__mips__)
    "machdep.loongson.frequency.current",
+#endif
+#if defined(__i386__) || defined(__x86_64__)
+   "machdep.est.frequency.current",
+   "machdep.powernow.frequency.current",
+#endif
```
BenBE, line 52: This artifact seems odd to be part of the PR …

pcp/PCPDynamicColumn.c
```diff
@@ -0,0 +1,330 @@
...
+      metric = &column->metrics[i];
+      if (String_eq(metric->name, metricName)) {
+         free(metricName);
+         return metric;
+      }
+   }
+
+   /* not an existing metric in this column - add it */
+   unsigned int n = column->totalMetrics + 1;
+   column->metrics = xReallocArray(column->metrics, n, sizeof(PCPDynamicColumnMetric));
```
BenBE, line 50: Beware of possibly quadratic behavior on repeated calls of this function.

pcp/PCPDynamicColumn.c
```diff
@@ -0,0 +1,330 @@
...
+   }
+}
+
+// Ensure a valid name for use in a PCP metric name and in htoprc
+static void PCPDynamicColumn_validateColumnName(char* key, const char* path, unsigned int line) {
+   char* p = key;
+   char* end = strrchr(key, ']');
+
+   if (end) {
+      *end = '\0';
```
BenBE, line 127: Beware of const correctness on caller-side of this function.

pcp/PCPDynamicColumn.c
```diff
@@ -0,0 +1,330 @@
...
+      } else {
+         if (!isalnum(*p) && *p != '_')
+            break;
+      }
+      p++;
+   }
+   if (*p != '\0') { /* badness */
+      char* note;
+      xAsprintf(&note,
+                "%s: invalid column....... name at %s line %u\n\"%s\"",
```
BenBE, line 151: Better message/less dots?

pcp/PCPDynamicColumn.c
```diff
@@ -0,0 +1,330 @@
...
+   for (;;) {
+      char* line = String_readLine(file);
+      if (!line)
+         break;
+      lineno++;
+
+      /* cleanup whitespace, skip comment lines */
+      char* trimmed = String_trim(line);
+      free(line);
+      if (trimmed[0] == '#' || trimmed[0] == '\0') {
```
BenBE, line 184:
```suggestion
      if (!trimmed || !trimmed[0] || trimmed[0] == '#') {
```

pcp/PCPDynamicColumn.c
```diff
@@ -0,0 +1,330 @@
...
+
+      size_t n;
+      char** config = String_split(trimmed, '=', &n);
+      free(trimmed);
+      if (config == NULL)
+         continue;
+
+      char* key = String_trim(config[0]);
+      char* value = n > 1 ? String_trim(config[1]) : NULL;
+      if (key[0] == '[') {  /* new section heading - i.e. new column */
```
BenBE, line 197: Shouldn't the heading check for `[` appear before trying to split the line at `=`?

pcp/PCPDynamicColumn.c
```diff
@@ -0,0 +1,330 @@
...
+   char* path;
+
+   columns->table = Hashtable_new(0, true);
+
+   /* search in the users home directory first of all */
+   if (xdgConfigHome) {
+      path = String_cat(xdgConfigHome, "/htop/columns/");
+   } else {
+      if (!home)
+         home = "";
```
BenBE, line 248: This effectively tries to read from the filesystem root, which is unlikely what's intended …

pcp/PCPDynamicColumn.c
```diff
@@ -0,0 +1,330 @@
...
+   PCPDynamicColumn_scanDir(columns, path);
+   free(path);
+
+   /* check the working directory, as a final option */
+   char cwd[PATH_MAX];
+   if (getcwd(cwd, sizeof(cwd)) != NULL) {
+      path = String_cat(cwd, "/pcp/columns/");
+      PCPDynamicColumn_scanDir(columns, path);
+      free(path);
+   }
```
BenBE, line 265: Given that the display expression may trigger arbitrarily complex calculations, this is possibly not a good idea IMHO.

pcp/PCPDynamicColumn.c
```diff
@@ -0,0 +1,330 @@
...
+}
+
+void PCPDynamicColumn_writeField(PCPDynamicColumn* this, const Process* proc, RichString* str, int field) {
+   const PCPProcess* pp = (const PCPProcess*) proc;
+   char buffer[30];
+   size_t size = sizeof(buffer);
+   int width = (abs(this->super.width) < 28 && this->super.width) ? this->super.width : -12;
```
BenBE, line 272: Switch order of conditions for consistency with the other location with this check in place.
```suggestion
   int width = (this->super.width && abs(this->super.width) < 28) ? this->super.width : -12;
```

pcp/PCPProcessList.h
```diff
@@ -56,14 +56,14 @@ typedef enum CPUMetric_ {
...
 typedef struct PCPProcessList_ {
    ProcessList super;
-   double timestamp;		/* previous sample timestamp */
-   pmAtomValue* cpu;		/* aggregate values for each metric */
-   pmAtomValue** percpu;	/* per-processor values for each metric */
-   pmAtomValue* values;		/* per-processor buffer for just one metric */
+   double timestamp;    /* previous sample timestamp */
+   pmAtomValue* cpu;    /* aggregate values for each metric */
+   pmAtomValue** percpu;   /* per-processor values for each metric */
+   pmAtomValue* values;    /* per-processor buffer for just one metric */
```
BenBE, line 62: Alignment of comments seems all over the place.

pcp/Platform.c
```diff
@@ -50,27 +51,33 @@ in the source distribution for its full text.
...
+   pmID* fetch;         /* enabled identifiers for sampling */
+   pmDesc* descs;    /* metric desc array indexed by Metric */
+   pmResult* result;    /* sample values result indexed by Metric */
+   PCPDynamicMeters meters;   /* dynamic meters via configuration files */
+   PCPDynamicColumns columns;    /* dynamic columns via configuration files */
+   struct timeval offset;  /* time offset used in archive mode only */
+   long long btime;     /* boottime in seconds since the epoch */
+   char* release;    /* uname and distro from this context */
+   int pidmax;       /* maximum platform process identifier */
+   int ncpu;         /* maximum processor count configured */
```
BenBE, line 73: Alignment of comments seems all over the place.

pcp/Platform.c
```diff
@@ -350,6 +362,7 @@ bool Metric_iterate(Metric metric, int* instp, int* offsetp) {
 
    *offsetp = offset;
    *instp = vset->vlist[offset].inst;
+   //fprintf(stderr, "offset %d PID %d\n", *offsetp, *instp);
```
BenBE, line 365: Debug code …

pcp/columns/cmd
```diff
@@ -0,0 +1,11 @@
+#
+# pcp-htop(1) configuration file - see pcp-htop(5)
+#
```
BenBE, line 3:
```suggestion
# pcp-htop(1) configuration file - see pcp-htop(5)
```
One comment line at the top oughta be enough …

BenBE (changes requested):

pcp/PCPDynamicColumn.c
```diff
@@ -5,20 +5,27 @@ htop - PCPDynamicColumn.c
...
 #include "config.h" // IWYU pragma: keep
 
 #include "pcp/PCPDynamicColumn.h"
 
-#include <math.h>
+#include <ctype.h>
+#include <dirent.h>
+#include <errno.h>
+#include <limits.h>
+#include <pcp/pmapi.h>
```
BenBE, line 17: Ordered as subdir …

BenBE (changes requested):

DynamicColumn.c
```diff
@@ -0,0 +1,50 @@
+/*
+htop - DynamicColumn.c
+(C) 2021 htop dev team
+(C) 2021 Red Hat, Inc.  All Rights Reserved.
```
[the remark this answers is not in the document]
smalinux: @BenBE any suggestions, please

netbsd/NetBSDProcessList.c
smalinux: @BenBE I think this happened accidentally while trying to resolve rebase conflicts.

pcp/PCPDynamicColumn.c
```diff
@@ -0,0 +1,338 @@
...
+         char* note;
+         xAsprintf(&note,
+                   "%s: failed to parse expression in %s at line %u\n%s\n",
+                   pmGetProgname(), path, line, error);
+         free(error);
+         errno = EINVAL;
+         CRT_fatalError(note);
+         free(note);
+      }
+   } else {
```
BenBE, line 89: Avoid this explicit `else` block, by exiting from the `true` path early.

pcp/PCPProcess.c
```diff
@@ -186,68 +193,71 @@ static int PCPProcess_compareByKey(const Process* v1, const Process* v2, Process
...
-      return SPACESHIP_NUMBER(p1->swapin_delay_percent, p2->swapin_delay_percent);
-   case CTXT:
-      return SPACESHIP_NUMBER(p1->ctxt_diff, p2->ctxt_diff);
-   case SECATTR:
-      return SPACESHIP_NULLSTR(p1->secattr, p2->secattr);
-   default:
-      return Process_compareByKey_Base(v1, v2, key);
-   }
+   if (key < LAST_STATIC_PROCESSFIELD) {
+      switch (key) {
```
BenBE, line 197: Avoid this extra indentation similar to the other large `switch` statement.

BenBE (changes requested):

Process.c
```diff
@@ -706,6 +707,13 @@ void Process_printLeftAlignedField(RichString* str, int attr, const char* conten
    RichString_appendChr(str, attr, ' ', width + 1 - columns);
 }
 
+static inline bool is_dynamicColumn( const ProcessList* pl, int key) {
+   const DynamicColumn* column = Hashtable_get(pl->dynamicColumns, key);
+   if (!column)
+      return 1;
+   return 0;
```
BenBE, line 714: There are constants ``true` and `false` in `stdbool.h` for this.

Process.c
```diff
@@ -726,12 +734,12 @@ void Process_writeField(const Process* this, RichString* str, ProcessField field
...
       int maxIndent = 0;
-      bool lastItem = (this->indent < 0);
-      int indent = (this->indent < 0 ? -this->indent : this->indent);
+      bool lastItem = this->indent < 0;
+      int indent = this->indent < 0 ? -this->indent : this->indent;
 
       for (int i = 0; i < 32; i++) {
          if (indent & (1U << i)) {
-            maxIndent = i + 1;
+            maxIndent = i+1;
```
BenBE, line 742: Accidental revert?

pcp/PCPProcess.c
```diff
@@ -186,68 +193,71 @@ static int PCPProcess_compareByKey(const Process* v1, const Process* v2, Process
...
+      case CTXT:
+         return SPACESHIP_NUMBER(p1->ctxt_diff, p2->ctxt_diff);
+      case SECATTR:
+         return SPACESHIP_NULLSTR(p1->secattr, p2->secattr);
+      default:
+         return Process_compareByKey_Base(v1, v2, key);
    }
+   } else
+      return PCPDynamicColumn_compareByKey(p1, p2, key);
 }
```
BenBE, line 261: Integrate the last check into the default case as with the other large `switch` statement …

Action.c
```diff
@@ -163,14 +164,27 @@ Htop_Reaction Action_setSortKey(Settings* settings, ProcessField sortKey) {
 
 // ----------------------------------------
 
+static inline bool has_dynamicColumns(int key) {
+   return key > LAST_STATIC_PROCESSFIELD;
+}
+
```
BenBE, line 170: `has_` vs. `is_`
 This is duplicated elsewhere too …

ProcessList.c
```diff
@@ -82,7 +84,18 @@ void ProcessList_setPanel(ProcessList* this, Panel* panel) {
    this->panel = panel;
 }
 
-static const char* alignedProcessFieldTitle(ProcessField field) {
+static void alignedDynamicColumnTitle(const ProcessList* this, int field, char buffer[], int size) {
```
BenBE, line 87:
```suggestion
static void alignedDynamicColumnTitle(const ProcessList* this, int field, char* buffer, int size) {
```

netbsd/Platform.h
```diff
@@ -94,6 +94,8 @@ static inline void Platform_gettime_monotonic(uint64_t* msec) {
    Generic_gettime_monotonic(msec);
 }
 
+static inline void Platform_dynamicColumnWriteField(ATTR_UNUSED const Process* proc, ATTR_UNUSED RichString* str, ATTR_UNUSED int param) { }
+
 static inline Hashtable* Platform_dynamicMeters(void) {
    return NULL;
 }
```
BenBE, line 101: Make things consistent with other platforms …

```suggestion
static inline Hashtable* Platform_dynamicMeters(void) { return NULL; }

static inline void Platform_dynamicColumnWriteField(ATTR_UNUSED const Process* proc, ATTR_UNUSED RichString* str, ATTR_UNUSED int param) { }
```

BenBE (changes requested):

pcp/PCPDynamicColumn.c
natoscott: Hyphen (minus) is the syntax used for subtraction in the PCP derived metrics specification, so we must exclude it here when checking name validity.

Makefile.am
natoscott: One alternative would be moving all *Meter.[ch] code into generic/ or a new meters/ ... either could make sense I think.  Might also serve to de-clutter the base htop source directory too I guess.

Either way, this is a topic for a follow-up series of commits.

pcp/PCPDynamicColumn.h
```diff
@@ -0,0 +1,40 @@
...
+   ColorElements color;
+   char* name;
+   char* label;
+   char* suffix;
+} PCPDynamicColumnMetric;
+
+typedef struct {
+   DynamicColumn super;
+   PCPDynamicColumnMetric* metrics;
+   unsigned int totalMetrics;
```
natoscott, line 24: Don't need 'totalMetrics' - there can be only one.

pcp/PCPDynamicMeter.c
```diff
@@ -18,12 +18,13 @@ in the source distribution for its full text.
...
 
-static PCPDynamicMetric* PCPDynamicMeter_lookupMetric(PCPDynamicMeters* meters, PCPDynamicMeter* meter, const char* name) {
+
+static PCPDynamicMeterMetric* PCPDynamicMeter_lookupMetric(PCPDynamicMeters* meters, PCPDynamicMeter* meter, const char* name) {
    size_t bytes = 8 + strlen(meter->super.name) + strlen(name);
    char* metricName = xMalloc(bytes);
    xSnprintf(metricName, bytes, "htop.%s.%s", meter->super.name, name);
 
-   PCPDynamicMetric* metric;
+   PCPDynamicMeterMetric* metric;
```
natoscott, line 27: So we don't need to make this change - PCPDynamicMetric can be used as before.

pcp/PCPProcess.c
```diff
@@ -246,7 +253,10 @@ static int PCPProcess_compareByKey(const Process* v1, const Process* v2, Process
    case SECATTR:
       return SPACESHIP_NULLSTR(p1->secattr, p2->secattr);
    default:
-      return Process_compareByKey_Base(v1, v2, key);
+      if(key > LAST_STATIC_PROCESSFIELD)
```
natoscott, line 256: Missing whitespace there.

pcp/PCPProcess.h
```diff
@@ -61,6 +60,9 @@ typedef struct PCPProcess_ {
    /* Storage data written (in kilobytes) */
    unsigned long long io_write_bytes;
 
+   /* Dynamic Columns */
+   pmAtomValue* dc;
```
natoscott, line 64: Hmm, not loving the name 'dc' - its a bit cryptic for me.  Maybe 'dynamic'?

pcp/Platform.c
```diff
@@ -501,6 +516,9 @@ void Platform_init(void) {
    Metric_enable(PCP_UNAME_MACHINE, true);
    Metric_enable(PCP_UNAME_DISTRO, true);
 
+   for(unsigned int i = pcp->columns.offset; i < pcp->columns.offset + pcp->columns.count; i++)
+      Metric_enable(i, true);
```
natoscott, line 520: Whitespace missing on 'for' - and this isn't a great approach in general.  We really only want to enable sampling of these metrics once they've been selected in the UI or read from htoprc (i.e. disabled by default, enabled only if used).  Otherwise we'll be sampling more and more metrics as more configurations are added, for no gain - only do it if needed.

pcp/columns/cmd
```diff
@@ -0,0 +1,9 @@
+# pcp-htop(1) configuration file - see pcp-htop(5)
```
natoscott, line 1: We shouldn't provide dynamic column configs that mirror static columns.  Remove this file & I'll send through some other realistic samples we should use shortly.

pcp/columns/cmd
```diff
@@ -0,0 +1,9 @@
...
+
+[syscmd]
+caption = myCMD
+width = -17
+description = PCP CMD description
+cmd.metric = proc.psinfo.cmd
```
natoscott, line 7: To clarify something mentioned earlier in the review, we should not have this cmd.metric syntax for columns - just use 'metric'.

pcp/columns/flags
```diff
@@ -0,0 +1,9 @@
+# pcp-htop(1) configuration file - see pcp-htop(5)
```
natoscott, line 1: Remove this file, not a useful DynamicColumn in practice.

pcp/columns/pid
```diff
@@ -0,0 +1,9 @@
+# pcp-htop(1) configuration file - see pcp-htop(5)
```
natoscott, line 1: Remove this file, not a useful DynamicColumn in practice.

pcp/columns/tty_name
```diff
@@ -0,0 +1,9 @@
+# pcp-htop(1) configuration file - see pcp-htop(5)
```
natoscott, line 1: Remove this file, not a useful DynamicColumn in practice - this is already a static column.

pcp/columns/wchan
```diff
@@ -0,0 +1,9 @@
+# pcp-htop(1) configuration file - see pcp-htop(5)
+
+[wchan]
+caption = WCHAN
+width = 8
+description = PCP wchan description
```
natoscott, line 6: Let's improve the description.  Use "pminfo -t proc.psinfo.wchan" output, as a sentence (start with upper case letter).

natoscott (changes requested):
Overall it looks good @smalinux - great work.  I think the htoprc change mentioned elsewhere is a critical next step before this can be merged, and also the dynamic config tweaks I mention in the review.

I'll send through some more practical dynamic column configuration files shortly.

Process.c
```diff
@@ -706,6 +707,11 @@ void Process_printLeftAlignedField(RichString* str, int attr, const char* conten
    RichString_appendChr(str, attr, ' ', width + 1 - columns);
 }
 
+static inline bool is_dynamicColumn( const ProcessList* pl, int key) {
+   const DynamicColumn* column = Hashtable_get(pl->dynamicColumns, key);
+   return !column? true : false;
```
BenBE, line 712: Doesn't this negate the actual check?

Better without the `?:` altogether …
```suggestion
   return !!column;
```

natoscott: Perhaps even more clear as "return column != NULL;" ... ?

BenBE: Fine with both `return !!column;` and `return column != NULL;`, though I prefer the first variant …

Action.c
```diff
@@ -169,18 +169,17 @@ static Htop_Reaction actionSetSortColumn(State* st) {
...
-      if (fields[i] > LAST_STATIC_PROCESSFIELD) {
-         int index = fields[i] - LAST_STATIC_PROCESSFIELD;
-         const DynamicColumn* column = Hashtable_get(st->pl->dynamicColumns, index);
-         if (column) {
-            Panel_add(sortPanel, (Object*) ListItem_new(column->caption, fields[i]));
-         }
+      if (fields[i] >= LAST_STATIC_PROCESSFIELD) {
+         DynamicColumn* column = Hashtable_get(dynamicColumns, fields[i]);
+         if (column)
+         name = column->caption ? xStrdup(column->caption) : xStrdup(column->name);
```
BenBE, line 178: Indentation

Action.c
```diff
@@ -169,18 +169,17 @@ static Htop_Reaction actionSetSortColumn(State* st) {
...
       } else {
          name = String_trim(Process_fields[fields[i]].name);
-         Panel_add(sortPanel, (Object*) ListItem_new(name, fields[i]));
       }
+      Panel_add(sortPanel, (Object*) ListItem_new(name, fields[i]));
```
BenBE, line 182: `name` can be `NULL` if the column is not found. Intentional? Checked that every code using that final list copes with `NULL` values?

AvailableColumnsPanel.c
```diff
@@ -61,14 +56,9 @@ static HandlerResult AvailableColumnsPanel_eventHandler(Panel* super, int ch) {
...
 
-         int key = selected->key;
          int at = Panel_getSelectedIndex(this->columns);
-         if (key > LAST_STATIC_PROCESSFIELD) {
-            AvailableColumnsPanel_insertDynamicColumnName(this, at, key);
-         } else {
-            Panel_insert(this->columns, at, (Object*) ListItem_new(Process_fields[key].name, key));
-         }
-         Panel_setSelected(this->columns, at+1);
+	 AvailableColumnsPanel_insert(this, at, selected->key);
```
BenBE, line 60: Indentation

CommandLine.c
```diff
@@ -386,8 +386,10 @@ int CommandLine_run(const char* name, int argc, char** argv) {
    if (flags.pidMatchList)
       Hashtable_delete(flags.pidMatchList);
 
-   /* Delete Settings last, since it can get accessed in the crash handler */
+   /* Delete these last, since they can get accessed in the crash handler */
    Settings_delete(settings);
+   Hashtable_delete(dc);
+   Hashtable_delete(dm);
```
BenBE, line 392: Maybe "unregister" those objects from the CRT for shutdown …

Hashtable.c
```diff
@@ -120,19 +120,23 @@ Hashtable* Hashtable_new(size_t size, bool owner) {
...
 
-   memset(this->buckets, 0, this->size * sizeof(HashtableItem));
-   this->items = 0;
+   if (this) {
+      memset(this->buckets, 0, this->size * sizeof(HashtableItem));
+      this->items = 0;
+   }
 
    assert(Hashtable_isConsistent(this));
 }
```
BenBE, line 142: Please explain these changes. Hashtable functions are always assumed to be called on live objects, thus `this!=NULL` …

pcp/PCPDynamicColumn.c
BenBE: Okay, that's a chance for cleanup of both locations in another PR then …

BenBE (changes requested):

Hashtable.c
smalinux: @BenBE I got errors from GitHub CI "error: null pointer dereference" [here](https://github.com/natoscott/htop/runs/3179357083)
So I added this last commit to trying to fix that

BenBE: Yep. Indicating that you at one place call it without a proper NULL check ;-)

smalinux: @BenBE I will revert this last commit now to see what will happen. thank you :)

CommandLine.c
```diff
@@ -384,8 +386,10 @@ int CommandLine_run(const char* name, int argc, char** argv) {
...
```
BenBE, line 392: These two calls lack a NULL check. Adding the check should resolve the compile issue.

An alternative would be to always initialize this Hashtable instead of using `NULL` when a platform does not support dynamic columns.

CommandLine.c
natoscott: Like Settings, we want these to hang around until the last pssible minute because the htoprc writing code may be called from segv handler and it needs to be able to map e.g. column identifiers to column names.

natoscott (approved):
I think we've covered everything (modulo some follow-up-PR worthy cleanups) - looks good.

Process.c
smalinux: @BenBE Done.

CommandLine.c
BenBE: The SIGSEGV handler should cope with any state of the system, in particular because it can be called in any situation, especially if the system is partly broken. Not properly unwinding here invites UAF bugs and other hardly reproducible states. I can live with some columns not being resolved in the SIGSEGV handler upon shutdown more than getting my system pwn'd by a UAF bug.

DynamicColumn.c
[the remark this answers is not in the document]
natoscott: @smalinux @BenBE I'll update this shortly, thanks

pcp/PCPDynamicColumn.c
natoscott: This behaviour is mirroring the way Settings_new() works over in Settings.c ... I agree its a bit odd though.  Will tweak it.

BenBE:
Let's do some reviewing of this patch series.

```
commit 87dad31ca396af8c677adff4bb25b54ef2002b45
Author: nia <nia@NetBSD.org>
Date:   Tue Jul 13 17:46:04 2021 +0200

    netbsd: Add dyanmicMeters stubs to fix the build
    
    Signed-off-by: Nia Alarie <nia@NetBSD.org>

diff --git a/netbsd/NetBSDProcessList.c b/netbsd/NetBSDProcessList.c
index 47bfeb23..f0394358 100644
--- a/netbsd/NetBSDProcessList.c
+++ b/netbsd/NetBSDProcessList.c
@@ -39,16 +39,6 @@ static long fscale;
 static int pageSize;
 static int pageSizeKB;
 
-static char const *freqSysctls[] = {
-   "machdep.est.frequency.current",
-   "machdep.powernow.frequency.current",
-   "machdep.intrepid.frequency.current",
-   "machdep.loongson.frequency.current",
-   "machdep.cpu.frequency.current",
-   "machdep.frequency.current",
-   NULL
-};
-
 ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* dynamicMeters, Hashtable* pidMatchList, uid_t userId) {
    const int mib[] = { CTL_HW, HW_NCPU };
    const int fmib[] = { CTL_KERN, KERN_FSCALE };
```

I don't think this belongs here ;-)

Same goes for this commit:

```
commit 4d0eeb0d9fb48bd6dd7e0b52fc4ce74bb088f28f
Author: nia <nia@NetBSD.org>
Date:   Wed Jul 14 20:53:43 2021 +0200

    netbsd: Support display of CPU frequency

diff --git a/netbsd/NetBSDProcessList.c b/netbsd/NetBSDProcessList.c
index f0394358..8a39df33 100644
--- a/netbsd/NetBSDProcessList.c
+++ b/netbsd/NetBSDProcessList.c
@@ -39,6 +39,22 @@ static long fscale;
 static int pageSize;
 static int pageSizeKB;
 
+static char const *freqSysctls[] = {
+#if defined(__powerpc__)
+   "machdep.intrepid.frequency.current",
+#endif
+#if defined(__mips__)
+   "machdep.loongson.frequency.current",
+#endif
+#if defined(__i386__) || defined(__x86_64__)
+   "machdep.est.frequency.current",
+   "machdep.powernow.frequency.current",
+#endif
+   "machdep.cpu.frequency.current",
+   "machdep.frequency.current",
+   NULL
+};
+
 ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* dynamicMeters, Hashtable* pidMatchList, uid_t userId) {
    const int mib[] = { CTL_HW, HW_NCPU };
    const int fmib[] = { CTL_KERN, KERN_FSCALE };
```

And this one:

```
commit e5e0c8663bec8e8982ab2b102a5a4e729e51de1a
Author: nia <nia@NetBSD.org>
Date:   Wed Jul 14 20:53:43 2021 +0200

    netbsd: Support display of CPU frequency

diff --git a/netbsd/NetBSDProcessList.c b/netbsd/NetBSDProcessList.c
index 8a39df33..f1d5a158 100644
--- a/netbsd/NetBSDProcessList.c
+++ b/netbsd/NetBSDProcessList.c
@@ -378,6 +378,7 @@ static void NetBSDProcessList_scanCPUTime(NetBSDProcessList* this) {
    kernelCPUTimesToHtop(avg, this->cpus);
 }
 
+<<<<<<< HEAD
 static void NetBSDProcessList_scanCPUFrequency(NetBSDProcessList* this) {
    unsigned int cpus = this->super.cpuCount;
    bool match = false;
@@ -422,6 +423,51 @@ static void NetBSDProcessList_scanCPUFrequency(NetBSDProcessList* this) {
    }
 }
 
+||||||| constructed merge base
+=======
+static void NetBSDProcessList_scanCPUFrequency(NetBSDProcessList* this) {
+   unsigned int cpus = this->super.cpuCount;
+   bool match = false;
+   char name[64];
+   int freq = 0;
+   size_t freqSize = sizeof(freq);
+
+   for (unsigned int i = 0; i <= cpus; i++) {
+         this->cpus[i].frequency = NAN;
+   }
+
+   /* newer hardware supports per-core frequency, for e.g. ARM big.LITTLE */
+   for (unsigned int i = 0; i <= cpus; i++) {
+      snprintf(name, sizeof(name), "machdep.cpufreq.cpu%u.current", i);
+      if (sysctlbyname(name, &freq, &freqSize, NULL, 0) != -1) {
+         this->cpus[i].frequency = freq;
+         match = true;
+      }
+   }
+
+   if (match) {
+      return;
+   }
+
+   /*
+    * Iterate through legacy sysctl nodes for single-core frequency until
+    * we find a match...
+    */
+   for (const char** s = freqSysctls; *s != NULL; ++s) {
+      if (sysctlbyname(*s, &freq, &freqSize, NULL, 0) != -1) {
+         match = true;
+         break;
+      }
+   }
+
+   if (match) {
+      for (unsigned int i = 0; i <= cpus; i++) {
+         this->cpus[i].frequency = freq;
+      }
+   }
+}
+
+>>>>>>> netbsd: Support display of CPU frequency
 void ProcessList_goThroughEntries(ProcessList* super, bool pauseProcessUpdate) {
    NetBSDProcessList* npl = (NetBSDProcessList*) super;
 
```

Commit 06d6e3bf97d70c0ec3639b7cec998478329fdaa9 has the following issues:
- Removes the broken code from the previous 3 commits.
- Some indentation issues
- Action.c @ `actionSetSortColumn`: `Panel_new` is called with `buffer` in call `ListItem_new(buffer, fields[i])` which is possibly truncated version of `column->caption` (IIRC fixed in later commit, but should be instantiated properly in the first place). Also shortens that code path
- AvailableColumnsPanel.c: Insufficient blank lines after headers
- AvailableColumnsPanel.c introduces commented-out function `AvailableColumnsPanel_insertDynamicColumnName`
- `AvailableColumnsPanel_eventHandler` has FIXME related to column caption
- `Makefile.am` introduces dependency on Linux ZRAM-related fiels (should be a separate commit if intentional and forgotten previously)
- `Process.c` contains large indentation change due to added `if` statement covering several screen pages. This changes is undone in a later commit AFAIR, but shouldn't be done in the first place to keep the patch set manageable and minimally covering.
- `ProcessList.c` @ `alignedProcessFieldTitle` has whitespace / styleguide issues
- `ProcessList.c` @ `ProcessList_scan` adds `//LOOKATME` comment should be resolved/removed
- `ProcessList.h` @ `typedef struct ProcessList_`: Comment alignment issue introduced for new field; comment should have `columns` spelled lower-case
- `Settings.c` @ `readFields`: Introduces unresolved(???) TODO to handle flags
- `pcp/PCPDynamicColumns.c` @ `PCPDynamicColumn_writeField`: Indentation issue
- `pcp/PCPDynamicColumn.h`: Insufficient blank lines after header includes
- `pcp/PCPDynamicColumn.h`: Introduced structs are anonymous
- `pcp/PCPProcess.c`: Insufficient blank lines after header includes
- `pcp/PCPProcess.c` @ `PCPProcess_compareByKey`: Code style issue regarding braces in final `else` clause
- `pcp/Platform.c` @ `typedef struct Platform_`: Newly introduced column `columns` undocumented
- `pcp/Platform.c` @ `Metric_desc`: extra blank line could be dropped
- `pcp/Platform.c` @ `Metric_iterate`: Introduces debug code
- `pcp/Platform.c` @ `Platform_init`: Introduces unresolved FIXME asking to move around code
- `pcp/Platform.c` @ `Platform_updateDynamicColumns`: Indentation / wrapping issues
- Inclusion of `pcp/columns` directory should be split into later commit; possibly merged with correction to only include the proper set of files.

commit 3e8179c4922b04dc166017cb3ccccad433db5e5b
- Contains changes unrelated to commit description.
- Should be squashed into previous commit, as changes are fixups to oversights there

commit 58668124401517c73edb92c1e4b61d87ce37e681
- Contains changes unrelated to this PR

commit 014dccbed4fd3c4fe067a6af60daa1787a8ee547
- Undoes those unrelated changes from previous commit 58668124401517c73edb92c1e4b61d87ce37e681

commit e46b9286d3bd2bd6240001eacb5606e1ba71164d
- Mixes code style and functional changes all over the place, thus making review unnecessrily hard
- Split into smaller pieces and squash those into originating commit that introduced the issue responsible for the fixup to exist
- Contains PR-unrelated changes (e.g. to the netbsd `freqSysctls` array
- Resolves several of the issues mentioned for previous commits
- Introduces new code in a cleanup commit

commit 78d9b3fd7c8277259839ccca7c17ec554a832343
- Should be combined with split out configuration file additions from commit 06d6e3bf97d70c0ec3639b7cec998478329fdaa9 (annotaded by fixups from e46b9286d3bd2bd6240001eacb5606e1ba71164d) AKA: Split out config additions where initially introduced, split out the related fixups and merge all three into this

commit 0204e1bd11ea677ef7f058bc22835e82213566cf
- Split changes for `pcp/columns` directory out and merge with introductory commit for those files; Move example config file commit to be last in patch series
- Typo `Wework` in commit message
- Consider splitting the changes from this commit into multiple parts to ease review
- `ColumnPanel.c` @ `ColumnsPanel_add`: Mixed(???) ownership of strings passed to `ListItem_new`: Either we have a double-free/UAF when freeing the list items of the panel, or we leak memory
- `ColumnPanel.c` @ `ColumnsPanel_new`: Consider putting the declaration/initialization of `fields` into the header of the `for` loop
- `CommandLine.c` @ `CommandLine_run`: Indentation for `Hashtable_delete(dm);`
- Updated prototypes in platform headers for `Platform_dynamicMeters` should come in separate commit

commit 33684c23ba601785ae235fa0812ce7638be91479
- Introduces changes that already should be on master. Missing rebase to upstream/master?

commit 3a47b538287f0959c4d28832eafb619e2d87ec61
- Introduces changes that already should be on master. Missing rebase to upstream/master?

commit 3d0c2e61a3fbdcec5ba6feb081a108cbaeddada6
- ACK.

commit df6c63427487c6e93943e22e1e773af693d8ddbf
- ACK.

Now there's no more concentration left after this review. The repeated touching of code blocks in repeated commits makes this patch set kinda noisy. Try to minimize changes per commit in a way that the individual commits tell a story. Also don't mix cleanup changes with functional changes in one commit: This easily allows for bugs to slip in unnoticed.

smalinux:
@BenBE Thank you for the review, and sorry for all these beginner mistakes 🙂

natoscott:
> Let's do some reviewing of this patch series.

Thanks @BenBE @smalinux - I'm working on tidying this up.  I'll drop the rebase leftovers & I'll likely merge most things into one or two commits (new code + any cleanup) for simpler reviewing and a clean git-log for future reference.  Having two people working on this has made attribution become a bit funky - since Sohaib did all of the early exploratory work and testing here, I'll likely simplify things by pushing everything in under his name.

BenBE:
> > Let's do some reviewing of this patch series.
> 
> I'll likely simplify things by pushing everything in under his name.

You should be able to just tell git to enter @smalinux as author and have yourself marked as committer. Have a look at `GIT_AUTHOR_NAME` and `GIT_AUTHOR_EMAIL` or `git commit --author` if you need.

natoscott:
> > > Let's do some reviewing of this patch series.
> > 
> > 
> > I'll likely simplify things by pushing everything in under his name.
> 
> You should be able to just tell git to enter @smalinux as author and have yourself marked as committer. Have a look at `GIT_AUTHOR_NAME` and `GIT_AUTHOR_EMAIL` or `git commit --author` if you need.

Good idea, will do.

natoscott:
@smalinux I've opened a PR for your dynamic columns branch - could you review and (--force) push here?  https://github.com/smalinux/htop/pull/2

BenBE:
```
FATAL PROGRAM ERROR DETECTED
============================

- Your htop version (htop --version)
- Likely steps to reproduce (How did it happened?)

Error information:
------------------
A signal 11 (Segmentation fault) was received.

Setting information:
--------------------
fields=0 3 48 17 18 38 39 40 43 2 50 46 47 49 1
sort_key=46
sort_direction=-1
tree_sort_key=39
tree_sort_direction=-1
hide_kernel_threads=1
hide_userland_threads=1
shadow_other_users=0
show_thread_names=0
show_program_path=0
highlight_base_name=1
highlight_deleted_exe=1
highlight_megabytes=1
highlight_threads=1
highlight_changes=1
highlight_changes_delay_secs=15
find_comm_in_cmdline=1
strip_exe_from_cmdline=1
show_merged_command=1
tree_view=1
tree_view_always_by_pid=0
all_branches_collapsed=0
header_margin=0
detailed_cpu_time=1
cpu_count_from_one=0
show_cpu_usage=1
show_cpu_frequency=1
show_cpu_temperature=0
degree_fahrenheit=0
update_process_names=1
account_guest_in_cpu_meter=1
color_scheme=0
enable_mouse=1
delay=5
left_meters=LeftCPUs2 RightCPUs2 Memory Swap Blank DiskIO NetworkIO
left_meter_modes=1 1 1 1 2 2 2
right_meters=Hostname System Tasks LoadAverage Uptime Battery Blank PressureStallMemorySome PressureStallMemoryFull PressureStallIOSome PressureStallIOFull
right_meter_modes=2 2 2 2 2 2 2 2 2 2 2
hide_function_bar=0

Backtrace information:
----------------------
The following function calls were active when the issue was detected:
---
./htop(CRT_handleSIGSEGV+0x14d)[0x56399f31ca80]
/lib/x86_64-linux-gnu/libc.so.6(+0x46210)[0x7f5c5b51e210] 
./htop(Hashtable_foreach+0x75)[0x56399f31f3cf]
./htop(+0x13e6d)[0x56399f317e6d]
./htop(AvailableColumnsPanel_new+0xb5)[0x56399f318039]
./htop(+0x14b05)[0x56399f318b05]
./htop(+0x14d22)[0x56399f318d22]
./htop(ScreenManager_run+0x3cb)[0x56399f3309d5]
./htop(+0x10b20)[0x56399f314b20]
./htop(+0x1198f)[0x56399f31598f]
./htop(+0x1e05a)[0x56399f32205a]
./htop(ScreenManager_run+0x3cb)[0x56399f3309d5]
./htop(CommandLine_run+0x3a8)[0x56399f31a581]
./htop(main+0x28)[0x56399f314771]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf3)[0x7f5c5b4ff0b3]
./htop(_start+0x2e)[0x56399f31468e]
---

To make the above information more practical to work with,
please also provide a disassembly of your htop binary.
This can usually be done by running the following command:

htop 3.1.0-dev aborting.

Segmentation fault (Core dumped)
```

```
 RBX  0x555555594b70 (__libc_csu_init) ◂— endbr64 |||||||||||||||||||||||||||||||32.9G/46.9G]
 RCX  0x3d                                                                             0K/0K] PSI some memory:  0.00%  0.00%  0.00% 
 RDX  0x555555c39420 —▸ 0x5555555a5960 (AvailableColumnsPanel_class) —▸ 0x5555555a6c00 (Panel_class) —▸ 0x555555597940 (Object_class) ◂— 0x0
 RDI  0x00.0% read: 0K write: 0K                                                              PSI some IO:      0.00%  0.00%  0.00% 
 RSI  0x555555567d32 (AvailableColumnsPanel_addDynamicColumn) ◂— endbr64                      PSI full IO:      0.00%  0.00%  0.00% 
 R8   0x555555ae8640 ◂— 'CWD - The current working directory of the process'                                                                                                                 
 R9   0x555555ae8930 ◂— 0x2c0lt                                                                                                                                                              
 R10  0x5555555ab010 ◂— 0x0ochromatic
 R11  0x7ffff7d95be0 (main_arena+96) —▸ 0x555555ca7520 ◂— 0x0
 R12  0x555555564660 (_start) ◂— endbr64 
 R13  0x7fffffffdee0 ◂— 0x1
 R14  0x0        [ ]    Black Night
 R15  0x0        [ ]    Broken Gray
 RBP  0x7fffffffd9f0 —▸ 0x7fffffffda10 —▸ 0x7fffffffda50 —▸ 0x7fffffffda80 —▸ 0x7fffffffdac0 ◂— ...
 RSP  0x7fffffffd9c0 —▸ 0x7fffffffd9f0 —▸ 0x7fffffffda10 —▸ 0x7fffffffda50 —▸ 0x7fffffffda80 ◂— ...
 RIP  0x55555556f3cf (Hashtable_foreach+117) ◂— mov    rax, qword ptr [rax]
─────────────────────────────────────────────────────────────────────────────────────────[ DISASM ]──────────────────────────────────────────────────────────────────────────────────────────
 ► 0x55555556f3cf <Hashtable_foreach+117>    mov    rax, qword ptr [rax]
   0x55555556f3d2 <Hashtable_foreach+120>    cmp    qword ptr [rbp - 0x10], rax
   0x55555556f3d6 <Hashtable_foreach+124>    jb     Hashtable_foreach+34 <Hashtable_foreach+34>
    ↓
   0x55555556f37c <Hashtable_foreach+34>     mov    rax, qword ptr [rbp - 0x18]
   0x55555556f380 <Hashtable_foreach+38>     mov    rcx, qword ptr [rax + 8]
   0x55555556f384 <Hashtable_foreach+42>     mov    rdx, qword ptr [rbp - 0x10]
   0x55555556f388 <Hashtable_foreach+46>     mov    rax, rdx
   0x55555556f38b <Hashtable_foreach+49>     add    rax, rax
   0x55555556f38e <Hashtable_foreach+52>     add    rax, rdx
   0x55555556f391 <Hashtable_foreach+55>     shl    rax, 3
   0x55555556f395 <Hashtable_foreach+59>     add    rax, rcx
──────────────────────────────────────────────────────────────────────────────────────────[ STACK ]──────────────────────────────────────────────────────────────────────────────────────────
00:0000│ rsp  0x7fffffffd9c0 —▸ 0x7fffffffd9f0 —▸ 0x7fffffffda10 —▸ 0x7fffffffda50 —▸ 0x7fffffffda80 ◂— ...
01:0008│      0x7fffffffd9c8 —▸ 0x555555c39420 —▸ 0x5555555a5960 (AvailableColumnsPanel_class) —▸ 0x5555555a6c00 (Panel_class) —▸ 0x555555597940 (Object_class) ◂— ...
02:0010│      0x7fffffffd9d0 —▸ 0x555555567d32 (AvailableColumnsPanel_addDynamicColumn) ◂— endbr64 
03:0018│      0x7fffffffd9d8 ◂— 0x0
... ↓
05:0028│      0x7fffffffd9e8 —▸ 0x555555c39470 ◂— 0x555500000011
06:0030│ rbp  0x7fffffffd9f0 —▸ 0x7fffffffda10 —▸ 0x7fffffffda50 —▸ 0x7fffffffda80 —▸ 0x7fffffffdac0 ◂— ...
07:0038│      0x7fffffffd9f8 —▸ 0x555555567e6d (AvailableColumnsPanel_addDynamicColumns+43) ◂— nop    
────────────────────────────────────────────────────────────────────────────────────────[ BACKTRACE ]────────────────────────────────────────────────────────────────────────────────────────
 ► f 0     55555556f3cf Hashtable_foreach+117
   f 1     555555567e6d AvailableColumnsPanel_addDynamicColumns+43
   f 2     555555568039 AvailableColumnsPanel_new+181
   f 3     555555568b05 CategoriesPanel_makeColumnsPage+69      F9      F10Done                                                                                                              
   f 4     555555568d22 CategoriesPanel_eventHandler+472
   f 5     5555555809d5 ScreenManager_run+971
   f 6     555555564b20 Action_runSetup+165
   f 7     55555556598f actionSetup+28
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
pwndbg> bt full
#0  0x000055555556f3cf in Hashtable_foreach ()
No symbol table info available.
#1  0x0000555555567e6d in AvailableColumnsPanel_addDynamicColumns ()
No symbol table info available.
#2  0x0000555555568039 in AvailableColumnsPanel_new ()
No symbol table info available.
#3  0x0000555555568b05 in CategoriesPanel_makeColumnsPage ()
No symbol table info available.
#4  0x0000555555568d22 in CategoriesPanel_eventHandler ()
No symbol table info available.
#5  0x00005555555809d5 in ScreenManager_run ()
No symbol table info available.
#6  0x0000555555564b20 in Action_runSetup ()
No symbol table info available.
#7  0x000055555556598f in actionSetup ()
No symbol table info available.
#8  0x000055555557205a in MainPanel_eventHandler ()
No symbol table info available.
#9  0x00005555555809d5 in ScreenManager_run ()
No symbol table info available.
#10 0x000055555556a581 in CommandLine_run ()
No symbol table info available.
#11 0x0000555555564771 in main ()
No symbol table info available.
#12 0x00007ffff7bd10b3 in __libc_start_main (main=0x555555564749 <main>, argc=1, argv=0x7fffffffdee8, init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffded8) at ../csu/libc-start.c:308
        self = <optimized out>
        result = <optimized out>
        unwind_buf = {
          cancel_jmp_buf = {{
              jmp_buf = {93824992496496, -3032261734600105369, 93824992298592, 140737488346848, 0, 0, 3032261733564981863, 3032278791375856231},
              mask_was_saved = 0
            }},
          priv = {
            pad = {0x0, 0x0, 0x1, 0x7fffffffdee8},
            data = {
              prev = 0x0,
              cleanup = 0x0,
              canceltype = 1
            }
          }
        }
        not_first_call = <optimized out>
#13 0x000055555556468e in _start ()
No symbol table info available.
```

Steps to reproduce:
1. Configure normal `htop` like `./configure --enable-linux-affinity --enable-delayacct --enable-dependency-tracking --enable-ancient-vserver --enable-vserver --enable-openvz --enable-sensors --enable-silent-rules`, but basically should not matter …
2. Then open up `htop`
3. Hit <kbd>F2</kbd> for setup
4. Move down to `Columns`.
5. *BOOM!*

smalinux:
@BenBE 
```
./configure --enable-pcp=yes
```

BenBE:
> @BenBE
> 
> ```
> ./configure --enable-pcp=yes
> ```

Exactly not. `htop` still needs to build and work properly without PCP being enabled. Otherwise there'd be not much sense in having that feature configurable … ;-)

smalinux:
@natoscott not sure, but I think this related https://github.com/htop-dev/htop/pull/685/commits/a4764902829d2a25a0e27c31824274362e11adea, we need to do the same for columns.

pcp/Platform.c
```diff
@@ -501,6 +511,9 @@ void Platform_init(void) {
    Metric_enable(PCP_UNAME_MACHINE, true);
    Metric_enable(PCP_UNAME_DISTRO, true);
 
+   for(unsigned int i = pcp->columns.offset; i < pcp->columns.offset + pcp->columns.count; i++)
```
BenBE, line 514:
```suggestion
   for (unsigned int i = pcp->columns.offset; i < pcp->columns.offset + pcp->columns.count; i++)
```

pcp/PCPDynamicColumn.h
```diff
@@ -0,0 +1,32 @@
...
+#include "CRT.h"
+#include "DynamicColumn.h"
+#include "Hashtable.h"
+#include "Process.h"
+#include "RichString.h"
+
+#include "pcp/PCPProcess.h"
+
+
+typedef struct {
```
BenBE, line 13:
```suggestion
typedef struct PCPDynamicColumn_ {
```

pcp/PCPDynamicColumn.h
```diff
@@ -0,0 +1,32 @@
...
+   DynamicColumn super;
+   char* metricName;
+   unsigned int id;  /* identifier for metric array lookups */
+} PCPDynamicColumn;
+
+typedef struct {
```
BenBE, line 19:
```suggestion
typedef struct PCPDynamicColumns_ {
```

pcp/PCPDynamicMeter.h
```diff
@@ -4,6 +4,7 @@
 #include "CRT.h"
 #include "DynamicMeter.h"
 
+
 typedef struct {
```
BenBE, line 8:
```suggestion
typedef struct PCPDynamicMetric_ {
```

Also fix `PCPDynamicMeter` and `PCPDynamicMeters`.

BenBE:
I'd prefer the following patch instead:

```diff
$ git diff
diff --git a/Settings.c b/Settings.c
index 07158e6f..8b366849 100644
--- a/Settings.c
+++ b/Settings.c
@@ -391,7 +391,7 @@ int Settings_write(const Settings* this, bool onCrash) {
 Settings* Settings_new(unsigned int initialCpuCount, Hashtable* dynamicColumns) {
    Settings* this = xCalloc(1, sizeof(Settings));
 
-   this->dynamicColumns = dynamicColumns;
+   this->dynamicColumns = dynamicColumns ? dynamicColumns : Hashtable_new(1, true);
    this->sortKey = PERCENT_CPU;
    this->treeSortKey = PID;
    this->direction = -1;
```

Much cleaner, less complexity added and offers DiD for NULL-deref issues down the line … The current code for the dynamic columns isn't generally written to cope with `dc` being NULL. Thus either fix it for every access or just act like an empty list of dynamic columns has been supplied.

pcp/PCPDynamicColumn.c
```diff
@@ -0,0 +1,324 @@
...
+   const PCPDynamicColumn* column = Hashtable_get(p1->super.processList->dynamicColumns, key);
+
+   unsigned int metric = column->id;
+   unsigned int type = Metric_type(metric);
+
+   pmAtomValue atom1 = {0}, atom2 = {0};
+   if (!Metric_instance(metric, p1->super.pid, p1->offset, &atom1, type) ||
+       !Metric_instance(metric, p2->super.pid, p2->offset, &atom2, type)) {
+      if (type == PM_TYPE_STRING)
+         free(atom1.cp);
```
BenBE, line 297: Shouldn't this also `free(atom2.cp)`?

BenBE (commented):
Things for follow-up cleanup:
- The column title alignment code may be updated to use a cache of previously aligned titles for columns. This should be done in conjunction with getting rid of the return of statically allocated module-global memory of those two functions.
- Check if `PCPDynamicColumn` and `PCPDynamicColumns` could utilize `size_t` instead of `unsigned int` for certain fields.
- Split the PCP Metric API (functions `Metric_*`) into their own module.

I'm probably missing something, but where's the `Metric_enable(FOO, false)` call for metrics being removed due to their column being removed in the settings?

pcp/PCPDynamicColumn.c
smalinux: please take a look @natoscott

natoscott:
> I'd prefer the following patch instead:

I think if we're going this route, the Hashtable allocation should happen earlier (up in CommandLine.c).
Putting it in Settings.c means we have other (NULL) pointers still in CommandLine and ProcessList to deal with, and also means the lifecycle of the Settings.c pointer has two different orbits and knowing the correctl place to free() the pointer is no longer possible.

The allocated-but-empty table is not my preferred solution because its just wasting memory (ok, very small amount) for platforms lacking support - the NULL pointer approach feels slightly more efficient to me.  Also if we allocate an empty hash for DynamicColumns we should probably do the same for DynamicMeters, for consistency there I guess.

> Much cleaner, less complexity added and offers DiD for NULL-deref issues down the line … The current code for the dynamic columns isn't generally written to cope with `dc` being NULL. Thus either fix it for every access or just act like an empty list of dynamic columns has been supplied.

There's not a huge number of accesses - I vote we just check for NULL and respond accordingly in each one that's missing the check.

natoscott:
> Things for follow-up cleanup:
> 
>     * The column title alignment code may be updated to use a cache of previously aligned titles for columns. 

For reference, which source files/lines are you talking about there?

> This should be done in conjunction with getting rid of the return of statically allocated module-global memory of those two functions.

@smalinux do you want to hack on this?

>     * Check if `PCPDynamicColumn` and `PCPDynamicColumns` could utilize `size_t` instead of `unsigned int` for certain fields.
> 
>     * Split the PCP Metric API (functions `Metric_*`) into their own module.

I'll take these two items on.  I'll do 'em as followup commits I think because they're likely to introduce patch conflicts with other PRs.
 
> I'm probably missing something, but where's the `Metric_enable(FOO, false)` call for metrics being removed due to their column being removed in the settings?

There isn't one - @smalinux do you know where that call would need to be?  I figured it wasn't a big deal to keep sampling one or two metrics we're not using anymore after interactive changes, but we should switch 'em off if we can.

BenBE:
@natoscott FYI: The minimalistic patch regarding `Settings.c` for the `Hashtable` was kinda retracted and that check placed in `CommandLine.c` instead. Cf. https://github.com/htop-dev/htop/compare/a0fefb6d7330539c73a3b10e13dbe96a5c4b3218..cf6d0ff4dad01747f668970697d1d0ea36a816de

I'd rather not follow the `NULL` approach for now, as this will lead to extra places to take care of possible `NULL` pointers where up to now we usually don't have this expectation in the code for our central structures.

But I'm fine if you make the `dynamicMeters` member non-`NULL` too for consistency too …

k, regarding header alignment and static buffers: I referred to `alignedDynamicColumnTitle` in `ProcessList.c`.

natoscott:
> @natoscott FYI: The minimalistic patch regarding `Settings.c` for the `Hashtable` was kinda retracted and that check placed in `CommandLine.c` instead. Cf. https://github.com/htop-dev/htop/compare/a0fefb6d7330539c73a3b10e13dbe96a5c4b3218..cf6d0ff4dad01747f668970697d1d0ea36a816de
> 

OK, fair enough.

> I'd rather not follow the `NULL` approach for now, as this will lead to extra places to take care of possible `NULL` pointers where up to now we usually don't have this expectation in the code for our central structures.
> 
> But I'm fine if you make the `dynamicMeters` member non-`NULL` too for consistency too …

OK, let's go that route then.

> k, regarding header alignment and static buffers: I referred to `alignedDynamicColumnTitle` in `ProcessList.c`.

Taa.

smalinux:
@natoscott 
> I'm probably missing something, but where's the Metric_enable(FOO, false) call for metrics being removed due to their column being removed in the settings?
>> There isn't one - @smalinux do you know where that call would need to be? I figured it wasn't a big deal to keep sampling one or two metrics we're not using anymore after interactive changes, but we should switch 'em off if we can.

- Columns: https://github.com/htop-dev/htop/blob/6974ce8e7982d061f26dbbe7c5ca48d7aa6f4dbc/AvailableColumnsPanel.c#L25
- Meters: https://github.com/htop-dev/htop/blob/6974ce8e7982d061f26dbbe7c5ca48d7aa6f4dbc/AvailableMetersPanel.c#L27

pcp/Platform.c
```diff
@@ -51,19 +52,20 @@ in the source distribution for its full text.
...
-   pmDesc* descs;		/* metric desc array indexed by Metric */
-   pmResult* result;		/* sample values result indexed by Metric */
-   PCPDynamicMeters meters;	/* dynamic meters via configuration files */
-   struct timeval offset;	/* time offset used in archive mode only */
-   long long btime;		/* boottime in seconds since the epoch */
-   char* release;		/* uname and distro from this context */
-   int pidmax;			/* maximum platform process identifier */
-   int ncpu;			/* maximum processor count configured */
+   int context;               /* PMAPI(3) context identifier */
+   unsigned int totalMetrics; /* total number of all metrics */
```
BenBE, line 56:
```suggestion
   size_t totalMetrics; /* total number of all metrics */
```

BenBE (commented):
LGTM. Didn't run it yet with those latest changes …

smalinux:
>Things for follow-up cleanup:
>* The column title alignment code may be updated to use a cache of previously aligned titles for columns. 
>>For reference, which source files/lines are you talking about there?

not sure this part is meant, anyway the column title code: https://github.com/htop-dev/htop/blob/8173856345d26e956d633a57da3e8c26a4b8c7b2/ProcessList.c#L88-L100

pcp/PCPDynamicMeter.c
```diff
@@ -139,12 +140,9 @@ static bool PCPDynamicMeter_validateMeterName(char* key, const char* path, unsig
 }
 
 // Ensure a meter name has not been defined previously
-static bool PCPDynamicMeter_uniqueName(char* key, const char* path, unsigned int line, PCPDynamicMeters* meters) {
+static bool PCPDynamicMeter_uniqueName(char* key, PCPDynamicMeters* meters) {
    if (DynamicMeter_search(meters->table, key, NULL) == false)
       return true;
```
BenBE, line 145:
```suggestion
   return !DynamicMeter_search(meters->table, key, NULL);
```

Remove the `return false;` accordingly.

pcp/Platform.c
```diff
@@ -51,19 +52,20 @@ in the source distribution for its full text.
...
-   pmDesc* descs;		/* metric desc array indexed by Metric */
-   pmResult* result;		/* sample values result indexed by Metric */
-   PCPDynamicMeters meters;	/* dynamic meters via configuration files */
-   struct timeval offset;	/* time offset used in archive mode only */
-   long long btime;		/* boottime in seconds since the epoch */
-   char* release;		/* uname and distro from this context */
-   int pidmax;			/* maximum platform process identifier */
-   int ncpu;			/* maximum processor count configured */
+   int context;               /* PMAPI(3) context identifier */
+   size_t totalMetrics; /* total number of all metrics */
```
BenBE, line 56: Indentation of line comment …

pcp/Platform.c
```diff
@@ -461,18 +467,21 @@ void Platform_init(void) {
 
    for (unsigned int i = 0; i < PCP_METRIC_COUNT; i++)
       Platform_addMetric(i, Platform_metricNames[i]);
-   pcp->meters.offset = PCP_METRIC_COUNT;
 
+   pcp->meters.offset = PCP_METRIC_COUNT;
```
BenBE, line 471: Why change this grouping of statements? Previously this was aligned with the loop that initializes this (IMHO better) vs. as prelude to the `PCPDynamicMeters_init` call (IMHO less optimal).

pcp/Platform.c
```diff
@@ -461,18 +467,21 @@ void Platform_init(void) {
...
    sts = pmLookupName(pcp->totalMetrics, pcp->names, pcp->pmids);
    if (sts < 0) {
       fprintf(stderr, "Error: cannot lookup metric names: %s\n", pmErrStr(sts));
       exit(1);
    }
 
-   for (unsigned int i = 0; i < pcp->totalMetrics; i++) {
-      pcp->fetch[i] = PM_ID_NULL;	/* default is to not sample */
+   for (size_t i = 0; i < pcp->totalMetrics; i++) {
+      pcp->fetch[i] = PM_ID_NULL;   /* default is to not sample */
```
BenBE, line 484: IMHO no real need to update the alignment of that line comment …

pcp/Platform.h
```diff
@@ -250,13 +251,15 @@ pmAtomValue* Metric_values(Metric metric, pmAtomValue* atom, int count, int type
...
+int Metric_type(Metric metric);
+
 int Metric_instanceCount(Metric metric);
 
 int Metric_instanceOffset(Metric metric, int inst);
 
 pmAtomValue* Metric_instance(Metric metric, int inst, int offset, pmAtomValue* atom, int type);
 
-int Platform_addMetric(Metric id, const char* name);
+size_t Platform_addMetric(Metric id, const char* name);
```
BenBE, line 262: Requires `#include <stddef.h>` …

BenBE (changes requested):
We're getting there …

pcp/Platform.c
smalinux: reverted.

BenBE (approved):
Okay. LGTM.

PR #709, opened 2021-07-26
netbsd: Add battery support

alarixnia:
This uses proplib and sysmon_envsys to determine the total charge
percentage of any number of connected batteries as well as the
AC adapter state. Should work with ACPI and non-ACPI systems.

netbsd-battery

### DIFF
--- a/netbsd/Platform.c
+++ b/netbsd/Platform.c
@@ -3,6 +3,7 @@ htop - netbsd/Platform.c
 (C) 2014 Hisham H. Muhammad
 (C) 2015 Michael McConville
 (C) 2021 Santhosh Raju
+(C) 2021 Nia Alarie
 (C) 2021 htop dev team
 Released under the GNU GPLv2, see the COPYING file
 in the source distribution for its full text.
@@ -11,12 +12,18 @@ in the source distribution for its full text.
 #include "netbsd/Platform.h"
 
 #include <errno.h>
+#include <fcntl.h>
+#include <paths.h>
+#include <unistd.h>
 #include <kvm.h>
 #include <limits.h>
 #include <math.h>
+#include <stdbool.h>
 #include <stdlib.h>
 #include <string.h>
 #include <time.h>
+#include <prop/proplib.h>
+#include <sys/envsys.h>
 #include <sys/resource.h>
 #include <sys/sysctl.h>
 #include <sys/time.h>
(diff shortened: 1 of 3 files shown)


netbsd/Platform.c
```diff
@@ -323,7 +330,100 @@ bool Platform_getNetworkIO(NetworkIOData* data) {
...
+   prop_object_iterator_t devIter, fieldsIter;
+   prop_object_t device, class, fieldsArray, curValue, maxValue, descField;
+
+   totalCharge = 0;
+   totalCapacity = 0;
+
+   *percent = NAN;
+   *isOnAC = AC_ERROR;
+
+   fd = open(_PATH_SYSMON, O_RDONLY);
```
BenBE, line 348: Declare here on first use …

netbsd/Platform.c
```diff
@@ -323,7 +330,100 @@ bool Platform_getNetworkIO(NetworkIOData* data) {
 }
 
 void Platform_getBattery(double* percent, ACPresence* isOnAC) {
-   // TODO
-   (void)percent;
-   (void)isOnAC;
+   int fd;
+   bool isACAD, isBattery;
```
BenBE, line 334: What does the ACAD stand for? Did you mean `isACAdapter`?

netbsd/Platform.c
```diff
@@ -323,7 +330,100 @@ bool Platform_getNetworkIO(NetworkIOData* data) {
...
+            /*
+             * After NetBSD 11's release NetBSD 9 will no longer be supported
+             * and these should be converted to prop_string_equals_string.
+             */
+
+            if (prop_string_equals_cstring(class, "ac-adapter")) {
+               isACAD = true;
+            } else if (prop_string_equals_cstring(class, "battery")) {
+               isBattery = true;
+            }
```
BenBE, line 391: Would checking for `prop_string_equals_string` work (using `configure.ac`)?

Thus having:
```c
#ifdef HAVE_PROP_STRING_EQUALS_STRING
#define proc_string_equals proc_string_equals_string
#else
#define proc_string_equals proc_string_equals_cstring
#endif
```

Or other question: How long is NetBSD 9 still to be supported, such that we could just ignore this old variant?

netbsd/Platform.c
```diff
@@ -323,7 +330,100 @@ bool Platform_getNetworkIO(NetworkIOData* data) {
...
+         }
+      }
+
+      if (isBattery && isPresent) {
+         totalCharge += curCharge;
+         totalCapacity += maxCharge;
+      }
+
+      if (isACAD) {
+         *isOnAC = isConnected ? AC_PRESENT : AC_ABSENT;
```
BenBE, line 420: What if there were multiple AC adapters present, but one disconnected?

BenBE (commented):
Mostly LGTM. Some questions though.

netbsd/Platform.c
alarixnia: Well, it's the current most recent release branch. proc_string_equals_cstring is only deprecated in the development branch.

netbsd/Platform.c
alarixnia: "AC Adapter". The abbreviation is used in envstat.

netbsd/Platform.c
alarixnia: What do other OSes do in this case?

netbsd/Platform.c
BenBE: Kinda non-obvious on first sight though. Took me the actual reference string a few lines down to figure it out. Maybe `isACAdapter` is a better choice?

netbsd/Platform.c
BenBE: Does the non-deprecated version exist at all in the old version?

alarixnia: It doesn't exist in 9.

BenBE: So we have to basically go via the check in `configure.ac` to make it compatible with both variants, without having to clean up the code a few years in …

netbsd/Platform.c
BenBE: Well, darwin leaks memory and takes the last one it finds … FreeBSD only seems to handle one battery, but depending on the return value of the syscall may treat everything with at least one connected source as on AC. Linux takes the ACPI reading for all batteries and terminates upon finding the first that's AC … Also the Linux code looks like it wants to receive some cleanups … PCP is not yet implemented … ;-) Neither is Solaris …

I think, the best behavior is to treat at least online AC connection as `AC_PRESENT` … Also don't leak memory while doing this. ;-)

BenBE (approved):

Issue #711, opened 2021-07-26
Memory leak in Darwin battery code


BenBE:
The Darwin code for reading the system battery state retains all suitable battery objects it can find, but only keeps track of the last one, thus introducing a memory leak on systems with more than one battery/AC connection present.

_Originally posted by @BenBE in https://github.com/htop-dev/htop/pull/709#discussion_r676970522_

PR #715, opened 2021-07-29
Properly handle multiple batteries on darwin

BenBE:
This makes the behaviour consistent with other platforms where AC is marked as present if at least one power source is marked as AC_PRESENT.

Fixes: #711


fasterit (approved):
Works on MacOS 11.3.1

fasterit:
Merged in e341217 just Github did not spot it

Issue #662, opened 2021-06-18
GNU extension in Makefile.am


BenBE:
When running `autogen.sh` I get the following warning:
```
$ ./autogen.sh
autoreconf: Entering directory `.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal --force --warnings=all 
autoreconf: configure.ac: tracing
autoreconf: configure.ac: not using Libtool
autoreconf: running: /usr/bin/autoconf --force --warnings=all
autoreconf: running: /usr/bin/autoheader --force --warnings=all
autoreconf: running: automake --add-missing --copy --force-missing --warnings=all
Makefile.am:396: warning: patsubst -O%,,$(CFLAGS: non-POSIX variable name
Makefile.am:396: (probably a GNU make extension)
autoreconf: Leaving directory `.'
```

The offending code is the `debug` invocation in `Makefile.am` (near the end).

PR #716, opened 2021-07-30
build: add -Wno-portability to the automake init call

natoscott:
This allows use of the non-POSIX variable name in a patsubst
call, which culls any optimiser flags in CFLAGS, without the
annoying warning at build time.

Resolves https://github.com/htop-dev/htop/issues/662

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -13,7 +13,7 @@ AC_CONFIG_AUX_DIR([build-aux])
 AC_CONFIG_HEADERS([config.h])
 
 AC_CANONICAL_TARGET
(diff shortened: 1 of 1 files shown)


Explorer09:
How about fixing the portability issue instead of suppressing the warning?

natoscott:
> How about fixing the portability issue instead of suppressing the warning?

Sure, I'll bite - how do you suggest doing that?

BenBE:
By using the non-bash syntax for subprocess content inclusion (which is also POSIX-compliant) the original warning can be avoided. Cf. 03705a20 for details.

Issue #720, opened 2021-08-04
Autogroup monitoring


unphased:
Here https://github.com/nlburgin/reallynice is an explanation for the problem related to autogroups in linux. Basically, the nice level of a process only affects that process's priority within its autogroup, and modern linux systems dole out autogroups based on session ID.

What this means is that for example when I run a build with `make -j32` on my terminal it makes not an iota of difference whether or not it has been `nice`d, the OS will schedule this task with the same relative priority as other processes on the system. In effect, the feature of nice itself has been broken for almost all of our command line workflows.

The nice level of a process still has a meaning (and works the same way), it only does so within an autogroup, however.

As such, I would recommend that a new column is added to htop that can show what autogroup each process is in. That way the `nice` readout (a default-enabled column!) could regain a semblance of utility on these systems which have autogroups switched on.

Furthermore, each autogroup itself has a nice level which the system will take into account. If you want to "background" something, previously you may `nice -n19` it, but now we usually need to `cat 19 > /proc/<pid>/autogroup` to achieve some semblance of similar behavior. The behavior isn't the same, though, as at this point all processes in that autogroup would still share CPU equally, and the regular nice level must be used to further manipulate that. 

This is spelled out very clearly under the autogroups section here: https://man7.org/linux/man-pages/man7/sched.7.html

Taking all of this into consideration, then, I'd say the default display of htop ought to be adjusted *for Linux systems to which this applies which have autogroups enabled*:

- remove process nice level column
- add autogroup column
- add autogroup nice level column

natoscott:
Quite surprising but this (CFS nice breakage) seems to be the case from a bit of reading around.

I can implement the metric extraction and display code tomorrow if noone else is super keen, and we can have a discussion about default display changes then once folks have had a chance to try it out.

BenBE:
I think we should add those two autogroup columns to new configurations when we detect them to be available.

Also IMHO we should keep the nice feature in place, but allow for the autogroup priority to be set too. Reprioritizing a whole group of processes when only one is selected may come as a surprise to the user, thus updating the group priority should be explicit (or by config to switch if nicing should apply to single processes/LWP or autogroups vs. default).

fasterit:
We should also make the F7 / F8 keys able to change auto-group niceness (e.g. as Shift-F7 / Shift-F8 and "{" / "}"  (as "[" / "]" are also bound to modify niceness))

BenBE:
Sound good to me with those keybindings …

PR #722, opened 2021-08-06
Add columns for process autogroup identifier and nice value

natoscott:
Adds AUTO (autogroup) and ANI (autogroup nice) columns that
report the information from /proc/PID/autogroup, as well as
handlers for '{' and '}' to change the autogroup nice value.

This is guarded by /proc/sys/kernel/sched_autogroup_enabled
such that sampling and/or changing values wont be attempted
unless the kernel feature is enabled.

Fixes: #720


fasterit (approved):

BenBE:
What about `AGRP` instead of `AUTO`? IMHO more descriptive random sequence of letters ;-)

linux/LinuxProcess.c
```diff
@@ -159,6 +161,40 @@ bool LinuxProcess_setIOPriority(Process* this, Arg ioprio) {
    return (LinuxProcess_updateIOPriority((LinuxProcess*)this) == ioprio.i);
 }
 
+bool LinuxProcess_isAutogroupEnabled(void) {
+   FILE* file = fopen(PROCDIR "/sys/kernel/sched_autogroup_enabled", "r");
+   if (!file)
+      return false;
+   unsigned int enabled;
+   int count = fscanf(file, "%u", &enabled);
+   fclose(file);
```
BenBE, line 170:
```suggestion
   char buf[16];
   if (xReadfile(PROCDIR "/sys/kernel/sched_autogroup_enabled", buf, sizeof(buf)) < 0)
      return false;

   unsigned int enabled;
   int count = sscanf(buf, "%u", &enabled);
```

linux/LinuxProcess.h
```diff
@@ -99,6 +100,10 @@ typedef struct LinuxProcess_ {
    unsigned long ctxt_diff;
    char* secattr;
    unsigned long long int last_mlrs_calctime;
+
+   /* Autogroup scheduling (CFS) information */
+   unsigned int autogroup;
```
BenBE, line 105: `autogroup_id` / `autogroup_gid`?

linux/LinuxProcessList.c
```diff
@@ -909,6 +909,24 @@ static void LinuxProcessList_readOomData(LinuxProcess* process, openat_arg_t pro
    fclose(file);
 }
 
+static void LinuxProcessList_readAutogroup(LinuxProcess* process, openat_arg_t procFd) {
+   FILE* file = fopenat(procFd, "autogroup", "r");
```
BenBE, line 913: Use `xReadfileAt`?

linux/Platform.c
```diff
@@ -150,8 +150,41 @@ static Htop_Reaction Platform_actionSetIOPriority(State* st) {
    return HTOP_REFRESH | HTOP_REDRAW_BAR | HTOP_UPDATE_PANELHDR;
 }
 
+
```
BenBE, line 153: No need for double blank here …

linux/ProcessField.h
```diff
@@ -44,8 +44,10 @@ in the source distribution for its full text.
    M_PSSWP = 121,                \
    CTXT = 122,                   \
    SECATTR = 123,                \
+   AUTOGROUP = 127,              \
+   AUTOGROUP_NICE = 128,         \
                                  \
-   DUMMY_BUMP_FIELD = CWD,       \
+   DUMMY_BUMP_FIELD = 128,       \
```
BenBE, line 50: Can be dropped, was only there to ensure the last ID defined here was the overall highest ID available.

linux/LinuxProcess.c
```diff
@@ -100,6 +100,8 @@ const ProcessFieldData Process_fields[LAST_PROCESSFIELD] = {
    [PROC_COMM] = { .name = "COMM", .title = "COMM            ", .description = "comm string of the process from /proc/[pid]/comm", .flags = 0, },
    [PROC_EXE] = { .name = "EXE", .title = "EXE             ", .description = "Basename of exe of the process from /proc/[pid]/exe", .flags = 0, },
    [CWD] = { .name = "CWD", .title = "CWD                       ", .description = "The current working directory of the process", .flags = PROCESS_FLAG_CWD, },
+   [AUTOGROUP] = { .name = "AUTOGROUP", .title = "AUTO", .description = "The autogroup identifier of the process", .flags = PROCESS_FLAG_LINUX_AUTOGROUP, },
```
BenBE, line 103: `AUTOGROUP` -> `AUTOGROUP_ID` as there's multiple fields on this topic.

BenBE (changes requested):

linux/LinuxProcess.c
```diff
@@ -159,6 +161,40 @@ bool LinuxProcess_setIOPriority(Process* this, Arg ioprio) {
...
+   xSnprintf(buffer, sizeof(buffer), PROCDIR "/%d/autogroup", this->pid);
+
+   FILE* file = fopen(buffer, "r+");
+   if (!file)
+      return false;
+
+   unsigned int identity;
+   long int nice;
+   bool success;
+   int ok = fscanf(file, "/autogroup-%u nice %ld", &identity, &nice);
```
BenBE, line 185: Type mismatch with kernel source:
https://github.com/torvalds/linux/blob/v5.13/kernel/sched/autogroup.c#L254

BenBE (changes requested):

linux/LinuxProcess.c
cgzones: Then we would not even need the `scanf()`, as we could just check `buf[0]` (which should be either '0' or '1').

linux/LinuxProcessList.c
```diff
@@ -909,6 +909,24 @@ static void LinuxProcessList_readOomData(LinuxProcess* process, openat_arg_t pro
    fclose(file);
 }
 
+static void LinuxProcessList_readAutogroup(LinuxProcess* process, openat_arg_t procFd) {
```
cgzones, line 912: Maybe set `process->autogroup` and `process->autogroup_nice` to some invalid value first and print `N/A` in `LinuxProcess_writeField()` on those?

linux/LinuxProcessList.c
```diff
@@ -1323,6 +1341,15 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
       return false;
    }
 
+   bool autogroup = (settings->flags & PROCESS_FLAG_LINUX_AUTOGROUP);
+   if (autogroup) {
+      // Refer to sched(7) 'autogroup feature' section
+      // The kernel feature can be enabled/disabled through procfs at
+      // any time, so check for it at the start of each sample - only
+      // read from per-process procfs files if it's globally enabled.
+      autogroup = LinuxProcess_isAutogroupEnabled();
```
cgzones, line 1350: This leads to `LinuxProcess_isAutogroupEnabled()` being called for each process and thread if `PROCESS_FLAG_LINUX_AUTOGROUP` is set.

natoscott:
> What about `AGRP` or `AGID` instead of `AUTO`? IMHO more descriptive random sequence of letters ;-)

Done, went with AGRP - thanks.

BenBE (approved):

PR #723, opened 2021-08-08
netbsd: Add support for DiskIOMeter

alarixnia:


BenBE (approved):

netbsd/Platform.c
```diff
@@ -329,9 +331,41 @@ FileLocks_ProcessData* Platform_getProcessLocks(pid_t pid) {
...
 bool Platform_getDiskIO(DiskIOData* data) {
-   // TODO
-   (void)data;
-   return false;
+   const int mib[] = { CTL_HW, HW_IOSTATS, sizeof(struct io_sysctl) };
+   struct io_sysctl *iostats = NULL;
+   size_t size = 0;
+
+   /* get the size of the IO statistic array */
+   (void)sysctl(mib, __arraycount(mib), iostats, &size, NULL, 0);
```
cgzones, line 339: Please check for failure:

```c
int r = sysctl(mib, __arraycount(mib), iostats, &size, NULL, 0);
if (r < 0 || size == 0)
   return false;
```

BenBE: Not quite. This `sysctl` is expected to fail with `ENOMEM` based on [sysctl(3)](https://man.netbsd.org/sysctl.3):
> The information is copied into the buffer specified by `oldp`.  The size of the buffer is given by the location specified by `oldlenp` before the call, and that location gives the amount of data copied after a successful call.  **If the amount of data available is greater than the size of the buffer supplied, the call supplies as much data as fits in the buffer provided and returns with the error code `ENOMEM`.**  If the old value is not desired, `oldp` and `oldlenp` should be set to NULL.

cgzones: The next paragraph:

>  The size of the available data can be determined by calling sysctl with a
     NULL parameter for oldp.  The size of the available data will be returned
     in the location pointed to by oldlenp.  For some operations, the amount
     of space may change often.  For these operations, the system attempts to
     round up so that the returned size is large enough for a call to return
     the data shortly thereafter.

So I would assume `sysctl(mib, __arraycount(mib), NULL, &size, NULL, 0)` does not fail with ENOMEM, as we do not supply any buffer, which might be too short.

BenBE: I'd still assume the `sysctl` to return `ENOMEM` in his case. Thus better yet, loop until you succeed to allocate a buffer large enough, but break on any non-`ENOMEM` error returned.

cgzones: https://github.com/htop-dev/htop/blob/aa0424ade806715ea0f410dc068a1eb648300bb9/darwin/DarwinProcessList.c#L111-L126

BenBE (approved):

PR #727, opened 2021-08-09
netbsd: handle repeated ENOMEM from HW_IOSTATS safely

alarixnia:
This isn't critical but it's a good example of why I shouldn't make commits early in the morning

### DIFF
--- a/netbsd/Platform.c
+++ b/netbsd/Platform.c
@@ -347,12 +347,17 @@ bool Platform_getDiskIO(DiskIOData* data) {
 
       iostats = xRealloc(iostats, size);
 
+      errno = 0;
+
       if (sysctl(mib, __arraycount(mib), iostats, &size, NULL, 0) == 0)
          break;
 
       if (errno != ENOMEM)
          CRT_fatalError("Unable to get disk IO statistics");
    }
+
+   if (errno == ENOMEM)
+      CRT_fatalError("Unable to get disk IO statistics");
 
    uint64_t bytesReadSum = 0;
    uint64_t bytesWriteSum = 0;


netbsd/Platform.c
```diff
@@ -347,13 +347,18 @@ bool Platform_getDiskIO(DiskIOData* data) {
...
+
       if (sysctl(mib, __arraycount(mib), iostats, &size, NULL, 0) == 0)
          break;
 
       if (errno != ENOMEM)
          CRT_fatalError("Unable to get disk IO statistics");
    }
 
+   if (errno == ENOMEM)
+      CRT_fatalError("Unable to get disk IO statistics");
```
cgzones, line 360: Please use a different error message than used 4 lines above, to distinguish them.

BenBE (approved):

BenBE:
@alarixnia Can you please test the following patch on NetBSD?

```diff
$ git diff
diff --git a/netbsd/Platform.c b/netbsd/Platform.c
index 9d57b149..ca747821 100644
--- a/netbsd/Platform.c
+++ b/netbsd/Platform.c
@@ -336,27 +336,21 @@ bool Platform_getDiskIO(DiskIOData* data) {
    size_t size = 0;
 
    for (int retry = 3; retry > 0; retry--) {
+      errno = 0;
+
       /* get the size of the IO statistic array */
-      if (sysctl(mib, __arraycount(mib), iostats, &size, NULL, 0) < 0)
-         CRT_fatalError("Unable to get size of io_sysctl");
+      if (sysctl(mib, __arraycount(mib), iostats, &size, NULL, 0) == 0)
+         break;
 
-      if (size == 0) {
+      if (size == 0 || errno != ENOMEM) {
          free(iostats);
          return false;
       }
 
       iostats = xRealloc(iostats, size);
-
-      errno = 0;
-
-      if (sysctl(mib, __arraycount(mib), iostats, &size, NULL, 0) == 0)
-         break;
-
-      if (errno != ENOMEM)
-         CRT_fatalError("Unable to get disk IO statistics");
    }
 
-   if (errno == ENOMEM)
+   if (errno || !size)
       CRT_fatalError("Unable to get disk IO statistics");
 
    uint64_t bytesReadSum = 0;
```

In particular the first iteration of that loop would be interesting if `sysctl` for `oldp==NULL` succeeds or fails with `ENOMEM`. If it fails (as I'd expect) then this safes quite a bit on source code …

PR #728, opened 2021-08-09
Read settings after applying defaults

bjpbakker:
Default settings are used as a base and only settings specified in `htoprc` are
applied on top of it. This patch removes the special case for applying some
defaults  when the config does not contain a `meters` key. All defaults are set
before any attempt to read settings, so only keys actually present in the config
file are overridden.

### DIFF
--- a/Settings.c
+++ b/Settings.c
@@ -131,12 +131,13 @@ static bool Settings_read(Settings* this, const char* fileName, unsigned int ini
       return false;
 
    bool didReadMeters = false;
-   bool didReadFields = false;
+   bool didReadAny = false;
    for (;;) {
       char* line = String_readLine(fd);
       if (!line) {
          break;
       }
+      didReadAny = true;
       size_t nOptions;
       char** option = String_split(line, '=', &nOptions);
       free (line);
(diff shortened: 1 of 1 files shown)


Settings.c
```diff
@@ -146,7 +147,6 @@ static bool Settings_read(Settings* this, const char* fileName, unsigned int ini
       }
       if (String_eq(option[0], "fields")) {
          readFields(this->fields, &(this->flags), option[1]);
-         didReadFields = true;
```
cgzones, line 149: Is the rework from `didReadFields` to `didReadAny` really needed?
The `fields` key defines the list of enabled process field columns, which every configuration file should contain.

bjpbakker: The reason for this patch is mostly that if the configuration file does not contain the `fields` key, the rest of the configuration keys are not applied either.

When manually editing the file and this can be confusing.

BenBE: To quote the source on this:
https://github.com/htop-dev/htop/blob/master/Settings.c#L327
> The parser is also very primitive, **and not human-friendly**.

fasterit: I still think the approach from @bjpbakker adds value as it cleans up the code and makes it more what you'd expect the relationship between compiled-in defaults and a setting file to be.
So ACK from DLange.

BenBE: Never really any objection from my side either. Just marked it as "Default change" due to this change of overall behavior.

PR #736, opened 2021-08-12
Add rounded corners to LEDMeterMode_digitsUtf8. Fixes #735

### LINKED ISSUE #735: Improve large digit readability by using rounded corners
Large digits in `Meter.c` use line drawing characters, but relies on "hard" corners originally present in IBM PC character ROMs. Most current systems, either by direct terminal support, or font support, provide rounded corners that would make the digits more readable:

image

instead of

image

rbanffy:
This should make the large digits look nicer in modern terminals.

image


BenBE:
Are there also some somewhat dented characters that could fill in for the right middle of the digit 3 and the middle left+right of the digit 8?

rbanffy:
> Are there also some somewhat dented characters that could fill in for the right middle of the digit 3 and the middle left+right of the digit 8?

Sadly no. Or, if they exist, I couldn't find them.

cgzones:
Probably the hole LED display mode needs some attention, e.g. the dots between the digits look odd.
Also personally I like the sharp corners more, as they remind me of alarm clocks.

rbanffy:
> Probably the hole LED display mode needs some attention, e.g. the dots between the digits look odd.
> Also personally I like the sharp corners more, as they remind me of alarm clocks.

I think it could be more consistent with a curved top segment for the 1 and the 7. And the dots can be moved down to the bottom row, but the ASCII char "." now looks too low.

image

Meter.c
```diff
@@ -408,9 +408,9 @@ static void LEDMeterMode_draw(Meter* this, int x, int y, ATTR_UNUSED int w) {
       } else {
 #ifdef HAVE_LIBNCURSESW
          const cchar_t wc = { .chars = { c, '\0' }, .attr = 0 }; /* use LED_COLOR from attrset() */
-         mvadd_wch(yText, xx, &wc);
+         mvadd_wch(yText + 1, xx, &wc);
 #else
-         mvaddch(yText, xx, c);
+         mvaddch(yText + 1, xx, c);
 #endif
```
BenBE, line 414: Should probably only be done for `.`; maybe replace with `⋅` (U+22C5: DOT OPERATOR) for Unicode versions

rbanffy:
Indeed. This works only for characters that align to the bottom of the
character cell (I didn't search for others, just wanted to make the
"." move down). A permanent solution would have to deal with other
characters and their alignments.

We could also add "." and "/" to the LED characters and do the range
check between "." and "9" instead of "0" and "9", but, when I tried it
the lines overflowed - perhaps it's assuming the width of the
non-numeric chars is 1 column.

On Mon, Aug 16, 2021 at 7:15 PM BenBE ***@***.***> wrote:
>
> @BenBE commented on this pull request.
>
> ________________________________
>
> In Meter.c:
>
> >  #ifdef HAVE_LIBNCURSESW
>
>           const cchar_t wc = { .chars = { c, '\0' }, .attr = 0 }; /* use LED_COLOR from attrset() */
>
> -         mvadd_wch(yText, xx, &wc);
>
> +         mvadd_wch(yText + 1, xx, &wc);
>
>  #else
>
> -         mvaddch(yText, xx, c);
>
> +         mvaddch(yText + 1, xx, c);
>
>  #endif
>
> Should probably only be done for .; maybe replace with ⋅ (U+22C5: DOT OPERATOR) for Unicode versions
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub, or unsubscribe.



-- 
Ricardo Bánffy
http://about.me/rbanffy

Meter.c
rbanffy: If we agree it looks better, I can clean it up and make a separate code path for the dot.

BenBE:
As there's pros and cons for both the pointy and rounded corners, what about having a setting for the style in the display options?

rbanffy:
> As there's pros and cons for both the pointy and rounded corners, what about having a setting for the style in the display options?

Sounds like a good idea. Does htop have a policy about being compatible with the Linux fbconsole? I don't think it has the rounder line drawing chars. 

Also, I haven't figured out how to create a new display style, so that's a next step.

BenBE:
> Does htop have a policy about being compatible with the Linux fbconsole?

None explicitly that I'm aware of. What I know is that there's some fallback for non-UTF8 terminals, so this may be copied for the new style too.

> I don't think it has the rounder line drawing chars.

Let's check and if necessary adapt …

> Also, I haven't figured out how to create a new display style, so that's a next step.

Currently mostly by copying that one text array + updating references to it. Going forward it might be wise to rewrite this access code a bit to have one ID distinguishing between ASCII mode, pointy and rounded display. When initializing htop we can then treat style 1 and 2 just as style 0 for ASCII mode:

```diff
diff --git a/Meter.c b/Meter.c
index 73e99ce5..3035de66 100644
--- a/Meter.c
+++ b/Meter.c
@@ -364,12 +364,18 @@ static const char* const LEDMeterMode_digitsAscii[] = {
 
 #ifdef HAVE_LIBNCURSESW
 
-static const char* const LEDMeterMode_digitsUtf8[] = {
+static const char* const LEDMeterMode_digitsPointyUtf8[] = {
    "┌──┐", "  ┐ ", "╶──┐", "╶──┐", "╷  ╷", "┌──╴", "┌──╴", "╶──┐", "┌──┐", "┌──┐",
    "│  │", "  │ ", "┌──┘", " ──┤", "└──┤", "└──┐", "├──┐", "   │", "├──┤", "└──┤",
    "└──┘", "  ╵ ", "└──╴", "╶──┘", "   ╵", "╶──┘", "└──┘", "   ╵", "└──┘", " ──┘"
 };
 
+static const char* const LEDMeterMode_digitsSmoothUtf8[] = {
+   "╭──╮", "  ╮ ", "╭──╮", "╭──╮", "╷  ╷", "┌──╴", "╭─  ", "╶──╮", "╭──╮", "╭──╮",
+   "│  │", "  │ ", "╭──╯", " ──┤", "└──┤", "└──╮", "├──╮", "   │", "├──┤", "╰──┤",
+   "╰──╯", "  ╵ ", "└──╴", "╰──╯", "   ╵", "╰──╯", "╰──╯", "   ╵", "╰──╯", " ──╯"
+};
+
 #endif
 
 static const char* const* LEDMeterMode_digits;
@@ -382,7 +388,11 @@ static void LEDMeterMode_drawDigit(int x, int y, int n) {
 static void LEDMeterMode_draw(Meter* this, int x, int y, int w) {
 #ifdef HAVE_LIBNCURSESW
    if (CRT_utf8)
-      LEDMeterMode_digits = LEDMeterMode_digitsUtf8;
+      if (true) { // check via this->settings for rounding flag
+         LEDMeterMode_digits = LEDMeterMode_digitsPointyUtf8;
+      } else {
+         LEDMeterMode_digits = LEDMeterMode_digitsSmoothUtf8;
+      }
    else
 #endif
       LEDMeterMode_digits = LEDMeterMode_digitsAscii;
```
(Plus the changes for a new setting).

The other option would be a new meter mode:
```diff
diff --git a/Meter.c b/Meter.c
index 73e99ce5..cb3fb07c 100644
--- a/Meter.c
+++ b/Meter.c
@@ -364,12 +364,18 @@ static const char* const LEDMeterMode_digitsAscii[] = {
 
 #ifdef HAVE_LIBNCURSESW
 
-static const char* const LEDMeterMode_digitsUtf8[] = {
+static const char* const LEDMeterMode_digitsPointyUtf8[] = {
    "┌──┐", "  ┐ ", "╶──┐", "╶──┐", "╷  ╷", "┌──╴", "┌──╴", "╶──┐", "┌──┐", "┌──┐",
    "│  │", "  │ ", "┌──┘", " ──┤", "└──┤", "└──┐", "├──┐", "   │", "├──┤", "└──┤",
    "└──┘", "  ╵ ", "└──╴", "╶──┘", "   ╵", "╶──┘", "└──┘", "   ╵", "└──┘", " ──┘"
 };
 
+static const char* const LEDMeterMode_digitsSmoothUtf8[] = {
+   "╭──╮", "  ╮ ", "╭──╮", "╭──╮", "╷  ╷", "┌──╴", "╭─  ", "╶──╮", "╭──╮", "╭──╮",
+   "│  │", "  │ ", "╭──╯", " ──┤", "└──┤", "└──╮", "├──╮", "   │", "├──┤", "╰──┤",
+   "╰──╯", "  ╵ ", "└──╴", "╰──╯", "   ╵", "╰──╯", "╰──╯", "   ╵", "╰──╯", " ──╯"
+};
+
 #endif
 
 static const char* const* LEDMeterMode_digits;
@@ -380,13 +386,6 @@ static void LEDMeterMode_drawDigit(int x, int y, int n) {
 }
 
 static void LEDMeterMode_draw(Meter* this, int x, int y, int w) {
-#ifdef HAVE_LIBNCURSESW
-   if (CRT_utf8)
-      LEDMeterMode_digits = LEDMeterMode_digitsUtf8;
-   else
-#endif
-      LEDMeterMode_digits = LEDMeterMode_digitsAscii;
-
    RichString_begin(out);
    Meter_displayBuffer(this, &out);
 
@@ -424,6 +423,28 @@ static void LEDMeterMode_draw(Meter* this, int x, int y, int w) {
    RichString_delete(&out);
 }
 
+static void LEDMeterMode_pointy_draw(Meter* this, int x, int y, int w) {
+#ifdef HAVE_LIBNCURSESW
+   if (CRT_utf8)
+         LEDMeterMode_digits = LEDMeterMode_digitsPointyUtf8;
+   else
+#endif
+      LEDMeterMode_digits = LEDMeterMode_digitsAscii;
+
+   LEDMeterMode_draw(this, x, y, w);
+}
+
+static void LEDMeterMode_smooth_draw(Meter* this, int x, int y, int w) {
+#ifdef HAVE_LIBNCURSESW
+   if (CRT_utf8)
+         LEDMeterMode_digits = LEDMeterMode_digitsSmoothUtf8;
+   else
+#endif
+      LEDMeterMode_digits = LEDMeterMode_digitsAscii;
+
+   LEDMeterMode_draw(this, x, y, w);
+}
+
 static MeterMode BarMeterMode = {
    .uiName = "Bar",
    .h = 1,
@@ -442,10 +463,16 @@ static MeterMode GraphMeterMode = {
    .draw = GraphMeterMode_draw,
 };
 
-static MeterMode LEDMeterMode = {
-   .uiName = "LED",
+static MeterMode LEDMeterModePointy = {
+   .uiName = "LED (pointy)",
+   .h = 3,
+   .draw = LEDMeterMode_pointy_draw,
+};
+
+static MeterMode LEDMeterModeSmooth = {
+   .uiName = "LED (smooth)",
    .h = 3,
-   .draw = LEDMeterMode_draw,
+   .draw = LEDMeterMode_smooth_draw,
 };
 
 const MeterMode* const Meter_modes[] = {
@@ -453,7 +480,8 @@ const MeterMode* const Meter_modes[] = {
    &BarMeterMode,
    &TextMeterMode,
    &GraphMeterMode,
-   &LEDMeterMode,
+   &LEDMeterModePointy,
+   &LEDMeterModeSmooth,
    NULL
 };
 
diff --git a/Meter.h b/Meter.h
index da5792f9..cd2ce916 100644
--- a/Meter.h
+++ b/Meter.h
@@ -128,7 +128,9 @@ typedef enum {
    BAR_METERMODE,
    TEXT_METERMODE,
    GRAPH_METERMODE,
-   LED_METERMODE,
+   LED_POINTY_METERMODE,
+   LED_METERMODE = LED_POINTY_METERMODE,
+   LED_SMOOTH_METERMODE,
    LAST_METERMODE
 } MeterModeId;
 
```

Both variants just quick hacks for demo …

rbanffy:
> Let's check and if necessary adapt 

I'll check that

> Both variants just quick hacks for demo …

I tend to favor the new meter mode. It's neater. I'll start work on that.

natoscott:
Small point (heh) on naming - I recommend we use "square" over "pointy" here, especially in anything user-facing.  No problem with "smooth" myself, I like that name - but a boring alternative would be "round".

rbanffy:
> Small point (heh) on naming - I recommend we use "square" over "pointy" here, especially in anything user-facing. No problem with "smooth" myself, I like that name - but a boring alternative would be "round".

I'd go with "square" and "rounded" in this case. Optimizing for boring is always good.

BenBE:
Another micro-optimization: Use the suggested dot replacement character instead of `.` when that character is followed by a digit AND place it in the bottom row.

BenBE:
As we wanted to give this proposal a fair chance we've created [a poll on Twitter](https://twitter.com/htopdev/status/1447705722414944258). This poll ran for a week an got only 18 votes, which was much less than we set as the required quorum. Even ignoring that only few participated we couldn't see a strong preference for either option. And while the results are nearly 50:50, they are not within the narrow range that we'd have been willing to support this choice as a setting.

We thus decided to not implement this proposal.

PR #746, opened 2021-08-16
Increase USER/ST_UID size to support 32-bit UIDs

silkeh:
While most Unix-like systems use 16-bit user IDs,
Linux supports 32-bit UIDs since version 2.6.
UIDs above 65535 are used for UID namespacing of containers,
where a container has its own set of 16-bit user IDs.
Processes in such containers will have (much) larger UIDs than 65535.

Because the current format strings for `ST_UID` and `USER`
are `%5d` and `%9d` respectively, processes with such UIDs
lead to misaligned columns.

Update the size of both columns to 10 characters (length of UINT32_MAX)
to ensure that the user ID always fits.


Explorer09:
Can we have a field width that's dynamically adjusted? 10-digit PID seems to waste a lot of terminal space when user does not use namespace or containers.

natoscott:
> Can we have a field width that's dynamically adjusted? 10-digit PID seems to waste a lot of terminal space when user does not use namespace or containers.

+1  (though I guess you mean UID not PID).  @silkeh you might be able to refer to the way the PID column width is dynamically sized for some code inspiration.

Explorer09:
@natoscott Um, yes, I confused the two. Sorry.
And you got what I mean. UID column should be dynamically sized the same way as PID can be.

silkeh:
I took a look at implementing something similar to how PID is scaled. One thing to note is that the maximum PID is a runtime setting in Linux (and other OSes), while the maximum UID is fixed. This means that (when implemented the same way as the PID) the UID column would still use 10 digits in Linux, because the maximum UID is always UINT32_MAX. The same is true for most BSDs, I think.

BenBE:
A workaround for this could be to track the highest seen UID and initialize it to some 5 digit number on start. If longer UIDs are encountered the column resizes automagically with that updated value. Tracking the maximum UID seen shouldn't add much runtime overhead.

silkeh:
> A workaround for this could be to track the highest seen UID and initialize it to some 5 digit number on start. If longer UIDs are encountered the column resizes automagically with that updated value. Tracking the maximum UID seen shouldn't add much runtime overhead.

Updated the PR with a commit that implements dynamic scaling of the UID column. I'm not sure if this is the cleanest/clearest implementation possible, let me know what can be improved! One thing I can think of is to rename `PROCESS_MAX_PID_DIGITS` and `.pidColumn` to something like `PROCESS_MAX_ID_DIGITS` and `.idColumn`, as that would deduplicate some of the code added.

BenBE (approved):

natoscott (approved):
LGTM too, nicely done.

Process.c
```diff
@@ -52,6 +53,14 @@ void Process_setupColumnWidths() {
    assert(Process_pidDigits <= PROCESS_MAX_PID_DIGITS);
 }
 
+void Process_setUidColumnWidth(unsigned int maxUid) {
+   Process_uidDigits = ceil(log10(maxUid));
```
Explorer09, line 57: I assume log10 would be expensive, so how about avoiding it when the UID is less than 5 digits?
```c
   if (maxUid < (unsigned int)pow(10, 5)) {
      Process_uidDigits = 5;
      return;
   }
```

Process.h
```diff
@@ -339,6 +341,9 @@ static inline bool Process_isThread(const Process* this) {
 
 void Process_setupColumnWidths(void);
 
+/* Sets the size of the UID column based on the passed UID */
+void Process_setUidColumnWidth(unsigned int maxUID);
```
Explorer09, line 345:
```suggestion
void Process_setUidColumnWidth(unsigned int maxUid);
```

ProcessList.c
```diff
@@ -105,13 +105,19 @@ static const char* alignedProcessFieldTitle(const ProcessList* this, ProcessFiel
    if (!title)
       return "- ";
 
-   if (!Process_fields[field].pidColumn)
-      return title;
+   if (Process_fields[field].pidColumn) {
+      static char titleBuffer[PROCESS_MAX_PID_DIGITS + /* space */ 1 + /* null-terminator */ + 1];
```
Explorer09, line 109: There are two plus signs in this line. I wonder why the compiler didn't warn or fail to build this.
Can we correct it as well? I know it's present in the old code.
```
static char titleBuffer[PROCESS_MAX_PID_DIGITS + /* space */ 1 + /* null-terminator */ + 1];
                                                               ^ ~~~~~~~~~~~~~~~~~~~~~~~
```

```
static char titleBuffer[PROCESS_MAX_PID_DIGITS + /* space & null terminator */ sizeof(" ")];
```

ProcessList.c
```diff
@@ -626,10 +632,16 @@ void ProcessList_scan(ProcessList* this, bool pauseProcessUpdate) {
 
    ProcessList_goThroughEntries(this, false);
 
+   uint maxUid = 0;
```
Explorer09, line 635: `uint`? Did you mean `unsigned int`?

Process.c
BenBE: In that case you could even shortcut that calculation alltogether:

```c
   if (maxUid < 100000) {
      Process_uidDigits = 5;
      return;
   }
```

ProcessList.c
BenBE: > There are two plus signs in this line. I wonder why the compiler didn't warn or fail to build this.

Because providing an extra sign for a number like `1 + -5` or `1 - -5` is totally legit. Just writing `1 ++5` isn't, because that could (and is) misinterpreted as a pre-increment operator (but ` 1 + +5` is valid).

> Can we correct it as well? I know it's present in the old code.

Sure. A PR may always clean up nearby code, as long as that cleanup is performed in a separate patch (if you didn't need to touch that line in the first place anyways).

> 
> ```
> static char titleBuffer[PROCESS_MAX_PID_DIGITS + /* space */ 1 + /* null-terminator */ + 1];
>                                                                ^ ~~~~~~~~~~~~~~~~~~~~~ ^
> ```
> 
> ```
> static char titleBuffer[PROCESS_MAX_PID_DIGITS + /* space & null terminator */ sizeof(" ")];
> ```

I'm not sure I like that variant …

ProcessList.c
BenBE: Prefer to use `uint32` in this case, as we're talking about support for 32 bit UIDs and using the native `uid_t` type wouldn't work in this platform neutral section.

Process.c
Explorer09: @BenBE Actually the compiler will shortcut that for us, but I'm not sure we should define a constant `#define PROCESS_MIN_UID_DIGITS 5` for this purpose. Any opinion?

ProcessList.c
Explorer09: So the second "+" represents positive sign. I get it now, but I would expect a warning anyway unless the plus sign is expanded from a macro or the number with the sign is parenthesized.

Process.c
BenBE: If the `maxUid < pow(10, PROCESS_MIN_UID_DIGITS+1)` version is shortcutted by the compiler to produce effectively my suggestion from above, then I'd prefer that one, as it keeps that condition in sync with all the moving parts.

Explorer09: It should be `maxUid < pow(10, PROCESS_MIN_UID_DIGITS)`. Because `pow(10, PROCESS_MIN_UID_DIGITS)` is 100000 and has 6 digits.

BenBE: Was still half asleep …

ProcessList.c
silkeh: Why not remove the inline comment entirely, and use the `sizeof` @Explorer09 suggested?
```
static char titleBuffer[PROCESS_MAX_PID_DIGITS + sizeof(" ")];
```

This already contains the information about the space and the null terminator, and I think is actually clearer than the version with comments.

ProcessList.c
silkeh: > uint? Did you mean unsigned int?

Yep :disappointed_relieved: 

> Prefer to use uint32_t in this case, as we're talking about support for 32 bit UIDs and using the native uid_t type wouldn't work in this platform neutral section.

Looks like `uid_t` is already used here, so I have used that.

Process.c
silkeh: Good point! Changed.

Process.h
silkeh: Fixed!

silkeh:
I have pushed a fixup commit incorporating the comments (except for the space & null terminator discussion). If there's consensus there I will add a commit for that, and squash the fixup.

Process.h
```diff
@@ -281,6 +281,7 @@ int Process_compare(const void* v1, const void* v2);
 void Process_delete(Object* cast);
 extern const ProcessFieldData Process_fields[LAST_PROCESSFIELD];
 #define PROCESS_MAX_PID_DIGITS 19
+#define PROCESS_MIN_UID_DIGITS 5
 #define PROCESS_MAX_UID_DIGITS 19
```
BenBE: Analog for PIDs too?

```suggestion
#define PROCESS_MIN_PID_DIGITS 7
#define PROCESS_MAX_PID_DIGITS 19
#define PROCESS_MIN_UID_DIGITS 5
#define PROCESS_MAX_UID_DIGITS 19
```

BenBE (approved):
One minor suggestion, apart from this LGTM.

BenBE (approved):

Process.h
silkeh: I added a commit with a minimum number of PID digits, but setting this to 7 actually increases the size on MacOS/FreeBSD/DragonFlyBSD (which have a default of 99999). Maybe this should be 5 as well?

ProcessList.c
silkeh: I added a commit with this change

Process.h
BenBE: If its determined dynamically anyway keeping it at 5 won't hurt.

silkeh: Ok! I updated the PR with `PROCESS_MIN_PID_DIGITS` set to `5`.

ProcessList.c
```diff
@@ -647,6 +659,9 @@ void ProcessList_scan(ProcessList* this, bool pauseProcessUpdate) {
       }
    }
 
+   // Set UID column width based on max UID.
+   Process_setUidColumnWidth(maxUid);
```
cgzones, line 663: Do we need to call this every cycle or just when `maxUid` changed, e.g. after line 642?

silkeh: This needs to be called once the (final) `maxUid` has been found. Putting it on line 642 would result in it being called multiple times during the scan.

BenBE:
As I can see this, the 4 commits can probably be merged/squashed into one commit to have a short&clean history for this change.

Issue #672, opened 2021-06-30
Feature request: show open files' read/write position


pjvandehaar:
In the open files view (`l`), I want to see a column of file read positions, like the `OFFSET` column of `lsof -o -p $pid`.  

Currently, htop shows `SIZE` like `lsof -s -p $pid`.  Could htop show both SIZE and OFFSET?

I'm happy to work on this if you would accept a PR.

BenBE:
We're always looking forward to receiving a PR for open issues. If you need any assistance or help with the source feel free to ask.

PR #750, opened 2021-08-20
Add OFFSET column to OpenFilesScreen

pjvandehaar:
I had to use `stat()` to get file size because `lsof -o -F -p <pid>` doesn't show the file size on linux.

Fixes https://github.com/htop-dev/htop/issues/672

### DIFF
--- a/OpenFilesScreen.c
+++ b/OpenFilesScreen.c
@@ -10,12 +10,14 @@ in the source distribution for its full text.
 #include "OpenFilesScreen.h"
 
 #include <fcntl.h>
+#include <inttypes.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
 #include <sys/types.h>
 #include <sys/wait.h>
+#include <sys/stat.h>
 
 #include "Macros.h"
 #include "Panel.h"
@@ -124,6 +126,7 @@ static OpenFiles_ProcessData* OpenFilesScreen_getProcessData(pid_t pid) {
 
    OpenFiles_Data* item = &(pdata->data);
    OpenFiles_FileData* fdata = NULL;
+   bool lsofIncludesFileSize = false;
 
    FILE* fd = fdopen(fdpair[0], "r");
    if (!fd) {
@@ -185,6 +188,10 @@ static OpenFiles_ProcessData* OpenFilesScreen_getProcessData(pid_t pid) {
          /* ignore */
          break;
       }
+
+      if (cmd == 's')
+         lsofIncludesFileSize = true;
+
       free(line);
    }
    fclose(fd);
@@ -199,6 +206,25 @@ static OpenFiles_ProcessData* OpenFilesScreen_getProcessData(pid_t pid) {
       pdata->error = 1;
    } else {
       pdata->error = WEXITSTATUS(wstatus);
+   }
+
+   /* We got all information we need; no post-processing needed */
+   if (lsofIncludesFileSize)
+      return pdata;
+
+   /* On linux, `lsof -o -F` omits SIZE, so add it back. */
+   /* On macOS, `lsof -o -F` includes SIZE, so this block isn't needed.  If no open files have a filesize, this will still run, unfortunately. */
+   size_t fileSizeIndex = getIndexForType('s');
+   for (fdata = pdata->files; fdata != NULL; fdata = fdata->next) {
+      item = &fdata->data;
+      const char* filename = getDataForType(item, 'n');
+
+      struct stat st;
+      if (stat(filename, &st) == 0) {
+         char fileSizeBuf[21]; /* 20 (long long) + 1 (NULL) */
+         xSnprintf(fileSizeBuf, sizeof(fileSizeBuf), "%"PRIu64, st.st_size); /* st.st_size is long long on macOS, long on linux */
+         free_and_xStrdup(&item->data[fileSizeIndex], fileSizeBuf);
+      }
    }
 
    return pdata;
--- a/OpenFilesScreen.c
+++ b/OpenFilesScreen.c
@@ -155,11 +155,19 @@ static OpenFiles_ProcessData* OpenFilesScreen_getProcessData(pid_t pid) {
       case 'n':  /* file name, comment, Internet address */
       case 's':  /* file's size */
       case 't':  /* file's type */
+      {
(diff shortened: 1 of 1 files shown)


OpenFilesScreen.c
```diff
@@ -191,6 +197,23 @@ static OpenFiles_ProcessData* OpenFilesScreen_getProcessData(pid_t pid) {
...
+     item = &(fdata->data);
+     char* filename = item->data[filenameIndex];
+     struct stat st;
+     if (stat(filename, &st) == 0) {
+       char sizeChar[11]; /* 10 + 1 (NULL) */
+       xSnprintf(sizeChar, sizeof(sizeChar), "%10lld", (long long)st.st_size); /* st.st_size is long long on macOS, long on linux */
+       item->data[sizeIndex] = xStrdup(sizeChar);
+     }
+     fdata = fdata->next;
+   }
```
BenBE, line 215: Indentation

OpenFilesScreen.c
```diff
@@ -191,6 +197,23 @@ static OpenFiles_ProcessData* OpenFilesScreen_getProcessData(pid_t pid) {
...
+   fdata = pdata->files;
+   size_t sizeIndex = getIndexForType('s');
+   size_t filenameIndex = getIndexForType('n');
+   while (fdata) {
+     item = &(fdata->data);
+     char* filename = item->data[filenameIndex];
+     struct stat st;
+     if (stat(filename, &st) == 0) {
+       char sizeChar[11]; /* 10 + 1 (NULL) */
+       xSnprintf(sizeChar, sizeof(sizeChar), "%10lld", (long long)st.st_size); /* st.st_size is long long on macOS, long on linux */
```
BenBE, line 211: `%10lld` (uint64_t) can be larger than 10 characters.

OpenFilesScreen.c
```diff
@@ -191,6 +197,23 @@ static OpenFiles_ProcessData* OpenFilesScreen_getProcessData(pid_t pid) {
...
+       item->data[sizeIndex] = xStrdup(sizeChar);
```
BenBE, line 212:
```suggestion
       free_and_xStrdup(&item->data[sizeIndex], sizeChar);
```

BenBE (changes requested):
`lsof` is FUBAR.

OpenFilesScreen.c
```diff
@@ -115,7 +118,7 @@ static OpenFiles_ProcessData* OpenFilesScreen_getProcessData(pid_t pid) {
       close(fdnull);
       char buffer[32] = {0};
       xSnprintf(buffer, sizeof(buffer), "%d", pid);
-      execlp("lsof", "lsof", "-P", "-p", buffer, "-F", NULL);
+      execlp("lsof", "lsof", "-P", "-o", "-p", buffer, "-F", NULL);
```
BenBE, line 121: This is <s>broken</s> FUBAR in `lsof`:
[TW: Atrocious code, click at your own risk! Cannot be unseen!](https://github.com/lsof-org/lsof/blob/234ac9ca6082084640cb1b762f6721eccc993c79/print.c#L1003-L1074)

OpenFilesScreen.c
pjvandehaar: 20 characters seems like too much for common terminal widths.  16 characters is enough for ext4’s max file size of 16TB. Or should I calculate what width is needed?

fasterit: Make it 20. 'cause everything else can be used to overrun the buffer. Safe coding is always preferred.

pjvandehaar:
I think I addressed all those issues now.  Sorry about the careless memory leak.

pjvandehaar:
I made the suggested changes, but github doesn't consider them done yet, so I'm re-requesting a review.

OpenFilesScreen.c
```diff
@@ -155,8 +158,17 @@ static OpenFiles_ProcessData* OpenFilesScreen_getProcessData(pid_t pid) {
...
+         break;
+      }
+      case 'o':  /* file's offset */
+      {
+         size_t index = getIndexForType(cmd);
+         if (String_startsWith(line+1, "0t")) {
+           free_and_xStrdup(&item->data[index], line+3);
+         } else {
+           free_and_xStrdup(&item->data[index], line+1);
+         }
```
BenBE, line 171: Indentation

BenBE (changes requested):

pjvandehaar:
Do you prefer that I rebase to make 3 clean commits, or 1 commit?

BenBE:
> Do you prefer that I rebase to make 3 clean commits, or 1 commit?

Please squash the cleanup commits into their respective feature commits. If I did look correctly this should leave 3 commits (add OFFSET column, strip prefix, resolve stat). TIA.

OpenFilesScreen.c
```diff
@@ -201,6 +202,22 @@ static OpenFiles_ProcessData* OpenFilesScreen_getProcessData(pid_t pid) {
...
+   /* On linux, `lsof -o -F` omits SIZE, so add it back. */
+   /* On macOS, `lsof -o -F` includes SIZE, but overwrite it for consistency. */
+   fdata = pdata->files;
+   size_t sizeIndex = getIndexForType('s');
+   size_t filenameIndex = getIndexForType('n');
+   while (fdata) {
+      item = &fdata->data;
+      struct stat st;
+      if (stat(item->data[filenameIndex], &st) == 0) {
+         char sizeChar[21]; /* 20 (long long) + 1 (NULL) */
```
BenBE, line 214: `sizeChar` is not a good identifier for this variable. Something a bit more descriptive like `fileSizeBuf` or `fileSizeStr` is IMHO better.

pjvandehaar: I changed this.  Thanks for all the reviews to try to make this code decent!  I should have noticed that this variable name didn't fit in earlier.

BenBE (approved):

BenBE:
Any chance to update the patch so it skips re-reading the file size via `stat` when it is already present in one record? That way we skip both the work on MacOS and allow for future versions of `lsof` to properly provide this information (currently I'm looking into writing a patch for `lsof` that would add this information naturally for Linux too).

pjvandehaar:
Thanks Ben and CGZones for your help– I’m excited for this feature!

PR #752, opened 2021-08-20
Fix macOS CPU time calculations

### LINKED ISSUE #368: CPU usage incorrect on macOS Big Sur with M1 chip

### LINKED ISSUE #751: Total CPU time is broken on M1 macs
#368 fixed the CPU percentages for M1 macs, but it didn't correctly address the calculations of CPU times.

Notice the discrepancy compared to Activity Monitor:

Screen Shot 2021-08-20 at 7 05 42 PM

amomchilov:
Resolves #751

#368 fixed the CPU percentages for M1 macs, but it didn't correctly address the calculations of CPU times.

As it turns out, the CPU percentage calculation is wrong, but since it was a ratio, two errors cancelled each other out.

This PR fixes that.

Comparison

Some notes:

* I did my development/testing on an M1 Mac Mini, but I also confirmed I haven't introduced regressions by double checking my final work against my Intel-powered MacBook Pro.
* I ran an infinite loop in a Swift REPL to peg one CPU core to 100%. I used this as a test-case to confirm CPU percentages were correct.
* I used Activity Monitor as a source-of-truth to confirm CPU times were working correctly.
* I set the "delay" to 10 (with `htop -d 10`, i.e. the refresh rate should be once per second, to make it easier to debug the `time_interval` mistakes.

With the existing code:

* The value of `time_interval` in `DarwinProcess_setFromLibprocPidinfo ` was wrong. Its variable name is useless, but it looks like it should nominally be a value in nanoseconds.

	* Its value was around `24090000.000000` nanoseconds (~`0.024` seconds).
	* It was expected to roughly `1e9` (around one second), since I was running with `-d 10` and I had approx one core at 100% and not much else.
	* It was off by a factor of `125/3` (`41.666`), our beloved `Platform_timebaseToNS`!

* The value of `total_current_time` in `DarwinProcess_setFromLibprocPidinfo` was wrong. It seems like it should also be a value in nanoseconds
	* Its value was around `1,147,173`, at a time when the correct value was around `43,752,916 ns` (a factor of `38.1`, which is close to our beloved `41.666`)

### DIFF
--- a/darwin/DarwinProcess.c
+++ b/darwin/DarwinProcess.c
@@ -269,6 +269,20 @@ ERROR_A:
    Process_updateCmdline(proc, k->kp_proc.p_comm, 0, strlen(k->kp_proc.p_comm));
 }
 
+// Converts ticks in the Mach "timebase" to nanoseconds.
+// See `mach_timebase_info`, as used to define the `Platform_timebaseToNS` constant.
+static uint64_t machTicksToNanoseconds(uint64_t schedulerTicks) {
+   double nanoseconds_per_mach_tick = Platform_timebaseToNS;
+   return (uint64_t) (nanoseconds_per_mach_tick * (double) schedulerTicks);
+}
+
+// Converts nanoseconds to hundreths of a second (centiseconds) as needed by the "time" field of the Process struct.
+static long long int nanosecondsToCentiseconds(uint64_t nanoseconds) {
+   uint64_t centiseconds_per_second = 100;
+   uint64_t nanoseconds_per_second = 1e9;
+   return nanoseconds / nanoseconds_per_second * centiseconds_per_second;
+}
+
 void DarwinProcess_setFromKInfoProc(Process* proc, const struct kinfo_proc* ps, bool exists) {
    DarwinProcess* dp = (DarwinProcess*)proc;
 
@@ -330,21 +344,25 @@ void DarwinProcess_setFromKInfoProc(Process* proc, const struct kinfo_proc* ps, 
    proc->updated = true;
 }
 
-void DarwinProcess_setFromLibprocPidinfo(DarwinProcess* proc, DarwinProcessList* dpl, double time_interval) {
+void DarwinProcess_setFromLibprocPidinfo(DarwinProcess* proc, DarwinProcessList* dpl, double time_interval_ns) {
    struct proc_taskinfo pti;
 
    if (sizeof(pti) == proc_pidinfo(proc->super.pid, PROC_PIDTASKINFO, 0, &pti, sizeof(pti))) {
-      uint64_t total_existing_time = proc->stime + proc->utime;
-      uint64_t total_current_time = pti.pti_total_system + pti.pti_total_user;
+      uint64_t total_existing_time_ns = proc->stime + proc->utime;
 
-      if (total_existing_time && 1E-6 < time_interval) {
-         uint64_t total_time_diff = total_current_time - total_existing_time;
-         proc->super.percent_cpu = ((double)total_time_diff / time_interval) * 100.0;
+      uint64_t user_time_ns = machTicksToNanoseconds(pti.pti_total_user);
+      uint64_t system_time_ns = machTicksToNanoseconds(pti.pti_total_system);
+
+      uint64_t total_current_time_ns = user_time_ns + system_time_ns;
+
+      if (total_existing_time_ns && 1E-6 < time_interval_ns) {
+         uint64_t total_time_diff_ns = total_current_time_ns - total_existing_time_ns;
+         proc->super.percent_cpu = ((double)total_time_diff_ns / time_interval_ns) * 100.0;
       } else {
          proc->super.percent_cpu = 0.0;
       }
 
-      proc->super.time = total_current_time / 10000000;
+      proc->super.time = nanosecondsToCentiseconds(total_current_time_ns);
       proc->super.nlwp = pti.pti_threadnum;
       proc->super.m_virt = pti.pti_virtual_size / ONE_K;
       proc->super.m_resident = pti.pti_resident_size / ONE_K;
@@ -352,8 +370,8 @@ void DarwinProcess_setFromLibprocPidinfo(DarwinProcess* proc, DarwinProcessList*
       proc->super.percent_mem = (double)pti.pti_resident_size * 100.0
                               / (double)dpl->host_info.max_mem;
 
-      proc->stime = pti.pti_total_system;
-      proc->utime = pti.pti_total_user;
+      proc->stime = system_time_ns;
+      proc->utime = user_time_ns;
 
       dpl->super.kernelThreads += 0; /*pti.pti_threads_system;*/
       dpl->super.userlandThreads += pti.pti_threadnum; /*pti.pti_threads_user;*/
--- a/darwin/DarwinProcess.c
+++ b/darwin/DarwinProcess.c
@@ -278,8 +278,8 @@ static uint64_t machTicksToNanoseconds(uint64_t schedulerTicks) {
 
 // Converts nanoseconds to hundreths of a second (centiseconds) as needed by the "time" field of the Process struct.
 static long long int nanosecondsToCentiseconds(uint64_t nanoseconds) {
-   uint64_t centiseconds_per_second = 100;
-   uint64_t nanoseconds_per_second = 1e9;
+   const uint64_t centiseconds_per_second = 100;
+   const uint64_t nanoseconds_per_second = 1e9;
    return nanoseconds / nanoseconds_per_second * centiseconds_per_second;
 }
 
@@ -344,7 +344,7 @@ void DarwinProcess_setFromKInfoProc(Process* proc, const struct kinfo_proc* ps, 
    proc->updated = true;
 }
 
-void DarwinProcess_setFromLibprocPidinfo(DarwinProcess* proc, DarwinProcessList* dpl, double time_interval_ns) {
+void DarwinProcess_setFromLibprocPidinfo(DarwinProcess* proc, DarwinProcessList* dpl, double timeIntervalNS) {
    struct proc_taskinfo pti;
 
    if (sizeof(pti) == proc_pidinfo(proc->super.pid, PROC_PIDTASKINFO, 0, &pti, sizeof(pti))) {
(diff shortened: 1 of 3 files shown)


darwin/DarwinProcess.c
```diff
@@ -330,30 +344,34 @@ void DarwinProcess_setFromKInfoProc(Process* proc, const struct kinfo_proc* ps,
...
-         proc->super.percent_cpu = ((double)total_time_diff / time_interval) * 100.0;
+      if (total_existing_time_ns && 1E-6 < time_interval_ns) {
+         uint64_t total_time_diff_ns = total_current_time_ns - total_existing_time_ns;
+         proc->super.percent_cpu = ((double)total_time_diff_ns / time_interval_ns) * 100.0;
       } else {
          proc->super.percent_cpu = 0.0;
       }
 
-      proc->super.time = total_current_time / 10000000;
+      proc->super.time = nanosecondsToCentiseconds(total_current_time_ns);
```
amomchilov, line 365: These magic numbers are absolute hell. Stop the madness.

darwin/DarwinProcess.c
```diff
@@ -330,30 +344,34 @@ void DarwinProcess_setFromKInfoProc(Process* proc, const struct kinfo_proc* ps,
    proc->updated = true;
 }
 
-void DarwinProcess_setFromLibprocPidinfo(DarwinProcess* proc, DarwinProcessList* dpl, double time_interval) {
+void DarwinProcess_setFromLibprocPidinfo(DarwinProcess* proc, DarwinProcessList* dpl, double time_interval_ns) {
```
amomchilov, line 347: I started incorporating units into the names of variables. Given that there's no proper higher-order "Measurement" class/abstraction that manages this, I figure Hungarian notation is the next best thing.

darwin/DarwinProcess.c
```diff
@@ -330,30 +344,34 @@ void DarwinProcess_setFromKInfoProc(Process* proc, const struct kinfo_proc* ps,
...
    struct proc_taskinfo pti;
 
    if (sizeof(pti) == proc_pidinfo(proc->super.pid, PROC_PIDTASKINFO, 0, &pti, sizeof(pti))) {
-      uint64_t total_existing_time = proc->stime + proc->utime;
-      uint64_t total_current_time = pti.pti_total_system + pti.pti_total_user;
+      uint64_t total_existing_time_ns = proc->stime + proc->utime;
+
+      uint64_t user_time_ns = machTicksToNanoseconds(pti.pti_total_user);
+      uint64_t system_time_ns = machTicksToNanoseconds(pti.pti_total_system);
```
amomchilov, line 354: The values of `pti_total_user` and `pti_total_system` that comeback from Libproc (specifically `proc_pidinfo` above) are in mach timebase ticks.

darwin/DarwinProcessList.c
```diff
@@ -192,7 +194,7 @@ void ProcessList_goThroughEntries(ProcessList* super, bool pauseProcessUpdate) {
       }
    }
 
-   const double time_interval = ticksToNanoseconds(dpl->global_diff) / (double) dpl->super.activeCPUs;
+   const double time_interval_ns = schedulerTicksToNanoseconds(dpl->global_diff) / (double) dpl->super.activeCPUs;
```
amomchilov, line 197: The values that come back from `host_processor_info` are *NOT* in Mach timebase tick units.

They're in "scheduler ticks" (as I call them), which are still the same 100 ticks/second on M1 as they used to be on Intel Macs.

darwin/DarwinProcessList.c
```diff
@@ -160,9 +160,11 @@ void ProcessList_delete(ProcessList* this) {
    free(this);
 }
 
-static double ticksToNanoseconds(const double ticks) {
+// Converts "scheduler ticks" to nanoseconds.
+// See `sysconf(_SC_CLK_TCK)`, as used to define the `Platform_clockTicksPerSec` constant.
+static double schedulerTicksToNanoseconds(const double ticks) {
    const double nanos_per_sec = 1e9;
-   return (ticks / Platform_timebaseToNS) * (nanos_per_sec / (double) Platform_clockTicksPerSec);
+   return ticks * (nanos_per_sec / (double) Platform_clockTicksPerSec);
```
amomchilov, line 167: The values that come back from `host_processor_info` are *NOT* in Mach timebase tick units.

They're in "scheduler ticks" (as I call them), which are still the same 100 ticks/second on M1 as they used to be on Intel Macs.

Converting with `Platform_timebaseToNS` here was mistake, but it cancels out after division, ended up making the percentage correct, but not the "CPU time" stat.

amomchilov: Given that this conversion function is the only use of `Platform_clockTicksPerSec`, I suggest we refactor:

1. Hide the `Platform_clockTicksPerSec` constant from `Platform.h`
2. Move `schedulerTicksToNanoseconds` to `Platform.c`
3. Expose `schedulerTicksToNanoseconds` from `Platform.h`, and call that from here.

darwin/DarwinProcess.c
```diff
@@ -269,6 +269,20 @@ static void DarwinProcess_updateCmdLine(const struct kinfo_proc* k, Process* pro
    Process_updateCmdline(proc, k->kp_proc.p_comm, 0, strlen(k->kp_proc.p_comm));
 }
 
+// Converts ticks in the Mach "timebase" to nanoseconds.
+// See `mach_timebase_info`, as used to define the `Platform_timebaseToNS` constant.
+static uint64_t machTicksToNanoseconds(uint64_t schedulerTicks) {
+   double nanoseconds_per_mach_tick = Platform_timebaseToNS;
+   return (uint64_t) (nanoseconds_per_mach_tick * (double) schedulerTicks);
+}
```
amomchilov, line 277: Similar to `Platform_clockTicksPerSec`, this conversion function is the only usage of `Platform_timebaseToNS`. I suggest we refactor:

1. Hide the `Platform_timebaseToNS` constant from `Platform.h`
2. Move `machTicksToNanoseconds` to `Platform.c`
3. Expose `machTicksToNanoseconds` from `Platform.h`, and call that from here.

amomchilov:
I can't request reviews, so I'm using mentions instead.

CC: @BenBE @cgzones @natoscott

darwin/DarwinProcessList.c
amomchilov: P.s. accidental trailing space here. I'll remove it with my "touch up" commit.

BenBE: > The values that come back from `host_processor_info` are _NOT_ in Mach timebase tick units.
> 
> They're in "scheduler ticks" (as I call them), which are still the same 100 ticks/second on M1 as they used to be on Intel Macs.

:fire: :fire: :fire:

BenBE: > P.s. accidental trailing space here. I'll remove it with my "touch up" commit.

Just amend/squash them in …

darwin/DarwinProcess.c
amomchilov: @BenBE Here's what the looks like https://github.com/amomchilov/htop/pull/1/files

Once this PR ships, I'll open a proper PR on this repo for that.

darwin/DarwinProcess.c
```diff
@@ -269,6 +269,20 @@ static void DarwinProcess_updateCmdLine(const struct kinfo_proc* k, Process* pro
...
+
+// Converts nanoseconds to hundreths of a second (centiseconds) as needed by the "time" field of the Process struct.
+static long long int nanosecondsToCentiseconds(uint64_t nanoseconds) {
+   uint64_t centiseconds_per_second = 100;
```
natoscott, line 281: Can prefix with 'const' here I think.

darwin/DarwinProcess.c
```diff
@@ -269,6 +269,20 @@ static void DarwinProcess_updateCmdLine(const struct kinfo_proc* k, Process* pro
...
+   uint64_t nanoseconds_per_second = 1e9;
```
natoscott, line 282: And here.

darwin/DarwinProcess.c
natoscott: Camel-case is the overriding style used for variables and parameters throughout htop, I'd stick with that here too.

natoscott (changes requested):
Thanks for working on this, what a mess this whole area has become on darwin - great to see it improving.

darwin/DarwinProcess.c
amomchilov: How thoroughly should I `const` things up? A whole bunch of local vars in `DarwinProcess_setFromLibprocPidinfo` could be made const, too.

amomchilov:
> Thanks for working on this, what a mess this whole area has become on darwin - great to see it improving.

It really has been. The clarity of the code is terrible. Cryptic short identifiers, no communication of expected units of values, magic numbers which aren't just conversion factors: they're a precomputed combination of multiple conversion factors.

If only you knew how long it took me to figure out that `proc->super.time = total_current_time / 10000000;` is converting ns to cs, and that the `10000000` is really just a pre-computation of `1e9 / 100`

I've made the requests you requested, take a look: https://github.com/htop-dev/htop/pull/752/commits/7a4d6fa4098b81a09df00a26aa15bf8768fade2a

In your code review, were you only reviewing the code at a glance, or did you also have a chance to run it and test it?

BenBE:
> In your code review, were you only reviewing the code at a glance, or did you also have a chance to run it and test it?

Lacking a Mac I mostly looked at the code itself. For actually running I pinged @fasterit for the review.

natoscott:
> > In your code review, were you only reviewing the code at a glance, or did you also have a chance to run it and test it?
> 
> Lacking a Mac I mostly looked at the code itself. For actually running I pinged @fasterit for the review.

Same here, I'm only able to look over the code here & offer emotional support.  :-)

darwin/DarwinProcess.c
natoscott: > How thoroughly should I `const` things up? A whole bunch of local vars in `DarwinProcess_setFromLibprocPidinfo` could be made const, too.

For now, just use const where it makes sense in the places you're changing/adding code.  Over time it'd be good to move other darwin/ code towards wider use of const like has been done in the htop core code.

fasterit (approved):
This patch doesn't the displayed value of "TIME+" for me on an Intel Mac. Which means it is probably good to go.

```
$ sw_vers
ProductName:	macOS
ProductVersion:	11.5.1
BuildVersion:	20G80
```

darwin/DarwinProcessList.c
amomchilov: > Given that this conversion function is the only use of `Platform_clockTicksPerSec`, I suggest we refactor:
> 
>     1. Hide the `Platform_clockTicksPerSec` constant from `Platform.h`
> 
>     2. Move `schedulerTicksToNanoseconds` to `Platform.c`
> 
>     3. Expose `schedulerTicksToNanoseconds` from `Platform.h`, and call that from here.

Done: https://github.com/htop-dev/htop/pull/766

darwin/DarwinProcess.c
amomchilov: > Similar to `Platform_clockTicksPerSec`, this conversion function is the only usage of `Platform_timebaseToNS`. I suggest we refactor:
> 
>     1. Hide the `Platform_timebaseToNS` constant from `Platform.h`
> 
>     2. Move `machTicksToNanoseconds` to `Platform.c`
> 
>     3. Expose `machTicksToNanoseconds` from `Platform.h`, and call that from here.

Done: https://github.com/htop-dev/htop/pull/766

PR #754, opened 2021-08-21
Linux: add combined memory, swap and zram meter

cgzones:


BenBE (approved):
LGTM.

I don't quite see the usecase here yet … (In contrast to the Mem+Swap meter)

fasterit:
Discussed this during the team meeting, seems a very small gain for yet another memory meter. So we'd lean towards not merging this one unless there is some rationale.

natoscott:
@cgzones we could add this as a config below pcp/meters/ if that helps?  No new code needed then - 'pminfo zram' will give you a list of the zram metric names.

PR #756, opened 2021-08-21
Fix glitches after adding or deleteing meters on resize

cgzones:
Fixes: https://www.youtube.com/watch?v=DVapUamvqO8&t=605s


BenBE (approved):
LGTM.

Issue #675, opened 2021-07-03
htop seems to count tmpfs into Disk IO


toralf:
I do bind-mount tmpfs here to my Gentoo images via bubblewrap [1]. It seems that it IO is counted by htop too (but shouldn't IMO)

[1] https://github.com/toralf/tinderbox/blob/master/bin/bwrap.sh#L139

BenBE:
Given there are numerous ways that fiels could be stored by a process to persist data, there's not much you can do except for special-casing each and every file system that could exist. While `tmpfs` is a quite obvious example of files just going into RAM there are examples like `afs` or `cifs` where it's not so obvious if you should call it still "Disk I/O" as targets of these file systems are networked systems. Even more complicated are cases like `fuse` where the high-level file system driver doesn't even give you a hint where things are going. They could be classical disk file systems like `ntfs3g` (NTFS), wrappers around other I/O like `encfs` or purely virtual ones like `ramfuse` and `nullfs`. Does storing to things like ntfs3g create I/O? Clearly yes. Actually double, because the kernel has to handle each block of data usually twice (once using FUSE and once for the underlaying block device. In this regard having `tmpfs` being counted as normal I/O is actually quite straight forward as it's just handling whatever data the kernel hands it and writes that into RAM. If you wanted to split that up by target device this would get complicated quite fast as this information is most likely not readily available on most systems.

cgzones:
Is this about the DiskIO meter in the top header or the process coulmns, like *DISK WRITE*, *DISK READ*?

toralf:
The former (meter in top header)

cgzones:
Can you share the output of `/proc/diskstats` while the tmpfs filesystems are mounted?

toralf:
`mr-fox ~ # cat /proc/diskstats
 259       0 nvme1n1 3863172 1200583 275328179 628920 53427106 1894831 3207270464 284059824 0 8403284 284688744 0 0 0 0 0 0
 259       2 nvme1n1p1 62 0 4904 14 0 0 0 0 0 28 14 0 0 0 0 0 0
 259       3 nvme1n1p2 291 0 16269 30 1 0 8 0 0 64 30 0 0 0 0 0 0
 259       4 nvme1n1p3 49362 27304 4589856 10611 563101 18148 52857320 30635580 0 92984 30646191 0 0 0 0 0 0
[...]

PR #758, opened 2021-08-21
Linux: rework disk-io parsing

cgzones:
Generalize sub-diskname handling, like sdb1/sdb2, to not count the
usage twice with the aggregate top-diskname, like sdb.
Rely on /proc/diskstats being ordered, e.g. no sub-diskname precedes its
top-diskname.

Closes: #675

### DIFF
--- a/linux/Platform.c
+++ b/linux/Platform.c
@@ -543,6 +543,8 @@ bool Platform_getDiskIO(DiskIOData* data) {
    if (!fd)
       return false;
 
+   char lastTopDisk[32] = { '\0' };
+
    unsigned long long int read_sum = 0, write_sum = 0, timeSpend_sum = 0;
    char lineBuffer[256];
    while (fgets(lineBuffer, sizeof(lineBuffer), fd)) {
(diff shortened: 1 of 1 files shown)


toralf:
It doesn't apply against sys-process/htop-3.0.5-r1 here at a stable hardened Gentoo.

cgzones:
> It doesn't apply against sys-process/htop-3.0.5-r1 here at a stable hardened Gentoo.

Probably missing commit 6974ce8e is the cause.

BenBE (approved):

linux/Platform.c
```diff
@@ -555,23 +557,10 @@ bool Platform_getDiskIO(DiskIOData* data) {
...
-         if (diskname[0] == 'm'
-             && diskname[1] == 'm'
-             && diskname[2] == 'c'
-             && diskname[3] == 'b'
-             && diskname[4] == 'l'
-             && diskname[5] == 'k'
-             && isdigit((unsigned char)diskname[6])
-             && diskname[7] == 'p')
-            continue;
+         if (lastTopDisk[0] && String_startsWith(diskname, lastTopDisk))
```
BenBE, line 560: Document the implicit requirement of disks being reported before their partitions in `/proc/diskstats`.

PR #763, opened 2021-08-24
HeaderLayout: save special ID in configuration

cgzones:
Use a special ID in the user configuration file instead of the compile
time enum value, so that future reorderings or insertions do not change
the user selected layout.


BenBE (approved):

fasterit:
I think we agreed in the past that we want to use mnemonic config formats over numeric ones in the future, so HF_TWO_50_50 or TWO_50_50 over 0. Because then - in this example - the sorting of options doesn't matter anymore.

PR #767, opened 2021-08-28
Misc rc1 improvements

cgzones:


linux/LinuxProcessList.c
```diff
@@ -1433,7 +1433,7 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
 
             uint64_t recheck = ((uint64_t)rand()) % 2048;
 
-            if (passedTimeInMs > 2000 || passedTimeInMs > recheck) {
+            if (passedTimeInMs > recheck) {
```
BenBE, line 1436: This is intentional as refreshs of that metric are randomized otherwise and this ensures some freshnes.

Settings.c
```diff
@@ -541,6 +541,9 @@ Settings* Settings_new(unsigned int initialCpuCount, Hashtable* dynamicColumns)
       this->changed = true;
       Settings_read(this, SYSCONFDIR "/htoprc", initialCpuCount);
    }
+   if (!ok) {
+      Settings_defaultMeters(this, initialCpuCount);
+   }
```
BenBE, line 546: What's the reason here?

.travis.yml
```diff
@@ -9,7 +9,8 @@ os:
...
 script:
   - ./autogen.sh
-  # clang might warn about C11 generic selections in isnan()
-  - CFLAGS=-Wno-c11-extensions ./configure --enable-werror
+  - ./configure --enable-werror
   - make -k
-  - CFLAGS=-Wno-c11-extensions make distcheck DISTCHECK_CONFIGURE_FLAGS=--enable-werror
+  - make distcheck DISTCHECK_CONFIGURE_FLAGS=--enable-werror
+  - sudo make install
+  - make installcheck
```
BenBE, line 16: What does this change?

Settings.c
cgzones: Try running `HTOPRC=./htoprc_new ./htop`.
The first time the header is empty, cause no default meters are added.
The second time meters are added, cause a configuration file already exists so a different path is taken.

linux/LinuxProcessList.c
cgzones: `recheck` is guaranteed to be at maximum `2047`.

.travis.yml
cgzones: `-Wno-c11-extensions` is added via the configure script.

`make install` and `make installcheck` ensures installing htop works.

linux/LinuxProcessList.c
BenBE: This slightly influences the distribution for the case where recheck has a value where the entry will now not be updated in a certain time window. This change is negligible, but that's why I asked what motivated this change now.

PR #769, opened 2021-08-30
Misc NetBSD update

cgzones:
* Color process state P as running
  On NetBSD state 'R' means runnable not running.

  Improve the color identifier name accordingly.

* Silence dropping const qualifier on define_key(3)
  On NetBSD define_key(3) has the prototype

    ```c
    int define_key(char *sequence, int key_symbol);
    ```

* Simplify Platform_setMemoryValues

* Fix reading environment variables of processes

  **kvm_getenvv**(3) seems not to work with `kvm_openfiles(..., KVM_NO_FILES, ...)`

* Scale CPU frequencies

  Convert the values from sysctl(3) to megahertz.

  Use a value type of 'long int' to avoid ENOMEM failures of sysctl(3).

  Also check for `machdep.tsc_freq`.

* Update process fields

  Populate fields to ttyname, minflt, majflt and processor.

/cc @fraggerfox @alarixnia


BenBE (approved):

alarixnia:
Not sure about the `machdep.tsc_freq` stuff. It's in a different unit, the other sysctl variables are in MHz already.

BenBE (approved):
LGTM. Also with the scaling change.

cgzones:
Thanks @alarixnia for the frequency scaling feedback. Could you please give the most recent version a try?

Process.c
```diff
@@ -864,8 +864,12 @@ void Process_writeField(const Process* this, RichString* str, ProcessField field
    case STATE:
       xSnprintf(buffer, n, "%c ", this->state);
       switch (this->state) {
```
natoscott, line 866: In the long term it'd be better to have a canonical set of state values (an enum), switch all the PROCESS_N_STATE macros to descriptive names (like RUN), have platform-specific code refresh the 'state' enum.  This will keep the htop core code cleaner and more platform-neutral once again.  But at this stage of a release, I think this smaller fix is best.

natoscott (approved):
LGTM, though would be good to tidy up further for next release.

Process.c
BenBE: Can you open an issue for this so we don't forget about this? Looks like a good first issue to me.

PR #771, opened 2021-08-31
linux/Platform.c

pghvlaans:
Currently, htop reports double disk IO for nvme drives on Linux systems because there is no check to pick out the root device. Adding a check for nvme drives in Platform_getDiskIO solves the problem.


BenBE:
There's already a similar PR open, cf. #758 …

pghvlaans:
Sorry, didn't see that. Thanks!

PR #773, opened 2021-09-02
Miscellaneous stuff

cgzones:


Affinity.c
```diff
@@ -63,9 +63,9 @@ Affinity* Affinity_get(const Process* proc, ProcessList* pl) {
             Affinity_add(affinity, i);
          }
       } else {
-         unsigned int id;
-         hwloc_bitmap_foreach_begin(id, cpuset);
-         Affinity_add(affinity, id);
+         int id;
+         hwloc_bitmap_foreach_begin(id, cpuset)
+            Affinity_add(affinity, (unsigned)id);
```
BenBE, line 68: Shouldn't this be put through to `Affinity_add` too?

BenBE (commented):
LGTM. Not sure about the `Affinity_add` change if this should also be put through there too.

Affinity.c
cgzones: You mean accepting `int` in `Affinity_add()`?

`hwloc_bitmap_foreach_begin` is defined as
```c
#define hwloc_bitmap_foreach_begin(id, bitmap) \
do { \
        assert(hwloc_bitmap_weight(bitmap) != -1); \
        for (id = hwloc_bitmap_first(bitmap); \
             (unsigned) id != (unsigned) -1; \
             id = hwloc_bitmap_next(bitmap, id)) {
```

so inside the loop the loop iterator `id` is non-negativ.

BenBE: Yes. I mean updating the signature of `Affinity_add` to allow `id` to be passed as `int` and either check inside there or explicitly cast there after appropriate checks before the final assignment.

natoscott (approved):
LGTM, nice.

BenBE (approved):

PR #775, opened 2021-09-02
PCP: do not set exe to empty string

cgzones:
In case the executable is an empty string, e.g. if pcp is run by an
unprivileged user, do not set procExe to an empty value, which breaks
the formatting of the PROCEXE column and the merged-cmdline logic.


natoscott (approved):
LGTM, thanks @cgzones

PR #777, opened 2021-09-03
Add a Process_printPercentage helper routine

natoscott:
Replace several open-coded variants of percentage formatting.
This function has been ported from Hishams old 'next' branch.

### DIFF
--- a/Process.c
+++ b/Process.c
@@ -707,6 +707,24 @@ void Process_printLeftAlignedField(RichString* str, int attr, const char* conten
    RichString_appendChr(str, attr, ' ', width + 1 - columns);
 }
 
+void Process_printPercentage(float val, char* buffer, int n, int* attr) {
+   if (val >= 0) {
+      if (val < 99.9F) {
+         if (val < 0.05F) {
+            *attr = CRT_colors[PROCESS_SHADOW];
+         }
+         xSnprintf(buffer, n, "%4.1f ", val);
+      } else if (val < 999) {
+         xSnprintf(buffer, n, "%3d. ", (int)val);
+      } else {
+         xSnprintf(buffer, n, "%4d ", (int)val);
+      }
+   } else {
+      *attr = CRT_colors[PROCESS_SHADOW];
+      xSnprintf(buffer, n, " N/A ");
+   }
+}
+
 void Process_writeField(const Process* this, RichString* str, ProcessField field) {
    char buffer[256];
    size_t n = sizeof(buffer);
@@ -821,34 +839,13 @@ void Process_writeField(const Process* this, RichString* str, ProcessField field
 
       xSnprintf(buffer, n, "%4ld ", this->nlwp);
       break;
-   case PERCENT_CPU:
+   case PERCENT_CPU: Process_printPercentage(this->percent_cpu, buffer, n, &attr); break;
    case PERCENT_NORM_CPU: {
-      float cpuPercentage = this->percent_cpu;
-      if (field == PERCENT_NORM_CPU) {
-         cpuPercentage /= this->processList->activeCPUs;
-      }
-      if (cpuPercentage > 999.9F) {
-         xSnprintf(buffer, n, "%4u ", (unsigned int)cpuPercentage);
-      } else if (cpuPercentage > 99.9F) {
-         xSnprintf(buffer, n, "%3u. ", (unsigned int)cpuPercentage);
-      } else {
-         if (cpuPercentage < 0.05F)
-            attr = CRT_colors[PROCESS_SHADOW];
-
-         xSnprintf(buffer, n, "%4.1f ", cpuPercentage);
-      }
+      float cpuPercentage = this->percent_cpu / this->processList->activeCPUs;
+      Process_printPercentage(cpuPercentage, buffer, n, &attr);
       break;
    }
-   case PERCENT_MEM:
-      if (this->percent_mem > 99.9F) {
-         xSnprintf(buffer, n, "100. ");
-      } else {
-         if (this->percent_mem < 0.05F)
-            attr = CRT_colors[PROCESS_SHADOW];
-
-         xSnprintf(buffer, n, "%4.1f ", this->percent_mem);
-      }
-      break;
+   case PERCENT_MEM: Process_printPercentage(this->percent_mem, buffer, n, &attr); break;
    case PGRP: xSnprintf(buffer, n, "%*d ", Process_pidDigits, this->pgrp); break;
    case PID: xSnprintf(buffer, n, "%*d ", Process_pidDigits, this->pid); break;
    case PPID: xSnprintf(buffer, n, "%*d ", Process_pidDigits, this->ppid); break;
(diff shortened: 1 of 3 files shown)


Process.c
```diff
@@ -707,6 +707,21 @@ void Process_printLeftAlignedField(RichString* str, int attr, const char* conten
    RichString_appendChr(str, attr, ' ', width + 1 - columns);
 }
 
+void Process_printPercentage(float val, char* buffer, int n, int* attr) {
+   if (val >= 0) {
+      if (val < 100) {
```
BenBE, line 712: Shadow out values below `0.1` to have interesting values (non-zero) values stick out more.

BenBE (approved):
LGTM. Maybe minor suggestion for further centralized highlighting/markup of large values. Can be added later …

Process.c
```diff
@@ -707,6 +707,21 @@ void Process_printLeftAlignedField(RichString* str, int attr, const char* conten
...
+         xSnprintf(buffer, n, "%4.1f ", val);
+      } else if (val < 1000) {
+         xSnprintf(buffer, n, "%3d. ", (unsigned int)val);
+      } else {
+         xSnprintf(buffer, n, "%4d ", (unsigned int)val);
```
Explorer09, line 717: "%d" expects an `int` argument, not `unsigned int`.
```suggestion
         xSnprintf(buffer, n, "%3d. ", (int)val);
      } else {
         xSnprintf(buffer, n, "%4d ", (int)val);
```

Process.c
```diff
@@ -821,34 +836,13 @@ void Process_writeField(const Process* this, RichString* str, ProcessField field
 
       xSnprintf(buffer, n, "%4ld ", this->nlwp);
       break;
-   case PERCENT_CPU:
+   case PERCENT_CPU: Process_printPercentage(this->percent_cpu, buffer, n, &attr); break;
```
Explorer09, line 839: Line wrapping and indent:
```suggestion
   case PERCENT_CPU:
      Process_printPercentage(this->percent_cpu, buffer, n, &attr);
      break;
```

Process.c
```diff
@@ -821,34 +836,13 @@ void Process_writeField(const Process* this, RichString* str, ProcessField field
...
-      if (this->percent_mem > 99.9F) {
-         xSnprintf(buffer, n, "100. ");
-      } else {
-         if (this->percent_mem < 0.05F)
-            attr = CRT_colors[PROCESS_SHADOW];
-
-         xSnprintf(buffer, n, "%4.1f ", this->percent_mem);
-      }
-      break;
+   case PERCENT_MEM: Process_printPercentage(this->percent_mem, buffer, n, &attr); break;
```
Explorer09, line 845: Line wrapping and indent (let's avoid formatting case statements as one line unless all other case statements are done so):
```suggestion
   case PERCENT_MEM:
      Process_printPercentage(this->percent_mem, buffer, n, &attr);
      break;
```

Process.c
BenBE: The case statement is formatted mixed already, and since we don't need to touch this line right now I'd favor to leave it alone at this time. We may decide to have a complete reformat of this (and related) switch statements, but IMHO this would introduce too much noise right now.

Process.c
BenBE: Same here. Maybe another time in another PR, but currently I'd prefer to avoid that noise.

Process.c
cgzones: Unrelated to the previous comment, but affects the same line:

The previous code used `99.9F` as limit; consider a value of `99.999`:
- currently it is bigger than `99.9F` and printed with `%3d. "` as `100. `.
- here it is smaller than `100` and printed with `"%4.1f "` it will print `100.0 `, which is one character too much for the columns

Process.c
```diff
@@ -707,6 +707,26 @@ void Process_printLeftAlignedField(RichString* str, int attr, const char* conten
...
 
+void Process_printPercentage(float val, char* buffer, int n, int* attr) {
+   if (val >= 0) {
+      if (val < 100) {
+         if (val < 0.05F) {
+            *attr = CRT_colors[PROCESS_SHADOW];
+         }
+         xSnprintf(buffer, n, "%4.1f ", val);
+      } else if (val < 1000) {
+         *attr = CRT_colors[PROCESS_MEGABYTES];
```
cgzones, line 718: I am not sure I like coloring these number...

cgzones (commented):
The delay accounting fields are one column wider, see

```c
"%4.1f "
```

vs

```c
"%4.1f  "
```

BenBE:
> The delay accounting fields are one column wider, see
> 
> ```c
> "%4.1f "
> ```
> 
> vs
> 
> ```c
> "%4.1f  "
> ```

That'd been broken on Linux before, because those columns aren't even created with the same width:

```c
   [PERCENT_CPU] = { .name = "PERCENT_CPU", .title = "CPU% ", .description = "Percentage of the CPU time the process used in the last sampling", .flags = 0, .defaultSortDesc = true, },
   [PERCENT_NORM_CPU] = { .name = "PERCENT_NORM_CPU", .title = "NCPU%", .description = "Normalized percentage of the CPU time the process used in the last sampling (normalized by cpu count)
", .flags = 0, .defaultSortDesc = true, },
   [PERCENT_MEM] = { .name = "PERCENT_MEM", .title = "MEM% ", .description = "Percentage of the memory the process is using, based on resident memory size", .flags = 0, .defaultSortDesc = t
rue, },
// ...
#ifdef HAVE_DELAYACCT
   [PERCENT_CPU_DELAY] = { .name = "PERCENT_CPU_DELAY", .title = "CPUD% ", .description = "CPU delay %", .flags = PROCESS_FLAG_LINUX_DELAYACCT, .defaultSortDesc = true, },
   [PERCENT_IO_DELAY] = { .name = "PERCENT_IO_DELAY", .title = "IOD% ", .description = "Block I/O delay %", .flags = PROCESS_FLAG_LINUX_DELAYACCT, .defaultSortDesc = true, },
   [PERCENT_SWAP_DELAY] = { .name = "PERCENT_SWAP_DELAY", .title = "SWAPD% ", .description = "Swapin delay %", .flags = PROCESS_FLAG_LINUX_DELAYACCT, .defaultSortDesc = true, },
#endif
```

Explorer09:
How about this?
```c
void Process_printPercentage(float val, char* buffer, int n, int* attr) {
   if (val >= 0.0F) {   
      if (val < 1.0F) {
         xSnprintf(buffer, n, "%4.1f ", val);
      } else if (val <= 999.0F) {
         xSnprintf(buffer, n, "%4.3g ", val);
      } else {
         if (val > 9999.0F) {
            val = 9999.0F;
         }
         xSnprintf(buffer, n, "%4f ", val);
      }
      if (val < 0.05F) {
         *attr = CRT_colors[PROCESS_SHADOW];
      } else if (val >= 100.0F) {
      } else if (val >= 1000.0F) {
         *attr = CRT_colors[PROCESS_GIGABYTES];
      }
   } else {
      *attr = CRT_colors[PROCESS_SHADOW];
      xSnprintf(buffer, n, " N/A ");
   }
}
```

PR #779, opened 2021-09-05
Highlight percentages similar to large numbers

BenBE:
This change introduces a highlighting of percentages above 100% as cyan and above 1000 as green, similar to the colors used for MB and GB for e.g. the amount of memory used by an application.

Split off from #777.

### DIFF
--- a/Process.c
+++ b/Process.c
@@ -718,7 +718,7 @@ void Process_printPercentage(float val, char* buffer, int n, int* attr) {
          *attr = CRT_colors[PROCESS_MEGABYTES];
          xSnprintf(buffer, n, "%3d. ", (int)val);
       } else {
-         *attr = CRT_colors[PROCESS_GIGABYTES];
+         *attr = CRT_colors[PROCESS_MEGABYTES];
          xSnprintf(buffer, n, "%4d ", (int)val);
       }
    } else {
--- a/Process.c
+++ b/Process.c
@@ -715,8 +715,10 @@ void Process_printPercentage(float val, char* buffer, int n, int* attr) {
          }
          xSnprintf(buffer, n, "%4.1f ", val);
       } else if (val < 999) {
+         *attr = CRT_colors[PROCESS_MEGABYTES];
          xSnprintf(buffer, n, "%3d. ", (int)val);
       } else {
+         *attr = CRT_colors[PROCESS_GIGABYTES];
          xSnprintf(buffer, n, "%4d ", (int)val);
       }
    } else {


natoscott (changes requested):
@BenBE conceptually OK to do some colouring of these large percentage values, but I don't like the use of PROCESS_GIGABYTES and PROCESS_MEGABYTES macro names here.  Could we instead use LARGE_NUMBER for both?  Seems fitting since these are values > 100% ... is there any situation where such large values might be an expected situation though?  I can't think of any... so the bold red might be OK?

BenBE:
I tried to avoid the `LARGE_NUMBER` color for at least 100-1000% as these may regularly be encountered in multi-threaded processes that utilize more than one CPU. Having such processes is not actually rare if your system is under load and thus having them in `LARGE_NUMBER` would emphasize them quite much.

Also we did introduce a few intermediate steps in the large number highlighting for memory sizes after all with the two constants I used in the patch, so actually seeing the `LARGE_NUMBER` color being used there is kinda rare: The only common situation I can think of is with processes using ASAN and thus taking up about 20T of (virtual) RAM.

cgzones:
Currently this function is used for the *used-memory-percent*, *used-cpu-percent*, *used-cpu-percent-normalized* and the delay accounting columns. Of those the *used-memory-percent*, *used-cpu-percent-normalized* and the delay accounting values should never exceed 100%.
Regarding the *used-cpu-percent* column, a value bigger than 100% is nothing critical (thus I am not a fan of using a red color), it just shows a multi-threaded program utilizing more than one full CPU.
Also 100% is a special value, which might justify using a different highlighting,  but 1000% is not. Using 9 times the maximum resource is not much different from using it 11 times (e.g. consider a multi-threaded process on a 128 core system).

natoscott:
I noticed just now another situation of using red with the process time field - once a process has been running long enough, htop is happy to use red (again in a non-critical situation, of course, just to draw the eye I believe).  Looking at the code its quite amusing - there is some precedence here of overloading use of each of the existing macros we've been talking about ...  :)

```
   int yearColor = coloring ? CRT_colors[LARGE_NUMBER]      : CRT_colors[PROCESS];
   int dayColor  = coloring ? CRT_colors[PROCESS_GIGABYTES] : CRT_colors[PROCESS];
   int hourColor = coloring ? CRT_colors[PROCESS_MEGABYTES] : CRT_colors[PROCESS];
   int defColor  = CRT_colors[PROCESS];
```

Since a graduated scheme with these macros has been used for process times too, I'm OK with any scheme that does a similar thing with percentages.  It may be worth considering the transition points more deeply if we consider all three colors as available, as in the above code.

BenBE:
I'd also be fine with splitting the highlighting definitions for those colors, but this might lead to unnecessary splits where single colors are only ever used at one specific place in the code. A somewhat more consistent reuse of things like with the large number highlighting should be able to simplify maintenance and visual consistency in the long run. Though I'm totally fine with discussing exact rules for those kinds of transitions and cut-over points.

natoscott:
> Though I'm totally fine with discussing exact rules for those kinds of transitions and cut-over points.

Let's start with what you have here already then, and we can refine it over time if need be.

natoscott (approved):

Explorer09:
How about just coloring 100% and above (with `PROCESS_MEGABYTES`) and forget about the `PROCESS_GIGABYTES` color? I think the reasoning above is clear - there's nothing special about the 1000% threshold.

PR #781, opened 2021-09-09
Settings: set ok when reading configuration from sysconfdir

eworm-de:
Without this htoprc from sysconfdir is ignored and default meters
are loaded.


eworm-de:
This is a regression from becd33795c99a3f1b26024dd9fa8cc880f4d399d.

natoscott (approved):
LGTM too.

Issue #784, opened 2021-09-10
empty column right of meters


eworm-de:
There's an empty and unused column right of meters, even if borders are disabled.
![Screenshot_2021-09-10_15-20-12](https://user-images.githubusercontent.com/970853/132860199-9cddbba8-1561-40eb-b248-ee1780a3435b.png)
Is that expected? Can we change that?

PR #786, opened 2021-09-10
Do not leave empty last column in header

cgzones:
Do not leave empty last column in header meters by refactoring the width and separator logic.

Closes: #784

### DIFF
--- a/Header.c
+++ b/Header.c
@@ -194,7 +194,8 @@ void Header_draw(const Header* this) {
    for (int y = 0; y < height; y++) {
       mvhline(y, 0, ' ', COLS);
    }
-   const int width = COLS - pad;
+   const int numCols = HeaderLayout_getColumns(this->headerLayout);
+   const int width = COLS - 2 * pad - (numCols - 1);
    int x = pad;
    float roundingLoss = 0.0F;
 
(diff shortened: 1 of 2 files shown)


cgzones:
Needs some rework regarding the *header margin* and allowing text meters to expand to blank right neighbors.

Header.h
```diff
@@ -23,7 +23,7 @@ typedef struct Header_ {
    int height;
 } Header;
 
-#define Header_forEachColumn(this_, i_) for (size_t (i_)=0; (i_) < HeaderLayout_getColumns((this_)->headerLayout); ++(i_))
+#define Header_forEachColumn(this_, i_) for (size_t (i_)=0, numColumns__ = HeaderLayout_getColumns((this_)->headerLayout); (i_) < numColumns__; ++(i_))
```
Explorer09, line 26: Just a reminder: Double underscores are reserved in C++ standard. It's a good idea to avoid them in C program as well.
```suggestion
#define Header_forEachColumn(this_, i_) for (size_t (i_)=0, numColumns_ = HeaderLayout_getColumns((this_)->headerLayout); (i_) < numColumns_; ++(i_))
```

BenBE: AFAIR this was just for identifiers starting in double underscore …

cgzones: https://en.cppreference.com/w/cpp/language/identifiers:
> the identifiers with a double underscore anywhere are reserved

However C++ compatibility is generally broken, e.g by the usage of `this` as variable name.

cgzones:
> Needs some rework regarding the header margin and allowing text meters to expand to blank right neighbors.

Those work, but the CPU and MemorySwap meters need adjustments for their internal multiple columns.

Since this is a minor issue and the fixes might cause regressions, I'll propose to delay this fix after the next release.

natoscott:
Previous few commits from @smalinux address the final remaining feedback items (incl. defaulting to tabs disabled on upgrade).  If there's other areas for improvement let us know please otherwise we'll go ahead and merge soon - thanks!

PR #782, opened 2021-09-10
Several tweaks while running under Valgrind

cgzones:
* track origin of uninitialised values

* rewrite ncurses suppressions
   Simplify and update valgrind suppressions for possible leak and reachable memory inside ncurses.

* close backup stderr file after reset
   Close the backup file descriptor of original stderr once it has been restored at stderr.

* zero CPU data after allocation
   Zero all the CPU data, like totalPeriod, after its memory allocation via realloc(3).

   ```
  Conditional jump or move depends on uninitialised value(s)
      at 0x132A9B: LinuxProcessList_scanCPUTime (LinuxProcessList.c:1928)
      by 0x1358C3: ProcessList_goThroughEntries (LinuxProcessList.c:2079)
      by 0x12A79A: ProcessList_scan (ProcessList.c:627)
      by 0x11CA67: CommandLine_run (CommandLine.c:357)
      by 0x4A81E49: (below main) (libc-start.c:314)
    Uninitialised value was created by a heap allocation
      at 0x48396C5: malloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
      by 0x12F633: xRealloc (XUtils.c:64)
      by 0x12F633: xReallocArray (XUtils.c:78)
      by 0x1325A8: LinuxProcessList_updateCPUcount (LinuxProcessList.c:207)
      by 0x134E0A: ProcessList_new (LinuxProcessList.c:284)
      by 0x11C8D0: CommandLine_run (CommandLine.c:301)
      by 0x4A81E49: (below main) (libc-start.c:314)
   ```

PR #788, opened 2021-09-10
Linux: zero CPU data after allocation

cgzones:
Zero all the CPU data, like totalPeriod, after its memory allocation via realloc(3).

```
Conditional jump or move depends on uninitialised value(s)
   at 0x132A9B: LinuxProcessList_scanCPUTime (LinuxProcessList.c:1928)
   by 0x1358C3: ProcessList_goThroughEntries (LinuxProcessList.c:2079)
   by 0x12A79A: ProcessList_scan (ProcessList.c:627)
   by 0x11CA67: CommandLine_run (CommandLine.c:357)
   by 0x4A81E49: (below main) (libc-start.c:314)
 Uninitialised value was created by a heap allocation
   at 0x48396C5: malloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
   by 0x12F633: xRealloc (XUtils.c:64)
   by 0x12F633: xReallocArray (XUtils.c:78)
   by 0x1325A8: LinuxProcessList_updateCPUcount (LinuxProcessList.c:207)
   by 0x134E0A: ProcessList_new (LinuxProcessList.c:284)
   by 0x11C8D0: CommandLine_run (CommandLine.c:301)
   by 0x4A81E49: (below main) (libc-start.c:314)
```

Split from #782

### DIFF
--- a/XUtils.c
+++ b/XUtils.c
@@ -78,6 +78,22 @@ void* xReallocArray(void* ptr, size_t nmemb, size_t size) {
    return xRealloc(ptr, nmemb * size);
 }
 
+void* xReallocArrayZero(void* ptr, size_t prevmemb, size_t newmemb, size_t size) {
+   assert((ptr == NULL) == (prevmemb == 0));
+
+   if (prevmemb == newmemb) {
+      return ptr;
+   }
+
+   void* ret = xReallocArray(ptr, newmemb, size);
+
+   if (newmemb > prevmemb) {
+      memset((unsigned char*)ret + prevmemb * size, '\0', (newmemb - prevmemb) * size);
+   }
+
+   return ret;
+}
+
 inline bool String_contains_i(const char* s1, const char* s2) {
    return strcasestr(s1, s2) != NULL;
 }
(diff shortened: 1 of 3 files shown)


XUtils.c
```diff
@@ -78,6 +78,22 @@ void* xReallocArray(void* ptr, size_t nmemb, size_t size) {
...
+   assert((ptr == NULL) == (prevmemb == 0));
+
+   if (prevmemb == newmemb) {
+      return ptr;
+   }
+
+   void* ret = xReallocArray(ptr, newmemb, size);
+
+   if (newmemb > prevmemb) {
+      memset((unsigned char*)ret + prevmemb * size, '\0', (newmemb - prevmemb) * size);
```
BenBE, line 91: Missing overflow checks for multiplications `prevmemb * size` and `(newmemb - prevmemb) * size`.

BenBE (changes requested):

XUtils.c
cgzones: `xReallocArray()` in line 88 does an overflow check on `newmemb * size`.
Since `newmemb > prevmemb` is a precondition for this branch `prevmemb * size` is safe.
Since `prevmemb > 0` (it is of type `size_t`) `(newmemb - prevmemb) * size` is also safe.

BenBE (approved):

Issue #796, opened 2021-09-12
"Memory & Swap" meter in Text style spans whole terminal width


Explorer09:
The "Memory & Swap" meter in Text style spans the whole terminal width.

This happens in the current master branch. Only Text style is affected; Bar, Graph and LED styles are all fine.

Screenshot:
![htop-memory-and-swap-meter](https://user-images.githubusercontent.com/2487263/133006578-a46aa5ad-60b0-4a4d-8758-b146356134d8.png)

BenBE:
Is this the only meter in that row? Does this happen if there's another meter right next to it?

Explorer09:
@BenBE I tested it just now. If I place a meter on the right of the Memory & Swap meter, The Memory & Swap meter reduces its display width of the Text style. So you won't see the bug if you place a meter right next to the Memory & Swap meter,

cgzones:
It's a feature, see #484.

The MemorySwap meter is a bit special cause it is not strictly left-aligned, but aligns itself on the given width.
We could exempt this special one though.


```diff
diff --git a/Header.c b/Header.c
index 8940b36..25bb81a 100644
--- a/Header.c
+++ b/Header.c
@@ -19,6 +19,7 @@ in the source distribution for its full text.
 #include "CPUMeter.h"
 #include "DynamicMeter.h"
 #include "Macros.h"
+#include "MemorySwapMeter.h"
 #include "Object.h"
 #include "Platform.h"
 #include "ProvideCurses.h"
@@ -212,7 +213,11 @@ void Header_draw(const Header* this) {
          Meter* meter = (Meter*) Vector_get(meters, i);
 
          float actualWidth = colWidth;
-         if (meter->mode == TEXT_METERMODE) {
+
+         /* Let meters in text mode expand to the right on empty neighbors;
+         * except for non-left-aligned ones. */
+         if (meter->mode == TEXT_METERMODE &&
+             !Object_isA(&meter->super, &MemorySwapMeter_class.super)) {
             for (int j = 1; j < meter->columnWidthCount; j++) {
                actualWidth += (float)width * HeaderLayout_layouts[this->headerLayout].widths[col + j] / 100.0F;
             }
```

Explorer09:
@cgzones The implementation is not consistent.
1. Why only MemorySwapMeter has this behavior and not other meters that split itself to multiple sub-columns (such as the multiple CPU meters)?
[...]

PR #798, opened 2021-09-18
Header: do not let multi-column meters expand to empty neighbors

cgzones:
Extending to right neighbors is intended for text meters with an
overlong content, so the whole text is shown if possible.
Multi column meters, like the combined memory and swap meter, position
its text depending on the given total width; keep the position to the
original assigned header slot.

Short term resolution for #796


BenBE (approved):

BenBE (approved):

natoscott (approved):
LGTM, thanks @cgzones

PR #807, opened 2021-09-22
Fix mouse wheel collision with autogroups nice adjustment

### LINKED ISSUE #805: [3.1.0] Mouse Wheel Up doesn't work in the process list
With 3.1.0 update Mouse Wheel Up doesn't work in the process list. Wheel Down works. Also it works as intended in the other menus.

fasterit:
Fixes #805


BenBE (approved):
LGTM. Doesn't introduce any breakage for me.
Also fixes the problem for someone else.

PR #809, opened 2021-09-22
Document minimum versions for glibc and libcap

### LINKED ISSUE #804: htop 3.1.0 bumps required glibc and libcap versions without updating documentation
htop 3.1.0 removes the workaround from 8149823d56653b7d23397a235edb609b358ff17d, so it now requires glibc 2.14 to build for the `O_PATH` definition. This is easy enough to fix by re-applying that patch (or by noting a glibc 2.14 minimum requirement in the README, I suppose).

It also requires the defined constant `CAP_IS_SUPPORTED` which requires `libcap` 2.21 or later. This is also not documented anywhere.

fasterit:
Thank you to James Brown (Roguelazer).
Closes #804

### DIFF
--- a/README
+++ b/README
@@ -37,6 +37,9 @@ List of build-time dependencies:
     - `autotools`
   * `ncurses`
 
+When building `htop` on Linux with the GNU C Library you will need `glibc` v2.14 or later.
+`pcp-htop` for Linux can still be built with older `glibc` versions.
+
 **Note about `ncurses`:**
 > `htop` requires `ncurses` 6.0. Be aware the appropriate package is sometimes still called libncurses5 (on Debian/Ubuntu). Also `ncurses` usually comes in two flavours:
 >* With Unicode support.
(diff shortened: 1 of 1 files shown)


BenBE (approved):

README
```diff
@@ -37,6 +37,8 @@ List of build-time dependencies:
     - `autotools`
   * `ncurses`
 
+When building against the GNU C Library you will need `glibc` v2.14 or later.
```
natoscott, line 40: This requirement only affects the Linux htop backend platform BTW - folks can still run pcp-htop on Linux with this glibc (i.e. on RHEL6 and older platforms).  Not sure if/how best to clarify the wording here though.  Perhaps simply:

```
When building htop on Linux with the GNU C Library you will need `glibc` v2.14 or later.  Earlier glibc versions are still supported with pcp-htop on Linux, however.
```

fasterit: Reworded a bit, as I'd prefer to avoid "supported". CentOS 6 is out of support etc.

README
```diff
@@ -37,6 +37,8 @@ List of build-time dependencies:
     - `autotools`
   * `ncurses`
 
+When building `htop` on Linux with the GNU C Library you will need `glibc` v2.14 or later.  `pcp-htop` for Linux can still be built with older `glibc` versions.
```
BenBE, line 40:
```suggestion
When building `htop` on Linux with the GNU C Library you will need `glibc` v2.14 or later.
`pcp-htop` for Linux can still be built with older `glibc` versions.
```

BenBE (approved):

PR #815, opened 2021-09-24
Linux: define O_PATH if not defined

cgzones:
Defining O_PATH if not defined by <fcntl.h> does not really add any
maintenance cost and might improve some portability.

Related: #804


fasterit:
I hope it will stay now (cf. faabbaa) :-). We can remove the glibc version requirement from README then again, right?

BenBE (approved):

BenBE:
As I understand the original description for the glibc requirement: yes …

PR #816, opened 2021-09-24
Linux: always compute procExeDeleted if already set

cgzones:
A process, whose executable has been replaced and thus marked by htop,
can be re-executed with the replaced executable, with the same PID, in
two ways: the Linux feature checkpoint/restore or re-execution of PID 1.
The actual check is just a string comparison, like the dropped
condition, leading to (almost) no computation overhead.


BenBE (approved):
LGTM.

PR #827, opened 2021-10-02
Release memory for cmdline when Process_updateCmdline() returns.

Ximalas:
This closes a nasty memory leak. There is at least another leak looming somewhere when Disk I/O is shown in the header area. That could very well be an issue within libdevstat native to FreeBSD.

### DIFF
--- a/freebsd/FreeBSDProcessList.c
+++ b/freebsd/FreeBSDProcessList.c
@@ -444,6 +444,8 @@ static void FreeBSDProcessList_updateProcessName(kvm_t* kd, const struct kinfo_p
    *at = '\0';
 
    Process_updateCmdline(proc, cmdline, 0, end);
+
+   free(cmdline);
 }
 
 static char* FreeBSDProcessList_readJailName(const struct kinfo_proc* kproc) {


freebsd/FreeBSDProcessList.c
```diff
@@ -444,6 +444,7 @@ static void FreeBSDProcessList_updateProcessName(kvm_t* kd, const struct kinfo_p
    *at = '\0';
 
    Process_updateCmdline(proc, cmdline, 0, end);
+   free(cmdline);
```
BenBE, line 447: I'd suggest to have a blank line to visually structure things a bit more.
```suggestion

   free(cmdline);
```
Please squash that change into your existing commit.

Apart from this LGTM.

Ximalas: Is this OK? Did I do it right?

BenBE: The second change itself is correct, but it would be nicer to have both lines changed as part of the same commit. Have a look at `git rebase -i` and the `fixup` operation. The result should be just one commit with both lines changed.

Ximalas: I hope I now managed to squash what was once HEAD~2.

BenBE (approved):

BenBE:
@Ximalas: Any reason, why you have still marked this as draft?

PR #830, opened 2021-10-05
Handle interrupted sampling from within libpcp PDU transfers

natoscott:
This situation can arise if pcp-htop screen is resized right
at the same time sampling from pmcd(1) is happening.  Have a
couple more goes at it before giving up entirely; once there
is no data available though we cannot proceed into accessing
the sample result data structure (segv will result) so a new
short-circuit guard is added there also.


BenBE (approved):
Have you checked if there are other places that might need such an additional check?

natoscott:
@BenBE thanks - this is the only place where the screen resize signal will interrupt us with such consequences.  All other PMAPI calls that could be affected are made before we enter curses mode and start sampling.

Issue #331, opened 2020-11-21
Early program termination only from main()


BenBE:
Early program termination (like when initializing platform-specific resources, handling invalid arguments, …) should be handled in a central place instead of procedures randomly calling exit themselves.

Noted while reviewing #324. Similar changes in #173 (6a03cd237a2e85600275a358430ea544f53aba3e) …

PR #833, opened 2021-10-06
Early program termination only from main()

vasvolodimir:
Fix for https://github.com/htop-dev/htop/issues/331

### DIFF
--- a/CommandLine.c
+++ b/CommandLine.c
@@ -86,9 +86,9 @@ typedef struct CommandLineSettings_ {
    bool readonly;
 } CommandLineSettings;
 
-static CommandLineSettings parseArguments(const char* program, int argc, char** argv) {
+static CommandLineStatus parseArguments(const char* program, int argc, char** argv, CommandLineSettings* flags) {
 
-   CommandLineSettings flags = {
+   *flags = (CommandLineSettings) {
       .pidMatchList = NULL,
       .commFilter = NULL,
       .userId = (uid_t)-1, // -1 is guaranteed to be an invalid uid_t (see setreuid(2))
@@ -131,10 +131,10 @@ static CommandLineSettings parseArguments(const char* program, int argc, char** 
       switch (opt) {
          case 'h':
             printHelpFlag(program);
-            exit(0);
+            return STATUS_OK_EXIT;
          case 'V':
             printVersionFlag(program);
-            exit(0);
+            return STATUS_OK_EXIT;
          case 's':
             assert(optarg); /* please clang analyzer, cause optarg can be NULL in the 'u' case */
             if (String_eq(optarg, "help")) {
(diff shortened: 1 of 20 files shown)


darwin/Platform.c
```diff
@@ -126,7 +126,7 @@ static double Platform_nanosecondsPerMachTick = 1.0;
 
 static double Platform_nanosecondsPerSchedulerTick = -1;
 
-void Platform_init(void) {
+int Platform_init(void) {
```
BenBE, line 129: Why change this, if the only return value is `0` anyways? If this shares the same semantics as `parseArguments` then a similar (or the same) enum should be used for marking the return value.

darwin/Platform.c
```diff
@@ -139,6 +139,8 @@ void Platform_init(void) {
 
    const double nanos_per_sec = 1e9;
    Platform_nanosecondsPerSchedulerTick = nanos_per_sec / scheduler_ticks_per_sec;
+
+   return 0;
```
BenBE, line 143: This return value doesn't make much sense, in particular as an int. IMHO `bool` (`stdbool.h`) would be better suited.

CommandLine.c
```diff
@@ -283,12 +283,17 @@ int CommandLine_run(const char* name, int argc, char** argv) {
    else
       setlocale(LC_CTYPE, "");
 
-   CommandLineSettings flags = parseArguments(name, argc, argv);
+   int status;
+   CommandLineSettings flags;
+
+   if ((status = parseArguments(name, argc, argv, &flags)))
+      return status < 0;
```
BenBE, line 290: This does an implicit `bool`-to-`int` conversion and is kinda crude IMO.

CommandLine.c
```diff
@@ -130,10 +130,10 @@ static CommandLineSettings parseArguments(const char* program, int argc, char**
       switch (opt) {
          case 'h':
             printHelpFlag(program);
-            exit(0);
+            return 1;
```
BenBE, line 133: Avoid magic numbers. Define a small enum for proper initialization outcomes and return the appropriate constant instead.

pcp/Platform.c
```diff
@@ -309,7 +309,8 @@ void Platform_init(void) {
    sts = pmLookupName(pcp->totalMetrics, pcp->names, pcp->pmids);
    if (sts < 0) {
       fprintf(stderr, "Error: cannot lookup metric names: %s\n", pmErrStr(sts));
-      exit(1);
+      Platform_done();
```
BenBE, line 312: @natoscott Can you take a look here for PCP? Not sure if this call to `Platform_done()` is correct here.

BenBE (changes requested):

vasvolodimir:
Hi @BenBE,
Rebase has done.

CommandLine.c
```diff
@@ -288,12 +290,17 @@ int CommandLine_run(const char* name, int argc, char** argv) {
    else
       setlocale(LC_CTYPE, "");
 
-   CommandLineSettings flags = parseArguments(name, argc, argv);
+   CommandLineStatus status;
+   CommandLineSettings flags;
```
BenBE, line 294: Better to not leave things uninitialized here.
```suggestion
   CommandLineStatus status = STATUS_OK;
   CommandLineSettings flags = 0;
```

CommandLine.c
```diff
@@ -288,12 +290,17 @@ int CommandLine_run(const char* name, int argc, char** argv) {
...
+
+   if ((status = parseArguments(name, argc, argv, &flags)) != STATUS_OK)
+      return status != STATUS_OK_EXIT ? 1 : 0;
 
    if (flags.readonly)
       Settings_enableReadonly();
 
-   Platform_init();
+   if (Platform_init() == false)
```
BenBE, line 302:
```suggestion
   if (!Platform_init())
```

BenBE (changes requested):

BenBE (approved):

PR #841, opened 2021-10-11
Plug the memory leak for the Disk I/O meter

Ximalas:
There are no functions in `libdevstat` to initialise or clean up memory. The simplest change is to mark the local variable `info` as `static`.


BenBE:
Can you add an appropriate block to [scripts/htop_suppressions.valgrind](https://github.com/htop-dev/htop/blob/main/scripts/htop_suppressions.valgrind) to ignore all memleaks in `libdevstat`?

PR #852, opened 2021-10-20
Linux: use proper way to detect kernel threads

RinCat:
Use `PF_KTHREAD` flag in `/proc/[pid]/stat` to detect kernel threads.
This fixed an issue when a process's cmdline is empty, htop think it is a kernel thread.


BenBE:
Please squash related commits together to make the history nice and tidy …

BenBE (approved):

cgzones (approved):

PR #854, opened 2021-10-21
Shorten CWD path

### LINKED ISSUE #462: Shorten path for CWD column when displaying
As the CWD column is usually shorter than the paths it needs to display it would be nice to shorten the displayed path such that more relevant parts are kept visible.

```
Full:     /var/lib/postgresql/12/main
Current:  /var/lib/postgresql/12/ma
Proposed: /var/lib/postgre~/12/main
```

The general algorithm could be something linke this:

0. If string fits into target width, keep as is. Terminate this algorithm as soon as the resulting string fits.

1. Shorten from back of last, non-terminal path component until only one unshortened character remains (skip two-char components):
    ```
    /var/lib/postgres~/12/main
    /var/lib/postgre~/12/main
    /var/lib/postgr~/12/main
    /var/lib/postg~/12/main
    /var/lib/post~/12/main
    /var/lib/pos~/12/main
    /var/lib/po~/12/main
    /var/lib/p~/12/main
    ```

2. Repeat process with all path components before this.
    ```
    /var/lib64/p~/12/main
    /var/lib~/p~/12/main
    /var/li~/p~/12/main
    /var/l~/p~/12/main
    /v~/l~/p~/12/main
    ```

3. combine multiple (shortened) path components
    ```
    /v~/l~/p~~/main
    /v~/l~~/main
    /v~~/main
    ```

4. Shorten the last label from the middle, prioritizing to keep the front (keeping at least first+last char)
    ```
    /v~~/m~n
    ```

If the length that can be gained from shortening the last label exceeds what can be gained from step 3 than step 4 is prioritized.

The overall concept can likely also be reused for the `CGROUP` column (cf. #217).

onlined:
For showing the CWD path in a more compact way, this function

- shortens path parts before last from right to left, until first
  character of the path stays (`directory` -> `direc~` or `d~`)
- combines parts before last from right with left (`a~/b~` -> `a~~`)
- shortens the last part from middle, preferring to keep characters
  from left (`longdirectoryname` -> `long~ame`)

in order (of 2nd and 3rd, the one that removes more characters is applied first) until given maximum length is reached.

Closes #462.

### DIFF
--- a/CwdUtils.c
+++ b/CwdUtils.c
@@ -1,0 +1,164 @@
+#include "config.h" // IWYU pragma: keep
+
+#include "CwdUtils.h"
+
+#include <stdbool.h>
+#include <stdlib.h>
+#include <wchar.h>
+
+#include "XUtils.h"
+
+
+typedef struct ShortenCwdContext {
+   size_t maxLength;
+   size_t len;
+   wchar_t **parts;
+   size_t partsLen;
+   size_t *partLengths;
+} ShortenCwdContext;
+
+static void ShortenCwdContext_init(ShortenCwdContext *ctx, wchar_t *cwd, size_t maxLength) {
+   ctx->maxLength = maxLength;
+   ctx->len = wcslen(cwd);
+   ctx->parts = Wstring_split(cwd, L'/', &ctx->partsLen);
+   ctx->partLengths = xCalloc(ctx->partsLen, sizeof(size_t));
+   for (size_t i = 0; i < ctx->partsLen; i++)
+      ctx->partLengths[i] = wcslen(ctx->parts[i]);
+}
+
+static void ShortenCwdContext_destroy(ShortenCwdContext *ctx) {
+   for (size_t i = 0; i < ctx->partsLen; i++)
+      free(ctx->parts[i]);
+   free(ctx->parts);
+   free(ctx->partLengths);
+   *ctx = (ShortenCwdContext){0};
+}
+
+static void shortenCwdParts(ShortenCwdContext *ctx) {
+   for (size_t i = ctx->partsLen - 2; i != (size_t)-1 && ctx->len > ctx->maxLength; i--) {
+      if (ctx->partLengths[i] < 3)
+         continue;
+
+      size_t extraChars = ctx->len - ctx->maxLength;
+      size_t maxRemovableChars = ctx->partLengths[i] - 2;
+      size_t charsToRemove = extraChars < maxRemovableChars ? extraChars : maxRemovableChars;
+
+      ctx->partLengths[i] -= charsToRemove;
+      ctx->len -= charsToRemove;
+
+      Wstring_safeWcsncpy(ctx->parts[i] + (ctx->partLengths[i] - 1), L"~", 2);
+   }
+}
+
+static size_t collapseCwdParts(ShortenCwdContext *ctx, bool doActualWork) {
+   if (ctx->len <= ctx->maxLength || ctx->partsLen <= 3)
+      return 0;
+
+   size_t len = ctx->len;
+
+   size_t i;
+   for (i = ctx->partsLen - 2; i > 1; i--) {
+      if (len + (3 - ctx->partLengths[i]) <= ctx->maxLength)
+         break;
+
+      len -= ctx->partLengths[i] + 1;
+
+      if (doActualWork) {
+         ctx->partLengths[i] = 0;
+         free(ctx->parts[i]);
+         ctx->parts[i] = NULL;
+      }
+   }
+
+   len += 3 - ctx->partLengths[i];
+   size_t diff = ctx->len - len;
+
+   if (doActualWork) {
+      wchar_t newPart[] = L"~~~";
+      newPart[0] = ctx->parts[i][0];
+      free(ctx->parts[i]);
+      ctx->parts[i] = xWcsdup(newPart);
+      ctx->partLengths[i] = 3;
+      ctx->len = len;
+   }
+
+   return diff;
+}
+
+static size_t shortenCwdLastPart(ShortenCwdContext *ctx, bool doActualWork) {
+   if (ctx->len <= ctx->maxLength)
+      return 0;
+
+   size_t lastPartLen = ctx->partLengths[ctx->partsLen - 1];
+   if (lastPartLen <= 3)
+      return 0;
+
+   wchar_t *lastPart = ctx->parts[ctx->partsLen - 1];
+   size_t extraChars = ctx->len - ctx->maxLength + 1;
+   size_t maxRemovableChars = lastPartLen - 2;
+   size_t charsToRemove = extraChars < maxRemovableChars ? extraChars : maxRemovableChars;
+
+   if (doActualWork) {
+      size_t charsAtBeginning = (lastPartLen - charsToRemove + 1) / 2;
+      size_t charsAtEnd = lastPartLen - charsToRemove - charsAtBeginning;
+      lastPart[charsAtBeginning] = '~';
+      wmemmove(lastPart + charsAtBeginning + 1, lastPart + lastPartLen - charsAtEnd, charsAtEnd);
+      lastPart[charsAtBeginning + charsAtEnd + 1] = '\0';
+      ctx->partLengths[ctx->partsLen - 1] = lastPartLen - charsToRemove + 1;
+      ctx->len -= charsToRemove - 1;
+   }
+
+   return charsToRemove - 1;
+}
+
+static wchar_t* buildCwdFromParts(ShortenCwdContext *ctx) {
+   size_t len = ctx->partsLen - 1;
+   for (size_t i = 0; i < ctx->partsLen; i++)
+      len += ctx->partLengths[i];
+
+   wchar_t *newCwd = xCalloc(len + 1, sizeof(wchar_t));
+
+   newCwd[0] = '\0';
+   for (size_t i = 0, writeIndex = 0; i < ctx->partsLen; i++) {
+      if (!ctx->parts[i])
+         continue;
+
+      Wstring_safeWcsncpy(newCwd + writeIndex, ctx->parts[i], ctx->partLengths[i] + 1);
+      writeIndex += ctx->partLengths[i];
+      if (i < ctx->partsLen - 1)
+         newCwd[writeIndex++] = L'/';
+   }
+
+   return newCwd;
+}
+
+char* CwdUtils_shortenCwd(const char *cwd, const size_t maxLength) {
+   wchar_t *wcwd = xMbstowcs(cwd);
+   size_t len = wcslen(wcwd);
+   if (len <= maxLength) {
+      free(wcwd);
+      return xStrdup(cwd);
+   }
+
+   ShortenCwdContext ctx;
+   ShortenCwdContext_init(&ctx, wcwd, maxLength);
+   free(wcwd);
+   wcwd = NULL;
+
+   shortenCwdParts(&ctx);
+   if (shortenCwdLastPart(&ctx, false) > collapseCwdParts(&ctx, false)) {
+      shortenCwdLastPart(&ctx, true);
+      collapseCwdParts(&ctx, true);
+   } else {
+      collapseCwdParts(&ctx, true);
+      shortenCwdLastPart(&ctx, true);
+   }
+
+   wchar_t *newWcwd = buildCwdFromParts(&ctx);
+   char *newCwd = xWcstombs(newWcwd);
+   free(newWcwd);
+
+   ShortenCwdContext_destroy(&ctx);
+
+   return newCwd;
+}
--- a/CwdUtils.h
+++ b/CwdUtils.h
@@ -1,0 +1,8 @@
+#ifndef HEADER_CwdUtils
+#define HEADER_CwdUtils
+
+#include <stddef.h>
+
+char* CwdUtils_shortenCwd(const char* cwd, size_t maxLength);
+
+#endif
--- a/Process.c
+++ b/Process.c
@@ -21,6 +21,7 @@ in the source distribution for its full text.
 #include <sys/resource.h>
 
 #include "CRT.h"
+#include "CwdUtils.h"
 #include "Hashtable.h"
 #include "Machine.h"
 #include "Macros.h"
@@ -630,7 +631,7 @@ void Process_writeField(const Process* this, RichString* str, RowField field) {
          attr = CRT_colors[PROCESS_SHADOW];
          cwd = "main thread terminated";
       } else {
-         cwd = this->procCwd;
+         cwd = this->procCwdShort;
       }
       Row_printLeftAlignedField(str, attr, cwd, 25);
       return;
@@ -770,6 +771,7 @@ void Process_done(Process* this) {
    free(this->procComm);
    free(this->procExe);
    free(this->procCwd);
+   free(this->procCwdShort);
    free(this->mergedCommand.str);
    free(this->tty_name);
 }
@@ -1081,6 +1083,11 @@ void Process_updateCPUFieldWidths(float percentage) {
 
    Row_updateFieldWidth(PERCENT_CPU, width);
    Row_updateFieldWidth(PERCENT_NORM_CPU, width);
+}
+
+void Process_updateShortCwd(Process* this) {
+   free(this->procCwdShort);
+   this->procCwdShort = this->procCwd ? CwdUtils_shortenCwd(this->procCwd, 25) : NULL;
 }
 
 const ProcessClass Process_class = {
--- a/XUtils.c
+++ b/XUtils.c
@@ -402,3 +402,84 @@ unsigned int countTrailingZeros(unsigned int x) {
    return mod37BitPosition[(-x & x) % 37];
 }
 #endif
+
+wchar_t* xMbstowcs(const char *mbs) {
+   size_t len = strlen(mbs);
+   wchar_t *wcs = xCalloc(len + 1, sizeof(wchar_t));
+   mbstate_t mbstate = {0};
+   mbsrtowcs(wcs, &mbs, len + 1, &mbstate);
+   if (mbs)
+      fail();
+   return wcs;
+}
+
+char* xWcstombs(const wchar_t *wcs) {
+   size_t len = wcslen(wcs);
+   if (SIZE_MAX / MB_CUR_MAX <= len)
+      fail();
+   size_t mbsSize = len * MB_CUR_MAX + 1;
+   char *mbs = xMalloc(mbsSize);
+   mbstate_t mbstate = {0};
+   wcsrtombs(mbs, &wcs, mbsSize, &mbstate);
+   if (wcs)
+      fail();
+   return mbs;
+}
+
+wchar_t* xWcsdup(const wchar_t* str) {
+   wchar_t* data = wcsdup(str);
+   if (!data)
+      fail();
+   return data;
+}
+
+wchar_t* xWcsndup(const wchar_t* str, size_t len) {
+   wchar_t* data = xCalloc(len + 1, sizeof(wchar_t));
+   if (!data)
+      fail();
+
+   Wstring_safeWcsncpy(data, str, len + 1);
+
+   return data;
+}
+
+wchar_t** Wstring_split(const wchar_t* s, wchar_t sep, size_t* n) {
+   const unsigned int rate = 10;
+   wchar_t** out = xCalloc(rate, sizeof(wchar_t*));
+   size_t ctr = 0;
+   unsigned int blocks = rate;
+   const wchar_t* where;
+   while ((where = wcschr(s, sep)) != NULL) {
+      size_t size = (size_t)(where - s);
+      out[ctr] = xWcsndup(s, size);
+      ctr++;
+      if (ctr == blocks) {
+         blocks += rate;
+         out = xReallocArray(out, blocks, sizeof(wchar_t*));
+      }
+      s += size + 1;
+   }
+   if (s[0] != L'\0') {
+      out[ctr] = xWcsdup(s);
+      ctr++;
+   }
+   out = xRealloc(out, sizeof(wchar_t*) * (ctr + 1));
+   out[ctr] = NULL;
+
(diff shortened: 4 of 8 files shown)


onlined:
@Explorer09 But I need to preprocess the string and decide which characters to keep or remove, before actually printing the characters. Do you suggest to printing to a non-visible or dummy terminal?

Explorer09:
@onlined My idea is to use the space of the CWD table cell -- you print the characters just as you process it. It's okay to overprint the space multiple time until you finally shorten the path.

BenBE:
> @onlined My idea is to use the space of the CWD table cell -- you print the characters just as you process it. It's okay to overprint the space multiple time until you finally shorten the path.

Problem is, this cell is not actually available for reference in that routine. The function should be independent of which screen position it is later on printed too. In particular, the printed string is only rendered into a `RichString` later on and that is finally written onto the display. So shortening to 25 code points with `wcwidth`/`wcswidth` is fully acceptable IMHO (even if this may differ somewhat from what the terminal will finally end up actually writing). Cf. `mbswidth`(3)

Explorer09:
@BenBE That's a structural problem. It would be better to fix the routine to accept a terminal position as the parameter. You don't need to prepare the string until it's printed at least once. So it's not unreasonable to request a terminal position for the routine.

BenBE:
> @BenBE That's a structural problem. It would be better to fix the routine to accept a terminal position as the parameter. You don't need to prepare the string until it's printed at least once. So it's not unreasonable to request a terminal position for the routine.

Yes and no. Yes, we should delay the preparation of this string to when we first actually write it to the display. But that's not how this works internally in `htop` right now and would take a bit of structural work to change it that way. That's not part of this PR and if we go that route, then we'd need to do quite a bit of related work elsewhere too to keep things consistent.

And no, you shouldn't print temporary things to a terminal. True story: I had a text-based app once that had a progress bar for showing the load progress when it was initializing different things. So far, so normal. One of the users had a braille display connected to the computer and sent in a complaint about the braille display going brrrrrrrrrr whenever the program started. Long story short: DON'T write temporary or fast-changing information to a terminal.

Explorer09:
@BenBE But htop is ncursed-based! If htop prints everything sequentially (to maintain compatibility with line terminals), I won't suggest this approach.
I know what you mean when you say progress bar as an example: wget is the example of having two ways to output the progress. One prints everything sequentially (incuding the progress indicator), the other uses overprinting to indicate the progress (and it prints the filename being downloaded in a scrolling text).

BenBE:
> @BenBE But htop is ncursed-based! If htop prints everything sequentially (to maintain compatibility with line terminals), I won't suggest this approach. I know what you mean when you say progress bar as an example: wget is the example of having two ways to output the progress. One prints everything sequentially (incuding the progress indicator), the other uses overprinting to indicate the progress (and it prints the filename being downloaded in a scrolling text).

Both methods cause the mal-behavior I described. The problem is not the sequential printing, but the printing itself.

Process.c
```diff
@@ -31,6 +31,7 @@ in the source distribution for its full text.
 #include "Settings.h"
 #include "Table.h"
 #include "XUtils.h"
+#include "CwdUtils.h"
```
BenBE, line 34: Header order

XUtils.c
```diff
@@ -402,3 +402,83 @@ unsigned int countTrailingZeros(unsigned int x) {
...
+   mbstate_t mbstate = {0};
+   mbsrtowcs(wcs, &mbs, len + 1, &mbstate);
+   if (mbs)
+      fail();
+   return wcs;
+}
+
+char* xWcstombs(const wchar_t *wcs) {
+   size_t len = wcslen(wcs);
+   size_t mbsSize = len * sizeof(wchar_t) + 1;
```
BenBE, line 418: Missing overflow check.

XUtils.c
```diff
@@ -402,3 +402,83 @@ unsigned int countTrailingZeros(unsigned int x) {
...
+   char *mbs = xMalloc(mbsSize);
+   mbstate_t mbstate = {0};
+   wcsrtombs(mbs, &wcs, mbsSize, &mbstate);
```
BenBE, line 421: I think this will fail for multi-byte string, where a sizable number of characters is more than 4 bytes. This may happen with strings containing many emoji or other characters that use more than 4 bytes for UTF-8 encoding.

XUtils.c
```diff
@@ -402,3 +402,83 @@ unsigned int countTrailingZeros(unsigned int x) {
...
+   return data;
+}
+
+wchar_t* xWcsndup(const wchar_t* str, size_t len) {
+   wchar_t* data = xCalloc(len + 1, sizeof(wchar_t));
+   if (!data)
+      fail();
+
+   for (size_t i = 0; i < len && str[i]; i++)
+      data[i] = str[i];
```
BenBE, line 440: Use `memcpy` and write explicit terminator?

XUtils.c
```diff
@@ -402,3 +402,83 @@ unsigned int countTrailingZeros(unsigned int x) {
...
+   size_t ctr = 0;
+   unsigned int blocks = rate;
+   const wchar_t* where;
+   while ((where = wcschr(s, sep)) != NULL) {
+      size_t size = (size_t)(where - s);
+      out[ctr] = xWcsndup(s, size);
+      ctr++;
+      if (ctr == blocks) {
+         blocks += rate;
+         out = (wchar_t**) xRealloc(out, sizeof(wchar_t*) * blocks);
```
BenBE, line 457: No need to cast `void*`. Also, there's `xReallocArray` for this case.

BenBE (changes requested):
Thank you very much for this update. I haven't yet taken a closer look at `CwdUtils.c`; but I noticed some points with the surrounding changes.

XUtils.c
Explorer09: @BenBE For your info, emoji sequences are considered multiple characters in Unicode. It is only that the _display_ of the characters would look combined together.

The real problem is we can't ensure the `sizeof(wchar_t)` is at least 4. For (hypothetical) systems that use 2 bytes for `wchar_t` (thus only be able to handle the Unicode BMP), this would surely overflow the `mbs` buffer.

XUtils.c
onlined: I preferred `wcsncpy`, but I can change it to `memcpy` if you like.

XUtils.c
onlined: Done.

XUtils.c
onlined: Done.

XUtils.c
onlined: Done.

Process.c
onlined: Done.

XUtils.c
Explorer09: @onlined 
`wcsncpy` has an additional behavior of padding `L'\0'` to the unused character space. If you don't require this behavior, it's better to avoid it as it wastes processor cycles.

CwdUtils.c
```diff
@@ -0,0 +1,155 @@
...
+      ctx.partLengths[i] = wcslen(ctx.parts[i]);
+
+   shortenCwdParts(&ctx);
+   if (shortenCwdLastPart(&ctx, false) > collapseCwdParts(&ctx, false)) {
+      shortenCwdLastPart(&ctx, true);
+      collapseCwdParts(&ctx, true);
+   } else {
+      collapseCwdParts(&ctx, true);
+      shortenCwdLastPart(&ctx, true);
+   }
```
Explorer09, line 143: The rules/logic in this area is arcane and hard to read. What's the reasoning behind these function calls and conditionals?

CwdUtils.c
```diff
@@ -0,0 +1,155 @@
...
+   ShortenCwdContext ctx = {
+      .maxLength = maxLength,
+      .len = len,
+   };
+   ctx.parts = Wstring_split(wcwd, L'/', &ctx.partsLen);
+   free(wcwd);
+   wcwd = NULL;
+   ctx.partLengths = xCalloc(ctx.partsLen, sizeof(size_t));
+   for (size_t i = 0; i < ctx.partsLen; i++)
+      ctx.partLengths[i] = wcslen(ctx.parts[i]);
```
Explorer09, line 134: This part of code looks like a constructor to `ShortenCwdContext`, but with complex `malloc` operations. Can we have a `ShortenCwdContext_new(wchar_t* wstr)` function?

CwdUtils.c
```diff
@@ -0,0 +1,155 @@
...
+
+   return diff;
+}
+
+static size_t shortenCwdLastPart(ShortenCwdContext *ctx, bool doActualWork) {
+   if (ctx->len <= ctx->maxLength)
+      return 0;
+
+   size_t lastPartLen = ctx->partLengths[ctx->partsLen - 1];
+   if (lastPartLen <= 3)
```
Explorer09, line 75: Unexplained magic number 3.

CwdUtils.c
```diff
@@ -0,0 +1,155 @@
...
+typedef struct ShortenCwdContext {
+   size_t maxLength;
+   size_t len;
+   wchar_t **parts;
+   size_t partsLen;
+   size_t *partLengths;
+} ShortenCwdContext;
+
+static void shortenCwdParts(ShortenCwdContext *ctx) {
+   for (int i = ctx->partsLen - 2; i >= 0 && ctx->len > ctx->maxLength; i--) {
```
Explorer09, line 20:
```suggestion
   for (size_t i = ctx->partsLen - 2; i != (size_t)-1 && ctx->len > ctx->maxLength; i--) {
```

CwdUtils.c
onlined: I added a `ShortenCwdContext_init` for not to allocate unnecessarily.

XUtils.c
onlined: `memcpy` would copy `len` bytes too, so should I go back to explicitly copying in a loop?

CwdUtils.c
onlined: Done.

XUtils.c
Explorer09: @onlined The compiler can inline `memcpy` when it's useful. When it comes to performance, `memcpy` is arguably the best.

onlined: Changed to `memcpy`.

onlined: I just realized a problem, for using `memcpy` we have to assume that `str` has more characters than `len` or we can risk accessing unallocated memory and crash. Because of that concern, I changed it to use `Wstring_safeWcsncpy`.

XUtils.c
BenBE: > @BenBE For your info, emoji sequences are considered multiple characters in Unicode. It is only that the _display_ of the characters would look combined together.

I was aware of this (after all 🐈‍⬛ is f0 9f 90 88  e2 80 8d  e2 ac 9b, which is 3 Unicode codepoints), but through a mix-up on my side I kinda had in mind that some of the non-BMP characters needed 5/6 bytes. Further fueled by MySQL, who had a utf8mb4 charset, which they added to support things beyond the BMP.

But …

> The real problem is we can't ensure the `sizeof(wchar_t)` is at least 4. For (hypothetical) systems that use 2 bytes for `wchar_t` (thus only be able to handle the Unicode BMP), this would surely overflow the `mbs` buffer.
> 
> The right element size to use is `MB_CUR_MAX`, not `sizeof(wchar_t)`.

Exactly. Or encode twice and use the first encoding to query how much space is actually needed.

CwdUtils.c
BenBE: A quick comment about the rationale of the check would be fine.

CwdUtils.h
```diff
@@ -0,0 +1,8 @@
+#ifndef HEADER_CwdUtils
+#define HEADER_CwdUtils
+
+#include <stddef.h>
+
```
BenBE, line 5: Second blank line after the header includes.

CwdUtils.c
```diff
@@ -0,0 +1,158 @@
+#include "config.h" // IWYU pragma: keep
+
+#include "CwdUtils.h"
+
+#include <stdbool.h>
+#include <stdlib.h>
+#include <wchar.h>
+
+#include "XUtils.h"
+
```
BenBE, line 10: Second blank line after the header includes.

CwdUtils.c
```diff
@@ -0,0 +1,158 @@
...
+typedef struct ShortenCwdContext {
+   size_t maxLength;
+   size_t len;
+   wchar_t **parts;
+   size_t partsLen;
+   size_t *partLengths;
```
BenBE, line 16: The fields `partsLen` and `partLengths` have quite similar names which may cause confusion.

To mitigate this, using a naming scheme like `strSize`/`partCount`/`partPtr`/`partSize` might be somewhat more easy to disambiguate.

CwdUtils.c
```diff
@@ -0,0 +1,158 @@
...
+} ShortenCwdContext;
+
+void ShortenCwdContext_init(ShortenCwdContext *ctx, wchar_t *cwd, size_t maxLength) {
```
BenBE, line 19: As this is not exported, this can/should be `static` …
```suggestion
static void ShortenCwdContext_init(ShortenCwdContext *ctx, wchar_t *cwd, size_t maxLength) {
```

CwdUtils.c
```diff
@@ -0,0 +1,158 @@
...
+   }
+
+   wchar_t *newWcwd = buildCwdFromParts(&ctx);
+   char *newCwd = xWcstombs(newWcwd);
+   free(newWcwd);
+
+   free(ctx.partLengths);
+   for (size_t i = 0; i < ctx.partsLen; i++)
+      free(ctx.parts[i]);
+   free(ctx.parts);
```
BenBE: If you have a `_init` function, you should consider using a dtor function for cleanup.

CwdUtils.c
```diff
@@ -0,0 +1,163 @@
...
+      writeIndex += ctx->partLengths[i];
+      if (i < ctx->partsLen - 1)
+         newCwd[writeIndex++] = L'/';
+   }
+
+   return newCwd;
+}
+
+char* CwdUtils_shortenCwd(const char *cwd, const size_t maxLength) {
+   wchar_t *wcwd = xMbstowcs(cwd);
```
BenBE, line 135: I quick pre-test could be:

```c
   if (strnlen(cwd, maxLength) < maxLength) {
      return xStrdup(cwd);
   }
```

CwdUtils.c
```diff
@@ -0,0 +1,163 @@
...
+   for (size_t i = 0; i < ctx->partsLen; i++)
+      free(ctx->parts[i]);
+   free(ctx->parts);
+   free(ctx->partLengths);
+   *ctx = (ShortenCwdContext){0};
+}
+
+static void shortenCwdParts(ShortenCwdContext *ctx) {
+   for (size_t i = ctx->partsLen - 2; i != (size_t)-1 && ctx->len > ctx->maxLength; i--) {
+      if (ctx->partLengths[i] < 3)
```
BenBE, line 38:
```suggestion
      if (ctx->partLengths[i] <= 3)
```

Also, documenting where the 3 comes from (initial, separator, trailer) would be nice.

CwdUtils.c
BenBE: Minimal length of directory/part strings (after shortening, if not shorter before).

BenBE (changes requested):

CwdUtils.c
onlined: Done.

CwdUtils.c
onlined: Done.

CwdUtils.c
onlined: Shouldn't the comparison be `<=`?

BenBE: No, because `==` would mean that `maxLength` bytes were read and no NUL byte was encountered. You couldn't distinguish the case of exactly `maxLength` bytes from the one with more bytes. This doesn't hurt though, as the `==` case is caught below in the propr check (with conversion) …

You could do `strnlen(cwd, maxLength + 1) < maxLength` if you wanted to catch that, but strictly speaking you'd then also have to include an overflow check for the `maxLength + 1` …

CwdUtils.c
Explorer09: Should be `partPtrs` and `partSizes` because the latter two are arrays.

Explorer09:
Before I comment on the code quality further, I noticed the entire PR had a wrong assumption that every Unicode character occupy the same width on the terminal, which is not the case.

Forget my previous comments (that was left 3 years ago) about printing the strings on the terminal twice. We need to determine the display width of the characters through [wcwidth(3)](https://man7.org/linux/man-pages/man3/wcwidth.3.html) or similar. Otherwise we might shorten the paths in all the wrong way.

Another thing that confused me is, why are we converting the wide character strings back to multi-byte strings (`xWcstombs`) when they are for display? I believe a better conversion would be to a `RichString` instance directly.

And by using `RichString` I have an idea to do one more thing: Color the abbreviated tildes (`~`) with a different color from surrounding text. This can give a stronger hint that the tildes are abbreviations and not part of directory names.

onlined:
Should I still store this short CWD (this time as `RichString`) or should I create it on the fly?

BenBE:
> Before I comment on the code quality further, I noticed the entire PR had a wrong assumption that every Unicode character occupy the same width on the terminal, which is not the case.

I remarked about this about 3 years ago re `wcwidth`/`wcswidth` …

So, this is one point that is on the todo list before inclusion.

> Forget my previous comments (that was left 3 years ago) about printing the strings on the terminal twice. We need to determine the display width of the characters through [wcwidth(3)](https://man7.org/linux/man-pages/man3/wcwidth.3.html) or similar. Otherwise we might shorten the paths in all the wrong way.

Furthermore, we should take into account when sequences of Unicode codepoints need to go together like 🐈‍⬛ or 🏳️‍⚧️. Also, for several characters, in particular in the CJK range, the width of single characters may be larger than 1.

> Another thing that confused me is, why are we converting the wide character strings back to multi-byte strings (`xWcstombs`) when they are for display? I believe a better conversion would be to a `RichString` instance directly.
> 
> And by using `RichString` I have an idea to do one more thing: Color the abbreviated tildes (`~`) with a different color from surrounding text. This can give a stronger hint that the tildes are abbreviations and not part of directory names.

That's actually a good idea. Given the metadata the functions collect internally, both outputting into a raw string and a formatted `RichString` should be feasible without too much extra work.

> Should I still store this short CWD (this time as `RichString`) or should I create it on the fly?

When creating a `RichString` you will need about 400 bytes of memory (for each instance). With about 3000 LWP on my kinda idle system, we are talking 1.2 MiB of extra memory (without optimization) for the shortened strings alone.

Contrary, if we assumed kinda optimized storage of the metadata to produce the `RichStrings` of like 10 parts max and 8 bytes per part, we'd be down to ~250 KiB. If we were to use bit masks instead, we could get this to 1 bit per wchar + string data (~30 bytes -> 90 KiB).

So the important step is less the output as `RichString`, but to be able to produce it from as little temporary data as possible.

NB: Storing an `uint8_t` per 8 wchars gives the last estimate, IFF you don't need to store the (maximum) length of the string. If done that way, you can just loop over each character in the output string and color it in (shade) when a bit in the bitmask is set. Extra handling for the OS directory prefixes may cause this scheme to take up more than 1 bit though.

Overall, there has been some discussion on IRC about this topic. During this discussion we found that we would like to give this feature a bit more time so the following things can be handled and implemented properly:
1. Proper character (display) width handling
2. RichString support for printing
3. Memory footprint
4. Handling together with highlighting default OS directories

Explorer09:
> Furthermore, we should take into account when sequences of Unicode codepoints need to go together like 🐈‍⬛ or 🏳️‍⚧️. Also, for several characters, in particular in the CJK range, the width of single characters may be larger than 1.

Just for everyone's info, I've seen [a wcswidth implementation](https://github.com/jquast/wcwidth/blob/master/wcwidth/wcwidth.py) that does address multiple character emojis (a la ZWJ sequences). But's it's Python only.

I believe it's still harmless if we miscalculated the width of emojis right now, because they happen more rarely in practice, and, depending on terminal display, they might not display as a combined icon but a separated sequence of emoji icons (which is not necessary a bug - Unicode permits such a fallback sequence display, especially for emojis that are added in later Unicode versions).

> 
> So the important step is less the output as `RichString`, but to be able to produce it from as little temporary data as possible.

My personal concern with the current state of the PR is that it might use too many `malloc` operations to hold the split tokens. `malloc` have chances to fail, and you have to `free` each of them when you have done. I wonder if it's feasible to hold a `wchar_t []` or a `RichString` buffer containing the whole path string, and shorten the strings in-place. Just an idea.

fasterit:
> > Furthermore, we should take into account when sequences of Unicode codepoints need to go together like 🐈‍⬛ or 🏳️‍⚧️. Also, for several characters, in particular in the CJK range, the width of single characters may be larger than 1.
> 
> Just for everyone's info, I've seen [a wcswidth implementation](https://github.com/jquast/wcwidth/blob/master/wcwidth/wcwidth.py) that does address multiple character emojis (a la ZWJ sequences). But's it's Python only.

wcswidth(3) is good enough™ for the `htop` use case. There will always be ways to cause glitches in a curses display on virtual terminals. We don't have to go to extremes to cover the (only) deliberately-constructed corner cases, too.
Just completely ignoring Unicode these days doesn't do either (as per the 2021 implementation).
/DLange

BenBE:
> > So the important step is less the output as `RichString`, but to be able to produce it from as little temporary data as possible.
> 
> My personal concern with the current state of the PR is that it might use too many `malloc` operations to hold the split tokens. `malloc` have chances to fail, and you have to `free` each of them when you have done. I wonder if it's feasible to hold a `wchar_t []` or a `RichString` buffer containing the whole path string, and shorten the strings in-place. Just an idea.

Yes, working with a single buffer for splitting is possible, but will require a second allocation to hold pointers for where each part begins (or some in-band information where the string terminates). Given you know the maximum final width, you can determine the number of splits for the result. For the input this is in the worst case `wcslen(input) / 2`.

NB: The actual conversion performance isn't too pressing, as the CWD doesn't change often AND we can cache the result when it stayed the same.

Issue #862, opened 2021-10-31
[feature request] Ability to change the default kill signal


Hexcede:
This is a very minor QoL thing, it's not particularly consequential but would be easily implemented and save me (and probably others) quite a lot of time. When killing a program, the default kill signal is 15 (SIGTERM). However, some programs such as ones under wine don't seem to be effected by the signal, and programs which are hanging are not effected by it either.

Would love to be able to switch this to pick a signal which is more aggressive such as SIGKILL (9) by default, as I tend to be repeatedly killing many wine processes from htop (usually when they are being problematic or don't create an interface)

This is semi-related to the similar request, #690. I have a similar issue in that I am repeatedly SIGKILLing many processes a day, as I do development under wine with a tool that doesn't quite work so well there, but unlike the related issue I am killing many processes with various lengthy pathnames I would need to type out in order to kill from my shell.

PR #865, opened 2021-11-02
Add option to change default signal sent on F9

cgzones:
Closes: #862

/cc @Hexcede

### DIFF
--- a/Action.c
+++ b/Action.c
@@ -344,9 +344,12 @@ static Htop_Reaction actionKill(State* st) {
    if (Settings_isReadonly())
       return HTOP_OK;
 
-   Panel* signalsPanel = SignalsPanel_new();
+   static int preSelectedSignal = SIGNALSPANEL_INITSELECTEDSIGNAL;
+
+   Panel* signalsPanel = SignalsPanel_new(preSelectedSignal);
    const ListItem* sgn = (ListItem*) Action_pickFromVector(st, signalsPanel, 14, true);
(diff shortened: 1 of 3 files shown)


DisplayOptionsPanel.c
```diff
@@ -135,6 +136,7 @@ DisplayOptionsPanel* DisplayOptionsPanel_new(Settings* settings, ScreenManager*
    Panel_add(super, (Object*) CheckItem_newByRef("Enable the mouse", &(settings->enableMouse)));
    #endif
    Panel_add(super, (Object*) NumberItem_newByRef("Update interval (in seconds)", &(settings->delay), -1, 1, 255));
+   Panel_add(super, (Object*) NumberItem_newByRef("Default selected signal on F9", &(settings->defaultSignal), 0, 1, Platform_numberOfSignals - 1));
```
BenBE, line 139: Not quite the best UX for the user …

fasterit:
I'm not a fan either. Why not default to the last signal sent (and may be even persist that in `htoprc`) ? Much easier, more intuitive and one setting less.

Hexcede:
> I'm not a fan either. Why not default to the last signal sent (and may be even persist that in `htoprc`) ? Much easier, more intuitive and one setting less.

I like this idea. It's fast, and solves the UX issue well. It also doesn't have to persist across sessions which could be confusing if you don't *regularly* use `htop`.

BenBE (approved):

PR #867, opened 2021-11-04
Cleanup: Fix misc styleguide issues

smalinux:


BenBE:
I'm fine with the blank after switch, but I'd prefer to keep the casted negative numbers without a space in between …

BenBE (approved):

PR #873, opened 2021-11-16
Reduce allocation size of cp_time_n and cp_time_o on FreeBSD and DragonFlyBSD

GuillaumeGomez:
Unless I'm missing something, these two arrays are only ever used to read one set of CPU data, not X (where X is the number of CPUs).

Usage is [here](https://github.com/htop-dev/htop/blob/f75a8bc3a1131151181d6794000b10063400221e/freebsd/FreeBSDProcessList.c#L190) and allocation is [here](https://github.com/htop-dev/htop/blob/f75a8bc3a1131151181d6794000b10063400221e/freebsd/FreeBSDProcessList.c#L112-L113).

### DIFF
--- a/freebsd/FreeBSDProcessList.c
+++ b/freebsd/FreeBSDProcessList.c
@@ -109,8 +109,8 @@ ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* dynamicMeters, H
 
    size_t sizeof_cp_time_array = sizeof(unsigned long) * CPUSTATES;
    len = 2; sysctlnametomib("kern.cp_time", MIB_kern_cp_time, &len);
-   fpl->cp_time_o = xCalloc(cpus, sizeof_cp_time_array);
-   fpl->cp_time_n = xCalloc(cpus, sizeof_cp_time_array);
+   fpl->cp_time_o = xCalloc(CPUSTATES, sizeof(unsigned long));
+   fpl->cp_time_n = xCalloc(CPUSTATES, sizeof(unsigned long));
    len = sizeof_cp_time_array;
 
    // fetch initial single (or average) CPU clicks from kernel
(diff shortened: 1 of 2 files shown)


freebsd/FreeBSDProcessList.c
```diff
@@ -109,8 +109,8 @@ ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* dynamicMeters, H
 
    size_t sizeof_cp_time_array = sizeof(unsigned long) * CPUSTATES;
    len = 2; sysctlnametomib("kern.cp_time", MIB_kern_cp_time, &len);
-   fpl->cp_time_o = xCalloc(cpus, sizeof_cp_time_array);
-   fpl->cp_time_n = xCalloc(cpus, sizeof_cp_time_array);
+   fpl->cp_time_o = xCalloc(1, sizeof_cp_time_array);
+   fpl->cp_time_n = xCalloc(1, sizeof_cp_time_array);
```
BenBE, line 113: Given this allocates an array of `unsigned long` this should better be:
```suggestion
   fpl->cp_time_o = xCalloc(CPU_STATES, sizeof(unsigned long));
   fpl->cp_time_n = xCalloc(CPU_STATES, sizeof(unsigned long));
```

BenBE (changes requested):

GuillaumeGomez:
I looked into all `*bsd` and into `solaris` but only `DragonFlyBSD` has the issue. Fixed it as well.

freebsd/FreeBSDProcessList.c
BenBE: There is a somewhat subtle issue here: `calloc` does not only an overflow check that the current version is missing, but also ensures proper alignment. Depending on the exact values used here, this [may make a difference](https://stackoverflow.com/a/501949), causing the actual allocation being larger due to proper alignment of allocated elements (See note [here](https://en.cppreference.com/w/c/memory/calloc)).

fasterit: `CPU_STATES` -> `CPUSTATES` in the suggested change

GuillaumeGomez: Thanks for the extra explanations @BenBE! Totally makes sense, I updated.

@fasterit: Thanks for the correction.

BenBE (approved):

cgzones (approved):

PR #874, opened 2021-11-19
Add a fix to use gettimeofday if monotonic is not available.

lutzmad:
Add an additional fix to support environments where monotonic is not available to Generic_gettime_monotonic.
The fix is necessary to support MacOS older than 10.13., tested down to 10.10..

See issue #373 for more information.

With regards, Lutz

### DIFF
--- a/generic/gettime.c
+++ b/generic/gettime.c
@@ -49,9 +49,13 @@ void Generic_gettime_monotonic(uint64_t* msec) {
    else
       *msec = 0;
 
-#else
+#else /* lower resolution gettimeofday() should be always available */
 
-# error "No monotonic clock available"
+   struct timeval tv;
+   if (gettimeofday(&tv, NULL) == 0)
+      *msec = ((uint64_t)tv.tv_sec * 1000) + ((uint64_t)tv.tv_usec / 1000);
+   else
+      *msec = 0;
 
 #endif
 }


cgzones:
@BenBE maybe warn with GNU compatible compilers?
```
#ifdef __GNUC__
#warning "No monotonic clock available, fall back to gettimeofday"
#endif
```

BenBE:
> @BenBE maybe warn with GNU compatible compilers?
> 
> ```
> #ifdef __GNUC__
> #warning "No monotonic clock available, fall back to gettimeofday"
> #endif
> ```

Don't quite like that variant because that's breaking `-Werror`. So either we treat that case as equally valid code or fully error out on it. With this approach you're potentially masking real compiler warnings due to prohibiting effective use of `-Werror`.

Explorer09:
@cgzones How about a warning at configure time? You know, when you check for `gettimeofday` in configure script.

natoscott:
> @cgzones How about a warning at configure time? You know, when you check for `gettimeofday` in configure script.

gettimeofday is available on all platforms, we don't check for it in configure.  Maybe you're thinking of clock_gettime - and we could add a warning if that's missing I guess (although configure does already report "checking for clock_gettime... no"), but it's not overly helpful to warn IMO as there's no action a developer can take to improve the situation.

I think this fallback is fine as implemented now.

natoscott (approved):
LGTM

BenBE:
I think the summary may display some information on the clock source like

```
Clock source: monotonic
Clock source: realtime
Clock source: gettimeofday (legacy)
```

That way it give's some obvious hint about the issue, without being a full-blown warning.

natoscott:
> That way it give's some obvious hint about the issue, without being a full-blown warning.

But there isn't an issue - the platform supports whatever APIs it supports, and we use the best available.  I think this extra diagnostic message isn't really needed.

BenBE (approved):

PR #878, opened 2021-11-23
add github workflow to build release tarball

### LINKED ISSUE #876: release tarballs without autoconf deps
it'd be great if you guys could upload release tarballs including pre-generated configure scripts as in 2.x time, so one doesn't have to install the full autoconf suite to build. (it's perfectly fine to depend on it for build-from-git, not so much for releases).

rofl0r:
the built release tarball provides a pre-built configure script and efficient packing with xz.
tested working.

this creates a release tarball attached to a *release*.
unfortunately i didnt find a way to attach it to a tag, that means after creating a tag one has to additionally go to tags tab, open tag and click "create release from tag".
apart from that no further action is required by the maintainer.

closes #876

### DIFF
--- a/.github/workflows/build_release.yml
+++ b/.github/workflows/build_release.yml
@@ -1,0 +1,51 @@
+name: Build Source Release
+
+# Trigger whenever a release is created
+on:
+  release:
+    types:
+      - created
+
+jobs:
+  build:
+    name: build
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+      with:
+        submodules: true
+
+    - name: archive
+      id: archive
+      run: |
+        VERSION=${{ github.event.release.tag_name }}
+        PKGNAME="htop-$VERSION"
+        SHASUM=$PKGNAME.tar.xz.sha256
+        autoreconf -i
+        mkdir -p /tmp/$PKGNAME
+        mv * /tmp/$PKGNAME
+        mv /tmp/$PKGNAME .
+        TARBALL=$PKGNAME.tar.xz
+        tar cJf $TARBALL $PKGNAME
+        sha256sum $TARBALL > $SHASUM
+        echo "::set-output name=tarball::$TARBALL"
+        echo "::set-output name=shasum::$SHASUM"
+    - name: upload tarball
+      uses: actions/upload-release-asset@v1
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        upload_url: ${{ github.event.release.upload_url }}
+        asset_path: ./${{ steps.archive.outputs.tarball }}
+        asset_name: ${{ steps.archive.outputs.tarball }}
+        asset_content_type: application/gzip
+
+    - name: upload shasum
+      uses: actions/upload-release-asset@v1
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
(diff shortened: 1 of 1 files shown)


fasterit:
Nice job! Please make it run on a tag event and only if the tag matches ~ `refs\/tags/v[0-9]+\.[0-9]+\.[0-9]+` .
See https://github.community/t/how-to-trigger-on-a-release-only-if-tag-name-matches-pattern/18514/ for a how-to.

Explorer09:
@rofl0r  I wonder why you are packing the tarball manually with the `tar` command. Isn't `make dist` available for the purpose and that it works better?

rofl0r:
> Nice job!

thanks!

> Please make it run on a tag event and only if the tag matches ~ refs\/tags/v[0-9]+\.[0-9]+\.[0-9]+

i've pushed several test branches to get github to make a release tarball just by pushing a tag, but was unable to achieve it. it seems build artifacts can only be uploaded to a tag when it's manually marked as "release".
the article you linked seems to use an external service on azure that creates artifacts, but imo it's preferable if it's done via github.

looking at git history, it appears past tags were labeled "3.x" and not "v3.x".
anyway, given that one has to manually click "create release from tag" i see no advantage of restricting this action to match a regex.

> I wonder why you are packing the tarball manually with the tar command. Isn't make dist available for the purpose and that it works better?

because it's simpler, doesn't require me to debug/fix make dist target, creates .xz, and allows to reuse the same github action for other repos with minor tweaks.

Explorer09:
> because it's simpler, doesn't require me to debug/fix make dist target, creates .xz, and allows to reuse the same github action for other repos with minor tweaks.

Well, make dist was designed to simplify the release process, as it builds the program, runs the testsuite (make check), and prepares the tarball all in one command. And there is one advantage of packing the tarball with make dist: automake by default uses a higher compression level than the compression program's default. It's the tarball for distribution, so higher compression is typically preferred.

rofl0r:
> It's the tarball for distribution, so higher compression is typically preferred.

we can achieve the same by exporting "XZ_OPT=-9e", though the 300kb result is already quite good https://github.com/rofl0r/htop/releases/tag/v9.99.99b

BenBE:
Any chance to also include zstd compression in the archive formats?

rofl0r:
> Any chance to also include zstd compression in the archive formats?

no big deal. it's a matter of adding a line in the build process and one artifact declaration.
though i wonder if it's that useful. in the benchmarks i've seen it's rarely if ever better than xz -9, and not commonly available in distros, unlike. but i can add it anyway, if you like.

rofl0r:
i added zstd max option and xz -9e for testing:  a81374d075b46d3807f6981389db2f75723f1bf8  but result isn't worth it imo:  https://github.com/rofl0r/htop/releases/tag/9.99.99f

natoscott:
> i added zstd max option and xz -9e for testing: [a81374d](https://github.com/htop-dev/htop/commit/a81374d075b46d3807f6981389db2f75723f1bf8) but result isn't worth it imo

Agreed.  Chatting to @BenBE he's not concerned if we drop this.  He also mentioned it might be worthwhile using the same format that Hisham used to provide (.tar.gz) - it looks like the commit you referenced above could be trivially changed to do that.

However, I suspect that means we'd have two .tar.gz files (the github auto-generated one, and this new one) which would be further confusing?  So for today's release (3.1.2) I'm planning to stick with the more tested code in this PR - if folks really want the extra .tar.gz file as well, let's add that in the next major release (3.2.0 - likely next year).

Issue #860, opened 2021-10-29
Disk IO:   value is not initialized


toralf:
The very first value seems to be ridiculous high, the following looks ok.

(version 3.1.1)

BenBE:
As this meter calculates its updates by reading the current stats for a device and calculating deltas this indicates that there is an initial update call missing for this value.

Can you check if the following patch resolves your issue?
```diff
$ git diff
diff --git a/DiskIOMeter.c b/DiskIOMeter.c
index c5f14768..11fb791e 100644
--- a/DiskIOMeter.c
+++ b/DiskIOMeter.c
@@ -81,6 +81,12 @@ static void DiskIOMeter_updateValues(Meter* this) {
       cached_msTimeSpend_total = data.totalMsTimeSpend;
    }
 
+   if (passedTimeInMs > 30000) {
+      // Triggers for the first initialization and
+      // when there was a long time we did not collect updates
+      hasData = false;
+   }
+
    this->values[0] = cached_utilisation_diff;
    this->total = MAXIMUM(this->values[0], 100.0); /* fix total after (initial) spike */
 
diff --git a/NetworkIOMeter.c b/NetworkIOMeter.c
index dcba78de..6c429c32 100644
--- a/NetworkIOMeter.c
+++ b/NetworkIOMeter.c
@@ -85,6 +85,12 @@ static void NetworkIOMeter_updateValues(Meter* this) {
       cached_txp_total = data.packetsTransmitted;
    }
 
+   if (passedTimeInMs > 30000) {
+      // Triggers for the first initialization and
+      // when there was a long time we did not collect updates
+      hasData = false;
+   }
+
    this->values[0] = cached_rxb_diff;
    this->values[1] = cached_txb_diff;
    if (cached_rxb_diff + cached_txb_diff > this->total) {
```

toralf:
indeed - solved

PR #883, opened 2021-12-05
IOMeters: rework initial display

cgzones:
Show a non highlighted string at the start of htop, not the failure
text. Also with ncurses the maximum time between updates is 25.5
seconds, so an interval of 30 seconds can only be reached by pausing
htop externally.
Also the original fix only handled the text mode, not the bar mode.

Improves: 2977414d ("Discard stale information from DiskIO and NetworkIO meters")
Related: #860

### DIFF
--- a/DiskIOMeter.c
+++ b/DiskIOMeter.c
@@ -12,6 +12,7 @@ in the source distribution for its full text.
 
 #include "CRT.h"
 #include "Macros.h"
+#include "Meter.h"
 #include "Object.h"
 #include "Platform.h"
 #include "ProcessList.h"
@@ -25,7 +26,7 @@ static const int DiskIOMeter_attributes[] = {
    METER_VALUE_IOWRITE,
 };
 
-static bool hasData = false;
+static MeterRateStatus status = RATESTATUS_INIT;
 static uint32_t cached_read_diff;
 static uint32_t cached_write_diff;
 static double cached_utilisation_diff;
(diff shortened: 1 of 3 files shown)


BenBE:
This reverts the check for stale data that was intentionally added previously.

cgzones:
So the stale check is intended to after htop being paused (externally) is resumed avoid spikes in the graph meter?
Otherwise seeing the accumulated IO usage over that paused time frame makes sense to me.

BenBE:
> So the stale check is intended to after htop being paused (externally) is resumed avoid spikes in the graph meter? Otherwise seeing the accumulated IO usage over that paused time frame makes sense to me.

Exactly. Or when due to high system load the last update is to old and thus using this data is prone to produce a spike if those information were processed.

DiskIOMeter.c
```diff
@@ -25,7 +25,7 @@ static const int DiskIOMeter_attributes[] = {
    METER_VALUE_IOWRITE,
 };
 
-static bool hasData = false;
+static enum { S_DATA, S_INIT, S_NODATA, S_STALE } status = S_INIT;
```
BenBE, line 28: Don't quite like this inline type declaration here. In particular due to the duplication with the other IO meter.

DiskIOMeter.c
```diff
@@ -36,20 +36,17 @@ static void DiskIOMeter_updateValues(Meter* this) {
...
       uint64_t diff;
 
-      cached_last_update = pl->realtimeMs;
-
       DiskIOData data;
-
-      hasData = Platform_getDiskIO(&data);
-      if (!hasData) {
-         this->values[0] = 0;
+      status = Platform_getDiskIO(&data) ? (cached_last_update == 0 ? S_INIT : S_DATA) : S_NODATA;
```
BenBE, line 47: This is not the easiest LOC to read. What about writing the if blocks explicitly?

BenBE, line 47:
```suggestion
      if (!Platform_getDiskIO(&data))
         status = S_NODATA;
      else if (cached_last_update == 0)
         status = S_INIT;
      else if (passedTimeInMs > 30000)
         status = S_STALE;
      else
         status = S_DATA;
```

BenBE (changes requested):
Changes in NetworkIO meter similar to DiskIO meter …

Meter.h
```diff
@@ -134,6 +134,13 @@ typedef enum {
    LAST_METERMODE
 } MeterModeId;
 
+typedef enum {
+   DATA_RATESTATUS,
+   INIT_RATESTATUS,
+   NODATA_RATESTATUS,
+   STALE_RATESTATUS
```
BenBE, line 141: I'd prefer to have the namespacing of those constants to be prefixed. This also aligns better with most of the remaining code (Meter.h being an exception here AFAICS.

```suggestion
   RATESTATUS_DATA,
   RATESTATUS_INIT,
   RATESTATUS_NODATA,
   RATESTATUS_STALE
```

BenBE (commented):
Minor naming issue, apart from that LGTM.

BenBE (approved):

eworm-de:
Thanks a lot! 😊

PR #887, opened 2021-12-06
{Disk,Network}IOMeter: degrade "no data" to warning

eworm-de:
Starting htop I am now faced with "no data", painted in red. This color
should be reserved to critical issues, like "degraded" in systemd meter.

So degrade "no data" to warning state and paint it in yellow.


BenBE:
I prefer the PR #883, but distinguishing between init, stale and actually no data should be easy to integrate there.

eworm-de:
Agreed, that looks a lot better. 😊 Thanks!

PR #891, opened 2021-12-09
RFC: dynamically adjust column width at runtime for specific fields

cgzones:
Adjust the column width for fields known to be large dynamically at each cycle to the longest value.

Also with the recent addition of multiple screens, over-long columns can be moved into their own screen.


linux/LinuxProcess.c
```diff
@@ -280,7 +280,9 @@ static void LinuxProcess_writeField(const Process* this, RichString* str, Proces
       }
       xSnprintf(buffer, n, "%5lu ", lp->ctxt_diff);
       break;
-   case SECATTR: snprintf(buffer, n, "%-30.30s ", lp->secattr ? lp->secattr : "?"); break;
+   case SECATTR:
+      snprintf(buffer, n, "%-*.*s ", Process_fieldWidths[SECATTR], Process_fieldWidths[SECATTR], lp->secattr ? lp->secattr : "N/A");
+      break;
```
BenBE, line 285: If you keep CGROUP and CCGROUP in one line you might as well keep SECATTR in one line too …

linux/LinuxProcessList.c
```diff
@@ -900,13 +900,17 @@ static void LinuxProcessList_readCGroupFile(LinuxProcess* process, openat_arg_t
 
    bool changed = !process->cgroup || !String_eq(process->cgroup, output);
 
+   Process_updateFieldWidth(CGROUP, strlen(output));
    free_and_xStrdup(&process->cgroup, output);
 
-   if (!changed)
+   if (!changed) {
+      Process_updateFieldWidth(CCGROUP, strlen(process->cgroup_short));
```
BenBE, line 907: `CCGROUP` given by `process->cgroup_short` can be `NULL` in which case this is aliased to `process->cgroup`.

linux/LinuxProcessList.c
```diff
@@ -900,13 +900,17 @@ static void LinuxProcessList_readCGroupFile(LinuxProcess* process, openat_arg_t
...
       return;
+   }
 
    char* cgroup_short = CGroup_filterName(process->cgroup);
    if (cgroup_short) {
+      Process_updateFieldWidth(CCGROUP, strlen(cgroup_short));
       free_and_xStrdup(&process->cgroup_short, cgroup_short);
       free(cgroup_short);
    } else {
```
BenBE, line 916: Missing call to `Process_updateFieldWidth` in the `else` branch …

BenBE (changes requested):

BenBE:
Should also adjust CPU usage column for values above 10000% as of #947 …

fasterit:
LGTM

PR #897, opened 2021-12-17
LinuxProcessList: do not collect LRS per thread

tanriol:
It's a memory map property, so it's process-wide and collecting it just
once should be enough.

### DIFF
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -1473,8 +1473,9 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
       {
          bool prev = proc->usesDeletedLib;
 
-         if ((ss->flags & PROCESS_FLAG_LINUX_LRS_FIX) ||
-             (settings->highlightDeletedExe && !proc->procExeDeleted && !proc->isKernelThread && !proc->isUserlandThread)) {
+         if (!proc->isKernelThread && !proc->isUserlandThread &&
+            ((ss->flags & PROCESS_FLAG_LINUX_LRS_FIX) || (settings->highlightDeletedExe && !proc->procExeDeleted))) {
+
             // Check if we really should recalculate the M_LRS value for this process
             uint64_t passedTimeInMs = pl->realtimeMs - lp->last_mlrs_calctime;
 
@@ -1487,6 +1488,7 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
          } else {
             /* Copy from process structure in threads and reset if setting got disabled */
             proc->usesDeletedLib = (proc->isUserlandThread && parent) ? parent->usesDeletedLib : false;
+            lp->m_lrs = (proc->isUserlandThread && parent) ? ((const LinuxProcess*)parent)->m_lrs : 0;
          }
 
          proc->mergedCommand.exeChanged |= prev ^ proc->usesDeletedLib;


linux/LinuxProcessList.c
```diff
@@ -1473,8 +1474,8 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
       {
          bool prev = proc->usesDeletedLib;
 
-         if ((ss->flags & PROCESS_FLAG_LINUX_LRS_FIX) ||
-             (settings->highlightDeletedExe && !proc->procExeDeleted && !proc->isKernelThread && !proc->isUserlandThread)) {
+         if (((ss->flags & PROCESS_FLAG_LINUX_LRS_FIX) || (settings->highlightDeletedExe && !proc->procExeDeleted))
```
cgzones, line 1477: Maybe move the now top-level conditions `!proc->isKernelThread && !proc->isUserlandThread` to the front for improved readability?

cgzones (commented):
One inline comment, otherwise LGTM.

linux/LinuxProcessList.c
tanriol: Done and rebased.

linux/LinuxProcessList.c
```diff
@@ -1366,11 +1366,12 @@ static char* LinuxProcessList_updateTtyDevice(TtyDriver* ttyDrivers, unsigned lo
    return out;
 }
 
-static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_t parentFd, const char* dirname, const Process* parent, double period) {
+static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_t parentFd, const char* dirname, const LinuxProcess* lparent, double period) {
    ProcessList* pl = (ProcessList*) this;
    const struct dirent* entry;
    const Settings* settings = pl->settings;
    const ScreenSettings* ss = settings->ss;
+   const Process *parent = (const Process *) lparent;
```
cgzones, line 1374: Style nit-picking: `const Process* parent = (const Process*) lparent;`

cgzones (approved):

linux/LinuxProcessList.c
tanriol: Sure, amended and pushed.

BenBE: Not fully, the left side also has the `*` aligned on the type …

tanriol: Sorry, I should have noticed, fixed.

linux/LinuxProcessList.c
```diff
@@ -1487,6 +1489,7 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
          } else {
             /* Copy from process structure in threads and reset if setting got disabled */
             proc->usesDeletedLib = (proc->isUserlandThread && parent) ? parent->usesDeletedLib : false;
+            lp->m_lrs = (proc->isUserlandThread && lparent) ? lparent->m_lrs : 0;
```
BenBE, line 1492: I'm not sure it's really worth changing the function signature for one place where we need access to the casted subclass …

```suggestion
            lp->m_lrs = (proc->isUserlandThread && parent) ? ((LinuxProcess_t *)parent)->m_lrs : 0;
```
And skip renaming the signature of this function …

tanriol: Done.

BenBE (approved):

PR #898, opened 2021-12-17
GH Actions: enable Werror in PCP build

cgzones:
Just exclude the singe warning type currently issued.

Avoids e64269df ("Fix process state handling compiler warning on PCP platform")


BenBE (approved):

natoscott (approved):
@cgzones /me waves from his holidays - LGTM - thanks, that's a better approach than what I had previously.

Issue #877, opened 2021-11-23
"Show custom thread names" doesn't work unless "Merge exe, comm and cmdline" is enabled


goffrie:
I have the "show custom thread names" option turned on, but here's what I see:
![regression](https://user-images.githubusercontent.com/1240398/142970136-1727e2c0-a731-4c20-af8c-78018c9299bf.png)

I bisected this to commit ce27f8379d143d64ab6a7787fdf38cb864e18edd (#724), part of release 3.1.0. If I check out the prior commit, htop looks like this, as I expect:
![expected behaviour](https://user-images.githubusercontent.com/1240398/142970196-25540637-28cb-49cb-a0e4-b9f99bd2366a.png)

If I turn on "Merge exe, comm and cmdline" on 3.1.0, thread names are shown again, at the expense of a lot more noise:
![with option](https://user-images.githubusercontent.com/1240398/142970294-c07d827e-86e3-4acc-a908-70b7b44b269b.png)

PR #899, opened 2021-12-18
Naive fix for issue #877 - getting back custom thread names

jan-konczak-cs-put:
Proposed fix for issue #877 - getting back custom thread names.

I attempted a 'git bisect' and ended up at: ce27f8379d143d64ab6a7787fdf38cb864e18edd

It seems that along with other changes replacing the command with thread name has been accidentally deleted.

Reverting this particular line:
https://github.com/htop-dev/htop/commit/ce27f8379d143d64ab6a7787fdf38cb864e18edd#diff-7b4ee00635e2355bc3b83ada6e57991d2a79ba37828dab7128c461c14645580aL1550
seems to fix the issue.

Please keep in mind that my PR is a naive revert attempt, and while I've tested some setups and it seems to fix the problem and did not break anything, it may not be the fix that one's looking for.

### DIFF
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -1552,6 +1552,10 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
             Process_updateCmdline(proc, NULL, 0, 0);
          } else if (!LinuxProcessList_readCmdlineFile(proc, procFd)) {
             Process_updateCmdline(proc, statCommand, 0, strlen(statCommand));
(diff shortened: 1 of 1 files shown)


BenBE:
Unfortunately this is the wrong place to address the root cause of this issue. There's a central routine `Process_makeCommandStr` where the logic is implemented for all platforms. Unfortunately in the short-circuit logic implemented there there's some issue I'm still looking into, that - although implemented - does not distinguish between processes and threads with the setting enabled.

PR #900, opened 2021-12-21
linux/SystemdMeter: use warning for some states

eworm-de:
Searching man systemd(1) for "is-system-running" we see these states:

initializing
starting
running
degraded
maintenance
stopping
offline
unknown

Let's differentiate some more: green (ok) if state is "running", red (error)
if the state is unknown to htop or "degraded", yellow (warning) for
everything else (incl. "initializing", "starting", "maintenance" &
"stopping").

### DIFF
--- a/linux/SystemdMeter.c
+++ b/linux/SystemdMeter.c
@@ -307,8 +307,12 @@ static int valueDigitColor(unsigned int value) {
 static void SystemdMeter_display(ATTR_UNUSED const Object* cast, RichString* out) {
    char buffer[16];
    int len;
+   int color = METER_VALUE_ERROR;
 
-   int color = (systemState && String_eq(systemState, "running")) ? METER_VALUE_OK : METER_VALUE_ERROR;
+   if (systemState) {
+      color = String_eq(systemState, "running") ? METER_VALUE_OK :
+              String_eq(systemState, "degraded") ? METER_VALUE_ERROR : METER_VALUE_WARN;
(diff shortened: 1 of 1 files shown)


BenBE:
Without spoiling it, I'm not really a fan of this change, as given the nature of the Systemd Meter there's already little difference between running and degraded (those other states never really being witnessed ;-)). What would be more interesting is warning only if certain non-critical services are not running, but reporting an error if certain critical services are marked as non-running. Though this requires additional code currently missing and which is likely out-of-scope.

Let's see what the others think …

eworm-de:
> Without spoiling it, I'm not really a fan of this change, as given the nature of the Systemd Meter there's already little difference between running and degraded (those other states never really being witnessed ;-)).

With *little difference* between running and degraded your reference that it becomes critical when just one of hundreds of services failed? Sure... Often it is the one **critical** service that fails. 😳

> What would be more interesting is warning only if certain non-critical services are not running, but reporting an error if certain critical services are marked as non-running. Though this requires additional code currently missing and which is likely out-of-scope.

I do not think this information is available. There is no weight, importance, significance or whatever for a systemd service, no?

But let me explain my motivation for this: After a system reboot starting `htop` is the first thing I do. It gives a good overview, now including the systemd state. Everything is fine if nothing is red. A starting system is expected to become ok within reasonable time, so it should not show red.

cgzones (approved):

PR #910, opened 2022-01-03
Altered --pids to show child/descendant processes of specified pid

jessesimpson36:
Hey everyone,

I've found that there's a lot of times that I use htop to track forking and joining of processes, and the current filtering functionality doesn't make it easy to show just one full tree of processes without a bunch of other processes bumping me up and down while I'm debugging. 

This patch will change the --pids  flag to display not just the  process with the pid specified, but the child processes.

Does this functionality sound like it would be useful? perhaps if it were separated out into a different flag?

### DIFF
--- a/ProcessList.c
+++ b/ProcessList.c
@@ -557,9 +557,8 @@ void ProcessList_rebuildPanel(ProcessList* this) {
 
       if ( (!p->show)
          || (this->userId != (uid_t) -1 && (p->st_uid != this->userId))
-         || (incFilter && !(String_contains_i(Process_getCommand(p), incFilter))) ) {
+         || (incFilter && !(String_contains_i(Process_getCommand(p), incFilter))) )
          continue;
-      }
 
       if (this->pidMatchList && !Hashtable_get(this->pidMatchList, p->tgid)) {
          /* Follow processes parents to find matches to process match list */
--- a/ProcessList.c
+++ b/ProcessList.c
@@ -557,9 +557,24 @@ void ProcessList_rebuildPanel(ProcessList* this) {
 
       if ( (!p->show)
          || (this->userId != (uid_t) -1 && (p->st_uid != this->userId))
-         || (incFilter && !(String_contains_i(Process_getCommand(p), incFilter)))
-         || (this->pidMatchList && !Hashtable_get(this->pidMatchList, p->tgid)) )
+         || (incFilter && !(String_contains_i(Process_getCommand(p), incFilter))) ) {
          continue;
+      }
+
+      if (this->pidMatchList && !Hashtable_get(this->pidMatchList, p->tgid)) {
+         /* Follow processes parents to find matches to process match list */
+         Process* currProc = p;
+         bool found = Hashtable_get(this->pidMatchList, currProc->tgid);
+         while (!Process_isChildOf(currProc, 0) && !found) {
+            pid_t ppid = Process_getParentPid(currProc);
+            currProc = ProcessList_findProcess(this, ppid);
+            if (!currProc)
+               break;
+            found = Hashtable_get(this->pidMatchList, currProc->tgid);
+         }
+         if (!found)
+            continue;
+      }
 
       Panel_set(this->panel, idx, (Object*)p);
 


jessesimpson36:
Spotted one problem already:  This patch would only work on linux because I imported that file I modified.

ProcessList.c
```diff
@@ -18,6 +18,7 @@ in the source distribution for its full text.
 #include "Platform.h"
 #include "Vector.h"
 #include "XUtils.h"
+#include "linux/LinuxProcess.h"
```
BenBE, line 21: Unnecessary import …

Apart from this import being forbidden in the (platform-independent) core code …

ProcessList.c
```diff
@@ -553,13 +554,39 @@ void ProcessList_rebuildPanel(ProcessList* this) {
...
+                   ppidProcess = nextProcess;
+               }
+               if (ppidProcess->tgid == 0 && !Hashtable_get(this->pidMatchList, ppidProcess->tgid)) {
+                   if (!preExisting) {
+                        free(ppidProcess);
+                   }
+                   continue;
+               }
+           }
+       }
```
BenBE, line 589: Code indentation and formatting

ProcessList.c
```diff
@@ -553,13 +554,39 @@ void ProcessList_rebuildPanel(ProcessList* this) {
...
-         || (incFilter && !(String_contains_i(Process_getCommand(p), incFilter)))
-         || (this->pidMatchList && !Hashtable_get(this->pidMatchList, p->tgid)) )
-         continue;
+       Process* p = (Process*) Vector_get(this->processes, i);
+
+       if ( (!p->show)
+               || (this->userId != (uid_t) -1 && (p->st_uid != this->userId))
+               || (incFilter && !(String_contains_i(Process_getCommand(p), incFilter))) ) {
+           continue;
+       } else {
```
BenBE, line 563: If there's no way the true branch continues, there's also no reason for making the false branch its own block. Reduce level of indentation where ever possible.

ProcessList.c
```diff
@@ -553,13 +554,39 @@ void ProcessList_rebuildPanel(ProcessList* this) {
...
+           if (this->pidMatchList && !Hashtable_get(this->pidMatchList, p->tgid)) {
+               // iterate over parents and check hashtable for entries
+               pid_t ppid = Process_getParentPid(p);
+               bool preExisting;
+               bool nextPreExisting;
+               Process *ppidProcess = ProcessList_getProcess(this, ppid, &preExisting, LinuxProcess_new);
+               Process *nextProcess;
```
BenBE, line 570: No reason to fetch processes from the platform API as there's a perfectly fine list of proceses currently tracked just part of the `ProcessList` instance you are operating on.

ProcessList.c
```diff
@@ -553,13 +554,39 @@ void ProcessList_rebuildPanel(ProcessList* this) {
...
+                   }
+                   preExisting = nextPreExisting;
+                   ppidProcess = nextProcess;
+               }
+               if (ppidProcess->tgid == 0 && !Hashtable_get(this->pidMatchList, ppidProcess->tgid)) {
+                   if (!preExisting) {
+                        free(ppidProcess);
+                   }
+                   continue;
+               }
```
BenBE, line 587: There's a `isChildOf` API available for this kind of filtering; please prefer to use the existing code.

BenBE (changes requested):

jessesimpson36:
I'll get on it. Thanks for the feedback!

ProcessList.c
jessesimpson36: This import is removed.

ProcessList.c
jessesimpson36: I changed the indentation to 3 and fixed the braces on the one-line if statements.

ProcessList.c
jessesimpson36: I removed the else block and de-indented the contents.  good catch.

ProcessList.c
jessesimpson36: I updated it to find processes from the process table.

ProcessList.c
jessesimpson36: I was able to use `isChildOf` for simplifying the while condition, however, I didn't find it helpful for removing the while loop because it didn't search it's parent's parent,  or  parent's grandparent, in the way that iterating over the while loop does. It's definitely simpler now than it was.

I considered adding a `Process_isDescendantOf` function, and pair that with `Hashmap_forEach` to replace the loop, but you'd need a process list built out to pass into a `isDescendantOf` function.

If you have ideas on how to simplify it further, I'd be happy to give them a try.

jessesimpson36:
Marking this as "Ready for review" again, since I've addressed existing comments.

jessesimpson36:
I made a couple more style fixes this morning.

ProcessList.c
```diff
@@ -557,9 +557,24 @@ void ProcessList_rebuildPanel(ProcessList* this) {
...
-         || (this->pidMatchList && !Hashtable_get(this->pidMatchList, p->tgid)) )
+         || (incFilter && !(String_contains_i(Process_getCommand(p), incFilter))) ) {
          continue;
+      }
+
+      if (this->pidMatchList && !Hashtable_get(this->pidMatchList, p->tgid)) {
+         /* Follow processes parents to find matches to process match list */
+         Process* currProc = p;
+         bool found = Hashtable_get(this->pidMatchList, currProc->tgid);
+         while (!Process_isChildOf(currProc, 0) && !found) {
```
tanriol, line 567: In theory it's possible for two processes in one snapshot to list each other as parents in case of really unfortunate PID reuse race. Not sure whether that's worth guarding against - it's really unlikely, but can hang htop completely with an implementation like this.

tanriol: On the other hand, I'm not really sure whether other parts of code are safe against that :-)

BenBE: Good catch and definitively worth investigating.

As reading the process list can take a significant amount of time, this race is a real chance on a VERY busy system. so guarding against it is adviseable, but should be done with minimum performance impact, as this is a hot code path.

jessesimpson36: What are your thoughts on having a max depth that if we hit (lets just say) 10  iterations, we exit the while loop? Maybe having it as an option passed in such as '--pid <pid> --max-child-depth 10'

It doesnt address the possibility of such a pid reuse error affecting the tree building process, but it would be cheap in cpu/memory usage, and avoid getting stuck in the loop.

PR #714, opened 2021-07-29
Graph meter coloring (with GraphData structure rework)

Explorer09:
This pull request implements dynamic scaling support and a "stacked" color for Graph meters. In addition, the Bar meter drawing routine is slightly improved for dynamically expanding the "total" value when needed.

This code would replace the current, one-color graph drawing algorithm so that all Graph meters are drawn with colors.

The data structures and algorithms for the color graphs and dynamically scaling are brand new, totally reworked from htop-dev#129 and the colors are intended to represent the meter data as close as possible. It still uses braille dots to display the graph, but they now have finer details (I think).

Screenshot

----

### Caveats

* The graph now shows only half the number of records after this PR. The old code used to show two records per terminal column (utilizing the two columns of a braille character), but due to coloring in terminal display, I have to change it to one record per terminal column, both in ASCII and Unicode terminal display.
* The new code no longer keeps the graph "values" in the C `double` (floating point) data type. It now stores the precomputed colors and braille dots instead, and the data are specific to a graph height setting.
* The "GRAPH_1" and "GRAPH_2" colors defined in `CRT.h` would become unused.

### Technical details

* The main method of determining the colors of all cells (terminal characters) is the [largest remainder method](https://en.wikipedia.org/wiki/Largest_remainder_method). It ensures the color cells chosen would be representative to the original values of a meter. (Just like how it is used in electing representatives in an electoral system)
* When dynamic scaling and color graph are combined, the graph color computing routine would perform multiple times to render the colors in multiple scales. The "color cells in multiple scales" are then stored in an array with an indexing method mimicking the "in-order traversal" of a complete binary tree. This allows at most 2 × "graph height" cells to be used for storing colors.

----

The following are potential features that can be implemented after this pull request, but I doubt if they are good ideas or if I have the ability to implement them:

* A new color for the scale unit text (`%`, `32K`, `16M`, etc.), separate from the caption color. I was considering making the text blue to separate it from the cyan text of the caption.
* Graph height adjustment at runtime. At least allow the `htoprc` file to specify the graph height. The settings UI would be the next step to implement.
* A "colorblind" display of the graph, for the "monochrome" color scheme. The bar mode characters (`|#*@$%&`) can be reused for this.

PR #913, opened 2022-01-09
Incorporate shared memory in bar text

### LINKED ISSUE #906: memory[bar] should display used+shared as the used memory
Shared memory accounts for memory in `tmpfs` among other things. If `tmpfs` is mounted in `/tmp` then adding a 1GB file will occupy 1GB of ram that is neither `buffer` or `cache` and cannot be freed by kernel on demand. However, htop's `memory[bar]` will only display the shared memory as purple bars but only write `used/total` in numbers at the end of the bar.

In case of a system running out of memory, the bar will look full but the number indicating the memory usage might be very low. `memory[text]` does not have this problem because it will display `available` memory, which will immediately show that the system is low on memory.

I think `memory[bar]` should display `used+shared/total` at the end of the bar to truly reflect the system state. An alternative would be to add an option to  display `available/total` or to display `used/used+shared/total`.

(running htop 3.1.2 on Arch Linux)

Edit: corrected some confusion where I wrote `free` instead of `used' as pointed out by @kjbracey

kjbracey2:
Shared memory is claimed, and as significant a part of the memory load as the "used" memory. Since "shared" was separated from the "used" value, the basic "used/total" display in the bar text has become less meaningful for Linux, as it only reflects a subset of the claimed memory. The difference often isn't huge, but it can become so if tmpfs is heavily used.

Improve the situation by adding used and shared for that text, making it "claimed/total".

In accordance with that logic, move the shared value leftwards next to the used value, so that we're summing the two leftmost values - a contiguous region.

Fixes #906.

(This could be separated into two PRs, but they would conflict - doing it as two patches).

### DIFF
--- a/MemoryMeter.c
+++ b/MemoryMeter.c
@@ -41,8 +41,10 @@ static void MemoryMeter_updateValues(Meter* this) {
       "MEMORY_METER_AVAILABLE is not the last item in MemoryMeterValues");
    this->curItems = MEMORY_METER_AVAILABLE;
 
-   /* we actually want to show "used + compressed" */
+   /* we actually want to show "used + shared + compressed" */
    double used = this->values[MEMORY_METER_USED];
+   if (isPositive(this->values[MEMORY_METER_SHARED]))
+      used += this->values[MEMORY_METER_SHARED];
    if (isPositive(this->values[MEMORY_METER_COMPRESSED]))
       used += this->values[MEMORY_METER_COMPRESSED];
 
--- a/MemoryMeter.c
+++ b/MemoryMeter.c
@@ -19,9 +19,9 @@ in the source distribution for its full text.
 
 static const int MemoryMeter_attributes[] = {
    MEMORY_USED,
-   MEMORY_BUFFERS,
    MEMORY_SHARED,
    MEMORY_COMPRESSED,
+   MEMORY_BUFFERS,
    MEMORY_CACHE
 };
 
@@ -66,10 +66,6 @@ static void MemoryMeter_display(const Object* cast, RichString* out) {
    RichString_appendAscii(out, CRT_colors[METER_TEXT], " used:");
    RichString_appendAscii(out, CRT_colors[MEMORY_USED], buffer);
 
-   Meter_humanUnit(buffer, this->values[MEMORY_METER_BUFFERS], sizeof(buffer));
-   RichString_appendAscii(out, CRT_colors[METER_TEXT], " buffers:");
-   RichString_appendAscii(out, CRT_colors[MEMORY_BUFFERS_TEXT], buffer);
-
    /* shared memory is not supported on all platforms */
    if (isNonnegative(this->values[MEMORY_METER_SHARED])) {
       Meter_humanUnit(buffer, this->values[MEMORY_METER_SHARED], sizeof(buffer));
@@ -83,6 +79,10 @@ static void MemoryMeter_display(const Object* cast, RichString* out) {
       RichString_appendAscii(out, CRT_colors[METER_TEXT], " compressed:");
       RichString_appendAscii(out, CRT_colors[MEMORY_COMPRESSED], buffer);
    }
+
+   Meter_humanUnit(buffer, this->values[MEMORY_METER_BUFFERS], sizeof(buffer));
+   RichString_appendAscii(out, CRT_colors[METER_TEXT], " buffers:");
+   RichString_appendAscii(out, CRT_colors[MEMORY_BUFFERS_TEXT], buffer);
 
    Meter_humanUnit(buffer, this->values[MEMORY_METER_CACHE], sizeof(buffer));
    RichString_appendAscii(out, CRT_colors[METER_TEXT], " cache:");
--- a/darwin/Platform.c
+++ b/darwin/Platform.c
@@ -292,9 +292,9 @@ void Platform_setMemoryValues(Meter* mtr) {
 
    mtr->total = dhost->host_info.max_mem / 1024;
    mtr->values[MEMORY_METER_USED] = (double)(vm->active_count + vm->wire_count) * page_K;
-   mtr->values[MEMORY_METER_BUFFERS] = (double)vm->purgeable_count * page_K;
    // mtr->values[MEMORY_METER_SHARED] = "shared memory, like tmpfs and shm"
    // mtr->values[MEMORY_METER_COMPRESSED] = "compressed memory, like zswap on linux"
+   mtr->values[MEMORY_METER_BUFFERS] = (double)vm->purgeable_count * page_K;
    mtr->values[MEMORY_METER_CACHE] = (double)vm->inactive_count * page_K;
    // mtr->values[MEMORY_METER_AVAILABLE] = "available memory"
 }
--- a/dragonflybsd/Platform.c
+++ b/dragonflybsd/Platform.c
@@ -214,9 +214,9 @@ void Platform_setMemoryValues(Meter* this) {
 
    this->total = host->totalMem;
    this->values[MEMORY_METER_USED] = host->usedMem;
-   this->values[MEMORY_METER_BUFFERS] = host->buffersMem;
    // this->values[MEMORY_METER_SHARED] = "shared memory, like tmpfs and shm"
-   // mtr->values[MEMORY_METER_COMPRESSED] = "compressed memory, like zswap on linux"
+   // this->values[MEMORY_METER_COMPRESSED] = "compressed memory, like zswap on linux"
+   this->values[MEMORY_METER_BUFFERS] = host->buffersMem;
    this->values[MEMORY_METER_CACHE] = host->cachedMem;
    // this->values[MEMORY_METER_AVAILABLE] = "available memory"
 }
--- a/Action.c
+++ b/Action.c
@@ -733,9 +733,9 @@ static Htop_Reaction actionHelp(State* st) {
    mvaddstr(line++, 0, "Memory bar:    ");
    addattrstr(CRT_colors[BAR_BORDER], "[");
    addbartext(CRT_colors[MEMORY_USED], "", "used");
-   addbartext(CRT_colors[MEMORY_BUFFERS_TEXT], "/", "buffers");
    addbartext(CRT_colors[MEMORY_SHARED], "/", "shared");
    addbartext(CRT_colors[MEMORY_COMPRESSED], "/", "compressed");
+   addbartext(CRT_colors[MEMORY_BUFFERS_TEXT], "/", "buffers");
    addbartext(CRT_colors[MEMORY_CACHE], "/", "cache");
    addbartext(CRT_colors[BAR_SHADOW], "          ", "used");
    addbartext(CRT_colors[BAR_SHADOW], "/", "total");
--- a/Action.c
+++ b/Action.c
@@ -733,9 +733,9 @@ static Htop_Reaction actionHelp(State* st) {
    mvaddstr(line++, 0, "Memory bar:    ");
    addattrstr(CRT_colors[BAR_BORDER], "[");
    addbartext(CRT_colors[MEMORY_USED], "", "used");
-   addbartext(CRT_colors[MEMORY_COMPRESSED], "/", "compressed");
    addbartext(CRT_colors[MEMORY_BUFFERS_TEXT], "/", "buffers");
    addbartext(CRT_colors[MEMORY_SHARED], "/", "shared");
+   addbartext(CRT_colors[MEMORY_COMPRESSED], "/", "compressed");
    addbartext(CRT_colors[MEMORY_CACHE], "/", "cache");
    addbartext(CRT_colors[BAR_SHADOW], "          ", "used");
    addbartext(CRT_colors[BAR_SHADOW], "/", "total");
--- a/MemoryMeter.h
+++ b/MemoryMeter.h
@@ -11,9 +11,9 @@ in the source distribution for its full text.
 
 typedef enum {
    MEMORY_METER_USED = 0,
-   MEMORY_METER_BUFFERS = 1,
-   MEMORY_METER_SHARED = 2,
-   MEMORY_METER_COMPRESSED = 3,
+   MEMORY_METER_SHARED = 1,
+   MEMORY_METER_COMPRESSED = 2,
+   MEMORY_METER_BUFFERS = 3,
    MEMORY_METER_CACHE = 4,
    MEMORY_METER_AVAILABLE = 5,
    MEMORY_METER_ITEMCOUNT = 6, // number of entries in this enum
--- a/freebsd/Platform.c
+++ b/freebsd/Platform.c
@@ -229,9 +229,9 @@ void Platform_setMemoryValues(Meter* this) {
 
    this->total = host->totalMem;
    this->values[MEMORY_METER_USED] = host->usedMem;
-   this->values[MEMORY_METER_BUFFERS] = host->buffersMem;
    this->values[MEMORY_METER_SHARED] = host->sharedMem;
    // this->values[MEMORY_METER_COMPRESSED] = "compressed memory, like zswap on linux"
+   this->values[MEMORY_METER_BUFFERS] = host->buffersMem;
    this->values[MEMORY_METER_CACHE] = host->cachedMem;
    // this->values[MEMORY_METER_AVAILABLE] = "available memory"
 
--- a/linux/Platform.c
+++ b/linux/Platform.c
@@ -360,9 +360,9 @@ void Platform_setMemoryValues(Meter* this) {
 
    this->total = host->totalMem;
    this->values[MEMORY_METER_USED] = host->usedMem;
-   this->values[MEMORY_METER_BUFFERS] = host->buffersMem;
    this->values[MEMORY_METER_SHARED] = host->sharedMem;
    this->values[MEMORY_METER_COMPRESSED] = 0; /* compressed */
+   this->values[MEMORY_METER_BUFFERS] = host->buffersMem;
    this->values[MEMORY_METER_CACHE] = host->cachedMem;
    this->values[MEMORY_METER_AVAILABLE] = host->availableMem;
 
--- a/netbsd/Platform.c
+++ b/netbsd/Platform.c
@@ -272,9 +272,9 @@ void Platform_setMemoryValues(Meter* this) {
    const Machine* host = this->host;
    this->total = host->totalMem;
    this->values[MEMORY_METER_USED] = host->usedMem;
-   this->values[MEMORY_METER_BUFFERS] = host->buffersMem;
    // this->values[MEMORY_METER_SHARED] = "shared memory, like tmpfs and shm"
    // this->values[MEMORY_METER_COMPRESSED] = "compressed memory, like zswap on linux"
+   this->values[MEMORY_METER_BUFFERS] = host->buffersMem;
    this->values[MEMORY_METER_CACHE] = host->cachedMem;
    // this->values[MEMORY_METER_AVAILABLE] = "available memory"
 }
--- a/openbsd/Platform.c
+++ b/openbsd/Platform.c
@@ -229,9 +229,9 @@ void Platform_setMemoryValues(Meter* this) {
    usedMem -= buffersMem + cachedMem;
    this->total = host->totalMem;
    this->values[MEMORY_METER_USED] = usedMem;
-   this->values[MEMORY_METER_BUFFERS] = buffersMem;
    // this->values[MEMORY_METER_SHARED] = "shared memory, like tmpfs and shm"
    // this->values[MEMORY_METER_COMPRESSED] = "compressed memory, like zswap on linux"
+   this->values[MEMORY_METER_BUFFERS] = buffersMem;
    this->values[MEMORY_METER_CACHE] = cachedMem;
    // this->values[MEMORY_METER_AVAILABLE] = "available memory"
 }
--- a/solaris/Platform.c
+++ b/solaris/Platform.c
@@ -240,9 +240,9 @@ void Platform_setMemoryValues(Meter* this) {
    const Machine* host = this->host;
    this->total = host->totalMem;
    this->values[MEMORY_METER_USED] = host->usedMem;
-   this->values[MEMORY_METER_BUFFERS] = host->buffersMem;
    // this->values[MEMORY_METER_SHARED] = "shared memory, like tmpfs and shm"
    // this->values[MEMORY_METER_COMPRESSED] = "compressed memory, like zswap on linux"
+   this->values[MEMORY_METER_BUFFERS] = host->buffersMem;
    this->values[MEMORY_METER_CACHE] = host->cachedMem;
    // this->values[MEMORY_METER_AVAILABLE] = "available memory"
 }


kjbracey:
Actually, when shared was separated, I think the bar text was actually displaying `(total - available)/total`, which  included tmpfs, but #838 changed it to `used/total`, losing tmpfs.

I think this is good compromise - tmpfs is back in there, and it's still based it on the displayed bars.

BenBE:
I'm not really happy with changing the order in which values are displayed as this will likely interfere with platforms other than Linux.

Apart from this I don't like introducing so much noise for just changing the values included in the calculation.

Why not `used+buffers+shared` instead?

kjbracey:
> changing the order in which values are displayed as this will likely interfere with platforms other than Linux.

My belief was that it wouldn't affect other systems at all - they don't have a `shared`, so it's a no-op for them. Internally, htop just skips value 1 instead of skipping value 2. This basically repeats #566, moving `shared` one more space.

> Why not `used+buffers+shared` instead?

Good question. As written, it's simply because `buffers` has never been part of `used`. I was thinking of only consequences of the `used`->`used+shared` split.

Research seems to confirm that the `buffers` are potentially reclaimable in Linux - they're part of the cache that /proc/meminfo chooses to separate out. Raw disk blocks or other filing system metadata that don't correspond to file contents, but still potentially as droppable.

Source: https://unix.stackexchange.com/questions/440558/30-of-ram-is-buffers-what-is-it

So I don't think it should be included. And it also changes the answer on other platforms. Included `shared` doesn't, as only Linux has that.

kjbracey:
There's a related point here with respect to the graph display - that basically shows full all the time on Linux, because it's summing `curItems`, so incorporates `cache`.

`curItems` being 4 is appropriate for the bar, which is showing a stack in different colours, so you can distinguish the reclaimable ram+buffers, but it's not good for a simple combined total like the graph.

I would suggest there should be a separate `simpleCurItems` set to 2 (`used` + `shared`) used to get the plotted graph value, and that then corresponds to the label value, and does assume the reordering.

Explorer09:
You should also swap the entries in the Memory Meter help text. The `actionHelp()` function in Action.c.

Explorer09:
@kjbracey 
1. You should also swap the `shared/buffer` entries in the Memory Meter help text. I mean the `actionHelp()` function in Action.c.
2. I have implemented a color graph display in a separate PR (#714), I think your adjustment of `curItems` in MemoryMeter would be unneeded and would potentially break the graph display after my PR is applied.

kjbracey:
Thanks for feedback - I like the graph colouring - that addresses the problem in a different way.

My proposal wouldn't break it. I was suggesting a separate `curSumItems` to indicate how many to show for a simple count (like a monochrome graph, or the text). A stacked display showing regions like the bars or your coloured graph could continue to use `curItems`.

Explorer09:
@kjbracey The `curItems` is useful for hiding certain items from bar and graph display. IIRC, the CPU Meters use it for hiding CPU temperatures as they are not use percentages and the temperatures are only shown in text.
I don't know what your `curSumItems` proposal is meant to do.

kjbracey:
Sounds like I may need to review more carefully the meaning of `curItems`, but if that's a "how many entries to display in a stacked graph", mine is "how many entries to add up to get a summary value". Those are potentially different.

Memory has 5 entries, of which you want 4 in a stacked graph. But given that the cache naturally fills all unused space, it's not that useful to just add the 4 together to get a total if you can't see how much is cache. Excluding the cache gives a more useful number. And that's what the text on the bar does.

`curSumItems` is API for that concept - `MemoryMeter` would set it to 2 (if including shared as per this PR), or 1 (reflecting current behavior). A monochrome graph (or simple bar?) or a generic equivalent of the bar label would use `curSumItems`. A stacking thing like the current bar or your coloured graph would use `curItems`.

Names could be clarified. Maybe `curItems` should be `curStackItems`. It only limits stacked graphical displays, right?

Explorer09:
@kjbracey No, the cache doesn't technically fill all unused space of the memory, in Linux at least.
You only need to change the `txtBuffer` (display text) of the Memory Meter to achieve what you want (add `shared` to the total "used" memory for summary display). I don't see a new `curSumItems` can be any useful for now. Let's keep the changes small.

kjbracey:
> No, the cache doesn't technically fill all unused space of the memory, in Linux at least. 

Indeed, but it tends to fill it up meaning that a simple total including it does not reasonably reflect current memory demands, so is generally less useful than it would be without including it.

And that's why the bar text has never included it. And it's why the monochrome graph isn't very useful, and why your coloured changes are a significant benefit compared to the monochrome, by making the cache perceivable.

I'm not doing the `curSumItems` in this patch - I'm just describing it as a concept to help justify the reordering, making the summed items contiguous. It would be useful to make the monochrome graph work better, making it graph the useful value shown in the bar label. It would have no impact on your graph display, except that that apparently totally replaces monochrome mode, so it would become redundant.

MemoryMeter.c
```diff
@@ -36,7 +37,12 @@ static void MemoryMeter_updateValues(Meter* this) {
    /* Do not print available memory in bar mode */
    this->curItems = 4;
 
-   written = Meter_humanUnit(buffer, this->values[0], size);
+   /* Include shared in basic readout */
+   simple_used = this->values[0];
+   if (!isnan(this->values[1]))
```
Explorer09, line 42: I would prefer this:
```suggestion
   if (this->values[1] > 0.0)
```
(With IEEE 754 semantics, NaN evaluate to `false` in all comparisons except `!=`.)

MemoryMeter.c
```diff
@@ -27,6 +27,7 @@ static void MemoryMeter_updateValues(Meter* this) {
    char* buffer = this->txtBuffer;
    size_t size = sizeof(this->txtBuffer);
    int written;
+   int simple_used;
```
Explorer09, line 30: 1. Use the `double` type. Don't use `int` as it could overflow (> 4 GiB).
2. I suggest the name `unavailable` here. It's difficult to recognize what `simple_used` means.
```suggestion
   double unavailable;
```

MemoryMeter.c
kjbracey: True, but simple `>` on NaNs generates a floating point exception (`==`/`!=` doesn't).  At best that sets a sticky flag, at worst it actually traps and signals.

 `if (isgreater(this->values[1], 0.0))` would be appropriate. That acknowledges you're prepared for the "unordered" result and doesn't generate the exception.

@BenBE thoughts? Neat as that is, I kind of feel the explicit NaN test is more human-friendly.

MemoryMeter.c
kjbracey: Good catch on the type. Not sure what I was thinking there.

`used` is what that number's called in the help text, so I think it should match that. But that is itself an issue, because then wouldn't be the same as the other `used` displays. But then that's also been the case in the past - it did used to be effectively "unavailable", based on the `available` value, but that was changed in favour of using a value based on the bars.

I think `unavailable` would also be a bit off, suggesting an association with `available` that isn't directly there.

How about `unreclaimable`? Or, simpler, just `claimed`? Get rid of the negative.

That's effectively I'm trying to distinguish here. The claimed bars we can't easily displace (except to swap) - used and shared - versus the scavenging bars that will yield - cache and buffers.

This would be a UI question first - would we be okay `claimed/total]` in the help?

kjbracey: Arguably, the problem starts with using the word `used` for only a fraction of the memory. Special areas have been carved out making `used` ever more specific.  But really `used` is the opposite of `free` - all 4 bars are `used`.

It's probably too late to do a full rework now, but more consistent naming would be

    |unshared|shared|buffers|cache|free|
    |-----claimed---|-reclaimable-|free|
    |-------------used------------|free|
    |---------------total--------------|

At least we can use `claimed` here.

MemoryMeter.c
Explorer09: @kjbracey Good point. I never thought of floating point exceptions before (I thought the exceptions won't matter when I wrote the color graph code). But I feel awkward when I am forced to use C99-specific, unreadable floating point macros, just to avoid exceptions anyway.

I always believed `this->values[1] > 0.0` is more straightforward than `isgreater(this->values[1], 0.0)`, and the configure script should provide a compatibility macro that looks something like this: 
```c
#ifndef isgreater
#define isgreater(x,y) ((x) == (x) && (y) == (y) && (x) > (y))
#endif
```
Just my opinion. I'm not a fan of such a macro.

kjbracey: I've worked on systems where FP exception traps (for invalid operation, overflow and divide by zero) were enabled by default, so I'm possibly unusually conscious of exceptions. I think most systems don't enable any traps by default though, so you'd probably get away with it in an application like this without raising `SIGFPE`.

Explorer09: IIRC, FP exception traps are controlled per application, and since htop uses NaNs extensively, enabling traps would generate a lot of noise. I'm not sure htop's stance on FP exception traps. Would @BenBE answer for this?

MemoryMeter.c
BenBE: The order is unshared, buffers, shared, cache, and free at the moment.

I'm impartial about the actual wording, but claimed seems fine with me, although this might cause yet another confusion because of yet another terminology introduced that's possibly in use elsewhere already.

MemoryMeter.c
kjbracey: Yes, it's per-application.  You _shouldn't_ get a lot of noise from using NaNs with traps enabled. The C `NAN` is a "quiet" NaN that just propagate through FP operations.   (IEEE also has signalling NaNs that generate exceptions whenever you touch them, but C doesn't support them).

I think ordering comparisons are the only thing that makes quiet NaNs cause exceptions.

MemoryMeter.c
kjbracey: > The order is unshared, buffers, shared, cache, and free at the moment.

Indeed, but that diagram shows the conceptual problem with that.  

Arguably there might be other ways of grouping the memory, but I can't actually think of one that justifies the current order. `shared` is associated with `used` by being unreclaimable - just a shared/unshared distinction.  `buffers` is associated with `cache` by being disk cache - a block/file distinction - and being reclaimable generally.  (Actually Linux's reclaimable slab is hiding inside `cache` at the minute - not disk, but reclaimable).

So separating `shared` from `used` and `buffers` from `cache` loses two associations, and I can't see anything gained.

It's just an error arising from the `shared` split not being initially fully considered. Original state was:

    |"used"-|buffers|cache|free|
    |claimed|-reclaimable-|free|
    |---------used--------|free|
    |-----------total----------|

`shared` was new, so just defaulted to the end, meaning it was wrongly separated from `used`.

    |unshared|buffers|cache|shared-|free|
    |claimed-|-reclaimable-|claimed|free|   !!! claimed discontiguous  !!! reclaimable not last
    |-------------used-------------|free|
    |---------------total---------------|

#566 noted that it wasn't reclaimable, and reclaimable should be last, so moved it left one. It got closer to `used` (good, but not perfect), and separated `buffers` from `cache` (bad).

    |unshared|buffers|shared-|-cache-|free|
    |claimed-|reclaim|claimed|reclaim|free|   !!! totally discontiguous
    |-------------used---------------|free|
    |---------------total-----------------|

Reordering shared to second just gets us back to where we started.

kjbracey: > I'm impartial about the actual wording, but claimed seems fine with me, although this might cause yet another confusion because of yet another terminology introduced that's possibly in use elsewhere already.

Just as you wrote this I pushed an update with `claimed`. It feels good to me. I can see Linux using "reclaimable" and "unreclaimable", so it uses "unreclaimable" for the same concept. But that feels a bit too heavy if used in isolation - they have uses of "reclaimable" so it's the natural counterpart. We don't have any uses of "reclaimable".

MemoryMeter.c
BenBE: We've been using the `isgreater` macro for ordering, thus in this situation I'd likely go with it to properly convey intent. This also solves unlikely cases where that value could accidently be negative.

FWIW: The overall attitude to NaNs in htop is to treat them as acceptable values (if actual value unknown/indeterminate) and regard them as zero if any calculations need to be performed on them. Thus silent handling as far as the source is concerned.

BenBE:
@kjbracey2 Please rebase this branch and resolve the issues in the discussion as you go. If there's further input you need, feel free to ask.

MemoryMeter.c
kjbracey2: This is now being rebased on top of code that's already using `isnan` for the "compressed" value in the same place. Shall I change both to `isgreater`?

kjbracey2:
Rebased - not sure what to address, aside from the `isnan` - check that.

Note that #1153 had left the help inconsistent with the actual display - this corrects that as it goes, which maybe makes it less clear. Maybe I should add an initial commit fixing that.

kjbracey2:
Okay, separated out the help correction.

The effect is now actually more to move "buffers" - the thing about "summing a discontiguous region" was now already happening with used+compressed and buffers in between.

MemoryMeter.c
kjbracey2: Updated visualisations now that "compressed" is in there too.

Before this patch:

```
|unshared|buffers|shared|compressed|-cache-|free|
|--used--|----unused----|---used---|---unused---|  <- discontiguous regions count as "used" in the label
|---------------------total---------------------|
```

After this patch
```
|unshared|shared|compressed|buffers|cache|free|
|--working set--|-inactive-|----unclaimed-----|
|---------claimed----------|----unclaimed-----|  <- contigious regions count as "claimed" in the label
|---------------------total-------------------|
```

Last chart shows justification for having shared to the left of compressed - it's grouping together the "live" working set.

darwin/Platform.c
```diff
@@ -292,9 +292,9 @@ void Platform_setMemoryValues(Meter* mtr) {
 
    mtr->total = dhost->host_info.max_mem / 1024;
    mtr->values[MEMORY_METER_USED] = (double)(vm->active_count + vm->wire_count) * page_K;
-   mtr->values[MEMORY_METER_BUFFERS] = (double)vm->purgeable_count * page_K;
    // mtr->values[MEMORY_METER_SHARED] = "shared memory, like tmpfs and shm"
    // mtr->values[MEMORY_METER_COMPRESSED] = "compressed memory, like zswap on linux"
+   mtr->values[MEMORY_METER_BUFFERS] = (double)vm->purgeable_count * page_K;
```
kjbracey2, line 297: Note that none of the Platform changes are any longer functionally required, unlike the original version of the patch, thanks to the `MEMORY_METER_` defines.

But they're still here as it seems worth keeping the order consistent everywhere.

MemoryMeter.c
Explorer09: @kjbracey2 Hello. Not sure if you can test it out, but I believe using `isgreater` can yields smaller and slightly faster code than `isnan`. I think htop's use of `isnan` was a mistake at the start, and if possible, be replaced with `isgreater`, `isless` or such operations.

The technical of `isnan` is that it's a type of `fpclassify` function, and it has special handling for "signaling" NaNs so it nevers throws an FP exception. Htop never uses "signaling" NaNs as far as I knew, and `isnan` would be more expensive than a `x != x` operation. Of course, it would be another issue report for replacing all `isnan` calls in htop with cheaper operations.

kjbracey2: C99/C11 don't specify anything about signalling NaNs, so that would have surprised me.

I've just checked, and no, GCC and Clang reduce both `isnan` and `isgreater` to a single floating-point comparison.. It's just `x != x` or `x >= 0`. 

https://godbolt.org/z/5P1jGMTdY

Unless you specify `-fsignaling-nans` to GCC, in which case then it makes a special call. That shouldn't be on for htop.

kjbracey2: Here's the actual code generated in this case:

```
    movsd   (%rax),%xmm0
    ucomisd %xmm1,%xmm1
    jp      skip_add
    addsd   %xmm1,%xmm0
skip_add:
```

kjbracey2: Ah, I've found the stuff about signaling NaNs in the C extension ISO/IEC TS 18661-1, which does mandate the "no exception for signalling NaNs" for `isnan` and others, but that's not part of the core C standard, which can't produce signaling NaNs and mandates nothing about their behaviour. So special `isnan` SNaN handling isn't on by default in compilers.

Explorer09: @kjbracey2 Yes. That's the tricky part of the standards. The IEEE floating point standard defined "signaling" NaNs with a special case that `fpclassify` functions would never throw exceptions on them despite that other operations would throw exceptions on "signaling" NaNs. That's why I add sarcastic quotes on the term "signaling".

Since we can never assume users would set `-fno-signaling-nans` on compiling htop, it might be better for us to avoid the tricky `isnan` operation altogether. Some of the conditionals in htop code would need to check for negative values in addition to NaNs. And for those cases using `isgreater`, `isgreaterequal` etc. would be the future proof way.

MemoryMeter.c
```diff
@@ -40,13 +40,16 @@ static void MemoryMeter_updateValues(Meter* this) {
       "MEMORY_METER_AVAILABLE is not the last item in MemoryMeterValues");
    this->curItems = MEMORY_METER_AVAILABLE;
 
-   /* we actually want to show "used + compressed" */
-   double used = this->values[MEMORY_METER_USED];
+   /* we actually want to show "used + shared + compressed" */
+   double claimed = this->values[MEMORY_METER_USED];
+   if (!isnan(this->values[MEMORY_METER_SHARED])) {
```
BenBE, line 45: Maybe use `isgreater` as suggested in the other discussion with @Explorer09 …

darwin/Platform.c
BenBE: Those order changes seem good. Now with the constants for the various fields you at least have no problem with consistency even if changing around the order.

dragonflybsd/Platform.c
```diff
@@ -213,9 +213,9 @@ void Platform_setMemoryValues(Meter* this) {
 
    this->total = host->totalMem;
    this->values[MEMORY_METER_USED] = host->usedMem;
-   this->values[MEMORY_METER_BUFFERS] = host->buffersMem;
    // this->values[MEMORY_METER_SHARED] = "shared memory, like tmpfs and shm"
    // mtr->values[MEMORY_METER_COMPRESSED] = "compressed memory, like zswap on linux"
```
BenBE, line 217: Potential cleanup commit: `mtr` --> `this`

BenBE (commented):
LGTM. Mostly the `isnan` vs `isgreater` change left.

MemoryMeter.c
kjbracey2: `-fno-signaling-nans` is the default.

If someone decides to use tougher compiler options that reduce optimisation opportunities, and it gets 0.01% slower as a result, I kind of think it's on them, and I think they'll probably cope. I don't see "this could be slightly slower if people used strange compiler options that serve no purpose for this application" as a very strong argument.

And I'm not a fan of `isgreater(x, 0)` in this bit of code because `if (isgreater(x,0)) { y += x; }` just looks like a weird pseudo-optimisation. "Let's not waste time adding 0". Temptation would be to remove the `if`. Or it could indicate that you are expecting negative values. But the actual intent is just to skip NaNs - yet for some reason we're not using `isnan` like everywhere else.

Having the first and only use of `isgreater` in the code base be to check for NaNs, when there are 50 other places using `isnan` seems a bad idea. Any move away from `isnan` should be codebase-wide, setting a clear new style to follow.  (You could have a local `IS_NAN()` macro that just does `x != x`?)

Explorer09: @kjbracey2 When it comes to rendering meter values, there is a `if (isgreater(x,0)) { y += x; }` check as you mentioned. And it's not just for checking NaNs. We are filtering out negative values just in case some meter implementations accidentally put them in.

Yes, I believe replacing `isnan` should be done codebase-wide, but that would be another pull request to be filed.

kjbracey2: There are no `isgreater`s in `main`. Are you thinking of a pending PR?

The rendering code currently has this:

https://github.com/htop-dev/htop/blob/5558c015f6699c9da47d1cb871318c23dedd4919/Meter.c#L336-L339

kjbracey2:
Awaiting clarification on the `isgreater` - @Explorer09 seems to be referring to code that isn't on main.

Explorer09:
> Awaiting clarification on the `isgreater` - @Explorer09 seems to be referring to code that isn't on main.

I stand corrected. The code surely isn't on main. When I was working with the color graph code in my personal branch, I have been using `isgreater` instead of `isnan` in many places, and I thought some of the `isgreater` uses have been merged back to main and so I was confused.

Here is my suggestion: start using `isgreater` for new code now when it makes sense. If the values are assumed to never go negative, then check using `isgreater`, not just `isnan` - it would make the code a lot safer.

BenBE:
@Explorer09 Can you check which PR the changes for `isgreater` are in and try to separate them to their own commit somehow, such that they can be pulled onto `main`? Though only necessary if the PR is somewhat larger otherwise.

intelfx:
Is there something that can be done to get this merged?

I believe the `isPositive()` rework has since landed in main, and #906 is a visible and user-affecting issue -- I just had to explain this exact situation to a very confused associate :-)

intelfx:
I like the ordering changes, +1 on that.

However, I disagree with the wording changes. "Claimed" is completely new terminology that is never used anywhere else, so I don't think it would be helpful to replace an established (even if ambiguous) term with a completely new and unknown one. More importantly, "used" memory is what coreutils' free(1) says. Thus I am of opinion that htop should just match that terminology.

(I have rebased the PR's branch on top of main, dropping wording changes, [here](https://github.com/intelfx/htop/tree/work/pr-913-include_shared).)

BenBE:
> I like the ordering changes, +1 on that.

PR LGTM.

> However, I disagree with the wording changes. "Claimed" is completely new terminology that is never used anywhere else, so I don't think it would be helpful to replace an established (even if ambiguous) term with a completely new and never-before-used word. More importantly, "used" memory is how coreutils' free(1) calls it. Thus I am of opinion that htop should just match that terminology.

ACK.

> (I have rebased the PR's branch on top of main, dropping wording changes, [here](https://github.com/intelfx/htop/tree/work/pr-913-include_shared).)

I pulled in your changes. Thx.

> Is there something that can be done to get this merged?

AFAICS there's some open discussions from @Explorer09 . If those could be marked resolved if no longer applicable we are GTG.

> I believe the `isPositive()` rework has since landed in main, and #906 is a visible and user-affecting issue -- I just had to explain this exact situation to a very confused associate :-)

ACK.

MemoryMeter.c
Explorer09: Since #1271 addressed the issue of `isnan` vs `isgreater` etc., I'm marking this one as resolved.

Explorer09:
@BenBE What open discussion did I have now?

Regarding `isnan` vs `isgreater` etc., the code has been rebased onto the new `main` which incorporated #1271, and thus it could be considered resolved.
The other discussion I saw is about using the "summary value" instead of the naïve sum of the meter. AFAIK, this is discussed separately in #928 (this PR doesn't include the "summary value" part in order to keep the changes small).

What else did I miss?

BenBE:
That was the only one. GTG …

kjbracey2:
> However, I disagree with the wording changes. "Claimed" is completely new terminology that is never used anywhere else, so I don't think it would be helpful to replace an established (even if ambiguous) term with a completely new and never-before-used word. More importantly, "used" memory is how coreutils' free(1) calls it. Thus I am of opinion that htop should just match that terminology.

The corollary then is that the _other_ usage of "used" should be changed, removing the ambiguity within htop, and avoiding using that coreutils term for something different.

PR #916, opened 2022-01-11
Fix out of boundary writes in XUtils

stoeckmann:
It is possible to exceed the unsigned int data type on 64 bit systems
with enough available RAM. Use size_t in all places instead.

Proof of Concept: Create a 4 GB line in .htoprc file and run htop

$ dd if=/dev/zero bs=1024 count=4194304 | tr '\0' 'a' > ~/.htoprc
$ htop
Segmentation fault

Also avoid overflow of stack based "match" array in String_getToken.

### DIFF
--- a/XUtils.c
+++ b/XUtils.c
@@ -161,35 +161,6 @@ void String_freeArray(char** s) {
    free(s);
 }
 
-char* String_getToken(const char* line, const unsigned short int numMatch) {
-   const size_t len = strlen(line);
-   char inWord = 0;
-   unsigned short int count = 0;
-   char match[50];
-
-   size_t foundCount = 0;
-
-   for (size_t i = 0; i < len; i++) {
-      char lastState = inWord;
-      inWord = line[i] == ' ' ? 0 : 1;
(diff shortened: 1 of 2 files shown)


XUtils.c
```diff
@@ -177,6 +178,8 @@ char* String_getToken(const char* line, const unsigned short int numMatch) {
 
       if (inWord == 1) {
          if (count == numMatch && line[i] != ' ' && line[i] != '\0' && line[i] != '\n' && line[i] != (char)EOF) {
+            if (foundCount == sizeof(match) / sizeof(match[0]) - 1)
```
BenBE, line 181: Better rewrite this function to copy the token in one go and allocate a properly sized target buffer instead of this incremental copy.

BenBE (changes requested):

XUtils.c
stoeckmann: I have removed the function after realizing that it is not used anymore.

BenBE (approved):

PR #919, opened 2022-01-12
Help: Linux swap consistency

kjbracey:
On the help screen's depiction of the swap bar, the `/` separator between used and cache should be coloured for consistency with the other bars.

I tried removing the coloured `/`s from the other bars to make them consistent, but found that less visually appealing.


BenBE:
You mean on the help screen?

kjbracey:
> You mean on the help screen?

Aye - that was in the title, but I've edited the body text say it too.

BenBE (approved):

Issue #905, opened 2021-12-27
Memory bar colors are difficult for color-blind people to see


mikehearn:
The memory bar is a bit hard to interpret because the choice of (???) green/yellow? is quite hard to understand. Fortunately there's a blue divider, or at least I think it's meant to be a divider, but it'd be better to use colors that are easier for people like us to tell apart. Stronger/bolder colors are easier, e.g. blue / yellow / red are easier to separate, and using wider bars would also make it easier as there's more pixels to help differentiate.

BenBE:
The colors used in h`htop` are based on the colors provided by the your terminal emulator.

The Memory meter uses green for used memory, dark blue for I/O buffers, magenta for shared memory and brown (dark yellow) for file system cache. Unfortunately there's little that can be done in `htop` as the set of available console colors is quite limited.

Explorer09:
@mikehearn Did you try the monochrome color scheme before reporting this? The point is, we are not sure how htop can improve regarding the use case for colorblind people. Htop has a monochrome mode. Is that not enough for colorblind people?

mikehearn:
Monochrome mode is of course much easier to read, but correspondingly uglier.

Fully understand that there's a limited selection of colors and it's up to me to customize my profile. Have you maybe considered using a full width bar rather than the thin stripes? Generally, colours are easier to see and distinguish in larger blocks.

Explorer09:
@mikehearn What do you mean by a "full width bar"? It might be better to explain it in a picture.

mikehearn:
█ instead of  ▏

Explorer09:
@mikehearn That filled block character isn't part of ASCII, so htop would be unlikely adopt your proposal.
And I don't think it would help colorblind people much.

PR #922, opened 2022-01-13
Color scheme for red/green colorblind

rrobgill:
As discussed in #905 some color contrasts used in the included color schemes are difficult to distinguish by colorblind people.

This additional color scheme avoids using red/green contrast, substituting yellow/blue where possible, and using brightness variation to further increase contrast. 

Signed-off-by: Rob Gill <rrobgill@protonmail.com>


Explorer09:
I don't know if it helps when I post an example screenshot:
Screenshot

I'm not sure I like it, but applying the color scheme seems to reveal problems (that applies even to default color scheme as well):
1. We didn't mark kernel threads with distinct color from user process threads.
2. The default color scheme tend to avoid dark blue color for displaying text (as it is hard to read) and replace them with light/bold blue (this color is different from cyan) through special exceptions in color definitions (e.g. `MEMORY_BUFFERS_TEXT` vs. `MEMORY_BUFFERS`). These special exceptions should be removed as we expand the possible color schemes (in a separate PR). Otherwise you can see in the screenshot, kernel CPU time (`sy:`) and cache in Swap (`Swp cache:`) look too dark.

rrobgill:
Light blue and Cyan are

rrobgill:
I may return after the expanded colorschemes PR

BenBE:
If we are going the route of catering for colorblind people we should not only take the common red/green variant into account, but the somewhat rarer yellow/blue variant too. Otherwise we're kinda back on square one.

PR #923, opened 2022-01-14
FreeBSD: Add support for showing process emulation

jrtc27:
This displays the same output as ps's -o emul, which is the system call
emulation environment, or ABI, in use. This will typically be FreeBSD
ELF32 or ELF64, but can also be Linux ELF32 or Linux ELF64 when running
Linux binaries under FreeBSD's Linuxulator binary compatibility layer.
The column width of 16 is chosen to match KI_EMULNAMELEN's value of 16,
most of which is normally used up as FreeBSD ELF32/64 is 13 characters.

### DIFF
--- a/freebsd/ProcessField.h
+++ b/freebsd/ProcessField.h
@@ -11,6 +11,7 @@ in the source distribution for its full text.
 #define PLATFORM_PROCESS_FIELDS  \
    JID   = 100,                  \
    JAIL  = 101,                  \
+   EMULATION = 102,              \
                                  \
    DUMMY_BUMP_FIELD = CWD,       \
    // End of list
--- a/freebsd/FreeBSDProcess.c
+++ b/freebsd/FreeBSDProcess.c
@@ -49,6 +49,7 @@ const ProcessFieldData Process_fields[LAST_PROCESSFIELD] = {
    [CWD] = { .name = "CWD", .title = "CWD                       ", .description = "The current working directory of the process", .flags = PROCESS_FLAG_CWD, },
    [JID] = { .name = "JID", .title = "JID", .description = "Jail prison ID", .flags = 0, .pidColumn = true, },
    [JAIL] = { .name = "JAIL", .title = "JAIL        ", .description = "Jail prison name", .flags = 0, },
+   [EMULATION] = { .name = "EMULATION", .title = "EMULATION        ", .description = "System call emulation environment (ABI)", .flags = 0, },
 };
 
 Process* FreeBSDProcess_new(const Settings* settings) {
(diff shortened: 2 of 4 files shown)


BenBE (approved):
LGTM.

BenBE:
Just noticed: given this fild is just some strdup of a value provided anyway, you can drop that extra flag in the column definition. Such flags are mainly intended for complex columns like M_LIBS that perform major calculations for each process.

jrtc27:
Oh ok, it looked like strdup was often hidden behind a flag, but if that's not expensive enough to warrant it then that's easy I'll drop the flag

BenBE:
For most such flags (especially on Linux) it's not so much the actual `free_and_xStrdup` that causes the usage of an additional flag, but the preceeding `xReadfile`. Also `free_and_xStrdup` includes a check with `String_eq` (mapping to `strcmp`) to see if the value actually changed. Thus given that the emulation value is provided by the kernel anyway it's essentially free when refreshing the data.

This aside: Do you know which FreeBSD version first provided this field in its `procinfo` structure? Mostly inquiring if it's worth checking for it in configure, because strictly speaking target is "any system with a suiting C99 compiler", which means potentially systems as old as 20+ years … (On Linux kernel baseline is around 2.6.8).

jrtc27:
https://github.com/freebsd/freebsd-src/commit/d688cb4c0db6368d9e83401ec52e20899c739111 says 5.3, which was released about a month after Linux 2.6.8, interestingly.

BenBE:
Any objections on calling the column `EMULATION` or even `EMULATION / ABI`?
(I think just `EMULATION` should suffice).

We've got enough space in a 16 character column, so there's no need to abbreviate things.

jrtc27:
Was apprehensive it might look out of place but it looks completely fine, still plenty of whitespace

freebsd/ProcessField.h
```diff
@@ -11,6 +11,7 @@ in the source distribution for its full text.
 #define PLATFORM_PROCESS_FIELDS  \
    JID   = 100,                  \
    JAIL  = 101,                  \
+   EMULATION = 102,              \
```
jrtc27, line 14: Process.h and linux/ProcessField.h don't bother aligning anything, it's all kept flush, so I didn't go and re-align the JID and JAIL. They should probably be de-aligned themselves though for consistency...

BenBE: If you do, have those stylistic changes as a separate commit.

jrtc27: I'll open a separate trivial PR for that once this is merged

PR #931, opened 2022-01-22
build: use AC_CANONICAL_HOST, not AC_CANONICAL_TARGET

thesamesam:
htop is a program which will be run on CHOST after cross-compilation;
CTARGET is only for a small number of cases where a program itself outputs
code (so you might cross-compile a compiler which spits out code for a third
architecture/platform).

We want to use AC_CANONICAL_HOST to check CHOST for the platform currently
being used to build htop.

The confusion around this issue was compounded by a mistake in autoconf-archive
which has since been fixed (AX_PTHREAD pulled it in incorrectly).

See: https://github.com/libstatgrab/libstatgrab/pull/131
See: https://github.com/fenrus75/powertop/pull/90#discussion_r705803725
Signed-off-by: Sam James <sam@gentoo.org>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -12,7 +12,7 @@ AC_CONFIG_SRCDIR([htop.c])
 AC_CONFIG_AUX_DIR([build-aux])
 AC_CONFIG_HEADERS([config.h])
 
-AC_CANONICAL_TARGET
+AC_CANONICAL_HOST
 AM_INIT_AUTOMAKE([-Wall std-options subdir-objects])
 
 # ----------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


Explorer09:
@thesamesam Why would AC_CANONICAL_TARGET used by htop cause a problem, when the default `--target` is the same as `--host` and when you specify `--host` without explicit `--target`?
There are purposes for configure's `--target` even when the software to be built is not a cross-compiler. For example, I might build htop without binding to any OS's `procfs` interface, in order to try porting to new system. For such a case `--target` is the right option to use.

fasterit:
Sounds reasonable (reviewed the discussion @thesamesam  pointed at).
/DLange

PR #938, opened 2022-02-04
darwin: lazily set process TTY name

charlievieth:
Fetching the TTY name of a process is extremely expensive on darwin and
the call to devname accounts for 95% of htop's CPU usage when there is
high process turnover (this is mostly due to devname calling lstat,
which is incredibly slow). This can make htop unresponsive.

To mitigate this only set the TTY name if the TTY ProcessField is
enabled in the Settings, which by default it is not.

This commit also adds the Settings_hasField helper function, which
returns true if a field is set.

This can be reproduced with the below script:

```sh

set -e

NUM_PROCS=2048
SLEEP_DUR=2

trap 'pkill -P $$' TERM INT
trap 'pkill -P $$' EXIT

while true; do
    echo "info: $NUM_PROCS procs: starting"
    for i in $(seq 1 $NUM_PROCS); do
        sleep 3600 &
    done
    echo "info: $NUM_PROCS procs: complete"
    sleep $SLEEP_DUR
    pkill -P "$$"
done
```

### DIFF
--- a/darwin/DarwinProcess.c
+++ b/darwin/DarwinProcess.c
@@ -12,6 +12,7 @@ in the source distribution for its full text.
 #include <stdlib.h>
 #include <string.h>
 #include <mach/mach.h>
+#include <sys/dirent.h>
 
 #include "CRT.h"
 #include "Process.h"
@@ -26,7 +27,7 @@ const ProcessFieldData Process_fields[LAST_PROCESSFIELD] = {
    [PPID] = { .name = "PPID", .title = "PPID", .description = "Parent process ID", .flags = 0, .pidColumn = true, },
    [PGRP] = { .name = "PGRP", .title = "PGRP", .description = "Process group ID", .flags = 0, .pidColumn = true, },
    [SESSION] = { .name = "SESSION", .title = "SID", .description = "Process's session ID", .flags = 0, .pidColumn = true, },
-   [TTY] = { .name = "TTY", .title = "TTY      ", .description = "Controlling terminal", .flags = 0, },
+   [TTY] = { .name = "TTY", .title = "TTY      ", .description = "Controlling terminal", .flags = PROCESS_FLAG_TTY, },
    [TPGID] = { .name = "TPGID", .title = "TPGID", .description = "Process ID of the fg process group of the controlling terminal", .flags = 0, .pidColumn = true, },
    [MINFLT] = { .name = "MINFLT", .title = "     MINFLT ", .description = "Number of minor faults which have not required loading a memory page from disk", .flags = 0, .defaultSortDesc = true, },
    [MAJFLT] = { .name = "MAJFLT", .title = "     MAJFLT ", .description = "Number of major faults which have required loading a memory page from disk", .flags = 0, .defaultSortDesc = true, },
@@ -276,6 +277,18 @@ static long long int nanosecondsToCentiseconds(uint64_t nanoseconds) {
    return nanoseconds / nanoseconds_per_second * centiseconds_per_second;
 }
 
+static char* DarwinProcess_getDevname(dev_t dev) {
+   if (dev == NODEV) {
+      return NULL;
+   }
+   char buf[sizeof("/dev/") + MAXNAMLEN];
+   char *name = devname_r(dev, S_IFCHR, buf, MAXNAMLEN);
+   if (name) {
+      return xStrdup(name);
+   }
+   return NULL;
+}
+
 void DarwinProcess_setFromKInfoProc(Process* proc, const struct kinfo_proc* ps, bool exists) {
    DarwinProcess* dp = (DarwinProcess*)proc;
 
@@ -306,15 +319,8 @@ void DarwinProcess_setFromKInfoProc(Process* proc, const struct kinfo_proc* ps, 
       proc->isKernelThread = false;
       proc->isUserlandThread = false;
       dp->translated = ps->kp_proc.p_flag & P_TRANSLATED;
-
       proc->tty_nr = ps->kp_eproc.e_tdev;
-      const char* name = (ps->kp_eproc.e_tdev != NODEV) ? devname(ps->kp_eproc.e_tdev, S_IFCHR) : NULL;
-      if (!name) {
-         free(proc->tty_name);
-         proc->tty_name = NULL;
-      } else {
-         free_and_xStrdup(&proc->tty_name, name);
-      }
+      proc->tty_name = NULL;
 
       proc->starttime_ctime = ep->p_starttime.tv_sec;
       Process_fillStarttimeBuffer(proc);
@@ -324,6 +330,23 @@ void DarwinProcess_setFromKInfoProc(Process* proc, const struct kinfo_proc* ps, 
 
       if (proc->settings->ss->flags & PROCESS_FLAG_CWD) {
          DarwinProcess_updateCwd(ep->p_pid, proc);
+      }
+   }
+
+   if (proc->tty_name == NULL && (dev_t)proc->tty_nr != NODEV) {
+      /* The call to devname() is extremely expensive (due to lstat)
+       * and represents ~95% of htop's CPU usage when there is high
+       * process turnover.
+       *
+       * To mitigate this we only fetch TTY information if the TTY
+       * field is enabled in the settings.
+       */
+      if (proc->settings->ss->flags & PROCESS_FLAG_TTY) {
+         proc->tty_name = DarwinProcess_getDevname(proc->tty_nr);
+         if (!proc->tty_name) {
+            /* devname failed: prevent us from calling it again */
+            proc->tty_nr = NODEV;
+         }
       }
    }
 
--- a/darwin/DarwinProcess.h
+++ b/darwin/DarwinProcess.h
@@ -13,6 +13,8 @@ in the source distribution for its full text.
 #include "darwin/DarwinProcessList.h"
 
 
+#define PROCESS_FLAG_TTY 0x00000100
+
 typedef struct DarwinProcess_ {
    Process super;
 


BenBE:
This PR introduces unnecessary code that is better handled with existing features already present: Each column definition has a `flags` member which can be queried for from the settings. HAve a look at that huge list in [DarwinProcess.c](https://github.com/htop-dev/htop/blob/main/darwin/DarwinProcess.c) and in particular compare with how e.g. the [`CWD` flag is handled](https://github.com/htop-dev/htop/blob/265a7b8a501174e2f5c8268a76b5e88d6688ad70/darwin/DarwinProcess.c#L325). Platforms can introduce their own custom flags, as [can be seen on Linux](https://github.com/htop-dev/htop/blob/265a7b8a501174e2f5c8268a76b5e88d6688ad70/linux/LinuxProcess.h#L22).

Settings.c
```diff
@@ -806,3 +806,25 @@ void Settings_setHeaderLayout(Settings* this, HeaderLayout hLayout) {
...
+static bool ProcessField_hasField(const ProcessField* fields, ProcessField field) {
+   if (fields) {
+      while (*fields) {
+         if (*fields++ == field) {
+            return true;
+         }
+      }
+   }
+   return false;
+}
```
BenBE, line 819: Use the `accumulated flags that each screen already has …`

Settings.c
```diff
@@ -806,3 +806,25 @@ void Settings_setHeaderLayout(Settings* this, HeaderLayout hLayout) {
...
+bool Settings_hasField(const Settings* this, ProcessField field) {
+   if (this->screens) {
+      for (unsigned int i = 0; this->screens[i]; i++) {
+         if (ProcessField_hasField(this->screens[i]->fields, field)) {
+            return true;
+         }
+      }
+   }
+   return false;
+}
```
BenBE, line 830: No need to look for information when the screen they are shown on is not visible either.

darwin/DarwinProcess.c
```diff
@@ -327,6 +320,25 @@ void DarwinProcess_setFromKInfoProc(Process* proc, const struct kinfo_proc* ps,
...
+      if (Settings_hasField(proc->settings, TTY)) {
+         const char *name = devname(proc->tty_nr, S_IFCHR);
+         if (name) {
+            proc->tty_name = xStrndup(name, 512);
+         } else {
+            /* devname failed: prevent us from calling it again */
+            proc->tty_nr = NODEV;
+         }
+      }
+   }
```
BenBE, line 340: Rationale good, but use the accumulated `flags` for checking instead. Compare with e.g. how the `CWD` column is handled just above this block of code.

BenBE (changes requested):

charlievieth:
Thanks for the quick review! I'll make these changes when I have some time this weekend or early next week.

darwin/DarwinProcess.c
charlievieth: Good point I totally missed the accumulated `flags`. Fixed and I also removed the `Settings_hasField` and `ProcessField_hasField` functions.

Settings.c
charlievieth: Done

Settings.c
charlievieth: Switched to using the accumulated flags as suggested and this function has been removed.

darwin/DarwinProcess.c
```diff
@@ -327,6 +320,25 @@ void DarwinProcess_setFromKInfoProc(Process* proc, const struct kinfo_proc* ps,
...
+
```
BenBE, line 341: By moving this code block from where it was before you are checking for a terminal on each process information update compared to just when the process was freshly created. And as `proc->tty_nr` doesn't change after the first initialization above this move does not make much sense.

Process.h
```diff
@@ -19,6 +19,7 @@ in the source distribution for its full text.
 
 #define PROCESS_FLAG_IO              0x00000001
 #define PROCESS_FLAG_CWD             0x00000002
+#define PROCESS_FLAG_TTY             0x00000004
```
BenBE, line 22: I don't quite like this flag being all-platform visible, when only Darwin does anything useful with it. Why not create this flag in `darwin/DarwinProcess.h` similar to what I referenced with those flags for Linux.

BenBE (changes requested):
Please squash your commits into one per logically related implementation step (in your case this should result in one commit).

Process.h
charlievieth: Good catch and fixed, I didn't realize that there were OS specific flags.

darwin/DarwinProcess.c
charlievieth: I moved it out of the `!exists` block so that changes to the settings (enabling/disabling the TTY column) will be reflected on each update. Otherwise, if the TTY column is enabled, after `htop` has been running, it will only be shown for processes that are created post-enablement (previously existing processes will not have the TTY field populated).

Also, the purpose of this block is lazily setting the `proc->tty_name` field. We only modify the `proc->tty_nr` if there is an error calling `devname()` and do so to prevent us from calling it again (on the assumption that if it failed once for a given `dev` number it'll continue to fail given that `dev`).

natoscott:
> Fetching the TTY name of a process is extremely expensive [...]

Even though its not quite so expensive everywhere, should we consider a cross-platform solution with a cache like we have for the UID to username mapping?  (see UsersTable.[ch] built on the HashTable code).  I'm pretty sure other platforms will also see speedups with a cache too so most of it could be platform independent.

charlievieth:
> > Fetching the TTY name of a process is extremely expensive [...]
> 
> Even though its not quite so expensive everywhere, should we consider a cross-platform solution with a cache like we have for the UID to username mapping? (see UsersTable.[ch] built on the HashTable code). I'm pretty sure other platforms will also see speedups with a cache too so most of it could be platform independent.

I agree that a `dev` => `TTY name` cache would be nice to have and potentially beneficial. The challenge is that unlike fetching the username, which is standard across OSs (`getpwuid`), getting the TTY name is OS specific (as is the logic for checking if a device number is valid `NODEV`, `PRNODEV`, etc...). So adding a cache requires either: duplicating the cache logic for each platform (which is subtle because we don't want to cache `NULL` values); or using a callback (which gets ugly because of Linux needs more than just the device number: [linux/LinuxProcessList.c#L1524-L1527](https://github.com/htop-dev/htop/blob/265a7b8a501174e2f5c8268a76b5e88d6688ad70/linux/LinuxProcessList.c#L1524-L1527))
 
One other thing to consider (which I don't know the answer to): do device numbers ever get recycled/reused?

That said, I'm happy to make this change, but it might be better to do it in a separate PR.

BenBE:
> One other thing to consider (which I don't know the answer to): do device numbers ever get recycled/reused?

At least AFAIK on Linux the device major/minor translate nearly 1:1 into the device name. Thus if you got `4,42` you can determine the name to be `tty42` even if you deleted and recreated that device. Similarly `4,64` is `ttyS0`. Although nothing prevents you from renaming the associated device file in `/dev`. Other systems may behave similar …

> That said, I'm happy to make this change, but it might be better to do it in a separate PR.

We should be fine without the cache for now.

darwin/DarwinProcess.c
```diff
@@ -327,6 +320,25 @@ void DarwinProcess_setFromKInfoProc(Process* proc, const struct kinfo_proc* ps,
...
+   if (proc->tty_name == NULL && (dev_t)proc->tty_nr != NODEV) {
+      /* The call to devname() is extremely expensive (due to lstat)
+       * and represents ~95% of htop's CPU usage when there is high
+       * process turnover.
+       *
+       * To mitigate this we only fetch TTY information if the TTY
+       * field is enabled in the settings.
+       */
+      if (proc->settings->ss->flags & PROCESS_FLAG_TTY) {
+         const char *name = devname(proc->tty_nr, S_IFCHR);
```
BenBE, line 332: Given macos seems to offer `devname_r` I'd prefer to use it. This also avoids that magic number below with `xStrndup` …

BenBE (changes requested):

natoscott:
> I agree that a `dev` => `TTY name` cache would be nice to have and potentially beneficial. The challenge is that unlike fetching the username, which is standard across OSs (`getpwuid`), getting the TTY name is OS specific (as is the logic for checking if a device number is valid `NODEV`, `PRNODEV`, etc...). So adding a cache requires either: duplicating the cache logic for each platform (which is subtle because we don't want to cache `NULL` values); or using a callback (which gets ugly because of Linux needs more than just the device number: [linux/LinuxProcessList.c#L1524-L1527](https://github.com/htop-dev/htop/blob/265a7b8a501174e2f5c8268a76b5e88d6688ad70/linux/LinuxProcessList.c#L1524-L1527))

Yeah, there's definitely some subtleties there needing careful consideration.  There are existing htop code patterns for handling these kinds of issues.  Platform-specific additions to a cache can be handled similarly to the way each platform handles its own custom system metrics (e.g. struct LinuxProcessList_ with it's 
first 'ProcessList super' field).  Each platform supplies its own implementation wrappers that call through to the underlying 'core' implementation - for platforms with simpler requirements the 'core' implementation might be able to provide pretty much everything beyond the initial OS devname mapping.

> One other thing to consider (which I don't know the answer to): do device numbers ever get recycled/reused?

Good question, not sure off the top of my head.

> That said, I'm happy to make this change, but it might be better to do it in a separate PR.

Definitely worth a separate PR if you take this on.

darwin/DarwinProcess.c
charlievieth: The magic number (512) is just a pedantic guard and can be removed (it should actually be `sizeof("/dev/") + MAXNAMLEN`) and can be removed.

Using `devname` is nice because we don't need thread-safety and not using it requires duplicating it's logic. The below implementation is essentially the same as in [`devname.c`](https://github.com/apple-oss-distributions/Libc/blob/Libc-1439.141.1/gen/devname.c#L103-L114) itself (note: `devename_r` returns NULL if there is no device hence the fixed sized buffer - otherwise we wouldn't be able to distinguish if NULL is returned because the buffer is too small or there is no device associated with `dev`):

```c
#include <stdlib.h>        // devname_r
#include <string.h>        // strcpy
#include <sys/dirent.h> // MAXNAMLEN

char *get_devname(dev_t dev) {
	char buf[sizeof("/dev/") + MAXNAMLEN];
	char *name = devname_r(dev, S_IFCHR, buf, MAXNAMLEN);
	if (name) {
		return strdup(buf);
	}
	return NULL;
}
```
That said, I don't have a strong opinion here.

BenBE: I think providing our own buffer to `devname_r` instead of copying from some hidden static buffer is preferable. And while we don't strictly need to be re-entrant it doesn't hurt and is basically in line what our code does already anyway.

charlievieth: Done

BenBE (approved):

charlievieth:
@BenBE thank you for the review and for merging this PR!

PR #939, opened 2022-02-04
ProcessList: fix quadratic process removal when scanning

charlievieth:
This commit changes ProcessList_scan to remove Processes by index, which
is known, instead of performing a brute-force search by pid.

Searching by pid is potentially quadratic (worst case is: O(n^2)/2) in
ProcessList_scan because the process we are searching for is always the
last process in the vector (the scan starts from the back of the
vector).

This commit also adds a new ProcessList_indexOf function to search for
processes by pid. This is more efficient than Vector_indexOf as it does
not require a comparison function, which cannot be inlined. The now
unused Vector_indexOf and Process_pidCompare functions have been
removed.

### DIFF
--- a/Process.h
+++ b/Process.h
@@ -394,7 +394,11 @@ bool Process_changePriorityBy(Process* this, Arg delta);
 
 bool Process_sendSignal(Process* this, Arg sgn);
 
-int Process_pidCompare(const void* v1, const void* v2);
+static inline int Process_pidEqualCompare(const void* v1, const void* v2) {
+   const pid_t p1 = ((const Process*)v1)->pid;
+   const pid_t p2 = ((const Process*)v2)->pid;
+   return p1 != p2; /* return zero when equal */
+}
 
 int Process_compareByKey_Base(const Process* p1, const Process* p2, ProcessField key);
 
--- a/ProcessList.c
+++ b/ProcessList.c
@@ -158,7 +158,7 @@ void ProcessList_printHeader(const ProcessList* this, RichString* header) {
 }
 
 void ProcessList_add(ProcessList* this, Process* p) {
-   assert(Vector_indexOf(this->processes, p, Process_pidCompare) == -1);
+   assert(Vector_indexOf(this->processes, p, Process_pidEqualCompare) == -1);
    assert(Hashtable_get(this->processTable, p->pid) == NULL);
    p->processList = this;
 
@@ -168,25 +168,22 @@ void ProcessList_add(ProcessList* this, Process* p) {
    Vector_add(this->processes, p);
    Hashtable_put(this->processTable, p->pid, p);
 
-   assert(Vector_indexOf(this->processes, p, Process_pidCompare) != -1);
+   assert(Vector_indexOf(this->processes, p, Process_pidEqualCompare) != -1);
    assert(Hashtable_get(this->processTable, p->pid) != NULL);
-   assert(Hashtable_count(this->processTable) == Vector_count(this->processes));
+   assert(Vector_countEquals(this->processes, Hashtable_count(this->processTable)));
 }
 
-void ProcessList_remove(ProcessList* this, const Process* p) {
-   assert(Vector_indexOf(this->processes, p, Process_pidCompare) != -1);
-   assert(Hashtable_get(this->processTable, p->pid) != NULL);
-
-   const Process* pp = Hashtable_remove(this->processTable, p->pid);
-   assert(pp == p); (void)pp;
-
+// ProcessList_removeIndex removes Process p from the list's map and soft deletes
+// it from its vector. Vector_compact *must* be called once the caller is done
+// removing items.
+static void ProcessList_removeIndex(ProcessList* this, const Process* p, int idx) {
    pid_t pid = p->pid;
-   int idx = Vector_indexOf(this->processes, p, Process_pidCompare);
-   assert(idx != -1);
 
-   if (idx >= 0) {
-      Vector_remove(this->processes, idx);
-   }
+   assert(p == (Process*)Vector_get(this->processes, idx));
+   assert(Hashtable_get(this->processTable, pid) != NULL);
+
+   Hashtable_remove(this->processTable, pid);
+   Vector_softRemove(this->processes, idx);
 
    if (this->following != -1 && this->following == pid) {
       this->following = -1;
@@ -194,7 +191,17 @@ void ProcessList_remove(ProcessList* this, const Process* p) {
    }
 
    assert(Hashtable_get(this->processTable, pid) == NULL);
-   assert(Hashtable_count(this->processTable) == Vector_count(this->processes));
+   assert(Vector_countEquals(this->processes, Hashtable_count(this->processTable)));
+}
+
+void ProcessList_remove(ProcessList* this, const Process* p) {
+   int idx = Vector_indexOf(this->processes, p, Process_pidEqualCompare);
+   assert(0 <= idx && idx < Vector_size(this->processes));
+   if (idx >= 0) {
+      ProcessList_removeIndex(this, p, idx);
+      // This function is public so must not require callers to compact the Vector
+      Vector_compact(this->processes);
+   }
 }
 
 static void ProcessList_buildTreeBranch(ProcessList* this, pid_t pid, int level, int indent, bool show) {
@@ -429,7 +436,7 @@ Process* ProcessList_getProcess(ProcessList* this, pid_t pid, bool* preExisting,
    Process* proc = (Process*) Hashtable_get(this->processTable, pid);
    *preExisting = proc != NULL;
    if (proc) {
-      assert(Vector_indexOf(this->processes, proc, Process_pidCompare) != -1);
+      assert(Vector_indexOf(this->processes, proc, Process_pidEqualCompare) != -1);
       assert(proc->pid == pid);
    } else {
       proc = constructor(this->settings);
@@ -484,7 +491,7 @@ void ProcessList_scan(ProcessList* this, bool pauseProcessUpdate) {
       if (p->tombStampMs > 0) {
          // remove tombed process
          if (this->monotonicMs >= p->tombStampMs) {
-            ProcessList_remove(this, p);
+            ProcessList_removeIndex(this, p, i);
          }
       } else if (p->updated == false) {
          // process no longer exists
@@ -493,10 +500,13 @@ void ProcessList_scan(ProcessList* this, bool pauseProcessUpdate) {
             p->tombStampMs = this->monotonicMs + 1000 * this->settings->highlightDelaySecs;
          } else {
             // immediately remove
-            ProcessList_remove(this, p);
+            ProcessList_removeIndex(this, p, i);
          }
       }
    }
+
+   // Compact the processes vector in case of any deletions
+   Vector_compact(this->processes);
 
    // Set UID column width based on max UID.
    Process_setUidColumnWidth(maxUid);
--- a/Vector.c
+++ b/Vector.c
@@ -29,6 +29,8 @@ Vector* Vector_new(const ObjectClass* type, bool owner, int size) {
    this->items = 0;
    this->type = type;
    this->owner = owner;
+   this->dirty_index = -1;
+   this->dirty_count = 0;
    return this;
 }
 
@@ -44,10 +46,21 @@ void Vector_delete(Vector* this) {
    free(this);
 }
 
+static inline bool Vector_isDirty(const Vector* this) {
+   if (this->dirty_count > 0) {
+      assert(0 <= this->dirty_index && this->dirty_index < this->items);
+      assert(this->dirty_count <= this->items);
+      return true;
+   }
+   assert(this->dirty_index == -1);
+   return false;
+}
+
 #ifndef NDEBUG
 
 static bool Vector_isConsistent(const Vector* this) {
    assert(this->items <= this->arraySize);
+   assert(!Vector_isDirty(this));
 
    if (this->owner) {
       for (int i = 0; i < this->items; i++) {
@@ -60,15 +73,14 @@ static bool Vector_isConsistent(const Vector* this) {
    return true;
 }
 
-unsigned int Vector_count(const Vector* this) {
-   unsigned int items = 0;
+bool Vector_countEquals(const Vector* this, unsigned int expectedCount) {
+   unsigned int n = 0;
    for (int i = 0; i < this->items; i++) {
       if (this->array[i]) {
-         items++;
+         n++;
       }
    }
-   assert(items == (unsigned int)this->items);
-   return items;
+   return n == expectedCount;
 }
 
 Object* Vector_get(const Vector* this, int idx) {
@@ -88,13 +100,16 @@ int Vector_size(const Vector* this) {
 void Vector_prune(Vector* this) {
    assert(Vector_isConsistent(this));
    if (this->owner) {
-      for (int i = 0; i < this->items; i++)
+      for (int i = 0; i < this->items; i++) {
          if (this->array[i]) {
             Object_delete(this->array[i]);
-            //this->array[i] = NULL;
+            this->array[i] = NULL;
          }
+      }
    }
    this->items = 0;
+   this->dirty_index = -1;
+   this->dirty_count = 0;
 }
 
 //static int comparisons = 0;
@@ -240,6 +255,58 @@ Object* Vector_remove(Vector* this, int idx) {
    } else {
       return removed;
    }
+}
+
+Object* Vector_softRemove(Vector* this, int idx) {
+   assert(idx >= 0 && idx < this->items);
+
+   Object* removed = this->array[idx];
+   assert(removed);
+   if (removed) {
+      this->array[idx] = NULL;
+
+      this->dirty_count++;
+      if (this->dirty_index < 0 || idx < this->dirty_index) {
+         this->dirty_index = idx;
+      }
+
+      if (this->owner) {
+         Object_delete(removed);
+         return NULL;
+      }
+   }
+
+   return removed;
+}
+
+void Vector_compact(Vector* this) {
+   if (!Vector_isDirty(this)) {
+      return;
+   }
+
+   const int size = this->items;
+   assert(0 <= this->dirty_index && this->dirty_index < size);
+   assert(this->array[this->dirty_index] == NULL);
+
+   int idx = this->dirty_index;
+
+   /* one deletion: use memmove, which should be faster */
+   if (this->dirty_count == 1) {
+      memmove(&this->array[idx], &this->array[idx + 1], (this->items - idx) * sizeof(this->array[0]));
+   } else {
+      /* multiple deletions */
+      for (int i = idx + 1; i < size; i++) {
+         if (this->array[i]) {
+            this->array[idx++] = this->array[i];
+         }
+      }
+   }
+
+   this->items -= this->dirty_count;
+   this->dirty_index = -1;
+   this->dirty_count = 0;
+
+   assert(Vector_isConsistent(this));
 }
 
 void Vector_moveUp(Vector* this, int idx) {
--- a/Vector.h
+++ b/Vector.h
@@ -22,6 +22,11 @@ typedef struct Vector_ {
    int arraySize;
    int growthRate;
    int items;
+   /* lowest index of a pending soft remove/delete operation,
+      used to speed up compaction */
+   int dirty_index;
+   /* count of soft deletes, required for Vector_count to work in debug mode */
+   int dirty_count;
    bool owner;
 } Vector;
 
@@ -44,6 +49,15 @@ Object* Vector_take(Vector* this, int idx);
 
 Object* Vector_remove(Vector* this, int idx);
 
+/* Vector_softRemove marks the item at index idx for deletion without
+   reclaiming any space. If owned, the item is immediately freed.
+
+   Vector_compact must be called to reclaim space.*/
+Object* Vector_softRemove(Vector* this, int idx);
+
+/* Vector_compact reclaims space free'd up by Vector_softRemove, if any. */
+void Vector_compact(Vector* this);
+
 void Vector_moveUp(Vector* this, int idx);
 
 void Vector_moveDown(Vector* this, int idx);
@@ -54,7 +68,11 @@ void Vector_set(Vector* this, int idx, void* data_);
 
 Object* Vector_get(const Vector* this, int idx);
 int Vector_size(const Vector* this);
-unsigned int Vector_count(const Vector* this);
+
+/* Vector_countEquals returns true if the number of non-NULL items
+   in the Vector is equal to expectedCount. This is only for debugging
+   and consistency checks. */
+bool Vector_countEquals(const Vector* this, unsigned int expectedCount);
 
 #else /* NDEBUG */
 
--- a/Process.c
+++ b/Process.c
@@ -1095,13 +1095,6 @@ bool Process_sendSignal(Process* this, Arg sgn) {
    return kill(this->pid, sgn.i) == 0;
 }
 
-int Process_pidCompare(const void* v1, const void* v2) {
-   const Process* p1 = (const Process*)v1;
-   const Process* p2 = (const Process*)v2;
-
-   return SPACESHIP_NUMBER(p1->pid, p2->pid);
-}
-
 int Process_compare(const void* v1, const void* v2) {
    const Process* p1 = (const Process*)v1;
    const Process* p2 = (const Process*)v2;


ProcessList.c
```diff
@@ -646,7 +660,7 @@ void ProcessList_scan(ProcessList* this, bool pauseProcessUpdate) {
       if (p->tombStampMs > 0) {
          // remove tombed process
          if (this->monotonicMs >= p->tombStampMs) {
-            ProcessList_remove(this, p);
+            ProcessList_removeIndex(this, p, i);
```
BenBE, line 663: This is still quadratic for large numbers of processes …

ProcessList.c
```diff
@@ -655,7 +669,7 @@ void ProcessList_scan(ProcessList* this, bool pauseProcessUpdate) {
             p->tombStampMs = this->monotonicMs + 1000 * this->settings->highlightDelaySecs;
          } else {
             // immediately remove
-            ProcessList_remove(this, p);
+            ProcessList_removeIndex(this, p, i);
```
BenBE, line 672: This is still quadratic for large numbers of processes …

ProcessList.c
```diff
@@ -592,7 +606,7 @@ Process* ProcessList_getProcess(ProcessList* this, pid_t pid, bool* preExisting,
    Process* proc = (Process*) Hashtable_get(this->processTable, pid);
    *preExisting = proc != NULL;
    if (proc) {
-      assert(Vector_indexOf(this->processes, proc, Process_pidCompare) != -1);
+      assert(ProcessList_indexOf(this, proc) != -1);
```
BenBE, line 609: The `indexOf` function is intentionally left with the `Vector` class.

A good optimizer will inline `Process_pidCompare` and optiize the resulting function to effectively what you wrote by hand. No need to maintain separate copies of this code for each scenario.

ProcessList.c
```diff
@@ -194,6 +204,10 @@ void ProcessList_remove(ProcessList* this, const Process* p) {
    assert(Hashtable_count(this->processTable) == Vector_count(this->processes));
 }
 
+void ProcessList_remove(ProcessList* this, const Process* p) {
+   ProcessList_removeIndex(this, p, ProcessList_indexOf(this, p));
+}
+
```
BenBE, line 210: Linear on `Vector_size` …

ProcessList.c
```diff
@@ -165,22 +179,18 @@ void ProcessList_add(ProcessList* this, Process* p) {
...
-   int idx = Vector_indexOf(this->processes, p, Process_pidCompare);
-   assert(idx != -1);
 
+   assert(((Process*)Vector_get(this->processes, idx))->pid == pid);
+   assert(Hashtable_get(this->processTable, pid) != NULL);
+
+   Hashtable_remove(this->processTable, pid);
    if (idx >= 0) {
       Vector_remove(this->processes, idx);
    }
```
BenBE, line 196:
```suggestion
   if (idx >= 0 && idx < Vector_size(this->processes))
      Vector_remove(this->processes, idx);

```

Alternative, make the index check an assertion and unconditionally remove the Vector entry.

ProcessList.c
```diff
@@ -165,22 +179,18 @@ void ProcessList_add(ProcessList* this, Process* p) {
...
-
-   const Process* pp = Hashtable_remove(this->processTable, p->pid);
-   assert(pp == p); (void)pp;
-
+static void ProcessList_removeIndex(ProcessList* this, const Process* p, int idx) {
    pid_t pid = p->pid;
-   int idx = Vector_indexOf(this->processes, p, Process_pidCompare);
-   assert(idx != -1);
 
+   assert(((Process*)Vector_get(this->processes, idx))->pid == pid);
```
BenBE, line 190:
```suggestion
   assert(p == Vector_get(this->processes, idx));
```

BenBE (changes requested):

tanriol:
Removing an entry from a vector while keeping the order of the other entries in it is `O(N)`, where `N` is the vector length, so removing an arbitrary subset of entries one by one will be `O(N^2)` even with this optimization. If you really want to bring it down to linear, you need to combine the removals into one "garbage collection" pass in some way.

charlievieth:
Thank you both for the quick reviews!

> If you really want to bring it down to linear, you need to combine the removals into one "garbage collection" pass in some way.

Agreed and done (https://github.com/htop-dev/htop/pull/939/commits/3d08536ed2b0ef3e305135fea2f30cdc9eb2743e). I didn't want to do this initially because I thought it might be a step too far without prior discussion.

ProcessList.c
charlievieth: Thanks for the quick review!

This commit removes `Vector_indexOf` because it was only used in `ProcessList.c` and only for this one exact task. 

Happy to add it back, but without Link Time Optimization GCC/Clang cannot optimize across translations units AFAIK and it was not inlined here (checked on Linux with gcc 10.3.0). Moving `Vector_indexOf` to `Vector.h` would allow the compiler to potentially inline the comparison function (but no guarantees).

ProcessList.c
charlievieth: Yep, I forgot about the vector reclaiming space via `memmove` as it was not prominent in profiles `sudo perf record -F 99 -g`. It was mostly `Vector_indexOf` that caught my attention.

Vector.h
```diff
@@ -22,6 +22,10 @@ typedef struct Vector_ {
    int arraySize;
    int growthRate;
    int items;
+   /* lowest index of a soft remove/deletes, used to speed up compaction */
+   int dirty_index;
+   /* count of soft deletes, required for Vector_count to work in debug mode */
+   int dirty_count;
```
charlievieth, line 28: Thoughts on "dirty_*" vs "delete_*"? I'm terrible at naming things.

I was hoping to avoid a dirty/delete count because it bloats the struct size, but it's required for `Vector_count` to be accurate. Originally, I wanted to use a `char` (no size penalty if slotted next to `owner`) that would store 3 states: 0 == no soft removes; 1 == one soft remove (so that we can shortcut to memmove in compact); 2 == more than one soft remove.

Vector.c
```diff
@@ -44,10 +46,15 @@ void Vector_delete(Vector* this) {
...
 
+static inline bool Vector_isDirty(const Vector* this) {
+   return this->dirty_count > 0;
+}
+
 #ifndef NDEBUG
 
 static bool Vector_isConsistent(const Vector* this) {
    assert(this->items <= this->arraySize);
+   assert(Vector_isDirty(this) == false);
```
BenBE, line 57:
```suggestion
   assert(!Vector_isDirty(this));
```

You don't say "if the weather is nice is true" either …

Vector.c
```diff
@@ -242,6 +255,56 @@ Object* Vector_remove(Vector* this, int idx) {
...
+   this->dirty_count++;
+   if (this->dirty_index < 0 || idx < this->dirty_index) {
+      this->dirty_index = idx;
+   }
+
+   if (this->owner) {
+      Object_delete(removed);
+      return NULL;
+   }
+   return removed;
```
BenBE, line 273:
```suggestion

   return removed;
```

Just add a blank here; better visual grouping.

Vector.h
```diff
@@ -22,6 +22,10 @@ typedef struct Vector_ {
    int arraySize;
    int growthRate;
    int items;
+   /* lowest index of a soft remove/deletes, used to speed up compaction */
```
BenBE, line 25:
```suggestion
   /* lowest index of a pending soft remove/delete operation, used to speed up compaction */
```

Vector.h
BenBE: I'm fine with `dirty_` unless any of our native speakers objects … ;-)

Also for the count: I feel more comfortable if it stores an absolute number.

Vector.h
```diff
@@ -44,6 +48,15 @@ Object* Vector_take(Vector* this, int idx);
 
 Object* Vector_remove(Vector* this, int idx);
 
+/* Vector_softRemove marks the item at index idx for deletion without
+   reclaiming any space. If owned, the item is immediately deleted.
```
BenBE, line 52:
```suggestion
   reclaiming any space. If owned, the item is immediately freed.
```

BenBE (changes requested):

ProcessList.c
BenBE: There was a subtle issue with your moved implementation: It lacked the consistency check that the original version had. But apart from this: I'm aware that LTO is required for a good optimizer to do its job™, but as things are now there'S quite a few other things that will tackle more runtime then trying to gain some linear speedup from having the optimizer work harder on a bad algorithm. ;-)

charlievieth: Sounds good and I added back `Vector_indexOf` and removed `ProcessList_indexOf`. 

Also for future reference, clang will inline both `Vector_indexOf` and the comparison function if both are defined `static inline ...` in their respective header files. I left `Vector_indexOf` in `Vector.c` as it is used only once and putting it in the header would require exposing `Vector_isConsistent`, which seems unnecessary.

ProcessList.c
charlievieth: Fixed now that we delay compacting the vector until we're done with all deletes in `ProcessList_scan`.

ProcessList.c
charlievieth: Good catch, there's an assertion before this that will fail if `idx` is out of range ([`devname.c#L103-L114`](https://github.com/apple-oss-distributions/Libc/blob/Libc-1439.141.1/gen/devname.c#L103-L114)) so we're good here. But the public function `ProcessList_remove` was not checking if the `Process` was found before calling `ProcessList_removeIndex` so I changed it to check before attempting to remove it. IMHO if `Vector_indexOf` returns a value that is past the end of the vector then we have bigger problems and a segfault might be appropriate.

```c
void ProcessList_remove(ProcessList* this, const Process* p) {
   int idx = Vector_indexOf(this->processes, p, Process_pidEqual);
   assert(idx >= 0);
   if (idx >= 0) {
      ProcessList_removeIndex(this, p, idx);
      // This is public function so do not require
      // callers to compact the Vector
      Vector_compact(this->processes);
   }
}
```

BenBE: Unfortunately an index just off the end of the Vector won't guarantee a SIGSEGV …

ProcessList.c
```diff
@@ -191,6 +188,17 @@ void ProcessList_remove(ProcessList* this, const Process* p) {
    assert(Hashtable_count(this->processTable) == Vector_count(this->processes));
 }
 
+void ProcessList_remove(ProcessList* this, const Process* p) {
+   int idx = Vector_indexOf(this->processes, p, Process_pidEqual);
+   assert(idx >= 0);
+   if (idx >= 0) {
+      ProcessList_removeIndex(this, p, idx);
+      // This is public function so do not require
+      // callers to compact the Vector
```
BenBE, line 197:
```suggestion
      // This function is public so must not require callers to compact the Vector
```

ProcessList.c
```diff
@@ -486,10 +494,12 @@ void ProcessList_scan(ProcessList* this, bool pauseProcessUpdate) {
             p->tombStampMs = this->monotonicMs + 1000 * this->settings->highlightDelaySecs;
          } else {
             // immediately remove
-            ProcessList_remove(this, p);
+            ProcessList_removeIndex(this, p, i);
          }
       }
    }
+   // Compact the processes vector after any deletions.
```
BenBE, line 501:
```suggestion

   // Compact the processes vector in case of any deletions
```

Vector.c
```diff
@@ -61,6 +74,9 @@ static bool Vector_isConsistent(const Vector* this) {
 }
 
 unsigned int Vector_count(const Vector* this) {
+   if (Vector_isDirty(this)) {
+      return this->items - this->dirty_count;
+   }
```
BenBE, line 79: While I know where this comes from, I'm not really happy about this, as this value is potentially used for the upper bound when iterating over the vector, thus hiding the last few items if the vector is iterated from zero to count …

charlievieth: Apologies for the slow response and yeah, this isn't great as it could easily lead to confusion. Since this function is only used in assertions by [`ProcessList_add()`](https://github.com/htop-dev/htop/blob/6133cac721f355aec0d6b5581fef90db34305f29/ProcessList.c#L167) and [`ProcessList_remove()`](https://github.com/htop-dev/htop/blob/6133cac721f355aec0d6b5581fef90db34305f29/ProcessList.c#L191) we could rename it to something like `Vector_countNonNull()`, add a comment about its usage, and fail with `CRT_fatalError()` if `NDEBUG` is set.

A more aggressive option would be to define macros for vector iteration and enforce their use.

```c
#define CONCATX(x, y)      x##y
#define CONCAT(x, y)       CONCATX(x, y)
#define UNIQUE_ALIAS(name) CONCAT(name, __COUNTER__)

/*
 * for_each_if - helper for handling conditionals in various for_each macros
 * @condition: The condition to check
 *
 * Typical use::
 *
 * #define for_each_foo_bar(x, y) \'
 *    list_for_each_entry(x, y->list, head) \'
 *       for_each_if(x->something == SOMETHING)
 *
 * The for_each_if() macro makes the use of for_each_foo_bar() less error
 * prone.
 */
#define for_each_if(condition) if (!(condition)) {} else

/*
 * _Vector_forEach - internal macro to iterate over all entries in Vector.
 *
 * @vec: vector to iterate
 * @value: *Object iteration cursor
 * @__items: int iteration limit, for macro-internal use
 * @__i: int iteration cursor, for macro-internal use
 */
#define _Vector_forEach(vec, value, __items, __i)            \
   const int (__items) = Vector_size(vec);                   \
   for (int (__i) = 0; (__i) < (__items); (__i)++)           \
      for_each_if(((value) = Vector_get((vec), (__i))))

/*
 * Vector_forEach - iterate over all entries in Vector
 * @vec: vector to iterate
 * @value: *Object iteration cursor
 */
#define Vector_forEach(vec, value)                           \
   _Vector_forEach(vec, value, UNIQUE_ALIAS(__items), UNIQUE_ALIAS(__i))

/*
 * _Vector_forEachIndex - internal macro to iterate over all entries in Vector
 * @vec: vector to iterate
 * @__items: int iteration limit, for macro-internal use
 * @__i: int iteration cursor, for macro-internal use
 */
#define _Vector_forEachIndex(vec, __items, __i)              \
   const int (__items) = Vector_size(vec);                   \
   for (int (__i) = 0; (__i) < (__items); (__i)++)           \
      for_each_if(((idx) == __i) || true);

/*
 * Vector_forEachIndex - iterate over all entries in Vector
 * @vec: vector to iterate
 * @idx: int iteration cursor
 */
#define Vector_forEachIndex(vec, idx)                                  \
   for (int __i = 0, __items = Vector_size(vec); __i < __items; __i++) \
      for_each_if(((idx) == __i) || true);

```

charlievieth:
Note: the build failure looks transient ([link](https://github.com/htop-dev/htop/runs/5369688918?check_suite_focus=true)).

BenBE:
The Clang Analyzer build on Ubuntu reports one issue (that is not transient):

```
ProcessList.c: In function ‘ProcessList_removeIndex’:
ProcessList.c:176:13: error: comparison of distinct pointer types lacks a cast [-Werror]
  176 |    assert(p == Vector_get(this->processes, idx));
      |             ^~
```

Vector.c
BenBE: What arguments do we have against simply ignoring the `NULL` elements and just return the number of (`NULL` and non-`NULL`) items?

Having all that macro boilerplate is IMO not helping with making the code clean as it hides quite a bit of control flow.

Vector.c
```diff
@@ -242,6 +261,56 @@ Object* Vector_remove(Vector* this, int idx) {
    }
 }
 
+Object* Vector_softRemove(Vector* this, int idx) {
+   assert(idx >= 0 && idx < this->items);
+
+   Object* removed = this->array[idx];
+   this->array[idx] = NULL;
+
+   this->dirty_count++;
```
BenBE, line 270: Miscounting if deleting an index that is already removed.
There should probably be an `assert(removed)` to check for this situation, plus some code to handle this case further down this function.

Vector.c
```diff
@@ -242,6 +261,56 @@ Object* Vector_remove(Vector* this, int idx) {
...
+   assert(this->array[this->dirty_index] == NULL);
+
+   int idx = this->dirty_index;
+
+   /* one deletion: use memmove, which should be faster */
+   if (this->dirty_count == 1) {
+      memmove(&this->array[idx], &this->array[idx + 1], (this->items - idx) * sizeof(this->array[0]));
+   } else {
+      /* multiple deletions */
+      for (int i = this->dirty_index + 1; i < size; i++) {
```
BenBE, line 299:
```suggestion
      for (int i = idx + 1; i < size; i++) {
```

You could even argue to drop the `+ 1` in the initializer, but that's purely a matter of taste (and one additional loop iteration).

Vector.c
```diff
@@ -242,6 +261,56 @@ Object* Vector_remove(Vector* this, int idx) {
...
+            this->array[idx++] = this->array[i];
+         }
+      }
+   }
+
+   this->items -= this->dirty_count;
+   this->dirty_index = -1;
+   this->dirty_count = 0;
+
+   assert(Vector_count(this) == (unsigned int)this->items);
```
BenBE, line 310: This can be dropped, because it is enforced implicitly by the `Vector_isConsistent` check below.

Vector.c
BenBE: On second thought:

Split `Vector_count` into two routines: `Vector_count` and `Vector_countUnsafe`.

```c
unsigned int Vector_countUnsafe(const Vector* this) {
   return this->items;
}

unsigned int Vector_count(const Vector* this) {
   return Vector_countUnsafe(this);
}
```

Rationale: Sticks with the current API of allowing access to any element, but doesn't throw you of if elements where deleted. The `*Unsafe` is for places that expect missing items, thus pointing out, that the following `Vector_get` calls may return `NULL`. On the other hand, code not expecting missing items will work as is and barf about the inconsistency in development builds. You could even drive this further by Introducing a `Vector_getUnsafe` function with the same semantics. Thus any code not expecting `NULL` values could throw a fatal error upon encountering such invalid indices.

Vector.c
```diff
@@ -88,13 +104,16 @@ int Vector_size(const Vector* this) {
 void Vector_prune(Vector* this) {
    assert(Vector_isConsistent(this));
    if (this->owner) {
-      for (int i = 0; i < this->items; i++)
+      for (int i = 0; i < this->items; i++) {
          if (this->array[i]) {
             Object_delete(this->array[i]);
-            //this->array[i] = NULL;
+            this->array[i] = NULL;
```
BenBE, line 110: NB: Not strictly thread-safe. But since we are single-threaded this is not a problem.
Better approach would be local variable to hold the item, remove it from the list and then `Object_delete''ing it.

Vector.c
charlievieth: Taking another look at this, `Vector_count()` is only used in two places: [`ProcessList_add()`](https://github.com/htop-dev/htop/blob/6133cac721f355aec0d6b5581fef90db34305f29/ProcessList.c#L167) and [`ProcessList_remove()`](https://github.com/htop-dev/htop/blob/6133cac721f355aec0d6b5581fef90db34305f29/ProcessList.c#L191) (to assert that the hash table and vector contain the same number of elements). And the concern is that future devs might use this as an upper bound when iterating over a vector, why don't we remove this function and replace it with something like `Vector_assertCountEquals()` that asserts the vector contains `N` non-null items? This would remove the possibility of confusing this with `Vector_size()`.

Vector.c
charlievieth: I prefer, the `+1` since it makes it clear that the value at `idx` is NULL and is correct. Without the `+1` a careful reader of this code could be left wondering: "why are they starting at an index that is known to be NULL - am I missing something?"

Vector.c
charlievieth: Yeah, nothing here is thread-safe and using a temporary variable won't fix that. Also, the compiler would have to re-order these statements for this to be invalid, which would be a bug.

Vector.c
charlievieth: Fixed

Vector.c
charlievieth: Great catch, fixed.

Vector.c
charlievieth: Also, fixed/

BenBE: The note on `+ 1` was purely a remark; I didn't intend for it to be removed, even though you validly could.

Vector.c
BenBE: The compiler is allowed to write the `NULL` into the memory before it actully performs the call to `Object_delete()` IF it keeps the old contents of that pointer around as the argument to that call.

Vector.c
charlievieth: Also, is this actually an issue since [`Vector_count()`](https://github.com/htop-dev/htop/blob/c4239335b9fb2da6dab1f53ff0e8eb32c78530de/Vector.h#L53-L57) is only defined when `NDEBUG` is set?

Regardless, I replaced `Vector_count` with `Vector_countEquals` in this commit: https://github.com/htop-dev/htop/pull/939/commits/79971f4980beee1086b0316f243e8ea2a191e53c. If you like this commit I'll squash it, otherwise we can revert it and find another solution.

BenBE: The commit message of that commit misses one negation. `Vector_count` is only define if `NDEBUG` is not defined. Also the sentence in that commit message might profit from some grammar simplification. Apart from this I think that renaming is an improvement. Will have to concentrated look at the overall PR again, to finalize my view on this. But I think we're getting close …

charlievieth: Sounds good and I'll fixup that commit message when I rebase/squash it.

charlievieth: Squashed the commit message

charlievieth:
@BenBE thanks for the review - is there anything else that needs to be changed here?

BenBE (approved):

ProcessList.c
```diff
@@ -168,33 +168,40 @@ void ProcessList_add(ProcessList* this, Process* p) {
...
 
    assert(Hashtable_get(this->processTable, pid) == NULL);
-   assert(Hashtable_count(this->processTable) == Vector_count(this->processes));
+   assert(Vector_countEquals(this->processes, Hashtable_count(this->processTable)));
+}
+
+void ProcessList_remove(ProcessList* this, const Process* p) {
+   int idx = Vector_indexOf(this->processes, p, Process_pidEqual);
+   assert(idx >= 0);
+   if (idx >= 0) {
```
BenBE, line 200: Missing bounds check: Should check both lower AND upper bound.

BenBE (commented):
Okay, going through the open discussions on this I noticed one last item that was not addressed yet, which is the bound checking in `ProcessList_remove` and the functions it calls.

ProcessList.c
charlievieth: Fixed!

charlievieth:
> Right now there's one vague idea still floating around, which I remarked upon [here](https://github.com/htop-dev/htop/pull/939#discussion_r820276613), regarding a split of the "safe" API with an explicit "unsafe" API. What are your thoughts on this?

Regarding having a "safe" and "unsafe" API: ideally public functions never leave the vector in an inconsistent state (even if doing so requires extra work such as compacting the vector after a single deletion like [`ProcessList_remove`](https://github.com/htop-dev/htop/blob/19823daf7f198c47fb42b4a692d6c202ac9780a3/ProcessList.c#L197-L205)). This is currently the case as the public functions of the existing ProcessList API never leave the vector in an inconsistent state (aka soft deletion without compaction) and we eliminated the need for `Vector_count`. Going forward, when vector operations need to handle "dirty" vectors (those with pending soft deletions) we shouldn't require the user to know if the vector is dirty or not since that is bug prone (`Vector_countEquals` does this).

TLDR: I think we should strive to have only a "safe" API and only consider adding an "unsafe" API if a situation arises where it is absolutely necessary.

BenBE:
Okay, with that last change fixed, I think this PR is good to go. Unless someone has any objections.

ProcessList.c
```diff
@@ -168,33 +168,40 @@ void ProcessList_add(ProcessList* this, Process* p) {
...
 
-void ProcessList_remove(ProcessList* this, const Process* p) {
-   assert(Vector_indexOf(this->processes, p, Process_pidCompare) != -1);
-   assert(Hashtable_get(this->processTable, p->pid) != NULL);
-
-   const Process* pp = Hashtable_remove(this->processTable, p->pid);
-   assert(pp == p); (void)pp;
-
+// ProcessList_removeIndex removes Process p from the list's map and soft deletes
+// it from it's vector. Vector_compact *must* be called once the caller is done
```
stoeckmann, line 177: it's -> its

Process.h
```diff
@@ -394,7 +394,11 @@ bool Process_changePriorityBy(Process* this, Arg delta);
 
 bool Process_sendSignal(Process* this, Arg sgn);
 
-int Process_pidCompare(const void* v1, const void* v2);
+static inline int Process_pidEqual(const void* v1, const void* v2) {
+   const pid_t p1 = ((const Process*)v1)->pid;
+   const pid_t p2 = ((const Process*)v2)->pid;
+   return p1 != p2; /* return zero when equal */
+}
```
stoeckmann, line 401: I would recommend to name this function "pidDiffer". Right now it returns 0 if pids are equal. From a boolean point of view, that would be false.

stoeckmann (changes requested):

Process.h
BenBE: This function is used as an argument to `Vector_indexOf` where semantics of the return value are like `strcmp`.

PR #942, opened 2022-02-13
LinuxProcessTree: avoid immediate removal

tanriol:
A process dying while being scanned can cause it to be removed from the process list immediately, which violates the "highlight dying processes" mode (if active) and can break the process tree structure due to deceased-but-still-displayed children missing a parent.

This PR closes the race window for `HAVE_OPENAT` systems (assuming that keeping a `/proc/$pid` directory FD causes the expected files in it to be present too) and keeps partially updated processes for `!HAVE_OPENAT` systems.

### DIFF
--- a/ProcessList.c
+++ b/ProcessList.c
@@ -176,6 +176,7 @@ void ProcessList_add(ProcessList* this, Process* p) {
 // ProcessList_removeIndex removes Process p from the list's map and soft deletes
 // it from its vector. Vector_compact *must* be called once the caller is done
 // removing items.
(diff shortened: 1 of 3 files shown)


BenBE (approved):

BenBE:
HAve you had a look if other platform implementations may suffer a similar issue?

tanriol:
According to `grep`, this was the only place in the codebase where `ProcessList_remove` was used in platform-specific code. The remaining uses are both in `ProcessList_scan` cleanup section.

BenBE (approved):

Issue #946, opened 2022-02-15
Incorrect CPU% when CPU utilization more than 100


etanot:
**Type:** Bug report

### Description
When we run a process which utilizes more than 100% of the CPU, `htop` shows only four characters of the value. For example, `100.x` and `247.x` becomes `100.` and `247.` respectively. This happens because of having fixed length of the `CPU%` column. It would be nice, if we could show CPU% value correctly, if not, at least trim the output properly and notify the user.

### Steps to reproduce
- Run a CPU intensive process. For example, run `ffmpeg` command.

```bash
yt-dlp -f 139 'https://youtu.be/a72VZmcRaFU' -o 'audio.m4a'
yt-dlp -f 160 'https://youtu.be/a72VZmcRaFU' -o 'video.mp4'
ffmpeg -i audio.m4a -i video.mp4 out.mp4
```
- Open `htop` and look for process named `ffmpeg`.

### Actual output
In my case `CPU%` value for `ffmpeg` process is `250.`.

![htop-cpu-utilization-greater-than-100](https://user-images.githubusercontent.com/28824347/154057553-cb60f68f-e5a5-4039-8604-d0a92e271cc6.png)

### Expected output
`CPU%` value should also contain value after the decimal, something like `250.2` etc.

PR #947, opened 2022-02-15
Process: Show only integer value when CPU% more than 99.9%

etanot:
When we run a process which utilizes CPU between 100.0% and 999.9%, htop
shows an unnecessary decimal character at the end of the value. For
example, '100.x' and '247.x' become '100.' and '247.' respectively.

When CPU utilization is less than and equal to '99.9%', show the result
with single digit precision and if result is less than four characters,
pad it with the blank space. When CPU utilization is greater than
'99.9%', show only integral part of the result and if it's less than
four characters, pad it with the blank space.

Closes: #946

### Screenshots
**Before:**
before

**After:**
after2

### DIFF
--- a/Process.c
+++ b/Process.c
@@ -741,7 +741,9 @@ void Process_printPercentage(float val, char* buffer, int n, int* attr) {
          xSnprintf(buffer, n, "%4.1f ", val);
       } else {
          *attr = CRT_colors[PROCESS_MEGABYTES];
-         xSnprintf(buffer, n, "%4d ", (int)val);
+         if (val < 100.0F)
+            val = 100.0F; // Don't round down and display "val" as "99".
+         xSnprintf(buffer, n, "%4.0f ", val);
       }
    } else {
       *attr = CRT_colors[PROCESS_SHADOW];
--- a/Process.c
+++ b/Process.c
@@ -739,9 +739,6 @@ void Process_printPercentage(float val, char* buffer, int n, int* attr) {
             *attr = CRT_colors[PROCESS_SHADOW];
          }
          xSnprintf(buffer, n, "%4.1f ", val);
-      } else if (val < 999) {
-         *attr = CRT_colors[PROCESS_MEGABYTES];
-         xSnprintf(buffer, n, "%3d. ", (int)val);
       } else {
          *attr = CRT_colors[PROCESS_MEGABYTES];
          xSnprintf(buffer, n, "%4d ", (int)val);


etanot:
> Also, as we enter the era of multi-core systems with many CPUs it might be worth allowing the CPU column to resize once values even larger than `9999` are needed.

Can I do this in a separate merge request? This merge request contains a simple bug fix, that's value ends with a decimal character, when CPU% is in between `99.9%` and `999.9%`. What you're asking is a feature request (/enhancement) which could be in separate merge request.

Process.c
```diff
@@ -734,14 +734,11 @@ void Process_printLeftAlignedField(RichString* str, int attr, const char* conten
 
 void Process_printPercentage(float val, char* buffer, int n, int* attr) {
    if (val >= 0) {
-      if (val < 99.9F) {
+      if (val <= 99.9F) {
```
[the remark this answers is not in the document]
etanot: My bad, I'm aware of floating comparison, missed this while coding. I just want to handle `99.9` case properly, let me think, whether I could come up with some solution.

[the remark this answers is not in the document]
BenBE: I think leaving this line be should be correct, as it errs on the "is not yet rounded up" side of things. On the "display as integer" side of things you could easily say "I'm rounding to full integers" without this being inconsistent as 99.9 would be rounded up by `printf` too.

Also, testing on godbolt you'll see the following return 100:
```c
#include <stdio.h>

int main(int, char**) {
   printf("%4.0f", 99.9f);
}
```

Thus not even integer conversion necessary to show it properly with rounding.

BenBE:
> > Also, as we enter the era of multi-core systems with many CPUs it might be worth allowing the CPU column to resize once values even larger than `9999` are needed.
> 
> Can I do this in a separate merge request?

Sure.

> This merge request contains a simple bug fix (when CPU% is in between `99.9%` and `999.9%`, value ends with a decimal character). What you're asking is a feature request (/enhancement) which could be in separate merge request.

Yes. It's just that the locations you'll need to edit there are (except for the tracking, which ideally needs to be in `ProcessList_scan`. And usually I'm one of the first to ask for splitting up things, but as long as that column change is a separate commit, I'd be totally fine in this case (otherwise wouldn't've suggested it). ;-)

Explorer09:
Can we please use the "%g" format instead of comparing with 99,9F so that any rounding can be handled by `printf` itself?
I have written an example here: https://github.com/htop-dev/htop/pull/777#issuecomment-913054510

etanot:
> Can we please use the "%g" format instead of comparing with 99,9F so that any rounding can be handled by `printf` itself? I have written an example here: [#777 (comment)](https://github.com/htop-dev/htop/pull/777#issuecomment-913054510)

@Explorer09 But depending on magnitude value, [formatting style of `%g`](https://stackoverflow.com/a/54162153) could be either `%f` (fixed format) or `%e`(scientific format). For example,
```c
#include<stdio.h>

int main()
{
    float a = 123456.654321;
    printf("%g\n", a); // 123457
    float b = 1234567.7654321;
    printf("%g\n", b); // 1.23457e+06

    return 0;
}
```

We could argue that for `%e` formatting, magnitude value is larger. But we still need to handle this, as we don't want to break the program for the systems with very large number of cores.

Explorer09:
@etanot 
Look carefully in my example. I use "%g" only when the value is less than 999.0. This way no scientific notation would be printed.

```c
      // ...
      if (val < 1.0F) {
         xSnprintf(buffer, n, "%4.1f ", val);
      } else if (val <= 999.0F) {
         xSnprintf(buffer, n, "%4.3g ", val);
      } else {
         if (val > 9999.0F) {
            val = 9999.0F;
         }
         xSnprintf(buffer, n, "%4f ", val);
      }
```

Explorer09:
@etanot 
Update: I think I need to revise the example to this. 
Thanks to the [link](https://stackoverflow.com/a/54162153) you provided, `%g` may strip trailing zeros, so `%g` should be used when trailing zeros can never happen. Here, the range where `%g` would be used is [99.1, 999.0).

```c
      // ...
      if (val < 99.1F) {
         xSnprintf(buffer, n, "%4.1f ", val);
      } else if (val < 999.0F) {
         xSnprintf(buffer, n, "%4.3g ", val);
      } else {
         if (val > 9999.0F)
            val = 9999.0F;
         xSnprintf(buffer, n, "%4f ", val);
      }
```

BenBE:
> > I'm not sure `%g` brings us any real benefits here over say hardcoding `%8.3e` for values larger 1e8, using `%*.0f` if not below 99.9 with `*` dynamically sized by the largest value seen in the current run (cf. dynamic column sizing in #891). Below just like now.
> 
> You misunderstood something. My suggestion of using `%g` is only to address the ambiguity of the rounding threshold between 99.9 and 100.0. The current behavior is when CPU percentage reaches 100% no fraction would be shown. But what if the percentage is 99.950, which `printf` could round it up (show as "100.0" and overflows the string buffer) or down (show as "99.9"; no overflow)?

I think this can be avoided much easier by cutting at `<99.9` and properly using `%4.1f` below and `%4.0f` above that threshold.

> Using `%g` on this interval can solve it, although it adds a little more code to htop. Note that this part is unrelated to the column size; it's just whether to show the fraction and the rounding ambiguity between 99.9 and 100.

Using `%g` on some intervals and not on others increases the cognitive hurdle when reading the source. Consistently using `%f` or `%e` to tell exactly what you are expecting makes the whole affair much more obvious IMHO.

Explorer09:
Ahh. I might be overthinking again. BenBE's idea is to just cut off at 99.9 and any number above 99.9 (e.g. `99.90000152587890625F` or `0x1.8F999AP+6F`) would display as 100 due to the round-up of the `%4.0f` format.

I didn't think carefully about that `%4.0f` would always round up the 99.9 value to 100. Sorry, my bad.

Explorer09:
Update: To be precise, some `printf` implementations do respect the rounding mode of the environment ([see this Stack Overflow question ](https://stackoverflow.com/questions/24120888)), so `printf("%4.0f", 99.9)` could display `99.9` when the rounding mode is set to down (`FE_DOWNWARD`) or towards zero (`FE_TOWARDZERO`). The default, round-to nearest mode will display 99.9 as 100, but this little matter is easy to workaround,

Process.c
```diff
@@ -739,9 +739,6 @@ void Process_printPercentage(float val, char* buffer, int n, int* attr) {
             *attr = CRT_colors[PROCESS_SHADOW];
          }
          xSnprintf(buffer, n, "%4.1f ", val);
-      } else if (val < 999) {
-         *attr = CRT_colors[PROCESS_MEGABYTES];
-         xSnprintf(buffer, n, "%3d. ", (int)val);
       } else {
          *attr = CRT_colors[PROCESS_MEGABYTES];
          xSnprintf(buffer, n, "%4d ", (int)val);
```
Explorer09, line 744:
```suggestion
      } else {
         *attr = CRT_colors[PROCESS_MEGABYTES];
         if (val < 100.0F)
            val = 100.0F; // Don't round down or display as "99".
         xSnprintf(buffer, n, "%4.0f ", val);
```

BenBE:
> @BenBE could you review this merge request, again? I'll create a separate merge request for multi-core systems.

Sure, can do. The only thing I see, we should currently still include is the small remark that @Explorer09 gave, as per the discussion on rounding. Apart from this I think we're good to go with this one.

Process.c
etanot: @Explorer09, Thank you! for the suggestion, but I couldn't figure out why do we need following if statement.

```c
if (val < 100.0F)
    val = 100.0F; // Don't round down or display as "99".
```

As we're using `%4.0f` for formatting, it would automatically round all value between `(99.49, 100.49]` to 100. And `if (val < 99.9F)` statement make sure our `val` always greater than `99.9`, so I don't think we above if statement, at all.

BenBE: Problem is that `printf` respects the rounding mode, which could be set to always round down. Under that circumstances `printf("%4.0f", 99.9);` would actually print `  99`, not ` 100` as you'd expect.

etanot:
@BenBE, applied all suggested changes.

BenBE (approved):

Issue #957, opened 2022-03-18
Allow single precision for CPU% greater than 99.9


etanot:
**Type:** Feature request

Since we're dynamically adjusting column width of `CPU%`, we can also show single precision value if `CPU%` greater than `99%`. 
Having single precision for values greater than `99%` makes `CPU%` column consistent for all values. In past, some users have asked for this, see https://github.com/hishamhm/htop/issues/231.

### Potential solution
In my opinion, it would be a good idea to increase the minimum column width of `CPU%` from `4` to `5` characters, unless there is a very good reason not to do so. Since probability of running a process which requires more than `999.9` percent of CPU is very lower than running a process which requires CPU between `99.9` and `999.9`, so width `5` would be a good choice as it would decrease the frequency of screen refresh/redraw.

BenBE:
Actually given a recent change that allowed this dynamic resize of the `CPU%` and `NCPU%` columns, the mentioned change actually isn't too big to implement; mostly boiling down to adjust the limits at which the display format is changed.

PR #959, opened 2022-03-21
Process: Single digit precision for CPU% greater than 99.9%

etanot:
Since commit edf319e53d1fb77546505e238d75160a3febe56e, we're dynamically adjusting column width of
"CPU%", showing single digit precision also for values greater than
"99.9%" makes "CPU%" column consistent with all other values.

Increase the minimum column width of "CPU%" to "5" from "4" characters.
"5" is an optimum choice because it would balance between the frequency
of screen refresh/redraw and minimum column width, as probability of
running a process which requires more than "999.99%" of CPU is lower
than running a process which requires CPU between "99.9%" and "999.9%".

Closes: #957

### Screenshots
cpu
mem
time

### DIFF
--- a/Process.c
+++ b/Process.c
@@ -739,17 +739,20 @@ void Process_printLeftAlignedField(RichString* str, int attr, const char* conten
 
 void Process_printPercentage(float val, char* buffer, int n, uint8_t width, int* attr) {
    if (val >= 0) {
-      if (val < 99.9F) {
-         if (val < 0.05F) {
-            *attr = CRT_colors[PROCESS_SHADOW];
-         }
-         xSnprintf(buffer, n, "%*.1f ", width, val);
-      } else {
+      if (val < 0.05F)
+         *attr = CRT_colors[PROCESS_SHADOW];
+      else if (val >= 99.9F)
          *attr = CRT_colors[PROCESS_MEGABYTES];
-         if (val < 100.0F)
-            val = 100.0F; // Don't round down and display "val" as "99".
-         xSnprintf(buffer, n, "%*.0f ", width, val);
+
+      int precision = 1;
+
+      // Display "val" as "100" for columns like "MEM%".
+      if (width == 4 && val > 99.9F) {
+         precision = 0;
+         val = 100.0F;
       }
+
+      xSnprintf(buffer, n, "%*.*f ", width, precision, val);
    } else {
       *attr = CRT_colors[PROCESS_SHADOW];
       xSnprintf(buffer, n, "%*.*s ", width, width, "N/A");
@@ -1280,13 +1283,14 @@ void Process_updateFieldWidth(ProcessField key, size_t width) {
 }
 
 void Process_updateCPUFieldWidths(float percentage) {
-   if (percentage < 99.9) {
+   if (percentage < 99.9F) {
       Process_updateFieldWidth(PERCENT_CPU, 4);
       Process_updateFieldWidth(PERCENT_NORM_CPU, 4);
       return;
    }
 
-   uint8_t width = ceil(log10(percentage + .2));
+   // Add additional two characters, one for "." and another for precision.
+   uint8_t width = ceil(log10(percentage + 0.1)) + 2;
 
    Process_updateFieldWidth(PERCENT_CPU, width);
    Process_updateFieldWidth(PERCENT_NORM_CPU, width);
--- a/darwin/DarwinProcess.c
+++ b/darwin/DarwinProcess.c
@@ -39,7 +39,7 @@ const ProcessFieldData Process_fields[LAST_PROCESSFIELD] = {
    [M_VIRT] = { .name = "M_VIRT", .title = " VIRT ", .description = "Total program size in virtual memory", .flags = 0, .defaultSortDesc = true, },
    [M_RESIDENT] = { .name = "M_RESIDENT", .title = "  RES ", .description = "Resident set size, size of the text and data sections, plus stack usage", .flags = 0, .defaultSortDesc = true, },
    [ST_UID] = { .name = "ST_UID", .title = "UID", .description = "User ID of the process owner", .flags = 0, },
-   [PERCENT_CPU] = { .name = "PERCENT_CPU", .title = "CPU% ", .description = "Percentage of the CPU time the process used in the last sampling", .flags = 0, .defaultSortDesc = true, .autoWidth = true, },
+   [PERCENT_CPU] = { .name = "PERCENT_CPU", .title = " CPU%", .description = "Percentage of the CPU time the process used in the last sampling", .flags = 0, .defaultSortDesc = true, .autoWidth = true, },
    [PERCENT_NORM_CPU] = { .name = "PERCENT_NORM_CPU", .title = "NCPU%", .description = "Normalized percentage of the CPU time the process used in the last sampling (normalized by cpu count)", .flags = 0, .defaultSortDesc = true, .autoWidth = true, },
    [PERCENT_MEM] = { .name = "PERCENT_MEM", .title = "MEM% ", .description = "Percentage of the memory the process is using, based on resident memory size", .flags = 0, .defaultSortDesc = true, },
    [USER] = { .name = "USER", .title = "USER       ", .description = "Username of the process owner (or user ID if name cannot be determined)", .flags = 0, },
--- a/linux/LinuxProcess.c
+++ b/linux/LinuxProcess.c
@@ -86,9 +86,9 @@ const ProcessFieldData Process_fields[LAST_PROCESSFIELD] = {
    [OOM] = { .name = "OOM", .title = " OOM ", .description = "OOM (Out-of-Memory) killer score", .flags = PROCESS_FLAG_LINUX_OOM, .defaultSortDesc = true, },
    [IO_PRIORITY] = { .name = "IO_PRIORITY", .title = "IO ", .description = "I/O priority", .flags = PROCESS_FLAG_LINUX_IOPRIO, },
 #ifdef HAVE_DELAYACCT
-   [PERCENT_CPU_DELAY] = { .name = "PERCENT_CPU_DELAY", .title = "CPUD%", .description = "CPU delay %", .flags = PROCESS_FLAG_LINUX_DELAYACCT, .defaultSortDesc = true, },
-   [PERCENT_IO_DELAY] = { .name = "PERCENT_IO_DELAY", .title = "IOD% ", .description = "Block I/O delay %", .flags = PROCESS_FLAG_LINUX_DELAYACCT, .defaultSortDesc = true, },
-   [PERCENT_SWAP_DELAY] = { .name = "PERCENT_SWAP_DELAY", .title = "SWPD%", .description = "Swapin delay %", .flags = PROCESS_FLAG_LINUX_DELAYACCT, .defaultSortDesc = true, },
+   [PERCENT_CPU_DELAY] = { .name = "PERCENT_CPU_DELAY", .title = "CPUD% ", .description = "CPU delay %", .flags = PROCESS_FLAG_LINUX_DELAYACCT, .defaultSortDesc = true, },
+   [PERCENT_IO_DELAY] = { .name = "PERCENT_IO_DELAY", .title = " IOD% ", .description = "Block I/O delay %", .flags = PROCESS_FLAG_LINUX_DELAYACCT, .defaultSortDesc = true, },
+   [PERCENT_SWAP_DELAY] = { .name = "PERCENT_SWAP_DELAY", .title = "SWPD% ", .description = "Swapin delay %", .flags = PROCESS_FLAG_LINUX_DELAYACCT, .defaultSortDesc = true, },
 #endif
    [M_PSS] = { .name = "M_PSS", .title = "  PSS ", .description = "proportional set size, same as M_RESIDENT but each page is divided by the number of processes sharing it", .flags = PROCESS_FLAG_LINUX_SMAPS, .defaultSortDesc = true, },
    [M_SWAP] = { .name = "M_SWAP", .title = " SWAP ", .description = "Size of the process's swapped pages", .flags = PROCESS_FLAG_LINUX_SMAPS, .defaultSortDesc = true, },
@@ -270,9 +270,9 @@ static void LinuxProcess_writeField(const Process* this, RichString* str, Proces
       break;
    }
    #ifdef HAVE_DELAYACCT
-   case PERCENT_CPU_DELAY: Process_printPercentage(lp->cpu_delay_percent, buffer, n, 4, &attr); break;
-   case PERCENT_IO_DELAY: Process_printPercentage(lp->blkio_delay_percent, buffer, n, 4, &attr); break;
-   case PERCENT_SWAP_DELAY: Process_printPercentage(lp->swapin_delay_percent, buffer, n, 4, &attr); break;
+   case PERCENT_CPU_DELAY: Process_printPercentage(lp->cpu_delay_percent, buffer, n, 5, &attr); break;
+   case PERCENT_IO_DELAY: Process_printPercentage(lp->blkio_delay_percent, buffer, n, 5, &attr); break;
+   case PERCENT_SWAP_DELAY: Process_printPercentage(lp->swapin_delay_percent, buffer, n, 5, &attr); break;
    #endif
    case CTXT:
       if (lp->ctxt_diff > 1000) {
--- a/linux/LinuxProcess.c
+++ b/linux/LinuxProcess.c
@@ -57,7 +57,7 @@ const ProcessFieldData Process_fields[LAST_PROCESSFIELD] = {
    [M_DRS] = { .name = "M_DRS", .title = " DATA ", .description = "Size of the data segment plus stack usage of the process", .flags = 0, .defaultSortDesc = true, },
    [M_LRS] = { .name = "M_LRS", .title = "  LIB ", .description = "The library size of the process (calculated from memory maps)", .flags = PROCESS_FLAG_LINUX_LRS_FIX, .defaultSortDesc = true, },
    [ST_UID] = { .name = "ST_UID", .title = "UID", .description = "User ID of the process owner", .flags = 0, },
-   [PERCENT_CPU] = { .name = "PERCENT_CPU", .title = "CPU% ", .description = "Percentage of the CPU time the process used in the last sampling", .flags = 0, .defaultSortDesc = true, .autoWidth = true, },
+   [PERCENT_CPU] = { .name = "PERCENT_CPU", .title = " CPU%", .description = "Percentage of the CPU time the process used in the last sampling", .flags = 0, .defaultSortDesc = true, .autoWidth = true, },
    [PERCENT_NORM_CPU] = { .name = "PERCENT_NORM_CPU", .title = "NCPU%", .description = "Normalized percentage of the CPU time the process used in the last sampling (normalized by cpu count)", .flags = 0, .defaultSortDesc = true, .autoWidth = true, },
    [PERCENT_MEM] = { .name = "PERCENT_MEM", .title = "MEM% ", .description = "Percentage of the memory the process is using, based on resident memory size", .flags = 0, .defaultSortDesc = true, },
    [USER] = { .name = "USER", .title = "USER       ", .description = "Username of the process owner (or user ID if name cannot be determined)", .flags = 0, },
--- a/pcp/PCPProcess.c
+++ b/pcp/PCPProcess.c
@@ -54,7 +54,7 @@ const ProcessFieldData Process_fields[] = {
    [M_LRS] = { .name = "M_LRS", .title = "  LIB ", .description = "The library size of the process (unused since Linux 2.6; always 0)", .flags = 0, .defaultSortDesc = true, },
    [M_DT] = { .name = "M_DT", .title = " DIRTY ", .description = "Size of the dirty pages of the process (unused since Linux 2.6; always 0)", .flags = 0, .defaultSortDesc = true, },
    [ST_UID] = { .name = "ST_UID", .title = "UID", .description = "User ID of the process owner", .flags = 0, },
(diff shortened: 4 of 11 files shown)


Process.c
```diff
@@ -1274,13 +1274,14 @@ void Process_updateFieldWidth(ProcessField key, size_t width) {
 }
 
 void Process_updateCPUFieldWidths(float percentage) {
-   if (percentage < 99.9) {
-      Process_updateFieldWidth(PERCENT_CPU, 4);
-      Process_updateFieldWidth(PERCENT_NORM_CPU, 4);
+   if (percentage < 999.9F) {
+      Process_updateFieldWidth(PERCENT_CPU, 5);
+      Process_updateFieldWidth(PERCENT_NORM_CPU, 5);
```
BenBE, line 1279: No need to increase this as this is already captured by the minimum width enforced by the column title.

darwin/DarwinProcess.c
```diff
@@ -39,9 +39,9 @@ const ProcessFieldData Process_fields[LAST_PROCESSFIELD] = {
...
    [ST_UID] = { .name = "ST_UID", .title = "UID", .description = "User ID of the process owner", .flags = 0, },
-   [PERCENT_CPU] = { .name = "PERCENT_CPU", .title = "CPU% ", .description = "Percentage of the CPU time the process used in the last sampling", .flags = 0, .defaultSortDesc = true, .autoWidth = true, },
+   [PERCENT_CPU] = { .name = "PERCENT_CPU", .title = " CPU%", .description = "Percentage of the CPU time the process used in the last sampling", .flags = 0, .defaultSortDesc = true, .autoWidth = true, },
    [PERCENT_NORM_CPU] = { .name = "PERCENT_NORM_CPU", .title = "NCPU%", .description = "Normalized percentage of the CPU time the process used in the last sampling (normalized by cpu count)", .flags = 0, .defaultSortDesc = true, .autoWidth = true, },
-   [PERCENT_MEM] = { .name = "PERCENT_MEM", .title = "MEM% ", .description = "Percentage of the memory the process is using, based on resident memory size", .flags = 0, .defaultSortDesc = true, },
+   [PERCENT_MEM] = { .name = "PERCENT_MEM", .title = " MEM% ", .description = "Percentage of the memory the process is using, based on resident memory size", .flags = 0, .defaultSortDesc = true, },
```
BenBE, line 44: I think we should discuss if we want to increase the size of the `MEM%` column too. I don't think it's strictly necessary.

linux/LinuxProcess.c
```diff
@@ -270,9 +270,9 @@ static void LinuxProcess_writeField(const Process* this, RichString* str, Proces
       break;
    }
    #ifdef HAVE_DELAYACCT
-   case PERCENT_CPU_DELAY: Process_printPercentage(lp->cpu_delay_percent, buffer, n, 4, &attr); break;
-   case PERCENT_IO_DELAY: Process_printPercentage(lp->blkio_delay_percent, buffer, n, 4, &attr); break;
-   case PERCENT_SWAP_DELAY: Process_printPercentage(lp->swapin_delay_percent, buffer, n, 4, &attr); break;
+   case PERCENT_CPU_DELAY: Process_printPercentage(lp->cpu_delay_percent, buffer, n, 5, &attr); break;
+   case PERCENT_IO_DELAY: Process_printPercentage(lp->blkio_delay_percent, buffer, n, 5, &attr); break;
+   case PERCENT_SWAP_DELAY: Process_printPercentage(lp->swapin_delay_percent, buffer, n, 5, &attr); break;
```
BenBE, line 275: Probably should be split into its own commit if not yet done (for clarity of related changes).

Process.c
```diff
@@ -742,8 +742,8 @@ void Process_printPercentage(float val, char* buffer, int n, uint8_t width, int*
       } else {
          *attr = CRT_colors[PROCESS_MEGABYTES];
          if (val < 100.0F)
-            val = 100.0F; // Don't round down and display "val" as "99".
-         xSnprintf(buffer, n, "%*.0f ", width, val);
+            val = 100.0F; // Don't round down and display "val" lower than "100.0".
+         xSnprintf(buffer, n, "%*.1f ", width, val);
       }
```
BenBE, line 747: If both branches of this if end up with the same number format string, we probably come better with restructuring the flow of this `if` statement to have the first part handle coloring and the second part the formatting.

Explorer09:
Can we make this thing tweakable in the display options menu? I'm not sure if it's a good idea to force a minimum width of 5 characters for all percentage columns. I mean, for columns like "NCPU%" and "MEM%" the percentages can't go over 100% and so there is no benefit to give them wider columns or re-format the percentage values to "100.0".

BenBE:
> Can we make this thing tweakable in the display options menu?

Unlikely to happen. Introducing a setting will require it to be supported into the future and might make later adjustments more complicated then necessary.

> I'm not sure if it's a good idea to force a minimum width of 5 characters for all percentage columns. I mean, for columns like "NCPU%" and "MEM%" the percentages can't go over 100% and so there is no benefit to give them wider columns or re-format the percentage values to "100.0".

ACK. Although for `NCPU%` it's forced to 5 digits by the column title …

Process.c
etanot: If column title enforce minimum width of the respective column, why do we need to update field width using `Process_updateFieldWidth` function? Title width of `CPU% ` and `NCPU%` is already `5`, so can't we do something like this:

```c
if (percentage < 999.9F) {
   return;
}
```

This would make sure, we don't do number of characters calculation, if percentage is less than `999.9F`.

BenBE: Actually most values are below 99.9, thus having that case specially hadled by setting it to 4 takes care of most calls. For 999.9 you want that distinction, because there potentially are columns with floating point values with just 4 digits and those would need the handling of 4 vs. 5 digits.

darwin/DarwinProcess.c
etanot: Actually, this is kind of required. If we don't put a space before and after `MEM%`, either `MEM%` column is misaligned or `▽` character is missing.
  - No left whitespace (`MEM% `)
      Title misaligned with its values.
       misaligned-mem

  - No right whitespace (` MEM%`)
       `▽` character is missing.
       missing-characters

Explorer09:
> Although for `NCPU%` it's forced to 5 digits by the column title …

I don't know if you guys would like this idea:
* Change the "normalized CPU percentage" column title from `NCPU%` to `CPU%`, and
* change the "CPU percentage" (non-normalized) column title from `CPU%` to `Core%`.

That would save a character for the former column.

Process.c
etanot: Sorry, I still didn't get it.

> thus having that case specially hadled by setting it to 4 takes care of most calls.

Why do we need to handle (/update) field width, at all? If we're sure minimum width (of `CPU% ` and `NCPU%` is `5`) which is greater than `4`, we can ignore `Process_updateFieldWidth()` function call, when value is less than `99.9F`. And do something like this:

```patch
 void Process_updateCPUFieldWidths(float percentage) {
-   if (percentage < 99.9) {
-      Process_updateFieldWidth(PERCENT_CPU, 4);
-      Process_updateFieldWidth(PERCENT_NORM_CPU, 4);
+   if (percentage < 999.9F) {
       return;
    }
```

BenBE: As there are some columns with minimum of 4 digits you can't do this simplification without forcing all columns to minimum of 5 characters.

Also, including those two calls doesn't hurt, as the `Process_updateFieldWidth` does no expensive processing to update the minimum width value.

darwin/DarwinProcess.c
BenBE: The right whitespace is usually used for the direction arrow.

etanot: @BenBE, how can I align column values with its title without increasing title width of `MEM%` to **6**? I tried different position for white space, but only this combination gives me indented result: `_CPU%` and `_MEM%_`, where `_` represents a white space.

etanot:
@Explorer09, it highly unlikely that title column is going to renamed. htop has been using it for years and renaming could confuse users. Maybe, we should await maintainers' reply.

darwin/DarwinProcess.c
BenBE: The sorting direction is marked in the last column directly after the value. Or to put it a bit different: If you have a column with N characters wide values, there's always another space character added. And the column that holds this space character is also the one that receives the marker for the sorting direction.

In the second screenshot above, actually `Time+` and `Command` start a column to early due to exactly that space character missing after the `MEM%` column heading.

BenBE:
As @etanot guessed correctly, renaming that columns is not planned for exactly that reason. Also renaming that column is something distinct from the topic of this PR, which should cover the proper handling of displaying single digit precision for larger values.

The implementation for this shouldn't assume a minimum column forced by the columns title, but rather should set that minimum width based on the values that should be printed. Hence the `<99` -> `minWidth = 4`. If the column title needs the column to be wider, then there's initialization code handling exactly this already.

NB: The suffix space for the sorting indicator column is not included in the minimal width recorded for the column.

Renaming columns – and in particular swapping names – is likely to cause more confusion with users and is thus unlikely to happen; especially with such thin arguments in support of it.

Explorer09:
@etanot 
Please try this patch. Apply it directly to your [feat/single-precision](https://github.com/etanot/htop/tree/feat/single-precision) branch.

* Reworked `Process_printPercentage()` logic so that the single digit precision would not apply to column widths less than 5.
* The column width of `PERCENT_MEM` is reverted back to 4.
* The "CPU%" column header is right-aligned like UID and PID column headers.
* Removed `autoWidth` property of `PERCENT_NORM_CPU` column.

[cpu-percent-improvement.patch.txt](https://github.com/htop-dev/htop/files/8449100/cpu-percent-improvement.patch.txt)

etanot:
@Explorer09, thank you! for the patch. It fix alignment issue, but I have a question.

> Reworked Process_printPercentage() logic so that the single digit precision would not apply to column widths less than 5.

```c
void Process_printPercentage(float val, char* buffer, int n, uint8_t width, int* attr) {
     ...
      int precision;
      if (width >= 5 || val < 99.9F) {
         precision = 1;
      } else {
         precision = 0;
         if (val < 100.0F)
            val = 100.0F; // Don't round down and display "val" as "99".
      }
     ...
}
```

I think, we don't need this if-else block. Else block never going to executed, `Process_updateCPUFieldWidths()` function (in `Process.c` file) make sure that width of `[99.9, 100.0)` is `5`. Please correct me, if I'm wrong.

```c
void Process_updateCPUFieldWidths(float percentage) {
   ...
   // Add additional two characters, one for "." and other for precision.
   int width = ceil(log10(percentage + 0.1)) + 2;
   ...
}
```

As far as I can understand, you're trying to handle a case where value is less than `100.0`, but rounding makes it `100.0` especially for columns like `MEM%`, where column width is fixed to `4` and showing `100.0` doesn't make any sense. IMHO, following could be a good solution:

- Split print percentage function into two functions.
  We could split `Process_printPercentage()` into two functions, one for columns whose value couldn't go beyond `100` like `MEM%` and other whose could be beyond `100` like `CPU%`. But now, we need to maintain two functions for same task and any change in one function needs change in other, accordingly.

- Handle fix width columns in print percentage function itself.
  We could add a `if` block to check, whether value greater than `99.9` and column title matched in list column whose value couldn't be greater than `100`.

Explorer09:
You're welcome.
 
> I think, we don't need this if-else block. Else block never going to executed, `Process_updateCPUFieldWidths()` function (in `Process.c` file) make sure that width of `[99.9, 100.0)` is `5`. Please correct me, if I'm wrong.

The else block is meant for `MEM%` column, which, unlike `CPU%` columns, has the width of 4.

> As far as I can understand, you're trying to handle a case where value is less than `100.0`, but rounding makes it `100.0` especially for columns like `MEM%`, where column width is fixed to `4` and showing `100.0` doesn't make any sense. IMHO, following could be a good solution:

The `MEM%` column won't expand its width, since the percentage of it never exceeds 100%. Hence the update width function won't apply to it. 
 
>     * Handle fix width columns in print percentage function itself.
>       We could add a `if` block to check, whether value greater than `99.9` and column title matched in list column whose value couldn't be greater than `100`.

Why do that? I have achieved what I want by removing the `autowidth` property of `NCPU%` and `MEM%` columns. What else did I miss?

darwin/DarwinProcess.c
```diff
@@ -39,7 +39,7 @@ const ProcessFieldData Process_fields[LAST_PROCESSFIELD] = {
    [M_VIRT] = { .name = "M_VIRT", .title = " VIRT ", .description = "Total program size in virtual memory", .flags = 0, .defaultSortDesc = true, },
    [M_RESIDENT] = { .name = "M_RESIDENT", .title = "  RES ", .description = "Resident set size, size of the text and data sections, plus stack usage", .flags = 0, .defaultSortDesc = true, },
    [ST_UID] = { .name = "ST_UID", .title = "UID", .description = "User ID of the process owner", .flags = 0, },
-   [PERCENT_CPU] = { .name = "PERCENT_CPU", .title = "CPU% ", .description = "Percentage of the CPU time the process used in the last sampling", .flags = 0, .defaultSortDesc = true, .autoWidth = true, },
+   [PERCENT_CPU] = { .name = "PERCENT_CPU", .title = " CPU%", .description = "Percentage of the CPU time the process used in the last sampling", .flags = 0, .defaultSortDesc = true, .autoWidth = true, },
```
Explorer09: Your next commit changes the same line. You should squash the two changes of the same line into one.

etanot:
Important highlights:
- Rework on `Process_printPercentage()` function's logic.
- Unlike `CPUD%` and `SWAP%`, keep field width of "IOD%" to 4.

darwin/DarwinProcess.c
etanot: Actually, I want to split `CPU%` and `NCPU%` column related changes in different commits. And those change in white space position between `CPU%` and `NCPU%` column is required to maintain alignment between title and their values for the respective commits. For example:

- If we would remove keep `.title = "CPU%"` and `.title = "NCPU%"` in commit https://github.com/htop-dev/htop/pull/959/commits/9d363d9cb34485db542eb2974570ad8511284a62, we'll get misaligned columns.
CPU%-u
CPU%-2-u

- Same things happens when keep title like this: `.title = "CPU%"` and `.title = "NPCU% "`.
NCPU%-u

So having same line change in two different commit allow me keep title aligned for respective commits.

Explorer09: @etanot I know you intended the "CPU%" and "NCPU%" column changes to reside in different commits. What I mean is, in 20fe4bc283f4f8c48eaa892f9250120696019061, you changed the column title of "CPU%" again. That part of the changes should be merged to 9d363d9cb34485db542eb2974570ad8511284a62, so that 20fe4bc283f4f8c48eaa892f9250120696019061 contains only "NCPU%" column change and no "CPU%" column change.

etanot: > https://github.com/htop-dev/htop/commit/20fe4bc283f4f8c48eaa892f9250120696019061 contains only "NCPU%" column change and no "CPU%" column change.

Reason for not doing so is that, commit 9d363d9 would have misaligned title with respect to their values. To reproduce this follow following steps:

- Clone this branch.
- `git checkout 9d363d9`.
- Apply [patch.txt](https://github.com/htop-dev/htop/files/8500983/patch.txt) patch (`git apply patch.txt`)
- Compile htop and run htop.
- Now run a CPU intensive process. For example, run `ffmpeg` command on a video (`ffmpeg -i in.mp4 out.mp4`).

We'll get an output something like this:
misaligned
MEM% column's values are far away from column title.

My question is that, **is it okay to have a commit in a pull request which breaks the alignment for that specific commit, but overall pull request doesn't have this problem?**

Process.c
```diff
@@ -1274,14 +1277,13 @@ void Process_updateFieldWidth(ProcessField key, size_t width) {
 }
 
 void Process_updateCPUFieldWidths(float percentage) {
-   if (percentage < 99.9) {
+   if (percentage < 99.9F) {
       Process_updateFieldWidth(PERCENT_CPU, 4);
```
Explorer09, line 1281:
```suggestion
      Process_updateFieldWidth(PERCENT_CPU, 5);
```

The cause of the misaligned title comes from here. You missed this change from the patch I attached.

etanot: Thank you!

etanot:
@BenBE merge request is ready.

BenBE:
Reduced the patchset somewhat and added some columns missing from the width adjustments.

Process.c
```diff
@@ -1274,13 +1277,14 @@ void Process_updateFieldWidth(ProcessField key, size_t width) {
...
       Process_updateFieldWidth(PERCENT_NORM_CPU, 4);
       return;
    }
 
-   uint8_t width = ceil(log10(percentage + .2));
+   // Add additional two characters, one for "." and other for precision.
+   uint8_t width = ceil(log10(percentage + 0.1)) + 2;
```
Explorer09, line 1287: I'm sorry that I made an imprecise algorithm in my last patch.
Here I make a better one. I avoid the 0.1 value in the addition as 0.1 doesn't have exact representation in binary floating point. I multiply the value by 10 instead (which is exact).

```suggestion
   // The width includes a decimal point and one fraction digit.
   uint8_t width = (int)ceil(log10(percentage * 10.0 + 1.0)) + 1;
```

pcp/PCPProcess.c
```diff
@@ -74,7 +74,7 @@ const ProcessFieldData Process_fields[] = {
    [CGROUP] = { .name = "CGROUP", .title = "    CGROUP ", .description = "Which cgroup the process is in", .flags = PROCESS_FLAG_LINUX_CGROUP, },
    [OOM] = { .name = "OOM", .title = " OOM ", .description = "OOM (Out-of-Memory) killer score", .flags = PROCESS_FLAG_LINUX_OOM, .defaultSortDesc = true, },
    [PERCENT_CPU_DELAY] = { .name = "PERCENT_CPU_DELAY", .title = "CPUD% ", .description = "CPU delay %", .flags = 0, .defaultSortDesc = true, },
-   [PERCENT_IO_DELAY] = { .name = "PERCENT_IO_DELAY", .title = "IOD% ", .description = "Block I/O delay %", .flags = 0, .defaultSortDesc = true, },
+   [PERCENT_IO_DELAY] = { .name = "PERCENT_IO_DELAY", .title = " IOD% ", .description = "Block I/O delay %", .flags = 0, .defaultSortDesc = true, },
```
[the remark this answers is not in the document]
BenBE: To have it the same width as the other two columns.

linux/LinuxProcess.c
```diff
@@ -270,9 +270,9 @@ static void LinuxProcess_writeField(const Process* this, RichString* str, Proces
       break;
    }
    #ifdef HAVE_DELAYACCT
-   case PERCENT_CPU_DELAY: Process_printPercentage(lp->cpu_delay_percent, buffer, n, 4, &attr); break;
-   case PERCENT_IO_DELAY: Process_printPercentage(lp->blkio_delay_percent, buffer, n, 4, &attr); break;
-   case PERCENT_SWAP_DELAY: Process_printPercentage(lp->swapin_delay_percent, buffer, n, 4, &attr); break;
+   case PERCENT_CPU_DELAY: Process_printPercentage(lp->cpu_delay_percent, buffer, n, 5, &attr); break;
+   case PERCENT_IO_DELAY: Process_printPercentage(lp->blkio_delay_percent, buffer, n, 5, &attr); break;
```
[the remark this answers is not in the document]
BenBE: Missed updating this in the commit message; intended behaviour was with it to be 5 characters; as per previous change.

Process.c
Explorer09: When @BenBE pushed a new commit, this width change I suggested to @etanot seems to be reverted. The current code of the PR has the minimum CPU% field width of 4.

darwin/DarwinProcess.c
Explorer09: When @BenBE pushed a new commit, this change I suggested to @etanot seems to be reverted.

There's no need for the beginning padding space for the "CPU%" column, since the CPU% column title is now printed with right alignment.

Process.c
BenBE: Based on the width of the column header it uses a minimum of 5, even though the code would allow it to be 4.

Explorer09: Why should the code allow it to be 4 anyway?

BenBE: Because I'm not special-casing things that do not need special-casing …

The minimum sensible width is 4, thus a check for 99.9 is sufficient. Any column that has a larger caption inherits a wider minimum width automatically without need to do any other checks. That's also why I dropped that one commit while updating the PR.

Explorer09: I thought the function name `Process_updateCPUFieldWidths` implies this is a special case anyway.

I would have been comviced if the function name were `Process_updatePercentageFieldWidth`.

BenBE: I won't mind renaming that function.

darwin/DarwinProcess.c
```diff
@@ -39,7 +39,7 @@ const ProcessFieldData Process_fields[LAST_PROCESSFIELD] = {
...
    [PERCENT_NORM_CPU] = { .name = "PERCENT_NORM_CPU", .title = "NCPU%", .description = "Normalized percentage of the CPU time the process used in the last sampling (normalized by cpu count)", .flags = 0, .defaultSortDesc = true, .autoWidth = true, },
```
Explorer09, line 43: I suggested in my last patch the `autowidth` property of "NCPU%" be removed. Why wasn't that incorporated? Is there a reason to keep the `autowidth` property of it?

Process.c
BenBE: The point of the `+0.1` is less about precision, but about over-estimating the needed number of digits.

darwin/DarwinProcess.c
BenBE: The width handling of columns with and without auto-width is subtly different and keeping it for now keeps the patch set simpler.

PR #960, opened 2022-03-24
README: Add Archlinux build dependency install command

HeshamTB:
Add Archlinux build dependency install command

### DIFF
--- a/README
+++ b/README
@@ -62,6 +62,11 @@ sudo apt install libncursesw5-dev autotools-dev autoconf build-essential
 sudo dnf install ncurses-devel automake autoconf gcc
 ~~~
 
+**Archlinux/Manjaro**
+~~~ shell
+sudo pacman -S ncurses automake autoconf gcc
+~~~
+
 ### Compile from source:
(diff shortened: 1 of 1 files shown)


BenBE (approved):

eworm-de:
Looks like my review was ignored...

BenBE:
Huh? Didn't notice any review comments … Sure you submitted it?

image
image

eworm-de:
Pretty sure...
Screenshot_20220324-162202

No idea why this is not visible for you...

BenBE:
> Pretty sure...
> 
> [Screenshot]
> 
> No idea why this is not visible for you...

You see that yellow "pending"? That's telling it's only prepared for review submission, but not yet actually submitted as a review. You also need to hit "Submit review" in the upper right corner (green button) to actually submit your comments.

As per your comment: Adding the `--needed` flag won't hurt I guess, but given base-devel (or corresponding other packages on other platforms) aren't mentioned I don't see a good argument for this change, unless we'd include it for all (i.e. `build-essential` on Debian and so on).

eworm-de:
Ah, did not notice... Thought is is pending due to missing answer. 😝

Debian does have `build-essential` listed, no?

PR #963, opened 2022-03-25
Allow multiple filters and search strings

### LINKED ISSUE #961: Is there a way to filter for more than one term?
If there are a handful of processes running that I wish to filter on, is there a way to display only those?

fasterit:
Closes #961

### DIFF
--- a/XUtils.c
+++ b/XUtils.c
@@ -94,8 +94,22 @@ void* xReallocArrayZero(void* ptr, size_t prevmemb, size_t newmemb, size_t size)
    return ret;
 }
 
-inline bool String_contains_i(const char* s1, const char* s2) {
-   return strcasestr(s1, s2) != NULL;
+inline bool String_contains_i(const char* s1, const char* s2, bool multi) {
+   // we have a multi-string search term, handle as special case for performance reasons
+   if (multi && strstr(s2, "|")) {
+      size_t nNeedles;
+      char** needles = String_split(s2, '|', &nNeedles);
+      for (size_t i = 0; i < nNeedles; i++) {
+         if (strcasestr(s1, needles[i]) != NULL) {
+            String_freeArray(needles);
+            return true;
+         }
+       }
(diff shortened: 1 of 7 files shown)


generic/uname.c
```diff
@@ -85,7 +85,7 @@ char* Generic_uname(void) {
 
       if (uname_result == 0) {
          size_t written = xSnprintf(savedString, sizeof(savedString), "%s %s [%s]", uname_info.sysname, uname_info.release, uname_info.machine);
-         if (!String_contains_i(savedString, distro) && sizeof(savedString) > written)
+         if (!String_eq(savedString, distro) && sizeof(savedString) > written)
```
BenBE, line 88: This is not case-insensitive IIRC.

XUtils.c
```diff
@@ -95,7 +95,21 @@ void* xReallocArrayZero(void* ptr, size_t prevmemb, size_t newmemb, size_t size)
 }
 
 inline bool String_contains_i(const char* s1, const char* s2) {
```
BenBE, line 97: I don't think this should be implemented as part of this function. Make the search logic it's own function as this may interfere with existing other callers.

BenBE (changes requested):

XUtils.c
fasterit: I moved the only other caller out of the way. That is the other commit :-)

BenBE: Ah, that other remark I had … ;-)

generic/uname.c
BenBE: Yes. And the previous call allowed for substrings, which your new implementation also forbids.

fasterit: And FreeBSD seems to need that, so dropped that commit

BenBE (approved):
The repeated `String_split` call may cause memory fragmentation and thus cause performance to degrade noticeably on a busy system with userland threads showing.

PR #971, opened 2022-04-05
Fix a compile warning: comparison of unsigned enum expression < 0 is …

hwangcc23:
…always false

Get this warning when compiling Settings.c on the Mac OS X with clang-800.0.42.1.
Settings.c:447:28: warning: comparison of unsigned enum expression < 0 is always false [-Wtautological-compare]
         if (this->hLayout < 0 || this->hLayout >= LAST_HEADER_LAYOUT)
             ~~~~~~~~~~~~~ ^ ~

This patch fixes the problem.

### DIFF
--- a/HeaderLayout.h
+++ b/HeaderLayout.h
@@ -18,6 +18,7 @@ in the source distribution for its full text.
 
 
 typedef enum HeaderLayout_ {
+   HF_INVALID = -1,
    HF_TWO_50_50,
    HF_TWO_33_67,
    HF_TWO_67_33,


HeaderLayout.h
```diff
@@ -18,6 +18,7 @@ in the source distribution for its full text.
 
 
 typedef enum HeaderLayout_ {
+   HEADER_LAYOUT_INVALID = -1,
```
BenBE, line 21: I'd prefer consistent naming. Apart from this LGTM.

```suggestion
   HF_INVALID = -1,
```

PR #978, opened 2022-04-18
Process: Fix PID & UID column widths off-by-one error

Explorer09:
If the max PID or UID value for a platform is exactly a power of ten
(10000, 100000, etc.) the column widths of PID and UID would be 1 char
less than the correct number of digits. This is caused by the wrong
rounding function (ceil(x)); change to the correct one (trunc(x) + 1).

### DIFF
--- a/Process.c
+++ b/Process.c
@@ -54,7 +54,7 @@ void Process_setupColumnWidths() {
       return;
    }
 
-   Process_pidDigits = ceil(log10(maxPid));
+   Process_pidDigits = (int)log10(maxPid) + 1;
    assert(Process_pidDigits <= PROCESS_MAX_PID_DIGITS);
 }
 
@@ -64,7 +64,7 @@ void Process_setUidColumnWidth(uid_t maxUid) {
       return;
    }
 
-   Process_uidDigits = ceil(log10(maxUid));
+   Process_uidDigits = (int)log10(maxUid) + 1;
    assert(Process_uidDigits <= PROCESS_MAX_UID_DIGITS);
 }
 


Explorer09:
P.S. This bug can be demonstrated by overriding the PID and UID of the Process objects of htop.
```diff
--- a/ProcessList.c
+++ b/ProcessList.c
@@ -477,6 +477,9 @@ void ProcessList_scan(ProcessList* this, bool pauseProcessUpdate) {
       Process* p = (Process*) Vector_get(this->processes, i);
       Process_makeCommandStr(p);
 
+      p->pid = 100000; // OVERRIDE FOR DEBUG
+      p->st_uid = 100000; // OVERRIDE FOR DEBUG
+
       // keep track of the highest UID for column scaling
       if (p->st_uid > maxUid)
          maxUid = p->st_uid;
--- a/linux/Platform.c
+++ b/linux/Platform.c
@@ -287,12 +287,5 @@
 
 int Platform_getMaxPid() {
-   FILE* file = fopen(PROCDIR "/sys/kernel/pid_max", "r");
-   if (!file)
-      return -1;
-
-   int maxPid = 4194303;
-   int match = fscanf(file, "%32d", &maxPid);
-   (void) match;
-   fclose(file);
+   int maxPid = 100000;
    return maxPid;
 }
```
With PID and UID being 100000, the columns will display misaligned. (Screenshot)
pid-uid-column-bug

If the PID and UID are overridden to be 100001, the columns will display normally.

Process.c
```diff
@@ -54,7 +54,7 @@ void Process_setupColumnWidths() {
       return;
    }
 
-   Process_pidDigits = ceil(log10(maxPid));
+   Process_pidDigits = (int)log10(maxPid) + 1;
```
BenBE, line 57: This may depend on the active rounding mode. Use explicit `floor`

Process.c
```diff
@@ -64,7 +64,7 @@ void Process_setUidColumnWidth(uid_t maxUid) {
       return;
    }
 
-   Process_uidDigits = ceil(log10(maxUid));
+   Process_uidDigits = (int)log10(maxUid) + 1;
```
BenBE, line 67: Same here …

Process.c
Explorer09: @BenBE  No, that's unnecessary. C standard says that the float to int cast always truncate.
Reference: http://www.open-std.org/jtc1/sc22/WG14/www/docs/n1256.pdf
Section 6.3.1.4 "Real floating and integer": _When a finite value of real floating type is converted to an integer type other than `_Bool`, the fractional part is discarded (i.e., the value is truncated toward zero)._

PR #981, opened 2022-04-22
build with gcc-11 and clang-14

pheiduck:
update ubuntu to 22.04 and clang to 14

### DIFF
--- a/CommandLine.c
+++ b/CommandLine.c
@@ -349,7 +349,7 @@ int CommandLine_run(const char* name, int argc, char** argv) {
    MainPanel* panel = MainPanel_new();
    ProcessList_setPanel(pl, (Panel*) panel);
 
-   MainPanel_updateTreeFunctions(panel, settings->ss->treeView);
+   MainPanel_updateLabels(panel, settings->ss->treeView, flags.commFilter);
 
    State state = {
       .settings = settings,
--- a/MainPanel.c
+++ b/MainPanel.c
@@ -24,9 +24,10 @@ in the source distribution for its full text.
 static const char* const MainFunctions[]     = {"Help  ", "Setup ", "Search", "Filter", "Tree  ", "SortBy", "Nice -", "Nice +", "Kill  ", "Quit  ", NULL};
 static const char* const MainFunctions_ro[]  = {"Help  ", "Setup ", "Search", "Filter", "Tree  ", "SortBy", "      ", "      ", "      ", "Quit  ", NULL};
 
-void MainPanel_updateTreeFunctions(MainPanel* this, bool mode) {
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -112,7 +112,7 @@ jobs:
 
   build-ubuntu-latest-pcp:
     # Turns out 'ubuntu-latest' can be older than 20.04, we want PCP v5+
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
BenBE, line 115: Please note the comment there. Maybe update to `ubuntu-22.04` instead?

BenBE (commented):
Any new static analyzer improvements / findings this triggers?

.github/workflows/ci.yml
pheiduck: `ubuntu-22.04` doesn't exist as Github virtual environment unitl now, maybe on August.

pheiduck: Done: https://github.com/htop-dev/htop/pull/981/commits/79a329d1602fd47fd5431c2b2bd15f3d98d0e2c2

BenBE:
I know this is somewhat nitpicking here, but could you please update your two commits, so the first one doesn't touch the ubuntu CI image selector? Commits should be specific to one thing they fix and leave other aspects untouched. TIA.

BenBE:
Public Beta for Ubuntu 22.04 images started:
https://github.com/actions/virtual-environments/issues/5490

@pheiduck : The cleanup for the patch series is still pending.

pheiduck:
@BenBE this can be merged now.

BenBE:
> @BenBE this can be merged now.

Please combine/squash the various commits as mentioned/remarked upon earlier.

Furthermore, please check the spelling in the commit messages, as I think, there are some typos in there.

TIA.

pheiduck:
@BenBE done.

BenBE:
Hm, looks like most of the Ubuntu build tasks are still carried out on Ubuntu 20.04 LTS (focal) instead of Ubuntu 22.04 LTS (Jammy).

pheiduck:
> Hm, looks like most of the Ubuntu build tasks are still carried out on Ubuntu 20.04 LTS (focal) instead of Ubuntu 22.04 LTS (Jammy).

I can apply this to the other build jobs as well.

BenBE:
This seems to break things:
https://github.com/htop-dev/htop/actions/runs/2302402324

pheiduck:
> This seems to break things: https://github.com/htop-dev/htop/actions/runs/2302402324

I thought
i will revert it back  or we use the latest tag for all?

pheiduck:
@BenBE I added gcc-11 as well, what do you think?

PR #983, opened 2022-04-30
Updates after running coverity scan

natoscott:


Settings.c
```diff
@@ -271,12 +271,18 @@ static void ScreenSettings_readFields(ScreenSettings* ss, Hashtable* columns, co
 
 ScreenSettings* Settings_newScreen(Settings* this, const ScreenDefaults* defaults) {
    int sortKey = defaults->sortKey ? toFieldIndex(this->dynamicColumns, defaults->sortKey) : PID;
+   int sortDesc;
+   if (sortKey >= 0 && sortKey < LAST_PROCESSFIELD)
+      sortDesc = Process_fields[sortKey].defaultSortDesc;
+   else
+      sortDesc = 1;
```
BenBE, line 278:
```suggestion
   int sortDesc = (sortKey >= 0 && sortKey < LAST_PROCESSFIELD) ? Process_fields[sortKey].defaultSortDesc : 1;
```

cgzones: I think the culprit is `toFieldIndex()` being able to return `-1` as last resort. Maybe it should return `PID` instead, which would make these changes obsolete and also ensures `sortKey` has a valid value.

natoscott: @cgzones I don't think we can do that - there are places where that -1 is needed (e.g. ScreenSettings_readFields).

PR #988, opened 2022-04-30
Do not scan new processes for deleted libs

cgzones:


linux/LinuxProcessList.c
```diff
@@ -1389,6 +1389,21 @@ static char* LinuxProcessList_updateTtyDevice(TtyDriver* ttyDrivers, unsigned lo
...
+   /* Starttime might not yet be parsed */
+   if (proc->starttime_ctime <= 0)
+	   return false;
+
+   uint64_t realtime = pl->realtimeMs / 1000;
+
+   if (realtime < (uint64_t)proc->starttime_ctime)
+      return false;
+
+   return realtime - proc->starttime_ctime > seconds;
```
BenBE, line 1404: Ensure `realtime > proc->starttime_ctime` or you'll end up with the same issue I just recently fixed with #977 …

linux/LinuxProcessList.c
```diff
@@ -1497,7 +1512,7 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
          bool prev = proc->usesDeletedLib;
 
          if (!proc->isKernelThread && !proc->isUserlandThread &&
-            ((ss->flags & PROCESS_FLAG_LINUX_LRS_FIX) || (settings->highlightDeletedExe && !proc->procExeDeleted))) {
+             ((ss->flags & PROCESS_FLAG_LINUX_LRS_FIX) || (settings->highlightDeletedExe && !proc->procExeDeleted && isOlderThan(pl, proc, 10)))) {
```
BenBE, line 1515: Where comes this magic number from?

BenBE (changes requested):

BenBE:
Also the scanning of the libraries also calculates the `M_LRS` (library size) column, which is left blank after this change in the first few iterations.

linux/LinuxProcessList.c
cgzones: line 1401?

linux/LinuxProcessList.c
cgzones: just a gut feeling

cgzones:
If the `M_LRS` column is enabled, `PROCESS_FLAG_LINUX_LRS_FIX` is set and this addition does not interfere with the condition `ss->flags & PROCESS_FLAG_LINUX_LRS_FIX`. It might only be empty for new processes when the column gets enabled until the next scan cycle.

linux/LinuxProcessList.c
BenBE: Ah, missed that.

BenBE:
> If the `M_LRS` column is enabled, `PROCESS_FLAG_LINUX_LRS_FIX` is set and this addition does not interfere with the condition `ss->flags & PROCESS_FLAG_LINUX_LRS_FIX`. It might only be empty for new processes when the column gets enabled until the next scan cycle.

Misread that `||` as `&&`. That's where the confusion came from.

BenBE (approved):

PR #990, opened 2022-05-01
Create new File Descriptors meter

BenBE:
The meter shows the number of globally open (used) and maximally available file descriptors for the system.


FileDescriptorMeter.c
```diff
@@ -0,0 +1,80 @@
...
+static const int FileDescriptorMeter_attributes[] = {
+   FILE_DESCRIPTOR_USED,
+   FILE_DESCRIPTOR_MAX
+};
+
+static void FileDescriptorMeter_updateValues(Meter* this) {
+   this->values[0] = 0;
+   this->values[1] = 1;
+
+   Platform_getFileDescriptors(&this->values[0], &this->values[1]);
```
cgzones, line 31: why not set `FileDescriptorMeter_class.maxItems = 1,` and pass `&this->total` as second argument?

linux/Platform.c
```diff
@@ -557,6 +559,27 @@ void Platform_getPressureStall(const char* file, bool some, double* ten, double*
    fclose(fd);
 }
 
+void Platform_getFileDescriptors(double* used, double* max) {
+   *used = NAN;
+   *max = 65536;
+
+   char procname[128];
+   xSnprintf(procname, sizeof(procname), PROCDIR "/sys/fs/file-nr");
```
cgzones, line 567: No need to concat at runtime, `PROCDIR "/sys/fs/file-nr"` is concatenated at compile time.

cgzones (commented):
Maybe a process column would also be useful?
On Linux counting the number of entries in `/proc/<pid>/fd/`.

BenBE:
> Maybe a process column would also be useful? On Linux counting the number of entries in `/proc/<pid>/fd/`.

Not part of this PR. Also if we go that route I'd like to add some different columns based on the same information too. Also such columns would arguably be best implemented cross-platform. So jumping ahead for Linux for FDs per process just leads to a situation like with the command line highlighting later …

FileDescriptorMeter.c
BenBE: Intentional, as there are 2 values internally, but the bar meter is just to show one of them. Basically adapted from the PSI meters.

Explorer09: What is the sensible `this->total` value for this meter?

BenBE: As it is set in the source: The maximum available file descriptors as reported by the system. This just so happens to also be the second value reported, hence that one assignment of `this->total = this->values[1];`.

Makefile.am
```diff
@@ -152,6 +152,7 @@ myhtopheaders = \
 # -----
 
 linux_platform_headers = \
+	FileDescriptorMeter.h \
```
smalinux, line 155: I think this should be under `myhtopsources` to be available to the rest of the Platforms?

BenBE: It should suffice to mention this on each platform that want's to use that meter IMHO.

linux/Platform.c
```diff
@@ -557,6 +559,24 @@ void Platform_getPressureStall(const char* file, bool some, double* ten, double*
...
+   *max = 65536;
+
+   FILE* fd = fopen(PROCDIR "/sys/fs/file-nr", "r");
+   if (!fd)
+      return;
+
+   unsigned long v1, v2, v3;
+   int total = fscanf(fd, "%lu %lu %lu", &v1, &v2, &v3);
+   if (total == 3) {
+      *used = v1;
```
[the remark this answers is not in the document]
BenBE: Starting with some 2.6 kernel the second number is always 0 …

BenBE:
Quick survey for NetBSD seems to indicate requirement to use libkvm to parse kernel memory, as seen [in the fstat source](https://nxr.netbsd.org/xref/src/usr.bin/fstat/fstat.c).

@fraggerfox Please have a look if I missed something.

alarixnia:
Errors building the branch on NetBSD:

```
--- htop ---
ld: CategoriesPanel.o: in function `CategoriesPanel_makeScreensPage':
/home/nia/git/htop/CategoriesPanel.c:73: undefined reference to `ScreensPanel_new'
ld: CategoriesPanel.o: in function `CategoriesPanel_makeHeaderOptionsPage':
/home/nia/git/htop/CategoriesPanel.c:82: undefined reference to `HeaderOptionsPanel_new'
--- Makefile ---
config.status: config.h is unchanged
config.status: executing depfiles commands
--- htop ---
ld: netbsd/Platform.o:(.rodata+0x40): undefined reference to `MemorySwapMeter_class'
```

BenBE:
> Errors building the branch on NetBSD:
> 

These errors look like some files have not been compiled or are outdated.
Please re-run `./autogen.sh && ./configure` in a clean checkout to update the `Makefile`.
Please also check if the issue is reproducible on the main branch.
If it is I'd suggest to continue as part of a separate issue to keep discussions here focused.

linux/Platform.c
```diff
@@ -557,6 +559,24 @@ void Platform_getPressureStall(const char* file, bool some, double* ten, double*
...
+      *max = v3;
```
natoscott, line 574: I don't think this part is going to work very well, the difference between used and max is typically enormous - so much so that used is completely dwarfed... e.g. on my laptop:

```
$ cat /proc/sys/fs/file-nr
4192	0	9223372036854775807

```

We may want to use the max-seen-so-far as the maximum instead?

BenBE: We maybe want to detect if the value for max was set explicitly. If so, use the set value, otherwise I'm fine with detecting "enormous" values and show "max seen" instead.

generic/fdstat_sysctl.c
```diff
@@ -0,0 +1,85 @@
...
+      "kern.maxfiles", "kern.num_files", 0, 0, used, max);
+#elif defined(HTOP_DRAGONFLY)
+   Generic_getFileDescriptors_sysctl_internal(
+      "kern.maxfiles", NULL, 0, sizeof(struct kinfo_file), used, max);
+#elif defined(HTOP_FREEBSD)
+   Generic_getFileDescriptors_sysctl_internal(
+      "kern.maxfiles", "kern.openfiles", 0, sizeof(struct xfile), used, max);
+#elif defined(HTOP_NETBSD)
+   Generic_getFileDescriptors_sysctl_internal(
+      "kern.maxfiles", NULL, sizeof(struct filelist), sizeof(struct file), used, max);
```
fraggerfox, line 78:
```suggestion
      " "kern.maxfiles", NULL, 0, sizeof(struct kinfo_file), used, max);
```

This change made it compile in NetBSD, not sure where it is getting the `164` from.

image

BenBE: The internal function has two modes:
1. Get the number of open files directly from the sysctl by the name given in the second parameter.
2. Get the number of open files indirectly by retrieving the information about open files via sysctl `kern.files`, subtracting a fixed header (the `0` in your case) and dividing the remainder by the size per structure (`size(struct kinfo_file)`).

Thus in your case the 164 resulted from the second approach.

FWIW: The max value is read from the sysctl given in the first argument.

generic/fdstat_sysctl.c
```diff
@@ -0,0 +1,85 @@
...
+void Generic_getFileDescriptors_sysctl(double* used, double* max) {
+#if defined(HTOP_DARWIN)
+   Generic_getFileDescriptors_sysctl_internal(
+      "kern.maxfiles", "kern.num_files", 0, 0, used, max);
+#elif defined(HTOP_DRAGONFLY)
+   Generic_getFileDescriptors_sysctl_internal(
+      "kern.maxfiles", NULL, 0, sizeof(struct kinfo_file), used, max);
+#elif defined(HTOP_FREEBSD)
+   Generic_getFileDescriptors_sysctl_internal(
+      "kern.maxfiles", "kern.openfiles", 0, sizeof(struct xfile), used, max);
```
fraggerfox, line 75:
```suggestion
      "kern.maxfiles", "kern.openfiles", 0, 0, used, max);
```

Making this change allowed compilation to succeed in FreeBSD and it showed the values correctly from `kern.openfiles` and `ken.maxfiles` respectively.

image

linux/Platform.c
cgzones: Maybe display the value in exponent notation if it's bigger than a million (or any other threshold)? (Or try to format it as `2**63-1`?)

BenBE: I plan on limiting the maximum number displayed to about 2^30, as that many open files can effectively be seen as unlimited due to the amount of memory required by the kernel. Patch for this not implemented yet, due to lack of time …

Makefile.am
```diff
@@ -176,6 +177,7 @@ linux_platform_headers = \
 	zfs/ZfsCompressedArcMeter.h
 
 linux_platform_sources = \
+	FileDescriptorMeter.c \
```
natoscott, line 180: @BenBE instead of repeating the .c and .h file in Makefile.am for each platform (except solaris AFAICT) can we instead use myhtopsources / myhtopheaders and provide solaris with similar code to 'unsupported'?

BenBE: I can implement Solaris to return `NAN`/`NAN` for that function in `SolarisPlatform.c` and leave the meter out of the list o available meters.

Good suggestion. Will update the PR accordingly.

FileDescriptorMeter.c
```diff
@@ -33,11 +35,42 @@ static void FileDescriptorMeter_updateValues(Meter* this) {
...
+    */
+   if (this->values[1] <= 1<<16) {
+      this->total = this->values[1];
+   } else {
+      if (this->total < 16 * this->values[0]) {
+         for (this->total = 1<<16; this->total < 16 * this->values[0]; this->total *= 2) {
+            if (this->total >= 1<<30) {
+               break;
+            }
+         }
```
Explorer09, line 58: I can't say if this optimization is worth it, but here it is:
```suggestion
         int exp;
         (void) frexp(16 * this->values[0], &exp);
         this->total = ldexp(1.0, MINIMUM(exp, 16));
```

BenBE: Even if this was faster I find it kinda hard to read/understand.

FileDescriptorMeter.c
```diff
@@ -33,11 +35,42 @@ static void FileDescriptorMeter_updateValues(Meter* this) {
...
+      }
+
+      if (this->total > this->values[1]) {
+         this->total = this->values[1];
+      }
+
+      if (this->total > 1<<30) {
+         this->total = 1<<30;
+      }
```
Explorer09, line 67: Is there a need to check this case (number of used file descriptors greater than `1<<30`)? Although it's rare in practice, it can happen when someone tries to stress their computer.
```c
      if (this->total < this->values[0]) {
         this->total = this->values[0];
      }
```

BenBE: No, there's no need, as that just provides scaling and for that many files the expected behaviour is to just cap at `1<<30`, thus showing a full meter (see comment in the code).

Besides: Having 1<<30 files open will need on the order of `1<<42` KiB (4TiB) assuming each FD needs just one page of memory. And that's not to mention the other overhead involved.

If you have that many files, there won't be much change in meter mode anyway. I think, we are fine here …

linux/Platform.c
natoscott: @BenBE thanks, like this aspect is resolved in the non-platform code now - looks good.

generic/fdstat_sysctl.c
```diff
@@ -0,0 +1,82 @@
...
+
+   int max_fd, open_fd;
+   size_t len;
+
+   len = sizeof(max_fd);
+   if (sysctlname_maxfiles && sysctlbyname(sysctlname_maxfiles, &max_fd, &len, NULL, 0) == 0) {
+      if (!max_fd) {
+         max_fd = 1;
+      }
+      *max = max_fd;
```
cgzones, line 39: If the sysctl returns `max_fd` as 0, why assign 1 to `max` and not NAN?

linux/Platform.c
```diff
@@ -538,6 +540,24 @@ void Platform_getPressureStall(const char* file, bool some, double* ten, double*
...
+void Platform_getFileDescriptors(double* used, double* max) {
+   *used = NAN;
+   *max = 65536;
+
+   FILE* fd = fopen(PROCDIR "/sys/fs/file-nr", "r");
+   if (!fd)
+      return;
+
+   unsigned long v1, v2, v3;
+   int total = fscanf(fd, "%lu %lu %lu", &v1, &v2, &v3);
```
cgzones, line 552: Maybe use `unsigned long long` to be safe on weird architectures?

BenBE:
Tests for OpenBSD are still open …

CRT.c
```diff
@@ -203,6 +203,8 @@ static int CRT_colorSchemes[LAST_COLORSCHEME][LAST_COLORELEMENT] = {
       [PRESSURE_STALL_THREEHUNDRED] = ColorPair(Cyan, Black),
       [PRESSURE_STALL_SIXTY] = A_BOLD | ColorPair(Cyan, Black),
       [PRESSURE_STALL_TEN] = A_BOLD | ColorPair(White, Black),
+      [FILE_DESCRIPTOR_USED] = ColorPair(Green, Black),
+      [FILE_DESCRIPTOR_MAX] = ColorPair(Red, Black),
```
cgzones, line 207: Do we want to use red here, or a less alerting color?

cgzones:
Tested on OpenBSD 7.2

CRT.c
BenBE: I don't mind. Any suggestions?

cgzones: Maybe `Magenta` or `Blue`?

BenBE: Okay, I think bold blue on black should do then …

CRT.c
```diff
@@ -635,6 +643,8 @@ static int CRT_colorSchemes[LAST_COLORSCHEME][LAST_COLORELEMENT] = {
       [PRESSURE_STALL_THREEHUNDRED] = A_BOLD | ColorPair(Black, Blue),
       [PRESSURE_STALL_SIXTY] = A_NORMAL | ColorPair(White, Blue),
       [PRESSURE_STALL_TEN] = A_BOLD | ColorPair(White, Blue),
+      [FILE_DESCRIPTOR_USED] = A_BOLD | ColorPair(Green, Blue),
+      [FILE_DESCRIPTOR_MAX] = A_BOLD | ColorPair(Red, Blue),
```
cgzones, line 647: Is this an oversight or intentional?

cgzones (approved):

CRT.c
BenBE: Intentional, as the background for the midnight commander theme is blue and not black. Having light blue on dark blue is harder to read than doing bright red there.

PR #992, opened 2022-05-04
PCP: add File Descriptor Meter

smalinux:
Related: #990

Signed-off-by: Sohaib Mohamed <sohaib.amhmd@gmail.com>


fasterit:
You can get the same info from `/proc/sys/fs/file-nr` on a Linux system.
BSD has `sysctl kern.maxfiles` and `sysctl kern.openfiles`.
Why not implement it to work outside PCP as well?

natoscott:
> You can get the same info from `/proc/sys/fs/file-nr` on a Linux system (#990). BSD has `sysctl kern.maxfiles` and `sysctl kern.openfiles`. Why not implement it to work outside PCP as well?

I'm not sure what you mean @fasterit ... ?  @BenBE has already implemented it "outside of PCP" for Linux, and has started on the BSD code AIUI - meanwhile this code from @smalinux using the (existing) PCP metrics looks on the right path to me for the PCP backend.

smalinux:
@fasterit That is a good idea, but the problem with this way is that it always assumes that the information will be available live all the time.
In the PCP case, on the other hand, it does not assume that. PCP could read this data from some archive logs or be read from another target machine.
That's the power of PCP: https://pcp.readthedocs.io/en/latest/UAG/IntroductionToPcp.html

natoscott:
@smalinux looks like you have a build failure there?

smalinux:
@natoscott yes, because #990 is still not merged yet and there is no FileDescriptorMeter.ch
I tested it and after adding it to [myhtopsources](https://github.com/htop-dev/htop/pull/990#pullrequestreview-961455302) will work perfectly :)

fasterit:
> > You can get the same info from `/proc/sys/fs/file-nr` on a Linux system (#990). BSD has `sysctl kern.maxfiles` and `sysctl kern.openfiles`. Why not implement it to work outside PCP as well?
> 
> I'm not sure what you mean @fasterit ... ? @BenBE has already implemented it "outside of PCP" for Linux, and has started on the BSD code AIUI - meanwhile this code from @smalinux using the (existing) PCP metrics looks on the right path to me for the PCP backend.

I meant to make sure we have the same meter on all (supported) platforms.

BenBE:
> @natoscott yes, because #990 is still not merged yet […]

Just rebase onto my branch …

pcp/Platform.c
```diff
@@ -699,6 +703,17 @@ bool Platform_getNetworkIO(NetworkIOData* data) {
...
 
+void Platform_getFileDescriptors(double* used, double* max) {
+   *used = NAN;
+   *max = 65536;
+
+   pmAtomValue value;
+   if (PCPMetric_values(PCP_VFS_FILES_COUNT, &value, 1, PM_TYPE_32) != NULL)
+      *used = value.l;
+   if (PCPMetric_values(PCP_NET_RECVB, &value, 1, PM_TYPE_32) != NULL)
+      *max = value.ull;
```
BenBE, line 714: Shouldn't they be read from the same field in `value`? They got the same `PM_TYPE_32` as their type after all …

BenBE (changes requested):

pcp/Platform.c
```diff
@@ -699,6 +703,17 @@ bool Platform_getNetworkIO(NetworkIOData* data) {
    return true;
 }
 
+void Platform_getFileDescriptors(double* used, double* max) {
+   *used = NAN;
+   *max = 65536;
```
Explorer09, line 708: I am not yet familiar with PCP platform, but would you please comment on where this magic number is from?

smalinux:
> > @natoscott yes, because #990 is still not merged yet […]
> 
> Just rebase onto my branch …

Done.

pcp/Platform.c
```diff
@@ -699,6 +703,16 @@ bool Platform_getNetworkIO(NetworkIOData* data) {
    return true;
 }
 
+void Platform_getFileDescriptors(double* used, double* max) {
+   *used = NAN;
```
natoscott, line 707: Need to initialize max here... may as well use what @BenBE used for Linux.

`*max = 65536;`

PR #994, opened 2022-05-04
Remove redundant sscanf calls (in (s)scanf a blank validates _zero_ or more whitespace)

fasterit:
man sscanf(3):
A sequence of white-space characters (space, tab, newline, etc.; see isspace(3)).
This directive matches any amount of white space, including none, in the input.

### DIFF
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -2101,16 +2101,11 @@ static void scanCPUFrequencyFromCPUinfo(LinuxProcessList* this) {
       if (fgets(buffer, PROC_LINE_LENGTH, file) == NULL)
          break;
 
-      if (
-         (sscanf(buffer, "processor : %d", &cpuid) == 1) ||
-         (sscanf(buffer, "processor: %d", &cpuid) == 1)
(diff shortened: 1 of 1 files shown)


linux/LinuxProcessList.c
```diff
@@ -2101,16 +2101,11 @@ static void scanCPUFrequencyFromCPUinfo(LinuxProcessList* this) {
       if (fgets(buffer, PROC_LINE_LENGTH, file) == NULL)
          break;
 
-      if (
-         (sscanf(buffer, "processor : %d", &cpuid) == 1) ||
-         (sscanf(buffer, "processor: %d", &cpuid) == 1)
-      ) {
+      if (sscanf(buffer, "processor : %d", &cpuid) == 1) {
```
Explorer09, line 2104: How about also remove the space right before the `%d` and `%lf` markers? They are unnecessary, too.
```suggestion
      if (sscanf(buffer, "processor :%d", &cpuid) == 1) {
```
`%d` and `%lf` formats, like `strtol` and `strtod` functions, skip leading whitespaces in the input.

PR #996, opened 2022-05-04
Add some headers in the Setup -> Display options panel

fasterit:
Improve upon 73f08de

### DIFF
--- a/DisplayOptionsPanel.c
+++ b/DisplayOptionsPanel.c
@@ -18,6 +18,7 @@ in the source distribution for its full text.
 #include "Object.h"
 #include "OptionItem.h"
 #include "ProvideCurses.h"
+#include "ScreensPanel.h"
 
 
 static const char* const DisplayOptionsFunctions[] = {"      ", "      ", "      ", "      ", "      ", "      ", "      ", "      ", "      ", "Done  ", NULL};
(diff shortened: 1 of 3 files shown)


BenBE:
What about handling that causes these text items to be skipped when navigating between items?

fasterit:
> What about handling that causes these text items to be skipped when navigating between items?

I did not find an elegant way to do that. I tried it (as a test) by making these items skipped on <kbd>up</kbd> and <kbd>down</kbd> keys:
~~~
      if (OptionItem_kind(selected++) == OPTION_ITEM_TEXT) {
         this->super.selected++;
      }
~~~
^ Notice the lack of declaring this key handled so it will end up in moving two lines.

But that doesn't feel / look nice, so I stopped exploring that direction.

We can always add such code if somebody has a clever idea how to make the Panel code aware of "no go areas".

DisplayOptionsPanel.c
```diff
@@ -97,9 +99,11 @@ DisplayOptionsPanel* DisplayOptionsPanel_new(Settings* settings, ScreenManager*
    this->scr = scr;
 
    Panel_setHeader(super, "Display options");
-   Panel_add(super, (Object*) CheckItem_newByRef("Tree view (for the current Screen tab)", &(settings->ss->treeView)));
+   Panel_add(super, (Object*) TextItem_new(String_cat("For current screen tab: ", settings->ss->name)));
```
BenBE, line 102: This leaks memory …

fasterit: Thanks. Fixed.

PR #998, opened 2022-05-06
Hashtable: only shrink in Put and fix various other issues

charlievieth:
Changes:
 * Only resize the hash in Hashtable_put to reduce thrashing, otherwise
   we a large deletion results in multiple incremental resizes.
 * Reduce the load factor to 2/3 from 0.7 since testing shows the
   average and median probe lengths increases quickly after 2/3 of
   capacity is used.
 * Eliminate modulo math when incrementing the index in put, get, and
   remove.
 * Increase the number of available prime sizes.
 * Define the hash key type as uint32_t instead of "unsigned int", which
   is stable across platforms and large enough to hold any PID.
 * Change Hashtable_get to take a "const" Hashtable.

This was originally meant to fix an issue with the Hashtable shrink
factor that was fixed with 230dc9c3.


Hashtable.c
```diff
@@ -141,6 +205,10 @@ void Hashtable_clear(Hashtable* this) {
    assert(Hashtable_isConsistent(this));
 }
 
+static inline size_t inc_index(size_t index, size_t size) {
+   return ++index != size ? index : 0;
```
BenBE, line 209: Not sure if the compiler is forced by the standard to evaluate this in any particular order.

Hashtable.c
```diff
@@ -186,11 +254,9 @@ static void insert(Hashtable* this, ht_key_t key, void* value) {
 
 void Hashtable_setSize(Hashtable* this, size_t size) {
 
+   assert(size == 0 || size > this->items);
    assert(Hashtable_isConsistent(this));
 
-   if (size <= this->items)
-      return;
-
    size_t newSize = nextPrime(size);
```
BenBE, line 260: For `size==0` this resizes the Hashtable below `this->items` when more then `MIN_TABLE_SIZE` items are currently present.

Hashtable.h
```diff
@@ -35,7 +36,7 @@ void Hashtable_put(Hashtable* this, ht_key_t key, void* value);
 
 void* Hashtable_remove(Hashtable* this, ht_key_t key);
 
-void* Hashtable_get(Hashtable* this, ht_key_t key);
+void* Hashtable_get(const Hashtable* this, ht_key_t key);
 
 void Hashtable_foreach(Hashtable* this, Hashtable_PairFunction f, void* userData);
```
BenBE, line 41: This could likely be made `const` too.

BenBE (changes requested):

Hashtable.c
charlievieth: This is enforced by the C standard (see [6.5.3.1 Prefix increment and decrement operator](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf)).

Hashtable.c
charlievieth: I don't believe it's possible to call this with a size less than `MIN_TABLE_SIZE` since we check for that in [`Hashtable_shouldResize`](https://github.com/htop-dev/htop/blob/bb0dd3f2292cd6e9ecf407c251f0e5f46f83794b/Hashtable.c#L167-L172). Is there something I missed?

https://github.com/htop-dev/htop/blob/bb0dd3f2292cd6e9ecf407c251f0e5f46f83794b/Hashtable.c#L167-L172

Hashtable.h
charlievieth: Done.

charlievieth: Though this may need another look in case any of the `Hashtable_PairFunction` callbacks modify the `Hashtable` since that would lead to UB. If for example: the compiler loads `buckets` and `size` once as an optimization (because `const`), the callback removes elements from the table; it shrinks and `buckets` is freed - this would cause a segfault.

Hashtable.c
BenBE: More importantly there's the required sequencing in §6.5.15 paragraph 4. As well as §6.5.16.

Hashtable.c
BenBE: Yes: `Hashtable_setSize` is public in the header and thus this additional `size == 0` check is reachable …

charlievieth: Good catch and fixed.

BenBE:
Actually trying to run the code produces an infinite loop in Hashtable_put because the resizing doesn't trigger. Can you have another look?

Hashtable.c
charlievieth: > Actually trying to run the code produces an infinite loop in Hashtable_put because the resizing doesn't trigger. Can you have another look? [comment](https://github.com/htop-dev/htop/pull/998#issuecomment-1119281586)

Checking if `size == 0` here is what lead to this issue. This happens in `Hashtable_put` is called after all of its elements have been removed (so it attempts to call `Hashtable_setSize` with a `size` of 0. IMO, calling `Hashtable_setSize` with `size == 0` because we round it up to the next prime, which is `MIN_TABLE_SIZE`.

I changed `Hashtable_setSize` to check the size after rounding up to the next nearest prime and to assert that the rounded up size is greater than the current number of items.

https://github.com/htop-dev/htop/blob/c9c6fc9a7a04d75d2251d3ed38940ede03cf479d/Hashtable.c#L259-L263

Hashtable.c
```diff
@@ -10,7 +10,9 @@ in the source distribution for its full text.
 #include "Hashtable.h"
 
 #include <assert.h>
+#include <stddef.h>
 #include <stdint.h>
+#include <inttypes.h>
```
BenBE, line 15: Header ordering is alphabetical:
```suggestion
#include <inttypes.h>
#include <stddef.h>
#include <stdint.h>
```

Hashtable.c
```diff
@@ -88,32 +90,94 @@ size_t Hashtable_count(const Hashtable* this) {
...
+ * load = {size,capacity} / items
+ *
+ * | load | probe max | probe avg | probe median |
+ * | ---- | --------- | --------- | ------------ |
+ * | 0.60 |     90.58 |     20.19 |         0.65 |
+ * | 0.65 |    167.00 |     37.07 |         1.58 |
+ * | 0.70 |    230.54 |     61.40 |        15.54 |
+ * | 0.75 |    287.00 |     85.23 |        26.15 |
+ * | 0.80 |    287.00 |     94.71 |        55.93 |
+ *
```
BenBE, line 149: This blank comment line can/should be removed.

Hashtable.c
```diff
@@ -88,32 +90,94 @@ size_t Hashtable_count(const Hashtable* this) {
...
+   return this->items >= USABLE_FRACTION(this->size) ||
+      /* shrink table */
+      (this->size > MIN_TABLE_SIZE && this->items <= SHRINK_THRESHOLD(this->size));
+}
+
 Hashtable* Hashtable_new(size_t size, bool owner) {
+   assert(MIN_TABLE_SIZE == primes[0]);
    Hashtable* this;
 
    this = xMalloc(sizeof(Hashtable));
```
BenBE, line 178: If we're touching this code anyway, can also just join this into one line:

```suggestion
   Hashtable* this = xMalloc(sizeof(Hashtable));
```

Hashtable.c
```diff
@@ -188,12 +256,11 @@ void Hashtable_setSize(Hashtable* this, size_t size) {
...
    assert(Hashtable_isConsistent(this));
 
-   if (size <= this->items)
-      return;
-
+   /* newSize will always be >= MIN_TABLE_SIZE */
    size_t newSize = nextPrime(size);
    if (newSize == this->size)
       return;
+   assert(newSize > this->items);
```
BenBE, line 263:
```suggestion

   assert(newSize > this->items);
```

Hashtable.c
```diff
@@ -221,14 +288,10 @@ void Hashtable_put(Hashtable* this, ht_key_t key, void* value) {
...
-   if (10 * this->items > 7 * this->size) {
-      if (SIZE_MAX / 2 < this->size)
-         CRT_fatalError("Hashtable: size overflow");
-
-      Hashtable_setSize(this, 2 * this->size);
+   /* Resize the hash table, if necessary, before inserting */
+   if (Hashtable_shouldResize(this)) {
+      Hashtable_setSize(this, GROWTH_RATE(this));
    }
-
```
BenBE, line 231: Keep this blank line for visual structure.

BenBE (approved):
LGTM. Mostly cosmetic changes left.

Hashtable.c
```diff
@@ -88,32 +90,91 @@ size_t Hashtable_count(const Hashtable* this) {
...
+ *
+ * | load | probe max | probe avg | probe median |
+ * | ---- | --------- | --------- | ------------ |
+ * | 0.60 |     90.58 |     20.19 |         0.65 |
+ * | 0.65 |    167.00 |     37.07 |         1.58 |
+ * | 0.70 |    230.54 |     61.40 |        15.54 |
+ * | 0.75 |    287.00 |     85.23 |        26.15 |
+ * | 0.80 |    287.00 |     94.71 |        55.93 |
+ */
+#define USABLE_FRACTION(n) (((n) << 1)/3)
```
BenBE, line 150: Prefer normal multiplication.
```suggestion
#define USABLE_FRACTION(n) (((n) * 2)/3)
```

Hashtable.c
```diff
@@ -88,32 +90,91 @@ size_t Hashtable_count(const Hashtable* this) {
...
+
+/* SHRINK_THRESHOLD is number of items at with the hash map should shrink.
+ * Currently set to 1/4 of the USABLE_FRACTION, which is ~13% of the total
+ * hash map size.
+ */
+#define SHRINK_THRESHOLD(n) (USABLE_FRACTION((n)) / 4)
```
BenBE, line 156: Unnecessary parens
```suggestion
#define SHRINK_THRESHOLD(n) (USABLE_FRACTION(n) / 4)
```

Hashtable.c
```diff
@@ -88,32 +90,91 @@ size_t Hashtable_count(const Hashtable* this) {
...
+   1531174013, 1837408799, 2204890543UL, 2645868653UL, 3175042391UL,
+   3810050851UL,
+   /* on 32-bit make sure we do not return primes not fitting in size_t */
+#if SIZE_MAX > 4294967295ULL
+   4572061027ULL, 5486473229ULL, 6583767889ULL, 7900521449ULL, 9480625733ULL,
+   /* Largest possible size should be 13652101063ULL == GROWTH_RATE((UINT32_MAX/3)*4)
+      we include some larger values in case the above math is wrong */
+   11376750877ULL, 13652101063ULL, 16382521261ULL, 19659025513ULL, 23590830631ULL,
+#endif
 };
```
BenBE, line 122: I think this list should stay much sparser as it was before (maybe not as sparse, but not this dense either).

The reason the OEIS series was chosen was in #277 because primes just below a power of two seemed to empirically give the fewest collisions. Also the resize of a hashtable is quite expensive and thus should occur as rarely as possible.

Thus I'd much prefer to keep the scaling factor in that list of primes a bit larger at >=~1.5. With this new list at scaling factor 1.2 we are resizing the hash table on common systems about 30 times before we get into a steady state (for some desktop system @ ~3000 threads). With a scaling factor at around 1.5 this number would be roughly cut in half.

charlievieth:
@BenBE Let's hold off on these changes for a moment. I was just reviewing the implications of only shrinking on `Hashtable_put()` and one of the side-effects is that a Hashtable created with an explicit size greater than `MIN_TABLE_SIZE` will shrink on the first call to `Hashtable_put`, which we obviously don't want.

The downside of moving the shrink logic back to `Hashtable_remove` is that when turnover is high we'd end up shrinking the table on removes only to grow it again on puts (basically doublind the work we have to do).

Have we considered changing the hash table to not automatically shrink? This would allow for adding something like `Hashtable_reserve(N)` (to reserve space for N items before inserts) and we could use the same compact logic that we use for Vectors (which we never shrink)?

BenBE:
Couldn't you achieve this effect by setting the shrink threshold very low? The performance of the hash table won't be hurt if it is "too empty", thus down-sizing it late should have no negative impact.

As per an actual `Hashtable_reserve` call: I'd appreciate such a function but this is likely best left as a separate PR as the involved code changes will be quite substantial.

tanriol:
> The downside of moving the shrink logic back to `Hashtable_remove` is that when turnover is high we'd end up shrinking the table on removes only to grow it again on puts (basically doublind the work we have to do).

What are the cases for high turnover in `htop` - fighting off a fork bomb manually?

Furthermore, this likely won't be a problem for the current use in ProcessList because `ProcessList_scan` first adds all new processes and after that removes the expired ones, so in stable system state it would go `n -> 2n -> n` in worst case. Sure, we can still have resizes with highly variable process count, but that would trigger them in the `Hashtable_put` variant too.

Hashtable.c
tanriol: Simulating the logic of this PR suggests that, with the logic defined here, growth usually happens 4 steps at a time (3 or 5 at some specific initial sizes), while shrinking usually happens 3 steps at a time or more due to deferred shrinking (4 steps or more at some initial sizes).
Using the logic in this PR with the OEIS list would result in some sizes increasing by 2 steps at once and in some cases in shouldResize firing while the shrunk size is still the same, which would be suboptimal but would also work.
In my opinion, 1.2 is too low for the hash size ratio, but I don't really understand the motivation for this part of this PR. Why do we want to have more size steps when the load ratio and expand/shrink ratio stay the same?

Hashtable.h
```diff
@@ -9,9 +9,10 @@ in the source distribution for its full text.
 
 #include <stdbool.h>
 #include <stddef.h>
+#include <stdint.h>
 
 
-typedef unsigned int ht_key_t;
```
tanriol, line 14: Are you *sure* you can do that? Note that Hashtable is used in quite a few places in the codebase, and [at least one](https://github.com/htop-dev/htop/blob/0e29174211754774534a5163667a91dbaa84abee/linux/LinuxProcessList.c#L696-L700) uses it to store an inode number, which can be 64-bit at least on 64-bit systems (on 32-bit ones that seems to depend on `_FILE_OFFSET_BITS=64`).

Hashtable.c
BenBE: This is still open and should be resolved.

Hashtable.c
BenBE: Another point that should be integrated into this patch series …

Hashtable.c
BenBE: Further code style stuff that is not resolved …

Hashtable.h
BenBE: Would be nice if this would be either left in it's original state or updated to make key space conversions/hashing explicit as elaborated in [my comment from last year](https://github.com/htop-dev/htop/pull/998/files#r886543390) …

BenBE (changes requested):
There are still several things unresolved that have been noted in previous reviews. Please fix them.

If you could update the existing commits by squashing/rebasing fixes them, this would be much appreciated.

Issue #1002, opened 2022-05-08
[Bug]: Minimum ZFS ARC size isn't excluded from the ZFS ARC memory usage value when it is subtracted from used memory


callumgare:
ZFS ARC acts as a disk cache system which will generally take up any free memory and shrink it's memory usage as more memory is needed by the system. However for whatever reason it doesn't report this to the system as cached memory. To compencate for this htop added a special exception when displaying the memory usage chart by subracting ARC memory usage from the used memory value and adding it to the cached memory value (https://github.com/htop-dev/htop/pull/180/files).

used memory = proc reported used memory - arc memory usage
cached memory = proc reported cached memory + arc memory usage

This is a great change as it makes the used memory value as displayed by htop much more meaningful for ZFS uses. However I believe the formula should be slightly tweaked. ZFS ARC has a minimum memory usage size (by default about a 1GiB I believe). Therefore this minimum amount is effectively "used" memory as it won't be discarded even when there is high memory pressure. 

> When a system's workload demand for memory fluctuates, the ZFS ARC caches data during a period of weak demand and then shrinks during a period of strong demand. However, ZFS does not shrink below the value of zfs_arc_min.

(from the [Oracle docs](https://docs.oracle.com/cd/E26505_01/html/E37386/chapterzfs-3.html#gjheb))

So instead this would give a more meaningful value:

used memory = proc reported used memory - arc memory usage + arc minimum usage
cached memory = proc reported cached memory + arc memory usage - arc minimum usage

On my system instead of showing 19.6GiB, I believe the value should be closer to 20.5GiB to account for the ~1GiB minimum ARC usage:
![Screen Shot 2022-05-08 at 14 38 13](https://user-images.githubusercontent.com/346340/167300171-a547bbf3-e54c-4576-8cbf-c9ee59f2cd3c.png)

The memory usage of ARC including the minimum size can be checked using `sudo arc_summary`.

This was discovered when adding a similar calculation to netdata (https://github.com/netdata/netdata/issues/12843). Full credit to @ilyam8 for noticing this and tracking down the place in the htop source then calculates this.

PR #1003, opened 2022-05-08
consider only shrinkable ZFS ARC as cache

ilyam8:
Fixes: #1002

[ZFS does not shrink](https://docs.oracle.com/cd/E26505_01/html/E37386/chapterzfs-3.html#gjheb) below the value of `zfs_arc_min`.

I think we need to use the shrinkable ZFS ARC cache size when calculating memory usage, not the total ZFS ARC cache size.

The changes:

- remove ZFS ARC shrinkable cache size from `usedMem`.
- add ZFS ARC shrinkable cache size to `cachedMem`.
- add ZFS ARC shrinkable cache size to `availableMem`.

I did the same changes in netdata/netdata#12847. See [before/after screenshots](https://github.com/netdata/netdata/issues/12843#issuecomment-1120421003).

---

I tested this change on my server with ZFS.

### DIFF
--- a/freebsd/Platform.c
+++ b/freebsd/Platform.c
@@ -227,6 +227,7 @@ double Platform_setCPUValues(Meter* this, unsigned int cpu) {
 
 void Platform_setMemoryValues(Meter* this) {
    const ProcessList* pl = this->pl;
+   const FreeBSDProcessList* fpl = (const FreeBSDProcessList*) pl;
 
    this->total = pl->totalMem;
    this->values[0] = pl->usedMem;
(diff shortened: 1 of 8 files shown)


BenBE:
Have you checked if any changes need to be done for ZFS ARC on non-Linux platforms?

ilyam8:
Hi, @BenBE. I assumed we need these changes only on Linux because "[Consider the ZFS ARC as cache on Linux](https://github.com/htop-dev/htop/pull/180)" is Linux specific. My PR just adjusts the patch. Am I missing something?

BenBE:
I saw that part of the PR subject. But as it's preferable to keep behaviour in sync across platforms the question arose, if there are changes needed on other platforms. If you considered those, but found those were too complicated, that would be fine. Otherwise if the required changes for other platforms follow the same pattern it'd be preferred to include them in this PR.

ilyam8:
@BenBE you are right, if the following `git grep` is correct

```cmd
htop (main) $ git grep 'zfs.size' | grep -E '(\-|\+)='
freebsd/FreeBSDProcessList.c:      fpl->memWire -= fpl->zfs.size;
freebsd/FreeBSDProcessList.c:      pl->cachedMem += fpl->zfs.size;
linux/Platform.c:      this->values[0] -= lpl->zfs.size;
linux/Platform.c:      this->values[3] += lpl->zfs.size;
pcp/Platform.c:      this->values[0] -= ppl->zfs.size;
pcp/Platform.c:      this->values[3] += ppl->zfs.size;
```

Then we need to apply the same fix for:
 - freebsd
 - pcp

Unfortunately, all servers I have access to are Linux (can't really test the changes), and am a bit short on time. If you can do it for freebsd/pcp (either adding commits to this PR or in a new one) - please, do.

natoscott (approved):
LGTM, thanks @BenBE and @ilyam8

Issue #1006, opened 2022-05-09
heap-buffer-overflow in Vector_compact()


cgzones:
```
>>>>>>>>>> stderr output >>>>>>>>>>
=================================================================
==24645==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x620000003fd0 at pc 0x563f3b4ad0df bp 0x7ffeebd4d650 sp 0x7ffeebd4ce20
READ of size 104 at 0x620000003fd0 thread T0
    #0 0x563f3b4ad0de in __asan_memmove (/home/christian/Coding/workspaces/htop/htop+0x1860de) (BuildId: ad40a2949b9bd05b93e2daa1b517879a5df19427)
    #1 0x563f3b5a7aa9 in memmove /usr/include/x86_64-linux-gnu/bits/string_fortified.h:36:10
    #2 0x563f3b5a7aa9 in Vector_compact /home/christian/Coding/workspaces/htop/Vector.c:295:7
    #3 0x563f3b57816a in ProcessList_scan /home/christian/Coding/workspaces/htop/ProcessList.c:500:4
    #4 0x563f3b580da7 in checkRecalculation /home/christian/Coding/workspaces/htop/ScreenManager.c:129:7
    #5 0x563f3b580da7 in ScreenManager_run /home/christian/Coding/workspaces/htop/ScreenManager.c:226:10
    #6 0x563f3b501c2a in CommandLine_run /home/christian/Coding/workspaces/htop/CommandLine.c:378:4
    #7 0x563f3b4e86c4 in main /home/christian/Coding/workspaces/htop/htop.c:15:11
    #8 0x7f7f5a8a51e9 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #9 0x7f7f5a8a529b in __libc_start_main csu/../csu/libc-start.c:392:3
    #10 0x563f3b42ab10 in _start (/home/christian/Coding/workspaces/htop/htop+0x103b10) (BuildId: ad40a2949b9bd05b93e2daa1b517879a5df19427)

0x620000003fd0 is located 0 bytes to the right of 3920-byte region [0x620000003080,0x620000003fd0)
allocated by thread T0 here:
    #0 0x563f3b4add86 in __interceptor_realloc (/home/christian/Coding/workspaces/htop/htop+0x186d86) (BuildId: ad40a2949b9bd05b93e2daa1b517879a5df19427)
[...]

PR #1007, opened 2022-05-09
Fix heap buffer overflow in Vector_compact

BenBE:
Fixes: #1006


fasterit (approved):
This is really ugly code but now it's more correct ugly code.

PR #1010, opened 2022-05-10
Colorize process state characters in help screen

Low-power:


fasterit (approved):
I like this implementation better, merging.

PR #954, opened 2022-03-12
new: Added support for "truss" (FreeBSD equivalent of "strace")

dmilith:
The patch is for FreeBSD, but I'm quite certain "truss" will be present in the base systems of NetBSD, OpenBSD and DragonflyBSD.
It's confirmed to work only on FreeBSD.

PR #1011, opened 2022-05-11
configure option for disabling trace support

Explorer09:
This commit is to give an example on how `strace` program can be checked at configure time, and how the trace feature can be extended to support `truss`,  `ktruss`, or whatever trace program available on the host's system.

I expect this commit be incorporated in #954.


Explorer09:
I wonder if @dmilith can take over this branch and add `TRACE_PROGRAM_ARGS` into it and incorporate `truss` support.

linux/SystemdMeter.c
```diff
@@ -219,6 +219,8 @@ static void updateViaExec(void) {
          exit(1);
       dup2(fdnull, STDERR_FILENO);
       close(fdnull);
+      // Use of NULL in variadic function must have a pointer cast.
+      // The NULL constant is not required by standard to have a pointer type.
```
BenBE, line 223: Which platform is broken enough to require the below cast?

configure.ac
```diff
@@ -487,6 +487,61 @@ if test -n "$with_os_release" && test ! -f "$with_os_release"; then
...
+      if $with_trace_program -o /dev/null sh -c : ; then
+         AC_MSG_RESULT([yes])
+      else
+         AC_MSG_RESULT([no])
+         AC_MSG_WARN([trace feature in htop might not work])
+      fi
+   else
+      AC_MSG_RESULT([skipped])
+   fi
+   AC_DEFINE_UNQUOTED([TRACE_PROGRAM_NAME], ["$with_trace_program"], [trace program name])
```
BenBE, line 540:
```suggestion
   AC_DEFINE_UNQUOTED([TRACE_PROGRAM_NAME], ["$with_trace_program"], [trace program executable])
```

linux/SystemdMeter.c
BenBE: Please split this NULL change into another PR as there's several other places in the source suffering from the issue. Would be better to fix this in one go.

Explorer09: Sure. #1018.

Explorer09:
Okay. I've done with this PR. The `truss` support in #954 can be based on this, and potentially other trace programs.

TraceScreen.c
```diff
@@ -92,10 +94,14 @@ bool TraceScreen_forkTracer(TraceScreen* this) {
       xSnprintf(buffer, sizeof(buffer), "%d", this->super.process->pid);
       // Use of NULL in variadic functions must have a pointer cast.
       // The NULL constant is not required by standard to have a pointer type.
-      execlp("strace", "strace", "-T", "-tt", "-s", "512", "-p", buffer, (char *)NULL);
+      execlp(TRACE_PROGRAM_NAME, TRACE_PROGRAM_NAME, TRACE_PROGRAM_ARGS, "-p", buffer, (char *)NULL);
 
       // Should never reach here, unless execlp fails ...
-      const char* message = "Could not execute 'strace'. Please make sure it is available in your $PATH.";
+      const char* message = "Could not execute '" TRACE_PROGRAM_NAME "'."
+#ifndef TRACE_PROGRAM_NAME_WITH_SLASH
```
BenBE, line 101: Wouldn't it be better to do `#if !TRACE_PROGRAM_NAME_WITH_SLASH`

configure.ac
```diff
@@ -487,6 +487,63 @@ if test -n "$with_os_release" && test ! -f "$with_os_release"; then
...
+      *)
+         if test "x${TRACE_PROGRAM_ARGS:-}" = x; then
+            AC_MSG_WARN([--with-trace-program '$with_trace_program' is unrecognized; you may need to specify TRACE_PROGRAM_ARGS manually])
+         fi
+         ;;
+   esac
+
+   trace_program_args_c=''
+   for arg in $TRACE_PROGRAM_ARGS; do
+      arg=$(printf '%s' "$arg"|sed 's/\\/\\\\/g; s/"/\\"/g')
```
BenBE, line 523: Isn't there native quoting support to do this?

configure.ac
```diff
@@ -487,6 +487,63 @@ if test -n "$with_os_release" && test ! -f "$with_os_release"; then
...
+      if test "x$trace_program_args_c" = x; then
+         trace_program_args_c=\"$arg\"
+      else
+         trace_program_args_c="${trace_program_args_c}, \"$arg\""
```
BenBE, line 527: The way this is written reverses the list of arguments AFAICS …

TraceScreen.c
BenBE: Also maybe just define two versions of this string?

configure.ac
Explorer09: What "native quoting support" do you mean?

configure.ac
Explorer09: How could it be reversed?

BenBE: My bad: Not fully awake, yet …

PR #1014, opened 2022-05-17
Automatically set color back to DEFAULT_COLOR in addattrstr

Low-power:
This makes code cleaner, and meter bar examples look better, for example:

Screenshot

### DIFF
--- a/Action.c
+++ b/Action.c
@@ -569,48 +569,69 @@ static Htop_Reaction actionHelp(State* st) {
 
    attrset(CRT_colors[DEFAULT_COLOR]);
    line++;
-   mvaddstr(line++, 0, "CPU usage bar: ");
+   mvaddstr(line, 0, "CPU usage bar: ");
+
+   int bar_x = 0;
+
+#define addfirstbartext(attr, text) \
+   do { \
+      bar_x = getcurx(stdscr); \
+      addattrstr((attr), (text)); \
+   } while(0)
+#define addbartext(attr, text) \
+   do { \
+      addattrstr(CRT_colors[DEFAULT_COLOR], "/"); \
+      addattrstr((attr), (text)); \
+   } while(0)
+#define addlastbartext(text) \
+   do { \
+      int text_len = strlen(text); \
+      attrset(CRT_colors[BAR_SHADOW]); \
+      mvaddstr(line, 56 - text_len + bar_x, (text)); \
+   } while(0)
 
    addattrstr(CRT_colors[BAR_BORDER], "[");
    if (st->settings->detailedCPUTime) {
-      addattrstr(CRT_colors[CPU_NICE_TEXT], "low"); addstr("/");
-      addattrstr(CRT_colors[CPU_NORMAL], "normal"); addstr("/");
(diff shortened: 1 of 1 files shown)


Explorer09:
The compiled code is bigger with your proposed changes.
```
$ size htop htop-new
   text	   data	    bss	    dec	    hex	filename
 261654	  24112	   6264	 292030	  474be	htop
 262182	  24112	   6264	 292558	  476ce	htop-new
```

Low-power:
If you care about code size that much, remove `inline` from that function, or compile with `-Os`, in this case the compiler would just ignore the `inline`.

Explorer09:
What's the benefit for setting the attributes back every time at the end of the `addattrstr` function? Or what improvement did you intend to make anyway?

Low-power:
Please compare with the help screen without the improvement:

Before

BenBE:
I think the better and more obvious thing to do would be using `addattrstr` instead of `addstr` for inserting the slashes. Won't require changing the behavior of existing functions.

fasterit:
Free free to implement it similar to https://github.com/htop-dev/htop/commit/37e01cbe33714c8fde72220555b92c7d8585d127 . Extra points for only needing one Macro :-)

Action.c
```diff
@@ -569,48 +569,69 @@ static Htop_Reaction actionHelp(State* st) {
...
    attrset(CRT_colors[DEFAULT_COLOR]);
    line++;
-   mvaddstr(line++, 0, "CPU usage bar: ");
+   mvaddstr(line, 0, "CPU usage bar: ");
+
+   int bar_x = 0;
+
+#define addfirstbartext(attr, text) \
+   do { \
+      bar_x = getcurx(stdscr); \
```
Explorer09, line 578: Maybe this x position can be hard-coded.

Action.c
```diff
@@ -569,48 +569,69 @@ static Htop_Reaction actionHelp(State* st) {
...
+#define addbartext(attr, text) \
+   do { \
+      addattrstr(CRT_colors[DEFAULT_COLOR], "/"); \
+      addattrstr((attr), (text)); \
+   } while(0)
+#define addlastbartext(text) \
+   do { \
+      int text_len = strlen(text); \
+      attrset(CRT_colors[BAR_SHADOW]); \
+      mvaddstr(line, 56 - text_len + bar_x, (text)); \
```
Explorer09, line 590: Since you made the macro, how about putting this line in the macro?
```c
addattrstr(CRT_colors[BAR_BORDER], "]");
```

Action.c
```diff
@@ -569,48 +569,69 @@ static Htop_Reaction actionHelp(State* st) {
...
+   mvaddstr(line, 0, "CPU usage bar: ");
+
+   int bar_x = 0;
+
+#define addfirstbartext(attr, text) \
+   do { \
+      bar_x = getcurx(stdscr); \
+      addattrstr((attr), (text)); \
+   } while(0)
+#define addbartext(attr, text) \
```
Explorer09, line 581: I think the macro name would be more readable if you name it in camelCase.
Otherwise I could mis-read it as "addBartExt"

PR #1015, opened 2022-05-17
Terminal mouse support is still active even with `enable_mouse=0` or `--no-mouse`

Low-power:
The current implementation just ignores any mouse events if mouse support is disabled by user, while still unconditionally enables ncurses mouse via **mousemask(3X)** on startup in `CRT_init`

This commit fixes the issue by calling **mousemask(3X)** with a conditional mask value on starting up as well as on changing settings.

### DIFF
--- a/CRT.c
+++ b/CRT.c
@@ -900,6 +900,20 @@ void CRT_resetSignalHandlers(void) {
    signal(SIGQUIT, SIG_DFL);
 }
 
+void CRT_setMouse(bool enabled) {
+#ifdef HAVE_GETMOUSE
+   if (enabled) {
+#if NCURSES_MOUSE_VERSION > 1
+      mousemask(BUTTON1_RELEASED | BUTTON4_PRESSED | BUTTON5_PRESSED, NULL);
+#else
+      mousemask(BUTTON1_RELEASED, NULL);
+#endif
+   } else {
+      mousemask(0, NULL);
+   }
+#endif
+}
+
 void CRT_init(const Settings* settings, bool allowUnicode) {
    redirectStderr();
 
@@ -993,13 +1007,7 @@ IGNORE_WCASTQUAL_END
 #endif
       CRT_treeStrAscii;
 
-#ifdef HAVE_GETMOUSE
-#if NCURSES_MOUSE_VERSION > 1
-   mousemask(BUTTON1_RELEASED | BUTTON4_PRESSED | BUTTON5_PRESSED, NULL);
-#else
-   mousemask(BUTTON1_RELEASED, NULL);
-#endif
-#endif
+   CRT_setMouse(settings->enableMouse);
 
    CRT_degreeSign = initDegreeSign();
 }
--- a/Action.c
+++ b/Action.c
@@ -88,6 +88,7 @@ static void Action_runSetup(State* st) {
    ScreenManager_run(scr, NULL, NULL, "Setup");
    ScreenManager_delete(scr);
    if (st->settings->changed) {
+      CRT_setMouse(st->settings->enableMouse);
       Header_writeBackToSettings(st->header);
    }
 }
--- a/CRT.h
+++ b/CRT.h
@@ -185,6 +185,8 @@ extern int CRT_scrollWheelVAmount;
 
 extern ColorScheme CRT_colorScheme;
 
+void CRT_setMouse(bool enabled);
+
 void CRT_init(const Settings* settings, bool allowUnicode);
 
 void CRT_done(void);


CRT.c
```diff
@@ -900,6 +900,19 @@ void CRT_resetSignalHandlers(void) {
...
+#ifdef HAVE_GETMOUSE
+   mmask_t mask = enabled ?
+#if NCURSES_MOUSE_VERSION > 1
+      BUTTON1_RELEASED | BUTTON4_PRESSED | BUTTON5_PRESSED
+#else
+      BUTTON1_RELEASED
+#endif
+      : 0;
+   mousemask(mask, NULL);
+#endif
```
Explorer09, line 913:
```suggestion
#ifdef HAVE_GETMOUSE
   mmask_t mask = 0;
   if (enabled) {
#if NCURSES_MOUSE_VERSION > 1
      mask = BUTTON1_RELEASED | BUTTON4_PRESSED | BUTTON5_PRESSED;
#else
      mask = BUTTON1_RELEASED;
#endif
   }
#endif
   return;
```

BenBE:
Please note that keeping the mouse responsive when disabling the mouse support in the setup screen is intentional and should still be possible after this patch is applied.

Low-power:
> keeping the mouse responsive when disabling the mouse support in the setup screen

I'm not quiet sure I understand this.
Did you mean that if user unchecks `Enable the mouse` in setup screen, mouse support should remain be active until the setup screen is closed (by pressing **F10** or **ESC**)?
If so, this code does exactly that.

CRT.c
```diff
@@ -900,6 +900,22 @@ void CRT_resetSignalHandlers(void) {
    signal(SIGQUIT, SIG_DFL);
 }
 
+void CRT_setMouse(bool enabled) {
+#ifdef HAVE_GETMOUSE
+   mmask_t mask;
```
BenBE, line 905: Avoid uninitialized variables.

BenBE:
> > keeping the mouse responsive when disabling the mouse support in the setup screen
> 
> I'm not quiet sure I understand this. Did you mean that if user unchecks `Enable the mouse` in setup screen, mouse support should remain be active until the setup screen is closed (by pressing **F10** or **ESC**)? If so, this code does exactly that.

Exactly. Wasn't sure if you were aware of this subtle feature or not, thus I mentioned it.

BenBE:
> But anyways...
> 
> > This code change can be committed by users with write permissions.
> 
> I sometimes wondering why there are so many 'change requested` for tiny things.
> 
> Why not simply commit with any change you seen fit?
> 
> I consider pull requests as suggestions, you don't need to merge my commit as-is.

While we do this occasionally, we somewhat avoid it for the benefit of providing feedback to those who take the time to create pull requests. Furthermore working and maintaining `htop` (and other projects) requires attention for details, which ensures both a somewhat consistent style of coding (which is BTW documented in the docs folder) and that most bugs/pitfalls can be avoided/mitigated before they are introduced in the code base. Many people also struggle to pick things up when commits are edited/rebased by someone else.

Also people don't always have their credentials available on all the computers they work from. Thus while I can work from nearly everywhere to have a look at recent activity I don't have the key for submitting things to the repository placed on every system (security considerations being part of why).

Or to summarize: We consider PRs as suggestions too, but prefer the submitter to be in charge until things are merged. Also access to push things to the repository is not always available. Even other reasons for each individual case may exist.

> @Explorer09, what is that `return;` statement used for? why it is needed?

I think this stray `return;` is a left-over from extracting this code block into its own function. You should be able to remove/drop it.

Explorer09:
> Explorer09, what is that `return;` statement used for? why it is needed?

It's my personal style to indicate the function may be blank/no-op. And I don't like leaving a function body completely empty without any indication that the empty body is intentional.

> I sometimes wondering why there are so many 'change requested` for tiny things.

It's about the coding style. When the code can be rewritten in the way that can be more readable, I'll suggest that to you, just like other developers would do.

> Why not simply commit with any change you seen fit?

I didn't know if you would give me permission to push the code towards your branch. Many developers don't like their branch (or pull request) to be edited by others, so the permission is not there by default. Besides, if I had a better idea on how a pull request can be done, I'll fork your code and make a separate PR already.

CRT.c
```diff
@@ -900,6 +900,20 @@ void CRT_resetSignalHandlers(void) {
    signal(SIGQUIT, SIG_DFL);
 }
 
+void CRT_setMouse(bool enabled) {
+#ifdef HAVE_GETMOUSE
+   if(enabled) {
```
Explorer09, line 905: Space, please?
```suggestion
   if (enabled) {
```

Low-power:
> Space, please?

I often forgot that, my finger didn't use to type that space, sorry.

This is just a minor style issue, please go ahead amend my commit(s).

PR #1018, opened 2022-05-18
Use of NULL in execlp() must have a pointer cast.

Explorer09:
See this article: https://ewontfix.com/11/
In short: The NULL constant is not required by the standards to have a pointer type. This was a design defect in C and C++ languages. In C++ before C++11, `NULL` has been traditionally defined as `0` (integer type, but conforming), which can cause problems when it's used as an argument in a variadic function. (C++11 introduced `nullptr` specifically to address this problem.) musl libc defines `NULL` as `0L` in the hopes of generating a compiler warning when `NULL` is used improperly (e.g. in a variadic function). We should solve this by always doing a pointer cast whenever we use `NULL` in a variadic function.

Note: [POSIX exec function](https://pubs.opengroup.org/onlinepubs/9699919799/functions/exec.html) example uses `(char *)0` as the terminating argument. [Linux exec(3) man page](https://man7.org/linux/man-pages/man3/exec.3.html) uses `(char *)NULL`. Both would work, but not plain `NULL` without a cast.

### DIFF
--- a/OpenFilesScreen.c
+++ b/OpenFilesScreen.c
@@ -120,7 +120,9 @@ static OpenFiles_ProcessData* OpenFilesScreen_getProcessData(pid_t pid) {
       close(fdnull);
       char buffer[32] = {0};
       xSnprintf(buffer, sizeof(buffer), "%d", pid);
-      execlp("lsof", "lsof", "-P", "-o", "-p", buffer, "-F", NULL);
+      // Use of NULL in variadic functions must have a pointer cast.
+      // The NULL constant is not required by standard to have a pointer type.
+      execlp("lsof", "lsof", "-P", "-o", "-p", buffer, "-F", (char *)NULL);
       exit(127);
    }
    close(fdpair[1]);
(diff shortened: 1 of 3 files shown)


Low-power:
In addition to the man page from the Linux man-pages project, the **exec(2)** man page from Solaris has followings:

> SYNOPSIS
>     #include <unistd.h>
>
>     int execl(const char *path, const char *arg0,
>          ...  /* const char *argn, (char *)0 */);

OpenFilesScreen.c
```diff
@@ -120,7 +120,9 @@ static OpenFiles_ProcessData* OpenFilesScreen_getProcessData(pid_t pid) {
       close(fdnull);
       char buffer[32] = {0};
       xSnprintf(buffer, sizeof(buffer), "%d", pid);
-      execlp("lsof", "lsof", "-P", "-o", "-p", buffer, "-F", NULL);
+      // Use of NULL in variadic function must have a pointer cast.
```
BenBE, line 123:
```suggestion
      // Use of NULL in variadic functions must have a pointer cast.
```

BenBE (approved):

Issue #1013, opened 2022-05-14
Default color scheme: Avoid using dark color fonts on black bg


darkblaze69:
htop 3.2.0
Dark color fonts on black bg shouldn't be used.
e.g. Previously in <3.2 threads where green, but now they are unreadable dark blue.

![htop](https://user-images.githubusercontent.com/6976920/168428136-c70030f7-8200-441b-9326-1c1588353980.png)

fasterit:
In the default color profile, we have:
~~~
      [PROCESS_THREAD] = ColorPair(Green, Black),
      [PROCESS_THREAD_BASENAME] = A_BOLD | ColorPair(Green, Black),
~~~

Showing custom thread names, which you have enabled in Setup (F2) -> Display Options, we chose another color to differentiate these. And that is Blue by default. Try a different color preset if that works better for you?

eworm-de:
I would second that. The default color preset is great, except for its usage of (non-bold) blue.
Any chance to switch that to bold blue? Should be perfectly visible with light and dark background then.

fasterit:
@BenBE Custom thread names is your turf...

fasterit:
~~~
diff --git a/CRT.c b/CRT.c
index 868f2377..4391f7c0 100644
--- a/CRT.c
+++ b/CRT.c
@@ -156,7 +156,7 @@ static int CRT_colorSchemes[LAST_COLORSCHEME][LAST_COLORELEMENT] = {
       [PROCESS_THREAD] = ColorPair(Green, Black),
       [PROCESS_THREAD_BASENAME] = A_BOLD | ColorPair(Green, Black),
       [PROCESS_COMM] = ColorPair(Magenta, Black),
-      [PROCESS_THREAD_COMM] = ColorPair(Blue, Black),
+      [PROCESS_THREAD_COMM] = A_BOLD | ColorPair(Blue, Black),
       [BAR_BORDER] = A_BOLD,
       [BAR_SHADOW] = A_BOLD | ColorPairGrayBlack,
       [SWAP] = ColorPair(Red, Black),
~~~

This looks much better, I second @eworm-de's proposal.

BenBE:
> @BenBE Custom thread names is your turf...

Bold (light) has not been used there, as bold is a marker for the executable basename.

fasterit:
> > @BenBE Custom thread names is your turf...
> 
> Bold (light) has not been used there, as bold is a marker for the executable basename.

[...]

PR #1023, opened 2022-05-20
default color preset: use bold blue for better visibility

eworm-de:
The default color preset works fine on dark background, with one
exception: blue characters are hardly readable. So let's make
(most of) these bold.
Even bold blue is fine on light background, so this sould not make
anything worse.

Addresses #1013


CRT.c
```diff
@@ -194,27 +194,27 @@ static int CRT_colorSchemes[LAST_COLORSCHEME][LAST_COLORELEMENT] = {
       [CPU_SOFTIRQ] = ColorPair(Magenta, Black),
       [CPU_STEAL] = ColorPair(Cyan, Black),
       [CPU_GUEST] = ColorPair(Cyan, Black),
-      [PANEL_EDIT] = ColorPair(White, Blue),
-      [SCREENS_OTH_BORDER] = ColorPair(Blue, Blue),
-      [SCREENS_OTH_TEXT] = ColorPair(Black, Blue),
+      [PANEL_EDIT] = A_BOLD | ColorPair(White, Blue),
```
fasterit, line 197: Why did you change that? Seems unrelated?

fasterit (changes requested):
see inline comment and review comments in the related issue tracker entry, cf. https://github.com/htop-dev/htop/issues/1013#issuecomment-1133565558 from @BenBE

CRT.c
eworm-de: My change was too quick an too dirty. 😜

eworm-de:
Taking the feedback from #1013 here...

> * The [] brackets around inactive tabs look strange. Also the dark-grey on blue is kinda hard to read. Would be better to have black there for the tab text).

Yes, that was ugly... I misread the settings.

However dark font (dark-grey or black) on (dark) blue is hardly readable. The current commit changes this to white (non-bold, so light grey) on dark blue. Can we agree on this?

> * The digits overlapping the text at the right of meters are still dark blue, making that area hard to read (could be updated when touching this anyway)

Not sure what you are referencing here...
Is this fixed in current commit already?

> * Can't check the ZRAM changes due to lack of having ZRAM configured.

Me neither. 😜
Well, could check in a VM...

BenBE:
> Taking @BenBE 's feedback from #1013 here...
> 
> > * The [] brackets around inactive tabs look strange. Also the dark-grey on blue is kinda hard to read. Would be better to have black there for the tab text).
> 
> Yes, that was ugly... I misread the settings.
> 
> However dark font (dark-grey or black) on (dark) blue is hardly readable. The current commit changes this to white (non-bold, so light grey) on dark blue. Can we agree on this?

Fine by me.

> > * The digits overlapping the text at the right of meters are still dark blue, making that area hard to read (could be updated when touching this anyway)
> 
> Not sure what you are referencing here... Is this fixed in current commit already?
> 

Reference to #930 …

cgzones (approved):

CRT.c
fasterit: Still in the latest iteration of the PR

fasterit:
Merged the good commit, ignored the color-change-for-tabs one that was unrelated and making it harder to differentiate the active tab
Thank you @eworm-de 
/DLange

eworm-de:
Did the unmerged commit cause  any fallout I did not notice? Or should it go to a separate pull request?

After all I do not have a strong opinion on this as my tabs are hidden. Just wondering.

PR #1024, opened 2022-05-22
Allocate GraphData buffer dynamically

zevweiss:
I occasionally run htop in a full-screen terminal on a wide monitor, which had been producing a clunkily truncated CPU graph in the header section:

htop

Bumping `METER_GRAPHDATA_SIZE` up to 1024 resolves the problem.

### DIFF
--- a/Meter.c
+++ b/Meter.c
@@ -86,7 +86,7 @@ void Meter_delete(Object* cast) {
    if (Meter_doneFn(this)) {
       Meter_done(this);
    }
-   free(this->drawData);
+   free(this->drawData.values);
    free(this->caption);
    free(this->values);
    free(this);
@@ -121,8 +121,9 @@ void Meter_setMode(Meter* this, int modeIndex) {
       }
    } else {
       assert(modeIndex >= 1);
-      free(this->drawData);
-      this->drawData = NULL;
+      free(this->drawData.values);
+      this->drawData.values = NULL;
+      this->drawData.nValues = 0;
 
       const MeterMode* mode = Meter_modes[modeIndex];
       this->draw = mode->draw;
@@ -289,12 +290,17 @@ static const char* const GraphMeterMode_dotsAscii[] = {
 
 static void GraphMeterMode_draw(Meter* this, int x, int y, int w) {
    const Machine* host = this->host;
+   GraphData* data = &this->drawData;
 
-   if (!this->drawData) {
-      this->drawData = xCalloc(1, sizeof(GraphData));
+   assert(w > 0);
+   if ((size_t)w * 2 > data->nValues) {
+      size_t oldNValues = data->nValues;
+      data->nValues = MAXIMUM(oldNValues + (oldNValues / 2), (unsigned int)w * 2);
+      data->values = xReallocArray(data->values, data->nValues, sizeof(*data->values));
+      memmove(data->values + (data->nValues - oldNValues), data->values, oldNValues * sizeof(*data->values));
+      memset(data->values, 0, (data->nValues - oldNValues) * sizeof(*data->values));
    }
-   GraphData* data = this->drawData;
-   const int nValues = METER_GRAPHDATA_SIZE;
+   const int nValues = data->nValues;
 
    const char* const* GraphMeterMode_dots;
    int GraphMeterMode_pixPerRow;
--- a/Meter.h
+++ b/Meter.h
@@ -20,7 +20,6 @@ in the source distribution for its full text.
 
 
 #define METER_TXTBUFFER_LEN 256
-#define METER_GRAPHDATA_SIZE 256
 
 #define METER_BUFFER_CHECK(buffer, size, written)          \
    do {                                                    \
@@ -97,7 +96,8 @@ typedef struct MeterClass_ {
 
 typedef struct GraphData_ {
    struct timeval time;
-   double values[METER_GRAPHDATA_SIZE];
+   size_t nValues;
+   double* values;
 } GraphData;
 
 struct Meter_ {
@@ -108,7 +108,7 @@ struct Meter_ {
    char* caption;
    int mode;
    unsigned int param;
-   GraphData* drawData;
+   GraphData drawData;
    int h;
    int columnWidthCount;      /**< only used internally by the Header */
    uint8_t curItems;


Explorer09:
I'm not sure of it's a good idea to expand `METER_GRAPHDATA_SIZE` unconditionally in all htop builds.

BenBE:
The better solution would dynamically allocate that memory and resize if necessary.

zevweiss:
> I'm not sure of it's a good idea to expand METER_GRAPHDATA_SIZE unconditionally in all htop builds.

Just due to the added memory consumption, or for some other reason?  It'll increase the `values` array in the `GraphData` struct from 2K to 8K, though it's not like there should be lots of instances of that running around (one per active graph-mode meter), so I'd expect the difference to still be in "pocket change" territory for the kinds of systems htop is typically used on.

> The better solution would dynamically allocate that memory and resize if necessary.

Ideally, sure, but that would also have to factor in reallocating on SIGWINCH and such; I'll take a glance at what implementing that would involve, but if it ends up taking longer than I have to invest in it in at the moment I'd still be in favor of a simple one-line patch to improve on the status quo...

zevweiss:
...okay, the dynamic-allocation approach turned out to be simpler to implement than expected, so I guess we might as well go with that.

Explorer09:
@zevweiss I think I'll incorporate your dynamic allocation change to my "colored graph" proposal. With the colored graph feature, the GraphData buffer would need to be dynamically allocated anyway. Let's avoid duplicated efforts.

Meter.h
```diff
@@ -107,7 +107,7 @@ struct Meter_ {
    char* caption;
    int mode;
    unsigned int param;
-   GraphData* drawData;
+   GraphData drawData;
```
Explorer09, line 110: I don't like embedding the `GraphData` structure into the `Meter` object. If the user has never enabled the graph mode for a meter, the graph data should remain unallocated.

Meter.c
```diff
@@ -288,12 +288,16 @@ static const char* const GraphMeterMode_dotsAscii[] = {
...
-   if (!this->drawData) {
-      this->drawData = xCalloc(1, sizeof(GraphData));
+   GraphData* data = &this->drawData;
+
+   if ((unsigned int)w * 2 > data->nValues) {
+      unsigned int oldNValues = data->nValues;
+      data->nValues = w * 2;
+      data->values = xReallocArray(data->values, data->nValues, sizeof(*data->values));
+      memmove(data->values + (data->nValues - oldNValues), data->values, oldNValues * sizeof(*data->values));
+      memset(data->values, 0, (data->nValues - oldNValues) * sizeof(*data->values));
```
Explorer09, line 298: There will be small performance problems if you reallocate buffer this way.

1.  `realloc(3)` internally moves the data from the old buffer to the new one, and you are going shift the graph data to the new location (through `memmove`). This moves the data twice.
2. If `htop` is run on a terminal emulator window, and the user resizes the window with the mouse, there would be a lot of window change events (the terminal width increases one by one for each event), and there would be a a lot of reallocations.

For the second point, I'm not sure if you like the buffer size to be rounded to a power of two.

BenBE: It doesn't hurt over-allocating here. Thus if a resize is needed, then just allocate a bunch of extra entries (like 1/8th of the existing buffer). This will reduce the number of necessary re-allocations by a lot.

The first case internally already checks if there is need to re-allocate, so the double-copying is only there in case of actual size changes for the buffer. I think if this is rare enough we should be able to live with that.

Meter.h
zevweiss: The bulk data does remain unallocated though, since the `values` member is now a pointer instead of an array.  (`GraphData` itself is only a tiny handful of bytes.)

Meter.c
zevweiss: Yeah, a multiplicative allocation increase would probably be better; I'll add that.

Explorer09: @zevweiss I'm working on another version of the `GraphData` allocation and may file a pull request separately from yours.

Meter.c
```diff
@@ -121,8 +121,8 @@ void Meter_setMode(Meter* this, int modeIndex) {
       }
    } else {
       assert(modeIndex >= 1);
-      free(this->drawData);
-      this->drawData = NULL;
+      free(this->drawData.values);
+      this->drawData.values = NULL;
```
Explorer09, line 125: Null pointer dereference is possible when you change the meter mode (style) to anything else and then back to graph.
You should set `this->drawData.nValues` to `0`. (I discovered this when working on my version of the patch (#1027))

zevweiss: Good point; now fixed.

Meter.h
```diff
@@ -97,7 +96,8 @@ typedef struct MeterClass_ {
 
 typedef struct GraphData_ {
    struct timeval time;
-   double values[METER_GRAPHDATA_SIZE];
+   unsigned int nValues;
```
BenBE, line 99: Prefer `size_t` over `unsigned int`, that way the type also conveys some meaning of the contained value.

Meter.c
```diff
@@ -288,12 +289,16 @@ static const char* const GraphMeterMode_dotsAscii[] = {
...
-      this->drawData = xCalloc(1, sizeof(GraphData));
+   GraphData* data = &this->drawData;
+
+   if ((unsigned int)w * 2 > data->nValues) {
+      unsigned int oldNValues = data->nValues;
+      data->nValues = MAXIMUM(oldNValues + (oldNValues / 2), (unsigned int)w * 2);
+      data->values = xReallocArray(data->values, data->nValues, sizeof(*data->values));
+      memmove(data->values + (data->nValues - oldNValues), data->values, oldNValues * sizeof(*data->values));
+      memset(data->values, 0, (data->nValues - oldNValues) * sizeof(*data->values));
    }
```
BenBE, line 300: Strictly speaking this should at least place an `assert(w > 0);` before casting.

Prefer `size_t` over `unsigned int`.

BenBE (changes requested):
Please rebase.

Meter.h
zevweiss: IMO that would convey the _wrong_ meaning.  `size_t` is for object sizes (i.e. a number of bytes), but that's not what this is -- it's the number of elements in the `values` array, which I think is reasonably clear as it stands.

Meter.c
zevweiss: > Strictly speaking this should at least place an assert(w > 0); before casting.

Okay, sure.

> Prefer `size_t` over `unsigned int`.

See above.

Meter.h
BenBE: Let me quote [cpp-reference.com](https://en.cppreference.com/w/c/types/size_t):
> Notes
> `size_t` can store the maximum size of a theoretically possible object of any type (including array).
>
> **`size_t` is commonly used for array indexing and loop counting.** Programs that use other types, such as `unsigned int`, for array indexing may fail on, e.g. 64-bit systems when the index exceeds [`UINT_MAX`](https://en.cppreference.com/w/c/types/limits) or if it relies on 32-bit modular arithmetic.

(Emphasis mine)

zevweiss: Commonality notwithstanding I still find it actively misleading, but if you insist...

PR #1025, opened 2022-05-22
Update LinuxProcessList.c

yong718100:
When HAVE_OPENAT is defined but HAVE_READLINKAT is not defined, at this time, the variable procFd is of type int, so if procFd use %s format,  it will cause the htop  program crash. I found this problem on Android 4.4.2 system.


BenBE:
Please use a descriptive subject for your commit&pull request.

Furthermore your PR just masks the issue as mixing openat and normal I/O API doesn't work and thus if one of those APIs is unsupported both places should act as if none of the newer was supported.

cgzones:
@yong718100 could you give #1066 a try?

PR #1029, opened 2022-05-29
Refactor code for rendering command line cache

### LINKED ISSUE #1008: Debian bug #1010780:  htop: markup of deleted executables not updated on change of color scheme
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1010780

Package: htop
Version: 3.2.0-1
Severity: minor

Recently htop started to highlight processes with deleted executables or libraries with a color which depends on the color scheme (red and yellow in the Default one). Also the command name is highlighted with a color when merged exe/comm/cmdline view is used.

Changing colors through the Setup > Color menu at the runtime does not immediately update these. For example, after switching from Default to Monochrome, the processes with deleted files are still marked with red and yellow and command names are magenta despite everything else displaying gray and white. Restarting htop or toggling some of the display options (e.g. the path display) updates it bringing display to a consistent state.

It would be expected for color scheme to apply consistently all the time.

-k

BenBE:
Fixes #1008

Note:
`Settings->lastUpdate` is implemented as an increment-only generation counter, but could also be realized as a timestamp. While working on the code though I noticed that most locations that update `Settings->changed` don't have access to `ProcessList->realtime_ms` or similar, thus there won't be an easy, consistent (monotonic) clock source available. As an monotonically incrementing counter suffices, I went with that implementation, though the actual logic does not care as long as `ProcessMergeCommand->lastUpdate < Settings->lastUpdate` is triggered whenever an update to the cached command line rendering is required.


fasterit (changes requested):
Three editorial / logical changes requested. Otherwise very nice work!

PR #1043, opened 2022-06-22
Write full buffer

cgzones:
Retry writing after a short write.

### DIFF
--- a/CRT.c
+++ b/CRT.c
@@ -834,7 +834,7 @@ static void dumpStderr(void) {
             fprintf(stderr, ">>>>>>>>>> stderr output >>>>>>>>>>\n");
             header = true;
          }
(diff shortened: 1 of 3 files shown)


CRT.c
```diff
@@ -834,7 +834,7 @@ static void dumpStderr(void) {
             fprintf(stderr, ">>>>>>>>>> stderr output >>>>>>>>>>\n");
             header = true;
          }
-         (void)! write(STDERR_FILENO, buffer, res);
+         (void)! full_write(STDERR_FILENO, buffer, res);
```
BenBE, line 837: Why keep the `!` when discarding the result anyway?

cgzones: Removed the void-cast; no longer necessary since full_write is not annotated with `__attribute__((warn_unused_result))`.

PR #1044, opened 2022-06-22
Tree tweaks

cgzones:


Process.c
```diff
@@ -786,9 +786,10 @@ void Process_writeField(const Process* this, RichString* str, ProcessField field
       bool lastItem = (this->indent < 0);
       int indent = (this->indent < 0 ? -this->indent : this->indent);
 
-      for (int i = 0; i < 32; i++) {
+      for (int i = 31; i >= 0; i--) {
          if (indent & (1U << i)) {
             maxIndent = i + 1;
```
tanriol, line 791: How about removing this loop together with `maxIndent` and replacing the next `for` loop with something the following?
```
while (indent > 1) {
   if (indent & 1) { ... } else { ... }
   indent >>= 1;
}
```

Explorer09: Is this equivalent to the following?
```c
if (indent == 0) {
   maxIndent = 0;
} else {
   maxIndent = __builtin_clz(indent) + 1;
} 
```

Explorer09: Is this equivalent to the following?
```c
if (indent == 0) {
   maxIndent = 0;
} else {
   maxIndent = (32 - __builtin_clz(indent));
} 
```

tanriol: IIUC, `__builtin_clz` is a GCC builtin, but `htop` does not require `gcc`. This could be an optimization, but does not sound worth it to me as we iterate over the same bit positions below in any case and perform string formatting, which is heavy enough to hide additional costs.

cgzones: > How about removing this loop together with `maxIndent` and replacing the next `for` loop with something the following?
> 
> ```
> while (indent > 1) {
>    if (indent & 1) { ... } else { ... }
>    ...
>    indent >>= 1;
> }
> ```

indeed, thanks!

Process.c
```diff
@@ -782,19 +782,11 @@ void Process_writeField(const Process* this, RichString* str, ProcessField field
...
+      const bool lastItem = (this->indent < 0);
 
-      for (int i = 0; i < 32; i++) {
-         if (indent & (1U << i)) {
-            maxIndent = i + 1;
-         }
-      }
-
-      for (int i = 0; i < maxIndent - 1; i++) {
+      for (int indent = (this->indent < 0 ? -this->indent : this->indent); indent > 1; indent >>= 1) {
```
tanriol, line 787: Do we have some reason to avoid `int indent = abs(this->indent)`?

BenBE: I don't see one …

cgzones: With the latest push `this->indent` is of type `int64_t`, which would depending on the system either require `labs(3)` or `llabs(3)`.
Also the manpage states:

> Trying to take the absolute value of the most negative integer is not defined.
> ...
> By default, GCC handles abs(), labs(), and (since GCC 3.0) llabs() and imaxabs() as built-in functions.

So I do not see any benefit, as modern compilers should recognize the pattern `a < 0 ? -a : a`.

Process.h
```diff
@@ -242,7 +242,7 @@ typedef struct Process_ {
    /*
     * Internal state for tree-mode.
     */
-   int indent;
+   int64_t indent;
```
BenBE, line 245: Isn't this somewhat overkill? Given that this mostly gives a level on screen? `int32_t` should suffice.

ProcessList.c
```diff
@@ -199,7 +199,7 @@ static void ProcessList_removeIndex(ProcessList* this, const Process* p, int idx
    assert(Vector_countEquals(this->processes, Hashtable_count(this->processTable)));
 }
 
-static void ProcessList_buildTreeBranch(ProcessList* this, pid_t pid, int level, int indent, bool show) {
+static void ProcessList_buildTreeBranch(ProcessList* this, pid_t pid, unsigned int level, int32_t indent, bool show) {
```
BenBE, line 202:
```suggestion
static void ProcessList_buildTreeBranch(ProcessList* this, pid_t pid, size_t level, int32_t indent, bool show) {
```

BenBE (commented):
LGTM.

ProcessList.c
cgzones: Is the depth to be expected to be bigger than 2^32-1? (Then the Process member tree_depth also needs to be updated.)

BenBE: As the `indent` bit field is limited to 32 bits right now, we could even suffice to use `unit8_t` - I thought about that too, but decided to suggest the `size_t` change instead for the implicit "number of things" semantic that comes with it.

Issue #1053, opened 2022-07-12
Deleting currently active screen segfaults


sfan5:
htop 3.1.2
Arch Linux x86_64

Steps to reproduce:
1. `HOME=/tmp htop` (empty default config)
2. F2 Down Down Down Right F9 F10

<pre>
Backtrace information:
----------------------
htop(+0x12dcc)[0x557c8e15fdcc]
htop(CRT_handleSIGSEGV+0xe9)[0x557c8e168269]
/usr/lib/libc.so.6(+0x3e8e0)[0x7f9d75e8c8e0]
htop(ProcessList_printHeader+0x60)[0x557c8e173c40]
htop(Panel_draw+0xb9)[0x557c8e1765d9]
htop(ScreenManager_run+0x17e)[0x557c8e177c3e]
htop(CommandLine_run+0x43e)[0x557c8e16875e]
/usr/lib/libc.so.6(+0x29290)[0x7f9d75e77290]
/usr/lib/libc.so.6(__libc_start_main+0x8a)[0x7f9d75e7734a]
htop(_start+0x25)[0x557c8e15e075]
</pre>

PR #1059, opened 2022-07-25
Reset ss and ssIndex Settings fields after screen removal

natoscott:
In the situation where the currently displayed screen is
removed we could end up with an invalid ss pointer (and
associated index) in the global Settings structure.  Any
subsequent access to these fields would make an invalid
pointer dereference.

Fixes: #1053

### DIFF
--- a/ScreensPanel.c
+++ b/ScreensPanel.c
@@ -132,7 +132,7 @@ static void startRenaming(Panel* super) {
    Panel_setCursorToSelection(super);
 }
 
-static void rebuildSettingsArray(Panel* super) {
+static void rebuildSettingsArray(Panel* super, int selected) {
    ScreensPanel* const this = (ScreensPanel*) super;
 
    int n = Panel_size(super);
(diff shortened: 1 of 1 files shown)


ScreensPanel.c
```diff
@@ -132,16 +132,25 @@ static void startRenaming(Panel* super) {
...
    int n = Panel_size(super);
    free(this->settings->screens);
    this->settings->screens = xMallocArray(n + 1, sizeof(ScreenSettings*));
    this->settings->screens[n] = NULL;
+   bool resetScreenIndex = true;
    for (int i = 0; i < n; i++) {
       ScreenListItem* item = (ScreenListItem*) Panel_get(super, i);
       this->settings->screens[i] = item->ss;
+      if (item->ss == this->settings->ss)
+         resetScreenIndex = false;
```
BenBE, line 147: Is there any necessaty to skip doing the re-assignment when these pointers match? Won't hurt AFAICS. I'd opt to skip the `resetScreenIndex` check ensuring that even in cases we don't anticipate the below re-assignment re-establishes a clean state.

BenBE (commented):
LGTM. Maybe one consideration.

ScreensPanel.c
natoscott: @BenBE yep, I think you're right.  I've updated the code to behave this way instead and tested its all OK still - looks good.

PR #1062, opened 2022-08-09
Add option to hide processes running in a container

### LINKED ISSUE #1036: Exclude processes running in a container (i.e. Docker)
I'm using htop on one of my servers also running various docker containers. In that case, htop can be very confusing due to the high amount of different processes.

It would be really helpful if there would be a switch to only show processes not running in a container (i.e. running directly on the host).

valdaarhun:
Might close #1036

The patch makes use of NSpid to check whether a process runs inside a pid namespace that is different from the host's init process' pid namespace. If this check is true, the cgroup name is checked to ensure it isn't a process running on the host that happens to run in a different pid namespace.

### DIFF
--- a/Process.h
+++ b/Process.h
@@ -134,6 +134,9 @@ typedef struct Process_ {
    /* This is a userland thread / LWP */
    bool isUserlandThread;
 
+   /* This process is running inside a container */
+   bool isRunningInContainer;
+
    /* Controlling terminal identifier of the process */
    unsigned long int tty_nr;
 
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -779,15 +779,15 @@ static bool LinuxProcessList_checkPidNamespace(Process *process, openat_arg_t pr
 
       char *ptr = buffer;
       int pid_ns_count = 0;
-      while(*ptr != '\0' && *ptr != '\n' && !isdigit(*ptr))
+      while (*ptr && *ptr != '\n' && !isdigit(*ptr))
          ++ptr;
 
-      while(*ptr != '\0' && *ptr != '\n') {
+      while (*ptr && *ptr != '\n') {
          if (isdigit(*ptr))
             pid_ns_count++;
-         while(isdigit(*ptr))
+         while (isdigit(*ptr))
             ++ptr;
-         while(*ptr != '\0' && *ptr != '\n' && !isdigit(*ptr))
+         while (*ptr && *ptr != '\n' && !isdigit(*ptr))
             ++ptr;
       }
 
@@ -1550,11 +1550,12 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
 
       LinuxProcessList_recurseProcTree(this, procFd, "task", proc, period);
 
-      if ((ss->flags & PROCESS_FLAG_LINUX_CGROUP) || hideRunningInContainer) {
+      if (ss->flags & PROCESS_FLAG_LINUX_CGROUP || hideRunningInContainer) {
          LinuxProcessList_readCGroupFile(lp, procFd);
-         if (hideRunningInContainer && lp->cgroup && isContainerOrVMSlice(lp -> cgroup)) {
-            if (!LinuxProcessList_checkPidNamespace(proc, procFd))
+         if (hideRunningInContainer && lp->cgroup && isContainerOrVMSlice(lp->cgroup)) {
+            if (!LinuxProcessList_checkPidNamespace(proc, procFd)) {
                goto errorReadingProcess;
+            }
          }
       }
 
@@ -1636,7 +1637,7 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
       char statCommand[MAX_NAME + 1];
       unsigned long long int lasttimes = (lp->utime + lp->stime);
       unsigned long int tty_nr = proc->tty_nr;
-      if (! LinuxProcessList_readStatFile(proc, procFd, statCommand, sizeof(statCommand)))
+      if (!LinuxProcessList_readStatFile(proc, procFd, statCommand, sizeof(statCommand)))
          goto errorReadingProcess;
 
       if (lp->flags & PF_KTHREAD) {
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -10,6 +10,7 @@ in the source distribution for its full text.
 #include "linux/LinuxProcessList.h"
 
 #include <assert.h>
+#include <ctype.h>
 #include <dirent.h>
 #include <errno.h>
 #include <fcntl.h>
@@ -763,6 +764,43 @@ static bool LinuxProcessList_readStatmFile(LinuxProcess* process, openat_arg_t p
    return r == 7;
 }
 
+static bool LinuxProcessList_checkPidNamespace(Process *process, openat_arg_t procFd) {
+   FILE *statusfile = fopenat(procFd, "status", "r");
+   if (!statusfile)
+      return false;
+
+   while (true) {
+      char buffer[PROC_LINE_LENGTH + 1];
+      if (fgets(buffer, sizeof(buffer), statusfile) == NULL)
+         break;
+
+      if (!String_startsWith(buffer, "NSpid:"))
+         continue;
+
+      char *ptr = buffer;
+      int pid_ns_count = 0;
+      while(*ptr != '\0' && *ptr != '\n' && !isdigit(*ptr))
+         ++ptr;
+
+      while(*ptr != '\0' && *ptr != '\n') {
+         if (isdigit(*ptr))
+            pid_ns_count++;
+         while(isdigit(*ptr))
+            ++ptr;
+         while(*ptr != '\0' && *ptr != '\n' && !isdigit(*ptr))
+            ++ptr;
+      }
+
+      if (pid_ns_count > 1)
+         process->isRunningInContainer = true;
+
+      break;
+   }
+
+   fclose(statusfile);
+   return true;
+}
+
 static bool LinuxProcessList_readSmapsFile(LinuxProcess* process, openat_arg_t procFd, bool haveSmapsRollup) {
    //http://elixir.free-electrons.com/linux/v4.10/source/fs/proc/task_mmu.c#L719
    //kernel will return data in chunks of size PAGE_SIZE or less.
@@ -890,6 +928,13 @@ static void LinuxProcessList_readOpenVZData(LinuxProcess* process, openat_arg_t 
 }
 
 #endif
+
+static bool isContainerOrVMSlice(char *cgroup) {
+   if (String_startsWith(cgroup, "/user") || String_startsWith(cgroup, "/system"))
+      return false;
+
+   return true;
+}
 
 static void LinuxProcessList_readCGroupFile(LinuxProcess* process, openat_arg_t procFd) {
    FILE* file = fopenat(procFd, "cgroup", "r");
@@ -1453,6 +1498,7 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
    const unsigned int activeCPUs = pl->activeCPUs;
    const bool hideKernelThreads = settings->hideKernelThreads;
    const bool hideUserlandThreads = settings->hideUserlandThreads;
+   const bool hideRunningInContainer = settings->hideRunningInContainer;
    while ((entry = readdir(dir)) != NULL) {
       const char* name = entry->d_name;
 
@@ -1504,6 +1550,14 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
 
       LinuxProcessList_recurseProcTree(this, procFd, "task", proc, period);
 
+      if ((ss->flags & PROCESS_FLAG_LINUX_CGROUP) || hideRunningInContainer) {
+         LinuxProcessList_readCGroupFile(lp, procFd);
+         if (hideRunningInContainer && lp->cgroup && isContainerOrVMSlice(lp -> cgroup)) {
+            if (!LinuxProcessList_checkPidNamespace(proc, procFd))
+               goto errorReadingProcess;
+         }
+      }
+
       /*
        * These conditions will not trigger on first occurrence, cause we need to
        * add the process to the ProcessList and do all one time scans
@@ -1523,6 +1577,12 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
          proc->show = false;
          pl->userlandThreads++;
          pl->totalTasks++;
+         Compat_openatArgClose(procFd);
+         continue;
+      }
+      if (preExisting && hideRunningInContainer && proc->isRunningInContainer) {
+         proc->updated = true;
+         proc->show = false;
          Compat_openatArgClose(procFd);
          continue;
       }
@@ -1639,10 +1699,6 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
          LinuxProcessList_readDelayAcctData(this, lp);
       }
       #endif
-
-      if (ss->flags & PROCESS_FLAG_LINUX_CGROUP) {
-         LinuxProcessList_readCGroupFile(lp, procFd);
-      }
 
       if (ss->flags & PROCESS_FLAG_LINUX_OOM) {
          LinuxProcessList_readOomData(lp, procFd);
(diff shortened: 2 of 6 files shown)


BenBE:
Please have a look at the compile issues mentioned by the CI. There seem to be some headers missing for `iswdigit` AFAICS.

linux/LinuxProcessList.c
```diff
@@ -763,6 +763,42 @@ static bool LinuxProcessList_readStatmFile(LinuxProcess* process, openat_arg_t p
...
+      char buffer[PROC_LINE_LENGTH + 1];
+      if (fgets(buffer, sizeof(buffer), statusfile) == NULL)
+         break;
+
+      if (!String_startsWith(buffer, "NSpid:"))
+         continue;
+
+      char *ptr = buffer;
+      int pid_ns_count = 1;
+      while(!iswdigit(*ptr))
```
cgzones, line 781: Why would `/proc/<pid>/status` contain non-ascii characters?
Htop generally does not support reading non-ascii values from the kernel.

valdaarhun: You are right. `/proc/<pid>/status` won't contain non-ascii characters. I think `isdigit` should be fine here. `iswdigit` is an overkill.

linux/LinuxProcessList.c
```diff
@@ -763,6 +764,42 @@ static bool LinuxProcessList_readStatmFile(LinuxProcess* process, openat_arg_t p
...
+      char buffer[PROC_LINE_LENGTH + 1];
+      if (fgets(buffer, sizeof(buffer), statusfile) == NULL)
+         break;
+
+      if (!String_startsWith(buffer, "NSpid:"))
+         continue;
+
+      char *ptr = buffer;
+      int pid_ns_count = 1;
+      while(!isdigit(*ptr))
```
BenBE, line 782: May overflow past end of the string if no digit found.

linux/LinuxProcessList.c
```diff
@@ -763,6 +764,42 @@ static bool LinuxProcessList_readStatmFile(LinuxProcess* process, openat_arg_t p
...
+         ++ptr;
+
+      while(1){
+         while(isdigit(*ptr))
+            ++ptr;
+         if (*ptr == '\n')
+            break;
+         ptr += 2;
+         pid_ns_count++;
```
BenBE, line 791: May access past end of buffer.

linux/LinuxProcessList.c
```diff
@@ -1640,8 +1684,13 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
       }
       #endif
 
-      if (ss->flags & PROCESS_FLAG_LINUX_CGROUP) {
-         LinuxProcessList_readCGroupFile(lp, procFd);
+      /* The process' cgroup name is required for checking whether it runs in a container */
+      LinuxProcessList_readCGroupFile(lp, procFd);
+      char *cgroup = lp->cgroup;
+      if ( lp->cgroup == NULL
+         || (!(String_startsWith(cgroup, "/user") || String_startsWith(cgroup, "/machine") || String_startsWith(cgroup, "/system"))) ){
```
BenBE, line 1691:
```suggestion
      if (!lp->cgroup || 
         !(String_startsWith(cgroup, "/user") || String_startsWith(cgroup, "/machine") || String_startsWith(cgroup, "/system")) ) {
```

Logic operators should appear at EOL.
Skip unnecessary (((parens))) …
There's a blank before the opening brace of a blcok (recheck the style guide and the existing source).

linux/LinuxProcessList.c
```diff
@@ -1640,8 +1684,13 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
...
+         if (!LinuxProcessList_checkPidNamespace((Process *)lp, procFd))
+            goto errorReadingProcess;
```
BenBE, line 1693: Not sure this should cause a full bail-out for reading this process …

@cgzones @fasterit What do you think?

linux/LinuxProcessList.c
```diff
@@ -1640,8 +1684,13 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
       }
       #endif
 
-      if (ss->flags & PROCESS_FLAG_LINUX_CGROUP) {
-         LinuxProcessList_readCGroupFile(lp, procFd);
+      /* The process' cgroup name is required for checking whether it runs in a container */
+      LinuxProcessList_readCGroupFile(lp, procFd);
+      char *cgroup = lp->cgroup;
```
BenBE, line 1689: This causes unnecessary increase of the (already huge) local stack frame. No need to have this copied as it's not modified. If really necessary this should have been `const` anyway. Also try to reduce scope of variables as much as possible.

linux/LinuxProcessList.c
```diff
@@ -1640,8 +1684,13 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
       }
       #endif
 
-      if (ss->flags & PROCESS_FLAG_LINUX_CGROUP) {
-         LinuxProcessList_readCGroupFile(lp, procFd);
+      /* The process' cgroup name is required for checking whether it runs in a container */
+      LinuxProcessList_readCGroupFile(lp, procFd);
```
BenBE, line 1688: Instead of reading the cgroup unconditional, make this check to succeed on the flag being selected OR the container setting active. You could then further put the below code as part of the block reading/processing the cgroup value.

linux/LinuxProcessList.c
```diff
@@ -763,6 +764,42 @@ static bool LinuxProcessList_readStatmFile(LinuxProcess* process, openat_arg_t p
...
+      if (!String_startsWith(buffer, "NSpid:"))
+         continue;
+
+      char *ptr = buffer;
+      int pid_ns_count = 1;
+      while(!isdigit(*ptr))
+         ++ptr;
+
+      while(1){
+         while(isdigit(*ptr))
```
BenBE, line 786: Code style

BenBE (changes requested):

linux/LinuxProcessList.c
valdaarhun: > Logic operators should appear at EOL.

Alright, got it. I think I got confused because of the convention used [here](https://github.com/htop-dev/htop/blob/main/ProcessList.c#L400)

linux/LinuxProcessList.c
```diff
@@ -778,21 +778,22 @@ static bool LinuxProcessList_checkPidNamespace(Process *process, openat_arg_t pr
...
 
       char *ptr = buffer;
-      int pid_ns_count = 1;
-      while(!isdigit(*ptr))
+      int pid_ns_count = 0;
+      while(*ptr != '\0' && *ptr != '\n' && !isdigit(*ptr))
          ++ptr;
 
-      while(1){
+      while(*ptr != '\0' && *ptr != '\n') {
```
BenBE, line 785:
```suggestion
      while (*ptr && *ptr != '\n') {
```

linux/LinuxProcessList.c
```diff
@@ -778,21 +778,22 @@ static bool LinuxProcessList_checkPidNamespace(Process *process, openat_arg_t pr
...
+         if (isdigit(*ptr))
+            pid_ns_count++;
          while(isdigit(*ptr))
```
BenBE, line 788:
```suggestion
         while (isdigit(*ptr))
```

linux/LinuxProcessList.c
```diff
@@ -778,21 +778,22 @@ static bool LinuxProcessList_checkPidNamespace(Process *process, openat_arg_t pr
...
             ++ptr;
-         if (*ptr == '\n')
-            break;
-         ptr += 2;
-         pid_ns_count++;
+         while(*ptr != '\0' && *ptr != '\n' && !isdigit(*ptr))
```
BenBE, line 790:
```suggestion
         while (*ptr && *ptr != '\n' && !isdigit(*ptr))
```

linux/LinuxProcessList.c
```diff
@@ -778,21 +778,22 @@ static bool LinuxProcessList_checkPidNamespace(Process *process, openat_arg_t pr
          continue;
 
       char *ptr = buffer;
-      int pid_ns_count = 1;
-      while(!isdigit(*ptr))
+      int pid_ns_count = 0;
+      while(*ptr != '\0' && *ptr != '\n' && !isdigit(*ptr))
```
BenBE, line 782:
```suggestion
      while (*ptr && *ptr != '\n' && !isdigit(*ptr))
```

linux/LinuxProcessList.c
```diff
@@ -1542,6 +1543,15 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
...
 
+      if ((ss->flags & PROCESS_FLAG_LINUX_CGROUP) || hideRunningInContainer) {
+         LinuxProcessList_readCGroupFile(lp, procFd);
+         if (hideRunningInContainer && lp->cgroup && 
+            !(String_startsWith(lp->cgroup, "/user") || String_startsWith(lp->cgroup, "/machine") || String_startsWith(lp->cgroup, "/system")) ) {
+            if (!LinuxProcessList_checkPidNamespace((Process *)lp, procFd))
+               goto errorReadingProcess;
+         }
+      }
+
```
BenBE, line 1554: Why did you move this block of code?

valdaarhun: Before moving the code, there was a significant lag between the moment one enables the "hide running in a container" option and the moment the change is actually seen on the screen.

valdaarhun: My theory is that it's because, initially, none of the processes are marked as "running in container". If the option is toggled just after this condition.

```
if ((ss->flags & PROCESS_FLAG_LINUX_CGROUP) || hideRunningInContainer)
``` 
The change will still not take effect in the next iteration of the loop, since the process will still not be marked as "running in a container" in this condition.

```
if (preExisting && hideRunningInContainer && Process_isRunningInContainer(proc))
```
It finally takes effect after one more iteration.

linux/LinuxProcessList.c
```diff
@@ -1504,6 +1543,15 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
 
       LinuxProcessList_recurseProcTree(this, procFd, "task", proc, period);
 
+      if (ss->flags & PROCESS_FLAG_LINUX_CGROUP || hideRunningInContainer) {
+         LinuxProcessList_readCGroupFile(lp, procFd);
+         if (hideRunningInContainer && lp->cgroup &&
+            !(String_startsWith(lp->cgroup, "/user") || String_startsWith(lp->cgroup, "/machine") || String_startsWith(lp->cgroup, "/system")) ) {
+            if (!LinuxProcessList_checkPidNamespace((Process *)lp, procFd))
```
cgzones, line 1550: Maybe factor the cgroup heuristic into a separate function so it's easier to understand the reason for the checks and tweak them. (Also `(Process *)lp` is just `proc`).

cgzones:
One downside seems to be that gathering the namespace information of a process of a foreign user only works for root, as `/proc/<pid>/status` (and also `/proc/<pid>/ns/`) is not world-readable.
So hiding container processes this way does not work for htop started as user.

linux/LinuxProcessList.c
valdaarhun: Got it.

linux/LinuxProcessList.c
```diff
@@ -891,6 +929,13 @@ static void LinuxProcessList_readOpenVZData(LinuxProcess* process, openat_arg_t
 
 #endif
 
+static bool isContainerOrVMSlice(char *cgroup) {
+   if (String_startsWith(cgroup, "/user") || String_startsWith(cgroup, "/system") || String_startsWith(cgroup, "-"))
```
valdaarhun, line 933: I have made a minor change in the condition based on [this article](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/resource_management_guide/sec-default_cgroup_hierarchies).

I haven't used `if (String_startsWith(cgroup, "/machine")` since, there may be processes running in a container with a different cgroup name.

BenBE: The `cgroup` name always starts with a slash. Thus if you want to match the root slice, it should be `/-.slice` as there should be exactly one. Have a look at the raw values in the (uncompressed) `CGROUP` column for examples.

As the root slice is somewhat different, I don't think it's necessary to include it in this check. Also LXC has it's markers directly below the root -> `/lxc.payload.somecontainername/system.slice/console-getty.service`.

linux/LinuxProcessList.c
```diff
@@ -1504,6 +1550,14 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
 
       LinuxProcessList_recurseProcTree(this, procFd, "task", proc, period);
 
+      if (ss->flags & PROCESS_FLAG_LINUX_CGROUP || hideRunningInContainer) {
+         LinuxProcessList_readCGroupFile(lp, procFd);
+         if (hideRunningInContainer && lp->cgroup && isContainerOrVMSlice(lp -> cgroup)) {
```
BenBE, line 1555:
```suggestion
         if (hideRunningInContainer && lp->cgroup && isContainerOrVMSlice(lp->cgroup)) {
```

No spaces around field dereferencing …

Process.h
```diff
@@ -324,6 +327,10 @@ static inline bool Process_isThread(const Process* this) {
    return Process_isUserlandThread(this) || Process_isKernelThread(this);
 }
 
+static inline bool Process_isRunningInContainer(const Process* this) {
+   return this->isRunningInContainer;
+}
+
```
BenBE, line 333: Why have this extra function/indirection when the only caller could just use the flag directly?

linux/LinuxProcessList.c
```diff
@@ -1504,6 +1550,14 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
 
       LinuxProcessList_recurseProcTree(this, procFd, "task", proc, period);
 
+      if (ss->flags & PROCESS_FLAG_LINUX_CGROUP || hideRunningInContainer) {
+         LinuxProcessList_readCGroupFile(lp, procFd);
+         if (hideRunningInContainer && lp->cgroup && isContainerOrVMSlice(lp->cgroup)) {
+            if (!LinuxProcessList_checkPidNamespace(proc, procFd))
+               goto errorReadingProcess;
+         }
```
BenBE, line 1558:
```suggestion
         if (hideRunningInContainer && lp->cgroup && isContainerOrVMSlice(lp->cgroup)) {
            if (!LinuxProcessList_checkPidNamespace(proc, procFd)) {
               goto errorReadingProcess;
            }
         }
```

No need to save on those braces. While technically the style guide allows to drop them, it's usually better to favor consistency and either keep them for all of the one-line conditions or drop them for all. In this particular block I think keeping them all looks nicer overall.

BenBE (changes requested):
The overall changes look okay now, but the commit history could manage some cleanup with `git rebase` (squashing the cleanups into the commits they apply to). This gives a much nicer history and allows for better understanding of what changes where done why (otherwise half of the changes land on some fixups unnecessarily).

P.S.: Feel free to clean up the history on your PR as much as you like. I think targeting about 3 commits (e.g.: 1. whitespace fix in `linux/LinuxProcessList.c` around line 1640, 2. Implementing the overall feature, 3. adding UI for it) might be good. Each commit for itself should be able to compile.

Process.h
valdaarhun: > Feel free to clean up the history on your PR as much as you like

Cool. I'll clean up the commit history and push it again.

valdaarhun:
> Feel free to clean up the history on your PR as much as you like

Cool. I'll clean up the commit history and push it again.

Process.h
valdaarhun: `Process_isKernelThread` and `Process_isUserlandThread` are there too. Since, those two features also have this option of hiding/displaying the option, I thought I would keep `Process_isRunningInContainer` there as well so that all three remain together. Shall I remove `Process_isRunningInContainer`?

valdaarhun: Although, now that I think about it, it might not be necessary since there is only one caller.

valdaarhun:
I have cleaned up the commit history as best as I could.

BenBE (approved):
LGTM.

cgzones (approved):

m0vie:
The `O` keyboard shortcut is not documented in the `F1` help. Could this also be added?

valdaarhun:
@m0vie Hi, thanks for pointing this out. This must have slipped my mind. It's been resolved in #1250 .

PR #1071, opened 2022-08-14
Empty environment handling

cgzones:


CRT.c
```diff
@@ -915,9 +915,8 @@ void CRT_setMouse(bool enabled) {
 #endif
 
 void CRT_init(const Settings* settings, bool allowUnicode) {
-   redirectStderr();
-
    initscr();
+   redirectStderr();
```
natoscott, line 919: It's not clear to me what problem this change is solving.  Maybe a note in the commit message for this aspect?

Settings.c
```diff
@@ -702,9 +703,10 @@ Settings* Settings_new(unsigned int initialCpuCount, Hashtable* dynamicColumns)
       this->filename = xStrdup(rcfile);
    } else {
       const char* home = getenv("HOME");
-      if (!home)
-         home = "";
-
+      if (!home) {
+         const struct passwd* pw = getpwuid(getuid());
+         home = pw ? pw->pw_dir : "";
+      }
```
natoscott, line 709: Makes sense.  Could you add a similar tweak to PCPDynamicMeters_init?  (else, np, I can do it as a follow-up)

natoscott (approved):

CRT.c
cgzones: when htop is run e.g. via `env -i htop` it just exists without any message, cause `initscr()` does not return and we have redirected stderr and the signal handlers to dump stderr are not yet installed (since they also cleanup up ncurses).

natoscott (commented):
@cgzones thanks - looks like a header file may be missing with that CI failure, otherwise all good and that initscr explanation all makes sense now.

BenBE (approved):

PR #1072, opened 2022-08-14
Linux: highlight username if process has elevated privileges

cgzones:
I.e. if binary has some file capabilities set.


BenBE:
Please implement for further platforms too.

Not sure if this is consistent with varying privs on htop's side … This currently looks just like marking all processes belonging to root. But shouldn't processes from a normal user also be marked, when they e.g. use namespace admin privs to spawn a new container? Could you please elaborate?

cgzones:
> Please implement for further platforms too.

Capabilities are Linux specific and I am not that familiar with other platforms in this regard.

> Not sure if this is consistent with varying privs on htop's side … This currently looks just like marking all processes belonging to root. But shouldn't processes from a normal user also be marked, when they e.g. use namespace admin privs to spawn a new container? Could you please elaborate?

It marks processes where `/proc/<pid>/` is owned by an unprivileged user and `/proc/<pid>/statm` by root, e.g. rtkit-daemon (user rtkit) or dbus-broker-launch (user messagebus).

BenBE:
> > Please implement for further platforms too.
> 
> Capabilities are Linux specific and I am not that familiar with other platforms in this regard.

So basically this could be interpreted as `euid != ruid`?

> > Not sure if this is consistent with varying privs on htop's side … This currently looks just like marking all processes belonging to root. But shouldn't processes from a normal user also be marked, when they e.g. use namespace admin privs to spawn a new container? Could you please elaborate?
> 
> It marks processes where `/proc/<pid>/` is owned by an unprivileged user and `/proc/<pid>/statm` by root, e.g. rtkit-daemon (user rtkit) or dbus-broker-launch (user messagebus).

That isn't clear from the description and documentation in the code alone.

PR #1081, opened 2022-09-12
PCP: DynamicMeters: Add Meter Choice framework

smalinux:
Related: https://github.com/htop-dev/htop/pull/200 && https://github.com/htop-dev/htop/pull/669

### DIFF
--- a/openbsd/Platform.c
+++ b/openbsd/Platform.c
@@ -377,13 +377,15 @@ void Platform_getBattery(double* percent, ACPresence* isOnAC) {
    }
 }
 
-char** Platform_getLocalIPv4addressChoices(void) {
+char** Platform_getLocalIPv4addressChoices(Meter* meter) {
    // TODO
+   (void) meter;
    return NULL;
 }
 
-char** Platform_getLocalIPv6addressChoices(void) {
+char** Platform_getLocalIPv6addressChoices(Meter* meter) {
    // TODO
+   (void) meter;
    return NULL;
 }
 
@@ -403,8 +405,9 @@ void Platform_getLocalIPv6address(const char* choice, char* buffer, size_t size)
    return;
 }
 
-char **Platform_getDiskUsageChoices(void) {
+char **Platform_getDiskUsageChoices(Meter* meter) {
    // TODO
+   (void) meter;
    return NULL;
 }
 
@@ -412,4 +415,10 @@ void Platform_getDiskUsage(const char* choice, DiskUsageData *data) {
    // TODO
    (void) choice;
    (void) data;
+}
+
+char **Platform_getDynamicMeterChoices(Meter* meter) {
+   // TODO
+   (void) meter;
+   return NULL;
 }
--- a/openbsd/Platform.c
+++ b/openbsd/Platform.c
@@ -375,3 +375,29 @@ void Platform_getBattery(double* percent, ACPresence* isOnAC) {
          *isOnAC = s.value;
    }
 }
+
+char** Platform_getLocalIPv4addressChoices(void) {
+   // TODO
+   return NULL;
+}
+
+char** Platform_getLocalIPv6addressChoices(void) {
+   // TODO
+   return NULL;
+}
+
+void Platform_getLocalIPv4address(const char* choice, char* buffer, size_t size) {
+   // TODO
+   (void) choice;
+   (void) buffer;
+   (void) size;
+   return;
+}
+
+void Platform_getLocalIPv6address(const char* choice, char* buffer, size_t size) {
+   // TODO
+   (void) choice;
+   (void) buffer;
+   (void) size;
+   return;
+}
--- a/openbsd/Platform.c
+++ b/openbsd/Platform.c
@@ -313,8 +313,9 @@ bool Platform_getDiskIO(DiskIOData* data) {
    return false;
 }
 
-bool Platform_getNetworkIO(NetworkIOData* data) {
+bool Platform_getNetworkIO(const char* choice, NetworkIOData* data) {
    // TODO
+   (void) choice;
    (void)data;
    return false;
 }
--- a/openbsd/Platform.c
+++ b/openbsd/Platform.c
@@ -401,3 +401,14 @@ void Platform_getLocalIPv6address(const char* choice, char* buffer, size_t size)
    (void) size;
    return;
 }
+
+char **Platform_getDiskUsageChoices(void) {
+   // TODO
+   return NULL;
+}
+
+void Platform_getDiskUsage(const char* choice, DiskUsageData *data) {
+   // TODO
+   (void) choice;
+   (void) data;
+}
--- a/pcp/PCPDynamicMeter.c
+++ b/pcp/PCPDynamicMeter.c
@@ -70,6 +70,9 @@ static void PCPDynamicMeter_parseMetric(PCPDynamicMeters* meters, PCPDynamicMete
       /* lookup a dynamic metric with this name, else create */
       metric = PCPDynamicMeter_lookupMetric(meters, meter, key);
 
+      /* fill metric's indom */
+      metric->indom = PCPMetric_lookupInDom(value);
+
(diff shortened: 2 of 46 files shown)


pcp/PCPDynamicMeter.c
```diff
@@ -458,3 +476,25 @@ void PCPDynamicMeter_display(PCPDynamicMeter* this, ATTR_UNUSED const Meter* met
...
+
+   count = pmGetInDom(indom, &in, &raw);
+   if (count < 0)
+      return NULL;
+
+   char** ret = xMalloc(count*sizeof(char*));
+
+   for (int i = 0; i < count; i++)
+      ret[i] = xStrdup(raw[i]);
+   ret[count] = NULL;
```
smalinux, line 494: I tried to return raw directly, but I got a Segfault. I don't know why! 
So I had to copy `raw` into `ret`

pcp/PCPDynamicMeter.c
```diff
@@ -458,3 +476,25 @@ void PCPDynamicMeter_display(PCPDynamicMeter* this, ATTR_UNUSED const Meter* met
...
+   unsigned int indom = this->metrics->indom;
+   int count = 0;
+   int* in;
+   char** raw;
+
+   count = pmGetInDom(indom, &in, &raw);
+   if (count < 0)
+      return NULL;
+
+   char** ret = xMalloc(count*sizeof(char*));
```
BenBE, line 490:
```suggestion
   char** ret = xCalloc(count, sizeof(char*));
```

pcp/PCPDynamicMeter.c
BenBE, line 494: OOB write

BenBE: Probably because `pmGetInDom` sets up references in `raw` to point into the same memory allocation.

solaris/Platform.c
```diff
@@ -325,8 +325,9 @@ bool Platform_getDiskIO(DiskIOData* data) {
    return false;
 }
 
-bool Platform_getNetworkIO(NetworkIOData* data) {
+bool Platform_getNetworkIO(const char* choice, NetworkIOData* data) {
    // TODO
+   (void) choice;
```
BenBE, line 330:
```suggestion
   (void)choice;
```

pcp/Platform.c
```diff
@@ -710,6 +713,38 @@ void Platform_getBattery(double* level, ACPresence* isOnAC) {
...
+
+void Platform_getLocalIPv4address(const char* choice, char* buffer, size_t size) {
+   // TODO
+   (void) choice; (void) buffer; (void) size;
+   return;
+}
+
+void Platform_getLocalIPv6address(const char* choice, char* buffer, size_t size) {
+   // TODO
+   (void) choice; (void) buffer; (void) size;
```
BenBE, line 734:
```suggestion
   (void)choice;
   (void)buffer;
   (void)size;
```

pcp/Platform.c
```diff
@@ -710,6 +713,38 @@ void Platform_getBattery(double* level, ACPresence* isOnAC) {
...
+}
+
+char **Platform_getDiskUsageChoices(ATTR_UNUSED Meter* meter) {
+   // TODO
+   return NULL;
+}
+
+void Platform_getDiskUsage(const char* choice, DiskUsageData *data) {
+   // TODO
+   (void) choice; (void) data;
```
BenBE, line 745:
```suggestion
   (void)choice;
   (void)data;
```

unsupported/Platform.c
```diff
@@ -162,3 +164,34 @@ void Platform_getHostname(char* buffer, size_t size) {
...
+}
+
+char** Platform_getLocalIPv6addressChoices(ATTR_UNUSED Meter* meter) {
+   return NULL;
+}
+
+void Platform_getLocalIPv4address(const char* choice, char* buffer, size_t size) {
+   (void) choice;
+   (void) buffer; (void)
+   size;
```
BenBE, line 179:
```suggestion
   (void)choice;
   (void)buffer;
   (void)size;
```

BenBE (changes requested):
Not quite happy with the `ATTR_UNUSED` littered all over the headers when it's actually just an implementation detail for the C files (except when `static inline`-ing the implementation in the header.

Also the `ATTR_UNUSED` usage and `(void)var;`-casts are mixed. Should only use one of both (unless there's an `assert` forcing the cast).

natoscott:
> Not quite happy with [...]

+1 - this feedback is more related to PR #200 though isn't it?

pcp/PCPMetric.c
```diff
@@ -178,3 +178,21 @@ bool PCPMetric_fetch(struct timeval* timestamp) {
...
+      return 0;
+   }
+
+   sts = pmLookupDesc(pmid, &desc);
+   if (sts < 0) {
+      return 0;
+   }
+
+   return desc.indom;
+}
```
natoscott, line 198: @smalinux I think we don't need this function - we have already done a descriptor lookup of each metric during initialization, and they're stashed off the main Platform structure for PCP.  It's better to use the cached values since pmLookupName and pmLookupDesc here involve round-trips to pmcd.  I'll send through some small cleanups including this little improvement shortly.

openbsd/Platform.c
```diff
@@ -375,3 +376,49 @@ void Platform_getBattery(double* percent, ACPresence* isOnAC) {
...
+   // TODO
+   (void) choice;
+   (void) data;
+}
+
+char **Platform_getDynamicMeterChoices(Meter* meter) {
+   // TODO
+   (void) meter;
+   return NULL;
+}
```
natoscott, line 424: Still need to do something for each of these.  It makes me wonder though, for PCP we have all of the component instance names readily available to us for a meter - could we make a generic implementation of getChoices() for PCP and have every meter that uses metrics with an instance domain be one of these 'choice' meters, automatically?

natoscott (changes requested):

openbsd/Platform.c
smalinux: > could we make a generic implementation of getChoices() for PCP and have every meter that uses metrics with an instance domain be one of these 'choice' meters, automatically?

Sorry, I don't understand this part! I have a generic function that returns all the choices here:
(that is actually return the external instance name list)
https://github.com/htop-dev/htop/blob/7d6d345e286fe890cd504be7522a02ed76bee8c8/DynamicMeter.c#L132
https://github.com/htop-dev/htop/blob/7d6d345e286fe890cd504be7522a02ed76bee8c8/pcp/Platform.c#L869
https://github.com/htop-dev/htop/blob/7d6d345e286fe890cd504be7522a02ed76bee8c8/pcp/PCPDynamicMeter.c#L480

smalinux: and the `curChoice` got saved in htoprc...

BenBE:
> > Not quite happy with [...]
> 
> +1 - this feedback is more related to PR #200 though isn't it?
> 

Saw it in the PCP commit, thus mentioned it here. IDK for the same in #200, but would apply there the same; haven't looked into this though.

openbsd/Platform.c
natoscott: @smalinux aha, right you are - so the remaining piece then is those TODO items in the code relating to specific meters like network/hostname ... can we just call Platform_getDynamicMeterChoices there too?

smalinux: @natoscott exactly :)

smalinux: > ... can we just call Platform_getDynamicMeterChoices there too?

I do not think so! these Meters (network/hostname) depend on these Metrics: https://github.com/htop-dev/htop/pull/200#discussion_r907890905
I think I should provide a static implementation for them like any normal Meter... like this: #992

or we could ignore these TODO items completely and provide the same Meters as DynamicMeters via pcp/meters/ config files...

pcp/PCPMetric.c
smalinux: @natoscott done!

PR #1083, opened 2022-09-13
Use pmLookupDescs(3) function if available from libpcp

natoscott:
This is a relative new, single-round-trip variant of the pmLookupDesc(3) function for metric descriptors lookup.

### DIFF
--- a/pcp/Platform.c
+++ b/pcp/Platform.c
@@ -1,8 +1,8 @@
 /*
 htop - linux/Platform.c
 (C) 2014 Hisham H. Muhammad
-(C) 2020-2021 htop dev team
-(C) 2020-2021 Red Hat, Inc.
+(C) 2020-2022 htop dev team
+(C) 2020-2022 Red Hat, Inc.
 Released under the GNU GPLv2+, see the COPYING file
 in the source distribution for its full text.
 */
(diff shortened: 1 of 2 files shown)


pcp/Platform.c
```diff
@@ -248,6 +248,25 @@ static const char* Platform_metricNames[] = {
...
+
+      int sts = pmLookupDesc(pmids[i], &descs[i]);
+      if (sts < 0) {
+         if (pmDebugOptions.appl0)
+            fprintf(stderr, "Error: cannot lookup metric %s(%s): %s\n",
+                    pcp->names[i], pmIDStr(pcp->pmids[i]), pmErrStr(sts));
+         pmids[i] = PM_ID_NULL;
+      }
+   }
+   return 0;
```
BenBE, line 266: This completely hides when all metrics failed to resolve. Shouldn't we at least check that at least one metric could be resolved in order to return "success"? What's the PCP API behaviour for partial success?

pcp/Platform.c
```diff
@@ -248,6 +248,25 @@ static const char* Platform_metricNames[] = {
    [PCP_METRIC_COUNT] = NULL
 };
 
+#ifndef HAVE_PMLOOKUPDESCS
```
BenBE, line 251: I know this works the way specified, but documenting the premise of providing the function if not available upstream would be a nice touch. ;-)

BenBE (commented):
LGTM. Some minor notes though …

pcp/Platform.c
natoscott: Good point, will do.

pcp/Platform.c
natoscott: Depends on the API, but in this case using metric identifier PM_ID_NULL indicates failure at some point of the lookup path (metric name lookup has happened just prior to this descriptor lookup).  I'll add some logic to handle the case where we end up with no metrics available across all of htop - as a final sanity check during startup.

BenBE (approved):

PR #1085, opened 2022-09-13
Avoid showing garbage values on startup (network).

tomasandrle:
Uninitialized "cached" values may be shown before "passedTimeInMs" is past the 500ms threshold.


tomasandrle:
Saw this happen when I added disk I/O and network to the screen & restarted.

Screenshot 2022-09-13 at 19 45 22

natoscott:
I'm pretty sure this code change has no effect (those global values are defined as zero by the C standard, there's no need to initialize them explicitly).

PR #1101, opened 2022-10-05
Linux: Add entropy meter

Geneo-5:
Add entropy avail less pcp.


cgzones:
This does not work on modern kernels, where those values are hardcoded to `256`, see https://www.zx2c4.com/projects/linux-rng-5.17-5.18/.

linux/RNGEntropy.c
```diff
@@ -0,0 +1,83 @@
...
+#include <sys/statvfs.h>
+
+#include "Object.h"
+#include "XUtils.h"
+
+static void getRNGEntropy(Meter* this) {
+   FILE* file = fopen(PROCDIR "/sys/kernel/random/entropy_avail", "r");
+   if (!file)
+      return;
+   int match = fscanf(file, "%le", &this->values[0]);
```
Explorer09, line 26: "%le", seriously?
I expect reading an integer and then converting to a floating point (for storing in `this->values[n]`). This would make the code more robust.
Also, don't skip the return value check, you probably need to set a placeholder value in case of read errors. Something like this:
```
if (!match)
   this->values[0] = NAN;
```

linux/RNGEntropy.c
```diff
@@ -0,0 +1,83 @@
...
+   (void) match;
+   fclose(file);
+}
+
+static void getRNGPoolsize(Meter* this) {
+   FILE* file = fopen(PROCDIR "/sys/kernel/random/poolsize", "r");
+   if (!file)
+      return;
+   int match = fscanf(file, "%le", &this->total);
```
Explorer09, line 35: Same as above. (Don't use "%le" but read it as integer, and do check the return value for read errors.)

linux/RNGEntropy.c
```diff
@@ -0,0 +1,83 @@
+/*
+htop - RNGEntropy.c
+(C) 2022 Loic Jourdheuil
+Released under the GNU GPLv2+, see the COPYING file
+in the source distribution for its full text.
+*/
+
+#include "linux/RNGEntropy.h"
+
+#include "CRT.h"
```
BenBE, line 10: Would normally go down alongside the include for `Object.h` and `XUtils.h`.

linux/RNGEntropy.c
```diff
@@ -0,0 +1,83 @@
...
+#include <stdio.h>
+#include <unistd.h>
+#include <sys/statfs.h>
+#include <sys/statvfs.h>
+
+#include "Object.h"
+#include "XUtils.h"
+
+static void getRNGEntropy(Meter* this) {
+   FILE* file = fopen(PROCDIR "/sys/kernel/random/entropy_avail", "r");
```
BenBE, line 26: Not sure if going with that whole `fopen` API and its overhead to read just a single integer is worth the overhead.

linux/RNGEntropy.c
```diff
@@ -0,0 +1,83 @@
...
+   if (!file)
+      return;
+   int match = fscanf(file, "%le", &this->total);
+   (void) match;
+   fclose(file);
+}
+
+static void RNGEntropy_updateValues(Meter* this) {
+   getRNGEntropy(this);
+   getRNGPoolsize(this);
```
BenBE: Why not have these in one function? They need to be updated in sync anyway and their code is easy enough.

linux/RNGEntropy.c
```diff
@@ -0,0 +1,83 @@
...
+
+   if ((this->total <= 0) || (this->values[0] <= 0)) {
+      xSnprintf(this->txtBuffer, sizeof(this->txtBuffer), "N/A");
+   }
+
+   xSnprintf(this->txtBuffer, sizeof(this->txtBuffer), "%.1f%%", 100.0 * this->values[0] / this->total);
```
BenBE: The `N/A` output is overwritten by the second `xSnprintf` due to a missing `return` statement.

linux/RNGEntropy.c
```diff
@@ -0,0 +1,83 @@
...
+   const Meter* this = (const Meter*)cast;
+   char buffer[50];
+   int len;
+
+   RichString_appendAscii(out, CRT_colors[METER_TEXT], " Entropy: ");
+   len = xSnprintf(buffer, sizeof(buffer), "%.0f", this->values[0]);
+   RichString_appendnAscii(out, CRT_colors[METER_VALUE_OK], buffer, len);
+   RichString_appendAscii(out, CRT_colors[METER_TEXT], " Poolsize: ");
+   len = xSnprintf(buffer, sizeof(buffer), "%.0f", this->total);
+   RichString_appendAscii(out, CRT_colors[METER_VALUE], buffer);
```
BenBE, line 68: Shouldn't this apply similar sanity checks as for the `updateValues` routine above?

linux/RNGEntropy.c
```diff
@@ -0,0 +1,83 @@
...
+   .updateValues = RNGEntropy_updateValues,
+   .defaultMode = BAR_METERMODE,
+   .maxItems = 1,
+   .total = 4096.0,
+   .attributes = RNGEntropy_attributes,
+   .name = "Entropy",
+   .uiName = "RNG Entropy",
+   .description = "RNG entropy overview",
+   .caption = "RNG"
+};
```
BenBE, line 90: Missing newline @ EOF.

linux/RNGEntropy.h
```diff
@@ -0,0 +1,15 @@
...
+Released under the GNU GPLv2+, see the COPYING file
+in the source distribution for its full text.
+*/
+
+#include "Meter.h"
+
+
+extern const MeterClass RNGEntropy_class;
+
+#endif /* HEADER_RNGEntropy */
```
BenBE, line 15: Missing newline @ EOF.

BenBE (commented):
I've not looked into the kernel sources regarding the actual current implementation, but the information in `entropy_avail` have been unreliable for quite a while already. Thus there is little value in putting in a meter for it (systems that lack enough entropy usually don't have a user interface connected and those that have a user interface usually also don't suffer from lack of entropy).

linux/RNGEntropy.c
```diff
@@ -0,0 +1,92 @@
...
+#include <stdlib.h>
+
+#include "CRT.h"
+#include "Object.h"
+#include "XUtils.h"
+
+#define INVALID_VALUE ((unsigned int)-1)
+
+static unsigned int entropy = INVALID_VALUE;
+static unsigned int poolsize = INVALID_VALUE;
```
BenBE, line 19:
```suggestion
static unsigned size_t = INVALID_VALUE;
static unsigned size_t = INVALID_VALUE;
```

Requires `<stddef.h>`

linux/RNGEntropy.c
```diff
@@ -0,0 +1,92 @@
...
+      RichString_appendAscii(out, CRT_colors[METER_TEXT], " Poolsize: ");
+      RichString_appendAscii(out, CRT_colors[METER_VALUE], "N/A");
+   } else {
+      RichString_appendAscii(out, CRT_colors[METER_TEXT], " Entropy: ");
+      len = xSnprintf(buffer, sizeof(buffer), "%u", entropy);
+      RichString_appendnAscii(out, CRT_colors[METER_VALUE_OK], buffer, len);
+      RichString_appendAscii(out, CRT_colors[METER_TEXT], " Poolsize: ");
+      len = xSnprintf(buffer, sizeof(buffer), "%u", poolsize);
+      RichString_appendAscii(out, CRT_colors[METER_VALUE], buffer);
+   }
```
BenBE, line 70: Why mark one with `METER_VALUE_OK` and the other one with `METER_VALUE`, when it would be more logical to have `METER_VALUE` in the error case and `METER_VALUE_OK` for normal display?

Also that last `RichString_appendAscii` in line 69 should better be `RichString_appendnAscii` like with line 66. Safes some cycles since we know the length of the string there from `xSnprintf`.

linux/RNGEntropy.c
```diff
@@ -0,0 +1,92 @@
...
+
+   if (poolsize == INVALID_VALUE) {
+      r = xReadfile(PROCDIR "/sys/kernel/random/poolsize", buffer, sizeof(buffer));
+      if (r <= 0) {
+         goto err;
+      }
+      poolsize = (unsigned int)strtoul(buffer, NULL, 10);
+      if (poolsize == 0) {
+         /* invalid poolsize value */
+         poolsize = INVALID_VALUE;
```
BenBE, line 33: This will continuously check the pool size again if it fails to read this value.

linux/RNGEntropy.c
```diff
@@ -0,0 +1,92 @@
...
+static void RNGEntropy_updateValues(Meter* this) {
+   char buffer[11];
+   ssize_t r;
+
+   if (poolsize == INVALID_VALUE) {
+      r = xReadfile(PROCDIR "/sys/kernel/random/poolsize", buffer, sizeof(buffer));
+      if (r <= 0) {
+         goto err;
+      }
+      poolsize = (unsigned int)strtoul(buffer, NULL, 10);
```
BenBE, line 30: See type change suggestion for `poolsize` above. This narrowing cast is a good indicator for this being a bad idea … ;-)

linux/RNGEntropy.c
```diff
@@ -0,0 +1,92 @@
...
+      entropy = (unsigned int)strtoul(buffer, NULL, 10);
+   }
+
+   this->values[0] = (100.0 * entropy) / poolsize;
+   xSnprintf(this->txtBuffer, sizeof(this->txtBuffer), "%.1f%%", this->values[0]);
+   return;
+
+err:
+   xSnprintf(this->txtBuffer, sizeof(this->txtBuffer), "N/A");
+   this->values[0] = 0.0;
```
BenBE, line 51:
```suggestion
   char buffer[33];
   ssize_t r;

   if (poolsize == INVALID_VALUE) {
      r = xReadfile(PROCDIR "/sys/kernel/random/poolsize", buffer, sizeof(buffer));
      poolsize = r > 0 ? strtoul(buffer, NULL, 10) : 0;
   }

   r = xReadfile(PROCDIR "/sys/kernel/random/entropy_avail", buffer, sizeof(buffer));
   entropy = r > 0 ? strtoul(buffer, NULL, 10) : INVALID_VALUE;
   
   if (poolsize && poolsize != INVALID_VALUE && poolsize > this->total) {
      this->total = poolsize;
   }

   this->values[0] =  ? entropy : NAN;
   if (entropy != INVALID_VALUE) {
      xSnprintf(this->txtBuffer, sizeof(this->txtBuffer), "%zu/%zu", entropy, poolsize);
      this->values[0] = entropy;
   } else {
      xSnprintf(this->txtBuffer, sizeof(this->txtBuffer), "N/A");
      this->values[0] = NAN;
   }
```

linux/RNGEntropy.c
Geneo-5: OK

PR #1102, opened 2022-10-06
PCP: DynamicScreens

smalinux:


DynamicScreen.c
```diff
@@ -0,0 +1,61 @@
...
+
+void DynamicScreens_delete(Hashtable* dynamics) {
+   if (dynamics) {
+      Platform_dynamicScreensDone(dynamics);
+      Hashtable_delete(dynamics);
+   }
+}
+
+typedef struct {
+   unsigned int key;
```
BenBE, line 35: Shouldn't this be `ht_key_t`, as this is effectively directly passed to the `Hashtable` implementation?

DynamicScreen.c
```diff
@@ -0,0 +1,61 @@
...
+static void DynamicScreen_compare(ht_key_t key, void* value, void* data) {
+   const DynamicScreen* screen = (const DynamicScreen*)value;
+   DynamicIterator* iter = (DynamicIterator*)data;
+   if (String_eq(iter->name, screen->name)) {
+      iter->found = true;
+      iter->key = key;
+   }
+}
+
+bool DynamicScreen_search(Hashtable* dynamics, const char* name, unsigned int* key) {
```
BenBE, line 49:
```suggestion
bool DynamicScreen_search(Hashtable* dynamics, const char* name, ht_key_t* key) {
```

DynamicScreen.c
```diff
@@ -0,0 +1,61 @@
...
+   DynamicIterator iter = { .key = 0, .name = name, .found = false };
+   if (dynamics)
+      Hashtable_foreach(dynamics, DynamicScreen_compare, &iter);
+   if (key)
+      *key = iter.key;
+   return iter.found;
+}
+
+const char* DynamicScreen_lookup(Hashtable* dynamics, unsigned int key) {
```
BenBE, line 58:
```suggestion
const char* DynamicScreen_lookup(Hashtable* dynamics, ht_key_t key) {
```

GenericData.c
```diff
@@ -0,0 +1,71 @@
...
+
+   return (ScreenSettings_getActiveDirection(ss) == 1) ? result : -result;
+}
+
+int GenericData_compareByKey_Base(const GenericData* g1, const GenericData* g2, ATTR_UNUSED ProcessField key) {
+
+   (void) g1;
+   (void) g2;
+
+   return 0;
```
BenBE, line 61: Missing implementation? Or missing documentation/comment?

GenericData.h
```diff
@@ -0,0 +1,61 @@
...
+
+} GenericData;
+
+
+// Implemented in platform-specific code:
+void GenericData_writeField(const GenericData* this, RichString* str, int field);
+int GenericData_compare(const void* v1, const void* v2);
+void GenericData_delete(Object* cast);
+
+typedef GenericData* (*GenericData_New)(const struct Settings_*);
```
BenBE, line 32:
```suggestion
typedef GenericData* (*GenericData_New)(const Settings* settings);
```

GenericData.h
```diff
@@ -0,0 +1,61 @@
...
+#include "RichString.h"
+#include "Settings.h"
+
+
+typedef struct GenericData_ {
+   /* Super object for emulated OOP */
+   Object super;
+
+   /* Pointer to quasi-global data structures */
+   const struct Settings_* settings;
```
BenBE, line 22:
```suggestion
   const Settings* settings;
```

GenericData.h
```diff
@@ -0,0 +1,61 @@
...
+   const ObjectClass super;
+   const GenericData_WriteField writeField;
+   const GenericData_CompareByKey compareByKey;
+} GenericDataClass;
+
+#define As_GenericData(this_)                              ((const GenericDataClass*)((this_)->super.klass))
+
+#define GenericData_compareByKey(g1_, g2_, key_)           (As_GenericData(g1_)->compareByKey ? (As_GenericData(g1_)->compareByKey(g1_, g2_, key_)) : GenericData_compareByKey_Base(g1_, g2_, key_))
+
+int GenericData_compareByKey_Base(const GenericData* g1, const GenericData* g2, ProcessField key);
```
BenBE, line 46: AFAICS the explicit import for `ProcessField` seems missing from this header.

GenericDataList.c
```diff
@@ -0,0 +1,89 @@
...
+
+   Vector_remove(this->genericDataRow, idx);
+   Hashtable_remove(this->genericDataTable, idx);
+
+   this->totalRows--;
+}
+
+GenericData* GenericDataList_getGenericData(GenericDataList* this, GenericData_New constructor) {
+   GenericData* g;
+   g = constructor(this->settings);
```
BenBE, line 54:
```suggestion
   GenericData* g = constructor(this->settings);
```

GenericDataList.h
```diff
@@ -0,0 +1,82 @@
...
+
+   Hashtable* fieldUI;
+
+   bool rebuildFields;
+
+   int offset;
+
+   int id; /* GenericDataList id == offset */
+
+   unsigned int totalRows;
```
BenBE, line 43: Why not use `size_t`?

ScreenManager.c
```diff
@@ -22,6 +22,7 @@ in the source distribution for its full text.
 #include "ProcessList.h"
 #include "ProvideCurses.h"
 #include "XUtils.h"
+#include "GenericDataList.h"
```
BenBE, line 25: Header sort order.

GenericDataList.c
```diff
@@ -0,0 +1,89 @@
...
+*/
+
+#include "config.h" // IWYU pragma: keep
+
+#include "GenericDataList.h"
+
+#include "GenericData.h"
+#include "Hashtable.h"
+#include "Object.h"
+#include "Vector.h"
```
BenBE, line 17:
```suggestion
#include "Object.h"
```

No need to repeat headers already present in the header file.

Settings.c
```diff
@@ -632,6 +638,8 @@ int Settings_write(const Settings* this, bool onCrash) {
    printSettingInteger("all_branches_collapsed", this->screens[0]->allBranchesCollapsed);
 
    for (unsigned int i = 0; i < this->nScreens; i++) {
+         if (this->screens[i]->generic)
+            continue;
```
BenBE, line 642: Check indentation

pcp/PCPGenericData.c
```diff
@@ -0,0 +1,321 @@
...
+                  TimeField(PM_TYPE_U32, atom.ul);
+                  break;
+               } else {
+                  CountField(PM_TYPE_U32, atom.ul);
+                  break;
+               }
+            } else {
+               PrintField("%*u ", gf->value->ul);
+               break;
+            }
```
BenBE, line 190: This looks very repetetive in the following blocks … Only thing changing is the accessed field and the `PM_TYPE_*` specifier. Care to use a `define` for this?

pcp/PCPGenericData.c
```diff
@@ -0,0 +1,321 @@
...
+            break;
+      }
+   }
+}
+
+static int PCPGenericData_compareByKey(const GenericData* v1, const GenericData* v2, int key) {
+   const PCPGenericData* g1 = (const PCPGenericData*)v1;
+   const PCPGenericData* g2 = (const PCPGenericData*)v2;
+
+   if (key < 0) return 0;
```
BenBE, line 267:
```suggestion
   if (key < 0)
      return 0;
```

pcp/PCPGenericData.h
```diff
@@ -0,0 +1,65 @@
...
+in the source distribution for its full text.
+*/
+
+#include "config.h" // IWYU pragma: keep
+
+#include "Object.h"
+#include "GenericData.h"
+#include "Settings.h"
+#include "Platform.h"
+#include "Hashtable.h"
```
BenBE, line 18: Header order

pcp/PCPGenericDataList.c
```diff
@@ -0,0 +1,215 @@
...
+
+   return count + 1;
+}
+
+static int getRowCount(int keyMetric) {
+   int count = PCPMetric_instanceCount(keyMetric);
+   if (count > 0)
+      return count;
+   else
+      return 0;
```
BenBE, line 54:
```suggestion
   return (count > 0) ? count : 0;
```

pcp/PCPGenericDataList.c
```diff
@@ -0,0 +1,215 @@
...
+}
+
+static void allocRows(GenericDataList* gl, int requiredRows) {
+   GenericData* g;
+   int currentRowsCount = Vector_size(gl->genericDataRow);
+
+   if ( currentRowsCount != requiredRows) {
```
BenBE, line 61:
```suggestion
   if (currentRowsCount != requiredRows) {
```

pcp/PCPGenericDataList.c
```diff
@@ -0,0 +1,215 @@
...
+      if (keyInDom != indom) {
+         return 0;
+      }
+   }
+
+   int requiredColumns = getColumnCount(fields);
+   int requiredRows = getRowCount(keyMetric->id);
+
+   /*
+    * alloc */
```
BenBE, line 134:
```suggestion
   // alloc
```

pcp/PCPGenericDataList.c
```diff
@@ -0,0 +1,215 @@
...
+   allocRows(gl, requiredRows);
+   if (requiredRows == 0)
+      return 0;
+
+   allocColumns(gl, requiredColumns, requiredRows);
+
+   /*
+    * fill */
```
BenBE, line 142:
```suggestion
   // fill
```

BenBE (commented):
I think some places that include `config.h` don't need it included. Please re-check.

natoscott:
@smalinux I tried building with --enable-debug and am now seeing an assert failure when I move to the first BPF tab (Opensnoop) - any ideas?  It looks like the Vector_add code is checking for something that has type Process_class (and these BPF entries are not that) ...
```
#0  0x00007ffff7a8ec4c in __pthread_kill_implementation () from /lib64/libc.so.6
#1  0x00007ffff7a3e9c6 in raise () from /lib64/libc.so.6
#2  0x00007ffff7a287f4 in abort () from /lib64/libc.so.6
#3  0x00007ffff7a2871b in __assert_fail_base.cold () from /lib64/libc.so.6
#4  0x00007ffff7a37576 in __assert_fail () from /lib64/libc.so.6
#5  0x000000000041f348 in Vector_set (this=0x713000, idx=idx@entry=0, data_=0x79ad30)
    at Vector.c:339
#6  0x00000000004134de in Panel_set (this=<optimized out>, i=i@entry=0, o=<optimized out>)
    at Panel.c:135
#7  0x000000000040c8f4 in GenericDataList_rebuildPanel (this=this@entry=0x6b08c0)
    at GenericDataList.c:85
#8  0x000000000041a07b in checkRecalculation [...]
(gdb) up 5
#5  0x000000000041f348 in Vector_set (this=0x713000, idx=idx@entry=0, data_=0x79ad30)
    at Vector.c:339
339	   assert(Object_isA(data, this->type));
(gdb) p this->type
$3 = (const ObjectClass *) 0x431700 <Process_class>

```

CommandLine.c
```diff
@@ -350,21 +358,28 @@ int CommandLine_run(const char* name, int argc, char** argv) {
...
-      .pauseProcessUpdate = false,
-      .hideProcessSelection = false,
+      .pauseUpdate = false,
+      .hideSelection = false,
    };
 
-   MainPanel_setState(panel, &state);
+   //MainPanel_setState(panel, &state);
+   panel->state = &state;
+   genericDataPanel->state = &state;
```
natoscott, line 377: Should this be ... ?
```
   MainPanel_setState(panel, &state);
   MainPanel_setState(genericDataPanel, &state);
```

pcp/screens/cgroups
```diff
@@ -0,0 +1,28 @@
...
+
+csystem.heading = STIME
+csystem.caption = Kernel CPU Time
+csystem.width = -8
+csystem.metric = 1000 * rate(cgroup.cpu.stat.system)
+
+cusage.heading = CPU
+cusage.caption = CPU Time
+cusage.width = -8
+cusage.metric = 1000 * rate(cgroup.cpu.stat.usage)
```
natoscott, line 22: @smalinux let's make this a percentage utilization - currently it's showing the sum of the other two (user / kernel) columns (i.e. its essentially repeating existing info there).
```
-cusage.heading = CPU
-cusage.caption = CPU Time
-cusage.width = -8
-cusage.metric = 1000 * rate(cgroup.cpu.stat.usage)
+cusage.heading = CPU%
+cusage.caption = CPU Utilization
+cusage.width = 6
+cusage.metric = 100 * (rate(cgroup.cpu.stat.usage) / hinv.ncpu)
```

pcp/screens/filesys
```diff
@@ -0,0 +1,77 @@
...
+
+[filesys]
+caption = Filesystem
+enabled = true
+
+avail.heading = AVAIL
+avail.caption = avail
+avail.width = -12
+avail.metric = filesys.avail
+avail.scale = true
```
natoscott, line 13: As discussed earlier, let's drop this 'scale' field and make this automatically do the right thing everywhere.

pcp/screens/filesys
```diff
@@ -0,0 +1,77 @@
+#
+# pcp-htop(1) configuration file - see pcp-htop(5)
+#
+
+[filesys]
+caption = Filesystem
+enabled = true
```
natoscott, line 7: For me, this screen doesn't function correctly.  The headings don't seem to match the data displayed.

natoscott (changes requested):
Looking good @smalinux - taking htop to another level.

Do you have a list of known issues?  Here's a start (no surprises here since we've chatted about these, just trying to track 'em so we can tick everything off before any merge) ...
- scaling of values needs to be more dynamic, drop from 'scale' config files
- the Available Columns from Setup should adjust per screen
- filesystem screen is a little confusing
- add a biotop screen as well

CommandLine.c
smalinux: @natoscott I think we need a new panel working in parallel with/instead MainPanel. This is a big part of what I'm talking about so far...
this panel will be minimal and doesn't have `Nice (f7)` or `Kill (f9)` panels [[dito](https://github.com/htop-dev/htop/blob/main/MainPanel.c#L25)]
This will fix a lot of issues, maybe one of them is the issue mentioned above `--enable-debug`, not sure, this part needs surgery ^^" 
but MainPanel has `Class(Process)`: https://github.com/htop-dev/htop/blob/bc22bee6505a4953d20eed96899d929e769a7baa/MainPanel.c#L215

also, @BenBE told me before how to do that: @natoscott I need your help here...
> BenBe: I think it might be worth looking into a direction where the current Panel code is just used as a derived class ProcessListPanel (showing a process list), while most of its code is split into a base class GenericPanel (doing mostly nothing). From that GenericPanel class you could then split another derived class "FancyPanel" for doing all the Fancy stuff ;-)
> 
> Sohaib: Could you explain this more please?
> 
> BenBe: Basically what I was going at was not adopting the current Panel class to also show generic stuff, but to use the existing inheritance stuff so you can just push different kinds of Panels ("Process Lists" vs. "Generic Panels") into the list of tabs handled by the ScreenManager code.
> The way to do this mostly boils down to split the current Panel implementation into the basic "draw the panel" code (extracted from current source BasePanel class) and the code drawing the process list (a new ProcessListPanel class).
> The ScreenManager then instead of listing plain Panels would list BasePanels, which could be ProcessListPanels (working like they do now) or GenricStuffPanels (doing something generic).

Action.c
```diff
@@ -504,8 +504,8 @@ static Htop_Reaction actionRedraw(ATTR_UNUSED State* st) {
    return HTOP_REFRESH | HTOP_REDRAW_BAR;
 }
 
-static Htop_Reaction actionTogglePauseProcessUpdate(State* st) {
-   st->pauseProcessUpdate = !st->pauseProcessUpdate;
+static Htop_Reaction actionTogglePauseUpdate(State* st) {
+   st->pauseUpdate = !st->pauseUpdate;
```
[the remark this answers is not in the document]
smalinux: Done! Thanks!

GenericData.h
smalinux: Done! Thanks!

BenBE:
@smalinux The rename changes from your second commit are partially included in the first commit (e.g. look for `pauseUpdate` in the first commit). I think putting that refactoring commit in front and ensuring the changes there are complete is most clear to get things work out correctly. A quick test is to try and compile each commit individually and see if it builds properly.

CommandLine.c
natoscott: @BenBE any chance you can help @smalinux here with some sample code changes to get him started on this aspect?  thanks!

BenBE: Will have to wait for next week; not in town tomorrow and the day after (giving a talk for a small conference). Changes-wise it boils down to creating inherited classes similar to what can be seen with `ProcessList` and `<Platform>ProcessList` with the Panel managing the `ProcessList` stuff getting most of the current process list draw code. Extra points when rendering the header/list items is split in a way so that for the generic tabs and the process list the actual rendering code is the same, but only creating/filling the content buffers differs. As said, I can elaborate next week …

pcp/screens/filesys
natoscott: I have a branch with this working now, will post an update soon.

pcp/PCPGenericData.c
natoscott: This was resolved with the refactoring awhile back that improved the generic value scaling algorithm used.  Code referred to here now lives in PCPDynamicColumn.c FWIW.

pcp/screens/filesys
natoscott: The generic value scaling work is now merged, marking this one resolved.

natoscott:
This functionality has now been incorporated into PR #1243 (after fairly substantial reworking).

Issue #1038, opened 2022-06-08
Filter function does not work with non-ASCII characters


Eucyon:
To reproduce, press F4 and insert e.g. ä, ö or ü.

Thank you!

fasterit:
The `getch()` we are using has no idea about how to process multi-byte input properly.
See https://github.com/ThomasDickey/dialog-snapshots/blob/7289c6063b40654f9818372cdbb9d17378ccc46a/ui_getc.c#L333 for an example of how to manage this better.
Code from `dialog` is LGPL v2.1 so compatible. Hint, hint.

PR #1105, opened 2022-10-13
Accept wide chars in filter input (#1038)

adsr:
(For hacktoberfest)

Notably this doesn't handle wcwidth!=1 or extended grapheme clusters, but you can filter for your emoji process now.


Panel.c
```diff
@@ -503,5 +506,46 @@ int Panel_getCh(Panel* this) {
...
+      ch = wchbuf[wchbuf_i++];
+   } else {
+      last_wkey = 0;
+      wchrv = get_wch(&wkey);
+      wch = (wchar_t)wkey;
+      if (wchrv == OK) {
+         last_wkey = wkey;
+         memset(&wst, 0, sizeof(wst));
+         wchbuf_i = 0;
+         wchbuf_len = (int)wcrtomb(wchbuf, wch, &wst);
```
BenBE, line 530: No need for this cast, cf. comment above.

Panel.c
```diff
@@ -503,5 +506,46 @@ int Panel_getCh(Panel* this) {
...
+         if (wchbuf_len < 0) {
```
BenBE, line 531: See return value for `wcrtomb` at https://en.cppreference.com/w/c/string/multibyte/wcrtomb; should check for `(size_t)-1`.

BenBE (changes requested):
Could you please merge the commits into one?

adsr:
Added your fixes and squashed. Thanks @BenBE.

IncSet.c
```diff
@@ -155,6 +157,28 @@ static bool IncMode_find(const IncMode* mode, Panel* panel, IncMode_GetPanelValu
...
+      return 1;
+   }
+
+   char mb[MB_LEN_MAX];
+   int clen = wcrtomb(mb, w[wlen - 1], NULL);
+   if (clen < 0) {
+      return 1;
+   } else if (clen > slen) {
+      clen = slen;
+   }
```
Explorer09, line 177: `clen` should be of type `size_t`. It would be better if `slen` is also a `size_t` type.

IncSet.c
```diff
@@ -12,6 +12,8 @@ in the source distribution for its full text.
 #include <ctype.h>
 #include <string.h>
 #include <stdlib.h>
+#include <limits.h>
```
Explorer09, line 15: The include lines should be ordered alphabetically for headers in the same group (standard libc headers in this case).

FunctionBar.c
```diff
@@ -92,6 +92,14 @@ int FunctionBar_draw(const FunctionBar* this) {
    return FunctionBar_drawExtra(this, NULL, -1, false);
 }
 
+static size_t FunctionBar_safeLen(const char *s) {
+   size_t slen = mbstowcs(NULL, s, 0);
+   if (slen == (size_t)-1) {
+      slen = strlen(s);
```
Explorer09, line 98: I would like some explanation on the logic. `mbstowcs` returns `(size_t)-1` when encountered an illegal byte, but why fall back to `strlen`? How is this function meant to be used?

Panel.c
```diff
@@ -503,5 +506,46 @@ int Panel_getCh(Panel* this) {
 #ifdef HAVE_SET_ESCDELAY
    set_escdelay(25);
 #endif
-   return getch();
+
+   int ch;
+#ifdef HAVE_LIBNCURSESW
+   static char wchbuf[MB_LEN_MAX];
+   static size_t wchbuf_len = 0, wchbuf_i = 0;
```
Explorer09, line 513: How about renaming the buffer to `mbbuf`?

FunctionBar.c
adsr: This function implements "get multi-byte string length, otherwise byte length". It is used below to position the cursor on the filter input field. In the happy path the string will have valid encoding, `mbstowcs` will succeed, and we'll position the cursor in the right place. If `s` is garbage, byte length is the only sensible fallback. `safeLen` is maybe not the best name.

Thanks for the review. I'll add your fixes shortly.

Explorer09: @adse So this function is meant to be getting a "display width"?
Unfortunately there is a complexity here. Not every Unicode character occupies one column in the terminal. CJK characters would occupy two. So assuming the "display width" of the string is the same as number of code points in the string (i.e. the return value of `mbstowcs`) is wrong. The best approach as far as I can find is to use [`wcwidth`](https://man7.org/linux/man-pages/man3/wcwidth.3.html) and [`wcswidth`](https://man7.org/linux/man-pages/man3/wcswidth.3.html).

IncSet.c
```diff
@@ -155,6 +157,28 @@ static bool IncMode_find(const IncMode* mode, Panel* panel, IncMode_GetPanelValu
    }
 }
 
+static int IncSet_lastCharLen(const char *s, size_t slen) {
+   if (slen < 1) {
+      return 0;
+   }
+
+   wchar_t w[INCMODE_MAX + 1];
+   size_t wlen = mbstowcs(w, s, slen);
```
Explorer09, line 166: The size limit of `mbstowcs` is number of wide characters (code points), not bytes.
The argument is meant to prevent buffer overflow of `w`, not the limit the buffer length of `s` .
Perhaps using [`mbrtowc`](https://en.cppreference.com/w/c/string/multibyte/mbrtowc) with a loop instead would do what you want. The size limit of `mbrtowc` is indeed in bytes and it limits the length of `s`.

adsr: Good call. Would it not suffice to change the 3rd param to `INCMODE_MAX + 1` and leave the rest? I believe the calling function ensures `s` is null-terminated.

FunctionBar.c
adsr: You're right. (I wrote this in the PR description.) This was only meant as an incremental improvement with proper handling and rendering as a follow-up.

adsr (commented):
Thanks for the reviews.

IncSet.c
Explorer09: @adsr
1. You need a NUL character, so the parameter should be `INCMODE_MAX` without `+1`, and you need this explicit line `w[INCMODE_MAX] = 0;` to ensure the wide string is null-terminated.

2. If you just the 3rd parameter of `mbstowcs` to `INCMODE_MAX`, then it would make no sense to pass in `slen` parameter to the function.

Explorer09: @adsr 
3. I've just read the caller code of `IncSet_handleKey`:

```
  } else if (ch == KEY_BACKSPACE || ch == 127) {
      if (mode->index > 0) {
         mode->index--;
         mode->index -= IncSet_lastCharLen(mode->buffer, mode->index);
         mode->buffer[mode->index] = 0;
```

`mode->index` is the (byte) index to the last character of the filter string. With `mbstowcs` you cannot convert part of the multi-byte string with the given number of bytes. To convert part of the string you have a few options:

(a) Use `mbtowc` or `mbrtowc` with a loop as mentioned in my last comment.
(b) Duplicate `mode->buffer`, write a NUL byte in the right position, and pass this buffer to `mbstowcs`. I don't like this approach as you need to handle when the buffer allocation fails, and you need to `free` the buffer afterward.
(c) Change `mode->index` so it becomes code point index rather than byte index. This requires much more work and more risk breaking something.

adsr: > With mbstowcs you cannot convert part of the multi-byte string with the given number of bytes.

As written, I don't think we can count on any of these `mb*` functions working unless we can guarantee the buffer is validly encoded, which we are not doing (yet). Therefore I don't see a big difference between using `mbstowcs` or looping with `mbtowc`. They both can fail and then you have employ some sort of fallback logic.

For example, consider the string "testλ" except let's remove the last byte, so `0x74 65 73 74 ce`. In this case the `mbstowcs` in `lastCharLen` will return `-1`, so the function will return `1` leading to use deleting the `0xce` byte. This is reasonable. Looping with `mbtowc`, we'd similarly fail on the last byte of the string. If the invalid encoding was in the middle of the string instead of the end, the behavior could be different, but it's not clear to me what the right thing to do there is. The end-user would have to be trying pretty hard to make that happen.

For the happy path, consider the full string "testλ" `0x74 65 73 74 ce bb`. In this case the `mbstowcs` succeeds and `w` holds ['t', 'e', 's', 't', 'λ', 0]. We measure the byte length of  `"λ"` using `wcrtomb` and return 2, leading us to delete the `"λ"`. Also good.

The proper solution is a lot more work as you point out. IMO what's here will work for most cases in what is already an edge case (non-ASCII process names).

Let me know if I've missed something. Thanks for the reviews @Explorer09.

Explorer09: @adsr How about this: If there is an invalid byte encoding in the middle of the string, cut the invalid byte and everything after it. You shouldn't trust the string input by the user, so there _would_ be a sanitization routine somewhere. Cutting just the last byte might not be sufficient. Consider a 4-byte UTF-8  character `F0 9F 8D 8E` (U+1F34E) and one last byte is Iost and the string becomes `F0 9F 8D`, a backspace should remove all three invalid bytes so that the rest of the sequence is valid.

adsr: Sure that also works. I renamed the function to `IncSet_backspaceLen` and changed it to loop with `mbtowc` as you describe.

Panel.c
```diff
@@ -13,6 +13,7 @@ in the source distribution for its full text.
 #include <stdlib.h>
 #include <string.h>
 #include <strings.h>
+#include <limits.h>
```
BenBE, line 16: Header order.

Panel.c
```diff
@@ -503,5 +506,46 @@ int Panel_getCh(Panel* this) {
 #ifdef HAVE_SET_ESCDELAY
    set_escdelay(25);
 #endif
-   return getch();
+
+   int ch;
```
BenBE, line 510: Given this value is more or less directly returned everywhere anyway, what about dropping `ch` entirely and instead just return the value we assign to it directly. The only place where this is not directly obvious to do is the case without wide char support, where things would need to be done as `return last_wkey = getch();`. The upshot of this is that the compiler will actually warn if any path through this functions accidentally missed that return value. Also this would allow to flatten the code nesting quite a bit and thus make the code easier to follow IMHO.

adsr: Incorporated this in latest. Thanks @BenBE.

FunctionBar.c
adsr: I renamed `FunctionBar_safeLen` to `FunctionBar_displayWidth` which sums `wcwidth` of each wchar and falls back on `strlen` on invalid encoding.

Explorer09: I need some time to review this, as I don't think falling back to `strlen` is a good idea. Ideally the string passed to `FunctionBar_displayWidth` should have been sanitized so that no invalid byte would exist (if there were invalid bytes or by sequences they should be replaced with U+FFFD as part of the sanitization). The other issue I can think of is that `wcwidth` might return -1 for non-printable characters and you need to handle that as a special case (maybe treat them as zero width?).

adsr: Is sanitizing the process name the right thing to do? The kernel doesn't put any encoding restrictions on process names, so users should ideally be able to filter by arbitrary binary as well.

For that reason, I'd vote for not sanitizing filter strings, and rendering anything non-printable or invalidly encoded as "?" (or U+FFFD). htop seems to do this already when printing comm lines. (Try `htop  -F $'\t'`.) `top` and `ps` both do something similar to what I'm describing.

`htop  -F $'\n'` is presently buggy. This will fix that as well.

I can implement this if people agree. It'll require changing some more of `FunctionBar.c`.

Explorer09:
I have a concern about `Panel_getCh` function. The API was designed to return a byte despite us implementing multi-byte support for it. And the function maintains a state (`mbbuf`, `mbbuf_i` and `mbbuf_len`) to make it work with multi-byte. I wonder if it is possible to avoid the state by changing the function to return the multi-byte sequence directly.
I'm thinking of a function prototype like this:
```c
int Panel_readMultiByteChar(Panel* this, char *mbBuf)
// mbBuf must have at least (MB_LEN_MAX + 1) bytes. The function outputs the
// multi-byte sequence in mbBuf and the sequence is terminated by a 0x00 byte.
```

This API would be more future-proof as it no longer returns a byte at the middle of the sequence, which could lead to a (potential) synchronization problem (I know this synchronization problem won't affect UTF-8 though).

FunctionBar.c
Explorer09: @adsr By "sanitizing" I mean filter out any Unicode control character that may come from user input, converting them into U+FFFD characters _or_ print them out in some form of escape sequences. You do need to filter that for display purposes (something that htop didn't implement yet, unfortunately). Otherwise you could see a spoofing attack like in [this article](https://www.vadesecure.com/en/blog/how-hackers-are-using-a-20-year-old-text-trick-to-phish-microsoft-365-users). I'm not sure if the Filter function of htop is going to support escape sequences, but accepting user input (from `getch` or `get_wch`) without some kind of checking is generally a bad idea.

Specifically, at least, never allow non-printable characters from `get_wch` input:
```
wint_t wkey;
int wchrv = get_wch(&wkey);
if (wchrv == OK) {
   if (!iswprint(wkey)) {
      // Disallow. Or convert this to U+FFFD
      wkey = (wint_t )0xFFFD;
   }
}
```

adsr:
@BenBE Could you add the `hacktoberfest-accepted` label? It takes a few days for it to register and I'd like to get credit for it if you think it's worthy of course. Sorry to bother.

FunctionBar.c
adsr: > You do need to filter that for display purposes (something that htop didn't implement yet, unfortunately).

I agree that filtering on display is needed, however I disagree with your code snippet (filtering input). As I stated, non-printables and invalid encodings are allowed in process names, so disallowing them as input would prevent users from matching such processes. Of course this is a very narrow edge case and naturally I will defer to the maintainers, but that is my view.

@BenBE or someone else want to weigh in?

Explorer09: > I agree that filtering on display is needed, however I disagree with your code snippet (filtering input). As I stated, non-printables and invalid encodings are allowed in process names, so disallowing them as input would prevent users from matching such processes. Of course this is a very narrow edge case and naturally I will defer to the maintainers, but that is my view.

My idea is that the invalid characters are disallowed, in raw, from user input. They may only be input via escape sequences.

BenBE: @Explorer09 Can you elaborate on the "input via escape sequences" part? Currently the filter/search functions don't do any escape sequence processing. If this was to be added, that'd be a different feature.

Also on the part of invalid encodings: As those are possible to occur as part of any process name/information we should not hinder filtering/searching for them. Thus if you limit inputs to valid character sequences only you'll also have to allow the user to specify if they wanted to look for an invalid encoding (intentionally).

Explorer09: @BenBE An example is embedded newlines in filenames, if you are adding that to the filter string, you should probably type it as `\n` and not the Enter key as the latter would mean something else. Likewise, you will type `\033` for the ASCII ESC character in the filename and not the Escape key. That's the filtering I was talking about.

You will need to filter this from the keyboard input (within the GUI), but not from the `-F` option argument (the escape sequence of the latter is processed by the shell).

BenBE: @Explorer09 Exactly. And this handling is currently not implemented. Same for when you need to look for `Hallö Wörld` in a process that uses ISO-8859-1 instead of UTF-8 …

IncSet.c
BenBE: @Explorer09 Is this one resolved?

Explorer09: It seems that the function has been reworked (to `IncSet_backspaceLen`) and no longer uses `wcrtomb`. I think I can close this part of the review as "no longer applicable".

Explorer09: P.S. I mean, mark this as resolved. (I can't close this by myself).

adsr:
> Also there's the `last_wkey` global I don't quite like. How much work would it take to avoid it?

I think the main alternative is to modify `Panel_getCh` and its callers to use `wint_t` instead of `int` or `char` as @Explorer09 suggested. We might be able limit the scope of the change by introducing some proxy functions that call the existing functions in a loop, e.g., `InfoScreen_run` invokes `Panel_getWch` and then `InfoScreen_run_inner` in a loop with the individual bytes of the wide char. IMO this would be uglier than the current implementation, so I'd vote either for changing it all or sticking with the current implementation.

Panel.c
```diff
@@ -503,5 +506,39 @@ int Panel_getCh(Panel* this) {
...
+   wchar_t wch = (wchar_t)wkey;
+
+   if (wchrv == OK) {
+      mbbuf_i = 0;
+      mbbuf_len = wcrtomb(mbbuf, wch, NULL);
+      if (mbbuf_len == (size_t)-1) {
+         mbbuf_len = 0;
+         return wkey;
+      }
+      last_wkey = wkey;
```
Explorer09, line 529: I tried to re-review this patch and I have some more comments on this:
1. I think we should upgrade the `Panel_getCh` function to `Panel_getWch` to avoid a state buffer in this function, making things more fragile.
2. Since this is the primary function we read user's keyboard input, if we want to sanitize user input, this would be the best place to do.
3. However, I've changed my mind on filtering all unprintable characters. Instead, filter only (a.) invalid characters, that is, when `wctomb` would fail to convert the character, and (b.) newlines of all kind. Replace the character with U+FFFD if `wctomb` fails. Don't replace other Unicode characters because user might input it through a paste from the clipboard to the terminal emulator, and those characters should be intact as much as possible.

IncSet.c
```diff
@@ -155,6 +157,23 @@ static bool IncMode_find(const IncMode* mode, Panel* panel, IncMode_GetPanelValu
...
 }
 
+static int IncSet_backspaceLen(const char *s, size_t slen) {
+   size_t i = 0;
+   int mblen = 0;
+
+   mbtowc(NULL, 0, 0);
+
+   while (i < slen) {
+      mblen = mbtowc(NULL, s + i, slen - i);
```
Explorer09, line 167: Re-reviewing this after I have more knowledge with the htop codebase (especially the `RichString` part).

As `RichString.c` is already using `mbrtowc` for wide character conversion, it's better to stick with `mbrtowc` here rather than downgrade to `mbtowc`. With `mbrtowc`, the string length can be properly addressed as `size_t` type.

Explorer09: Also. It's good to rename the local variable `mblen` to `mbbuf_len` or something to avoid name collision with [the standard library function](https://man7.org/linux/man-pages/man3/mblen.3.html). (Name collision is not an error in C, but it can produce noise when other programmers `grep` for the keyword in code.)

FunctionBar.c
```diff
@@ -92,6 +92,28 @@ int FunctionBar_draw(const FunctionBar* this) {
...
 
+static size_t FunctionBar_displayWidth(const char *s) {
+   size_t slen = strlen(s);
+   size_t width = 0, i = 0;
+
+   mbtowc(NULL, 0, 0);
+
+   while (i < slen) {
+      wchar_t wch;
+      int mblen = mbtowc(&wch, s + i, slen - i);
```
Explorer09, line 103: Ditto. (Use `mbrtowc` here rather than downgrade to `mbtowc`. With `mbrtowc`, the string length can be properly addressed as `size_t` type.)

IncSet.c
BenBE: Shadowing should also be avoided where possible, as this may introduce hard-to-notice issues.

FunctionBar.c
```diff
@@ -92,6 +92,28 @@ int FunctionBar_draw(const FunctionBar* this) {
...
+      if (mblen <= 0) {
+         return slen;
+      }
+      i += mblen;
+      int w = wcwidth(wch);
+      if (w > 0) {
+         width += w;
```
Explorer09, line 110: Don't leave the case of `wcwidth(wch) < 0` out of this.

`wcwidth(wch) < 0` can happen when `!iswprint(wch)` and that means `wch` is a C0 or C1 control character, U+2028 (line separator) or U+2029 (paragraph separator). This is according to [ISO/IEC 30112](https://www.open-std.org/jtc1/sc35/wg5/docs/30112d12.pdf) (look for characters in the `cntrl` category).

PR #1107, opened 2022-10-16
Initial notes on cross compiling for Windows

dnabre:
main document in librewriter
 markdown version of same

 scripts referred in above, {env,build_htop,build_ncurses}.sh

 current errors when building (htop_build_error.log)

 configure.ac that doesn't check for sys/utsname.h


dnabre:
Didn't mean to create this request here. Meant to do it on my own fork. 

Looking up how to remove.

Issue #1108, opened 2022-10-19
Only CPU bars on a high CPU machine


lucasew:
I am having a first world problem here.

I have access to a machine with only 256 cores. The problem is that when I run HTOP on it it's only CPU bars.

I can't open the HTOP settings by pressing F2 because it doesn't appear at all.

What can I do in this situation xD

![Captura de tela_2022-10-19_19-17-59](https://user-images.githubusercontent.com/15693688/196815332-91571b3e-e6f5-47e9-8ef7-e490a4b002c6.png)

BenBE:
Recent versions of `htop` (IIRC starting with 3.0.5) automatically detect if there are many CPU cores and default to multiple columns (up to 8) or just one overall CPU meter. To reset to the default settings you can remove `~/.config/htop/htoprc`

If you can't update you can:
1. Reduce your terminal font size
2. Hit <kbd>F2</kbd> to enter the setup screen
3. Select the Meters configuration
4. Remove the CPU meters in the left and right columns
5. Hit <kbd>F10</kbd> to save the configuration
6. Restore your terminal font size.

If you can't go that route, there's also the chance of editing the configuration file by hand (note the remarks at the top of that file). Just adjust the entries for `column_meters_N` and `column_meter_modes_N` (The space-separated items to the right of the `=` are paired: Thus if removing the X<sup>th</sup> item in `column_meters_N` be sure to also remove the X<sup>th</sup> item from `column_meter_modes_N`).

OlafvdSpek:
> Recent versions of `htop` (IIRC starting with 3.0.5) automatically detect if there are many CPU cores and default to multiple columns (up to 8) or just one overall CPU meter.

> What's new in version 3.1.0

> * Improve default display for systems with many CPUs

lucasew:
> > What's new in version 3.1.0
> 
> > * Improve default display for systems with many CPUs

This explains everything. The OS is Ubuntu 20.04.3.

```
$ htop --version
htop 2.2.0 - (C) 2004-2019 Hisham Muhammad
Released under the GNU GPL.
```

lucasew:
![Captura de tela_2022-11-04_10-39-06](https://user-images.githubusercontent.com/15693688/199986266-98f04c65-5e63-4894-b30d-c6b9d975eb46.png)

Tested with the Alpine Linux HTOP using a Docker container. That's the best I can get without hiding all the cores. It hides the "each core block" and only shows the average usage by default.

[...]

PR #1110, opened 2022-10-20
Toggle the header meters with pound key

cgzones:
Show/hide the header meters with the pound ('#') key.  Useful in cases where the header is too large and occupies essential parts of the screen, especially in settings (see #1108).

It is only stored as a runtime state, not a persistent setting; to remove the header permanently one can delete all active meters.

### DIFF
--- a/ScreenManager.c
+++ b/ScreenManager.c
@@ -24,7 +24,7 @@ in the source distribution for its full text.
 #include "XUtils.h"
 
 
-ScreenManager* ScreenManager_new(Header* header, const Settings* settings, const State* state, bool owner) {
+ScreenManager* ScreenManager_new(Header* header, const Settings* settings, State* state, bool owner) {
    ScreenManager* this;
    this = xMalloc(sizeof(ScreenManager));
    this->x1 = 0;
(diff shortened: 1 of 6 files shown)


BenBE (changes requested):

BenBE (approved):
LGTM.

Not sure if `hideMeters` should be explicitly initialized when `state` is created.

ScreenManager.c
```diff
@@ -24,7 +24,7 @@ in the source distribution for its full text.
 #include "XUtils.h"
 
 
-ScreenManager* ScreenManager_new(Header* header, const Settings* settings, const State* state, bool owner) {
+ScreenManager* ScreenManager_new(Header* header, const Settings* settings, State* state, bool owner) {
```
Explorer09, line 27: What's the reason of removing this `const`?

cgzones: The ScreenManager now changes the state in the event of the pound key pressed:
```c
case '#':
         this->state->hideMeters = !this->state->hideMeters;
         ScreenManager_resize(this);
         force_redraw = true;
```

The action code only works in the MainPanel, not in settings etc..

BenBE:
@cgzones re my last comment: Is that initialization missing or covered by `xCalloc` of the state struct?

cgzones:
It was implicitly initialized to 0 via the C99 struct designator expression; now explicit.

BenBE:
Is it intentionally limited to the setup screen?

Would prefer this to also work on the main screen (currently does not do anything there for me).

cgzones:
The action flags were incorrect (`HTOP_RESIZE` was missing), but I tested it before https://github.com/htop-dev/htop/commit/d83948ed09039864cdaf6066a1bd64476211ecaa, when it regardless worked.

PR #1111, opened 2022-10-20
CI: add GitHub CodeQL action 

cgzones:
LGTM.com will shutdown in December 2022, use CodeQL instead.


BenBE:
Shouldn't this also remove LGTM.com as one of the tasks?

cgzones:
LGTM.com is enabled via GitHub apps in the repository settings, not via any in-repository files.

PR #1117, opened 2022-10-22
Do not render a margin around the header if there are any meters

BenBE:


Header.c
```diff
@@ -286,10 +288,17 @@ int Header_calculateHeight(Header* this) {
       }
       maxHeight = MAXIMUM(maxHeight, height);
    }
+
+   if (maxHeight == pad) {
+      maxHeight = 0;
+      this->pad = 0;
+   }
```
cgzones, line 295: Maybe instead of assigning `this->pad` in different locations use an else branch here:
```c
} else
    this->pad = pad;
```
?

cgzones (approved):

Issue #858, opened 2021-10-27
Option to exclude HugePages from memory


SunRed:
Commit 04871a2d04361517f2e539810cb11e9e9898f0d0 reverted a change I actually welcomed and didn't know would be quite handy. I use htop on a kvm host where I use hugepages for the vms and having the memory bar filled up almost entirely with huge pages isn't quite useful, also since there is a separate bar for huge pages I can see the currently used and allocated hugepages in so that it's quite redundant information.
Therefore it would be quite useful to actually have a simple option in the display options to exclude hugepages from memory. If this is a welcome change I could also just submit a pull request if you don't mind. :)

PR #1123, opened 2022-10-31
Add toggle to exclude hugepages from memory bar

SunRed:
This implements #858 what I completely forgot about. One addition one could do would be coloring the hugepages part of the memory bar when not excluding hugepages in a different color.


cgzones:
Since the handling of hugepages is Linux only, maybe the new setting should also be only available on Linux.

PR #1127, opened 2022-11-07
DiskIOMeter: convert to per second & use complete units

eworm-de:
The value was in per-interval, let's convert to per-second and show complete units. This brings the meter in line with the NetworkIOMeter.

### DIFF
--- a/DiskIOMeter.c
+++ b/DiskIOMeter.c
@@ -106,7 +106,7 @@ static void DiskIOMeter_updateValues(Meter* this) {
    char bufferRead[12], bufferWrite[12];
    Meter_humanUnit(bufferRead, cached_read_diff, sizeof(bufferRead));
    Meter_humanUnit(bufferWrite, cached_write_diff, sizeof(bufferWrite));
-   snprintf(this->txtBuffer, sizeof(this->txtBuffer), "%siB/s %siB/s %.1f%%", bufferRead, bufferWrite, cached_utilisation_diff);
+   snprintf(this->txtBuffer, sizeof(this->txtBuffer), "r:%siB/s w:%siB/s %.1f%%", bufferRead, bufferWrite, cached_utilisation_diff);
 }
 
 static void DiskIOMeter_display(ATTR_UNUSED const Object* cast, RichString* out) {
--- a/DiskIOMeter.c
+++ b/DiskIOMeter.c
@@ -64,7 +64,7 @@ static void DiskIOMeter_updateValues(Meter* this) {
 
       if (data.totalBytesRead > cached_read_total) {
          diff = data.totalBytesRead - cached_read_total;
-         diff /= 1024; /* Meter_humanUnit() expects unit in kilo */
+         diff /= ONE_K; /* convert to KiB/s */
          cached_read_diff = (uint32_t)diff;
       } else {
          cached_read_diff = 0;
@@ -73,7 +73,7 @@ static void DiskIOMeter_updateValues(Meter* this) {
 
       if (data.totalBytesWritten > cached_write_total) {
          diff = data.totalBytesWritten - cached_write_total;
-         diff /= 1024; /* Meter_humanUnit() expects unit in kilo */
+         diff /= ONE_K; /* convert to KiB/s */
          cached_write_diff = (uint32_t)diff;
       } else {
          cached_write_diff = 0;
--- a/DiskIOMeter.c
+++ b/DiskIOMeter.c
@@ -106,7 +106,7 @@ static void DiskIOMeter_updateValues(Meter* this) {
    char bufferRead[12], bufferWrite[12];
    Meter_humanUnit(bufferRead, cached_read_diff, sizeof(bufferRead));
    Meter_humanUnit(bufferWrite, cached_write_diff, sizeof(bufferWrite));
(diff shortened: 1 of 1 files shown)


DiskIOMeter.c
```diff
@@ -64,7 +64,8 @@ static void DiskIOMeter_updateValues(Meter* this) {
 
       if (data.totalBytesRead > cached_read_total) {
          diff = data.totalBytesRead - cached_read_total;
-         diff /= 1024; /* Meter_humanUnit() expects unit in kilo */
+         diff /= ONE_K; /* Meter_humanUnit() expects unit in kilo */
+         diff = (1000 * diff) / passedTimeInMs; /* convert to per second */
```
BenBE, line 68: Possible (minor) error due to order of operations.

The first division drops 10 bits of precision, then things are shifted (no additional information) with the multiplication by 1000, and another ~1 bit is removed (due to the factor `1000/passedTimeInMs`). It might be worth swapping the two divisions, as in that case the division by `passedTimeInMs` would not introduce any error that would be enlarged by scaling the number in the second step (actually the second step would always scale by a factor <1, thus shrinking the error introduced in the first step.

No need to change; just food for thought/discussion.

DiskIOMeter.c
```diff
@@ -73,7 +74,8 @@ static void DiskIOMeter_updateValues(Meter* this) {
 
       if (data.totalBytesWritten > cached_write_total) {
          diff = data.totalBytesWritten - cached_write_total;
-         diff /= 1024; /* Meter_humanUnit() expects unit in kilo */
+         diff /= ONE_K; /* Meter_humanUnit() expects unit in kilo */
+         diff = (1000 * diff) / passedTimeInMs; /* convert to per second */
```
BenBE, line 78: Same reasoning as above.

BenBE (approved):

DiskIOMeter.c
OlafvdSpek: > actually the second step would always scale by a factor <1

What if passedTimeInMs >= 1000?

> It might be worth swapping the two divisions

Why not just `diff = (1000 * diff) / (ONE_K * passedTimeInMs)`?

DiskIOMeter.c
```diff
@@ -73,7 +74,8 @@ static void DiskIOMeter_updateValues(Meter* this) {
...
          cached_write_diff = (uint32_t)diff;
```
OlafvdSpek, line 79: What if diff >= 4 GiB?

DiskIOMeter.c
BenBE: Given `passedTimeInMs` is always around 500ms or greater, there's no real precision to be lost by the second operation, which scales by 1000 / 1024 (which is <1). The error introduced by the initial division is thus scaled down.

Doing the calculation in one step works too, provided that (`diff` being `uint64_t`) it's unlikely we get a diff of ~16EiB in one step.

DiskIOMeter.c
BenBE: AFAICS `diff` is `uint64_t`, thus we'd need to be `diff > ~16EiB` for that overflow …

OlafvdSpek: `diff` is per second (on the last line), so the `uint32_t` cast would truncate it at 4 GiB/s wouldn't it?

eworm-de: Actually the value is in kilo. So the truncation would happen with 4TiB/s, no?

OlafvdSpek: Ah, my bad, it's in KiB/s..

eworm-de:
Force-pushed commits to swap calculation.

DiskIOMeter.c
OlafvdSpek: Why no single division?

eworm-de: Trying to keep the code in sync with network Io meter. Any opinions from upstream?

DiskIOMeter.c
```diff
@@ -64,7 +64,8 @@ static void DiskIOMeter_updateValues(Meter* this) {
 
       if (data.totalBytesRead > cached_read_total) {
          diff = data.totalBytesRead - cached_read_total;
-         diff /= 1024; /* Meter_humanUnit() expects unit in kilo */
+         diff = (1000 * diff) / passedTimeInMs; /* convert to per second */
+         diff /= ONE_K; /* Meter_humanUnit() expects unit in kilo */
```
BenBE, line 68:
```suggestion
         diff /= passedTimeInMs; /* convert to B/ms */
         diff = (1000 * diff) / ONE_K; /* convert to KiB/s */
```

DiskIOMeter.c
```diff
@@ -73,7 +74,8 @@ static void DiskIOMeter_updateValues(Meter* this) {
 
       if (data.totalBytesWritten > cached_write_total) {
          diff = data.totalBytesWritten - cached_write_total;
-         diff /= 1024; /* Meter_humanUnit() expects unit in kilo */
+         diff = (1000 * diff) / passedTimeInMs; /* convert to per second */
+         diff /= ONE_K; /* Meter_humanUnit() expects unit in kilo */
```
BenBE, line 78:
```suggestion
         diff /= passedTimeInMs; /* convert ti B/ms */
         diff = (1000 * diff) / ONE_K; /* convert to KiB/s */
```

BenBE (commented):
Same as with the other PR. :)

PR #1132, opened 2022-11-17
Refresh settings

BenBE:
This works around an issue where updated display settings are not considered because the display strings in the `Command` column are cached.

This also allows to keep following when changing the Merged Command view or display of full path names.

### DIFF
--- a/Action.c
+++ b/Action.c
@@ -213,26 +213,36 @@ static Htop_Reaction actionSortByTime(State* st) {
 
 static Htop_Reaction actionToggleKernelThreads(State* st) {
    st->settings->hideKernelThreads = !st->settings->hideKernelThreads;
+   st->settings->lastUpdate++;
+
    return HTOP_RECALCULATE | HTOP_SAVE_SETTINGS | HTOP_KEEP_FOLLOWING;
 }
 
 static Htop_Reaction actionToggleUserlandThreads(State* st) {
    st->settings->hideUserlandThreads = !st->settings->hideUserlandThreads;
+   st->settings->lastUpdate++;
+
    return HTOP_RECALCULATE | HTOP_SAVE_SETTINGS | HTOP_KEEP_FOLLOWING;
 }
 
 static Htop_Reaction actionToggleRunningInContainer(State* st) {
    st->settings->hideRunningInContainer = !st->settings->hideRunningInContainer;
+   st->settings->lastUpdate++;
+
    return HTOP_RECALCULATE | HTOP_SAVE_SETTINGS | HTOP_KEEP_FOLLOWING;
 }
 
 static Htop_Reaction actionToggleProgramPath(State* st) {
    st->settings->showProgramPath = !st->settings->showProgramPath;
+   st->settings->lastUpdate++;
+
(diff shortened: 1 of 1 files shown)


cgzones:
It seems `HTOP_KEEP_FOLLOWING` should almost be used everywhere, e.g. in `actionHigherPriority()`, `actionLowerPriority()`, `actionLsof()`, `actionShowLocks()`, `actionStrace()`, `actionRedraw()`, `actionTogglePauseProcessUpdate()`, `actionUntagAll()`, `actionTagAllChildren()`, `actionShowEnvScreen()`.

Maybe it is worth to invert the logic: keep following by default and only stop for a couple of actions?

Action.c
```diff
@@ -213,26 +213,36 @@ static Htop_Reaction actionSortByTime(State* st) {
 
 static Htop_Reaction actionToggleKernelThreads(State* st) {
    st->settings->hideKernelThreads = !st->settings->hideKernelThreads;
+   st->settings->lastUpdate++;
```
cgzones, line 216: Is this necessary? What does change about the merged command line on toggle?

Action.c
```diff
@@ -213,26 +213,36 @@ static Htop_Reaction actionSortByTime(State* st) {
...
+
    return HTOP_RECALCULATE | HTOP_SAVE_SETTINGS | HTOP_KEEP_FOLLOWING;
 }
 
 static Htop_Reaction actionToggleUserlandThreads(State* st) {
    st->settings->hideUserlandThreads = !st->settings->hideUserlandThreads;
+   st->settings->lastUpdate++;
```
cgzones, line 223: dito

Action.c
```diff
@@ -213,26 +213,36 @@ static Htop_Reaction actionSortByTime(State* st) {
...
+
    return HTOP_RECALCULATE | HTOP_SAVE_SETTINGS | HTOP_KEEP_FOLLOWING;
 }
 
 static Htop_Reaction actionToggleRunningInContainer(State* st) {
    st->settings->hideRunningInContainer = !st->settings->hideRunningInContainer;
+   st->settings->lastUpdate++;
```
cgzones, line 230: dito

BenBE:
`lsof`, `strace` and `locks` are those that don't need the flag, as there's a new screen opening for those where navigation with the cursor keys is expected.

Also I don't think inverting the logic brings too much advantages as this should be quite deliberate when handling an action.

Action.c
BenBE: The color rendering information for the threads: With kernel threads it's green, while without it's rendered as a normal thread.

Action.c
BenBE: Same as above

Action.c
BenBE: No change re the merge command rendering per se, but added for consistency such that all (batched) config updates can be triggered this way.

Action.c
BenBE: Yes. Strictly speaking, the update for user/kernel threads is not needed. I added it here regardless, to have the `lastUpdate` stamp consistently applied for batched updates (i.e. avoid having to think too much about if it is needed in any given place).

PR #1137, opened 2022-11-26
Fix locks screen

nabijaczleweli:
See commit messages, but in short, given:
```
$ head /proc/$$/fdinfo/{9,10,11}
==> /proc/797393/fdinfo/9 <==
pos:    0
flags:  0100000
mnt_id: 2775
lock:   1: FLOCK  ADVISORY  WRITE 818754 00:69:7 0 EOF

==> /proc/797393/fdinfo/10 <==
pos:    0
flags:  0100000
mnt_id: 127
lock:   1: FLOCK  ADVISORY  WRITE 828244 00:30:509028 0 EOF
lock:   2: OFDLCK ADVISORY  READ -1 00:30:509028 0 199

==> /proc/797393/fdinfo/11 <==
pos:    0
flags:  0100000
mnt_id: 2775
lock:   1: FLOCK  ADVISORY  WRITE 935925 00:69:1187 0 EOF
```

You get (before above, after below):

### DIFF
--- a/ProcessLocksScreen.c
+++ b/ProcessLocksScreen.c
@@ -27,7 +27,7 @@ ProcessLocksScreen* ProcessLocksScreen_new(const Process* process) {
       this->pid = process->tgid;
    else
       this->pid = process->pid;
-   return (ProcessLocksScreen*) InfoScreen_init(&this->super, process, NULL, LINES - 2, "        ID  TYPE       EXCLUSION  READ/WRITE DEVICE:INODE                              START                  END  FILENAME");
+   return (ProcessLocksScreen*) InfoScreen_init(&this->super, process, NULL, LINES - 2, "   FD TYPE       EXCLUSION  READ/WRITE DEVICE:INODE                              START                  END  FILENAME");
 }
 
 void ProcessLocksScreen_delete(Object* this) {
@@ -64,16 +64,16 @@ static void ProcessLocksScreen_scan(InfoScreen* this) {
 
          char entry[512];
          if (ULLONG_MAX == data->end) {
-            xSnprintf(entry, sizeof(entry), "%10d  %-10s %-10s %-10s %02x:%02x:%020"PRIu64" %20"PRIu64" %20s  %s",
-               data->id,
+            xSnprintf(entry, sizeof(entry), "%5d %-10s %-10s %-10s %02x:%02x:%020"PRIu64" %20"PRIu64" %20s  %s",
+               data->fd,
                data->locktype, data->exclusive, data->readwrite,
                data->dev[0], data->dev[1], data->inode,
                data->start, "<END OF FILE>",
                data->filename ? data->filename : "<N/A>"
             );
          } else {
-            xSnprintf(entry, sizeof(entry), "%10d  %-10s %-10s %-10s %02x:%02x:%020"PRIu64" %20"PRIu64" %20"PRIu64"  %s",
-               data->id,
+            xSnprintf(entry, sizeof(entry), "%5d %-10s %-10s %-10s %02x:%02x:%020"PRIu64" %20"PRIu64" %20"PRIu64"  %s",
+               data->fd,
                data->locktype, data->exclusive, data->readwrite,
                data->dev[0], data->dev[1], data->inode,
                data->start, data->end,
--- a/ProcessLocksScreen.c
+++ b/ProcessLocksScreen.c
@@ -27,7 +27,8 @@ ProcessLocksScreen* ProcessLocksScreen_new(const Process* process) {
       this->pid = process->tgid;
    else
       this->pid = process->pid;
-   return (ProcessLocksScreen*) InfoScreen_init(&this->super, process, NULL, LINES - 2, "   FD TYPE       EXCLUSION  READ/WRITE DEVICE:INODE                              START                  END  FILENAME");
+
+   return (ProcessLocksScreen*) InfoScreen_init(&this->super, process, NULL, LINES - 2, "   FD TYPE       EXCLUSION  READ/WRITE DEVICE       NODE               START                 END  FILENAME");
 }
 
 void ProcessLocksScreen_delete(Object* this) {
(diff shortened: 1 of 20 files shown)


nabijaczleweli:
Not sure what the errors are about, scanf has been required to support `%ms` for [like 14 years](https://pubs.opengroup.org/onlinepubs/9699919799/functions/scanf.html#tag_16_170_14)

ProcessLocksScreen.c
```diff
@@ -64,18 +65,18 @@ static void ProcessLocksScreen_scan(InfoScreen* this) {
...
                data->filename ? data->filename : "<N/A>"
             );
          } else {
-            xSnprintf(entry, sizeof(entry), "%10d  %-10s %-10s %-10s %02x:%02x:%020"PRIu64" %20"PRIu64" %20"PRIu64"  %s",
-               data->id,
+            xSnprintf(entry, sizeof(entry), "%5d %-10s %-10s %-10s %#6"PRIx64" %10"PRIu64" %19"PRIu64" %19"PRIu64"  %s",
+               data->fd,
                data->locktype, data->exclusive, data->readwrite,
-               data->dev[0], data->dev[1], data->inode,
+               (uint64_t) data->dev, data->inode,
```
BenBE, line 79: Why are you changing the width of the FD column? The width of 10 is intentional, due to the maximum value of `2^31-1` you could have there.

Also I don't think converting the `DEV` field in this way is portable.

linux/Platform.c
```diff
@@ -430,117 +431,91 @@ char* Platform_getProcessEnv(pid_t pid) {
    return env;
 }
 
-/*
- * Return the absolute path of a file given its pid&inode number
- *
- * Based on implementation of lslocks from util-linux:
- * https://sources.debian.org/src/util-linux/2.36-3/misc-utils/lslocks.c/#L162
- */
-char* Platform_getInodeFilename(pid_t pid, ino_t inode) {
```
BenBE, line 439: Please move the removal of the `Platform_getInodeFilename` implementation for the Linux platform into the other patch with the general cleanup of this; just re-order the commits accordingly in your PR. Makes it easier to review and cleans this bugfix patch.

BenBE (changes requested):
Please fix up the errors reported for the automated checks from the CI.

BenBE:
What about keeping both the lock ID AND the FD the lock is applicable to?

ProcessLocksScreen.c
nabijaczleweli: it is, dev_t is defined to be an integer type, and that's what we already write for lsof output (well, lsof `makedev()`s for us).

likewise, the FD column is 5-wide, like in lsof output. the limit here is not 2^31, it's "whatever the soft ulimit is", which for most systems is a sensible 1024. systemd sets its own ulimit to 1048576, besides that:
```
$ grep 'Max open files' /proc/*/limits | awk '{print $4}' | sort -n | uniq -c
      1 50
      1 256
    684 1024
      1 8192
      1 10000
      2 16384
      3 16424
      8 65535
      9 524288
      6 1048576
```
*If* you saw a ≥10^6 *ever* you'd have gotten reports to widen the FD column in the file listing. But you don't, so you haven't.

linux/Platform.c
nabijaczleweli: applied

nabijaczleweli:
How do you mean. This information doesn't exist. In the /proc/PID/fdinfo/ model, file descriptors have a list of locks; in the general case where you have between 0 and 1 lock per fd, you'll have a column with just "1"s. The "position in the global lock list" is a property of the output from /proc/locks, because the first number before the colon is just the ordinal in the current file.

ProcessLocksScreen.c
nabijaczleweli: Also, notably, I didn't change the width of the FD column, I took care to *not* do it, by replacing the 10-wide ID column (which, sure, you could have that many locks, I guess), with the 5-wide FD column copied from the file listing.

linux/Platform.c
```diff
@@ -430,117 +432,95 @@ char* Platform_getProcessEnv(pid_t pid) {
...
-      pdata->error = true;
-      return pdata;
-   }
+         FileLocks_LockData* ldata = xCalloc(1, sizeof(FileLocks_LockData));
+         FileLocks_Data* data = &ldata->data;
+
+         int _;
+         unsigned int maj, min;
+         char lock_end[25];
+         if (10 != sscanf(buffer + strlen("lock:\t"), "%d: %ms %ms %ms %d %x:%x:%"PRIu64" %"PRIu64" %24s",
```
BenBE, line 487: I'd personally avoid `%ms` in favor of `%31s` and appropriately sized buffers (`char foo[32]`) as this delays allocation until we are sure the data is valid. Also we aim for the code to be able to compile on any C99-compliant compiler, which would be broken by using this POSIX.1:2013 extension (or it's GNU extension counterpart).

Delaying these allocation even has the advantage, that the above lines 481/482 can be moved below too, furthermore streamlining the code flow.

BenBE (changes requested):

nabijaczleweli:
Both applied. CI failure on latest-minimal-gcc looks like an unrelated configuration error.

linux/Platform.c
```diff
@@ -430,117 +432,91 @@ char* Platform_getProcessEnv(pid_t pid) {
...
-      data->dev[0] = lock_dev[0];
-      data->dev[1] = lock_dev[1];
-      data->inode = lock_inode;
-      data->start = strtoull(lock_start, NULL, 10);
-      if (!String_eq(lock_end, "EOF")) {
-         data->end = strtoull(lock_end, NULL, 10);
-      } else {
-         data->end = ULLONG_MAX;
+         *data_ref = xCalloc(1, sizeof(FileLocks_LockData));
+         memcpy(&(*data_ref)->data, &data, sizeof(data));
```
BenBE, line 508:
```suggestion
         (*data_ref)->data = data;
```

linux/Platform.c
```diff
@@ -430,117 +432,91 @@ char* Platform_getProcessEnv(pid_t pid) {
...
 
+   FileLocks_LockData** data_ref = &pdata->locks;
    while ((de = readdir(dirp))) {
       if (String_eq(de->d_name, ".") || String_eq(de->d_name, ".."))
          continue;
 
-      /* care only for numerical descriptors */
-      if (!strtoull(de->d_name, (char **) NULL, 10))
+      errno = 0;
+      char *end;
```
BenBE, line 460: Avoid uninitialized memory access if `strtoull` does not set `end` (DiD):
```suggestion
      char *end = de->d_name;
```

nabijaczleweli:
sure; both applied

linux/Platform.c
BenBE: Doing just this causes the build failure.

Please add this patch to resolve the build issue:
```diff
diff --git a/linux/Platform.c b/linux/Platform.c
index f5aa6b96..64f25c42 100644
--- a/linux/Platform.c
+++ b/linux/Platform.c
@@ -434,7 +434,6 @@ char* Platform_getProcessEnv(pid_t pid) {
 
 FileLocks_ProcessData* Platform_getProcessLocks(pid_t pid) {
    FileLocks_ProcessData* pdata = xCalloc(1, sizeof(FileLocks_ProcessData));
-   const struct dirent* de;
    DIR* dirp;
    int dfd;
 
@@ -452,7 +451,7 @@ FileLocks_ProcessData* Platform_getProcessLocks(pid_t pid) {
    }
 
    FileLocks_LockData** data_ref = &pdata->locks;
-   while ((de = readdir(dirp))) {
+   for (struct dirent* de; (de = readdir(dirp)); ) {
       if (String_eq(de->d_name, ".") || String_eq(de->d_name, ".."))
          continue;
 
```

nabijaczleweli:
applied

BenBE (approved):

PR #1139, opened 2022-11-27
Add options to display parents and children processes of filtered pro…

yurenchen000:
…cesses

Co-authored-by: yurenchen000 <yurenchen@yeah.net>

### DIFF
--- a/ProcessList.c
+++ b/ProcessList.c
@@ -370,6 +370,16 @@ void ProcessList_collapseAllBranches(ProcessList* this) {
    }
 }
 
+static void ProcessList_filterChildern(ProcessList *this, pid_t pid, Hashtable *processFilter) {
+   for (int i = Vector_size(this->processes) - 1; i >= 0; i--) {
+      Process *p = (Process*) (Vector_get(this->processes, i));
+      if (p->pid != pid && Process_isChildOf(p, pid)) {
+         Hashtable_put(processFilter, p->pid, (void*) 1);
+         ProcessList_filterChildern(this, p->pid, processFilter);
+      }
+   }
+}
+
 void ProcessList_rebuildPanel(ProcessList* this) {
    ProcessList_updateDisplayList(this);
 
@@ -392,6 +402,25 @@ void ProcessList_rebuildPanel(ProcessList* this) {
 
    const int processCount = Vector_size(this->displayList);
    int idx = 0;
+
+   // Mark Children and Parent for F4-Filter
+   Hashtable* filteredProcesses = Hashtable_new(processCount, false);
+   for (int i = 0; i < processCount; i++) {
+      Process* p = (Process*)Vector_get(this->processes, i);
+      pid_t ppid = Process_getParentPid(p);
+
+      if (incFilter && !(String_contains_i(Process_getCommand(p), incFilter, true)))
+         continue;
+
+      if (this->settings->showChildrenInFilter)
+         ProcessList_filterChildern(this, p->pid, filteredProcesses);
+      do {
+         Hashtable_put(filteredProcesses, p->pid, (void*) 1);
+         ppid = Process_getParentPid(p);
+         p = Hashtable_get(this->processTable, ppid);
+      } while (this->settings->showParentsInFilter && p && p->pid != p->ppid);
+   }
+
    bool foundFollowed = false;
 
    for (int i = 0; i < processCount; i++) {
@@ -399,7 +428,7 @@ void ProcessList_rebuildPanel(ProcessList* this) {
 
       if ( (!p->show)
          || (this->userId != (uid_t) -1 && (p->st_uid != this->userId))
-         || (incFilter && !(String_contains_i(Process_getCommand(p), incFilter, true)))
+         || (incFilter && !Hashtable_get(filteredProcesses, p->pid))
          || (this->pidMatchList && !Hashtable_get(this->pidMatchList, p->tgid)) )
          continue;
 
@@ -428,6 +457,7 @@ void ProcessList_rebuildPanel(ProcessList* this) {
 
       this->panel->scrollV = currScrollV;
    }
+   Hashtable_delete(filteredProcesses);
 }
 
 Process* ProcessList_getProcess(ProcessList* this, pid_t pid, bool* preExisting, Process_New constructor) {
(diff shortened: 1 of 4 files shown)


vlada-dudr:
Exactly what I wanted. Works good for me.

ProcessList.c
```diff
@@ -370,6 +370,16 @@ void ProcessList_collapseAllBranches(ProcessList* this) {
    }
 }
 
+static void ProcessList_filterChildern(ProcessList *this, pid_t pid, Hashtable *processFilter) {
```
fasterit, line 373: Typo Childern -> Children

ProcessList.c
```diff
@@ -392,14 +402,33 @@ void ProcessList_rebuildPanel(ProcessList* this) {
...
+   Hashtable* filteredProcesses = Hashtable_new(processCount, false);
+   for (int i = 0; i < processCount; i++) {
+      Process* p = (Process*)Vector_get(this->processes, i);
+      pid_t ppid = Process_getParentPid(p);
+
+      if (incFilter && !(String_contains_i(Process_getCommand(p), incFilter, true)))
+         continue;
+
+      if (this->settings->showChildrenInFilter)
+         ProcessList_filterChildern(this, p->pid, filteredProcesses);
```
fasterit, line 416: Typo Childern -> Children

fasterit (changes requested):
Very nice work @yurenchen000.
This should have a key binding to be quickly toggled as well. May be Shift-F4 + Ctrl-F4?
(Marked the typo Childern, please fix)

yurenchen000:
> Very nice work @yurenchen000. This should have a key binding to be quickly toggled as well. May be Shift-F4 + Ctrl-F4? (Marked the typo Childern, please fix)

got it, I'll fix typo right now.
but I'm not sure if I can add keybinding for it

ProcessList.c
yurenchen000: fixed it

ProcessList.c
yurenchen000: fixed

BenBE (commented):
I think those options should probably be exclusive: Either show parents or children, but showing both does not make any sense with filtering.

yurenchen000:
> I think those options should probably be exclusive: Either show parents or children, but showing both does not make any sense with filtering.

I think it has practical uses, imagine this scenario:
 - filter by `bash`
   - view It's parents (ancestors): who open it, a ssh or tmux
   - view It's children (descendants): what it doing, run vim or some service

// equivalent to viewing its family tree

BenBE:
I was assuming this refered to the whole tree, not just the next level in each direction. In that case having both makes sense.

ProcessList.c
```diff
@@ -392,14 +402,34 @@ void ProcessList_rebuildPanel(ProcessList* this) {
...
+         continue;
+
+      if (this->settings->showChildrenInFilter)
+         ProcessList_filterChildren(this, p->pid, filteredProcesses);
+
+      do {
+         Hashtable_put(filteredProcesses, p->pid, (void*) 1);
+         ppid = Process_getParentPid(p);
+         p = Hashtable_get(this->processTable, ppid);
+      } while (this->settings->showParentsInFilter && p && p->pid != p->ppid);
```
tanriol, line 422: As we know from #1017 (which I still need to rebase/reimplement), in some cases two processes may be parents of each other and thus this loop becomes an infinite one.

tanriol (changes requested):
There's a couple problems with this implementation. Also, I'm a bit worried about growing another thing dealing with process trees completely separate from tree view implementation. Would be nice to share the hard part in some reasonable way.

tanriol:
Also, do these options make sense in flat view or are they intended for tree mode only?

yurenchen000:
> I was assuming this refered to the whole tree, not just the next level in each direction. In that case having both makes sense.

agree, use whole tree is my primary usage,
if matched item can highlight it'll perfect

yurenchen000:
> Also, do these options make sense in flat view or are they intended for tree mode only?

@tanriol 
yes, The current implementation can filter normally in non-tree view,
But I also think it doesn't make much sense in non-tree view.

Did you mean that limit this functionality to tree view mode, 
And figure out a way to reuse the process tree information from the tree view?

ProcessList.c
yurenchen000: got it, 
I will keep an eye on the progress of that issue

ProcessList.c
```diff
@@ -370,6 +370,16 @@ void ProcessList_collapseAllBranches(ProcessList* this) {
    }
 }
 
+static void ProcessList_filterChildren(ProcessList *this, pid_t pid, Hashtable *processFilter) {
+   for (int i = Vector_size(this->processes) - 1; i >= 0; i--) {
+      Process *p = (Process*) (Vector_get(this->processes, i));
+      if (p->pid != pid && Process_isChildOf(p, pid)) {
+         Hashtable_put(processFilter, p->pid, (void*) 1);
+         ProcessList_filterChildren(this, p->pid, processFilter);
```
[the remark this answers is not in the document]
eMPee584: Do you have an idea how to do better?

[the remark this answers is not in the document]
tanriol: Even a single loop over processes with walking up from every process would be better (although it would need some cycle detection too) as it can easily be made worst case linear in the total process number. I personally think something like the tree mode "sort by parent" would be even better, but I'm a bit out of context at the moment and not sure how feasible it would be to share that part instead of reimplementing it.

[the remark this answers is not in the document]
BenBE: You could make the hit detection (in subtree/any parent) in O(1) if we tracked some item ranges for nested sets (similar to what Joomla is doing for its menus). Unfortunately, when we tried to implement this in-place while building the tree, this made tree creation kinda slow (`O(n²)`). One option could be to calculate the nested set ranges on the fly from the `Table->displayList` based on the indent columns; takes at most `O(n*m)` with `m` being the maximum tree depth (which is fixed at 32).

BenBE:
> Also, do these options make sense in flat view or are they intended for tree mode only?

Depending on the task it can make sense to have this available outside tree view too.

thapakazi:
anyone working on this ?

for now I am achieving same with:
```
process_name=tmux; 
watch -n 1 -d "pgrep $process_name | xargs -I {} pstree -p {}"
```

darkweaver87:
Thanks @thapakazi for the tip. I'm also interested in it because @thapakazi workaround doesn't work for processes spawning subprocesses after `htop` is launched.

PR #1141, opened 2022-11-28
Add a systemd user meter to monitor user units.

### LINKED ISSUE #1140: Add systemd user meter
Hello,

I currently enjoy having htop tell me at a glance whether there are any failed services, but it would be nice to have this also work for systemd user units.  The addition of a systemd user meter would do this, and taking a quick look at the code it seems like it would just require adding `--user` to the `systemctl` command in the existing `SystemdMeter`.  What other changes would be required for adding a new meter based on an existing one?

Thanks.

Edit:
After a bit of testing just modifying the existing meter it looks like I also need to use `sd_bus_open_user` instead of `sd_bus_open_system`.

aereaux:
Fixes #1140 

Let me know if I missed anything.

### DIFF
--- a/linux/SystemdMeter.c
+++ b/linux/SystemdMeter.c
@@ -9,6 +9,7 @@ in the source distribution for its full text.
 
 #include <dlfcn.h>
 #include <fcntl.h>
+#include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -39,6 +40,7 @@ in the source distribution for its full text.
 typedef void sd_bus;
 typedef void sd_bus_error;
 static int (*sym_sd_bus_open_system)(sd_bus**);
+static int (*sym_sd_bus_open_user)(sd_bus**);
 static int (*sym_sd_bus_get_property_string)(sd_bus*, const char*, const char*, const char*, const char*, sd_bus_error*, char**);
 static int (*sym_sd_bus_get_property_trivial)(sd_bus*, const char*, const char*, const char*, const char*, sd_bus_error*, char, void*);
 static sd_bus* (*sym_sd_bus_unref)(sd_bus*);
@@ -46,37 +48,43 @@ static void* dlopenHandle = NULL;
 
 #endif /* BUILD_STATIC */
 
-#if !defined(BUILD_STATIC) || defined(HAVE_LIBSYSTEMD)
-static sd_bus* bus = NULL;
-#endif /* !BUILD_STATIC || HAVE_LIBSYSTEMD */
-
 
 #define INVALID_VALUE ((unsigned int)-1)
 
-static char* systemState = NULL;
-static unsigned int nFailedUnits = INVALID_VALUE;
-static unsigned int nInstalledJobs = INVALID_VALUE;
-static unsigned int nNames = INVALID_VALUE;
-static unsigned int nJobs = INVALID_VALUE;
+typedef struct SystemdMeterContext {
+#if !defined(BUILD_STATIC) || defined(HAVE_LIBSYSTEMD)
+   sd_bus* bus;
+#endif /* !BUILD_STATIC || HAVE_LIBSYSTEMD */
+   char* systemState;
+   unsigned int nFailedUnits;
+   unsigned int nInstalledJobs;
+   unsigned int nNames;
+   unsigned int nJobs;
+} SystemdMeterContext_t;
+
+static SystemdMeterContext_t ctx_system;
+static SystemdMeterContext_t ctx_user;
 
 static void SystemdMeter_done(ATTR_UNUSED Meter* this) {
-   free(systemState);
-   systemState = NULL;
+   SystemdMeterContext_t* ctx = String_eq(Meter_name(this), "SystemdUser") ? &ctx_user : &ctx_system;
+
+   free(ctx->systemState);
+   ctx->systemState = NULL;
 
 #ifdef BUILD_STATIC
 # ifdef HAVE_LIBSYSTEMD
-   if (bus) {
-      sym_sd_bus_unref(bus);
+   if (ctx->bus) {
+      sym_sd_bus_unref(ctx->bus);
    }
-   bus = NULL;
+   ctx->bus = NULL;
 # endif /* HAVE_LIBSYSTEMD */
 #else /* BUILD_STATIC */
-   if (bus && dlopenHandle) {
-      sym_sd_bus_unref(bus);
+   if (ctx->bus && dlopenHandle) {
+      sym_sd_bus_unref(ctx->bus);
    }
-   bus = NULL;
+   ctx->bus = NULL;
 
-   if (dlopenHandle) {
+   if (!ctx_system.systemState && !ctx_user.systemState && dlopenHandle) {
       dlclose(dlopenHandle);
       dlopenHandle = NULL;
    }
@@ -84,7 +92,8 @@ static void SystemdMeter_done(ATTR_UNUSED Meter* this) {
 }
 
 #if !defined(BUILD_STATIC) || defined(HAVE_LIBSYSTEMD)
-static int updateViaLib(void) {
+static int updateViaLib(bool user) {
+   SystemdMeterContext_t* ctx = user ? &ctx_user : &ctx_system;
 #ifndef BUILD_STATIC
    if (!dlopenHandle) {
       dlopenHandle = dlopen("libsystemd.so.0", RTLD_LAZY);
@@ -101,6 +110,7 @@ static int updateViaLib(void) {
       } while(0)
 
       resolve(sd_bus_open_system);
+      resolve(sd_bus_open_user);
       resolve(sd_bus_get_property_string);
       resolve(sd_bus_get_property_trivial);
       resolve(sd_bus_unref);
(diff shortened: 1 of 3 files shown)


BenBE:
Hi, thank you for your PR. As both the Systemd Meter and its user session counterpart are quite similar (and topically related), I'd much prefer to share the code (and put things over in the existing `SystemdMeter.[ch]` files.

BenBE:
> That makes total sense.

Absolutely. For one avoids much code duplication.

> I'm not a huge C expert, so I'll see what I can do …

No problem. If there are specific questions feel free to ask them. I'm marking this PR as draft for now; once things are sorted you just need to update your branch with the new code.

> … but let me know if you have any suggestions for how to structure that.

In the existing code there's a list of function signatures/types to hold the dynamically determined function pointers from systemd. You should go and extend these lists to include the additional pointers you need. Further down, in the section, where function pointers are loaded, proper initialization of these new pointers needs to be added.

After that (I didn't check the implementation in detail) you need to extend the function that reads the desired values, to read both values: AFAICS this is affects `updateViaLib` and `updateViaExec`. Given these two functions are basically identical (except for how the dbus connection s established) it should suffice to pass a `bool` (`#include <stdbool.h>`) and switch between the appropriate function calls towards systemd.

At the end of the file, you can duplicate that data structure `const MeterClass SystemdUserMeter_class` (one for each of the meters). From there you either need to have separate display/update functions OR you can check for the `name` member of the provided `Meter` and decide whether you are running as global or user meter from there.

Sounds like a lot, but most of the stuff already exists. There shouldn't be much more changes necessary in places besides those you already needed to touch for this first version of the patch (and those I mentioned above).

Note: Be sure to properly manage both bus connections properly.

aereaux:
OK, still a bit of duplication, and I'm not entirely sure I'm managing the two buses right, but it seems to work.

linux/SystemdMeter.c
```diff
@@ -48,6 +50,7 @@ static void* dlopenHandle = NULL;
 
 #if !defined(BUILD_STATIC) || defined(HAVE_LIBSYSTEMD)
 static sd_bus* bus = NULL;
+static sd_bus* userbus = NULL;
```
BenBE, line 53: Maybe better name them:
```suggestion
static sd_bus* bus_system = NULL;
static sd_bus* bus_user = NULL;
```

linux/SystemdMeter.c
```diff
@@ -14,6 +14,7 @@ in the source distribution for its full text.
 #include <string.h>
 #include <unistd.h>
 #include <sys/wait.h>
+#include <stdbool.h>
```
BenBE, line 17: Header order (alphabetical, cf. style guide in docs directory for details).

linux/SystemdMeter.c
```diff
@@ -83,8 +88,34 @@ static void SystemdMeter_done(ATTR_UNUSED Meter* this) {
...
+#else /* BUILD_STATIC */
+   if (userbus && dlopenHandle) {
+      sym_sd_bus_unref(userbus);
+   }
+   userbus = NULL;
+
+   if (dlopenHandle) {
+      dlclose(dlopenHandle);
+      dlopenHandle = NULL;
+   }
```
BenBE, line 113: This must not be freed unless neither `bus_system` nor `bus_user` are also already freed.

linux/SystemdMeter.c
```diff
@@ -180,8 +218,13 @@ static int updateViaLib(void) {
...
 busfailure:
-   sym_sd_bus_unref(bus);
-   bus = NULL;
+   if(user) {
+      sym_sd_bus_unref(userbus);
+      userbus = NULL;
+   } else {
+      sym_sd_bus_unref(bus);
+      bus = NULL;
+   }
```
BenBE, line 227: Should be freed independently.

linux/SystemdMeter.c
```diff
@@ -221,16 +264,30 @@ static void updateViaExec(void) {
...
-         "--property=NNames",
-         "--property=NJobs",
-         "--property=NInstalledJobs",
-         (char *)NULL);
+      if (user) {
+         execlp(
+            "systemctl",
+            "systemctl",
+            "show",
+	    "--user",
```
BenBE, line 272: Indentation, cf. style guide.

aereaux:
Actually, with a couple further changes it looks like the meters get reset to default on restart when I include a SystemdUserMeter.  Not sure what the problem is

BenBE (changes requested):
NB: While reviewing the changes I noticed a potential issue (of the original code) when more than one instance of the meter is added (and removed). The easiest fix (which would also address the code duplication in the `Systemd*Meter_done` function would be ref-counting instances and only freeing resources, one all references are gone (keeping the user and system bus handles around even if one of them would be unused).

Apart from this, the changes already look quite good.

linux/SystemdMeter.c
aereaux: I think this does what you mean

linux/SystemdMeter.c
aereaux: Not sure what you mean here.

aereaux:
Is it OK to reuse the state variables (`systemState`, etc)?  It appears to only be showing the user state in each of the meters although I think it was working before.

linux/SystemdMeter.c
```diff
@@ -221,16 +265,30 @@ static void updateViaExec(void) {
...
+            "systemctl",
+            "systemctl",
+            "show",
+            "--property=SystemState",
+            "--property=NFailedUnits",
+            "--property=NNames",
+            "--property=NJobs",
+            "--property=NInstalledJobs",
+            (char *)NULL);
+      }
```
BenBE, line 291: Given that there's an implied default `--system`, we can simplify here. :)
```suggestion
      execlp(
         "systemctl",
         "systemctl",
         "show",
         user ? "--user" : "--system",
         "--property=SystemState",
         "--property=NFailedUnits",
         "--property=NNames",
         "--property=NJobs",
         "--property=NInstalledJobs",
         (char *)NULL);
```

BenBE (commented):
One minor suggestion, otherwise LGTM. Haven't tested this yet though.

BenBE:
> Is it OK to reuse the state variables (`systemState`, etc)? It appears to only be showing the user state in each of the meters although I think it was working before.

No, needs to be split, so that both contexts (system & user) have their own variables for these values.

```c
typedef struct SystemdMeterContext {
   sd_bus bus;
   int32_t systemState;
   // ...
} SystemdMeterContext_t;

static SystemdMeterContext_t ctx_system;
static SystemdMeterContext_t ctx_user;
```

You can then pass around the appropriate context variable. Would use `ctx_system.bus` instead of `bus_system` and so on (or `my_ctx->bus`).

aereaux:
How do I check the name of the passed Meter?  I tried `SystemdMeterContext_t ctx = this->name == "SystemdUser" ? ctx_user: ctx_system;` but then I get this error:
```
linux/SystemdMeter.c: In function ‘SystemdMeter_done’:
linux/SystemdMeter.c:69:36: error: ‘Meter’ {aka ‘struct Meter_’} has no member named ‘name’
   69 |    SystemdMeterContext_t ctx = this->name == "SystemdUser" ? ctx_user: ctx_system;
      |                                    ^~
```

BenBE:
> How do I check the name of the passed Meter? I tried
> 
> ```
> static void SystemdMeter_done(ATTR_UNUSED Meter* this) {
>    SystemdMeterContext_t ctx = this->name == "SystemdUser" ? ctx_user : ctx_system;
> ```
> 
> but then I get this error:
> 
> ```
> linux/SystemdMeter.c: In function ‘SystemdMeter_done’:
> linux/SystemdMeter.c:69:36: error: ‘Meter’ {aka ‘struct Meter_’} has no member named ‘name’
>    69 |    SystemdMeterContext_t ctx = this->name == "SystemdUser" ? ctx_user: ctx_system;
>       |                                    ^~
> ```

HAve a look at https://github.com/htop-dev/htop/blob/1707642f5b2536dc5a30d54163f6ce6123754331/CPUMeter.c#L190

Note, that in C strings are compared with `strcmp`, but preferrably use the existing wrapper `String_eq` (cf. https://github.com/htop-dev/htop/blob/1707642f5b2536dc5a30d54163f6ce6123754331/XUtils.h#L45-L47).

aereaux:
Just pushed another version.  I think I took care of all your comments so far but let me know if not.

linux/SystemdMeter.c
```diff
@@ -39,52 +40,60 @@ in the source distribution for its full text.
...
-   if (bus && dlopenHandle) {
-      sym_sd_bus_unref(bus);
+   if (ctx->bus && dlopenHandle) {
+      sym_sd_bus_unref(ctx->bus);
    }
-   bus = NULL;
+   ctx->bus = NULL;
 
-   if (dlopenHandle) {
+   if (ctx_system.systemState == NULL && ctx_user.systemState == NULL && dlopenHandle) {
```
BenBE, line 87: Can be shortened:
```suggestion
   if (!ctx_system.systemState && !ctx_user.systemState && dlopenHandle) {
```

BenBE (approved):
LGTM.

BenBE:
> One could use the pseudo class hierarchy to check what meter to operate on:

@cgzones In that case it would probably be better to just put the referenced context and the `isUser` flag into the derived meter classes and access it from there. But I don't think that would be any improvement over how things are now.

I think I'll merge this as is, unless anyone objects …

Issue #1144, opened 2022-12-02
Linux: CPU frequency not displayed in "text" mode for meter


dare-to-recurse:
When running htop 3.2.1 (compiled from source) the CPU meters don't seem to show the frequency when set to "text" mode, but they do display the frequency in the "bar" mode.

Not sure if it worked differently in any previous versions as I haven't tested them.

PR #1145, opened 2022-12-02
CPUMeter: show frequency in text mode

cgzones:
Closes: #1144


BenBE (approved):

cgzones:
padded `N/A` such that the following `temp` columns are aligned in case frequency in known for some cores but not all

PR #1151, opened 2022-12-26
Use float for non-precision values

cgzones:
Use single precision floating point numbers for calculations where the reduced precision is sufficient enough.

May save some cycles on architectures not supporting double precision arithmetic in hardware.


BenBE:
This touches quite a lot of source for probably little benefit. Any particular measurements that show this is worth all this noise?

Explorer09:
Hardware _not_ supporting double precision float is rare, since C89 already defined double precision floating point functions (math.h). The single precision variants of math.h functions came later, in C99.

PR #1153, opened 2022-12-27
Implement zswap support on Linux

### LINKED ISSUE #104: Feature request: Account for zswap in memory usage

### LINKED ISSUE #721: swap and zram
Maybe this is a question, but the output of zram and swp is confusing. I'd like to see which swap is used at a glance more than total swap, like the output of `swapon` on my system.

Also the 12k/16k isn't making much sense. I assume that has something to do with the compressed size, but I don't know why there are two numbers.

```
swapon                                                                    # ~
NAME           TYPE       SIZE  USED PRIO
/dev/nvme0n1p6 partition 32.2G    0B   -2
/dev/zram0     partition    4G 15.1M  100
```

Htop showing 16k zram used an 15.1M swap total

intelfx:
When zswap is enabled and in use on a Linux system, htop output does not reflect this in any way. This leads to two issues:

1. Compressed pool state is not communicated to the user (this is less important),
2. Zswapped pages are accounted to swap usage, which is misleading (this is more important).

As of Linux 5.19, zswap metrics can be read from `/proc/meminfo` and no longer require debugfs. Use this to provide zswap support in two ways:

- highlight the compressed pages on the RAM gauge as a separate type of RAM usage (but still account them as "used" RAM),
- highlight the frontswapped pages on the Swp gauge as a separate type of swap usage, and discount them from the "used" swap because these pages are not physically residing on the swap device.

(Note: we still want to highlight the frontswapped pages on the Swp gauge, because frontswap will only work when there is enough physical swap to evict the frontswapped pages all the way to the backing device. Thus it is important to give the user an insight into this metric.)

Fixes #104.
Potentially closes #721.


MemoryMeter.c
```diff
@@ -18,6 +18,7 @@ in the source distribution for its full text.
 
 static const int MemoryMeter_attributes[] = {
    MEMORY_USED,
+   MEMORY_COMPRESSED,
```
BenBE, line 21: I don't quite like how this indexing change causes a cascade of changes even on unrelated platforms that do not implement this.

linux/LinuxProcessList.c
```diff
@@ -1821,6 +1824,10 @@ static inline void LinuxProcessList_scanMemoryInfo(ProcessList* this) {
             break;
          }
          break;
+      case 'Z':
+	      tryRead("Zswap:", zswapCompMem);
+	      tryRead("Zswapped:", zswapOrigMem);
+	      break;
```
BenBE, line 1830: Indentation

linux/LinuxProcessList.c
```diff
@@ -1904,6 +1913,24 @@ static void LinuxProcessList_scanHugePages(LinuxProcessList* this) {
...
+	r = xReadfile("/sys/module/zswap/parameters/max_pool_percent", buf, 256);
+	if (r <= 0) {
+		return;
+	}
+	max_pool_percent = strtol(buf, NULL, 10);
+	if (max_pool_percent < 0 || max_pool_percent > 100) {
+		return;
+	}
+
+	this->zswap.totalZswapPool = this->super.totalMem * max_pool_percent / 100;
```
BenBE, line 1930: Indentation

linux/Platform.c
```diff
@@ -352,34 +352,55 @@ void Platform_setMemoryValues(Meter* this) {
...
 }
 
 void Platform_setZramValues(Meter* this) {
    const LinuxProcessList* lpl = (const LinuxProcessList*) this->pl;
-   this->total = lpl->zram.totalZram;
-   this->values[0] = lpl->zram.usedZramComp;
-   this->values[1] = lpl->zram.usedZramOrig;
+
+   if (lpl->zram.totalZram > 0) {
+	   this->total = lpl->zram.totalZram;
```
BenBE, line 396: Indentation

linux/ZswapStats.h
```diff
@@ -0,0 +1,13 @@
+#ifndef HEADER_ZswapStats
+#define HEADER_ZswapStats
+
+typedef struct ZswapStats_ {
+	/* maximum size of the zswap pool */
+	memory_t totalZswapPool;
+	/* amount of RAM used by the zswap pool */
+	memory_t usedZswapComp;
+	/* amount of data stored inside the zswap pool */
+	memory_t usedZswapOrig;
```
BenBE, line 10: Indentation

linux/ZswapStats.h
```diff
@@ -0,0 +1,13 @@
+#ifndef HEADER_ZswapStats
```
BenBE, line 1: Missing copyright header

BenBE (changes requested):
A quick glance at the set of changes looks mostly good, except that for the way they are done they mostly cause cascading changes to unrelated parts in the code where they should not be needed and thus avoided. That doesn't mean they are wrong, but they should be properly reasoned and justified.

Also this PR comments out quite a bit of code, which is normally a sign for something strange going on.

linux/LinuxProcessList.c
intelfx: Fixed, along with the rest of the indentation fails. Sorry for the noise, my editor didn't apply the editorconfig file entirely correctly and I just missed all the tabs.

linux/ZswapStats.h
intelfx: Fixed

MemoryMeter.c
BenBE: While not perfect either, putting the compressed memeroy value after shared (and before cached) should avoid most of the edits down the line while keeping the "used" section of memory together. This isn't quite without complications either, but should at least reduce the edits on unrelated platforms quite a bit.

intelfx:
There is actually an issue with the swap accounting change that I have just discovered after running the patched htop on a different machine.

Linux has two types of occupied swap pages: normal and SwapCached. Htop treats them separately, meaning that `SwapUsed = SwapTotal - SwapFree - SwapCached`, or `SwapTotal - SwapFree = SwapUsed + SwapCached`. And it so happens that the frontswapped pages may span both categories:

```
|<-------(SwapTotal-SwapFree)------>|<---SwapFree--->|
|--------------------|--------------|----------------|
|      SwapUsed      |  SwapCached  |                |
|--------------------|--------------|----------------|
|<------Zswapped------>|
```

TL;DR is that sometimes `Zswapped > SwapUsed`, so this means I can't just subtract Zswapped from SwapUsed. 

However, Linux doesn't tell us which exact proportion of SwapUsed vs SwapCached pages are Zswapped.

Any ideas?

MemoryMeter.c
intelfx: I have reworked the patch as you suggested. IMO, the new patch isn't all that smaller (21 KiB vs 23 KiB email-formatted), but anyway, it's up to you.

Here's a before/after screenshot (top is v2, bottom is v1):
Снимок экрана от 2022-12-27 17-53-10

intelfx:
I have applied a gross-ish hack to work around the issue I described (just so that we don't hit negative swap usage figures), but I don't like it.

intelfx:
ping?

BenBE:
Two notes:

1. There was a patch recently to make the MemoryMeter/SwapMeter values more accessible. Please rebase your patch to include those changes. Note that the SwapMeter changes are not yet landed.

2. Can you somehow have a look at the kernel/its documentation if there really is no statistics on the amount of swapped-out pages related to Zswap'd memory? If there really is no information re that metric, it's kinda bad behaviour just assuming everything is in RAM, as the displayed bar should roughly correspond to what memory is actually used.

intelfx:
@BenBE my apologies, I got sidetracked.

1. I have rebased these patches on top of the main branch and while at it, fixed up the last remaining places using magic constants instead of the newly added `(MEMORY|SWAP)_METER_*` macros.
2. I spent a few days reading the kernel sources and I unfortunately see no apparent way to tell Zswap&&SwapUsed from Zswapped&&SwapCached.

intelfx:
@BenBE ping?

Compat.h
```diff
@@ -13,6 +13,7 @@ in the source distribution for its full text.
 #include <stddef.h>
 #include <unistd.h>
 #include <sys/stat.h> // IWYU pragma: keep
+#include <assert.h> // for _Static_assert() and/or static_assert()
```
BenBE, line 16: Header order

linux/ZswapStats.h
```diff
@@ -0,0 +1,19 @@
...
+/*
+htop - ZswapStats.h
+(C) 2022 htop dev team
+Released under the GNU GPLv2+, see the COPYING file
+in the source distribution for its full text.
+*/
+
+typedef struct ZswapStats_ {
+   /* maximum size of the zswap pool */
+   memory_t totalZswapPool;
```
BenBE, line 12: Header should be compile-able stand-alone. Needs to include the necessary headers.

linux/LinuxProcessList.h
```diff
@@ -16,6 +16,7 @@ in the source distribution for its full text.
 #include "ProcessList.h"
 #include "UsersTable.h"
 #include "ZramStats.h"
+#include "ZswapStats.h"
```
BenBE, line 19:
```suggestion
#include "linux/ZswapStats.h"
```

Compat.h
intelfx: Fixed, thanks.

linux/LinuxProcessList.h
intelfx: Fixed; I've also corrected the similar issue with `ZramStats.h` (which is what I used as a reference for the change I made).

linux/ZswapStats.h
intelfx: Fixed; I've also corrected the similar issue with `ZramStats.h` (which is what I used as a reference for the change I made).

Compat.h
```diff
@@ -9,6 +9,7 @@ in the source distribution for its full text.
 
 #include "config.h" // IWYU pragma: keep
 
+#include <assert.h> // for _Static_assert() and/or static_assert()
```
BenBE, line 12: No need to note the included symbols; at most tag for IWYU if the use of the included symbol is optional (likely the case here).

Please rebase the fixup into the appropriate commit.

BenBE (commented):
One minor thing left. LGTM apart from that.

Compat.h
intelfx: Done. iwyu 0.19 didn't complain for me but I have put in the tag anyway because it will likely change in future as various branches of the compat shims get exercised.

BenBE (approved):

Issue #1161, opened 2023-01-06
Display scheduler class


kanashimia:
Same as `cls` option in `ps`

Manpage citation:
```
cls         CLS       
    scheduling class of the process.  (alias policy, cls).  
    Field's possible values are:

    -   not reported
    TS  SCHED_OTHER
    FF  SCHED_FIFO
    RR  SCHED_RR
    B   SCHED_BATCH
    ISO SCHED_ISO
    IDL SCHED_IDLE
    DLN SCHED_DEADLINE
    ?   unknown value
```

PR #1162, opened 2023-01-07
Add support for scheduling policies

cgzones:
Show the current scheduling class of the process.

Currently implemented on Linux only but should be portable, since `sched_getscheduler(2)` is part of POSIX.1-2001.

Closes: #1161

/cc @kanashimia

### DIFF
--- a/Action.c
+++ b/Action.c
@@ -29,6 +29,7 @@ in the source distribution for its full text.
 #include "Process.h"
 #include "ProcessLocksScreen.h"
 #include "ProvideCurses.h"
+#include "Scheduling.h"
 #include "ScreenManager.h"
 #include "SignalsPanel.h"
 #include "TraceScreen.h"
@@ -407,6 +408,51 @@ static Htop_Reaction actionSetAffinity(State* st) {
 
 }
 
+#ifdef SCHEDULER_SUPPORT
+static Htop_Reaction actionSetSchedPolicy(State* st) {
+   if (Settings_isReadonly())
+      return HTOP_KEEP_FOLLOWING;
+
+   static int preSelectedPolicy = SCHEDULINGPANEL_INITSELECTEDPOLICY;
+   static int preSelectedPriority = SCHEDULINGPANEL_INITSELECTEDPRIORITY;
+
+   Panel* schedPanel = Scheduling_newPolicyPanel(preSelectedPolicy);
+
+   const ListItem* policy;
+   for(;;) {
+      policy = (const ListItem*) Action_pickFromVector(st, schedPanel, 18, true);
+
+      if (!policy || policy->key != -1)
+         break;
+
+      Scheduling_togglePolicyPanelResetOnFork(schedPanel);
+   }
+
+   if (policy) {
+      preSelectedPolicy = policy->key;
+
+      Panel* prioPanel = Scheduling_newPriorityPanel(policy->key, preSelectedPriority);
+      if (prioPanel) {
+         const ListItem* prio = (const ListItem*) Action_pickFromVector(st, prioPanel, 14, true);
+         if (prio)
+            preSelectedPriority = prio->key;
+
+         Panel_delete((Object*) prioPanel);
+      }
+
+      SchedulingArg v = { .policy = preSelectedPolicy, .priority = preSelectedPriority };
+
+      bool ok = MainPanel_foreachProcess(st->mainPanel, Scheduling_setPolicy, (Arg) { .v = &v }, NULL);
+      if (!ok)
+         beep();
+   }
+
+   Panel_delete((Object*)schedPanel);
+
+   return HTOP_REFRESH | HTOP_REDRAW_BAR | HTOP_KEEP_FOLLOWING;
+}
+#endif  /* SCHEDULER_SUPPORT */
+
 static Htop_Reaction actionKill(State* st) {
    if (Settings_isReadonly())
       return HTOP_OK;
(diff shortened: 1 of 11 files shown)


Process.c
```diff
@@ -894,6 +895,48 @@ void Process_writeField(const Process* this, RichString* str, ProcessField field
...
+            #ifdef SCHED_DEADLINE
+            case SCHED_DEADLINE:
+               xSnprintf(buffer, n, "DDL ");
+               break;
+            #endif
+            default:
+               CRT_debug("Unexpected scheduling class: %d", this->scheduling_policy);
+               xSnprintf(buffer, n, "??? ");
+               break;
+         }
```
BenBE, line 937: I think as most of the entries can be reasonably well described with 4 or 5 characters, I'd suggest resizing this column accordingly, as I'd find the following mapping more descriptive:

```
SCHED_OTHER    : OTH OTHER
SCHED_FIFO     : FF  FIFO
SCHED_RR       : RR  RR
SCHED_BATCH    : BAT BATCH
SCHED_IDLE     : IDL IDLE
SCHED_DEADLINE : DDL EDF
default        : ??? ???
```

EDF AKA Earliest Deadline First, which is (+CBS) what Linux uses in this case.

Process.c
```diff
@@ -894,6 +895,48 @@ void Process_writeField(const Process* this, RichString* str, ProcessField field
...
+         #else
+         int pure = this->scheduling_policy;
+         #endif
+
+         switch (pure) {
+            case SCHED_OTHER:
+               xSnprintf(buffer, n, "OTHER ");
+               break;
+            case SCHED_FIFO:
+               xSnprintf(buffer, n, "FF    ");
```
BenBE, line 913:
```suggestion
               xSnprintf(buffer, n, "FIFO  ");
```

BenBE (approved):
LGTM.

One minor remark re `CRT_debug_once` in `Process.c:934`: Shouldn't this output one line for each distinct unknown value instead just the first one?

cgzones:
> One minor remark re CRT_debug_once in Process.c:934: Shouldn't this output one line for each distinct unknown value instead just the first one?

That would be an improvement, but requires storing all value already logged about.

Explorer09:
I also have opinions about the current implementation of `CRT_debug_once`: I think different kinds of debug messages should be separated, with a flag for each kind indicating whether the debug message of it should appear. Sharing different kinds of debug messages with one global flag looks like a bad idea, as it limit how the function could be useful.

Process.c
```diff
@@ -894,6 +895,48 @@ void Process_writeField(const Process* this, RichString* str, ProcessField field
...
+               break;
+            case SCHED_RR:
+               xSnprintf(buffer, n, "RR    ");
+               break;
+            #ifdef SCHED_BATCH
+            case SCHED_BATCH:
+               xSnprintf(buffer, n, "BATCH ");
+               break;
+            #endif
+            #ifdef SCHED_BATCH
```
Explorer09, line 923: Is this `#ifdef SCHED_IDLE` or `#ifdef SCHED_BATCH`? Looks like a typo.

Process.h
```diff
@@ -49,6 +50,7 @@ typedef enum ProcessField_ {
    TGID = 52,
    PERCENT_NORM_CPU = 53,
    ELAPSED = 54,
+   SCHEDPOL = 55,
    PROC_COMM = 124,
    PROC_EXE = 125,
```
Explorer09, line 55: Is there any reason you abbreviate `SCHED_POLICY` to `SCHEDPOL`? Because I would like the `SCHED_POLICY` name which is more readable.

Process.c
BenBE: Should be `SCHED_IDLE`, good catch!
```suggestion
            #ifdef SCHED_IDLE
```

Process.h
BenBE: I think skipping the underscore here to avoid intrusion into the `SCHED_` pseudo-namespace, to avoid confusion where this constant belongs to.

Explorer09: @BenBE Good pont about the namespace collision. How about naming it like `SCHEDPOLICY` or `SCHEDULEPOLICY`?

BenBE: `SCHEDULERPOLICY` would be fine by me; though I don't quite care either way.

Action.c
```diff
@@ -407,6 +408,51 @@ static Htop_Reaction actionSetAffinity(State* st) {
...
+   Panel* schedPanel = Scheduling_newPolicyPanel(preSelectedPolicy);
+
+   const ListItem* policy;
+   for(;;) {
+      policy = (const ListItem*) Action_pickFromVector(st, schedPanel, 18, true);
+
+      if (policy && policy->key == -1)
+         Scheduling_togglePolicyPanelResetOnFork(schedPanel);
+      else
+         break;
```
Explorer09, line 428: Slightly more readable if you write like this:
```suggestion
      if (!policy || policy->key != -1)
         break;
      Scheduling_togglePolicyPanelResetOnFork(schedPanel);
```

Scheduling.c
```diff
@@ -0,0 +1,156 @@
...
+      return "???";
+   }
+}
+
+void Scheduling_readProcessPolicy(Process* proc) {
+   proc->scheduling_policy = sched_getscheduler(proc->pid);
+}
+
+#else
+extern int ISO_C_forbids_an_empty_translation_unit;
```
Explorer09, line 155: Do we really need this line?

cgzones: apparently not

PR #1164, opened 2023-01-07
Handle invalid process columns from configuration

cgzones:
Skip over invalid process columns, such that later columns are displayed fine.

Replace invalid sort key identifiers with the default sort column PID.

### DIFF
--- a/Settings.c
+++ b/Settings.c
@@ -262,10 +262,9 @@ static void ScreenSettings_readFields(ScreenSettings* ss, Hashtable* columns, co
       }
       int id = toFieldIndex(columns, ids[i]);
       if (id >= 0)
-         ss->fields[j] = id;
+         ss->fields[j++] = id;
       if (id > 0 && id < LAST_PROCESSFIELD)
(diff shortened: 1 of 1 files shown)


Settings.c
```diff
@@ -261,10 +261,10 @@ static void ScreenSettings_readFields(ScreenSettings* ss, Hashtable* columns, co
          memset(&ss->fields[j], 0, sizeof(ProcessField));
       }
       int id = toFieldIndex(columns, ids[i]);
-      if (id >= 0)
-         ss->fields[j] = id;
-      if (id > 0 && id < LAST_PROCESSFIELD)
-         ss->flags |= Process_fields[id].flags;
+      if (id < 0 || id >= LAST_PROCESSFIELD)
```
BenBE, line 264: This check will fail for dynamically added columns, e.g. with pcp-htop. Also why not make this validity check part of the post-conditions of `toFieldIndex` instead?

BenBE (approved):
LGTM.

PR #1173, opened 2023-01-22
README: affinity syscalls are Linux-specific

9ary:
From https://linux.die.net/man/2/sched_getaffinity:
> ## Conforming To
> These system calls are Linux-specific.

### DIFF
--- a/README
+++ b/README
@@ -92,9 +92,6 @@ To install on the local system run `make install`. By default `make install` ins
     enable Unicode support
     - dependency: *libncursesw*
     - default: *yes*
-  * `--enable-affinity`:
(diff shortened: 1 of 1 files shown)


fasterit:
Seems wrong.
https://cgit.freebsd.org/src/tree/lib/libc/gen/sched_getaffinity.c and
https://cgit.freebsd.org/src/tree/lib/libc/gen/sched_setaffinity.c

9ary:
Right, freebsd implements a lot of Linux APIs for compatibility. In this case, maybe documenting this would be a good idea, I still wouldn't call this generic, unlike libhwloc. Feel free to close otherwise.

Low-power:
It is also available on DragonFly BSD.

PR #1174, opened 2023-01-22
Disable delayacct columns in I/O tab when unsupported

9ary:
Unsupported columns cause every column to their right to be missing. This results in a missing Command column in the default configuration, with no way to add it back short of creating a new tab.

See: https://github.com/NixOS/nixpkgs/pull/212134

### DIFF
--- a/linux/Platform.c
+++ b/linux/Platform.c
@@ -95,7 +95,11 @@ const ScreenDefaults Platform_defaultScreens[] = {
    },
    {
       .name = "I/O",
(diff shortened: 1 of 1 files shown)


9ary:
For what it's worth maybe a more proper fix is to address the underlying issue by ignoring unsupported columns at runtime instead of breaking.

9ary:
Nevermind, I see this was already fixed in 490cedba04aef27c2a3c0bc3a9d6190b05082717.

PR #1176, opened 2023-01-27
Temperature reporting fix for AMD Zen CPUs

JohnsonGlenT:
Reports Package temperature across CPU dies for
AMD Zen CPU's. Temperature is now selected by
package die Tccd[X] for  for temperature reporting
and ignores the Tctl reporting.

Tested and Validated with:
Ryzen 3600 on Artix Linux
Ryzen Threadripper 3970x on Arch Linux

### DIFF
--- a/linux/LibSensors.c
+++ b/linux/LibSensors.c
@@ -243,10 +243,27 @@ void LibSensors_getCPUTemperatures(CPUData* cpus, unsigned int existingCPUs, uns
       /* Check for further adjustments */
    }
 
-   /* Only temperature for core 0, maybe Ryzen - copy to all other cores */
-   if (coreTempCount == 1 && !isnan(data[0])) {
-      for (unsigned int i = 1; i <= existingCPUs; i++)
-         data[i] = data[0];
+
+   /* Check AMD Zen CPUs packages, and mirror temps across packages */
+   if (coreTempCount != existingCPUs) {
+      int coresInDie = existingCPUs / coreTempCount;
+      double temps[coreTempCount];
+      unsigned int count = 0;
+      for (unsigned int i = 0; i <= existingCPUs; i++) {
+         if (!isnan(data[i])) {
+            temps[count] = data[i];
+            count++;
+            if (count == coreTempCount) {
+               break;
+            }
+         }
+      }
+
+      for (unsigned int die = 0; die < coreTempCount; die++) {
+         for (int i = 0; i <= coresInDie; i++) {
+            data[(die*coresInDie)+i] = temps[die];
+         }
+      }
 
       /* No further adjustments */
       goto out;
--- a/linux/LibSensors.c
+++ b/linux/LibSensors.c
@@ -243,8 +243,17 @@ void LibSensors_getCPUTemperatures(CPUData* cpus, unsigned int existingCPUs, uns
       /* Check for further adjustments */
    }
 
+   /* Half the temperatures, probably HT/SMT - copy to second half */
+   const unsigned int delta = activeCPUs / 2;
+   if (coreTempCount == delta) {
+      memcpy(&data[delta + 1], &data[1], delta * sizeof(*data));
+
+      /* No further adjustments */
+      goto out;
+   }
+
    /* Check AMD Zen CPUs packages, and mirror temps across packages */
-   if (coreTempCount != existingCPUs) {
+   if (coreTempCount > 0 && coreTempCount != existingCPUs) {
       double temp[coreTempCount];
       unsigned int count = 0;
       unsigned int coresInDie = existingCPUs / coreTempCount;
@@ -260,25 +269,19 @@ void LibSensors_getCPUTemperatures(CPUData* cpus, unsigned int existingCPUs, uns
          }
       }
 
-      // Set Temperature
-      data[0] = temp[0];
-      for (unsigned int die = 0; die < coreTempCount; die++) {
-         for (unsigned int i = 1; i <= coresInDie; i++) {
-            data[(die*coresInDie)+i] = temp[die];
+      /* Returns to conditional checking if no temp is found  */
+      if (!isnan(temp[0])) {
+         // Set Temperature
+         data[0] = temp[0];
+         for (unsigned int die = 0; die < coreTempCount; die++) {
+            for (unsigned int i = 1; i <= coresInDie; i++) {
+               data[(die * coresInDie) + i] = temp[die];
+            }
          }
+
+         /* No further adjustments */
+         goto out;
       }
-
-      /* No further adjustments */
-      goto out;
-   }
-
-   /* Half the temperatures, probably HT/SMT - copy to second half */
-   const unsigned int delta = activeCPUs / 2;
-   if (coreTempCount == delta) {
-      memcpy(&data[delta + 1], &data[1], delta * sizeof(*data));
-
-      /* No further adjustments */
-      goto out;
    }
 
 out:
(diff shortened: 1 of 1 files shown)


linux/LibSensors.c
```diff
@@ -243,10 +243,30 @@ void LibSensors_getCPUTemperatures(CPUData* cpus, unsigned int existingCPUs, uns
...
+               break;
+            }
+         }
+      }
+
+      // Set Temperature
+      data[0] = temp[0];
+      for (unsigned int die = 0; die < coreTempCount; die++) {
+         for (unsigned int i = 1; i <= coresInDie; i++) {
+            data[(die*coresInDie)+i] = temp[die];
```
BenBE, line 267:
```suggestion
            data[die * coresInDie + i] = temp[die];
```

linux/LibSensors.c
```diff
@@ -243,10 +243,30 @@ void LibSensors_getCPUTemperatures(CPUData* cpus, unsigned int existingCPUs, uns
       /* Check for further adjustments */
    }
 
-   /* Only temperature for core 0, maybe Ryzen - copy to all other cores */
-   if (coreTempCount == 1 && !isnan(data[1])) {
-      for (unsigned int i = 2; i <= existingCPUs; i++)
-         data[i] = data[1];
+   /* Check AMD Zen CPUs packages, and mirror temps across packages */
+   if (coreTempCount != existingCPUs) {
+      double temp[coreTempCount];
```
BenBE, line 248: This is left partially uninitialized if `data` contains `NAN` values.

linux/LibSensors.c
JohnsonGlenT: Change applied in latest commit

linux/LibSensors.c
```diff
@@ -261,6 +252,38 @@ void LibSensors_getCPUTemperatures(CPUData* cpus, unsigned int existingCPUs, uns
...
+         // Set Temperature
+         data[0] = temp[0];
+         for (unsigned int die = 0; die < coreTempCount; die++) {
+            for (unsigned int i = 1; i <= coresInDie; i++) {
+               data[(die * coresInDie) + i] = temp[die];
+            }
+         }
+
+         /* No further adjustments */
+         goto out;
```
BenBE, line 283: This should be one level up, because the special-casing should end here regardless of a global temperature found or not.

JohnsonGlenT: I think that the goto statement should stay here and a fallback conditional should be created for the case where coreTempCount is 1 and data[0] has the only value in it. 

This check can be added into the current conditional statement, however, from looking at all the other conditional statements, we should have that moved out into its own check so that it remains a fallback and is not coupled with the section specified for AMD Zen CPU temperature reporting. (change idea added below)

I may be needlessly worried about a impossible or niche hardware/platform config. 

```suggestion

         /* No further adjustments */
         goto out;
      }
      /* AMD Zen CPU can report coreTempCount as 1 - use fallback conditional */
   }

   /* Fallback for single temperature count reporting */
   if (coreTempCount == 1 && !isnan(data[0])) {
      for (unsigned int i = 1; i <= existingCPUs; i++) {
         data[i] = data[0];

      /* No further adjustments */
      goto out;
   }

```

JohnsonGlenT:
> Hi, I compiled with these patches and it seems to not work correctly.
> 
> This is on a Ryzen 5900X.
> 
> Screenshot_2023-02-27_16-35-01

can you submit the output of sensors

CyberX5:
Oh sorry about that. I was thinking of including that initially.

```
sensors k10temp-pci-00c3

k10temp-pci-00c3
Adapter: PCI adapter
Tctl:         +41.4°C
Tccd1:        +41.2°C
Tccd2:        +38.2°C
```

```
sensors -u k10temp-pci-00c3

k10temp-pci-00c3
Adapter: PCI adapter
Tctl:
  temp1_input: 44.125
Tccd1:
  temp3_input: 37.750
Tccd2:
  temp4_input: 39.250
```

JohnsonGlenT:
@CyberX5 
Can you run and upload the ryzen.log that https://github.com/JohnsonGlenT/htop/tree/ryzen5900X branch creates.

C0rn3j:
Just adding a test result (compiled your fresh logging branch) 

7600X works great - top is current stable, bottom is your branch
image

CyberX5:
> @CyberX5 Can you run and upload the ryzen.log that https://github.com/JohnsonGlenT/htop/tree/ryzen5900X branch creates.

Sure :slightly_smiling_face: [ryzen.log](https://github.com/htop-dev/htop/files/10950816/ryzen.log)

JohnsonGlenT:
@CyberX5 What OS/kernel/Distro are you on?

From the log, your system seems to be reporting coreTempCount as 3. I am assuming it is counting the Tctl, which so far seems to be an issue not with your CPU, but it may be how htop is dealing with polling the k10temp sensors output on older kernels. If this issue will exist on major versions of the kernel, namely LTS, additional changes may need to be made.

CyberX5:
@JohnsonGlenT Well, I do want to apologize it does seem to be an issue on my end, I'm on Arch Linux, 6.1 LTS Kernel, but I run a custom kernel config and I'm guessing that's the problem, now I do keep that in mind so I did test this on the stock Arch kernel before reporting, but must have messed something up...

[ryzen-6.1.19-1-lts.log](https://github.com/htop-dev/htop/files/10963833/ryzen-6.1.19-1-lts.log) Here's the log from the stock Arch LTS kernel. CoreTempCount is 2 now.

Also, I have no idea what config option could be causing this... Haven't seen anything like this in any other program :shrug:

Now, I have encountered a different issue.

Screenshot_2023-03-14_04-43-55

To my understanding the CCD layout of this CPU is 0-5,12-17 = CCD1 - 6-11,18-23 = CCD2.

lstopo for reference:
Screenshot_2023-03-14_04-58-08

I actually have the columns set to 2 for this reason, so visually 1 row would contain both threads of a core. For example thread 0 and 12 would be core 0.

And again sorry about that  :heart:

C0rn3j:
My test log above with stable htop that is mostly broken is from I believe Arch Linux with stock 6.2.2 kernel.

JohnsonGlenT:
> To my understanding the CCD layout of this CPU is 0-5,12-17 = CCD1 and 6-11,18-23 = CCD2.
> 
> lstopo for reference: Screenshot_2023-03-14_04-58-08
> 
> I actually have the columns set to 2 for this reason, so visually 1 row would contain both threads of a core. For example thread 0 and 12 would be core 0.

@CyberX5  This is likely a problem. However, if htop reports CPU based on logical numbering, not physical, then the current temperature readout is correct. I will take a look into this, and will update on my findings.

JohnsonGlenT:
> > @CyberX5 This is likely a problem. However, if htop reports CPU based on logical numbering, not physical, then the current temperature readout is correct. I will take a look into this, and will update on my findings.
> 
> Ah I see, interesting, didn't think of that. slightly_smiling_face

I have confirmed that htop reads CPU number by physical id. I will update when I have a fix that is ready for testing.

JohnsonGlenT:
> > > @CyberX5 This is likely a problem. However, if htop reports CPU based on logical numbering, not physical, then the current temperature readout is correct. I will take a look into this, and will update on my findings.
> > 
> > 
> > Ah I see, interesting, didn't think of that. slightly_smiling_face
> 

> I have confirmed that htop reads CPU number by physical id. I will update when I have a fix that is ready for testing.

[siblist.txt](https://github.com/htop-dev/htop/files/10986106/siblist.txt)

So, as of doing looking with other zen architectures (threadripper 3970X) there seems to be no uniform way to ensure mapping of the CPU thread siblings will receive the same temperature reading, except for polling the thread siblings on every update tick. In this case I think mapping the temp by logical process number is a good enough solution without possibly affecting other systems, or needing to overhaul CPU startup/update loop.

Issue #955, opened 2022-03-14
Wrong memory usage on BSD and derivatives.


dmilith:
So here's an example: https://s.verknowsys.com/1cfc4d97db733029aa928d94c78d0207.png

influxd is using over 48G of RSS, but total memory usage of the system in htop is 28G.

dmilith:
Shouldn't htop count Total Memory as Active + Inactive + Laundry + Wired?

BenBE:
What does `vmstat` report?

Which *BSD port exactly do you use? Or have you confirmed this issue with multiple platforms? If so, which?

dmilith:
All BSDs (and MacOS too) will have a very similar (or the same) memory model, but I use FreeBSD 13.0 arm64/amd64 and MacOS 12.3 arm64.

The `vmstat` on that host shows:

```
 procs    memory    page                      disks     faults       cpu
 r  b  w  avm  fre  flt  re  pi  po   fr   sr nv0 nv1   in   sy   cs us sy id
 1  0  1  27G 4.0G 2.2K 171  78   1 4.4K  14K   0 3212 5593  33K  30K 23 10 67
```
The `top` utility shows this:

```
Mem: 18G Active, 28G Inact, 2028K Laundry, 11G Wired, 4102M Free
```

(where htop shows "23.8G/63.7G" in Mem line)

grahamperrin:
The Reddit post mentioned at <https://github.com/htop-dev/htop/issues/1039#issuecomment-1157206516> includes this comment: 

> … think that it subtracts the arc cache from the actual memory usage when it updates, then it's wraps around backward all the way to the top. …

grahamperrin:
If we're to compare with top(1) <https://www.freebsd.org/cgi/man.cgi?query=top&sektion=1&manpath=FreeBSD>, then a most significant difference might be the perception of what's _free_. 

![image](https://user-images.githubusercontent.com/192271/174495841-b96ff19c-f25f-46c8-977e-3d9faee332f5.png)

* 372 M free according to top
* much more free according to htop (altered according to <https://github.com/htop-dev/htop/issues/1039#issuecomment-1157564395>)

hstk30:
In my MacOS Monterey 12.4 with apple M1 pro, compare `top` and `htop` also have this problem.

<img width="1498" alt="image" src="https://user-images.githubusercontent.com/42362747/175912512-98651c6c-2344-4449-a7ba-e47365a5981d.png">

[...]

PR #1178, opened 2023-02-02
Ignore NaN values when drawing graphs

### LINKED ISSUE #1040: Memory graph flatline on FreeBSD
I'm not sure when this issue began, but I no longer get any reading in the Mem graph: 

image

That is, htop from the FreeBSD repository. Not a custom build. 

```text
% pkg info -x htop
bashtop-0.9.25_1
htop-3.2.1
% pkg -vv | grep -e url -e enabled -e priority
    url             : "pkg+http://pkg0.pkt.freebsd.org/FreeBSD:14:amd64/latest",
    enabled         : yes,
    priority        : 2,
    url             : "https://alpha.pkgbase.live/current/FreeBSD:14:amd64/latest",
    enabled         : no,
    priority        : 0,
    url             : "file:///usr/local/poudriere/data/packages/main-default",
    enabled         : yes,
    priority        : 3
% uname -aKU
FreeBSD mowa219-gjp4-8570p-freebsd 14.0-CURRENT FreeBSD 14.0-CURRENT #13 main-n256147-1c665e95d44-dirty: Wed Jun 15 07:49:44 BST 2022     root@mowa219-gjp4-8570p-freebsd:/usr/obj/usr/src/amd64.amd64/sys/GENERIC-NODEBUG amd64 1400061 1400061
% 
```

er-azh:
Fixes #1040.

The issue was that shared memory is not supported in FreeBSD and thus its value was NaN, When drawing a graph it's adding all the values together and that makes `values` also NaN. I added a check to see if a value is NaN or not when adding it.
This issue must've affected all platforms that don't support shared memory but I did not find any other related issues.

### DIFF
--- a/Meter.c
+++ b/Meter.c
@@ -325,7 +325,7 @@ static void GraphMeterMode_draw(Meter* this, int x, int y, int w) {
 
       double value = 0.0;
       for (uint8_t i = 0; i < this->curItems; i++)
-         value += this->values[i];
+         value += !isnan(this->values[i]) ? this->values[i] : 0;
       data->values[nValues - 1] = value;
    }
 


Meter.c
```diff
@@ -324,8 +324,10 @@ static void GraphMeterMode_draw(Meter* this, int x, int y, int w) {
          data->values[i] = data->values[i + 1];
 
       double value = 0.0;
-      for (uint8_t i = 0; i < this->curItems; i++)
-         value += this->values[i];
+      for (uint8_t i = 0; i < this->curItems; i++) {
+         if (!isnan(this->values[i]))
+            value += this->values[i];
+      }
```
BenBE, line 330: Braces for all or none of these lines; in this case without would be fine.

Alternatively `value += !isnan(this->values[i]) ? this->values[i] : 0;` would also be acceptable.

BenBE (commented):
Change looks reasonable. Please amend your commit with one of the suggested alternatives. TIA.

Explorer09:
Wait. When the shared memory is not supported in FreeBSD, shouldn't it's value be 0.0 and not NaN? We should fix the problem from the source. The meter values should never be negative or NaN.

Explorer09:
And while this PR is about ignoring NaN in meter values, I think negative values should be ignored as well, it's better to fix the thing probably.

Explorer09:
And while this PR is about ignoring NaN in meter values, I think negative values should be igno1red as well, it's better to fix the thing properly.

BenBE:
I agree with the negative values needing proper handling. But IMO `NaN` is perfectly valid for a value to not exist in the first place. Look at it that way: FreeBSD will most likely have some sort of "shared memory" somewhere but we do not know the exact amount. We could assign a value of `0` in this situation but would indicate that we had at least some certainty about its value. Using `NaN` here makes sense in that it encapsulates this "there might be, but the might also be nothing, we don't know" state of affairs of this situation.

And as some values areonly added later on to platforms, you can very well have the situation of some values not actually implemented on one platform and the meters implementation should be able to cope with that situation gracefully. After all, `NaN` is a semantic feature underlying the data, which should be represented properly (i.e. by not flat-lining the things we DO know) …

P.S.: This has intentionally been postponed for 3.3.x to find a proper solution, which would not fit in the time left in the current release schedule …

er-azh:
I also think `NaN` is the best way of representing this especially in FreeBSD. FreeBSD actually supports POSIX memory sharing but we don't know how much shared memory there is since it's not implemented yet. The example below shows the number of shared pages.
```c
#include <stdio.h>
#include <sys/types.h>
#include <sys/sysctl.h>
#include <sys/vmmeter.h>
#include <vm/vm_param.h>

int main() {
  int mib[2];
  struct vmtotal vmtotal;
  size_t size;

  mib[0] = CTL_VM;
  mib[1] = VM_TOTAL;
  size = sizeof(vmtotal);

  if (sysctl(mib, 2, &vmtotal, &size, NULL, 0) < 0) {
    perror("sysctl");
    return 1;
  }

  printf("number of shared pages: %ld\n", vmtotal.t_vmshr);
  return 0;
}
```
And shouldn't negative values be something meters must never return?

I'm also starting to suspect shared memory support might be one of the culprits in causing #955.

Explorer09:
I'm not convinced that NaN may be used here. My impression is that NaN is reserved for arithmetic errors or the situation when "not any value would make sense for this data when drawing the meter/graph".

(When I drafted the code for the graph meter coloring feature (#714), I have considered special casing the NaN values but later decided against it. The existence of NaN values in meters does complicate the drawing as any NaN value in the meter would result in NaN as the total (sum).)

> FreeBSD actually supports POSIX memory sharing but we don't know how much shared memory there is since it's not implemented yet.

Why not just implement it, or hide the "shared memory" field from the meter?

er-azh:
About shared memory, I'll implement it. But we still need a placeholder for values that are not supported or implemented to hide them. As far as I know `NaN` was that replacement. I also don't think it's safe to hide anything that is zero.

Explorer09:
@er-azh 

> About shared memory, I'll implement it. But we still need a placeholder for values that are not supported or implemented to hide them. As far as I know `NaN` was that replacement. I also don't think it's safe to hide anything that is zero.

Meter objects have a `curItems` property that specifies how many items that should be drawn on the bar or graph.
Take MemoryMeter as an example, when `curItems = 4`, the `MEMORY_METER_AVAILABLE` item will not be drawn on the bar or graph.

er-azh:
That won't work unless we swap the value of `MEMORY_METER_SHARED` and `MEMORY_METER_CACHE` since changing `curItems` to `2` will also ignore `MEMORY_METER_CACHE`. and what if another platform doesn't support another value?

https://github.com/htop-dev/htop/blob/623ee31cfe3a39c87e6066c3044c7c38ce676e6b/MemoryMeter.h#L13-L17

BenBE:
Swapping those fields won't work as the order is shared across platforms (and is also used for stacking things in the meter. _So using `NaN` is the way to go here._

@er-azh : You can convert from pages to amount of memory by multiplying by the system's page size (which is traditionally 4KiB). IIRC there's even a constant somewhere in htop (maybe look at the Linux platform code for how it's called) …

natoscott:
>  [...] there's even a constant somewhere in htop

The page size is accessed via a system call - its different for each platform (and even different architectures within one platform) - on FreeBSD it seems to be accessed as:

```
sysctlbyname("vm.stats.vm.v_page_size", &pageSize, &len, NULL, 0)
```

(see freebsd/FreeBSDProcessList.c)

PR #1181, opened 2023-02-03
Add -n option similar to -n option in top

### LINKED ISSUE #377: Option for process list acquisition and data dumping
There should be some option for `htop` to be run in a "CLI one-shot" mode, that only gathers all data for the process list (and maybe meters?) and dumps them to `stdout` in a machine-readable format (bonus points, if its also human-readable).

The dump content provided may depend on some flavour given; like basic (only shown columns), full (gather all columns, even ones normally disabled when not shown), extended (like full, but also including internal data like highlighting offsets).

valdaarhun:
Closes #377 .

While this patch implements the -n option available in `top`, I still haven't implemented the part that allows the output to remain on the screen after `htop` quits.


CommandLine.c
```diff
@@ -176,6 +180,15 @@ static CommandLineStatus parseArguments(const char* program, int argc, char** ar
                return STATUS_ERROR_EXIT;
             }
             break;
+         case 'n':
+            if (sscanf(optarg, "%16d", &(flags->maxIterations)) == 1) {
+               if (flags->maxIterations < 1)
+                  flags->maxIterations = -1;
```
BenBE, line 186: This fails to give a proper error message on invalid numbers put in (like `-n -42`).

Also: the inner `if` should include braces too.

CommandLine.c
```diff
@@ -338,6 +351,8 @@ int CommandLine_run(const char* name, int argc, char** argv) {
 
    if (flags.delay != -1)
       settings->delay = flags.delay;
+   if (flags.maxIterations != -1)
+      settings->maxIterations = flags.maxIterations;
```
BenBE, line 355: Why not put this as part of the `ScreenManager` state directly?

valdaarhun:
Based on what I have read, I don't think it's possible to prevent the screen from getting cleared by using the `ncurses` APIs directly. But I think it can be done by using lower-level APIs that make use of `term` and `terminfo`.

I believe `less` and `vim` do something similar. According to the man page for `less`:

> -X or --no-init
              Disables  sending  the termcap initialization and deinitialization strings to the terminal.  This is sometimes desirable if the deinitialization string does something un‐necessary, like clearing the screen.

Similarly, `vim` by default restores the previous screen on quitting. This can be [prevented](https://vi.stackexchange.com/questions/432/prevent-vim-from-clearing-the-terminal-after-exit#435) by setting `t_te` and `t_ti` accordingly.

What is your take on this?

CommandLine.c
valdaarhun: > This fails to give a proper error message on invalid numbers put in (like `-n -42`)

Instead of throwing an error message, I thought I would simply set `maxIteraitons` to -1 to "disable" the `-n` option. This is somewhat similar to what's been done for the `--delay` option:

```c
if (sscanf(optarg, "%16d", &(flags->delay)) == 1) {
    if (flags->delay < 1)
        flags->delay = 1;
```

valdaarhun: > Also: the inner `if` should include braces too.

Sure, I'll fix that.

CommandLine.c
valdaarhun: That can be done too. I never thought of that.

BenBE:
There was some feature request to print the acquired data to the screen and exit immediately, cf. #377 …

BTW: You are free to do code rebasing/amending/squashing on your branch as much as you like to keep the commits of the PR as clean and straight forward as possible.

valdaarhun:
> There was some feature request to print the acquired data to the screen and exit immediately, cf. #377 …

I took a look at that issue. I think I'll open another PR for that issue since the requirements are slightly different.

valdaarhun:
Hi. I think I have understood how to let htop's output remain on the screen after quitting. After a lot more reading, I found this [comment](https://www.reddit.com/r/C_Programming/comments/n1310b/comment/gwbcuty/?utm_source=share&utm_medium=web2x&context=3) on reddit. I started going through `dialog`'s codebase and found some interesting stuff. (I am still not very sure about the portability of this solution with other terminals though).

Based on what I have understood, an ncurses application makes use of an alternate screen where output is displayed. After the application stops running, the original screen is restored and the contents of the alternate screen is lost. By default, in most terminals, `initscr` uses a terminal capability called `enter_ca_mode` which makes the application use an alternate screen.

`vim` and `less` both have options to either use the alternate screen or the same screen. `vim` by default uses an alternate screen which causes the original screen to be restored after quitting it. `:set t_ti= t_te` prevents the original screen from being restored. Similarly, `less` also uses the alternate screen by default. Using the `-X` option prevents the original screen from being restored.

In case of `htop`, the `enter_ca_mode` is written to the terminal when `initscr` is called at the beginning of the TUI setup. This enables the alternate screen. We will need to disable this manually by writing `exit_ca_mode` to the terminal immediately after `initscr`. This can't be done directly using the ncurses API and we'll need to use lower-level terminfo APIs to implement this.

_Refs_:
- https://jameshfisher.com/2017/12/04/how-less-works/
- https://www.linuxquestions.org/questions/linux-software-2/xterm-alternate-screen-and-hidden-lines-when-scrolling-back-to-main-screen-4175681241-print/
- https://www.reddit.com/r/C_Programming/comments/n1310b/comment/gwbcuty/?utm_source=share&utm_medium=web2x&context=3
- https://vimhelp.org/term.txt.html
- https://www.man7.org/linux/man-pages/man5/terminfo.5.html
- https://www.man7.org/linux/man-pages/man3/del_curterm.3x.html

CRT.c
```diff
@@ -22,6 +22,9 @@ in the source distribution for its full text.
 #include "ProvideCurses.h"
 #include "XUtils.h"
 
+#include <ncurses.h>
```
BenBE, line 25: Because (n)curses support is cursed, there's the `ProvideCurses.h` header available, which internally switches the various places where the `curses`/`ncurses` headers (with and without wide char support) can live. Thus please don't include `ncurses.h` directly: It will break builds. ;-)

CRT.c
```diff
@@ -22,6 +22,9 @@ in the source distribution for its full text.
...
+#include <term.h>
```
BenBE, line 26: Same goes for `term.h`: Please extend `ProvideCurses.h` accordingly instead. There should be defines available from `configure.ac` …

BenBE (changes requested):

CRT.c
valdaarhun: I don't think `configure.ac` currently has defines for `term.h`. Also, I couldn't find any configuration options in the [documentation](https://www.gnu.org/software/autoconf-archive/The-Macros.html) that are to be used for `term.h`. I could only find options for `curses.h`, and its extensions.

https://www.gnu.org/software/autoconf-archive/ax_with_curses.html
https://www.gnu.org/software/autoconf-archive/ax_with_curses_extra.html

BenBE: There actually should be a check for `term.h` be added to `configure.ac` (near where it looks for `ncurses`/`ncursesw`. Also `term.h` should then be added through `ProvideCurses.h`. In the folder `iwyu` the ncurses rules need to be updated accordingly to declare `ProvideCurses.h` is providing the respective symbols from `term.h` instead.

BenBE: Okay. Sounds good.

valdaarhun: For example, when `term.h` checks are kept in `ProvideCurses.h`, the following compilation error gets thrown:

```
In file included from ProvideCurses.h:33,
                 from CRT.h:15,
                 from CRT.c:12:
HeaderLayout.h:35:12: error: expected ‘)’ before ‘*’ token
   35 |    uint8_t columns;
      |            ^~~~~~~
```

This is because `columns` is a defined variable in the `term.h` header file but refers to another variable in `HeaderLayout.h`

I thought of using `#ifdef` guards to make `term.h` get included only in the `CRT.c` file. But that gives issues as well. Suppose, I use a macro `#ifdef INCLUDE_TERM` and use that in `ProvideCurses.h` to include `term.h` only when required.

In `CRT.h`, the header files are included as shown below:
```
1    #include "config.h" // IWYU pragma: keep
2    
3    #include "CRT.h"
4    
5    #include <errno.h>
6    #include <fcntl.h>
7    #include <langinfo.h>
8    #include <signal.h>
9    #include <stdarg.h>
10   #include <stdio.h>
11   #include <stdlib.h>
12   #include <string.h>
13   #include <unistd.h>
14    
15   #include "ProvideCurses.h"
16   #include "XUtils.h"
```

If I insert `#define INCLUDE_TERM` on line 14, it won't have any effect because `ProvideCurses.h` is included implicitly through `CRT.h`. If I insert the include statement before line 3, it'll will clash with other header files that are implicitly included through `CRT.h`.

BenBE: As said in my previous note, splitting this into `ProvideCurses.h` and `ProvideTerm.h` is fine if that makes things easier.

iwyu/htop.imp
```diff
@@ -4,6 +4,9 @@
     { include: ["<ncurses/curses.h>", "private", "\"ProvideCurses.h\"", "public"] },
     { include: ["<ncurses/ncurses.h>", "private", "\"ProvideCurses.h\"", "public"] },
 
+    { include: ["<term.h>", "private", "\"ProvideCurses.h\"", "public"] },
+    { include: ["<ncurses/term.h>", "private", "\"ProvideCurses.h\"", "public"] },
```
BenBE, line 8:
```suggestion
    { include: ["<term.h>", "private", "\"ProvideTerm.h\"", "public"] },
    { include: ["<ncurses/term.h>", "private", "\"ProvideTerm.h\"", "public"] },
```

configure.ac
```diff
@@ -359,13 +359,19 @@ else
...
 
    # check if additional linker flags are needed for keypad(3)
    # (at this point we already link against a working ncurses library)
    AC_SEARCH_LIBS([keypad], [tinfo])
 fi
+
+AC_CHECK_HEADERS([ncursesw/term.h], [],
+   [AC_CHECK_HEADERS([ncurses/term.h], [],
+      [AC_CHECK_HEADERS([term.h], [],
+         [AC_MSG_ERROR([can not find required term header file])])])])
```
BenBE, line 373: Is there a distribution providing `ncursesw/term.h` without Unicode support? If there is not this should be moved up to the above `AC_ARG_ENABLE` check (without `ncursesw/term.h` if no Unicode support was requested.

BenBE:
You also need to include `ProvideTerm.h` in `Makefile.am` in order for `make distcheck` to succeed.

configure.ac
valdaarhun: I couldn't find any info on `ncursesw/term.h` with no Unicode support. But, I think all ncurses-related libraries will be in the same lib folder. So if `ncursesw/ncurses` is in the `AC_ARG_ENABLE` check, then probably the check for `ncursesw/term.h` should be there too. Do let me know I am wrong.

ProvideTerm.h
```diff
@@ -0,0 +1,18 @@
+#ifndef HEADER_ProvideTerm
+#define HEADER_ProvideTerm
+
+#include "config.h"
```
BenBE, line 4:
```suggestion
#include "config.h" // IWYU pragma: keep
```

ScreenManager.c
```diff
@@ -229,6 +229,7 @@ void ScreenManager_run(ScreenManager* this, Panel** lastFocus, int* lastKey, con
    bool rescan = false;
    int sortTimeout = 0;
    int resetSortTimeout = 5;
+   int maxIterations = this->settings->maxIterations;
```
BenBE, line 232: I'm not sure if this variable really makes things easier to read.

CommandLine.c
```diff
@@ -176,6 +180,17 @@ static CommandLineStatus parseArguments(const char* program, int argc, char** ar
...
             }
             break;
+         case 'n':
+            if (sscanf(optarg, "%16d", &(flags->maxIterations)) == 1) {
+               if (flags->maxIterations <= 0) {
+                  fprintf(stderr, "Error: number of iterations should be at least one.\n");
+                  return STATUS_ERROR_EXIT;
+               }
+            } else {
+               fprintf(stderr, "Error: invalid value for number of iterations \"%s\".\n", optarg);
```
BenBE, line 190:
```suggestion
               fprintf(stderr, "Error: invalid maximum iteration count \"%s\".\n", optarg);
```

BenBE (changes requested):
I did some cleanup in your commits and also did some re-ordering of things. Will have to test if things work properly.

Also changed the visibility of some variables that are only used in the same compilation unit and thus do not need to have public visibility.

ScreenManager.c
valdaarhun: `maxIterations` is being used thrice in the `if` block on line 244. I think it'll become cluttered if `this->settings->maxIterations` is used in all three places.

But maybe, line 232 can be placed between the `if` block and `force_redraw = false`. What are your thoughts on this?

BenBE: First off, `maxIterations` is a bad name for how the variable currently is used. And unless the value is stored back into the setting, this doesn't help with reducing the scope.

If you, on the second note though, just use the fully qualified reference to the field in the settings, you can skip line 232 entirely and just need the three lines below. You could even put the decrement and the second check into one line:

```c
if (this->settings->maxIterations != -1) {
   if (!--this->settings->maxIterations) {
      // Exit the program
   }
}
```

Regarding the name: while the setting is called `max-iterations` the internal field could better be called `iterationsLeft` or something along those lines.

BenBE (commented):
Not sure if having something volatile, that is not actually stored to disk, should be part of the settings.

@cgzones @fasterit @natoscott : What do you guys think?

fasterit:
> > @cgzones @fasterit @natoscott : What do you guys think?
> 
> This reminds me of the recent discussion about 'readonly' - in that case the preference was to have a global variable separate to Settings. This seems sufficiently similar to that case that its probably the way to go here too.

Seconded!
/DLange

ScreenManager.c
valdaarhun: Hi. You're right. `iterationsLeft` is a better name given the way it is being handled. I'll change this.

valdaarhun:
> Not sure if having something volatile, that is not actually stored to disk, should be part of the settings.

Hi. Sorry for the late reply. I think I have understood what this means. I have made some changes based on what I have understood. Please let me know if something else was expected.

BenBE (changes requested):
Already looks much better.

Please cleanup the patchset's commits a bit. The changes from commits 1, 2, 5, and 6 should all fit together nicely in one commit (no need to split them this much). As commit 5 is using `ProvideTerm.h`, you can place the squashed after 3 and 4:

```
3
4
1+2+5+6+7
```

(7 being the additional changes re `extern`)

CRT.c
```diff
@@ -793,12 +794,16 @@ static int CRT_colorSchemes[LAST_COLORSCHEME][LAST_COLORELEMENT] = {
...
 
+static bool retainScreenOnExit = false;
+
 int CRT_scrollHAmount = 5;
 
 int CRT_scrollWheelVAmount = 10;
 
 ColorScheme CRT_colorScheme = COLORSCHEME_DEFAULT;
 
+extern int iterationsRemaining;
```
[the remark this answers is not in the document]
natoscott: Another option might be to store it in the new struct Machine, alongside the other sample-time related fields there (e.g. realtime, realtimeMs, monotonicMs)?  That way its accessible in all the places that need it.

valdaarhun:
> Already looks much better.
> 
> Please cleanup the patchset's commits a bit. The changes from commits 1, 2, 5, and 6 should all fit together nicely in one commit (no need to split them this much). As commit 5 is using `ProvideTerm.h`, you can place the squashed between 3 and 4:
> 
> ```
> 3
> 1+2+5+6+7
> 4'
> ```
> 
> (7 being the additional changes re `extern`, 4' as it will need small modifications)
> 
> P.S.: Note that commits should compile on their own though. 3 best should stay separate, mix and match the others as needed.

Sure thing. I'll rearrange the commits after making the final changes in the code.

CRT.c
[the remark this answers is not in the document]
BenBE: @natoscott I'm not sure if `CRT.c` has access to the `Machine` structure (nor if it should). AFAICS this is currently not available there (and I'm not really convinced that we should change that).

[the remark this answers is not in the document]
valdaarhun: Based on what I have understood, the `Machine` structure is being used to keep track of machine-related stats such as memory usage and active CPUs.

> I'm not sure if CRT.c has access to the Machine structure

I concur. Currently, CRT.c doesn't seem to have access to the `Machine` struct but including Machine.h in CRT.c should fix that.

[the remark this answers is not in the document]
natoscott: @BenBE yeah, you're right - its definitely there where we need it in ScreenManager but not in Settings (and agree, its not the right thing to do adding it there).

@valdaarhun struct Machine has information about sampling too, which is what 'iterationsRemaining' is, so it could find a logical home there.

| static bool retainScreenOnExit = false;

This variable name for consistency with many other globals in CRT.c should be prefixed with CRT_ I guess.  I think the cleanest approach might be to have another flag passed into CRT_init, alongside allowUnicode, like 'retainScreen' to set the CRT.c global, and maintain the 'remainingIterations' counter within struct Machine alongside the sample timing fields there, for ScreenManager to update.  But, just my 2c... there's plenty of other ways too, I'm not strongly of any one opinion here.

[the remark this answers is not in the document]
valdaarhun: > struct Machine has information about sampling too, which is what 'iterationsRemaining' is, so it could find a logical home there.

Ah, ok. I have understood this now.

> I think the cleanest approach might be to have another flag passed into CRT_init, alongside allowUnicode, like 'retainScreen' to set the CRT.c global, and maintain the 'remainingIterations' counter within struct Machine alongside the sample timing fields there, for ScreenManager to update

I think this approach is quite neat.

Machine.h
```diff
@@ -45,6 +45,8 @@ typedef struct Machine_ {
    struct timeval realtime;   /* time of the current sample */
    uint64_t realtimeMs;       /* current time in milliseconds */
    uint64_t monotonicMs;      /* same, but from monotonic clock */
+   
+   int iterationsRemaining;
```
valdaarhun, line 49: I am not sure if this should be of type `int64_t` or if `int` is good enough.

BenBE: The bit-sized types are preferred, thus I'd suggest `int64_t`.

BenBE:
The changes from that last update look really good.

What do you think about grouping them into the following three commits like this:
1. Configure changes (`ProvideTerm.h`)
2. Actual counter logic + `-n` option
3. Retain screen logic related changes

valdaarhun:
> The changes from that last update look really good.
> 
> What do you think about grouping them into the following three commits like this:
> 
> 1. Configure changes (`ProvideTerm.h`)
> 2. Actual counter logic + `-n` option
> 3. Retain screen logic related changes

This looks good to me. I'll clean up the patchset.

valdaarhun:
Hi. I have cleaned up the patchset as best I could. Please let me know if additional changes are required.

BenBE:
Looks good and clean.

Last thing I noticed: in the first commit the iwyu update is mentioned twice. But the configure changes for detecting term.h are not.

Once you fixed that, we are good to go.

valdaarhun:
@BenBE Sorry, I must have missed that. I have changed the description now.

BenBE (approved):

valdaarhun:
I enjoyed working on this feature. Thanks for the help and feedback!

PR #1185, opened 2023-02-04
Annotate functions with access attribute

cgzones:
Supported by GCC since version 10.

### DIFF
--- a/Macros.h
+++ b/Macros.h
@@ -61,6 +61,27 @@
 
 #endif /* HAVE_ATTR_ALLOC_SIZE */
 
+#ifdef HAVE_ATTR_ACCESS
+
+#define ATTR_ACCESS2(mode, ref)         __attribute__((access (mode, ref)))
+#define ATTR_ACCESS3(mode, ref, size)   __attribute__((access (mode, ref, size)))
+
+#else
+
+#define ATTR_ACCESS2(mode, ref)
+#define ATTR_ACCESS3(mode, ref, size)
+
+#endif /* HAVE_ATTR_ACCESS */
+
+#define ATTR_ACCESS2_R(ref)              ATTR_ACCESS2(read_only, ref)
+#define ATTR_ACCESS3_R(ref, size)        ATTR_ACCESS3(read_only, ref, size)
+
+#define ATTR_ACCESS2_RW(ref)             ATTR_ACCESS2(read_write, ref)
+#define ATTR_ACCESS3_RW(ref, size)       ATTR_ACCESS3(read_write, ref, size)
+
+#define ATTR_ACCESS2_W(ref)              ATTR_ACCESS2(write_only, ref)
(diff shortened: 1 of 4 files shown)


Macros.h
```diff
@@ -61,6 +61,18 @@
 
 #endif /* HAVE_ATTR_ALLOC_SIZE */
 
+#ifdef HAVE_ATTR_ACCESS
+
+#define ATTR_ACCESS2(mode, ref)         __attribute__((access (mode, ref)))
+#define ATTR_ACCESS3(mode, ref, size)   __attribute__((access (mode, ref, size)))
```
BenBE, line 67: Why not pre-define `ATTR_ACCESS_R`, `ATTR_ACCESS_RW`, and `ATTR_ACCESS_W`?

Explorer09: I can see why you define two versions of the macro `ATTR_ACCESS2`, `ATTR_ACCESS3`, but consider using `__VA_ARGS__`.

```suggestion
// Usage: ATTR_ACCESS(mode, refIndex, sizeIndex) while sizeIndex is optional
#define ATTR_ACCESS(mode, ...)          __attribute__((access (mode, __VA_ARGS__)))
```

Or, as @BenBE suggested:

```suggestion
// Usage: ATTR_ACCESS_X(refIndex, sizeIndex) while sizeIndex is optional
#define ATTR_ACCESS_R(...)              __attribute__((access (read_only, __VA_ARGS__)))
#define ATTR_ACCESS_RW(...)             __attribute__((access (read_write, __VA_ARGS__)))
#define ATTR_ACCESS_W(...)              __attribute__((access (write_only, __VA_ARGS__)))
```

BenBE: I'd prefer to have 6 macros instead, as the `__VA_ARGS__` semantic is not intended and misleading in this case. Yes, that's some more macros, but clarity is better in this case. I even could see a reason to keep all 8 variants:
```suggestion
#define ATTR_ACCESS2(mode, ref)         __attribute__((access (mode, ref)))
#define ATTR_ACCESS3(mode, ref, size)   __attribute__((access (mode, ref, size)))

#define ATTR_ACCESS2_R(ref)         ATTR_ACCESS2(read_only, ref)
#define ATTR_ACCESS3_R(ref, size)   ATTR_ACCESS3(read_only, ref, size)

#define ATTR_ACCESS2_RW(ref)         ATTR_ACCESS2(read_write, ref)
#define ATTR_ACCESS3_RW(ref, size)   ATTR_ACCESS3(read_write, ref, size)

#define ATTR_ACCESS2_W(ref)         ATTR_ACCESS2(write_only, ref)
#define ATTR_ACCESS3_W(ref, size)   ATTR_ACCESS3(write_only, ref, size)
```

BenBE (approved):
LGTM.

PR #1186, opened 2023-02-04
Misc cleanup

cgzones:
* declare read-only pointer parameter const
* drop duplicate include of <errno.h>
* use String_startsWith

### DIFF
--- a/ScreenManager.c
+++ b/ScreenManager.c
@@ -153,7 +153,7 @@ static void checkRecalculation(ScreenManager* this, double* oldTime, int* sortTi
    *rescan = false;
 }
 
-static inline bool drawTab(int* y, int* x, int l, const char* name, bool cur) {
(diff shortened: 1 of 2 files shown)


ScreenManager.c
```diff
@@ -153,7 +153,7 @@ static void checkRecalculation(ScreenManager* this, double* oldTime, int* sortTi
    *rescan = false;
 }
 
-static inline bool drawTab(int* y, int* x, int l, const char* name, bool cur) {
+static inline bool drawTab(const int* y, int* x, int l, const char* name, bool cur) {
```
BenBE, line 156: Shouldn't `x` be `const` too?

BenBE (approved):

ScreenManager.c
cgzones: All tab titles are in the same row, so `y` is const. But for writing the titles the row, `x`, has to be advanced.

PR #1189, opened 2023-02-13
Rework some methods of Vector

### LINKED ISSUE #1188: assertion in Vector_checkArraySize fails when setting the core affinity
The function  `Vector_checkArraySize` seems to be originally designed to be called when the vector is full (hence the `assert(Vector_isConsistent())`). This works fine, when `Vector_checkArraySize` is called every time **after** adding (exactly one) element. This is the case in `Vector_set`

https://github.com/htop-dev/htop/blob/e3481a9846ef01bb27c169b71eafeee704e10c68/Vector.c#L203-L214

However, the function `Vector_splice`uses this function to reallocate for (possibly) more than one element. 

https://github.com/htop-dev/htop/blob/e3481a9846ef01bb27c169b71eafeee704e10c68/Vector.c#L396-L406

When the new size is larger than the current array size, the assertion `Vector_isConsistent` in `Vector_checkArraySize` fails.

This can be triggered by pressing `a` (setting the core affinity)  in a debug build of htop because `Vector_splice` is called then. 
However, the assertion does probably only trigger on machines with more than 10 cores (this is the default vector size).

lbckmnn:
closes #1188.

reworks some methods of Vector.

See #1188 for details.

### DIFF
--- a/Vector.c
+++ b/Vector.c
@@ -62,14 +62,6 @@ static bool Vector_isConsistent(const Vector* this) {
    assert(this->items <= this->arraySize);
    assert(!Vector_isDirty(this));
 
-   if (this->owner) {
-      for (int i = 0; i < this->items; i++) {
-         if (!this->array[i]) {
-            return false;
-         }
-      }
-   }
-
    return true;
 }
 
@@ -396,7 +388,7 @@ int Vector_indexOf(const Vector* this, const void* search_, Object_Compare compa
 void Vector_splice(Vector* this, Vector* from) {
    assert(Vector_isConsistent(this));
    assert(Vector_isConsistent(from));
-   assert(!(this->owner && from->owner));
+   assert(!this->owner);
 
    int olditems = this->items;
    this->items += from->items;
--- a/Vector.c
+++ b/Vector.c
@@ -192,12 +192,12 @@ void Vector_insertionSort(Vector* this) {
    assert(Vector_isConsistent(this));
 }
 
-static void Vector_checkArraySize(Vector* this) {
-   assert(Vector_isConsistent(this));
-   if (this->items >= this->arraySize) {
-      //int i;
-      //i = this->arraySize;
-      this->arraySize = this->items + this->growthRate;
+static void Vector_resizeIfNecessary(Vector* this, int newSize) {
+   assert(newSize >= 0);
+   if (newSize > this->arraySize) {
+      assert(Vector_isConsistent(this));
+      int oldSize = this->arraySize;
+      this->arraySize = newSize + this->growthRate;
       this->array = (Object**) xRealloc(this->array, sizeof(Object*) * this->arraySize);
       //for (; i < this->arraySize; i++)
       //   this->array[i] = NULL;
@@ -215,7 +215,7 @@ void Vector_insert(Vector* this, int idx, void* data_) {
       idx = this->items;
    }
 
-   Vector_checkArraySize(this);
+   Vector_resizeIfNecessary(this, this->items + 1);
    //assert(this->array[this->items] == NULL);
    if (idx < this->items) {
       memmove(&this->array[idx + 1], &this->array[idx], (this->items - idx) * sizeof(this->array[0]));
@@ -331,14 +331,15 @@ void Vector_set(Vector* this, int idx, void* data_) {
    assert(Object_isA(data, this->type));
    assert(Vector_isConsistent(this));
 
-   Vector_checkArraySize(this);
+   Vector_resizeIfNecessary(this, idx + 1);
    if (idx >= this->items) {
       this->items = idx + 1;
    } else {
       if (this->owner) {
          Object* removed = this->array[idx];
-         assert (removed != NULL);
-         Object_delete(removed);
+         if (removed != NULL) {
+            Object_delete(removed);
+         }
       }
    }
    this->array[idx] = data;
@@ -391,8 +392,8 @@ void Vector_splice(Vector* this, Vector* from) {
    assert(!this->owner);
 
    int olditems = this->items;
+   Vector_resizeIfNecessary(this, this->items + from->items);
    this->items += from->items;
-   Vector_checkArraySize(this);
    for (int j = 0; j < from->items; j++) {
       this->array[olditems + j] = from->array[j];
    }
--- a/Vector.c
+++ b/Vector.c
@@ -95,13 +95,13 @@ void Vector_prune(Vector* this) {
       for (int i = 0; i < this->items; i++) {
          if (this->array[i]) {
             Object_delete(this->array[i]);
-            this->array[i] = NULL;
          }
       }
    }
    this->items = 0;
    this->dirty_index = -1;
    this->dirty_count = 0;
+   memset(this->array, '\0', this->arraySize * sizeof(Object *));
 }
 
 //static int comparisons = 0;
@@ -198,9 +198,7 @@ static void Vector_resizeIfNecessary(Vector* this, int newSize) {
       assert(Vector_isConsistent(this));
       int oldSize = this->arraySize;
       this->arraySize = newSize + this->growthRate;
-      this->array = (Object**) xRealloc(this->array, sizeof(Object*) * this->arraySize);
-      //for (; i < this->arraySize; i++)
-      //   this->array[i] = NULL;
+      this->array = (Object **)xReallocArrayZero(this->array, oldSize, this->arraySize, sizeof(Object*));
    }
    assert(Vector_isConsistent(this));
 }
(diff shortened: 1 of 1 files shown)


cgzones (approved):

lbckmnn:
@cgzones should i still implement the suggested changes from the #1188 ?

Vector.c
```diff
@@ -293,13 +285,19 @@ void Vector_compact(Vector* this) {
...
       memmove(&this->array[idx], &this->array[idx + 1], (this->items - idx - 1) * sizeof(this->array[0]));
+      this->array[this->items - 1] = NULL;
    } else {
       /* multiple deletions */
       for (int i = idx + 1; i < size; i++) {
          if (this->array[i]) {
             this->array[idx++] = this->array[i];
          }
       }
+      /* idx is now at the end of the vector and on the first index which should be set to NULL */
```
BenBE, line 296: -
```suggestion
      // idx is now at the end of the vector and on the first index which should be set to NULL
```

Also note the brace style in the `for` loop below.

Vector.c
```diff
@@ -200,15 +192,15 @@ void Vector_insertionSort(Vector* this) {
...
-      //int i;
-      //i = this->arraySize;
-      this->arraySize = this->items + this->growthRate;
+   if (newSize > this->arraySize) {
+      int oldSize = this->arraySize;
+      this->arraySize = newSize + this->growthRate;
       this->array = (Object**) xRealloc(this->array, sizeof(Object*) * this->arraySize);
-      //for (; i < this->arraySize; i++)
-      //   this->array[i] = NULL;
+      for (int i = oldSize; i < this->arraySize; i++)
```
BenBE, line 202: Maybe adding braces here?

BenBE (commented):
LGTM, some minor formatting notes.

Vector.c
lbckmnn: Should i change the style of the other comments in this function too?

Vector.c
```diff
@@ -290,16 +283,22 @@ void Vector_compact(Vector* this) {
...
-      /* multiple deletions */
+      // multiple deletions
       for (int i = idx + 1; i < size; i++) {
          if (this->array[i]) {
             this->array[idx++] = this->array[i];
          }
       }
+      // idx is now at the end of the vector and on the first index which should be set to NULL
+      for (int i = idx; i < size; i++)
+      {
```
BenBE, line 299: Should be one line too …
```suggestion
      for (int i = idx; i < size; i++) {
```

Vector.c
```diff
@@ -200,15 +192,16 @@ void Vector_insertionSort(Vector* this) {
...
-      //int i;
-      //i = this->arraySize;
-      this->arraySize = this->items + this->growthRate;
+   if (newSize > this->arraySize) {
+      int oldSize = this->arraySize;
+      this->arraySize = newSize + this->growthRate;
       this->array = (Object**) xRealloc(this->array, sizeof(Object*) * this->arraySize);
-      //for (; i < this->arraySize; i++)
-      //   this->array[i] = NULL;
+      for (int i = oldSize; i < this->arraySize; i++) {
```
cgzones, line 202: I think we can use `xReallocArrayZero(this->array, oldSize, this->arraySize, sizeof(Object*))` here

Vector.c
```diff
@@ -62,14 +62,6 @@ static bool Vector_isConsistent(const Vector* this) {
...
    assert(!Vector_isDirty(this));
 
-   if (this->owner) {
-      for (int i = 0; i < this->items; i++) {
-         if (!this->array[i]) {
-            return false;
-         }
-      }
-   }
-
```
cgzones, line 72: We could count the non empty slots:
```
int items = 0;
for (int i = 0; i < this->arraySize; i++)
   if (this->array[i])
      items++;
assert(this->items == items);
```

This requires `memset(this->array, '\0', this->arraySize * sizeof(Object *));` in `Vector_prune()`

Vector.c
cgzones: or even `memset(&this->array[idx], '\0', (size - idx) * sizeof(this->array[0]));`?

Vector.c
lbckmnn: I don't think that this will work.

`this->items` does currently just contain the actual length of the vector and not the number of used slots.

Look for example at `Vector_set()`:
`Vector_set()` allows calls with `idx` >`this->items`. In this case `this->items` is just set to `idx +1`. 
`this->items` would then also include empty slots and the assertion would fail.

lbckmnn: @cgzones any thoughts on this?

cgzones (approved):

Vector.c
BenBE: @cgzones There are open questions that you might want to take a look at.

Vector.c
```diff
@@ -200,15 +192,13 @@ void Vector_insertionSort(Vector* this) {
...
-      //int i;
-      //i = this->arraySize;
-      this->arraySize = this->items + this->growthRate;
-      this->array = (Object**) xRealloc(this->array, sizeof(Object*) * this->arraySize);
-      //for (; i < this->arraySize; i++)
-      //   this->array[i] = NULL;
+   if (newSize > this->arraySize) {
+      int oldSize = this->arraySize;
+      this->arraySize = newSize + this->growthRate;
+      this->array =(Object **) xReallocArrayZero(this->array, oldSize, this->arraySize, sizeof(Object*));
```
BenBE, line 201:
```suggestion
      this->array = (Object **)xReallocArrayZero(this->array, oldSize, this->arraySize, sizeof(Object*));
```
Minor code style note …

Vector.c
```diff
@@ -339,14 +332,15 @@ void Vector_set(Vector* this, int idx, void* data_) {
...
-   Vector_checkArraySize(this);
+   Vector_resizeIfNecessary(this, idx + 1);
    if (idx >= this->items) {
       this->items = idx + 1;
    } else {
       if (this->owner) {
          Object* removed = this->array[idx];
-         assert (removed != NULL);
-         Object_delete(removed);
+         if(removed != NULL) {
```
BenBE, line 341:
```suggestion
         if (removed != NULL) {
```

Vector.c
cgzones: > I don't think that this will work.
>
>this->items does currently just contain the actual length (or in other words the first unused index) of the array and >not the number of used slots.
>
>Look for example at Vector_set():
>Vector_set() allows calls with idx >this->items. In this case this->items is just set to idx +1.
>this->items would then also include empty slots and the assertion would fail.

True, so disregard my suggestion.

> The memset in Vector_prune() might be a good idea anyway.

Could you add this?

cgzones:
I think https://github.com/htop-dev/htop/issues/1188#issuecomment-1440886128 should be addressed as well:

in the current implementation `Vector_splice` should probably only take non-owning vectors as source (e.g. `assert(!this->owner);`, since this should be satisfied by its only caller (the 7th parameter is false): https://github.com/htop-dev/htop/blob/e207c8aebdcdb88bc8ab838e2ac3dd1774d6a618/AffinityPanel.c#L363

On a side note: the whole functions `Vector_splice` just asks for memory corruptions (double-free,...), luckily its only used once.

Vector.c
lbckmnn: is there any way to check the formatting? 
The astyle command from the style guide results in another different format:
`this->array = (Object**)xReallocArrayZero(this->array, oldSize, this->arraySize, sizeof(Object*));`

lbckmnn:
> I think [#1188 (comment)](https://github.com/htop-dev/htop/issues/1188#issuecomment-1440886128) should be addressed as well:
> 
> in the current implementation `Vector_splice` should probably only take non-owning vectors as source (e.g. `assert(!this->owner);`, since this should be satisfied by its only caller (the 7th parameter is false):
> 
> https://github.com/htop-dev/htop/blob/e207c8aebdcdb88bc8ab838e2ac3dd1774d6a618/AffinityPanel.c#L363
> 
> On a side note: the whole functions `Vector_splice` just asks for memory corruptions (double-free,...), luckily its only used once.

Yes, that is true. 
I adjusted the assertions in `Vector_splice`

Vector.c
lbckmnn: done

Vector.c
```diff
@@ -200,15 +192,13 @@ void Vector_insertionSort(Vector* this) {
    assert(Vector_isConsistent(this));
 }
 
-static void Vector_checkArraySize(Vector* this) {
+static void Vector_resizeIfNecessary(Vector* this, int newSize) {
+   assert(newSize >= 0);
    assert(Vector_isConsistent(this));
```
BenBE, line 197: Minor optimization: Only check `Vector_isConsistent` once the resizing is about to happen.

Thus either leave this line and move the one below to the end of the `if` statement, or move this one to the start of the `if` block.

BenBE (approved):
Small optimization. Apart from this LGTM.

Vector.c
lbckmnn: good catch

cgzones (approved):

cgzones:
Thanks, merged.

Issue #1195, opened 2023-02-24
[Regression in 3.2.1] Wrong CPU threads are shown when running in a container


C0rn3j:
I have a host server with 48 threads running a container using LXC 5.0.2 where I am executing htop.

This container has the following LXC config set: `limits.cpu: "16"`, meaning it gets assigned 16 random threads.

htop 3.0.5 shows the threads and their usage correctly:
![image](https://user-images.githubusercontent.com/1641362/221320992-7ef133ef-6f14-43bd-9ff1-eb9d2cb9c31e.png)

htop 3.1.0 to 3.2.0 show things correctly while showing the unused host CPU threads too:
![image](https://user-images.githubusercontent.com/1641362/221321043-83e77964-1685-42c1-bd3d-51a428523da0.png)

htop 3.2.1 to 3.2.2 however seem to show completely wrong threads by simply showing threads 1 to 16 as they go in order, irrelevant of the actual assignment. This is obvious when comparing against the previous screenshot.
![image](https://user-images.githubusercontent.com/1641362/221321095-f1164000-e71b-43ca-aa8e-ccc4e7a20635.png)

fasterit:
Can you please provide a `cat /proc/cpuinfo` from inside the LXC container?
Probably /linux/LinuxProcessList.c:scanAvailableCPUsFromCPUinfo needs to be tinkered with a bit more to accommodate (more) LXC oddities.

C0rn3j:
[cpuinfo.txt](https://github.com/htop-dev/htop/files/10829821/cpuinfo.txt)

fasterit:
ok, that cpuinfo only has 16 "processors" so we need to find a way to identify which real cores are the active ones. Can you please upload a `tar -czf LXC_cpus.tar.gz /sys/devices/system/cpu/` ?

C0rn3j:
Sure thing
[LXC_cpus.tar.gz](https://github.com/htop-dev/htop/files/10830573/LXC_cpus.tar.gz)

fasterit:
`/sys/devices/system/cpu/online` seems useful:
4,7,12,15-16,18-19,24,26-27,32,39,42-43,45,47

Ofc, the individual /cpu*/online are all "1" despite LXC configured to not use them

fasterit:
[...]

Issue #993, opened 2022-05-04
Too many CPUs in LXC


odo2063:
htop 3.2.0 running in LXC shows all Cores available on Host, but statistics only for the available CPUs in the container.
![Bildschirmfoto von 2022-05-04 14-33-17@2x](https://user-images.githubusercontent.com/15078667/166685145-f2f6f7fa-2612-41c0-bcf2-76e847dd9105.png)
.

odo2063:
V 3.0.5 is ok
V 3.1.2 has same problem

fasterit:
That is because your `.htoprc` has that layout saved. Delete it or select better meters in the Setup menu (F2).

odo2063:
nope, removed all configs and tried all meter settings. It stays in 2 lines. which is simply not enough...

odo2063:
but this is maybe more related to the original problem.

I have bzip2 running which is single threaded but it shows 200% usage. (I applied your patch)
![Bildschirmfoto von 2022-05-07 08-37-24@2x](https://user-images.githubusercontent.com/15078667/167242225-a409b0d0-163e-466e-8ab5-b82394c803dd.png)

fasterit:
The initial htoprc is generated to account for the `active` CPUs and these are again not properly managed by LXC.
Try with https://github.com/htop-dev/htop/pull/995/commits/33973f7e40b6b24e219f5c9d19b205f6f9ddeed4 applied.

odo2063:
seems fine now. Thanks.

fasterit:
Cool, thanks @odo2063 for reporting and testing. I'll merge this into main now.

Issue #757, opened 2021-08-21
[sparc64] limit cpu shown on active (online) cpus only?


mator:
Hello!

After PR #656 htop shows all available CPUs , instead of active ones. How do i limit it to show it to only active CPUs?

For example, on sparc64 linux platform, under LDOM (read virtual machine), there's only particular list of CPUs available for LDOM. I.e. my LDOM has 8 active CPU (allocated for VM from hypervisor):
```
$ nproc  --all
256
$ nproc
8
```
and there's no point to show unavailable (256-8 == 248) CPUs (which would 0 load anyway).

This is how `htop` was looking before:
```

    0[                                       0.0%]     4[                                       0.0%]
    1[|||                                    3.8%]     5[|                                      0.6%]
    2[                                       0.0%]     6[||                                     1.3%]
    3[                                       0.0%]     7[|                                      0.6%]
  Mem[||||||||                         641M/15.7G]   Tasks: 29, 6 thr; 1 running
  Swp[                                    0K/769M]   Load average: 0.02 0.01 0.00
                                                     Uptime: 1 day, 09:05:00
                                                     Date & Time: 2021-08-21 22:43:56

    PID USER      PRI  NI  VIRT   RES   SHR S CPU%▽MEM%   TIME+  Command
  77472 mator      25   5 29816  4896  3640 R  3.2  0.0  0:00.25 htop
      1 root       20   0  185M 10336  7256 S  0.0  0.1  0:12.01 /sbin/init
    282 root       20   0  164M 75944 74616 S  0.0  0.5  0:07.75 /lib/systemd/systemd-journald
```
and this is how it looks now:
```
    0[                                                                                0.0%]   128[                                                                                0.0%]
    1[||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||100.0%]   129[                                                                                0.0%]
    2[                                                                                0.0%]   130[                                                                                0.0%]
    3[                                                                                0.0%]   131[                                                                                0.0%]
    4[                                                                                0.0%]   132[                                                                                0.0%]
    5[                                                                                0.0%]   133[                                                                                0.0%]
    6[                                                                                0.0%]   134[                                                                                0.0%]
    7[                                                                                0.0%]   135[                                                                                0.0%]
    8[                                                                                0.0%]   136[                                                                                0.0%]
    9[                                                                                0.0%]   137[                                                                                0.0%]
   10[                                                                                0.0%]   138[                                                                                0.0%]
[...]

PR #1197, opened 2023-02-25
Remove LXC special handling for the CPU count

fasterit:
LXC shows the real host CPU ids but can be limited in configuration as to which cores are used. Still the sysfs files are visible and the CPUs (stay) marked online. We will need to parse
/sys/devices/system/cpu/online to follow LXC's logic. Revert for now until we can come up with a better handling of the LXC hacks.

Cf. issue #1195
Essentially reverting 33973f7e and 0d53245c (#993, #995)

### DIFF
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -210,28 +210,6 @@ static void LinuxProcessList_initNetlinkSocket(LinuxProcessList* this) {
 
 #endif
 
-static unsigned int scanAvailableCPUsFromCPUinfo(LinuxProcessList* this) {
-   FILE* file = fopen(PROCCPUINFOFILE, "r");
-   if (file == NULL)
-      return this->super.existingCPUs;
-
-   unsigned int availableCPUs = 0;
-
-   while (!feof(file)) {
-      char buffer[PROC_LINE_LENGTH];
-
-      if (fgets(buffer, PROC_LINE_LENGTH, file) == NULL)
-         break;
-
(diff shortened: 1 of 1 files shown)


C0rn3j:
Applying this against 3.2.2 tarball, that does indeed go back to the old behavior.

image

Napsty:
Actually if I apply this patch to htop 3.2.2, I get the WRONG number of cpus in the htop output. 

Inside a LXC container with a cgroup limit to 2 cpus. 

htop 3.2.2 compiled from the release tar.xz:

image

htop 3.2.2 with this patch added:

image

To me this looks wrong. Why are now all cpus showing up again when the cgroup limit is in place?
The same behaviour can be seen in the current htop package in the Debian repositories. htop 3.2.2 for Bookworm contains this pull request as patch, too.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1057466

C0rn3j:
@Napsty Two reasons.

1) There is a bug introduced in 3.1.0 that does not mark trailing offline threads offline, which https://github.com/htop-dev/htop/pull/1394 fixes and it is *kinda* tracked here https://github.com/htop-dev/htop/issues/757

2) htop isn't able to show CPUs properly dynamically, this depends on https://github.com/htop-dev/htop/issues/1198  
Currently, if you assign a machine a singular thread ID 5, the removed handling caused you to see a singular thread ID 0 in htop, which was incorrect, so now you see all threads, but due to the bug you don't even see them marked as offline.

PR #1202, opened 2023-03-02
Increase filter string length to 254

### LINKED ISSUE #1201: Increase filter string length
Looks like filter string is limited to 40 characters. Can it be increased to something like 255 characters?

gabrieljones:
fixes #1201

### DIFF
--- a/IncSet.h
+++ b/IncSet.h
@@ -15,7 +15,7 @@ in the source distribution for its full text.
 #include "Vector.h"
 
 
-#define INCMODE_MAX 254
+#define INCMODE_MAX 255
 
 typedef enum {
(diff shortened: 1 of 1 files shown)


IncSet.h
```diff
@@ -15,7 +15,7 @@ in the source distribution for its full text.
 #include "Vector.h"
 
 
-#define INCMODE_MAX 40
+#define INCMODE_MAX 254
```
BenBE, line 18:
```suggestion
#define INCMODE_MAX 255
```

This has the added bonus of also causing the buffer in the `IncSet` structure to have proper alignment with no additional padding.

PR #1209, opened 2023-03-15
Fix scroll relative to followed process

### LINKED ISSUE #1180: Fixed process view
Add a feature allowing the user to keep the process/thread view pane fixed relative to the highlighted process, not the first listed process:
- Changes (processes or threads added or removed) above or below the highlighted process according to the current sort/tree configuration do not move the highlighted process/thread up or down.
- There could be a race condition depending on how this feature is implemented or the surrounding code:  The user moves highlight while a change in the process view alters placement of highlight possibly resulting in user surprise.  This may not be an issue because the feature is new and expectations will be set by the first implementation.  (For example, the user moves the highlight up one row but it jumps more than one (or doesn't move, or moves down) because some processes appeared between to previously adjacent procs according to current sort/tree view.)

Alternate but less useful idea: fix process viewport to process at the top (or bottom).  The point of fixing the process view to the highlighted process is to prevent the user from, for example, accidentally sending a signal to the wrong proc due to a shift in the view between highlighting a process and typing `k`.

gandalf3:
Pertains to #1180

### DIFF
--- a/ProcessList.c
+++ b/ProcessList.c
@@ -408,7 +408,8 @@ void ProcessList_rebuildPanel(ProcessList* this) {
       if (this->following != -1 && p->pid == this->following) {
          foundFollowed = true;
          Panel_setSelected(this->panel, idx);
-         this->panel->scrollV = currScrollV;
+         /* Keep scroll position relative to followed process */
+         this->panel->scrollV = idx - (currPos-currScrollV);
       }
       idx++;
    }


cgzones (approved):
LGTM code-wise.
One comment:
The commit descriptions contains `Fixes #1180`, which would close #1180. Does it really address all issues mentioned?

gandalf3:
good point thank you i've taken it out. i think technically it might address the issues from the original comment? but it could be nice if follow stayed following when changing nice value or sending signals or scrolling to see child processes etc, which this Pr does not address.

BenBE (approved):
LGTM.

PR #1213, opened 2023-03-29
Do not pollute main thread CPU and IO statistics with secondary threads statistics

tkhai:
This patch is for "Hide userland process threads mode is off" case.

Below is a small example of two threads program, whose main thread statistics is incorrect in htop. **Despite main thread is permanently sleeping, its CPU load is 100% (must be 0%)**. CPU load of secondary thread is correct (100%).

```
void *thread_func(void *data)
{
    while(1) {
    }
    return NULL;
}
int main()
{
    pthread_t thread;
    pthread_create(&thread, NULL, thread_func, NULL);
    pthread_join(thread, NULL);
}
```

The reason is in there is a difference in behavior of some files in **/proc/pid** and **/proc/pid/task/pid** directories. For example, **/proc/pid/stat** shows **agregated user and sys times for all threads of a process**. For the details please see _do_task_stat_() realization and this function behavior difference in dependence of last argument:

https://elixir.bootlin.com/linux/v6.2.8/source/fs/proc/array.c#L652

So, the problem occurs because of user and sys times of main thread are polluted by secondary thread statistics. This patch fixes the problem by reading correct stat from **/proc/pid/task/pid/stat** for main thread.

Looking at other files from **/proc/pid** directory I found **/proc/pid/io** file with the same problem:

https://elixir.bootlin.com/linux/v6.2.8/source/fs/proc/base.c#L3029

This problem is also fixed in this patch by reading correct data from **/proc/pid/task/pid/io** file for main thread.

**/proc/pid** directory files are declared in tgid_base_stuff, while **/proc/pid/task/tid** directory files are declared in tid_base_stuff in kernel's **fs/proc/base.c** file:

https://elixir.bootlin.com/linux/v6.2.8/source/fs/proc/base.c#L3238

I checked the difference between rest of declarations and it looks like there is no more problems of such type affecting htop.

### DIFF
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -612,31 +612,31 @@ static bool LinuxProcessList_updateUser(ProcessList* processList, Process* proce
    return true;
 }
 
-static void LinuxProcessList_readIoFile(LinuxProcess* process, openat_arg_t procFd, bool scanMainThread, unsigned long long realtimeMs) {
-   Process *proc = &process->super;
+static void LinuxProcessList_readIoFile(LinuxProcess* lp, openat_arg_t procFd, bool scanMainThread, unsigned long long realtimeMs) {
+   Process *process = &lp->super;
    char path[20] = "io";
    char buffer[1024];
    if (scanMainThread) {
-      xSnprintf(path, sizeof(path), "task/%"PRIi32"/io", (int32_t)proc->pid);
+      xSnprintf(path, sizeof(path), "task/%"PRIi32"/io", (int32_t)process->pid);
    }
    ssize_t r = xReadfileat(procFd, path, buffer, sizeof(buffer));
    if (r < 0) {
-      process->io_rate_read_bps = NAN;
-      process->io_rate_write_bps = NAN;
-      process->io_rchar = ULLONG_MAX;
-      process->io_wchar = ULLONG_MAX;
-      process->io_syscr = ULLONG_MAX;
-      process->io_syscw = ULLONG_MAX;
-      process->io_read_bytes = ULLONG_MAX;
-      process->io_write_bytes = ULLONG_MAX;
-      process->io_cancelled_write_bytes = ULLONG_MAX;
-      process->io_last_scan_time_ms = realtimeMs;
+      lp->io_rate_read_bps = NAN;
+      lp->io_rate_write_bps = NAN;
+      lp->io_rchar = ULLONG_MAX;
+      lp->io_wchar = ULLONG_MAX;
+      lp->io_syscr = ULLONG_MAX;
+      lp->io_syscw = ULLONG_MAX;
+      lp->io_read_bytes = ULLONG_MAX;
+      lp->io_write_bytes = ULLONG_MAX;
+      lp->io_cancelled_write_bytes = ULLONG_MAX;
+      lp->io_last_scan_time_ms = realtimeMs;
       return;
    }
 
-   unsigned long long last_read = process->io_read_bytes;
-   unsigned long long last_write = process->io_write_bytes;
-   unsigned long long time_delta = realtimeMs > process->io_last_scan_time_ms ? realtimeMs - process->io_last_scan_time_ms : 0;
+   unsigned long long last_read = lp->io_read_bytes;
+   unsigned long long last_write = lp->io_write_bytes;
+   unsigned long long time_delta = realtimeMs > lp->io_last_scan_time_ms ? realtimeMs - lp->io_last_scan_time_ms : 0;
 
    char* buf = buffer;
    const char* line;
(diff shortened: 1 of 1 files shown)


linux/LinuxProcessList.c
```diff
@@ -1531,8 +1540,9 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
          continue;
       }
 
+      bool isMainThread = !hideUserlandThreads && !Process_isKernelThread(proc) && !parent;
```
cgzones, line 1543: Maybe name the variable `scanMainThread`, since it might not be the main thread depending on the active setting?

tkhai: Seems good for me

tkhai:
> Some minor nitpicking … If I'm counting correctly, the allocated buffers for the path are one byte short for negative PIDs (which technically are allowed as `pid_t` on Linux is signed).

Good point, thanks. I've increased buffers 1 byte more.
 
> Suggestion: Maybe use `PRIi32` for the format specifier …

Ok

tkhai:
After pushing to my own repository the pull request here updated automatically. It seems it's all I had to do to provide new version from my local repo to htop-dev repository. If some more actions are required, please, feel free to say.
It's my first pull request on github, so possible some magic is hidden for me. Thanks! :)

linux/LinuxProcessList.c
```diff
@@ -607,9 +612,14 @@ static bool LinuxProcessList_updateUser(ProcessList* processList, Process* proce
    return true;
 }
 
-static void LinuxProcessList_readIoFile(LinuxProcess* process, openat_arg_t procFd, unsigned long long realtimeMs) {
+static void LinuxProcessList_readIoFile(Process* proc, openat_arg_t procFd, bool scanMainThread, unsigned long long realtimeMs) {
```
BenBE, line 615: I think having both `LinuxProcessList_readStatFile` and `LinuxProcessList_readIoFile` take `LinuxProcess*` as the argument and accessing `process->super` to get the embedded `struct Process_` is a somewhat cleaner solution in my eyes. Otherwise your code assumes that you pass the right structure even if that's not a given; using a `LinuxProcess*` this is the responsibility of the caller. Also avoids the local variable (if you like).

BenBE (changes requested):

linux/LinuxProcessList.c
```diff
@@ -382,11 +383,15 @@ static inline ProcessState LinuxProcessList_getProcessState(char state) {
...
 
-static bool LinuxProcessList_readStatFile(Process* process, openat_arg_t procFd, char* command, size_t commLen) {
+static bool LinuxProcessList_readStatFile(Process* process, openat_arg_t procFd, bool scanMainThread, char* command, size_t commLen) {
    LinuxProcess* lp = (LinuxProcess*) process;
 
    char buf[MAX_READ + 1];
-   ssize_t r = xReadfileat(procFd, "stat", buf, sizeof(buf));
+   char path[22] = "stat";
+   if (scanMainThread) {
+      snprintf(path, sizeof(path), "task/%"PRIi32"/stat", (int32_t)process->pid);
```
BenBE, line 392: Use `xSnprintf`, otherwise the return value is unchecked …

linux/LinuxProcessList.c
```diff
@@ -607,9 +612,14 @@ static bool LinuxProcessList_updateUser(ProcessList* processList, Process* proce
...
 }
 
-static void LinuxProcessList_readIoFile(LinuxProcess* process, openat_arg_t procFd, unsigned long long realtimeMs) {
+static void LinuxProcessList_readIoFile(Process* proc, openat_arg_t procFd, bool scanMainThread, unsigned long long realtimeMs) {
+   LinuxProcess* process = (LinuxProcess*) proc;
+   char path[20] = "io";
    char buffer[1024];
-   ssize_t r = xReadfileat(procFd, "io", buffer, sizeof(buffer));
+   if (scanMainThread) {
+      snprintf(path, sizeof(path), "task/%"PRIi32"/io", (int32_t)proc->pid);
```
BenBE, line 620: Same here: use `xSnprintf`

BenBE (changes requested):

linux/LinuxProcessList.c
tkhai: Fixed in new commit.

linux/LinuxProcessList.c
tkhai: Done

linux/LinuxProcessList.c
tkhai: Changed

linux/LinuxProcessList.c
```diff
@@ -26,6 +26,7 @@ in the source distribution for its full text.
 #include <unistd.h>
 #include <sys/stat.h>
 #include <sys/types.h>
+#include <inttypes.h>
```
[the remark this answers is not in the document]
tkhai: Ok

tkhai:
It looks like I resolve all @BenBE comments by clicking the button after pushing a new version of pull request, but "Changes requested" header is still highlighted. Don't know what to do here, possible it will disappear after second review...

linux/LinuxProcessList.c
```diff
@@ -607,9 +612,14 @@ static bool LinuxProcessList_updateUser(ProcessList* processList, Process* proce
    return true;
 }
 
-static void LinuxProcessList_readIoFile(LinuxProcess* process, openat_arg_t procFd, unsigned long long realtimeMs) {
+static void LinuxProcessList_readIoFile(LinuxProcess* process, openat_arg_t procFd, bool scanMainThread, unsigned long long realtimeMs) {
+   Process *proc = &process->super;
```
BenBE, line 616: Naming convention: `LinuxProcess *lp` vs. `Process *process`.

BenBE:
> Don't know what to do here, possible it will disappear after second review...

Yes, that will disappear, once I did another review which fully approves the changes. Looks good so far (one further minor nitpick, but we're getting there. :))

linux/LinuxProcessList.c
tkhai: You want this?

```
diff --git a/linux/LinuxProcessList.c b/linux/LinuxProcessList.c
index b94c24a1..351c545f 100644
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -612,8 +612,8 @@ static bool LinuxProcessList_updateUser(ProcessList* processList, Process* proce
    return true;
 }
 
-static void LinuxProcessList_readIoFile(LinuxProcess* process, openat_arg_t procFd, bool scanMainThread, unsigned long long realtimeMs) {
-   Process *proc = &process->super;
+static void LinuxProcessList_readIoFile(LinuxProcess* lp, openat_arg_t procFd, bool scanMainThread, unsigned long long realtimeMs) {
+   Process *proc = &lp->super;
    char path[20] = "io";
    char buffer[1024];
    if (scanMainThread) {
@@ -621,16 +621,16 @@ static void LinuxProcessList_readIoFile(LinuxProcess* process, openat_arg_t proc
    }
    ssize_t r = xReadfileat(procFd, path, buffer, sizeof(buffer));
    if (r < 0) {
-      process->io_rate_read_bps = NAN;
-      process->io_rate_write_bps = NAN;
-      process->io_rchar = ULLONG_MAX;
-      process->io_wchar = ULLONG_MAX;
-      process->io_syscr = ULLONG_MAX;
-      process->io_syscw = ULLONG_MAX;
-      process->io_read_bytes = ULLONG_MAX;
-      process->io_write_bytes = ULLONG_MAX;
-      process->io_cancelled_write_bytes = ULLONG_MAX;
-      process->io_last_scan_time_ms = realtimeMs;
+      lp->io_rate_read_bps = NAN;
+      lp->io_rate_write_bps = NAN;
+      lp->io_rchar = ULLONG_MAX;
+      lp->io_wchar = ULLONG_MAX;
+      lp->io_syscr = ULLONG_MAX;
+      lp->io_syscw = ULLONG_MAX;
+      lp->io_read_bytes = ULLONG_MAX;
+      lp->io_write_bytes = ULLONG_MAX;
+      lp->io_cancelled_write_bytes = ULLONG_MAX;
+      lp->io_last_scan_time_ms = realtimeMs;
       return;
    }
 

```

BenBE: Even though this is somewhat more extensive than the initial patch, from a naming convention consistency PoV this would be better. You may put this as a separate commit as this is mostly a stylistic change.

P.S.: You can put a programming language after the 3 backticks in markdown to use syntax highlighting. In case you wondered what that edit was about.

tkhai: >Even though this is somewhat more extensive than the initial patch, from a naming convention consistency PoV this would be better. You may put this as a separate commit as this is mostly a stylistic change.

Ok, then, possible, on separate pull request, since current LinuxProcessList_readIoFile() @process argument name is not a blame of current pull request ;)

BenBE: You can split this, but as offered, you may include it here as a second commit too.

BenBE (approved):

Issue #1216, opened 2023-04-03
Zram bar shows compressed value instead of uncompressed value


Tristan79:
Zram bar shows compressed memory usage instead of uncompressed usage.
I expected to bar to be full. But the bar will never be bigger and never reach 100%

I searched the issues. No mention. So I think this is a bug?

```
Mem[||||||||||||||||||||||||||||||||||||||||||||3.68G/3.88G]
zrm[||||||||||||                          431M(2.00G)/2.00G]
Swp[|||||||||                                   2.29G/14.5G]

$swapon

NAME       TYPE       SIZE   USED PRIO
/dev/sda5  partition 12.5G 264.3M   -2
/dev/zram0 partition    2G     2G  100

htop: 3.0.5
machine: 5.10.0-21-amd64 #1 SMP Debian 5.10.162-1 (2023-01-21)
```

Also is there an option to have the zram in the swp section colored yellow so you can make distinction between hdd swap and zram swap?

Tristan79:
Last thought. Would be nice to have both zrm compressed and uncompressed displayed with two colors.
Where you overlay the compressed bar on top of the uncompressed bar.

cgzones:
@Tristan79 please give #1217 a try

PR #1217, opened 2023-04-03
ZramMeter: update bar mode

cgzones:
Show both compressed and uncompressed bars.

Closes: #1216


BenBE (approved):
LGTM.

CRT.c
```diff
@@ -653,7 +657,8 @@ static int CRT_colorSchemes[LAST_COLORSCHEME][LAST_COLORELEMENT] = {
       [ZFS_OTHER] = A_BOLD | ColorPair(Magenta, Blue),
       [ZFS_COMPRESSED] = A_BOLD | ColorPair(White, Blue),
       [ZFS_RATIO] = A_BOLD | ColorPair(Magenta, Blue),
-      [ZRAM] = A_BOLD | ColorPair(Yellow, Blue),
+      [ZRAM_COMPRESSED] = ColorPair(Blue, Blue),
```
BenBE, line 660: This color choice will not be visible …

BenBE (commented):
Mostly LGTM. Once minor color scheme issue requiring attention …

BenBE (approved):

Explorer09:
If I mention @cgzones here would he be notified of the new post? I have a question for the ZramMeter.

The ZramMeter consists of "Compressed" and "Uncompressed" values, but is it possible that a Compressed value be greater than the Uncompressed value? That is, that memory compression (or whatever it is) make the memory usage higher than the uncompressed one.

How would you handle that in the bar meter drawing? (Or do you simply assume it won't happen or set "uncompress value = compressed value" if "uncompress value > compressed value" just for safety?)

Explorer09:
I'm thinking that if compressed > uncompressed ever happens, it's more likely an OS bug. The OS should have been able whether pages of memory are worth compressing and not try hard to compress uncompressable stuff (it would be a performance hit as well as a waste of memory).

One more question: Should htop add an indication when compressed > uncompressed happens (e.g. turning the ZramMeter bar red to inform the user about the performance drop or OS bug), or is it enough to just clamp the value?

BenBE:
Just clamping should suffice IMO. Only add color if that's more or less for free in the existing code …

Explorer09:
I made pull request #1286 as a follow-up of this.

PR #1218, opened 2023-04-04
Refactor dynamic Hashtables pointers

natoscott:
Code refactoring (no-op) in several areas noticed during dynamic screens development.

### DIFF
--- a/Settings.h
+++ b/Settings.h
@@ -51,7 +51,8 @@ typedef struct Settings_ {
    int config_version;
    HeaderLayout hLayout;
    MeterColumnSetting* hColumns;
-   Hashtable* dynamicColumns;
+   Hashtable* dynamicColumns; /* runtime-discovered columns */
+   Hashtable* dynamicMeters;  /* runtime-discovered meters */
 
    ScreenSettings** screens;
    unsigned int nScreens;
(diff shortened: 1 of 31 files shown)


Settings.h
```diff
@@ -51,7 +51,8 @@ typedef struct Settings_ {
    int config_version;
    HeaderLayout hLayout;
    MeterColumnSetting* hColumns;
-   Hashtable* dynamicColumns;
+   Hashtable* dynamicColumns; /* runtime-discovered Columns */
+   Hashtable* dynamicMeters;  /* runtime-discovered meters */
```
BenBE, line 55: Consistent letter-casing …
```suggestion
   Hashtable* dynamicColumns; /* runtime-discovered columns */
   Hashtable* dynamicMeters;  /* runtime-discovered meters */
```

pcp-htop.c
```diff
@@ -14,13 +14,14 @@ in the source distribution for its full text.
 #include "Platform.h"
 
 
+const char* program = "pcp-htop";
```
BenBE, line 17: Maybe try to get this name from `autotools` too?

BenBE (approved):
LGTM. One minor note and one suggestion …

cgzones (approved):

Settings.h
natoscott: +1, thanks

pcp-htop.c
natoscott: Hmm, not seeing an obvious way to do that after a quick look, so I may leave this idea if that's OK?

BenBE: Sure, just was a suggestion in case it's available. Would have been a bit more consistent with `htop.c` where it's fetched from `PACKAGE_NAME` directly.

Issue #1211, opened 2023-03-20
htop 3.2.2 itself using a lot of CPU (20+%)


joshuaboniface:
*I'm not 100% sure if this is a bug or a configuration issue, but reporting nonetheless.*

**Problem:**

I recently updated to `htop` version 3.2.2 (after upgrading my system from Debian 11 Bullseye to Debian 12 Bookworm testing). After doing so, the `htop` process itself is using quite a lot of CPU power. For instance, I have two instances open right now, one for CPU and one for I/O process monitoring, and both process are using >20%:

![image](https://user-images.githubusercontent.com/4031396/226368492-e23c6ef6-1b49-4111-a92b-3db0a46f52f7.png)

If I check an older system running `htop` version 3.0.5, even with (almost exactly) the same configs, CPU usage is much lower at a much more reasonable ~2-3%:

![image](https://user-images.githubusercontent.com/4031396/226369417-0d1727e5-e4f1-4609-b2ca-a006ecb73380.png)

I have a fair number of basic tweaks (e.g. enabling CPU frequency and temperature, showing all kernel threads), but the big one is I have an interval of 1.0s instead of 1.5s. However changing this back to 1.5s still results in 10-15% CPU utilization for the `htop` process, more than 10x more than `htop` 3.0.5 with a 1.0s interval. I've tried adjusting a few random settings but none that I've seen make much of a difference.

**Expected Behaviour:**

I don't think `htop` should be using nearly this much CPU power to display the system state, so it seems like either a bug or possible something wrong with my setup. Reporting in the hopes that this is resolvable so my power-sensitive system isn't using so much of its CPU resources on simply monitoring said CPU resources.

**Additional Information:**

I've attached my `htoprc`, and can provide other details such as `strace` if requested.

[htoprc.txt](https://github.com/htop-dev/htop/files/11019285/htoprc.txt)

joshuaboniface:
I'm not all that familiar with `valgrind` @fasterit, can you let me know how to invoke it to get what you'd need? If I do a basic `valgrind --tool=callgrind --log-file=valgrind-htop.log htop` and let it run for a bit then ctrl+C it, the log doesn't show much of anything.

fasterit:
that's all good, @joshuaboniface . `callgrind_annotate` will get you the info where the CPU burns cycles through processing the log file

joshuaboniface:
This invocation:

```
valgrind --log-file=valgrind-htop.log callgrind_annotate $( which htop )
```

[...]

PR #1222, opened 2023-04-12
Linux: `/proc/<pid>/status` refactor no2

cgzones:
#1211 showed reading `/proc/<pid>/status` might have a significant
performance impact.  Do not read by default only if actually needed,
since the permitted capabilities are no longer gather from it.

Improves: https://github.com/htop-dev/htop/commit/8ea144df7494bad0c46d3bf4f16c9a6556c500d2 ("Linux: Refactor /proc/<pid>/status parsing")
Improves: https://github.com/htop-dev/htop/issues/1211

### DIFF
--- a/Process.h
+++ b/Process.h
@@ -24,6 +24,13 @@ in the source distribution for its full text.
 
 #define DEFAULT_HIGHLIGHT_SECS 5
 
+typedef enum Tristate_ {
+   TRI_INITIAL = 0,
+   TRI_OFF = -1,
+   TRI_ON = 1,
+} Tristate;
+
+
 /* Core process states (shared by platforms)
  * NOTE: The enum has an ordering that is important!
  * See processStateChar in process.c for ProcessSate -> letter mapping */
@@ -106,7 +113,7 @@ typedef struct Process_ {
     *   - from file capabilities
     *   - inherited from the ambient set
     */
-   bool elevated_priv;
+   Tristate elevated_priv;
 
    /* Process runtime (in hundredth of a second) */
    unsigned long long int time;
--- a/Process.h
+++ b/Process.h
@@ -94,7 +94,7 @@ typedef struct Process_ {
    bool isUserlandThread;
 
    /* This process is running inside a container */
-   bool isRunningInContainer;
+   Tristate isRunningInContainer;
 
    /* Controlling terminal identifier of the process */
    unsigned long int tty_nr;
--- a/linux/LinuxProcess.h
+++ b/linux/LinuxProcess.h
@@ -30,6 +30,7 @@ in the source distribution for its full text.
 #define PROCESS_FLAG_LINUX_DELAYACCT 0x00040000
 #define PROCESS_FLAG_LINUX_AUTOGROUP 0x00080000
 #define PROCESS_FLAG_LINUX_GPU       0x00100000
+#define PROCESS_FLAG_LINUX_CONTAINER 0x00200000
 
 typedef struct LinuxProcess_ {
    Process super;
--- a/linux/LinuxProcessTable.c
+++ b/linux/LinuxProcessTable.c
@@ -1590,7 +1590,7 @@ static bool LinuxProcessTable_recurseProcTree(LinuxProcessTable* this, openat_ar
       }
 
       if (ss->flags & PROCESS_FLAG_LINUX_CTXT
-         || (hideRunningInContainer && proc->isRunningInContainer == TRI_INITIAL)
+         || ((hideRunningInContainer || ss->flags & PROCESS_FLAG_LINUX_CONTAINER) && proc->isRunningInContainer == TRI_INITIAL)
 #ifdef HAVE_VSERVER
          || ss->flags & PROCESS_FLAG_LINUX_VSERVER
 #endif
--- a/linux/LinuxProcessTable.c
+++ b/linux/LinuxProcessTable.c
@@ -20,7 +20,9 @@ in the source distribution for its full text.
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <syscall.h>
 #include <unistd.h>
+#include <linux/capability.h> // raw syscall, no libcap  // IWYU pragma: keep // IWYU pragma: no_include <sys/capability.h>
 #include <sys/stat.h>
 
 #include "Compat.h"
@@ -390,14 +392,6 @@ static bool LinuxProcessTable_readStatusFile(Process* process, openat_arg_t proc
 
          if (pid_ns_count > 1)
             process->isRunningInContainer = true;
-
-      } else if (String_startsWith(buffer, "CapPrm:")) {
-         char* ptr = buffer + strlen("CapPrm:");
-         while (*ptr == ' ' || *ptr == '\t')
-            ptr++;
-
-         uint64_t cap_permitted = fast_strtoull_hex(&ptr, 16);
-         process->elevated_priv = cap_permitted != 0 && process->st_uid != 0;
 
       } else if (String_startsWith(buffer, "voluntary_ctxt_switches:")) {
          unsigned long vctxt;
@@ -1478,6 +1472,19 @@ static bool LinuxProcessTable_recurseProcTree(LinuxProcessTable* this, openat_ar
             } else if (!LinuxProcessTable_readCmdlineFile(proc, procFd)) {
                Process_updateCmdline(proc, statCommand, 0, strlen(statCommand));
             }
+         }
+      }
+
+      /* Gather permitted capabilities (thread-specific data) for non-root process. */
+      if (proc->st_uid != 0 && proc->elevated_priv != TRI_OFF) {
+         struct __user_cap_header_struct header = { .version = _LINUX_CAPABILITY_VERSION_3, .pid = Process_getPid(proc) };
+         struct __user_cap_data_struct data;
+
+         long res = syscall(SYS_capget, &header, &data);
+         if (res == 0) {
+            proc->elevated_priv = (data.permitted != 0) ? TRI_ON : TRI_OFF;
+         } else {
+            proc->elevated_priv = TRI_OFF;
          }
       }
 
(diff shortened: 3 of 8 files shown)


linux/LinuxProcessList.c
```diff
@@ -1616,8 +1617,18 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
       if (! LinuxProcessList_updateUser(pl, proc, procFd))
          goto errorReadingProcess;
 
-      if (!LinuxProcessList_readStatusFile(proc, procFd))
-         goto errorReadingProcess;
+      if (ss->flags & PROCESS_FLAG_LINUX_CGROUP || hideRunningInContainer)
+         LinuxProcessList_readCGroupFile(lp, procFd);
```
tanriol, line 1621: Does this mean that in the `hideRunningInContainer` case we trade reading `status` for reading cgroup? Is that supposed to be better?
Also, are the paths being checked systemd-specific?

linux/LinuxProcessList.c
```diff
@@ -1616,8 +1617,18 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
...
+
+      if (ss->flags & PROCESS_FLAG_LINUX_CTXT
+          || (hideRunningInContainer && lp->cgroup && isContainerOrVMSlice(lp->cgroup))
+#ifdef HAVE_VSERVER
+          || ss->flags & PROCESS_FLAG_LINUX_VSERVER
```
BenBE, line 1626: Indentation
```suggestion
         || (hideRunningInContainer && lp->cgroup && isContainerOrVMSlice(lp->cgroup))
#ifdef HAVE_VSERVER
         || ss->flags & PROCESS_FLAG_LINUX_VSERVER
```

linux/LinuxProcessList.c
```diff
@@ -1436,6 +1430,13 @@ static bool isOlderThan(const ProcessList* pl, const Process* proc, unsigned int
    return realtime - proc->starttime_ctime > seconds;
 }
 
+static bool isContainerOrVMSlice(const char* cgroup) {
+   if (String_startsWith(cgroup, "/user") || String_startsWith(cgroup, "/system"))
```
BenBE, line 1434: I'm not sure this works as expected. This won't catch containers on Proxmox, but might accidently match non-containers otherwise.

tanriol:
Can we just detect container status once per process instead of per thread? According to [`pid_namespaces(7)`](https://www.man7.org/linux/man-pages/man7/pid_namespaces.7.html), threads are required to be in the same PID
namespace, so doing that once per thread seems to be extra work with no gain.

linux/LinuxProcessList.c
```diff
@@ -1646,6 +1640,14 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
          }
       }
 
+      if (proc->st_uid != 0) {
+         struct __user_cap_header_struct header = { .version = _LINUX_CAPABILITY_VERSION_3, .pid = proc->pid };
+         struct __user_cap_data_struct data;
+
+         int res = syscall(SYS_capget, &header, &data);
+         proc->elevated_priv = (res == 0) ? (data.permitted != 0) : 0;
```
Explorer09, line 1665:
```suggestion
         proc->elevated_priv = res == 0 && data.permitted != 0;
```

BenBE (approved):
LGTM.

Process.h
```diff
@@ -155,8 +155,13 @@ typedef struct Process_ {
     * Linux:
     *   - from file capabilities
     *   - inherited from the ambient set
+    *
+    * Values:
+    *    0:  inital value
```
Explorer09, line 160: There's a typo here. And it's better to state the meaning of the value in addition to how we can get the value.
```suggestion
    *    0:  initial value / undetermined
```

cgzones: thanks, adjusted

linux/LinuxProcessList.c
```diff
@@ -1535,11 +1636,12 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
          continue;
       }
 
-      bool scanMainThread = !hideUserlandThreads && !Process_isKernelThread(proc) && !parent;
+      const bool scanMainThread = !hideUserlandThreads && !Process_isKernelThread(proc) && !mainTask;
```
Explorer09: What's the purpose of the `const` qualifier here?

Explorer09: @BenBE There is no benefit to tag a local variable as a `const`, even for the purpose of compiler optimization. And while I agree the tag prevents the programmer from accidentally reassign the value later in the code, the tag is no longer needed when the code is submitted for version control. You know, any lines added for reassigning a value of a variable would have been intended by the programmer, or otherwise the patch is faulty already and should not be merged to `main`.

BenBE: I kinda disagree regarding the "does not belong in the patch": It's not just about the state of the code when it was written, but also going forward. And needing to consciously remove the const when you want to write to this variable in some later code change is better, as it communicates intent.

cgzones: It was added as a mental shortcut: when looking over the code (maybe after some time has passed) it directly states the variable is not going to change, and one can skip checking all possible branches in between the assignment and any of the usages for potential modifications.
So it's more of a maintainability thing than a (compiler) optimization.

BenBE:
@cgzones What's missing from this branch to be ready?

linux/LinuxProcess.h
```diff
@@ -18,17 +18,18 @@ in the source distribution for its full text.
...
+#define PROCESS_FLAG_LINUX_VSERVER     0x00000400
+#define PROCESS_FLAG_LINUX_CGROUP      0x00000800
+#define PROCESS_FLAG_LINUX_OOM         0x00001000
+#define PROCESS_FLAG_LINUX_SMAPS       0x00002000
+#define PROCESS_FLAG_LINUX_CTXT        0x00004000
+#define PROCESS_FLAG_LINUX_SECATTR     0x00008000
+#define PROCESS_FLAG_LINUX_LRS_FIX     0x00010000
+#define PROCESS_FLAG_LINUX_DELAYACCT   0x00040000
+#define PROCESS_FLAG_LINUX_AUTOGROUP   0x00080000
+#define PROCESS_FLAG_LINUX_ISCONTAINER 0x00100000
```
BenBE, line 32: I'd prefer if we could avoid this alignment change by keeping the new name short enough. `PROCESS_FLAG_LINUX_CONTAINER` would fit the bill IMHO.

BenBE (commented):
The various casts from `Process*` to `LinuxProcess*` for `maintask` looks kinda ugly when we know `maintask` is a `LinuxProcess*` and thus could already be provided as such; would match with the `process` argument we usually provide alongside too.

linux/LinuxProcessTable.c
```diff
@@ -20,7 +20,9 @@ in the source distribution for its full text.
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <syscall.h>
 #include <unistd.h>
+#include <linux/capability.h>
```
BenBE, line 25: Shouldn't this be behind a check for when capabilities are actually needed?

Also, this header is covered by an IWYU rule asking to import `sys/capability.h` instead. Or is there explicit reason to use `linux/capability.h` instead here?

cgzones: Fixed the include path.

There is currently no option to disable username coloring for elevated non-root processes, see https://github.com/htop-dev/htop/issues/1379.

cgzones: Reverted the path change, since *<sys/capability.h>* is provided by libcap-dev, which we don't require, and *<linux/capability.h>* is by linux-libc-dev.

BenBE: AFAICS libcap-dev is checked for in `configure.ac` if `--enable-capabilities` is given. And that check looks for `libcap`(-dev) …

linux/LinuxProcessTable.c
```diff
@@ -1639,10 +1645,38 @@ static bool LinuxProcessTable_recurseProcTree(LinuxProcessTable* this, openat_ar
          }
       }
 
-      /*
-       * Section gathering non-critical information that is independent from
-       * each other.
-       */
+      /* Reset, causes processes might enter namespaces at runtime. */
```
tanriol, line 1648: According to [pid_namespaces(7)](https://www.man7.org/linux/man-pages/man7/pid_namespaces.7.html):

> To put things another way: a process's PID namespace membership
       is determined when the process is created and cannot be changed
       thereafter.

<details>

> Calls to [setns(2)](https://www.man7.org/linux/man-pages/man2/setns.2.html) that specify a PID namespace file descriptor
       and calls to [unshare(2)](https://www.man7.org/linux/man-pages/man2/unshare.2.html) with the CLONE_NEWPID flag cause children
       subsequently created by the caller to be placed in a different
       PID namespace from the caller.  (Since Linux 4.12, that PID
       namespace is shown via the /proc/pid/ns/pid_for_children file, as
       described in [namespaces(7)](https://www.man7.org/linux/man-pages/man7/namespaces.7.html).)  
> These calls do not, however, change
       the PID namespace of the calling process, because doing so would
       change the caller's idea of its own PID (as reported by
       getpid()), which would break many applications and libraries.


</details>

cgzones: Thanks, fixed.

kartoffelheinz:
@cgzones Thank you very much for these Patches. We would love to try these in production. This PR only applies to current upstream and not latest 3.3.0 release. Since we are unfortunately not as firm with C we can only do so much as to rebuild the debian package for 3.3.0 with a patch applied. This fails trying this PR for obvious reasons since it has been based on latest upstream master, which apparently includes some significant changes compared to 3.3.0.

So may I ask you in case this is not a lot of effort, could you provide with a patch which includes your improvements and cleanly applies to release 3.3.0? Thanks so much in advance!

cgzones:
@kartoffelheinz The code in *linux/LinuxProcessTable.c* changed quite a lot with conflicting changes, so at the moment I am unable to supply rebased patches for 3.3.0.

BenBE (approved):

Issue #1138, opened 2022-11-26
Implement Network & Disk IO support on MacOS


pbrena:
Hello the sudo htop running in my MacMini, macOS Catalina cannot be setup to display Network & Disk IO as the one running in my raspberry Linux can. How do I enable it?

BenBE:
For MacOS, the necessary functions to determine the disk and network I/O are not implement.

Thus in order to activate these meters, a working implementation for `Platform_getDiskIO` and `Platform_getNetworkIO` in `darwin/Platform.c` is required.

pbrena:
I see...
Should I close this?

Is it still possible installing any (maybe 3rd party) SW?
or... is it something that could be suggested as improvement?

BenBE:
We can just update this ticket as a feature request to implement these missing functions, although I doubt there will be any work from inside the core team as noone here really works on darwin …

pbrena:
ok, Thx!

UeiWang:
I'm interested in implementing `Platform_getDiskIO` for Darwin. This may take some time, since I'm kind of busy recently.

pbrena:
That sounds great, please keep me posted.
    On Sunday, 16 April 2023, 04:45:16 AM GMT-6, UeiWang ***@***.***> wrote:  
 
 


I'm interested in implementing Platform_getDiskIO for Darwin. This may take some time, since I'm kind of busy recently.

UeiWang:
@pbrena you can now test the Disk IO Meter for Darwin by building the latest commit of #1228 (if your local repository is already configured, run `./autogen.sh` and `./configure` before building). Please pay special attention to if "Disk IO" in the Disk IO Meter may exceed 100% even when no external drive is connected to your machine.

pbrena:
i will, thx
    On Tuesday, 18 April 2023, 08:10:14 AM GMT-6, UeiWang ***@***.***> wrote:  
 
 


@pbrena you can now test the Disk IO Meter for Darwin by building the latest commit of #1228 (if your local repository is already configured, run ./autogen.sh and ./configure before building). Please pay special attention to if "Disk IO" in the Disk IO Meter may exceed 100% even when no external drive is connected to your machine.

UeiWang:
I'll also work on the Network IO Meter for Darwin :)

pbrena:
geat, thx
    On Saturday, 22 April 2023, 11:33:10 PM GMT-6, UeiWang ***@***.***> wrote:  
 
 


[...]

PR #1228, opened 2023-04-17
Add DiskIOMeter support for Darwin

UeiWang:
Relevant issue: #1138

### DIFF
--- a/darwin/Platform.c
+++ b/darwin/Platform.c
@@ -14,10 +14,19 @@ in the source distribution for its full text.
 #include <math.h>
 #include <stdlib.h>
 #include <unistd.h>
+#include <sys/_types/_mach_port_t.h>
+
+#include <CoreFoundation/CFBase.h>
+#include <CoreFoundation/CFDictionary.h>
+#include <CoreFoundation/CFNumber.h>
 #include <CoreFoundation/CFString.h>
 #include <CoreFoundation/CoreFoundation.h>
+
+#include <IOKit/IOKitLib.h>
+#include <IOKit/IOTypes.h>
 #include <IOKit/ps/IOPowerSources.h>
 #include <IOKit/ps/IOPSKeys.h>
+#include <IOKit/storage/IOBlockStorageDriver.h>
 
 #include "ClockMeter.h"
 #include "CPUMeter.h"
@@ -128,6 +137,7 @@ const MeterClass* const Platform_meterTypes[] = {
    &RightCPUs8Meter_class,
    &ZfsArcMeter_class,
    &ZfsCompressedArcMeter_class,
+   &DiskIOMeter_class,
    &FileDescriptorMeter_class,
    &BlankMeter_class,
    NULL
@@ -136,6 +146,9 @@ const MeterClass* const Platform_meterTypes[] = {
 static double Platform_nanosecondsPerMachTick = 1.0;
 
 static double Platform_nanosecondsPerSchedulerTick = -1;
+
+static bool iokit_available = false;
+static mach_port_t iokit_port; // the mach port used to initiate communication with IOKit
 
 bool Platform_init(void) {
    Platform_nanosecondsPerMachTick = Platform_calculateNanosecondsPerMachTick();
@@ -150,6 +163,17 @@ bool Platform_init(void) {
 
    const double nanos_per_sec = 1e9;
    Platform_nanosecondsPerSchedulerTick = nanos_per_sec / scheduler_ticks_per_sec;
+
+   // Since macOS 12.0, IOMasterPort is deprecated, and one should use IOMainPort instead
+   #if defined(HAVE_DECL_IOMAINPORT) && HAVE_DECL_IOMAINPORT
+   if (!IOMainPort(bootstrap_port, &iokit_port)) {
+      iokit_available = true;
+   }
+   #elif defined(HAVE_DECL_IOMASTERPORT) && HAVE_DECL_IOMASTERPORT
+   if (!IOMasterPort(bootstrap_port, &iokit_port)) {
+      iokit_available = true;
+   }
+   #endif
 
    return true;
 }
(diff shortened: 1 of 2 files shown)


darwin/Platform.c
```diff
@@ -357,9 +378,75 @@ void Platform_getFileDescriptors(double* used, double* max) {
...
+               CFNumberGetValue(number, kCFNumberSInt64Type, &value);
+               timeSpend_sum += value;
+            }
+         }
+         CFRelease(properties);
+      }
+      IOObjectRelease(drive);
+   }
+   data->totalBytesRead = read_sum;
+   data->totalBytesWritten = write_sum;
```
BenBE, line 443: No need to cramp everything together …
```suggestion
         }

         CFRelease(properties);
      }

      IOObjectRelease(drive);
   }

   data->totalBytesRead = read_sum;
   data->totalBytesWritten = write_sum;
```

BenBE (changes requested):

BenBE:
Please rebase any changes into the corresponding commits to keep the patch series clean and straight forward. TIA.

UeiWang:
> 

OK

darwin/Platform.c
```diff
@@ -151,6 +163,15 @@ bool Platform_init(void) {
...
    Platform_nanosecondsPerSchedulerTick = nanos_per_sec / scheduler_ticks_per_sec;
 
+   // Since macOS 12.0, IOMasterPort is deprecated, and one should use IOMainPort instead
+   #ifdef __MAC_12_0
+   if (!IOMainPort(bootstrap_port, &htop_main_port)) {
+   #else
+   if (!IOMasterPort(bootstrap_port, &htop_main_port)) {
+   #endif
+      IOKit_available = true;
+   }
```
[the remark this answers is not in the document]
BenBE: You can ask `configure` to check if certain functions are available and work from there.
```c
#if defined(HAS_SYMBOL1)
if (call_symbol1(42)) {
   // new and available
} else
#elif defined(HAS_SYMBOL2)
if (call_symbol2(42)) {
   // old, but available
} else
#endif
{
   // unavailable
}
```

BenBE (approved):
LGTM.

Can you have a look at the check for the version-dependent functions via `configure`? Would be nice, if we could include that.

darwin/Platform.c
```diff
@@ -357,9 +383,78 @@ void Platform_getFileDescriptors(double* used, double* max) {
...
+   if (IOServiceGetMatchingServices(iokit_port, IOServiceMatching("IOBlockStorageDriver"), &drive_list))
+      return false;
+
+   unsigned long long int read_sum = 0, write_sum = 0, timeSpend_sum = 0;
+   io_registry_entry_t drive;
+   while ((drive = IOIteratorNext(drive_list)) != 0) {
+      CFDictionaryRef properties = NULL;
+
+      /* Get the properties of this drive */
+      if (IORegistryEntryCreateCFProperties(drive, (CFMutableDictionaryRef*) &properties, kCFAllocatorDefault, 0)) {
```
cgzones, line 401:
```
CFMutableDictionaryRef properties = NULL;

      /* Get the properties of this drive */
      if (IORegistryEntryCreateCFProperties(drive, &properties, kCFAllocatorDefault, 0)) {
```
to avoid the const-cast

UeiWang:
> LGTM.
> 
> 
> 
> Can you have a look at the check for the version-dependent functions via `configure`? Would be nice, if we could include that.

Sure. Please have a look at my latest commit :)

darwin/Platform.c
UeiWang: But this produces a new `clang` warning:

`passing 'CFDictionaryRef *' (aka 'const struct __CFDictionary **') to parameter of type 'CFMutableDictionaryRef *' (aka 'struct __CFDictionary **') discards qualifiers in nested pointer types [-Wincompatible-pointer-types-discards-qualifiers]`

UeiWang:
There's something curious: when I run `htop` built from the latest commit of this pull request on my MacBook Pro 2019 running macOS Ventura 13.3.1, "Disk IO" in the Disk IO Meter may be over 100% (see the screenshot below), whereas `lldb` shows that in `Platform_getDiskIO` the body of `while ((drive = IOIteratorNext(drive_list)) != 0)` is only executed once, so that the statistics of only one disk is read.

Screenshot 2023-04-18 at 15 15 31

What do you think may be the cause?

darwin/Platform.c
BenBE: There's a difference regarding constness in those two types …

UeiWang: Since `IORegistryEntryCreateCFProperties` requires `CFMutableDictionaryRef *` and `CFDictionaryGetValue` requires `CFDictionaryRef`, to avoid  `clang` warnings, we can change this part into

`CFMutableDictionaryRef mut_properties = NULL;

      /* Get the properties of this drive */
      if (IORegistryEntryCreateCFProperties(drive, &mut_properties, kCFAllocatorDefault, 0)) {
         IOObjectRelease(drive);
         IOObjectRelease(drive_list);
         return false;
      }

      CFDictionaryRef properties = mut_properties;

      if (properties != NULL) {
         /* Get the statistics of this drive */
         CFDictionaryRef statistics = (CFDictionaryRef) CFDictionaryGetValue(properties, CFSTR(kIOBlockStorageDriverStatisticsKey));`

UeiWang: (Don't know why "`CFMutableDictionaryRef mut_properties = NULL;" wasn't in the code block)

darwin/Platform.c
```diff
@@ -357,9 +381,80 @@ void Platform_getFileDescriptors(double* used, double* max) {
...
+      /* Get the properties of this drive */
+      if (IORegistryEntryCreateCFProperties(drive, &properties_tmp, kCFAllocatorDefault, 0)) {
+         IOObjectRelease(drive);
+         IOObjectRelease(drive_list);
+         return false;
+      }
+
+      CFDictionaryRef properties = properties_tmp;
+
+      if (properties != NULL) {
```
BenBE, line 407: Why not invert this condition and flatten the block below.

```c
if (!properties) {
   continue;
}
```

You could even postpone the `CFMutableDictionaryRef` -> `CFDictionaryRef` conversion to after this check.

darwin/Platform.c
```diff
@@ -357,9 +381,83 @@ void Platform_getFileDescriptors(double* used, double* max) {
...
+         IOObjectRelease(drive);
+         continue;
+      }
+
+      CFDictionaryRef properties = properties_tmp;
+
+      /* Get the statistics of this drive */
+      CFDictionaryRef statistics = (CFDictionaryRef) CFDictionaryGetValue(properties, CFSTR(kIOBlockStorageDriverStatisticsKey));
+
+      if (statistics != NULL) {
```
BenBE, line 416: I think the same inversio can be applied here too.

BenBE (approved):

PR #1229, opened 2023-04-17
show commit in version info

ryenus:
This is to include commit hash in the output of `htop --version`.

`htop --version` output:

| Before           | After
|------------------|-------------
| `htop 3.3.0-dev` | `htop 3.3.0-dev (commit d8e779d9)`

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -6,7 +6,9 @@
 # ----------------------------------------------------------------------
 
 AC_PREREQ([2.69])
-AC_INIT([htop], [3.3.0-dev], [htop@groups.io], [], [https://htop.dev/])
+AC_INIT([htop],
+   m4_esyscmd_s(printf "%s (commit %s)" "3.3.0-dev" `git rev-parse --short HEAD`),
+   [htop@groups.io], [], [https://htop.dev/])
 
 AC_CONFIG_SRCDIR([htop.c])
 AC_CONFIG_AUX_DIR([build-aux])


BenBE:
I have two main issues with this:

1. This breaks for builds outside of git

2. The commit part is determined at `configure` time, not at build time. Thus if I don't re-`configure` after changing to a different commit, this gives stale information.

Also, there's `git describe` for this exact use case, which reduces work even further, by providing some git history information regarding the set of commits since the last tag.

ryenus:
> test -d "htop-3.3.0-dev (commit cd681ab)" || mkdir "htop-3.3.0-dev (commit cd681ab)"
> cp: cannot create regular file 'htop-3.3.0-dev (commit cd681ab)/build-aux/compile': No such file or directory

Hmm, not sure how to fix this in the automake generated scripts, any help?
About `git describe`, currently it generates `3.2.2-44-d8e779d9`, which by default is based on the last annotated tag meanwhile the current version is `3.3.0-dev`, my initial idea is only to  keep the existing versioning scheme and just mention the commit. I guess I'm open for other ideas. Nevertheless thank you @BenBE for looking into this.

fasterit:
There is another issue to think about:

The git repo the build in running in may not be htop's.

E.g. Debian packaging or PCP have their own git repos.

Explorer09:
Mentioning the git commit ID in `htop --version` might not be as useful as you think.
It is only useful for testing and reporting bugs of the `main` branch, but people often make custom branches when developing htop or packaging it for distros. With custom branches, the commit ID would be useless unless the bug reporter provides the full source code of htop.

PR #1240, opened 2023-05-09
Machine class fixes (Darwin, *BSD)

Ximalas:
A series of changes to get rid of errors and warnings for FreeBSD and NetBSD, introduced by the new Machine class.

### DIFF
--- a/freebsd/FreeBSDMachine.c
+++ b/freebsd/FreeBSDMachine.c
@@ -59,7 +59,7 @@ Machine* Machine_new(UsersTable* usersTable, uid_t userId) {
    char errbuf[_POSIX2_LINE_MAX];
    size_t len;
 
-   Machine_init(this, usersTable, userId);
+   Machine_init(super, usersTable, userId);
 
    // physical memory in system: hw.physmem
    // physical page size: hw.pagesize
@@ -146,13 +146,13 @@ Machine* Machine_new(UsersTable* usersTable, uid_t userId) {
       CRT_fatalError("kvm_openfiles() failed");
    }
 
-   return this;
+   return super;
 }
 
 void Machine_delete(Machine* super) {
    FreeBSDMachine* this = (FreeBSDMachine*) super;
 
-   ProcessList_done(this);
+   Machine_done(super);
 
    if (this->kd) {
       kvm_close(this->kd);
--- a/freebsd/FreeBSDProcessList.c
+++ b/freebsd/FreeBSDProcessList.c
@@ -29,7 +29,6 @@ in the source distribution for its full text.
 
 #include "CRT.h"
 #include "Compat.h"
-#include "FreeBSDProcess.h"
 #include "Macros.h"
 #include "Object.h"
 #include "Process.h"
@@ -38,6 +37,9 @@ in the source distribution for its full text.
 #include "Settings.h"
 #include "XUtils.h"
 
+#include "freebsd/FreeBSDMachine.h"
+#include "freebsd/FreeBSDProcess.h"
+
 
 ProcessList* ProcessList_new(Machine* host, Hashtable* pidMatchList) {
    FreeBSDProcessList* this = xCalloc(1, sizeof(FreeBSDProcessList));
@@ -45,7 +47,7 @@ ProcessList* ProcessList_new(Machine* host, Hashtable* pidMatchList) {
 
    ProcessList_init(super, Class(FreeBSDProcess), host, pidMatchList);
 
-   return this;
+   return super;
 }
 
 void ProcessList_delete(ProcessList* super) {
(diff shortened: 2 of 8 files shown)


natoscott:
@Ximalas can you check if this is a subset of #1239 please?

Ximalas:
> @Ximalas can you check if this is a subset of #1239 please?

Yes, it is.

natoscott:
OK, thanks, I'll squash & merge 'em as best I ca.

BenBE:
@Ximalas Can you cherry-pick the NetBSD changes from #1239 (Updat the commit message accordingly)? Those are related to the build failures (introduced through changes from #1234). AFAICS the patch from there should apply without (major) collisions.

Regarding your patches: In `freebsd/FreeBSDProcessList.h` please move `freebsd/FreeBSDMachine.h` and `freebsd/FreeBSDProcess.h` (path missing) to below `XUtils.h`. Amend your patch.

Also in `freebsd/FreeBSDMachine.c` for `Machine_new` you can return `super` directly (instead of casting `this`.

@natoscott : I think this PR should superseed #1239 as your PR is missing some of the header changes included here.

BenBE:
@Ximalas Can you cherry-pick the NetBSD changes from #1239 (Updat the commit message accordingly)? Those are related to the build failures (introduced through changes from #1234). AFAICS the patch from there should apply without (major) collisions.

Regarding your patches: In `freebsd/FreeBSDProcessList.h` please move `freebsd/FreeBSDMachine.h` and `freebsd/FreeBSDProcess.h` (path missing) to below `XUtils.h`. Amend your patch.

Also in `freebsd/FreeBSDMachine.c` for `Machine_new` you can return `super` directly (instead of casting `this`.

@natoscott : I think this PR should superseed #1239 as your PR is missing some of the header changes included here.

freebsd/FreeBSDProcessList.c
```diff
@@ -7,6 +7,7 @@ in the source distribution for its full text.
 
 #include "config.h" // IWYU pragma: keep
 
+#include "freebsd/FreeBSDMachine.h"
```
BenBE, line 10: Header order, should go below `XUtils.h` as this is not paired with `FreeBSDMachine.c`.

On that note: `FreeBSDProcess.h` should be `freebsd/FreeBSDProcess.h` instead (sorted below `XUtils.h` too.

freebsd/FreeBSDMachine.c
```diff
@@ -146,13 +146,13 @@ Machine* Machine_new(UsersTable* usersTable, uid_t userId) {
       CRT_fatalError("kvm_openfiles() failed");
    }
 
-   return this;
+   return (Machine *)this;
```
BenBE, line 149: Return `super` to avoid cast.

BenBE (changes requested):
Please include the NetBSD changes from #1239 as noted previously.

BenBE:
I had a closer look at the other platforms too and found some more minor things to update for consistency and correctnes.

The commits are mostly squashed from yours, except for the additional changes for Darwin, DragonFlyBSD and OpenBSD. I picked the NetBSD changes from @natoscott and reworded the commit to reflect the new commit contents.

@Ximalas: Note that I updated your branch ref accordingly. Please follow up further changes from there.

freebsd/FreeBSDProcessList.c
```diff
@@ -38,14 +37,17 @@ in the source distribution for its full text.
 #include "Settings.h"
 #include "XUtils.h"
 
+#include "freebsd/FreeBSDMachine.h"
+#include "freebsd/FreeBSDProcess.h"
+
 
 ProcessList* ProcessList_new(Machine* host, Hashtable* pidMatchList) {
```
BenBE, line 44: Following along with other platforms this should actually be called `FreeBSDProcessList_new`.

BenBE (approved):
I think only minor style issues left.

@natoscott: I think we should have a closer look at the `Machine_done` calls on the various platforms. I know, historically, we had the super class free all memory (fine so far), but in that cast we should also consider freeing the super class members before inherited members, i.e. `free(this->…);` before calling `Machine_done(super);` closed by `free(this);`. Nothing that blocks this PR, but an item for the overall refactoring list.

freebsd/FreeBSDProcessList.c
natoscott: @BenBE that makes sense, but its not correct ;)  ...

```
grep ProcessList_new */*ProcessList.c
```

natoscott:
> [...] also consider freeing the super class members before inherited members

+1  ... I'm pretty sure we're not consistent about this.  I'll add it to the cleanup queue, thanks @BenBE

natoscott (commented):
LGTM, thanks all.  :)

PR #1242, opened 2023-05-10
UptimeMeter: "days(!)," -> "days," always

nabijaczleweli:
This appears to've been inadvertently added in 33113fe0d73ea3299843b483f108ef5bcfcc944f


BenBE:
This is intentional as keeping your system up for that long is not a good practice as your are likely missing security updates for the running kernel. Thus, if htop remarks on this long runtime you really should consider to restart your system to free up resources and possibly boot a new kernel.

Issue #1191, opened 2023-02-14
Linux/PSI: Add cgroup variants for PSI meters


tamiko:
I would like to contribute new "cgroup variants" of the PSI meters for htop. For this I would need some quick feedback on the design:

Background:
---
The linux kernel has per cgroup variants ([documentation](https://docs.kernel.org/accounting/psi.html#cgroup2-interface)) for pressure stall information. Those are of particular interest for containers (or services) running in a cgroup with reduced access to system resources.

For example, let's say I start an lxc container with a very small CPU quota of 10% of one core and issue some CPU intensive task.
I then observe the following within the container:
```
# tail -n +1 /proc/pressure/cpu /sys/fs/cgroup/cpu.pressure
==> /proc/pressure/cpu <==
some avg10=10.78 avg60=2.27 avg300=1.08 total=435591433
full avg10=0.00 avg60=0.00 avg300=0.00 total=0

==> /sys/fs/cgroup/cpu.pressure <==
some avg10=14.26 avg60=3.23 avg300=1.38 total=16380126
full avg10=14.26 avg60=3.23 avg300=1.38 total=16372809
```
Here, arguably I do observe that something is amiss in the (system wide) `/proc/pressure/cpu` file. But much more relevant for the container is the information provided in `/sys/fs/cgroup/cpu.pressure`.

It would thus be nice to have a variant of the PSI meters in htop that can display PSI information from `/sys/fs/cgroup/(cpu|io|memory|irq).pressure` instead of the one found at `/proc/pressure/(cpu|io|memory|irq)`.

Questions:
---
Two solutions come to mind:

 - A natural candidate (for containers at least) would be to query `/sys/fs/cgroup/(cpu|io|memory|irq).pressure`. This works as long as an lxc guest creates a new cgroup namespace instance and mounts `/sys/fs/cgroup` accordingly.
 - I think a more flexible an robust approach instead would <b>to query for the current cgroup htop is located in</b> and then use this to show all relevant information:
    ```
    # cat /proc/self/cgroup
    0::/user.slice/user-0.slice/user@0.service/app.slice/tmux.service
    # tail -n +1 /sys/fs/cgroup/user.slice/user-0.slice/user@0.service/app.slice/tmux.service/(cpu|io|memory|irq).pressure
    ==> /sys/fs/cgroup/user.slice/user-0.slice/user@0.service/app.slice/tmux.service/cpu.pressure <==
[...]

PR #1243, opened 2023-05-11
Introduce Row and Table classes for tabular screens beyond top-processes

natoscott:
This commit refactors the Process and ProcessList structures such they each have a new parent - Row and Table, respectively.  These new classes handle screen updates relating to anything that could be represented in tabular format, e.g. cgroups, filesystems, etc, without us having to reimplement the display logic repeatedly for each new entity.

### DIFF
--- a/Action.c
+++ b/Action.c
@@ -165,6 +165,17 @@ Htop_Reaction Action_setSortKey(Settings* settings, ProcessField sortKey) {
 
 // ----------------------------------------
 
+static bool Action_writeableProcess(State* st) {
+   const Settings* settings = st->host->settings;
+   bool readonly = Settings_isReadonly() || settings->ss->dynamic;
+   return !readonly;
+}
+
+static bool Action_readableProcess(State* st) {
+   const Settings* settings = st->host->settings;
+   return !settings->ss->dynamic;
+}
+
 static Htop_Reaction actionSetSortColumn(State* st) {
    Htop_Reaction reaction = HTOP_OK;
    Panel* sortPanel = Panel_new(0, 0, 0, 0, Class(ListItem), true, FunctionBar_newEnterEsc("Sort   ", "Cancel "));
@@ -303,7 +314,7 @@ static Htop_Reaction actionIncSearch(State* st) {
 }
 
 static Htop_Reaction actionHigherPriority(State* st) {
-   if (Settings_isReadonly())
+   if (!Action_writeableProcess(st))
       return HTOP_OK;
 
    bool changed = changePriority(st->mainPanel, -1);
@@ -311,7 +322,7 @@ static Htop_Reaction actionHigherPriority(State* st) {
 }
 
 static Htop_Reaction actionLowerPriority(State* st) {
-   if (Settings_isReadonly())
+   if (!Action_writeableProcess(st))
       return HTOP_OK;
 
    bool changed = changePriority(st->mainPanel, 1);
@@ -345,13 +356,27 @@ static Htop_Reaction actionExpandCollapseOrSortColumn(State* st) {
    return st->host->settings->ss->treeView ? actionExpandOrCollapse(st) : actionSetSortColumn(st);
 }
 
+static inline void setActiveScreen(Settings* settings, State* st, unsigned int ssIdx) {
+   assert(settings->ssIndex == ssIdx);
+   Machine* host = st->host;
+
+   settings->ss = settings->screens[ssIdx];
+   if (!settings->ss->table)
+      settings->ss->table = host->processTable;
+   host->activeTable = settings->ss->table;
+
+   // set correct functionBar - readonly if requested, and/or with non-process screens
+   bool readonly = Settings_isReadonly() || (host->activeTable != host->processTable);
+   MainPanel_setFunctionBar(st->mainPanel, readonly);
+}
+
 static Htop_Reaction actionNextScreen(State* st) {
    Settings* settings = st->host->settings;
    settings->ssIndex++;
    if (settings->ssIndex == settings->nScreens) {
       settings->ssIndex = 0;
    }
-   settings->ss = settings->screens[settings->ssIndex];
+   setActiveScreen(settings, st, settings->ssIndex);
    return HTOP_UPDATE_PANELHDR | HTOP_REFRESH | HTOP_REDRAW_BAR;
 }
 
@@ -362,21 +387,22 @@ static Htop_Reaction actionPrevScreen(State* st) {
    } else {
       settings->ssIndex--;
    }
-   settings->ss = settings->screens[settings->ssIndex];
+   setActiveScreen(settings, st, settings->ssIndex);
    return HTOP_UPDATE_PANELHDR | HTOP_REFRESH | HTOP_REDRAW_BAR;
 }
 
-Htop_Reaction Action_setScreenTab(Settings* settings, int x) {
+Htop_Reaction Action_setScreenTab(State* st, int x) {
+   Settings* settings = st->host->settings;
    int s = 2;
    for (unsigned int i = 0; i < settings->nScreens; i++) {
       if (x < s) {
          return 0;
       }
-      const char* name = settings->screens[i]->name;
-      int len = strlen(name);
+      const char* tab = settings->screens[i]->heading;
+      int len = strlen(tab);
       if (x <= s + len + 1) {
          settings->ssIndex = i;
-         settings->ss = settings->screens[i];
+         setActiveScreen(settings, st, i);
          return HTOP_UPDATE_PANELHDR | HTOP_REFRESH | HTOP_REDRAW_BAR;
       }
       s += len + 3;
@@ -389,7 +415,7 @@ static Htop_Reaction actionQuit(ATTR_UNUSED State* st) {
 }
 
 static Htop_Reaction actionSetAffinity(State* st) {
-   if (Settings_isReadonly())
+   if (!Action_writeableProcess(st))
       return HTOP_OK;
 
    Machine* host = st->host;
@@ -426,7 +452,7 @@ static Htop_Reaction actionSetAffinity(State* st) {
 
 #ifdef SCHEDULER_SUPPORT
 static Htop_Reaction actionSetSchedPolicy(State* st) {
-   if (Settings_isReadonly())
+   if (!Action_writeableProcess(st))
       return HTOP_KEEP_FOLLOWING;
 
    static int preSelectedPolicy = SCHEDULINGPANEL_INITSELECTEDPOLICY;
@@ -470,7 +496,7 @@ static Htop_Reaction actionSetSchedPolicy(State* st) {
 #endif  /* SCHEDULER_SUPPORT */
 
 static Htop_Reaction actionKill(State* st) {
-   if (Settings_isReadonly())
+   if (!Action_writeableProcess(st))
       return HTOP_OK;
 
    static int preSelectedSignal = SIGNALSPANEL_INITSELECTEDSIGNAL;
@@ -521,7 +547,7 @@ static Htop_Reaction actionSetup(State* st) {
 }
 
 static Htop_Reaction actionLsof(State* st) {
-   if (Settings_isReadonly())
+   if (!Action_writeableProcess(st))
       return HTOP_OK;
 
    const Process* p = (Process*) Panel_getSelected((Panel*)st->mainPanel);
@@ -539,6 +565,9 @@ static Htop_Reaction actionLsof(State* st) {
 }
 
 static Htop_Reaction actionShowLocks(State* st) {
+   if (!Action_readableProcess(st))
+      return HTOP_OK;
+
    const Process* p = (Process*) Panel_getSelected((Panel*)st->mainPanel);
    if (!p)
       return HTOP_OK;
@@ -554,7 +583,7 @@ static Htop_Reaction actionShowLocks(State* st) {
 }
 
 static Htop_Reaction actionStrace(State* st) {
-   if (Settings_isReadonly())
+   if (!Action_writeableProcess(st))
       return HTOP_OK;
 
    const Process* p = (Process*) Panel_getSelected((Panel*)st->mainPanel);
@@ -805,6 +834,9 @@ static Htop_Reaction actionTagAllChildren(State* st) {
 }
 
 static Htop_Reaction actionShowEnvScreen(State* st) {
+   if (!Action_readableProcess(st))
+      return HTOP_OK;
+
    Process* p = (Process*) Panel_getSelected((Panel*)st->mainPanel);
    if (!p)
       return HTOP_OK;
@@ -820,6 +852,9 @@ static Htop_Reaction actionShowEnvScreen(State* st) {
 }
 
 static Htop_Reaction actionShowCommandScreen(State* st) {
+   if (!Action_readableProcess(st))
+      return HTOP_OK;
+
    Process* p = (Process*) Panel_getSelected((Panel*)st->mainPanel);
    if (!p)
       return HTOP_OK;
--- a/Action.c
+++ b/Action.c
@@ -56,22 +56,22 @@ Object* Action_pickFromVector(State* st, Panel* list, int x, bool follow) {
    Panel* panelFocus;
    int ch;
    bool unfollow = false;
-   int pid = follow ? MainPanel_selectedPid(mainPanel) : -1;
-   if (follow && host->pl->following == -1) {
-      host->pl->following = pid;
+   int row = follow ? MainPanel_selectedRow(mainPanel) : -1;
+   if (follow && host->activeTable->following == -1) {
+      host->activeTable->following = row;
       unfollow = true;
    }
    ScreenManager_run(scr, &panelFocus, &ch, NULL);
    if (unfollow) {
-      host->pl->following = -1;
+      host->activeTable->following = -1;
    }
    ScreenManager_delete(scr);
    Panel_move((Panel*)mainPanel, 0, y);
    Panel_resize((Panel*)mainPanel, COLS, LINES - y - 1);
    if (panelFocus == list && ch == 13) {
       if (follow) {
-         const Process* selected = (const Process*)Panel_getSelected((Panel*)mainPanel);
-         if (selected && selected->pid == pid)
+         const Row* selected = (const Row*)Panel_getSelected((Panel*)mainPanel);
+         if (selected && selected->id == row)
             return Panel_getSelected(list);
 
          beep();
@@ -99,7 +99,7 @@ static void Action_runSetup(State* st) {
 
 static bool changePriority(MainPanel* panel, int delta) {
    bool anyTagged;
-   bool ok = MainPanel_foreachProcess(panel, Process_changePriorityBy, (Arg) { .i = delta }, &anyTagged);
+   bool ok = MainPanel_foreachRow(panel, Process_rowChangePriorityBy, (Arg) { .i = delta }, &anyTagged);
    if (!ok)
       beep();
    return anyTagged;
@@ -121,36 +121,36 @@ bool Action_setUserOnly(const char* userName, uid_t* userId) {
    return false;
 }
 
-static void tagAllChildren(Panel* panel, Process* parent) {
+static void tagAllChildren(Panel* panel, Row* parent) {
    parent->tag = true;
-   pid_t ppid = parent->pid;
+   int parent_id = parent->id;
    for (int i = 0; i < Panel_size(panel); i++) {
-      Process* p = (Process*) Panel_get(panel, i);
-      if (!p->tag && Process_isChildOf(p, ppid)) {
-         tagAllChildren(panel, p);
+      Row* row = (Row*) Panel_get(panel, i);
+      if (!row->tag && Row_isChildOf(row, parent_id)) {
+         tagAllChildren(panel, row);
       }
    }
 }
 
 static bool expandCollapse(Panel* panel) {
-   Process* p = (Process*) Panel_getSelected(panel);
-   if (!p)
+   Row* row = (Row*) Panel_getSelected(panel);
+   if (!row)
       return false;
 
-   p->showChildren = !p->showChildren;
+   row->showChildren = !row->showChildren;
    return true;
 }
 
 static bool collapseIntoParent(Panel* panel) {
-   const Process* p = (Process*) Panel_getSelected(panel);
-   if (!p)
+   const Row* r = (Row*) Panel_getSelected(panel);
+   if (!r)
       return false;
 
-   pid_t ppid = Process_getParentPid(p);
+   int parent_id = Row_getGroupOrParent(r);
    for (int i = 0; i < Panel_size(panel); i++) {
-      Process* q = (Process*) Panel_get(panel, i);
-      if (q->pid == ppid) {
-         q->showChildren = false;
+      Row* row = (Row*) Panel_get(panel, i);
+      if (row->id == parent_id) {
+         row->showChildren = false;
          Panel_setSelected(panel, i);
          return true;
       }
@@ -159,7 +159,7 @@ static bool collapseIntoParent(Panel* panel) {
 }
 
 Htop_Reaction Action_setSortKey(Settings* settings, ProcessField sortKey) {
-   ScreenSettings_setSortKey(settings->ss, sortKey);
+   ScreenSettings_setSortKey(settings->ss, (RowField) sortKey);
    return HTOP_REFRESH | HTOP_SAVE_SETTINGS | HTOP_UPDATE_PANELHDR | HTOP_KEEP_FOLLOWING;
 }
 
@@ -171,11 +171,11 @@ static Htop_Reaction actionSetSortColumn(State* st) {
    Panel_setHeader(sortPanel, "Sort by");
    Machine* host = st->host;
    Settings* settings = host->settings;
-   const ProcessField* fields = settings->ss->fields;
+   const RowField* fields = settings->ss->fields;
    Hashtable* dynamicColumns = settings->dynamicColumns;
    for (int i = 0; fields[i]; i++) {
       char* name = NULL;
-      if (fields[i] >= LAST_PROCESSFIELD) {
+      if (fields[i] >= ROW_DYNAMIC_FIELDS) {
          DynamicColumn* column = Hashtable_get(dynamicColumns, fields[i]);
          if (!column)
             continue;
@@ -195,7 +195,7 @@ static Htop_Reaction actionSetSortColumn(State* st) {
    }
    Object_delete(sortPanel);
 
-   host->pl->needsSort = true;
+   host->activeTable->needsSort = true;
 
    return reaction | HTOP_REFRESH | HTOP_REDRAW_BAR | HTOP_UPDATE_PANELHDR;
 }
@@ -262,9 +262,9 @@ static Htop_Reaction actionToggleTreeView(State* st) {
    ss->treeView = !ss->treeView;
 
    if (!ss->allBranchesCollapsed)
-      ProcessList_expandTree(host->pl);
+      Table_expandTree(host->activeTable);
 
-   host->pl->needsSort = true;
+   host->activeTable->needsSort = true;
 
    return HTOP_REFRESH | HTOP_SAVE_SETTINGS | HTOP_KEEP_FOLLOWING | HTOP_REDRAW_BAR | HTOP_UPDATE_PANELHDR;
 }
@@ -282,9 +282,9 @@ static Htop_Reaction actionExpandOrCollapseAllBranches(State* st) {
    }
    ss->allBranchesCollapsed = !ss->allBranchesCollapsed;
    if (ss->allBranchesCollapsed)
-      ProcessList_collapseAllBranches(host->pl);
+      Table_collapseAllBranches(host->activeTable);
    else
-      ProcessList_expandTree(host->pl);
+      Table_expandTree(host->activeTable);
    return HTOP_REFRESH | HTOP_SAVE_SETTINGS;
 }
 
@@ -292,7 +292,7 @@ static Htop_Reaction actionIncFilter(State* st) {
    Machine* host = st->host;
    IncSet* inc = (st->mainPanel)->inc;
    IncSet_activate(inc, INC_FILTER, (Panel*)st->mainPanel);
-   host->pl->incFilter = IncSet_filter(inc);
+   host->activeTable->incFilter = IncSet_filter(inc);
    return HTOP_REFRESH | HTOP_KEEP_FOLLOWING;
 }
 
@@ -321,7 +321,7 @@ static Htop_Reaction actionLowerPriority(State* st) {
 static Htop_Reaction actionInvertSortOrder(State* st) {
    Machine* host = st->host;
    ScreenSettings_invertSortOrder(host->settings->ss);
-   host->pl->needsSort = true;
+   host->activeTable->needsSort = true;
    return HTOP_REFRESH | HTOP_SAVE_SETTINGS | HTOP_KEEP_FOLLOWING | HTOP_UPDATE_PANELHDR;
 }
 
@@ -397,11 +397,11 @@ static Htop_Reaction actionSetAffinity(State* st) {
       return HTOP_OK;
 
 #if (defined(HAVE_LIBHWLOC) || defined(HAVE_AFFINITY))
-   const Process* p = (const Process*) Panel_getSelected((Panel*)st->mainPanel);
-   if (!p)
+   const Row* row = (const Row*) Panel_getSelected((Panel*)st->mainPanel);
+   if (!row)
       return HTOP_OK;
 
-   Affinity* affinity1 = Affinity_get(p, host);
+   Affinity* affinity1 = Affinity_rowGet(row, host);
    if (!affinity1)
       return HTOP_OK;
 
@@ -412,7 +412,7 @@ static Htop_Reaction actionSetAffinity(State* st) {
    const void* set = Action_pickFromVector(st, affinityPanel, width, true);
    if (set) {
       Affinity* affinity2 = AffinityPanel_getAffinity(affinityPanel, host);
-      bool ok = MainPanel_foreachProcess(st->mainPanel, Affinity_set, (Arg) { .v = affinity2 }, NULL);
+      bool ok = MainPanel_foreachRow(st->mainPanel, Affinity_rowSet, (Arg) { .v = affinity2 }, NULL);
       if (!ok)
          beep();
       Affinity_delete(affinity2);
@@ -422,7 +422,6 @@ static Htop_Reaction actionSetAffinity(State* st) {
 #else
    return HTOP_OK;
 #endif
-
 }
 
 #ifdef SCHEDULER_SUPPORT
@@ -459,7 +458,7 @@ static Htop_Reaction actionSetSchedPolicy(State* st) {
 
       SchedulingArg v = { .policy = preSelectedPolicy, .priority = preSelectedPriority };
 
-      bool ok = MainPanel_foreachProcess(st->mainPanel, Scheduling_setPolicy, (Arg) { .v = &v }, NULL);
+      bool ok = MainPanel_foreachRow(st->mainPanel, Scheduling_rowSetPolicy, (Arg) { .v = &v }, NULL);
       if (!ok)
          beep();
    }
@@ -483,7 +482,7 @@ static Htop_Reaction actionKill(State* st) {
       Panel_setHeader((Panel*)st->mainPanel, "Sending...");
       Panel_draw((Panel*)st->mainPanel, false, true, true, State_hideFunctionBar(st));
       refresh();
-      MainPanel_foreachProcess(st->mainPanel, Process_sendSignal, (Arg) { .i = sgn->key }, NULL);
+      MainPanel_foreachRow(st->mainPanel, Process_rowSendSignal, (Arg) { .i = sgn->key }, NULL);
       napms(500);
    }
    Panel_delete((Object*)signalsPanel);
@@ -511,7 +510,7 @@ static Htop_Reaction actionFilterByUser(State* st) {
 }
 
 Htop_Reaction Action_follow(State* st) {
-   st->host->pl->following = MainPanel_selectedPid(st->mainPanel);
+   st->host->activeTable->following = MainPanel_selectedRow(st->mainPanel);
    Panel_setSelectionColor((Panel*)st->mainPanel, PANEL_SELECTION_FOLLOW);
    return HTOP_KEEP_FOLLOWING;
 }
@@ -529,6 +528,8 @@ static Htop_Reaction actionLsof(State* st) {
    if (!p)
       return HTOP_OK;
 
+   assert(Object_isA((const Object*) p, (const ObjectClass*) &Process_class));
+
    OpenFilesScreen* ofs = OpenFilesScreen_new(p);
    InfoScreen_run((InfoScreen*)ofs);
    OpenFilesScreen_delete((Object*)ofs);
@@ -541,6 +542,9 @@ static Htop_Reaction actionShowLocks(State* st) {
    const Process* p = (Process*) Panel_getSelected((Panel*)st->mainPanel);
    if (!p)
       return HTOP_OK;
+
+   assert(Object_isA((const Object*) p, (const ObjectClass*) &Process_class));
+
    ProcessLocksScreen* pls = ProcessLocksScreen_new(p);
    InfoScreen_run((InfoScreen*)pls);
    ProcessLocksScreen_delete((Object*)pls);
@@ -557,6 +561,8 @@ static Htop_Reaction actionStrace(State* st) {
    if (!p)
       return HTOP_OK;
 
+   assert(Object_isA((const Object*) p, (const ObjectClass*) &Process_class));
+
    TraceScreen* ts = TraceScreen_new(p);
    bool ok = TraceScreen_forkTracer(ts);
    if (ok) {
@@ -569,11 +575,11 @@ static Htop_Reaction actionStrace(State* st) {
 }
 
 static Htop_Reaction actionTag(State* st) {
-   Process* p = (Process*) Panel_getSelected((Panel*)st->mainPanel);
-   if (!p)
+   Row* r = (Row*) Panel_getSelected((Panel*)st->mainPanel);
+   if (!r)
       return HTOP_OK;
 
-   Process_toggleTag(p);
+   Row_toggleTag(r);
    Panel_onKey((Panel*)st->mainPanel, KEY_DOWN);
    return HTOP_OK;
 }
@@ -783,18 +789,18 @@ static Htop_Reaction actionHelp(State* st) {
 
 static Htop_Reaction actionUntagAll(State* st) {
    for (int i = 0; i < Panel_size((Panel*)st->mainPanel); i++) {
-      Process* p = (Process*) Panel_get((Panel*)st->mainPanel, i);
-      p->tag = false;
+      Row* row = (Row*) Panel_get((Panel*)st->mainPanel, i);
+      row->tag = false;
    }
    return HTOP_REFRESH;
 }
 
 static Htop_Reaction actionTagAllChildren(State* st) {
-   Process* p = (Process*) Panel_getSelected((Panel*)st->mainPanel);
-   if (!p)
+   Row* row = (Row*) Panel_getSelected((Panel*)st->mainPanel);
+   if (!row)
       return HTOP_OK;
 
-   tagAllChildren((Panel*)st->mainPanel, p);
+   tagAllChildren((Panel*)st->mainPanel, row);
    return HTOP_OK;
 }
 
@@ -802,6 +808,8 @@ static Htop_Reaction actionShowEnvScreen(State* st) {
    Process* p = (Process*) Panel_getSelected((Panel*)st->mainPanel);
    if (!p)
       return HTOP_OK;
+
+   assert(Object_isA((const Object*) p, (const ObjectClass*) &Process_class));
 
    EnvScreen* es = EnvScreen_new(p);
    InfoScreen_run((InfoScreen*)es);
@@ -815,6 +823,8 @@ static Htop_Reaction actionShowCommandScreen(State* st) {
    Process* p = (Process*) Panel_getSelected((Panel*)st->mainPanel);
    if (!p)
       return HTOP_OK;
+
+   assert(Object_isA((const Object*) p, (const ObjectClass*) &Process_class));
 
    CommandScreen* cmdScr = CommandScreen_new(p);
    InfoScreen_run((InfoScreen*)cmdScr);
--- a/Affinity.c
+++ b/Affinity.c
@@ -12,6 +12,7 @@ in the source distribution for its full text.
 
 #include <stdlib.h>
 
+#include "Process.h"
 #include "XUtils.h"
 
 #if defined(HAVE_LIBHWLOC)
@@ -49,12 +50,11 @@ void Affinity_add(Affinity* this, unsigned int id) {
    this->used++;
 }
 
-
 #if defined(HAVE_LIBHWLOC)
 
-Affinity* Affinity_get(const Process* proc, Machine* host) {
+static Affinity* Affinity_get(const Process* p, Machine* host) {
    hwloc_cpuset_t cpuset = hwloc_bitmap_alloc();
-   bool ok = (hwloc_get_proc_cpubind(host->topology, proc->pid, cpuset, HTOP_HWLOC_CPUBIND_FLAG) == 0);
+   bool ok = (hwloc_get_proc_cpubind(host->topology, Process_getPid(p), cpuset, HTOP_HWLOC_CPUBIND_FLAG) == 0);
    Affinity* affinity = NULL;
    if (ok) {
       affinity = Affinity_new(host);
@@ -73,22 +73,22 @@ Affinity* Affinity_get(const Process* proc, Machine* host) {
    return affinity;
 }
 
-bool Affinity_set(Process* proc, Arg arg) {
+static bool Affinity_set(Process* p, Arg arg) {
    Affinity* this = arg.v;
    hwloc_cpuset_t cpuset = hwloc_bitmap_alloc();
    for (unsigned int i = 0; i < this->used; i++) {
       hwloc_bitmap_set(cpuset, this->cpus[i]);
    }
-   bool ok = (hwloc_set_proc_cpubind(this->host->topology, proc->pid, cpuset, HTOP_HWLOC_CPUBIND_FLAG) == 0);
+   bool ok = (hwloc_set_proc_cpubind(this->host->topology, Process_getPid(p), cpuset, HTOP_HWLOC_CPUBIND_FLAG) == 0);
    hwloc_bitmap_free(cpuset);
    return ok;
 }
 
 #elif defined(HAVE_AFFINITY)
 
-Affinity* Affinity_get(const Process* proc, Machine* host) {
+static Affinity* Affinity_get(const Process* p, Machine* host) {
    cpu_set_t cpuset;
-   bool ok = (sched_getaffinity(proc->pid, sizeof(cpu_set_t), &cpuset) == 0);
+   bool ok = (sched_getaffinity(Process_getPid(p), sizeof(cpu_set_t), &cpuset) == 0);
    if (!ok)
       return NULL;
 
@@ -101,15 +101,31 @@ Affinity* Affinity_get(const Process* proc, Machine* host) {
    return affinity;
 }
 
-bool Affinity_set(Process* proc, Arg arg) {
+static bool Affinity_set(Process* p, Arg arg) {
    Affinity* this = arg.v;
    cpu_set_t cpuset;
    CPU_ZERO(&cpuset);
    for (unsigned int i = 0; i < this->used; i++) {
       CPU_SET(this->cpus[i], &cpuset);
    }
-   bool ok = (sched_setaffinity(proc->pid, sizeof(unsigned long), &cpuset) == 0);
+   bool ok = (sched_setaffinity(Process_getPid(p), sizeof(unsigned long), &cpuset) == 0);
    return ok;
 }
 
 #endif
+
+#if defined(HAVE_LIBHWLOC) || defined(HAVE_AFFINITY)
+
+bool Affinity_rowSet(Row* row, Arg arg) {
+   Process* p = (Process*) row;
+   assert(Object_isA((const Object*) p, (const ObjectClass*) &Process_class));
+   return Affinity_set(p, arg);
+}
+
+Affinity* Affinity_rowGet(const Row* row, Machine* host) {
+   const Process* p = (const Process*) row;
+   assert(Object_isA((const Object*) p, (const ObjectClass*) &Process_class));
+   return Affinity_get(p, host);
+}
+
+#endif /* HAVE_LIBHWLOC || HAVE_AFFINITY */
--- a/Affinity.h
+++ b/Affinity.h
@@ -16,7 +16,7 @@ in the source distribution for its full text.
 #include <stdbool.h>
 
 #include "Object.h"
-#include "Process.h"
+#include "Row.h"
 #endif
 
 
@@ -40,9 +40,9 @@ void Affinity_add(Affinity* this, unsigned int id);
 
 #if defined(HAVE_LIBHWLOC) || defined(HAVE_AFFINITY)
 
-Affinity* Affinity_get(const Process* proc, Machine* host);
+Affinity* Affinity_rowGet(const Row* row, Machine* host);
 
-bool Affinity_set(Process* proc, Arg arg);
+bool Affinity_rowSet(Row* row, Arg arg);
 
 #endif /* HAVE_LIBHWLOC || HAVE_AFFINITY */
 
--- a/CommandLine.c
+++ b/CommandLine.c
@@ -25,6 +25,7 @@ in the source distribution for its full text.
 #include "CRT.h"
 #include "DynamicColumn.h"
 #include "DynamicMeter.h"
+#include "DynamicScreen.h"
 #include "Hashtable.h"
 #include "Header.h"
 #include "IncSet.h"
@@ -337,15 +338,12 @@ int CommandLine_run(int argc, char** argv) {
    UsersTable* ut = UsersTable_new();
    Hashtable* dm = DynamicMeters_new();
    Hashtable* dc = DynamicColumns_new();
-   if (!dc)
-      dc = Hashtable_new(0, true);
+   Hashtable* ds = DynamicScreens_new();
 
    Machine* host = Machine_new(ut, flags.userId);
    ProcessList* pl = ProcessList_new(host, flags.pidMatchList);
-   Settings* settings = Settings_new(host->activeCPUs, dm, dc);
-
-   host->settings = settings;
-   Machine_addTable(host, &pl->super, true);
+   Settings* settings = Settings_new(host->activeCPUs, dm, dc, ds);
+   Machine_populateTablesFromSettings(host, settings, &pl->super);
 
    Header* header = Header_new(host, 2);
    Header_populateFromSettings(header);
@@ -377,7 +375,7 @@ int CommandLine_run(int argc, char** argv) {
    CRT_init(settings, flags.allowUnicode, flags.iterationsRemaining != -1);
 
    MainPanel* panel = MainPanel_new();
-   Table_setPanel(&pl->super, (Panel*) panel);
+   Machine_setTablesPanel(host, (Panel*) panel);
 
    MainPanel_updateLabels(panel, settings->ss->treeView, flags.commFilter);
 
@@ -435,6 +433,7 @@ int CommandLine_run(int argc, char** argv) {
    Settings_delete(settings);
    DynamicColumns_delete(dc);
    DynamicMeters_delete(dm);
+   DynamicScreens_delete(ds);
 
    return 0;
 }
--- a/CommandLine.c
+++ b/CommandLine.c
@@ -303,11 +303,11 @@ static void CommandLine_delay(Machine* host, unsigned long millisec) {
 }
 
 static void setCommFilter(State* state, char** commFilter) {
-   ProcessList* pl = state->host->pl;
+   Table* table = state->host->activeTable;
    IncSet* inc = state->mainPanel->inc;
 
    IncSet_setFilter(inc, *commFilter);
-   pl->incFilter = IncSet_filter(inc);
+   table->incFilter = IncSet_filter(inc);
 
    free(*commFilter);
    *commFilter = NULL;
@@ -334,8 +334,6 @@ int CommandLine_run(int argc, char** argv) {
    if (!Platform_init())
       return 1;
 
-   Process_setupColumnWidths();
-
    UsersTable* ut = UsersTable_new();
    Hashtable* dm = DynamicMeters_new();
    Hashtable* dc = DynamicColumns_new();
@@ -347,7 +345,7 @@ int CommandLine_run(int argc, char** argv) {
    Settings* settings = Settings_new(host->activeCPUs, dm, dc);
 
    host->settings = settings;
-   Machine_addList(host, pl);
+   Machine_addTable(host, &pl->super, true);
 
    Header* header = Header_new(host, 2);
    Header_populateFromSettings(header);
@@ -379,7 +377,7 @@ int CommandLine_run(int argc, char** argv) {
    CRT_init(settings, flags.allowUnicode, flags.iterationsRemaining != -1);
 
    MainPanel* panel = MainPanel_new();
-   ProcessList_setPanel(pl, (Panel*) panel);
+   Table_setPanel(&pl->super, (Panel*) panel);
 
    MainPanel_updateLabels(panel, settings->ss->treeView, flags.commFilter);
 
@@ -400,13 +398,13 @@ int CommandLine_run(int argc, char** argv) {
    ScreenManager_add(scr, (Panel*) panel, -1);
 
    Machine_scan(host);
-   ProcessList_scan(pl);
+   Machine_scanTables(host);
    CommandLine_delay(host, 75);
    Machine_scan(host);
-   ProcessList_scan(pl);
+   Machine_scanTables(host);
 
    if (settings->ss->allBranchesCollapsed)
-      ProcessList_collapseAllBranches(pl);
+      Table_collapseAllBranches(&pl->super);
 
    ScreenManager_run(scr, NULL, NULL, NULL);
 
@@ -421,7 +419,6 @@ int CommandLine_run(int argc, char** argv) {
    }
 
    Header_delete(header);
-   ProcessList_delete(pl);
    Machine_delete(host);
 
    ScreenManager_delete(scr);
--- a/DynamicScreen.c
+++ b/DynamicScreen.c
@@ -1,0 +1,65 @@
+/*
+htop - DynamicScreen.c
+(C) 2022 Sohaib Mohammed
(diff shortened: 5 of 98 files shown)


Explorer09:
I didn't read the entire commit, but I have one comment here:
I don't like the class naming "Table" and "Row" here. Considering we have `Hashtable` class already, the name Table could draw confusion for "what does this Table class do exactly and how is it different from Hashtable".
I can't suggest a better name because I'm not sure of the goals of these classes for now. `UITable`? `UIDataTable`? Would these names be better?

Machine.c
```diff
@@ -54,7 +54,23 @@ void Machine_done(Machine* this) {
 #endif
 }
 
-void Machine_addList(Machine* this, struct ProcessList_ *pl) {
-   // currently only process lists are supported
-   this->pl = pl;
+void Machine_addTable(Machine* this, Table* table, bool processes) {
+   if (processes)
```
Explorer09, line 58: The argument `processes` doesn't match the prototype declared in `Machine.h`

Process.h
```diff
@@ -106,19 +108,7 @@ typedef struct ProcessMergedCommand_ {
...
-
-   /* Process identifier */
-   pid_t pid;
-
-   /* Parent process identifier */
-   pid_t ppid;
-
-   /* Thread group identifier */
-   pid_t tgid;
+   Row super;
```
Explorer09, line 111: The `Process` class should not inherit the `Row` class. Make it a composition relation instead.
You know in OOP, the inheritance is better used for "is a kind of" relation. A inherits B when A is a kind of B.

Row.h
```diff
@@ -0,0 +1,141 @@
...
+    */
+   int32_t indent;
+   unsigned int tree_depth;
+} Row;
+
+typedef unsigned int RowField;
+
+typedef Row* (*Row_New)(const struct Machine_*);
+typedef bool (*Row_Filtered)(const Row*, const struct Table_*);
+typedef bool (*Row_Obscured)(const Row*, const struct Table_*);
```
Explorer09, line 66: Can we add an "is" word on these functions that return boolean?
* `Row_IsObscured`
* `Row_IsFiltered`

By the way, what are the purposes of these two new virtual functions?

Table.c
```diff
@@ -0,0 +1,276 @@
...
+   // (e.g. on OpenBSD the kernel thread 'swapper' has pid 0.)
+   if (rowid == 0)
+      return;
+
+   // The vector is sorted by parent, find the start of the range by bisection
+   int vsize = Vector_size(this->rows);
+   int l = 0;
+   int r = vsize;
+   while (l < r) {
+      int c = (l + r) / 2;
```
Explorer09, line 92: Potential integer overflow. Maybe fix this as well?
```suggestion
      int c = l + (r - l) / 2;
```

Row.c
```diff
@@ -0,0 +1,340 @@
...
+         precision = 0;
+         val = 100.0F;
+      }
+
+      xSnprintf(buffer, n, "%*.*f ", width, precision, val);
+   } else {
+      *attr = CRT_colors[PROCESS_SHADOW];
+      xSnprintf(buffer, n, "%*.*s ", width, width, "N/A");
+   }
+}
```
BenBE, line 298: Future work: Split these into some separate UI helpers since output of values and quantities is not intrinsically linked to the `Row` or the DT.

Row.h
BenBE: `Row_IsFiltered` is used to determine if the `Row`  matches a filter defined by the `IncSet` filter for the active `Panel`.

`Row_IsObscured` is a refactor of the check to see if a given `Row` needs to be shown or not; e.g. a LWP being hidden because of `Settings->hideUserlandThreads` being active.

Row.h
```diff
@@ -0,0 +1,141 @@
...
+
+/* Takes time in hundredths of a seconds. Prints 9 columns. */
+void Row_printTime(RichString* str, unsigned long long totalHundredths, bool coloring);
+
+/* Takes rate in bare unit (base 1024) per second. Prints 12 columns. */
+void Row_printRate(RichString* str, double rate, bool coloring);
+
+void Row_printLeftAlignedField(RichString* str, int attr, const char* content, unsigned int width);
+
+void Row_printPercentage(float val, char* buffer, int n, uint8_t width, int* attr);
```
BenBE, line 115: Future work: Cf. remark on implementation of these functions; should be split into an UI helper utilities collection.

Process.h
```diff
@@ -15,6 +15,7 @@ in the source distribution for its full text.
 #include "Object.h"
 #include "ProcessField.h"
 #include "RichString.h"
+#include "Row.h"
```
smalinux, line 18: The `Table_` symbol has been used in this file.
```suggestion
#include "Row.h"
#include "Table.h"
```

Row.h
```diff
@@ -0,0 +1,141 @@
...
+
+void Row_toggleTag(Row* this);
+
+static inline int Row_idEqualCompare(const void* v1, const void* v2) {
+   const int p1 = ((const Row*)v1)->id;
+   const int p2 = ((const Row*)v2)->id;
+   return p1 != p2; /* return zero when equal */
+}
+
+static inline int Row_getGroupOrParent(const Row* this) {
```
BenBE, line 131: Got some docs for where this is used and what for?

Scheduling.c
```diff
@@ -97,7 +97,7 @@ Panel* Scheduling_newPriorityPanel(int policy, int preSelectedPriority) {
    return this;
 }
 
-bool Scheduling_setPolicy(Process* proc, Arg arg) {
+bool Scheduling_setPolicy(Row* row, Arg arg) {
```
BenBE, line 100: Keep API to use processes. Scheduling NICs or namespaces does not make sense. It's the task for the caller to ensure we have a process.

Scheduling.c
```diff
@@ -112,7 +112,7 @@ bool Scheduling_setPolicy(Process* proc, Arg arg) {
       policy &= SCHED_RESET_ON_FORK;
    #endif
 
-   int r = sched_setscheduler(proc->pid, policy, &param);
+   int r = sched_setscheduler(row->id, policy, &param);
```
BenBE, line 115: Be consistent. Use `Process_getPid`.

Scheduling.h
```diff
@@ -38,7 +39,7 @@ typedef struct {
    int priority;
 } SchedulingArg;
 
-bool Scheduling_setPolicy(Process* proc, Arg arg);
+bool Scheduling_setPolicy(Row* proc, Arg arg);
```
BenBE, line 42: Keep API to `Process*`. The caller needs to ensure we got a process.

ScreenManager.c
```diff
@@ -133,13 +133,16 @@ static void checkRecalculation(ScreenManager* this, double* oldTime, int* sortTi
...
          *sortTimeout = 1;
       }
       // sample current values for system metrics and processes if not paused
       Machine_scan(host);
       if (!this->state->pauseUpdate)
-         ProcessList_scan(host->pl);
+         Machine_scanTables(host);
+
+// TODO --         ProcessList_scan(host->activeTable); // TODO - each table needs a scan function
+//
```
BenBE, line 145: Unresolved TODOs in PR? Something missing?

Settings.h
```diff
@@ -104,9 +104,9 @@ typedef struct Settings_ {
 
 #define Settings_cpuId(settings, cpu) ((settings)->countCPUsFromOne ? (cpu)+1 : (cpu))
 
-static inline ProcessField ScreenSettings_getActiveSortKey(const ScreenSettings* this) {
+static inline RowField ScreenSettings_getActiveSortKey(const ScreenSettings* this) {
    return (this->treeView)
```
BenBE, line 108: This should only affect sort order if the presented table contains a process list to keep with the spirit of this setting.

Table.c
```diff
@@ -0,0 +1,276 @@
...
+   assert(Vector_indexOf(this->rows, row, Row_idEqualCompare) != -1);
+   assert(Hashtable_get(this->table, row->id) != NULL);
+   assert(Vector_countEquals(this->rows, Hashtable_count(this->table)));
+}
+
+// Table_removeIndex removes a given row from the lists map and soft deletes
+// it from its vector. Vector_compact *must* be called once the caller is done
+// removing items.
+// Note: for processes, should only be called from ProcessList_scan to avoid
+// breaking dying process highlighting.
```
BenBE, line 62: Future work: May implement new/removed highlighting for other table styles too (e.g. hot-plugging of CPUs, HW components, …

Table.c
BenBE: Due to rounding this might not be equivalent. Needs further checking …

Table.c
```diff
@@ -0,0 +1,276 @@
...
+
+   Panel_prune(this->panel);
+
+   /* Follow main group row instead if following a row that is occluded (hidden) */
+   if (this->following != -1) {
+      const Row* followed = (const Row*) Hashtable_get(this->table, this->following);
+      if ( (followed != NULL)
+         && Hashtable_get(this->table, followed->group)
+         && Row_obscured(followed, this) )
+         this->following = followed->group;
```
BenBE, line 237: Possibly use braces here.
```suggestion
      if (
         (followed != NULL) &&
         Hashtable_get(this->table, followed->group) &&
         Row_obscured(followed, this)
      ) {
         this->following = followed->group;
      }
```
I know that's not the main style we use for wrapping, but IMHO that's much clearer on what part belongs to the condition …

Table.h
```diff
@@ -0,0 +1,87 @@
...
+struct Machine_;
+struct Table_;
+
+typedef void (*Table_scan)(struct Table_ *);
+typedef void (*Table_foreachRowFn)(struct Table_*, struct Row_*, int, void*);
+
+typedef struct Table_ {
+   Vector* rows;          /* all known; sort order can vary and differ from display order */
+   Vector* displayList;   /* row tree flattened in display order (borrowed); updated
+                             in Table_updateDisplayList when rebuilding panel */
```
BenBE, line 36:
```suggestion
   Vector* displayList;   /* row tree flattened in display order (borrowed);
                             updated in Table_updateDisplayList when rebuilding panel */
```

Table.h
```diff
@@ -0,0 +1,87 @@
...
+   Hashtable* table;      /* fast known row lookup by identifier */
+
+   struct Machine_* host;
+   const char* incFilter;
+   bool needsSort;
+   int following;         /* -1 or row being visually tracked in the user interface */
+
+   Table_scan scan;
```
BenBE, line 44: Should be virtual …

BenBE (changes requested):
Only checked the platform-independent code so far. Several remarks and annotations.

One open question is distributing a `Table`'s contents onto different `Screen`s while keeping single instances of various table types as needed (n:m linkage).

Action.c
```diff
@@ -262,9 +262,9 @@ static Htop_Reaction actionToggleTreeView(State* st) {
    ss->treeView = !ss->treeView;
 
    if (!ss->allBranchesCollapsed)
-      ProcessList_expandTree(host->pl);
+      Table_expandTree(host->activeTable);
```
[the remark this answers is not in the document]
smalinux: If one instance from "Table" represents the whole screen... maybe "Screen" is a better name...

Process.h
Explorer09: I think I'm still confused about the purpose of the `Row` class here. Because in the current PR, the `Row` object contains members that describe tree relations - the tree releations may not be applicable to future things that could derive from this `Row` class. Such members shouldn't belong to the `Row`, AFAICT.

Is the `Row` class supposed to be like a `ListBoxItem`, which manages the display logic and the handler routine upon mouse click? Or is it supposed to be something else?

Row.h
Explorer09: I have to update and suggest better names for these functions:

* `Row_Filtered` renamed to `Row_MatchesFilter`
* `Row_Obscured` renamed to `Row_IsVisible` and the semantics reversed.

Process.h
BenBE: If `ProcessList` was not used before, `Table` shouldn't be used now either …

natoscott:
> [...] Would these names be better?

I don't think so, personally.  I've chosen Table and Row to match with the existing use of Column throughout the code base, which is not prefixed in any way to refine/clarify its meaning - it's self evident in the Column case and I think the same is true for Row and Table.  These are also the simplest possible names for these things, which is best for a base class IMO.

> I didn't read the entire commit, but I have one comment here:  [...]
> I think I'm still confused about the purpose of the Row class here.  [...]

These aspects will be clearer once there's a second user of these classes (final commit after this one, which will be done as soon as time is available - it's an adaptation of #1102 to this though).  This PR is not for approving/merging until that arrives, more to get early feedback (thanks!) and to give a heads-up that's (still) progressing.

> [...] Make it a composition relation instead.

I agree with @BenBE here, I'm finding its a much better fit to have these as 'inherited' base classes.

I'm juggling a few projects here, context-switching alot, but will get updated code and the final PR in this series open ASAP for more discussion.

Row.h
natoscott: Yep, agreed, that's neater - I'll make those name changes, thanks @Explorer09

Table.c
natoscott: Unrelated to this commit FWIW, this is not new code - it's been refactored from ProcessList.c.

Action.c
```diff
@@ -56,22 +56,22 @@ Object* Action_pickFromVector(State* st, Panel* list, int x, bool followProcess)
    Panel* panelFocus;
    int ch;
    bool unfollow = false;
-   int pid = followProcess ? MainPanel_selectedPid(mainPanel) : -1;
-   if (followProcess && host->pl->following == -1) {
-      host->pl->following = pid;
+   int row = follow ? MainPanel_selectedRow(mainPanel) : -1;
```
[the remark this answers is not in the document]
natoscott: I'll see what I can do.

Action.c
```diff
@@ -121,36 +121,36 @@ bool Action_setUserOnly(const char* userName, uid_t* userId) {
...
-static void tagAllChildren(Panel* panel, Process* parent) {
+static void tagAllChildren(Panel* panel, Row* parent) {
    parent->tag = true;
-   pid_t ppid = parent->pid;
+   int parent_id = parent->id;
    for (int i = 0; i < Panel_size(panel); i++) {
-      Process* p = (Process*) Panel_get(panel, i);
-      if (!p->tag && Process_isChildOf(p, ppid)) {
-         tagAllChildren(panel, p);
+      Row* row = (Row*) Panel_get(panel, i);
```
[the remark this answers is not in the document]
natoscott: I don't think that's going to fly because Panels hold other things too beyond just Process/Row.

Action.c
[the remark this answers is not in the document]
natoscott: Heh, yes - that 'problem' exists already with 'Process**List**' which is also not a tree :) ... I don't think this is fixable (or even needs to be, as you say - its a feature not a bug that htop can manage rows so cleverly).

Action.c
```diff
@@ -529,6 +529,8 @@ static Htop_Reaction actionLsof(State* st) {
    if (!p)
       return HTOP_OK;
 
+   assert(Object_isA((const Object*) p, (const ObjectClass*) &Process_class));
```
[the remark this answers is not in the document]
natoscott: Sure thing, see where else I can add this pattern.

Table.c
natoscott: OK, maybe - I'll see if I can make it clearer somehow.

Table.h
natoscott: Yep, I'm totally reworking this aspect - this function will be removed/reworked to be more in the htop OO style (we need a couple of other Table methods too, I'm finding).

Process.h
natoscott: Thanks, I'll drop this include in the next update.

Machine.c
natoscott: Thanks, I've fixed this.

Process.h
natoscott: As mentioned elsewhere, I'm with @BenBE on this one.  Leaving this conversation as 'open' for now though, until the next classes using Row & Table (beyond Process & ProcessList) arrive.

Row.h
natoscott: This is done now in latest update, thanks @Explorer09

Action.c
```diff
@@ -56,22 +56,22 @@ Object* Action_pickFromVector(State* st, Panel* list, int x, bool followProcess)
...
    }
    ScreenManager_delete(scr);
    Panel_move((Panel*)mainPanel, 0, y);
    Panel_resize((Panel*)mainPanel, COLS, LINES - y - 1);
    if (panelFocus == list && ch == 13) {
-      if (followProcess) {
-         const Process* selected = (const Process*)Panel_getSelected((Panel*)mainPanel);
-         if (selected && selected->pid == pid)
+      if (follow) {
+         const Row* selected = (const Row*)Panel_getSelected((Panel*)mainPanel);
```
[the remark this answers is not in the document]
natoscott: As discussed elsewhere, there are non-Row entries in Panels, so marking this resolved now.

Table.c
natoscott: @BenBE agreed.  (cgroups too)

Table.c
natoscott: I think the single-command with braces part makes this sufficiently clear now.  I don't really like the round-brackets-on-own-lines thing, its not really used anywhere else in htop.  There was also some extra parenthesis in there I've removed now that weren't helping.

Table.h
natoscott: This is resolve now - Table became a "proper" base class with its own additional methods relating to Row scanning, and all the remaining presentation-related code moved from ProcessList into Table as a result.  Looks much better now IMO, thanks @BenBE

Explorer09:
> I don't think so, personally. I've chosen Table and Row to match with the existing use of Column throughout the code base, which is not prefixed in any way to refine/clarify its meaning - it's self evident in the Column case and I think the same is true for Row and Table.

What existing Column code base are you referring to? Because I don't see any Column.c or Column.h files.

> These are also the simplest possible names for these things, which is best for a base class IMO.

I disagree when any one can look at the code and ask "why isn't a Hashtable derived from a Table", for example. Even a class name like `DataTable` can clarify a lot of things that a generic `Table` name cannot. And this new code is meant to be maintained long term.

> These aspects will be clearer once there's a second user of these classes (final commit after this one, which will be done as soon as time is available - it's an adaptation of #1102 to this though). This PR is not for approving/merging until that arrives, more to get early feedback (thanks!) and to give a heads-up that's (still) progressing.

I'll wait and see. But for now, I don't think the current code refactoring make sense.

Explorer09:
Two more comments:
1. When a `Process` object can have PID as a [primary key](https://en.wikipedia.org/wiki/Primary_key) that uniquely identifies the object and be use for sorting, I don't know if it make sense to have that `id` applied to any other kind of things htop can monitor (cgroups, filesystems, etc.) Maybe other objects should use different primary keys. This, I kind of disagree when `Row` contains an `id` member and that `Process.pid` derives from it.
2. The `RowField` name. Maybe you meant `Field` instead (or `DataField` if clarity is needed) so that the `Row` class can become `Record` (`DataRecord`). That's one of the awkward class naming I was complaining about.

solaris/SolarisProcessList.c
```diff
@@ -46,17 +46,18 @@ static char* SolarisProcessList_readZoneName(kstat_ctl_t* kd, SolarisProcess* sp
 
 ProcessList* ProcessList_new(Machine* host, Hashtable* pidMatchList) {
    SolarisProcessList* spl = xCalloc(1, sizeof(SolarisProcessList));
```
BenBE, line 48:
```suggestion
   SolarisProcessList* this = xCalloc(1, sizeof(SolarisProcessList));
```

solaris/SolarisProcessList.c
```diff
@@ -46,17 +46,18 @@ static char* SolarisProcessList_readZoneName(kstat_ctl_t* kd, SolarisProcess* sp
...
-   ProcessList* pl = (ProcessList*) spl;
+   Object_setClass(this, Class(ProcessList));
 
+   ProcessList* pl = (ProcessList*) spl;
    ProcessList_init(pl, Class(SolarisProcess), host, pidMatchList);
 
    return pl;
```
BenBE, line 54:
```suggestion
   ProcessList_init(super, Class(SolarisProcess), host, pidMatchList);

   return super;
```

unsupported/UnsupportedProcessList.c
```diff
@@ -15,15 +15,18 @@ in the source distribution for its full text.
 
 
 ProcessList* ProcessList_new(Machine* host, Hashtable* pidMatchList) {
-   ProcessList* this = xCalloc(1, sizeof(ProcessList));
+   UnsupportedProcessList* this = xCalloc(1, sizeof(UnsupportedProcessList));
+   Object_setClass(this, Class(ProcessList));
 
-   ProcessList_init(this, Class(Process), host, pidMatchList);
+   ProcessList* pl = (ProcessList*) this;
```
BenBE, line 21:
```suggestion
   ProcessList* super = &this->super;
```

linux/LinuxProcess.c
```diff
@@ -404,10 +406,15 @@ static int LinuxProcess_compareByKey(const Process* v1, const Process* v2, Proce
...
+         .display = Process_display,
+         .delete = Process_delete,
+         .compare = Process_compare
+      },
+      .isVisible = Process_isVisible,
+      .matchesFilter = Process_matchesFilter,
+      .compareByParent = Process_compareByParent,
    },
    .writeField = LinuxProcess_writeField,
    .compareByKey = LinuxProcess_compareByKey
```
BenBE, line 420:
```suggestion
   .compareByKey = LinuxProcess_compareByKey,
```

BenBE (changes requested):
Okay, some more comments for the stuff I didn't look at previously. Mostly consistency fixes, nothing too special. Overlaps in big parts with things already noted for the platform-independent stuff.

Not too happy yet to have tgpid (group) and uid as part of the Row class …

Action.c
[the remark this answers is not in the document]
natoscott: I've pushed this followProcess change in as an independent commit.

Row.h
```diff
@@ -0,0 +1,218 @@
...
+typedef struct Row_ {
+   /* Super object for emulated OOP */
+   Object super;
+
+   /* Pointer to quasi-global data */
+   const struct Machine_* host;
+
+   int id;
+   uid_t uid;
+   int group;
```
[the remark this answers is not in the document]
BenBE: Ah, k. I think for the sake of simplicity™ of this PR, we can leave them there for now.

Future work: Keep uid/gid as part of the `Process`, but provide an interface from `Row` to control the associated behaviour. In particular for the group handling this could be mostly loosely mirroring the `gid` field onto `group` (not equal, as things like "applications" and other concepts could hijack the group field for advanced usage.

For the `uid` field, the highlighting aspect would probably be better suited as its own API (for passing display attributes like background, new, dead, …) too; mostly triggered from the Process class instead.

natoscott:
| What existing Column code base are you referring to? Because I don't see any Column.c or Column.h files.

Ah that's a literal interpretation of what I meant - sorry, wasn't clear - what I more meant was we don't see naming like "ColumnUI" - there's lots of things that just talk just about Columns (in various contexts).

Your suggestion of "Record" may well be a better name than "Row" - it was one of the options considered earlier.  Row is shorter, which is goodness, and "Record" is slightly ambiguous in that it also has "saving data to disk" connotations.  Naming is hard.

In terms of the "not liking" the refactoring, maybe it will help to forget about the PCP use case and think more from the perspective of the cgroups discussion in #1191 - i.e. how would we need to refactor the code so that cgroups could be shown in the way processes are today (with tree-view as well, etc)?  We definitely need to do exactly the sorts of things being done in this PR.

This is how we've ended up with this refactoring - we have the initial PCP implementation that doesn't do this (ends up duplicating lots of Process.c/ProcessList.c code) and doesn't help the other htop platforms to move forward and allow for other types of Screens in the future.  Hence, we're heading this way to try prepare the way for other platforms that might want to do this kind of thing too.

Row.h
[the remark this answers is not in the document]
Explorer09: @natoscott 
How about you add a virtual function `IsHighlighted` on the `Row` class so that the `uid` member be moved to `Process`?
`Process_IsHighlighted` would then control whether a data row would be highlighted.

Explorer09:
@natoscott I'm okay when you have something in mind when you refactor the code with the new classes. I just didn't know how would, e.g, cgroups be based on these new classes, and thus missed the goal you have for this refactoring.

Can you provide us a link to the ”initial PCP implementation” you are talking about, just to get a clearer vision on how the base class `Row` can be structured?

> Your suggestion of "Record" may well be a better name than "Row" - it was one of the options considered earlier. Row is shorter, which is goodness, and "Record" is slightly ambiguous in that it also has "saving data to disk" connotations. Naming is hard.

I was also considering the name like `DataEntry` or maybe even `Node` (the term "node" makes sense in the context of tree data structures). But I think it's better to forget the naming for now, and focus on finishing the code structure. A good class name may come out later naturally when things are almost finished.

natoscott:
An example Filesystem dynamic screen.
Screenshot 2023-06-19 at 4 15 03 pm

natoscott:
An example of what the Screens setup page looks like with dynamic screens.
Screenshot 2023-06-19 at 4 16 05 pm

Explorer09:
@natoscott The dynamic screen feature for PCP looks interesting. Before I try to review the code, I have one comment on the Setup screen UI:

The "Available" list (list of available screens) may be merged with the "Available columns" list (list of available columns of that screen). This would save some horizontal space as there are too many columns on that particular Setup screen.

DynamicScreen.c
```diff
@@ -0,0 +1,64 @@
...
+static void DynamicScreen_compare(ht_key_t key, void* value, void* data) {
+   const DynamicScreen* screen = (const DynamicScreen*)value;
+   DynamicIterator* iter = (DynamicIterator*)data;
+   if (String_eq(iter->name, screen->name)) {
+      iter->found = true;
+      iter->key = key;
+   }
+}
+
+bool DynamicScreen_search(Hashtable* dynamics, const char* name, ht_key_t* key) {
```
BenBE, line 52: `dynamics` should be better reflect what it contains. What about `screens` or `dyn_screens`?

Machine.c
```diff
@@ -15,16 +15,23 @@ in the source distribution for its full text.
...
 
+   this->htopUserId = getuid();
+
+   // discover fixed column width limits
+   Row_setPidColumnWidth(Platform_getMaxPid());
+
    // always maintain valid realtime timestamps
    Platform_gettime_realtime(&this->realtime, &this->realtimeMs);
 
+
```
BenBE, line 34: No need for this extra blank line …

Machine.c
```diff
@@ -49,12 +56,73 @@ void Machine_done(Machine* this) {
    if (this->topologyOk) {
       hwloc_topology_destroy(this->topology);
    }
-#else
-   (void)this;
 #endif
+   Object_delete(this->processTable);
```
BenBE, line 60: Maybe add some blank line before …

MainPanel.c
```diff
@@ -30,25 +30,25 @@ void MainPanel_updateLabels(MainPanel* this, bool list, bool filter) {
    FunctionBar_setLabel(bar, KEY_F(4), filter ? "FILTER" : "Filter");
 }
 
-static void MainPanel_pidSearch(MainPanel* this, int ch) {
+static void MainPanel_idSearch(MainPanel* this, int ch) {
    Panel* super = (Panel*) this;
-   pid_t pid = ch - 48 + this->pidSearch;
+   int id = ch - 48 + this->idSearch;
```
BenBE, line 35: Integer truncation compared to previously on systems with pid_t being 64-bit.

Settings.h
```diff
@@ -34,15 +37,17 @@ typedef struct {
 } MeterColumnSetting;
 
 typedef struct {
```
BenBE, line 39: Shouldn't this be called like this?
```suggestion
typedef struct ScreenSettings_ {
```

pcp/Instance.c
```diff
@@ -0,0 +1,160 @@
+/*
+htop - Instance.c
+(C) 2022-2023 Sohaib Mohammed
+(C) 2022-2023 htop dev team
```
BenBE, line 4: Please check consistency/order of copyright headers for the newly created PCP platform files.

Also the filename above should include the pcp/ path prefix, as present with other files.

pcp/PCPDynamicColumn.h
```diff
@@ -11,15 +11,22 @@
...
 
+struct pmDesc;
+
 typedef struct PCPDynamicColumn_ {
    DynamicColumn super;
    char* metricName;
+   char* format;
    size_t id;  /* identifier for metric array lookups */
+   int width;  /* optional width from configuration file */
+   bool enabled;  /* default enabled in dynamic screen */
```
BenBE, line 22: Missing include for `<stdbool.h>` …

pcp/screens/filesys
```diff
@@ -0,0 +1,50 @@
+#
```
BenBE, line 1: Rename to `filesystem`, `filesystems` or just `fs`.

unsupported/UnsupportedProcess.c
```diff
@@ -94,11 +94,17 @@ static int UnsupportedProcess_compareByKey(const Process* v1, const Process* v2,
...
+         .compare = Process_compare
+      },
+      .isVisible = Process_isVisible,
+      .matchesFilter = Process_matchesFilter,
+      .compareByParent = Process_compareByParent,
+      .sortKeyString = Process_getSortKey,
+      .writeField = UnsupportedProcess_writeField
    },
-   .writeField = UnsupportedProcess_writeField,
    .compareByKey = UnsupportedProcess_compareByKey
```
BenBE, line 109:
```suggestion
   .compareByKey = UnsupportedProcess_compareByKey,
```

natoscott:
> The "Available" list (list of available screens) may be merged with the "Available columns" list (list of available columns of that screen). This would save some horizontal space as there are too many columns on that particular Setup screen.

We would like to keep the existing htop functionality of allowing the user to choose the displayed screen name (i.e. the string shown in the tab above each Screen).  Ultimately, this means these two columns need to be separate - we need a list of (not editable) screen names (i.e. 'Processes' from htop core, followed by the list of dynamic screens from configuration files) and a list of (editable) names that the user selects - the latter is based on the former, but they are different things.

In the example Setup screenshot I posted, we can see "Main" and "I/O" are two differently-configured "Processes" Screens.  These names can be edited as the user wishes, but since we now have more than just Processes, a separate (non-editable) column is needed to indicate what the underlying Screen is.

In a similar way to Main and I/O, we can also have zero or more of each of the dynamic screens.  So we could have users with multiple different Filesystem tab configurations, for example, each with different display names and different visible columns.

Explorer09:
@natoscott 
> In the example Setup screenshot I posted, we can see "Main" and "I/O" are two differently-configured "Processes" Screens. These names can be edited as the user wishes, but since we now have more than just Processes, a separate (non-editable) column is needed to indicate what the underlying Screen is.
> 

You know, I cannot tell that from the screenshot you provided. I think there is a usability problem here that users cannot tell the type of the screen ("Processes") from the screen with the custom name. And the column of available screens won't help with the problem.

My idea is to list the active screens with the type of the screen as well as the name. Under the "Active Screens" heading, show something like this:

```
Main (Processes)
I/O (Processes)
Cgroups (Cgroups)
Filesystems (Filesystems)
```

The "Available Screens" should only be there if the user has pressed the "New" button to select a screen they wish to add.

Explorer09:
Screenshot

P.S. From this screenshot I saw another confusion. When the "Filesystems" screen is highlighted in the "Active" column, the "Available" column does not reflect to what the type of the "Active" screen is. Here, "Processes" is highlighted but I expect "Filesystems" be highlighted. I'm just to tell there is usability bug even with the current approach of listing available screens and available columns separately.

natoscott:
| [...] usability problem here that users cannot tell the type of the screen ("Processes") from the screen with the custom name.

Yep, I agree - I'll look into improving this.  I don't love the idea of merging the non-editable Available column with the editable Active column using brackets - I think that'll make for a confusing interface as well.  We may be able to solve it more simply though if we make a completely left-to-right interaction, such that the Active Screens only shows those Screens active for the currently selected row in the Available column to its left.  This approach addresses the point you make re "I saw another confusion" too I think.

I'll hack some more on this and update the PR soon.  Thanks for all the feedback!

BenBE:
As changing the type once a screen has been added is likely undesired, what about hiding the list of available types unless a new screen is currently being added (add the type after the name of the screen). The available+active columns should only be shown when a screen is currently edited, but not added (read: created).

FWIW: Force the ID column to be always included; make the PID of a process a distinct entity.

Explorer09:
@natoscott 
> I don't love the idea of merging the non-editable Available column with the editable Active column using brackets 

No, you mistook my idea. I was saying that the ”Available Screens" column and "Available Columns" column may share the same UI space. Since you would only use one UI and not both at the same time. (Select "New Screen" function -> show Available Screens -> user selects one type of screen -> UI changes to the list of Active and Available Columns)

natoscott:
> 

Oh interesting, and yep, I misunderstood.  Let me experiment further.  I'm away on holidays for a bit, but will get back to you (both) in a couple of weeks - thanks again!

natoscott:
Hey all - I've been pondering all the ideas for awhile and think I may have a new, simpler way forward.

@Explorer09 I've spent a long time thinking about your UI idea.  If I understand correctly, I think changing the contents and meaning of a column on-the-fly like that is going to cause more usability problems than it solves.  I can't find anything in htop that behaves in any way similar to that - columns are always fixed in terms of what they represent once initially displayed.  IMO that's good UI design and follows the principle of least surprise.

However, I think the underlying point is valid - having 5 columns makes for quite a busy setup screen needing plenty of screen real estate, and it may be too much information to be presenting all at once.

What I suggest we do instead is as follows:
- keep the F2/Setup -> Screens page exactly as it is today, for all platforms that don't have dynamic screens.
- for those platforms that do (just PCP, so far) remove the 'F5 New' option when something in the Screens column is selected (since there's no way to associate a new screen with anything other than processes).
- for those platforms, add an additional F2/Setup Categories column entry, like "Tabs", "Tab layout" or something like that, in-between the Meters and Screens categories.
- selecting this would then go to a new 2-column setup screen with just the Available and Active screens columns, providing the ability to add and (re-)configure Active screens from the available dynamic screens.

Thoughts?  Thanks.

BenBE:
LGTM. Just one question: Why split out PCP here instead of refactoring into that new UI too?

natoscott:
> 

@BenBE I don't understand the question - my explanation was probably unclear.  There's nothing PCP-specific about it, the decision as to whether the additional UI element is used is based on whether the platform has support for dynamic screens (i.e. displaying values for non-process tables).  Currently that's only PCP but any platform can use this, e.g. we could add code to the Linux platform for cgroups, in which case Linux would also have this additional UI configuration.

BenBE:
@natoscott Framed that way this makes more sense. Also there's network interfaces+I/O, disks/other storage media, …

BTW: Re generalized table cf. #1270 for one application where generic table handling could come in handy …

Explorer09:
>
> * for those platforms, add an additional F2/Setup Categories column entry, like "Tabs", "Tab layout" or something like that, in-between the Meters and Screens categories.
> 
> * selecting this would then go to a new 2-column setup screen with just the Available and Active screens columns, providing the ability to add and (re-)configure Active screens from the available dynamic screens.

I'm not a fan of this idea. The reason is that, you'll not be able to easily tell which "Tab" / "Screen" you are configuring when you are adjusting the Columns of that Screen.

It should be one setup category for adjust the number of Screens and the Columns of each Screen. That way the user's flow of creating one Screen would be streamlined, like this:

1. The user presses "New Screen" on the function bar.
2. Select the screen to add from the popped-up "Available screens" list.
3. The "Available screens" list disappears and the default list of columns would show in it's place. The user can then add, remove, or reorder the columns of the screen.

natoscott:
|  [...] you'll not be able to easily tell which "Tab" / "Screen" you are configuring when you are adjusting the Columns of that Screen.

You will, the name is shown in the first column of the Screens setup page (no change there).  This is the users chosen Tab name, not the underlying dynamic screen name, but I think that's OK TBH.  Also, the Available Columns list will reflect the choices for the selected screen, so no mistakes can be made there.

We can't have it both ways IMO - we tradeoff either more information on one (busy) screen, or split the configuration in two stages like I've proposed here.  For me, changing the semantics of a column on-the-fly is confusing UI design, I don't think we can go that path.

Explorer09:
@natoscott But the flow would no longer be straightforward.

Yours flow for creating a new Tab/Screen would look like:
1. In "Tabs" setup category, select a new screen to add in the "Available screens" column.
2. Press "Add".
3. Navigate to the "Columns" setup category manually, select the screen you just added, and see the default list of columns.
4. Edit the list of columns there.

natoscott:
> 

That's correct - making it a two-step process is the tradeoff to reduce the busy-ness of the Screens setup page.

BenBE:
> […] is confusing UI design, I don't think we can go that path.

We _can_, but we absolutely **should not** … ;-)

Jokes aside.

I've been thinking about some minimalistic notion of "dialogs" for a while, like Midnight Commander does for various operations that require user input. In the case of htop we could do something akin to:

```

   0[||||||||||||          42%] Load: 0.00 0.00 0.00

   Setup
Categories       +-> Screen setup for Main [Processes] ---------------
Display Options  | Active Columns       Available Columns
Header Layout    | PID                  PID - Process/thread ID
Meters           | PPID                 STATE - Process state
Screens          | USER                 PPID - Parent process ID
Colors           | PRIORITY             PGRP - Process group ID
                 | …                    …
```

On the main `Screens` selector you just have the list of screens and the list of screen types available:

```

   0[||||||||||||          42%] Load: 0.00 0.00 0.00

   Setup
Categories       Screens                Screen Types
Display Options  Main [Process]         Processes
Header Layout    I/O [Process]          CGroups
Meters           Delay [Process]        Network
Screens          Network [Network]      Storage
Colors           Mounts [Storage]       GPU
                 …                      …
```

That way the UI is minimized for the actual task at hand and the user is only presented with the options necessary at each moment.

Explorer09:
@BenBE I'm not sure if I get it. The "Screen" and "Screen Types" lists in your UI proposal looks okay, but I have one question:
When is the "Screen Types" list shown, and how would the user's flow be for adding a screen, modifying a screen, and deleting a screen?

@natoscott In your proposal, how would the user's flow be for deleting a screen? How about renaming a screen (changing the display name of the tab)?

BenBE:
> @BenBE I'm not sure if I get it. The "Screen" and "Screen Types" lists in your UI proposal looks okay, but I have one question: When is the "Screen Types" list shown, and how would the user's flow be for adding a screen, modifying a screen, and deleting a screen?

Adding screen: Press function key according to Status Bar, enter name, select type.

Renaming screen: Press function key according to Status Bar, update name, confirm with Enter.

Modifying type: Ensure screen has no columns, select new type from list of available types

Delete screen: Press function key according to Status Bar.

Add/Modify/Remove columns: Navigate to screen, hit Enter.

> @natoscott In your proposal, how would the user's flow be for deleting a screen? How about renaming a screen (changing the display name of the tab)?

Explorer09:
> Modifying type: Ensure screen has no columns, select new type from list of available types

My expectation is the screen type would not be modifiable after creation. The user would have to delete the screen and then create another with the right type.
The reason? There would be a lot of confusion if we allow the user to modify the type, when the user would see all the columns they have customized just go away, without a warning or a backtrack.

BenBE:
> > Modifying type: Ensure screen has no columns, select new type from list of available types
> 
> My expectation is the screen type would not be modifiable after creation. The user would have to delete the screen and then create another with the right type. The reason? There would be a lot of confusion if we allow the user to modify the type, when the user would see all the columns they have customized just go away, without a warning or a backtrack.

Thus my note about "requires all columns deleted" for changing the type. Effectively boils down to deleting and re-creating that screen (sans re-entering the name).

natoscott:
@BenBE interesting stuff - I like your idea of dialogs.

It's quite a change from the current model though, so I'm not going to attempt it within this PR.  :)

> @natoscott how would the user's flow be for deleting a screen? How about renaming a screen (changing the display name of the tab)?

No change to how its done today from ScreensPanel.  It can also be achieved from the (new) ScreenTabsPanel now, for platforms with dynamic screens.

I've got an initial implementation going, working sufficiently well that the interaction can be experimented with and it feels quite intuitive to me (and is consistent with existing htop UI).  There's still some bugs to be sorted out though, I'll continue working on those.

Affinity.c
```diff
@@ -52,9 +52,9 @@ void Affinity_add(Affinity* this, unsigned int id) {
 
 #if defined(HAVE_LIBHWLOC)
 
-Affinity* Affinity_get(const Process* proc, Machine* host) {
+Affinity* Affinity_get(const Row* row, Machine* host) {
```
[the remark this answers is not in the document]
natoscott: The problem is Affinity_set is a callback used with MainPanel_foreachRow (we could change Affinity_get to take Process* and leave Affinity_set as-is, but that looks pretty strange and I'd prefer not that).

I'll add an explicit cast and type-check at the start of these functions to verify process-ness of the Row pointer.

Affinity.h
```diff
@@ -16,7 +16,7 @@ in the source distribution for its full text.
 #include <stdbool.h>
 
 #include "Object.h"
-#include "Process.h"
+#include "Row.h"
```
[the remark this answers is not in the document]
natoscott: Same explanation as above - Affinity_set is a panel callback.

Action.c
```diff
@@ -556,6 +561,7 @@ static Htop_Reaction actionStrace(State* st) {
    const Process* p = (Process*) Panel_getSelected((Panel*)st->mainPanel);
    if (!p)
       return HTOP_OK;
+   assert(Object_isA((const Object*) p, (const ObjectClass*) &Process_class));
```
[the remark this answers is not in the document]
natoscott: Done.

Action.c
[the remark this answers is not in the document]
natoscott: Bunch more places have this check now, everywhere I can see that makes sense.

CommandLine.c
```diff
@@ -363,7 +363,7 @@ int CommandLine_run(int argc, char** argv) {
    CRT_init(settings, flags.allowUnicode);
 
    MainPanel* panel = MainPanel_new();
-   ProcessList_setPanel(pl, (Panel*) panel);
+   Table_setPanel(&pl->super, (Panel*) panel);
```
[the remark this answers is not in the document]
natoscott: If that was the case it would be a problem already, independent of Dynamic Screens, since this is how processes and (multiple) screen panels are setup today.  Changing this model would be a major deviation from the current code.  I'm inclined to leave this to a possible follow-up commit, if anyone wants to rework this: it's not a new issue that has been introduced here.

CommandLine.c
```diff
@@ -384,13 +384,12 @@ int CommandLine_run(int argc, char** argv) {
    ScreenManager_add(scr, (Panel*) panel, -1);
 
    Machine_scan(host);
-   ProcessList_scan(pl);
+   ProcessList_scan(&pl->super);
```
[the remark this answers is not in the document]
natoscott: Agreed.  This is resolved by further refactoring and cleanup, with list scanning moving to Machine.c and becoming more generic.

CommandLine.c
```diff
@@ -384,13 +384,12 @@ int CommandLine_run(int argc, char** argv) {
...
    CommandLine_delay(host, 75);
-   Machine_scan(host);
-   ProcessList_scan(pl);
+   ProcessList_scan(&pl->super);
```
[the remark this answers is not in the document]
natoscott: +1 - resolved now.

Machine.c
```diff
@@ -54,7 +54,23 @@ void Machine_done(Machine* this) {
...
+   Table** tables = (Table**) xReallocArray(this->tables, nmemb, sizeof(Table*));
+   if (tables) {
+      tables[nmemb - 1] = table;
+      this->tables = tables;
+      this->tableCount++;
+   }
+}
+
+void Machine_scanTables(Machine* this) {
+   for (unsigned int i = 0; i < this->tableCount; i++) {
```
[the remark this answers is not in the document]
natoscott: +1 - done.

Machine.c
```diff
@@ -54,7 +54,23 @@ void Machine_done(Machine* this) {
...
-void Machine_addList(Machine* this, struct ProcessList_ *pl) {
-   // currently only process lists are supported
-   this->pl = pl;
+void Machine_addTable(Machine* this, Table* table, bool processes) {
+   if (processes)
+      this->processTable = table;
+   this->activeTable = table;
+
+   size_t nmemb = this->tableCount + 1;
+   Table** tables = (Table**) xReallocArray(this->tables, nmemb, sizeof(Table*));
```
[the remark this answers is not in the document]
natoscott: Yep, done.

Machine.h
```diff
@@ -68,9 +69,10 @@ typedef struct Machine_ {
    UsersTable* usersTable;
    uid_t userId;
 
-   /* To become an array of lists - processes, cgroups, filesystems,... etc */
-   /* for now though, just point back to the one list we have at the moment */
-   struct ProcessList_ *pl;
+   unsigned int tableCount;
```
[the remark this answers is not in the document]
natoscott: Fixed up, thanks.

Machine.h
```diff
@@ -68,9 +69,10 @@ typedef struct Machine_ {
...
+   Table **tables;
```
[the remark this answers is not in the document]
natoscott: Because that'd be overkill? - same reason we don't use Vectors for the settings file and screens and other simple things.  We don't need complex operations on this array like insertion in the middle, sorting of elements, complex resizing with compaction, and so on.  It is just a simple old array that gets setup once and then occasionally walked left-to-right.

MainPanel.c
```diff
@@ -142,35 +142,35 @@ static HandlerResult MainPanel_eventHandler(Panel* super, int ch) {
...
-int MainPanel_selectedPid(MainPanel* this) {
-   const Process* p = (const Process*) Panel_getSelected((Panel*)this);
-   if (p) {
-      return p->pid;
+int MainPanel_selectedRow(MainPanel* this) {
+   const Row* row = (const Row*) Panel_getSelected((Panel*)this);
+   if (row) {
+      return row->id;
    }
    return -1;
```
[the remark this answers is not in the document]
natoscott: Much neater, done!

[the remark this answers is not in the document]
natoscott: Done.

MainPanel.c
```diff
@@ -142,35 +142,35 @@ static HandlerResult MainPanel_eventHandler(Panel* super, int ch) {
...
    Panel* super = (Panel*) this;
    bool ok = true;
    bool anyTagged = false;
    for (int i = 0; i < Panel_size(super); i++) {
-      Process* p = (Process*) Panel_get(super, i);
-      if (p->tag) {
-         ok = fn(p, arg) && ok;
+      Row* row = (Row*) Panel_get(super, i);
+      if (row->tag) {
+         ok = fn(row, arg) && ok;
```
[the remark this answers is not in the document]
natoscott: OK, sure, done.

OpenFilesScreen.c
```diff
@@ -75,9 +75,9 @@ OpenFilesScreen* OpenFilesScreen_new(const Process* process) {
    OpenFilesScreen* this = xMalloc(sizeof(OpenFilesScreen));
```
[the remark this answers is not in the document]
natoscott: Sure, done.

Process.c
```diff
@@ -1084,30 +815,59 @@ void Process_done(Process* this) {
...
+      return Process_isThread(this);
+   return false;
+}
+
+/* Test whether display must filter out this process (various mechanisms) */
+bool Process_filtered(const Row* super, const Table* table) {
+   const Process* this = (const Process*) super;
+   const Machine* host = table->host;
+
+   if (host->userId != (uid_t) -1 && (this->st_uid != host->userId))
```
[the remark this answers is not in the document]
natoscott: Agreed, done, thanks.

Process.c
```diff
@@ -1084,30 +815,59 @@ void Process_done(Process* this) {
...
+      return true;
+   const char* incFilter = table->incFilter;
+   if (incFilter && !(String_contains_i(Process_getCommand(this), incFilter, true)))
```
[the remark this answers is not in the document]
natoscott: Yep, done.

Process.c
```diff
@@ -1084,30 +815,59 @@ void Process_done(Process* this) {
...
+      return true;
+   const ProcessList* pl = (const ProcessList*) host->activeTable;
```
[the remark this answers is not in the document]
natoscott: Good idea, added.

Process.c
```diff
@@ -1084,30 +815,59 @@ void Process_done(Process* this) {
...
+   if (pl->pidMatchList && !Hashtable_get(pl->pidMatchList, Process_getThreadGroup(this)))
+      return true;
+   return false;
+}
+
 const ProcessClass Process_class = {
```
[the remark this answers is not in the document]
natoscott: Done.

Process.c
```diff
@@ -1134,45 +890,57 @@ bool Process_isTomb(const Process* this) {
    return this->tombStampMs > 0;
 }
 
-bool Process_setPriority(Process* this, int priority) {
+bool Process_setPriority(Row* super, int priority) {
```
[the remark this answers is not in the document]
natoscott: That's a good point - and this may be a better way to handle the Affinity.c situation as well.  I'll fix 'em all up, thanks.

Row.h
natoscott: Added.

Scheduling.c
natoscott: Done, using same model as Affinity.c discussion earlier now.

Scheduling.c
natoscott: Yep, done.

Scheduling.h
natoscott: Reworked this now, thanks.

Settings.h
natoscott: I think it depends on whether any other code might want to use the struct without pulling in the header, but sure, I'll update it - may as well be same as struct Settings here (though ScreenDefaults and MeterColumnSetting in the same file don't use this technique - *shrug*)

Settings.h
natoscott: Hmm?  Any table can have an active sort key (column), its not limited to processes.

Table.c
natoscott: @Explorer09 @BenBE as this is existing code and unrelated to this commit, I'm going to leave this with you to follow up on (please).

linux/LinuxProcess.c
```diff
@@ -143,22 +143,23 @@ static int LinuxProcess_effectiveIOPriority(const LinuxProcess* this) {
...
-bool LinuxProcess_setIOPriority(Process* this, Arg ioprio) {
+bool LinuxProcess_setIOPriority(Row* super, Arg ioprio) {
 // Other OSes masquerading as Linux (NetBSD?) don't have this syscall
 #ifdef SYS_ioprio_set
-   syscall(SYS_ioprio_set, IOPRIO_WHO_PROCESS, this->pid, ioprio.i);
+   syscall(SYS_ioprio_set, IOPRIO_WHO_PROCESS, super->id, ioprio.i);
 #endif
-   return (LinuxProcess_updateIOPriority((LinuxProcess*)this) == ioprio.i);
+   return (LinuxProcess_updateIOPriority(super) == ioprio.i);
 }
```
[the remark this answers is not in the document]
natoscott: Agreed, done.

linux/LinuxProcess.c
```diff
@@ -168,9 +169,9 @@ bool LinuxProcess_isAutogroupEnabled(void) {
    return buf[0] == '1';
 }
 
-bool LinuxProcess_changeAutogroupPriorityBy(Process* this, Arg delta) {
+bool LinuxProcess_changeAutogroupPriorityBy(Row* super, Arg delta) {
    char buffer[256];
-   xSnprintf(buffer, sizeof(buffer), PROCDIR "/%d/autogroup", this->pid);
+   xSnprintf(buffer, sizeof(buffer), PROCDIR "/%d/autogroup", super->id);
```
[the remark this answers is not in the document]
natoscott: Yep, done.

linux/LinuxProcess.c
natoscott: Hmm, nak (for now at least) - I'd prefer it your suggested way, but it'll be inconsistent with so many other htop class definitions then.  I think a strong case could be made though - a single cleanup commit to do this would be better (separate PR to this one though).

linux/LinuxProcess.h
```diff
@@ -122,13 +122,13 @@ Process* LinuxProcess_new(const Machine* host);
...
-IOPriority LinuxProcess_updateIOPriority(LinuxProcess* this);
+IOPriority LinuxProcess_updateIOPriority(Row* super);
 
-bool LinuxProcess_setIOPriority(Process* this, Arg ioprio);
+bool LinuxProcess_setIOPriority(Row* super, Arg ioprio);
 
 bool LinuxProcess_isAutogroupEnabled(void);
 
-bool LinuxProcess_changeAutogroupPriorityBy(Process* this, Arg delta);
+bool LinuxProcess_changeAutogroupPriorityBy(Row* super, Arg delta);
```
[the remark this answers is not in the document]
natoscott: Yep, done, thanks.

openbsd/OpenBSDProcessList.c
```diff
@@ -34,17 +34,17 @@ in the source distribution for its full text.
 
 ProcessList* ProcessList_new(Machine* host, Hashtable* pidMatchList) {
    OpenBSDProcessList* this = xCalloc(1, sizeof(OpenBSDProcessList));
-   ProcessList* super = (ProcessList*) this;
+   Object_setClass(this, Class(ProcessList));
 
+   ProcessList* super = (ProcessList*) this;
```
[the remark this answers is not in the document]
natoscott: Done, thanks.

pcp/Instance.c
natoscott: Ack, thanks.

pcp/PCPProcessList.c
```diff
@@ -33,16 +33,17 @@ in the source distribution for its full text.
 
 ProcessList* ProcessList_new(Machine* host, Hashtable* pidMatchList) {
    PCPProcessList* this = xCalloc(1, sizeof(PCPProcessList));
-   ProcessList* super = &(this->super);
+   Object_setClass(this, Class(ProcessList));
 
+   ProcessList* super = &(this->super);
```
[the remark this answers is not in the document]
natoscott: Good catch, thanks.

Row.h
```diff
@@ -0,0 +1,218 @@
...
+   /* Platform specific fields, defined in ${platform}/ProcessField.h */
+   PLATFORM_PROCESS_FIELDS
+
+   /* Do not add new fields after this entry (dynamic entries follow) */
+   LAST_PROCESSFIELD
+} ProcessField;
+
+/* Extend ProcessField with dynamic entries - i.e. defined at runtime */
+#define ROW_DYNAMIC_FIELDS LAST_PROCESSFIELD
+typedef enum ProcessField_ RowField;
```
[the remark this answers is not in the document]
natoscott: This idea has worked quite well @BenBE - I'm much happier with it now, thanks.  I kept the type names as ProcessField and RowField though (instead of adding the 'Id' suffix) to reduce impact across the code base in all the functions and declarations using the existing type names.

DynamicScreen.h
```diff
@@ -0,0 +1,32 @@
...
+#define HEADER_DynamicScreen
+
+#include <stdbool.h>
+
+#include "Hashtable.h"
+#include "Panel.h"
+#include "Settings.h"
+
+
+typedef struct DynamicScreen_ {
```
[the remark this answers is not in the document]
natoscott: @BenBE I've audited Hashtable.c and don't see any requirement to use Objects - empirically it all works without em, based on our experience too.  Have you remembered more details on this aspect?

Affinity.c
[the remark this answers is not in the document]
natoscott: This is done now and worked quite nicely with the separation between Row/Process calls I think.

DynamicScreen.c
natoscott: Sure - I like 'screens' better I think, will go with that, thanks.

DynamicScreen.c
```diff
@@ -0,0 +1,64 @@
...
+   DynamicIterator iter = { .key = 0, .name = name, .found = false };
+   if (dynamics)
+      Hashtable_foreach(dynamics, DynamicScreen_compare, &iter);
+   if (key)
+      *key = iter.key;
+   return iter.found;
+}
+
+const char* DynamicScreen_lookup(Hashtable* dynamics, ht_key_t key) {
```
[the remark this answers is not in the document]
natoscott: Done.

MainPanel.c
natoscott: Hmm?  There are no systems with 64 bit PIDs, are there?

linux/LinuxProcess.c
```diff
@@ -143,22 +143,29 @@ static int LinuxProcess_effectiveIOPriority(const LinuxProcess* this) {
...
 
-bool LinuxProcess_setIOPriority(Process* this, Arg ioprio) {
+static bool LinuxProcess_setIOPriority(Process* p, Arg ioprio) {
 // Other OSes masquerading as Linux (NetBSD?) don't have this syscall
 #ifdef SYS_ioprio_set
-   syscall(SYS_ioprio_set, IOPRIO_WHO_PROCESS, this->pid, ioprio.i);
+   syscall(SYS_ioprio_set, IOPRIO_WHO_PROCESS, Process_getPid(p), ioprio.i);
 #endif
-   return (LinuxProcess_updateIOPriority((LinuxProcess*)this) == ioprio.i);
+   return (LinuxProcess_updateIOPriority(p) == ioprio.i);
```
BenBE, line 162:
```suggestion
   return LinuxProcess_updateIOPriority(p) == ioprio.i;
```

pcp/PCPDynamicScreen.c
```diff
@@ -0,0 +1,413 @@
...
+(C) 2022-2023 htop dev team
+Released under the GNU GPLv2+, see the COPYING file
+in the source distribution for its full text.
+*/
+
+#include "pcp/PCPDynamicScreen.h"
+
+#include <ctype.h>
+#include <dirent.h>
+#include <pcp/pmapi.h>
```
BenBE, line 13: Order? Or specific reason for this order? Comment if so.

ScreenTabsPanel.c
```diff
@@ -0,0 +1,420 @@
...
+(C) 2023 htop dev team
+Released under the GNU GPLv2+, see the COPYING file
+in the source distribution for its full text.
+*/
+
+#include "ScreenTabsPanel.h"
+
+#include <stdlib.h>
+#include <ctype.h>
+#include <string.h>
```
BenBE, line 12: Header order:
```suggestion
#include <ctype.h>
#include <stdlib.h>
#include <string.h>
```

Action.c
```diff
@@ -538,9 +563,15 @@ static Htop_Reaction actionLsof(State* st) {
...
 
 static Htop_Reaction actionShowLocks(State* st) {
+   if (!Action_readableProcess(st))
+      return HTOP_OK;
+
    const Process* p = (Process*) Panel_getSelected((Panel*)st->mainPanel);
    if (!p)
       return HTOP_OK;
+
+   assert(Object_isA((const Object*) p, (const ObjectClass*) &Process_class));
```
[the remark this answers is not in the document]
natoscott: Object_isA is not a macro, its a function with types ... so, nope, we need the casts.

Machine.c
natoscott: Fixed, thanks.

MainPanel.c
natoscott: OK, I'll go back to using pid_t here.

pcp/screens/filesys
natoscott: Sure, sounds better.

Scheduling.h
```diff
@@ -13,6 +13,7 @@ in the source distribution for its full text.
 #include <stdbool.h>
 
 #include "Panel.h"
+#include "Process.h"
```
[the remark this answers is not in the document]
natoscott: Scheduling_readProcessPolicy is passed a Process, so I think its correct as-is.

linux/LinuxProcess.c
natoscott: Yep, thanks.

Settings.h
natoscott: Ah, I see now - I'll revert it.

Machine.c
```diff
@@ -49,12 +56,71 @@ void Machine_done(Machine* this) {
...
+   Object_delete(this->processTable);
+   free(this->tables);
+}
+
+static void Machine_addTable(Machine* this, Table* table) {
+   /* check that this table has not been seen previously */
+   for (size_t i = 0; i < this->tableCount; i++) {
+      if (this->tables[i] == table)
+         return;
+   }
```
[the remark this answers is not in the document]
natoscott: OK, done, picked option 1.

ScreensPanel.c
```diff
@@ -43,23 +44,19 @@ ScreenListItem* ScreenListItem_new(const char* value, ScreenSettings* ss) {
 }
 
 static const char* const ScreensFunctions[] = {"      ", "Rename", "      ", "      ", "New   ", "      ", "MoveUp", "MoveDn", "Remove", "Done  ", NULL};
+static const char* const DynamicFunctions[] = {"      ", "Rename", "      ", "      ", "      ", "      ", "MoveUp", "MoveDn", "Remove", "Done  ", NULL};
 
 static void ScreensPanel_delete(Object* object) {
    Panel* super = (Panel*) object;
    ScreensPanel* this = (ScreensPanel*) object;
-
+
```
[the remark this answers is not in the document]
natoscott: Got it, thanks.

pcp/PCPDynamicScreen.c
natoscott: @BenBE not sure what you mean here - the order looks correct to me (alphabetical).

ScreenTabsPanel.c
natoscott: Yep, fixed, thanks.

ScreenTabsPanel.h
```diff
@@ -0,0 +1,55 @@
...
+/*
+htop - ScreenTabsPanel.h
+(C) 2023 htop dev team
+Released under the GNU GPLv2+, see the COPYING file
+in the source distribution for its full text.
+*/
+
+#include <stdbool.h>
+
+#include "ScreensPanel.h"
```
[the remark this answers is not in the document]
natoscott: Yep, done, thanks!

pcp/PCPDynamicScreen.c
BenBE: I think when I marked this, this block also contained some headers out of sort order. But this still applies for the block directly below. Just re-check for the pcp/ files for the header order.

natoscott: Ah yes, I see it now - thanks.

BenBE (approved):
LGTM.

Action.c
```diff
@@ -345,13 +356,27 @@ static Htop_Reaction actionExpandCollapseOrSortColumn(State* st) {
    return st->host->settings->ss->treeView ? actionExpandOrCollapse(st) : actionSetSortColumn(st);
 }
 
+static inline void setActiveScreen(Settings* settings, State* st, unsigned int ssIdx) {
+   assert(settings->ssIndex == ssIdx);
+   Machine* host = st->host;
+
+   settings->ss = settings->screens[ssIdx];
+   if (settings->ss->table == NULL)
```
BenBE, line 364:
```suggestion
   if (!settings->ss->table)
```

DynamicScreen.h
```diff
@@ -0,0 +1,33 @@
+#ifndef HEADER_DynamicScreen
+#define HEADER_DynamicScreen
+
```
BenBE, line 3: Copyright header?

ScreenTabsPanel.c
```diff
@@ -0,0 +1,372 @@
...
+      const ScreenSettings* ss = settings->screens[i];
+
+      if (ds == NULL) {
+         if (ss->dynamic != NULL)
+            continue;
+         /* built-in (processes, not dynamic) - e.g. Main or I/O */
+      } else {
+         if (ss->dynamic == NULL)
+            continue;
+         if (String_eq(ds->name, ss->dynamic) == false)
```
BenBE, line 46:
```suggestion
         if (!String_eq(ds->name, ss->dynamic))
```

ScreenTabsPanel.c
```diff
@@ -0,0 +1,372 @@
...
+
+static HandlerResult ScreenNamesPanel_eventHandler(Panel* super, int ch) {
+   ScreenNamesPanel* const this = (ScreenNamesPanel*) super;
+
+   if (!this->renamingItem)
+      return ScreenNamesPanel_eventHandlerNormal(super, ch);
+   return ScreenNamesPanel_eventHandlerRenaming(super, ch);
+}
+
+PanelClass ScreenNamesPanel_class = {
```
BenBE, line 341: Maybe collect all classes at the end of the file?

pcp/PCPDynamicScreen.h
```diff
@@ -0,0 +1,50 @@
+#ifndef HEADER_PCPDynamicScreen
+#define HEADER_PCPDynamicScreen
+
```
BenBE, line 3: Copyright header?

Action.c
natoscott: Fixed, thanks

DynamicScreen.h
natoscott: Added, thanks

ScreenTabsPanel.c
natoscott: Sure - updated now, thanks

ScreenTabsPanel.c
natoscott: Hmm, I think I'll leave as-is cos this pattern occurs a few times (keeping Panel sub-class definitions in particular close to the code for simple panels) across existing htop code.

pcp/PCPDynamicScreen.h
natoscott: Added, thanks!

natoscott:
This is merged - if there's any platform specific build issues (Linux and PCP only verified so far), please let me know.

Explorer09:
Ah. I missed the chance to review this code a little bit. There are a few styling issues that I'm not sure I like it. Would you mind if I make pull requests as follows ups of this one?

Here are two issues I can see at the moment:
1. The `printBytes`, `printCount`, `printPercentage`, `printTime` etc. functions should reside in the `RowField` class, not on the `Row` class, as these functions are more for printing fields and not for printing a row.
2. Would you mind if I rename the class `RowField` to just `Field` (`RowField.h` to `Field.h`)? Because I still believe the name `RowField` is silly. Also that enum type should be `FieldID`. I expect that `Field` refer to a class and not an enum or an identifier. How do you think?

natoscott:
@Explorer09 personally, I don't think those changes will be helpful - the reason for the RowField naming is to match up with the ProcessField naming, 'Field' is maybe a little too vague IMO.

Explorer09:
@natoscott I know the `ProcessField` name. And I think naming the new enum type `RowField` is no longer intuitive now. I was saying that the enum should be named `FieldID` so that the name `Field` could refer to a class or a module on its own.

`ProcessField` will inherit `FieldID` (becoming the same integer/enum type). There is no need to change the `ProcessField` name for now.

PR #1244, opened 2023-05-12
Add NetworkIOMeter support for Darwin

UeiWang:
Relevant issue: #1138

### DIFF
--- a/darwin/Platform.c
+++ b/darwin/Platform.c
@@ -14,6 +14,10 @@ in the source distribution for its full text.
 #include <math.h>
 #include <stdlib.h>
 #include <unistd.h>
+#include <net/if.h>
+#include <net/if_types.h>
+#include <net/route.h>
+#include <sys/socket.h>
 #include <sys/_types/_mach_port_t.h>
 
 #include <CoreFoundation/CFBase.h>
@@ -138,6 +142,7 @@ const MeterClass* const Platform_meterTypes[] = {
    &ZfsArcMeter_class,
    &ZfsCompressedArcMeter_class,
    &DiskIOMeter_class,
+   &NetworkIOMeter_class,
    &FileDescriptorMeter_class,
    &BlankMeter_class,
    NULL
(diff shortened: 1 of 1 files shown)


darwin/Platform.c
```diff
@@ -464,10 +469,58 @@ bool Platform_getDiskIO(DiskIOData* data) {
...
+   }
+
+   unsigned long long int bytesReceived_sum = 0, packetsReceived_sum = 0, bytesTransmitted_sum = 0, packetsTransmitted_sum = 0;
+
+   for (char *next = buf; next < buf + len;) {
+      IGNORE_WCASTALIGN_BEGIN
+      struct if_msghdr *ifm = (struct if_msghdr*) next;
+      IGNORE_WCASTALIGN_END
+
+      next += ifm->ifm_msglen;
```
BenBE, line 502: You can avoid the cast alignment warning by first casting to `void*` in an intermediate step.

darwin/Platform.c
```diff
@@ -464,10 +469,58 @@ bool Platform_getDiskIO(DiskIOData* data) {
...
+   /* Determine len */
+   if (sysctl(mib, ARRAYSIZE(mib), NULL, &len, NULL, 0) < 0)
+      return false;
+
+   char *buf = xMalloc(len);
+
+   if (sysctl(mib, ARRAYSIZE(mib), buf, &len, NULL, 0) < 0) {
+      free(buf);
+      return false;
+   }
```
BenBE, line 493: As the allocated buffer may be too small due to a race condition, try to allocate enough memory in a loop (with some slack); and only fail after 3 or so attempts to get the data.

Similar to what is done for listing processes/LWP …

BenBE (changes requested):

darwin/Platform.c
Explorer09: 1. I wonder, is there an API that allocates the buffer and returns the `sysctl` information in one atomic step? Because it looks like we could not avoid the rece condition otherwise.
2. Is it possible that we reuse the `buf` buffer and avoid the repeated `xMalloc` and `free` routines? I mean, track the allocated buffer and it's size in global variables, and use `xRealloc` when the buffer needs expanding. I don't know if htop had used this practice before.

BenBE: > 1. I wonder, is there an API that allocates the buffer and returns the `sysctl` information in one atomic step? Because it looks like we could not avoid the race condition otherwise.

We can't fully avoid this if the interface to query the information does not provide the required size directly. Additionallythis size can change at any time, thus insufficient buffer space is not a bug in itself. Thus the code should be able to cope with the situation and increase the buffer if necessary; but break out of the loop, if repeated attempts fail. As mentioned: That's what the process information code (e.g. on Darwin) does.

> 2. Is it possible that we reuse the `buf` buffer and avoid the repeated `xMalloc` and `free` routines? I mean, track the allocated buffer and it's size in global variables, and use `xRealloc` when the buffer needs expanding. I don't know if htop had used this practice before.

That's what the `Platform_init` and `Platform_done` functions are for. Also the code should just re-allocate the buffer as needed for subsequent requests.

BenBE (changes requested):

darwin/Platform.c
```diff
@@ -15,6 +15,10 @@ in the source distribution for its full text.
 #include <stdlib.h>
 #include <unistd.h>
 #include <sys/_types/_mach_port_t.h>
+#include <sys/socket.h>
+#include <net/if.h>
+#include <net/if_types.h>
+#include <net/route.h>
```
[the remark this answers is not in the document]
UeiWang: What do you suggest?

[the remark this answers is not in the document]
BenBE: Have a look at the style guide. Alphabetical, parent directory first, system headers before project headers (exception: associated file header).

darwin/Platform.c
```diff
@@ -464,10 +469,65 @@ bool Platform_getDiskIO(DiskIOData* data) {
...
-   (void)data;
-   return false;
+   int mib[6] = {CTL_NET,
+      PF_ROUTE, /* routing messages */
+      0, /* protocal number, currently always 0 */
+      0, /* select all address families */
+      NET_RT_IFLIST2, /* interface list with addresses */
+      0};
+
+   for (unsigned int retry = 0; retry < 4; retry++) {
```
BenBE, line 483:
```suggestion
   for (size_t retry = 0; retry < 4; retry++) {
```

darwin/Platform.c
```diff
@@ -464,10 +469,65 @@ bool Platform_getDiskIO(DiskIOData* data) {
...
+      size_t len;
```
BenBE, line 484:
```suggestion
      size_t len = 0;
```

Avoid uninitialized stack variables …

darwin/Platform.c
```diff
@@ -464,10 +469,65 @@ bool Platform_getDiskIO(DiskIOData* data) {
...
+
+      if (sysctl(mib, ARRAYSIZE(mib), buf, &len, NULL, 0) < 0) {
+         free(buf);
+         if (errno == ENOMEM && retry < 3)
+            continue;
+         else
+            return false;
+      }
+
+      unsigned long long int bytesReceived_sum = 0, packetsReceived_sum = 0, bytesTransmitted_sum = 0, packetsTransmitted_sum = 0;
```
BenBE, line 501:
```suggestion
      uint64_t bytesReceived_sum = 0, packetsReceived_sum = 0, bytesTransmitted_sum = 0, packetsTransmitted_sum = 0;
```

BenBE (changes requested):

darwin/Platform.c
```diff
@@ -464,10 +469,65 @@ bool Platform_getDiskIO(DiskIOData* data) {
...
+      0};
+
+   for (unsigned int retry = 0; retry < 4; retry++) {
+      size_t len;
+
+      /* Determine len */
+      if (sysctl(mib, ARRAYSIZE(mib), NULL, &len, NULL, 0) < 0 || len == 0)
+         return false;
+
+      len += 16 * retry * retry * sizeof(struct if_msghdr2);
```
Explorer09, line 490: What does this `16 * retry * retry` come from?

BenBE: That's for exponential back-off to over-allocate in case of rapid fluctuations. Not strictly necessary for network stuff, but doesn't hurt either.

darwin/Platform.c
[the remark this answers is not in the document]
UeiWang: OK. I didn't know alphabetical order is required since this requirement is missing from the style guide.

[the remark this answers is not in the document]
BenBE: I'm certain it's in the style guide. But given some systems have broken headers, it's worded less strictly (alphabetical, unless required otherwise, but add comment in that case).

BenBE (approved):

PR #1246, opened 2023-05-15
Minor style guide clarifications

BenBE:
This PR contains a few minor updates to our style guide used when contributing code to htop.

Apart from a few minor changes in wording, the most significant update is the addition of two items to take care of when writing a PR:

1. Your patchset should have descriptive commit subjects and be minimal
2. Follow up with your PRs and take care of change requests from reviews

Both items are nothing new (both were in there before); they just have been amended.

Also, the second item absolutely means PR authors can (and should) rebase their commits to incorporate changes.
If feedback from our side is missing, a quick note suffices.

NB: Even if it may seem daunting to receive lots of comments from a review, we are always open for discussion of the raised points and suggestions like alternative approaches.

Comments and further additions for this style guide update welcome.
If you are missing something in this update, drop a note by 2023-05-22 00:00 UTC.

### DIFF
--- a/docs/styleguide.md
+++ b/docs/styleguide.md
@@ -8,12 +8,12 @@ Names are important to convey what all those things inside the project are for.
 Filenames for source code traditionally used camel-case naming with the first letter written in uppercase.
 The file extension is always lowercase.
 
-The only exception here is `htop.c` which is the main entrance point into the code.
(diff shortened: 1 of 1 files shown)


natoscott (approved):
@BenBE looks good - all makes sense to me

PR #1249, opened 2023-05-21
Linux/LibSensors: adjust data for k10temp

cgzones:
Adjust the temperature data gathered from the k10temp driver. 
Only a control value, an optional die value, and one value per CCD are provided.

See https://www.kernel.org/doc/html/latest/hwmon/k10temp.html for details.

### DIFF
--- a/linux/LibSensors.c
+++ b/linux/LibSensors.c
@@ -153,7 +153,7 @@ void LibSensors_getCPUTemperatures(CPUData* cpus, unsigned int existingCPUs, uns
    assert(existingCPUs > 0 && existingCPUs < 16384);
 
    double* data = xMallocArray(existingCPUs + 1, sizeof(double));
-   for (size_t i = 0; i < existingCPUs + 1; i++)
+   for (size_t i = 0; i <= existingCPUs; i++)
       data[i] = NAN;
 
 #ifndef BUILD_STATIC
--- a/linux/LibSensors.c
+++ b/linux/LibSensors.c
@@ -29,12 +29,12 @@ in the source distribution for its full text.
 
 #ifdef BUILD_STATIC
 
-#define sym_sensors_init sensors_init
-#define sym_sensors_cleanup sensors_cleanup
+#define sym_sensors_init               sensors_init
+#define sym_sensors_cleanup            sensors_cleanup
 #define sym_sensors_get_detected_chips sensors_get_detected_chips
-#define sym_sensors_get_features sensors_get_features
-#define sym_sensors_get_subfeature sensors_get_subfeature
-#define sym_sensors_get_value sensors_get_value
+#define sym_sensors_get_features       sensors_get_features
+#define sym_sensors_get_subfeature     sensors_get_subfeature
+#define sym_sensors_get_value          sensors_get_value
 
 #else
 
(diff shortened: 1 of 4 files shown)


linux/LibSensors.c
```diff
@@ -120,26 +120,29 @@ int LibSensors_reload(void) {
    return sym_sensors_init(NULL);
 }
 
-static int tempDriverPriority(const sensors_chip_name* chip) {
-   static const struct TempDriverDefs {
+enum TempDriver {
```
BenBE, line 123: Please create a proper typedef for this enum.

BenBE (commented):
I'm all for dropping the first (align macro definition) commit.

BenBE:
Re the `double`->`float` patch: While we are not showing anything more than 1 fractional digit, the data acquisition is done using doubles, thus sticking to keeping the native type here is IMO preferable (avoids rounding errors due to type conversion).

linux/LibSensors.c
```diff
@@ -120,26 +120,29 @@ int LibSensors_reload(void) {
...
-}
+      TempDriver driver;
+} tempDrivers[] =  {
+   { "coretemp",    0, TD_CORETEMP },
+   { "via_cputemp", 0, TD_CPUTEMP },
+   { "cpu_thermal", 0, TD_CPUTHERMAL },
+   { "k10temp",     0, TD_K10TEMP },
+   { "zenpower",    0, TD_ZENPOWER },
+   /* Low priority drivers */
+   { "acpitz",      1, TD_ACPITZ },
```
Explorer09, line 144: Is there a possibility that a temperature driver comes with two names?
Because I wonder if the table can be simplified to this:

```c
   [TD_CORETEMP] = { "coretemp", 0 },
   [TD_CPUTEMP] = { "via_cputemp", 0 },
   [TD_CPUTHERMAL] = { "cpu_thermal", 0 },
   [TD_K10TEMP] = { "k10temp", 0 },
   [TD_ZENPOWER] = { "zenpower", 0 },
   /* Low priority drivers */
   [TD_ACPITZ] = { "acpitz", 1 },
```

linux/LinuxMachine.c
```diff
@@ -56,13 +56,15 @@ static void LinuxMachine_updateCPUcount(LinuxMachine* this) {
...
       super->existingCPUs = 1;
+      this->maxCoreId = 0;
    }
 
    DIR* dir = opendir("/sys/devices/system/cpu");
    if (!dir)
       return;
 
    unsigned int currExisting = super->existingCPUs;
+   unsigned short maxCoreId = 0;
```
Explorer09, line 67: Why do we bother with the `unsigned short` data type here (rather than `unsigned int`)? There is no space constraint here (it's not an array) and `short` data types typically process slower in most CPU architectures.

cgzones: To save two bytes in the `CPUData` struct.

Explorer09: @cgzones Not worth it when your other code becomes larger due to the instructions processing 16-bit registers. (At least in IA-32 & x86-64, instructions processing 16-bit registers need a prefix and each becomes one byte larger than 32-bit counterparts.)

fasterit:
@cgzones: does #1352 fix the k10temp use case as well? I.e. can this be closed as resolved without a merge?

PR #1250, opened 2023-05-23
Document 'O' keyboard shortcut

### LINKED ISSUE #1236: Keyboard shortcut to hide containerized processes not documented
The `O` keyboard shortcut is not documented in the `F1` help. Could this also be added?
Same for the man page.

_Originally posted by @m0vie in https://github.com/htop-dev/htop/issues/1062#issuecomment-1537134535_

valdaarhun:
Closes #1236.

Document the 'O' keyboard shortcut in the man page and `F1` help screen.

Reported by @m0vie.

### DIFF
--- a/Action.c
+++ b/Action.c
@@ -606,6 +606,7 @@ static const struct {
    { .key = "      u: ",  .roInactive = false, .info = "show processes of a single user" },
    { .key = "      H: ",  .roInactive = false, .info = "hide/show user process threads" },
    { .key = "      K: ",  .roInactive = false, .info = "hide/show kernel threads" },
+   { .key = "      O: ",  .roInactive = false, .info = "hide/show processes in containers" },
    { .key = "      F: ",  .roInactive = false, .info = "cursor follows process" },
    { .key = "  + - *: ",  .roInactive = false, .info = "expand/collapse tree/toggle all" },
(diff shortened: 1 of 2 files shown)


valdaarhun:
> Where are the changes to the offset for the right column coming from?

The `-es` in "processes" is replaced with the characters in the right column when using the original offset. Increasing the offset by two prevents that.

Action.c
```diff
@@ -606,6 +606,7 @@ static const struct {
    { .key = "      u: ",  .roInactive = false, .info = "show processes of a single user" },
    { .key = "      H: ",  .roInactive = false, .info = "hide/show user process threads" },
    { .key = "      K: ",  .roInactive = false, .info = "hide/show kernel threads" },
+   { .key = "      O: ",  .roInactive = false, .info = "hide/show containerized processes" },
```
fasterit, line 609: How about "hide/show processes in containers" ?
Same length, seems more understandable / correct English.

fasterit:
> 



> > Where are the changes to the offset for the right column coming from?
> 
> The `-es` in "processes" is replaced with the characters in the right column when using the original offset. Increasing the offset by two prevents that.

We're at 88 columns width now, this change makes it 90.
Somebody did not pay attention that we stay within 80 columns, so ... seems 90 is fine now as well.
We just should make sure it does not continue to grow beyond that...

Action.c
valdaarhun: Looks good to me.

PR #208, opened 2020-10-05
Update license and copyright info

fasterit:

PR #1251, opened 2023-05-24
Style/header consistency / copyright notes

BenBE:
@cgzones @fasterit @ginggs @hishamhm @natoscott Please check completeness of the copyright statements.

Also I'll take another look at other places where the include order is inconsistent with the style guide.


Explorer09:
In ProvideCurses.h and ProvideTerm.h, the header file name as mentioned in the comments are wrong.
This is just what I've found for now.

natoscott (approved):
LGTM.

hishamhm:
Ah, nice — the names and date ranges are based on the git history, right?

I haven't kept track of the copyright dates for the other files — a quick check of a random file, Vector.c, confirms it; it still says `2004-2011` even if there were later commits of mine into 2018 plus a bunch of more recent ones from you folks.

Also (and sorry if I open a can of worms here): I have no memory if it ever being GPL v2+, I was pretty sure I had licensed it GPL v2 only; searching for `later` in older tags such as 2.2.0, I don't see it being v2+ mentioned anywhere (other than `If the Program specifies a version number of this License which applies to it and "any later version" [...]`, which was not the case). It did say "GNU GPL" only in the files themselves, but the "COPYING" file was for version 2, and the README stated "GNU General Public License version 2" only).

Back in the day I made a conscious decision to keep it at version 2 only, like the Linux kernel, because (a) that kept it anchored to a known license, not subject to unforeseen changes in FSF leadership and/or behavior, (b) htop's typical use-cases are not subject to tivoization, which was the main benefit of GPLv3.

fasterit:
You're about 2-3 years late for the license discussion ... cf. 079c2abf, 976fa3b1, 94ad1113, 466d6f99 and 469ae7a0.

BenBE:
> Ah, nice — the names and date ranges are based on the git history, right?

I created the copyright headers based on the commit history from git, filtering out any commits that could be marked as trivial. I did not update any headers for files which already had them in place.

> I haven't kept track of the copyright dates for the other files — a quick check of a random file, Vector.c, confirms it; it still says `2004-2011` even if there were later commits of mine into 2018 plus a bunch of more recent ones from you folks.

I can update those too; although this wasn't the initial goal of this PR. But if needed I can take the time to go through all of them and update as necessary (filtering trivial edits as necessary).

> Also (and sorry if I open a can of worms here): I have no memory if it ever being GPL v2+, I was pretty sure I had licensed it GPL v2 only; searching for `later` in older tags such as 2.2.0, I don't see it being v2+ mentioned anywhere (other than `If the Program specifies a version number of this License which applies to it and "any later version" [...]`, which was not the case). It did say "GNU GPL" only in the files themselves, but the "COPYING" file was for version 2, and the README stated "GNU General Public License version 2" only).
> 
> Back in the day I made a conscious decision to keep it at version 2 only, like the Linux kernel, because (a) that kept it anchored to a known license, not subject to unforeseen changes in FSF leadership and/or behavior, (b) htop's typical use-cases are not subject to tivoization, which was the main benefit of GPLv3.

As @fasterit referenced, and furthermore #208, #808, #814. AFAIR the issue at hand was the text of the license file matching the GPLv2+ variant although the code stated GPLv2 everywhere.

FWIW, for personal projects I started to go all-out AGPLv3 and I can quite well see, why this does not bring many advantages for htop (although depending on the lawyer you ask you could make a point of displaying htop over SSH constituting "networked usage" ;-)). As for tivoization: Sure the risk is negligible for htop, but regardless having that protection doesn't hurt. Remains the issue re FSF leadership. Unlikely, but valid concern.

Let's ask differently:
- Do you have hard objections to using GPLv2 **or later** for the portions based on your code?
- Do you allow us to use the code with GPLv2+ as its license, i.e. including the "or any later" term?

NB: A "no" for the second question will get VERY messy … ;-)

hishamhm:
> But if needed I can take the time to go through all of them and update as necessary (filtering trivial edits as necessary).

I don't think it's necessary (but IANAL of course), but in any case it's not in scope for this PR.

> AFAIR the issue at hand was the text of the license file matching the GPLv2+ variant although the code stated GPLv2 everywhere.

What's in COPYING is the standard text for GPLv2; I don't think there are different versions of COPYING for GPLv2 and GPLv2+. The text does state:

```
Each version is given a distinguishing version number.  If the Program
specifies a version number of this License which applies to it and "any
later version", you have the option of following the terms and conditions
either of that version or of any later version published by the Free
Software Foundation.  If the Program does not specify a version number of
this License, you may choose any version ever published by the Free Software
Foundation.
```

...but that's a conditional ("If the program..."), which as you said, wasn't triggered anywhere else in the codebase or documentation, so the "or later" note does not apply, and the README did say "version 2" specifically. Note that the [GPL-2 text in the Linux Kernel](https://github.com/torvalds/linux/blob/master/LICENSES/preferred/GPL-2.0) source code (which notably does not apply "or later") contains that same paragraph.

So, I think the license change here happened as a result of believing there was an inconsistency in the licensing, but there wasn't any.

> Remains the issue re FSF leadership. Unlikely, but valid concern.

When I started htop back in 2004 I also found that to be unlikely. But some missteps by the FSF in the years since have moved the needle of the unlikeliness-meter in my perception a bit.

> NB: A "no" for the second question will get VERY messy … ;-)

Would it, really? You folks changed the license once, even without the explicit go-ahead from the previous contributors (and I wasn't the only copyright holder either, as seen in `git checkout 2.2.0 | grep -r '(C)' *`).

To make it clear, I hope I don't come across as annoyed by the situation — I don't see any ill-will, of course: the messages in this thread and the older commits all indicate that the changes were out of a desire for consistency, not to explicitly change the license into something else for, say, adding GPL3 compatibility or any other reason. I see it an honest mistake, given the inconsistency wasn't really there, and one I hope is just as easy to revert. If the current code is GPL "version 2, or at your choice, any later version", why can't the next version be, at our choice, version 2 again? These links seem to indicate it's fine to do so:

* (explanation summary) https://opensource.stackexchange.com/a/7996
* (primary source) https://www.gnu.org/licenses/gpl-faq.en.html#AllCompatibility

For past releases in the 3.x series we can let it go as water under the bridge, but I'd really like my license choice to be respected in the code that is my legacy to the project.

Issue #1248, opened 2023-05-20
Empty header in configuration not understood


mizlan:
I would like to have no header entries. Using <kbd>F2</kbd> (setup), I deleted every entry under my two columns in `Meters`, which displays live properly, and when I exit with <kbd>F10</kbd>, the following lines are set in my `~/.config/htop/htoprc`:

```
column_meters_0=
column_meter_modes_0=
column_meters_1=
column_meter_modes_1=
```

However, even though `htop` writes these lines to its configuration file, I don't believe it knows how to parse this, because when I reopen `htop`, the header is back. I think this is a bug.

BenBE:
Cf. #1227 #168 #445 for previous discussion of this.

PR #1253, opened 2023-06-01
Settings: preserve empty header

cgzones:
If the user removes all meters from the header on exit htop will write no values for the settings column_meters_X and column_meter_modes_X. The parser skips all configuration settings with no value, so on the next start no header meter related key is processed and htop will add a set of default meters to the header.

Write instead an invalid value of `!`, so the keys are parsed and htop will not add the default meters back.

Closes: #1248

### DIFF
--- a/Settings.c
+++ b/Settings.c
@@ -588,15 +588,25 @@ static void writeList(FILE* fd, char** list, int len, char separator) {
 }
 
 static void writeMeters(const Settings* this, FILE* fd, char separator, unsigned int column) {
-   writeList(fd, this->hColumns[column].names, this->hColumns[column].len, separator);
+   if (this->hColumns[column].len) {
+      writeList(fd, this->hColumns[column].names, this->hColumns[column].len, separator);
+   } else {
+      fputc('!', fd);
+      fputc(separator, fd);
+   }
 }
 
 static void writeMeterModes(const Settings* this, FILE* fd, char separator, unsigned int column) {
-   const char* sep = "";
-   for (size_t i = 0; i < this->hColumns[column].len; i++) {
-      fprintf(fd, "%s%d", sep, this->hColumns[column].modes[i]);
-      sep = " ";
+   if (this->hColumns[column].len) {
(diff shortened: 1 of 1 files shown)


BenBE (approved):
LGTM.

Explorer09:
This looks like a dirty hack to me. Although you let `writeMeters` and `writeMeterModes` output an `!` mark when no meters are used, I think you should also update the `Settings_readMeterModes` part to special-case the `!` mark.

The changed code would look something like this:

```c
static void Settings_readMeterModes(Settings* this, const char* line, unsigned int column) {
   char* trim = String_trim(line);
   char** ids = String_split(trim, ' ', NULL);
   free(trim);
   int len = 0;
   for (int i = 0; ids[i]; i++) {
      len++;
   }
   column = MINIMUM(column, HeaderLayout_getColumns(this->hLayout) - 1);
   int* modes = len ? xCalloc(len, sizeof(int)) : NULL;
   for (int i = 0; i < len; i++) {
      if (String_eq(ids[i], "!")) {
         modes[i] = 0;
         len = i + 1;
         break;
      }
      modes[i] = atoi(ids[i]);
   }
   this->hColumns[column].len = len;
   String_freeArray(ids);
   this->hColumns[column].modes = modes;
}
```

Settings.c
```diff
@@ -544,15 +544,25 @@ static void writeList(FILE* fd, char** list, int len, char separator) {
 }
 
 static void writeMeters(const Settings* this, FILE* fd, char separator, unsigned int column) {
-   writeList(fd, this->hColumns[column].names, this->hColumns[column].len, separator);
+   if (this->hColumns[column].len)
+      writeList(fd, this->hColumns[column].names, this->hColumns[column].len, separator);
+   else {
```
Explorer09, line 549: Braces.
```c
   if (this->hColumns[column].len) {
      writeList(fd, this->hColumns[column].names, this->hColumns[column].len, separator);
   } else {
   ...
   }
```

PR #1255, opened 2023-06-03
Update process list on thread visibility toggling

### LINKED ISSUE #1254: CPU usage of some processes incorrectly reported as using 100% of all cores until refresh after pressing Shift+H
It seems like after pressing shift+h, a bunch of processes show up as using 1600% CPU (my system has 16 cores) before waiting the two seconds to refresh

I'd expect after shift+h, the CPU usage of processes is updated and correct without waiting for a refresh

System: Fully updated Arch Linux as of 03 June 2023, hyprland + alacritty
Htop: b810678a8dcb708b452864f5991f007920161bdb

Thank you for your time :), and of course reach out if there's more info that would be helpful. I'm happy to bisect the issue as well

https://github.com/htop-dev/htop/assets/7105082/44f455de-2888-4e12-a8aa-e03b6289dbd8

fasterit:
Fixes #1254

### DIFF
--- a/Action.c
+++ b/Action.c
@@ -220,6 +220,7 @@ static Htop_Reaction actionToggleKernelThreads(State* st) {
    Settings* settings = st->host->settings;
    settings->hideKernelThreads = !settings->hideKernelThreads;
    settings->lastUpdate++;
+   ProcessList_scan(st->host->pl); // needed to not have a visible delay showing wrong data
 
    return HTOP_RECALCULATE | HTOP_SAVE_SETTINGS | HTOP_KEEP_FOLLOWING;
 }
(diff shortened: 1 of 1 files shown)


BenBE:
Shouldn't it suffice to return the need for re-scanning in those two places?

fasterit:
Well, there is still the delay before the next redraw if we set `HTOP_REFRESH`. If you have a better idea, pls. update this PR or make a better one!

BenBE:
I just had something in mind regarding a similar fix recently in https://github.com/htop-dev/htop/commit/f50944c3d563238c9843057978041f2cd7c3b285

PR #1257, opened 2023-06-05
Fix Linux running tasks count in the Tasks Meter, a recent regression

### LINKED ISSUE #1256: htop tip shows "0 running" processes in the task counter header
Tasks: 151, 1497 thr, 329 kthr; **0 running**
... regardless of how many processes are in **R** state

natoscott:
The recent split of machine-wide metric gathering from process metrics gathering introduced a regression in the Linux running task accounting. The way we extract this value from /proc/stat (which typically is used for machine-wide metrics) conspired with a now-incorrectly placed init- to-zero at the start of the per-process sampling, caused this value to always be zero by the time the Tasks Meter used it.

Fix this by making it much more clear that Linux has this special case handling of runningTasks.

Resolves #1256

### DIFF
--- a/linux/LinuxMachine.c
+++ b/linux/LinuxMachine.c
@@ -501,7 +501,7 @@ static void LinuxMachine_scanCPUTime(LinuxMachine* this) {
    char buffer[PROC_LINE_LENGTH + 1];
    while (fgets(buffer, sizeof(buffer), file)) {
       if (String_startsWith(buffer, "procs_running")) {
-         super->pl->runningTasks = strtoul(buffer + strlen("procs_running"), NULL, 10);
+         this->runningTasks = strtoul(buffer + strlen("procs_running"), NULL, 10);
          break;
       }
    }
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -1312,6 +1312,9 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
    const ScreenSettings* ss = settings->ss;
    const struct dirent* entry;
 
+   /* set runningTasks from /proc/stat (from Machine_scanCPUTime) */
+   pl->runningTasks = lhost->runningTasks;
+
 #ifdef HAVE_OPENAT
    int dirFd = openat(parentFd, dirname, O_RDONLY | O_DIRECTORY | O_NOFOLLOW);
    if (dirFd < 0)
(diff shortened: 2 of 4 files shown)


linux/LinuxProcessList.c
```diff
@@ -1312,6 +1312,9 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, openat_arg_
    const ScreenSettings* ss = settings->ss;
    const struct dirent* entry;
 
+   /* set runningTasks from /proc/stat (from Machine_scanCPUTime) */
+   pl->runningTasks = lhost->running;
+
```
BenBE, line 1317: With `ProcessList_goThroughEntries` being the entry point calling `LinuxProcessList_recurseProcTree` it would be better to place this initialization there, thus executing it only once (`PL_gTE` recurses, thus executing this code potentially multiple times; doesn't hurt, but isn't necessary either).

linux/LinuxMachine.c
```diff
@@ -501,7 +501,7 @@ static void LinuxMachine_scanCPUTime(LinuxMachine* this) {
    char buffer[PROC_LINE_LENGTH + 1];
    while (fgets(buffer, sizeof(buffer), file)) {
       if (String_startsWith(buffer, "procs_running")) {
-         super->pl->runningTasks = strtoul(buffer + strlen("procs_running"), NULL, 10);
+         this->running = strtoul(buffer + strlen("procs_running"), NULL, 10);
```
BenBE, line 504: `running` is too generic; maybe stick with `runningTasks`?

BenBE (commented):
Did you check this only affects Linux?

Darwin has the init-to-zero in `PL_gTE` too, which can be removed given it's already globally in `PL_scan`.

natoscott:
> Did you check this only affects Linux?

Yes, its definitely only Linux affected.

> Darwin has the init-to-zero in `PL_gTE` too, which can be removed given it's already globally in `PL_scan`.

Yep - I've added a followup cleanup commit for that.

linux/LinuxMachine.c
natoscott: Done.

linux/LinuxProcessList.c
natoscott: Hmm, maybe - I did prefer this location because it keeps the two matched comments about runningTasks in that same routine (the second is towards the end of the loop, where the runningTasks++ used to be, back in the day).

I don't think its a problem to execute an integer assignment multiple times, especially given everything else going on in that routine - better to keep the code intent clear and understandable IMO.

BenBE (approved):

PR #1263, opened 2023-06-19
default color preset: use bold blue for better visibility

eworm-de:
This one for compressed zram was missing in
2ca75625ee5c2ac0ef1571e6918d7c94f3aa011c... Let's add it.


BenBE:
As far as I can see, this is only ever used for the meter, but not any numbers when printing (unless I'm missing something on first glance). For the normal meters the usual dark blue on black is fine, I guess …

eworm-de:
While this is more important for numbers I still think the visibility of meters can benefit here. Anyway, the best argument is consistency - all other meters do use bold blue instead of plain blue now.

PR #1271, opened 2023-07-26
Replace isnan() with better comparisons (isgreater(), etc.)

Explorer09:
The standard `isnan()` function is defined to never throw FP exceptions even when the argument is a "signaling" NaN. This makes `isnan()` more expensive than `(x != x)` expression unless the compiler flag `-fno-signaling-nans` is given.

Introduce functions `isNaN()`, `isNonnegative()`, `isPositive()`, `sumPositiveValues()` and `compareRealNumbers()`, and replace `isnan()` in htop's codebase with the new functions. These functions utilize `isgreater()` and `isgreaterequal()` comparisons, which do not throw FP exceptions on "quiet" NaNs, which htop uses extensively.

With `isnan()` removed, there is no need to suppress the warning `-Wno-c11-extensions` in FreeBSD. Remove the code from `configure.ac`.

### DIFF
--- a/Macros.h
+++ b/Macros.h
@@ -2,6 +2,8 @@
 #define HEADER_Macros
 
 #include <assert.h> // IWYU pragma: keep
+#include <math.h>
+#include <stdbool.h>
 
 #ifndef MINIMUM
 #define MINIMUM(a, b)                  ((a) < (b) ? (a) : (b))
@@ -97,6 +99,24 @@
 #define IGNORE_WCASTQUAL_BEGIN
 #define IGNORE_WCASTQUAL_END
 #endif
+
+/* Cheaper function for checking NaNs. Unlike the standard isnan(), this may
+   throw an FP exception on a "signaling" NaN.
+   (ISO/IEC TS 18661-1 and the C23 standard stated that isnan() throws no
+   exceptions even with a "signaling" NaN) */
+static inline bool isNaN(double x) {
+   return !isgreaterequal(x, x);
+}
+
+/* Checks if x is nonnegative. Returns false if x is NaN. */
+static inline bool isNonnegative(double x) {
+   return isgreaterequal(x, 0.0);
+}
+
+/* Checks if x is positive. Returns false if x is NaN. */
+static inline bool isPositive(double x) {
+   return isgreater(x, 0.0);
+}
 
 /* This subtraction is used by Linux / NetBSD / OpenBSD for calculation of CPU usage items. */
 static inline unsigned long long saturatingSub(unsigned long long a, unsigned long long b) {
--- a/Process.c
+++ b/Process.c
@@ -791,7 +791,7 @@ void Process_printLeftAlignedField(RichString* str, int attr, const char* conten
    RichString_appendChr(str, attr, ' ', width + 1 - columns);
 }
 
-void Process_printPercentage(float val, char* buffer, int n, uint8_t width, int* attr) {
+void Process_printPercentage(float val, char* buffer, size_t n, uint8_t width, int* attr) {
    if (val >= 0) {
       if (val < 0.05F)
          *attr = CRT_colors[PROCESS_SHADOW];
--- a/Process.c
+++ b/Process.c
@@ -1184,7 +1184,7 @@ int Process_compareByKey_Base(const Process* p1, const Process* p2, ProcessField
    switch (key) {
    case PERCENT_CPU:
    case PERCENT_NORM_CPU:
-      return SPACESHIP_NUMBER(p1->percent_cpu, p2->percent_cpu);
+      return compareRealNumbers(p1->percent_cpu, p2->percent_cpu);
    case PERCENT_MEM:
       return SPACESHIP_NUMBER(p1->m_resident, p2->m_resident);
    case COMM:
--- a/Process.c
+++ b/Process.c
@@ -759,7 +759,7 @@ void Process_printRate(RichString* str, double rate, bool coloring) {
       processMegabytesColor = CRT_colors[PROCESS];
    }
 
-   if (isnan(rate)) {
+   if (!isNonnegative(rate)) {
       RichString_appendAscii(str, shadowColor, "        N/A ");
    } else if (rate < 0.005) {
       int len = snprintf(buffer, sizeof(buffer), "%7.2f B/s ", rate);
@@ -792,7 +792,7 @@ void Process_printLeftAlignedField(RichString* str, int attr, const char* conten
 }
 
 void Process_printPercentage(float val, char* buffer, size_t n, uint8_t width, int* attr) {
-   if (val >= 0) {
+   if (isNonnegative(val)) {
       if (val < 0.05F)
          *attr = CRT_colors[PROCESS_SHADOW];
       else if (val >= 99.9F)
--- a/XUtils.c
+++ b/XUtils.c
@@ -12,6 +12,7 @@ in the source distribution for its full text.
 #include <assert.h>
 #include <errno.h>
 #include <fcntl.h>
+#include <math.h>
 #include <stdarg.h>
 #include <stdint.h>
 #include <stdlib.h>
@@ -337,6 +338,16 @@ ssize_t full_write(int fd, const void* buf, size_t count) {
    }
 
    return written;
+}
+
+/* Compares floating point values for ordering data entries. In this function,
+   NaN is considered "less than" any other floating point value (regardless of
+   sign), and two NaNs are considered "equal" regardless of payload. */
+int compareRealNumbers(double a, double b) {
+   int result = isgreater(a, b) - isgreater(b, a);
+   if (result)
+      return result;
+   return !isNaN(a) - !isNaN(b);
 }
 
 /* Computes the sum of all positive floating point values in an array.
--- a/XUtils.c
+++ b/XUtils.c
@@ -19,6 +19,7 @@ in the source distribution for its full text.
 #include <unistd.h>
 
 #include "CRT.h"
+#include "Macros.h"
 
 
 void fail(void) {
@@ -336,4 +337,16 @@ ssize_t full_write(int fd, const void* buf, size_t count) {
    }
 
    return written;
+}
+
+/* Computes the sum of all positive floating point values in an array.
+   NaN values in the array are skipped. The returned sum will always be
+   nonnegative. */
+double sumPositiveValues(const double* array, size_t count) {
+   double sum = 0.0;
+   for (size_t i = 0; i < count; i++) {
+      if (isPositive(array[i]))
+         sum += array[i];
+   }
+   return sum;
 }
--- a/configure.ac
+++ b/configure.ac
@@ -149,7 +149,7 @@ dnl been updated in Autoconf 2.69, so use a workaround:
 m4_version_prereq([2.70], [],
 [if test "x$ac_cv_header_sys_mkdev_h" != xyes; then
    AC_CHECK_HEADER([sys/sysmacros.h], [AC_DEFINE([MAJOR_IN_SYSMACROS], [1],
-      [Define to 1 if `major', `minor', and `makedev' are declared in <sys/sysmacros.h>.])])
+      [Define to 1 if 'major', 'minor', and 'makedev' are declared in <sys/sysmacros.h>.])])
 fi])
 
 # Optional Section
@@ -428,7 +428,7 @@ if test "x$enable_affinity" = xcheck; then
          AC_MSG_RESULT([yes])],
       [enable_affinity=no
          AC_MSG_RESULT([no])],
-      [AC_MSG_RESULT([yes (assumed while cross compiling)])])
+      [AC_MSG_RESULT([assume yes (cross compiling)])])
    fi
 fi
 if test "x$enable_affinity" = xyes; then
@@ -493,7 +493,7 @@ case "$enable_hwloc" in
          PKG_CHECK_MODULES(HWLOC, hwloc, [
                CFLAGS="$CFLAGS $HWLOC_CFLAGS"
                LIBS="$LIBS $HWLOC_LIBS"
-               AC_DEFINE([HAVE_LIBHWLOC], [1], [Define to 1 if you have the `hwloc' library (-lhwloc).])
+               AC_DEFINE([HAVE_LIBHWLOC], [1], [Define to 1 if you have the 'hwloc' library (-lhwloc).])
             ], [
                AC_CHECK_LIB([hwloc], [hwloc_get_proc_cpubind], [], [AC_MSG_ERROR([can not find required library libhwloc])])
                AC_CHECK_HEADERS([hwloc.h], [], [AC_MSG_ERROR([can not find require header file hwloc.h])])
--- a/configure.ac
+++ b/configure.ac
@@ -203,21 +203,42 @@ AC_COMPILE_IFELSE([
 CFLAGS="$old_CFLAGS"
 
 AC_MSG_CHECKING(for NaN support)
-AC_RUN_IFELSE([
+dnl Note: AC_RUN_IFELSE does not try compiling the program at all when
+dnl $cross_compiling is 'yes'.
+AC_LINK_IFELSE([
    AC_LANG_PROGRAM(
       [[
-       #include <math.h>
+#include <math.h>
       ]],
       [[
-       double x = NAN; return !isnan(x);
+         double x = NAN;
+         /* Both should evaluate to false -> 0 (exit success) */
+         return isgreater(x, x) || isgreaterequal(x, x);
       ]]
    )],
-   [AC_MSG_RESULT(yes)],
-   [
+   [flag_finite_math_only=unknown
+   if test "$cross_compiling" = yes; then
+      AC_COMPILE_IFELSE([
+         AC_LANG_SOURCE([[
+/* __FINITE_MATH_ONLY__ is documented in Clang. */
+#ifdef __FINITE_MATH_ONLY__
+#error "should not enable -ffinite-math-only"
+#endif
+         ]])],
+         AC_MSG_RESULT([assume yes (cross compiling)]),
+	 flag_finite_math_only=yes)
+   elif ./conftest$EXEEXT >&AS_MESSAGE_LOG_FD; then
+      flag_finite_math_only=no
+      AC_MSG_RESULT(yes)
+   else
+      flag_finite_math_only=yes
+   fi
+   if test "$flag_finite_math_only" = yes; then
       AC_MSG_RESULT(no)
-      AC_MSG_WARN([Compiler does not respect NaN, some functionality might break; consider using '-fno-finite-math-only'])
-   ],
-   [AC_MSG_RESULT(skipped)])
+      AC_MSG_WARN([runtime behavior with NaN is not compliant - some functionality might break; consider using '-fno-finite-math-only'])
+   fi],
+   [AC_MSG_RESULT(no)
+   AC_MSG_ERROR([can not find required macros: NAN, isgreater() and isgreaterequal()])])
 
 # ----------------------------------------------------------------------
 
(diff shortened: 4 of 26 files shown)


BenBE:
Maybe introduce a `ispositive` and `isnegative` function for the special cases of `isgreater(x, 0.)` and `isless(x, 0.)`.

Furthermore you may update `configure.ac` to check for support for `-fno-signaling-nans` and set it as appropriate, as that's basically what we assume in the code anyway.

BenBE:
Any chance to avoid the `#pragma GCC` in `configure.ac`? You can't guarantee that you're dealing with GCC/LLVM after all, because the requirement is C99 compliant, which technicall includes `tcc` and `icc` too …

Explorer09:
1. I'm considering introducing 3 inline functions in Macros.h:

```c
// Checks if a floating point value is positive. Returns false on a NaN.
static inline bool isPositive(double x) {
   return isgreater(x, 0.0);
}

// Checks if a floating point value is nonnegative. Returns false on a NaN.
static inline bool isNonNegative(double x) {
   return isgreaterequal(x, 0.0);
}

// Cheaper version of C99 isnan(). Unlike isnan(), this may throw FP exception on a "signaling" NaN.
static inline bool isNan(double x) {
   return !isgreaterequal(x, x);
}
```

2. I'm not a fan of overriding CFLAGS specified by the user, but I can add a configure-time check if `-fsignaling-nans` is set and print a warning message saying the flag is not recommended. What do you say?

3.

> Any chance to avoid the `#pragma GCC` in `configure.ac`? You can't guarantee that you're dealing with GCC/LLVM after all, because the requirement is C99 compliant, which technicall includes `tcc` and `icc` too …

That `#pragma` was to silence a GCC warning when checking for `(x != x)` behavior when it comes to NaN. (And I think the warning is a false positive.) The alternative would be to not check for the behavior at all, if we avoid the `(x != x)` expression everywhere.

4. I have a question with this commit 0e922d4085f0285cf353323fd5a96edae604d5da. I don't know why the `isnan` checks are needed. Specifically in what cases the CPU percentages will sum to a NaN value?

BenBE:
> > What's the behaviour for `isNan(inf)`?
> 
> Infinity is technically not NaN. It would return false even on the C99 standard `isnan` function.
> 
> The difference between standard `isnan` and the custom macro is this:
> 
> ```
>               | isnan | (x != x)  | custom macro
> --------------+-------+-----------+-------------
> -Wfloat-equal | no    | warn      | no
> --------------+-------+-----------+-------------
> Finite number | false | false     | false
> Infinity      | false | false     | false
> NaN (quiet)   | true  | true      | true
> SNaN          | true  | FPE; true | FPE; true
> (FPE means floating point exception)
> 
>               | (x >= x)   | isgreaterequal(x,x)
> --------------+------------+--------------------
> Finite number | true       | true
> Infinity      | true       | true
> NaN (quiet)   | FPE; false | false
> SNaN          | FPE; false | FPE; false
> ```

Okay, just wanted to make sure, those edge cases are covered.

> > I think we better avoid the `==`/`!=` comparisons on floats and thus don't need to play around with those `#pragma` as they are inherently non-portable AND that style of comparison usually indicates a bug (thus would raise red flags in a code review).
> 
> Okay. So the inline function `isNan` would be the way to go.

ACK.

> > > 4. I have a question with this commit [0e922d4](https://github.com/htop-dev/htop/commit/0e922d4085f0285cf353323fd5a96edae604d5da). I don't know why the `isnan` checks are needed. Specifically in what cases the CPU percentages will sum to a NaN value?
> > 
> > That's an implementation detail you see when looking at the `CLAMP` macro. Also having that additional check is cleaner IMO.
> 
> I didn't get it. When using the `CLAMP` macro, the second and third operand should never be NaN (there's an assertion for that), and when the first operand is NaN, it throws an FP exception or (if it doesn't trap) simply uses the second (`low`) operand.
> 
> My question was, when could the summing operation result in a NaN? With the code like the following:
> 
> ```c
> percent = v[CPU_METER_NICE] + v[CPU_METER_NORMAL] + v[CPU_METER_KERNEL] + v[CPU_METER_IRQ] + v[CPU_METER_SOFTIRQ]
> ```
> 
> all the values are supposed to be defined and non-NaN, aren't they?

There are some platforms that do not provide all of those values. Thus some of those values can be `NaN`.

Explorer09:
> > > What's the behaviour for `isNan(inf)`?
> > 
> > 
> > Infinity is technically not NaN. It would return false even on the C99 standard `isnan` function.
> > The difference between standard `isnan` and the custom macro is this:
> > ```
> >               | isnan | (x != x)  | custom macro
> > --------------+-------+-----------+-------------
> > -Wfloat-equal | no    | warn      | no
> > --------------+-------+-----------+-------------
> > Finite number | false | false     | false
> > Infinity      | false | false     | false
> > NaN (quiet)   | true  | true      | true
> > SNaN          | true  | FPE; true | FPE; true
> > (FPE means floating point exception)
> > 
> >               | (x >= x)   | isgreaterequal(x,x)
> > --------------+------------+--------------------
> > Finite number | true       | true
> > Infinity      | true       | true
> > NaN (quiet)   | FPE; false | false
> > SNaN          | FPE; false | FPE; false
> > ```
> 
> Okay, just wanted to make sure, those edge cases are covered.
> 
> > > I think we better avoid the `==`/`!=` comparisons on floats and thus don't need to play around with those `#pragma` as they are inherently non-portable AND that style of comparison usually indicates a bug (thus would raise red flags in a code review).
> > 
> > 
> > Okay. So the inline function `isNan` would be the way to go.
> 
> ACK.
> 
> > > > 4. I have a question with this commit [0e922d4](https://github.com/htop-dev/htop/commit/0e922d4085f0285cf353323fd5a96edae604d5da). I don't know why the `isnan` checks are needed. Specifically in what cases the CPU percentages will sum to a NaN value?
> > > 
> > > 
> > > That's an implementation detail you see when looking at the `CLAMP` macro. Also having that additional check is cleaner IMO.
> > 
> > 
> > I didn't get it. When using the `CLAMP` macro, the second and third operand should never be NaN (there's an assertion for that), and when the first operand is NaN, it throws an FP exception or (if it doesn't trap) simply uses the second (`low`) operand.
> > My question was, when could the summing operation result in a NaN? With the code like the following:
> > ```c
> > percent = v[CPU_METER_NICE] + v[CPU_METER_NORMAL] + v[CPU_METER_KERNEL] + v[CPU_METER_IRQ] + v[CPU_METER_SOFTIRQ]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > all the values are supposed to be defined and non-NaN, aren't they?
> 
> There are some platforms that do not provide all of those values. Thus some of those values can be `NaN`.

Explorer09:
> There are some platforms that do not provide all of those values. Thus some of those values can be `NaN`.

And it looks like a bug when I read it. NaNs propagate during addition, and if any value of the `v` array is NaN the whole result would be NaN.

I was considering a function to be added in XUtils.c, but I'm not sure which is the best name for it:
`sumPositiveValues`, `sumPositiveDoubles`, or `sumPositiveDoubleValues`
```c
static double sumPositiveValues(const double* array, size_t count) {
   double sum = 0.0;
   for (size_t i = 0; i < count; i++) {
      if (!isPositive(array[i]))
         continue; // Skip negative and NaN values
      sum += array[i];
      if (!(sum < HUGE_VAL))
         break;
   }
   return sum;
}
```

BenBE:
> I was considering a function to be added in XUtils.c, but I'm not sure which is the best name for it: `sumPositiveValues`, `sumPositiveDoubles`, or `sumPositiveDoubleValues`

`sumNonNaN` … SCNR

Though `sumPositiveValues` is probably fine …

Explorer09:
> `sumNonNaN` … SCNR

This `sumPositiveValues` function would be used in meter code as well (in my branch, at least) and I really intended it to filter out negative values.

linux/LinuxProcessList.c
Explorer09, line 1: I've fixed the logic of computing the "delay percentages" here. The `isnan()` check is technically unnecessary, but the subtractions should have been using `saturatingSub` to avoid negative results. I've also added a conditional to ensure `timeDelta` would never be zero.

linux/LibSensors.c
```diff
@@ -229,22 +229,20 @@ void LibSensors_getCPUTemperatures(CPUData* cpus, unsigned int existingCPUs, uns
    }
 
    /* No package temperature - set to max core temperature */
-   if (isnan(data[0]) && coreTempCount != 0) {
-      double maxTemp = NAN;
+   if (coreTempCount > 0 && isNaN(data[0])) {
+      double maxTemp = -HUGE_VAL;
```
BenBE, line 233: Ensure `data[0]` is initialized as `NAN` to ensure defined value.

Explorer09: I think that is already done in line 147:
```c
   for (size_t i = 0; i < existingCPUs + 1; i++)
      data[i] = NAN;
```

BenBE: k. ACK. Didn't catch that earlier.

Somehow I prefer the original version of that part of the code, as it kept the general pattern closer together. That's not that this version is bad, but a bit more confusing than need be IMHO.

linux/LinuxProcessList.c
```diff
@@ -1003,14 +1003,14 @@ static int handleNetlinkMsg(struct nl_msg* nlmsg, void* linuxProcess) {
...
-      #define BOUNDS(x) (isnan(x) ? 0.0 : ((x) > 100) ? 100.0 : (x))
-      #define DELTAPERC(x,y) BOUNDS((float) ((x) - (y)) / timeDelta * 100)
-      lp->cpu_delay_percent = DELTAPERC(stats.cpu_delay_total, lp->cpu_delay_total);
-      lp->blkio_delay_percent = DELTAPERC(stats.blkio_delay_total, lp->blkio_delay_total);
-      lp->swapin_delay_percent = DELTAPERC(stats.swapin_delay_total, lp->swapin_delay_total);
-      #undef DELTAPERC
-      #undef BOUNDS
+      if (stats.ac_etime * 1000 > lp->delay_read_time) {
+         unsigned long long int timeDelta = stats.ac_etime * 1000 - lp->delay_read_time;
+         #define DELTAPERC(x, y) MINIMUM((float)saturatingSub(x, y) / timeDelta * 100.0F, 100.0F)
```
BenBE, line 1008: Please keep these lowercase …
```suggestion
         #define DELTAPERC(x, y) MINIMUM((float)saturatingSub(x, y) / timeDelta * 100.0f, 100.0f)
```

linux/LinuxProcessList.c
```diff
@@ -1003,14 +1003,14 @@ static int handleNetlinkMsg(struct nl_msg* nlmsg, void* linuxProcess) {
...
-      lp->swapin_delay_percent = DELTAPERC(stats.swapin_delay_total, lp->swapin_delay_total);
-      #undef DELTAPERC
-      #undef BOUNDS
+      if (stats.ac_etime * 1000 > lp->delay_read_time) {
+         unsigned long long int timeDelta = stats.ac_etime * 1000 - lp->delay_read_time;
+         #define DELTAPERC(x, y) MINIMUM((float)saturatingSub(x, y) / timeDelta * 100.0f, 100.0f)
+         lp->cpu_delay_percent = DELTAPERC(stats.cpu_delay_total, lp->cpu_delay_total);
+         lp->blkio_delay_percent = DELTAPERC(stats.blkio_delay_total, lp->blkio_delay_total);
+         lp->swapin_delay_percent = DELTAPERC(stats.swapin_delay_total, lp->swapin_delay_total);
+         #undef DELTAPERC
```
Explorer09, line 1012: There's a change of behavior in my version of the code. When `timeDelta` is zero, the old code would set all `*_delay_percent` members to zero (due to the result of the `BOUNDS` macro). In my version, the percentage values would be left untouched. Should I go back to the old behavior of always setting the value?

BenBE: If changing the behaviour, I think the only correct change would be to use NAN there. The old behaviour would be fine too IMHO.

Explorer09:
One more question. In the PCP platform, I didn't see any code that updates any of the following members of the `PCPProcess` object:
```c
float cpu_delay_percent;
float blkio_delay_percent;
float swapin_delay_percent;
```
Besides they are initialized to 0 in `PCPProcess_new()`, it seems that nothing updates them. They are not even updated to `NAN` when they make sense.

BenBE:
That's a question for @natoscott , but I guess they lack implementation for PCP yet …

Process.c
```diff
@@ -787,7 +787,7 @@ void Process_printLeftAlignedField(RichString* str, int attr, const char* conten
 }
 
 void Process_printPercentage(float val, char* buffer, size_t n, uint8_t width, int* attr) {
-   if (val >= 0) {
+   if (isNonnegative(val)) {
```
Explorer09, line 790: Personal note. With this expression changed to using `isNonnegative`, there is an implicit promotion from `float` to `double`, is the compiler smart enough to realize the promotion can be optimized out and make the expression identical to `isgreaterequal((float)val, 0.0F)` ?

Explorer09: Update. I figured it out by myself. The answer is yes.
`isNonnegative(val)` optimizes to `isgreaterequal((float)val, 0.0F)` with GCC and Clang with `-O1` or higher.

https://godbolt.org/z/h5G1rWf5h

linux/LinuxProcessList.c
Explorer09: I'm considering revising this patch. There are a few issues with the current one.

1. [The Linux Kernel documentation about struct taskstats](https://docs.kernel.org/accounting/taskstats-struct.html) stated that `cpu_delay_total`, `blkio_delay_total` and `swapin_delay_total` may wrap around on overflow. Although the wraparound is unlikely until 2^64 nanoseconds (≈584.5 years, much later than Y2038) of uptime, it is enough to say that using `saturatingSub` would be too defensive on our part. So I might just leave these values wrap around during subtraction as the kernel originally expected.
2. There is still a need to check if the "delta time" is zero, but I don't consider the case being an error. There is (theoretically) a possibility that we scan the process states too fast before the kernel updates the data and thus we get the exact same data as the last scan. My original idea was to keep the percentages computed in the last scan (not turn them to zeros or NaNs) is this case happens.
3. I wish the behaviors of handling "zero delta times" be consistent across platforms, so may tweak the platform codes of Solaris, DragonFlyBSD and others.

BenBE: Can you catalog the behaviours on the various platforms before the change? So we can compare and chose from that?

XUtils.c
```diff
@@ -337,3 +339,25 @@ ssize_t full_write(int fd, const void* buf, size_t count) {
 
    return written;
 }
+
+/* Compares floating point values for ordering data entries. In this function,
+   NaN is considered "less than" any other floating point value (regardless of
+   sign), and two NaNs are considered "equal" regardless of payload. */
+int compareRealNumbers(double a, double b) {
+   int result = (int)isgreater(a, b) - isgreater(b, a);
```
BenBE, line 347:
```suggestion
   int result = (int)isgreater(a, b) - (int)isgreater(b, a);
```

Shouldn't this actually include both casts. And AFAICS both should be implicit anyway.

BenBE:
@Explorer09 Anything left you want to change before we can go ahead with this one?

Explorer09:
@BenBE 
1. The "delta time" calculation code. When I tried to eliminate the usage of `isnan` calls, this is the only code I haven't finished changing. The goal was to eliminate division by zeros so that the NaN values can be reserved for "not applicable", "not supported by OS" or "error retrieving said data". (Note: I can defer this to another PR if you guys are in a hurry for merging this PR.)
2. `PCPDynamicColumn_compareByKey()`: The `float` and `double` values may be compared using `compareRealNumbers()`, but I am not sure if the ordering is desireble for that part of comparison. Can PCP support code developers look at it?

XUtils.c
Explorer09: The cast may be removed or both made explicit. I don't know which looks neater.

BenBE:
> @BenBE
> 
> 1. The "delta time" calculation code. When I tried to eliminate the usage of `isnan` calls, this is the only code I haven't finished changing. The goal was to eliminate division by zeros so that the NaN values can be reserved for "not applicable", "not supported by OS" or "error retrieving said data". (Note: I can defer this to another PR if you guys are in a hurry for merging this PR.)

How much code does this affect? And is this sufficiently distinct from the existing change to go in a different PR? I mean, if this is basically a code change mostly similar to the other changes we've seen so far, i'm quite inclined to say to keep it in this PR. Though if the changes do end up causing a noticeable difference in behaviour, like they probably will, there's good reason to split them just in case we need to partially undo those changes.

And although I'd prefer to go ahead with this PR soonish™ there's actually no preasure to have it been done by yesterday. ;-) Call is yours, just don't drag it on for too long. :)

> 2. `PCPDynamicColumn_compareByKey()`: The `float` and `double` values may be compared using `compareRealNumbers()`, but I am not sure if the ordering is desireble for that part of comparison. Can PCP support code developers look at it?

This should be best directed at @natoscott and @smalinux, who mostly worked on that part.

linux/LinuxProcessList.c
Explorer09: Splitting this part to issue #1274 in order to simplify this patch of the PR.

BenBE (commented):
Looks reasonable.

natoscott:
> That's a question for @natoscott , but I guess they lack implementation for PCP yet …

@Explorer09 @BenBE yep, thanks - these are leftovers from using Linux headers/code at one point in the PCP platform.  We'd need some PCP metrics to provide those netlink stats, but that doesn't exist currently so I'll open a separate PR to drop these fields for now.

Explorer09:
Just to give a note: This PR is ready for review and merging.

natoscott (approved):
Small thing, rest LGTM.

XUtils.c
natoscott: Removed is neater.

Macros.h
```diff
@@ -98,6 +100,24 @@
...
+/* Cheaper function for checking NaNs. Unlike the standard isnan(), this may
+   throw an FP exception on a "signaling" NaN.
+   (ISO/IEC TS 18661-1 and the C23 standard stated that isnan() throws no
+   exceptions even with a "signaling" NaN) */
+static inline bool isNaN(double x) {
+   return !isgreaterequal(x, x);
+}
+
+/* Checks if x is nonnegative. Returns false if x is NaN. */
+static inline bool isNonnegative(double x) {
```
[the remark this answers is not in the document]
Explorer09: Well…
https://math.stackexchange.com/questions/3342643/

natoscott:
> PCPDynamicColumn_compareByKey(): The float and double values may be compared using compareRealNumbers(), but I am not sure if the ordering is desireble for that part of comparison. Can PCP support code developers look at it?

Leave this with me, these can use compareRealNumbers but I need to look more closely at the ordering aspect (any change here can be done subsequently).  Thanks!

Macros.h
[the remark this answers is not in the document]
natoscott: Interesting!  Fine to leave it as-is then IMO, thanks for the note.

configure.ac
```diff
@@ -203,21 +203,42 @@ AC_COMPILE_IFELSE([
...
+   if test "$cross_compiling" = yes; then
+      AC_COMPILE_IFELSE([
+         AC_LANG_SOURCE([[
+/* __FINITE_MATH_ONLY__ is documented in Clang. */
+#ifdef __FINITE_MATH_ONLY__
+#error "should not enable -ffinite-math-only"
+#endif
+         ]])],
+         AC_MSG_RESULT([assume yes (cross compiling)]),
+	 flag_finite_math_only=yes)
```
Explorer09, line 229: Is @BenBE able to see this even though the PR is closed?
I accidentally left a tab character in this line in the commit edc7c40a1df50eec2ac85b3ab33dc30fa54685b8. It should be spaces indent instead. This issue might be too small to file a new PR for it.

```suggestion
         flag_finite_math_only=yes)
```

BenBE: @Explorer09 Yes, I can see comments; especially when they mention me … ;-)

I think I'll do some whitespace / formatting run at some time. There's likely more that slipped in.

PR #1272, opened 2023-07-26
Add cpufreq support on Darwin for Apple Silicon machines

### LINKED ISSUE #280: CPU Frequency shows as "N/A" on macOS 10.14 Catalina 2015 MacBook Pro 4 core, 8 thread.
image

Not sure if it's a known issue or something that I'm maybe doing wrong. Anyway, thought I'd post it to get some feedback.

Thanks!

bakaid:
This PR adds CPU frequency support on Darwin for Apple Silicon machines.

After doing research on this, it seems to me that the only way to get cpufreq data is to gather some information from the IODeviceTree using IORegistry and then get performance statistics using libIOReport. Unfortunately, the former's structure is not guaranteed to be stable, and the latter has a private API, so again, it can break at any update.
That being said, based on my research on powermetrics and the available device trees of Apple Silicon SoCs, the solution presented here seems to be the current best option available.

I tested this on vanilla M1 and M2, but not on Max/Pro (different core count) or Ultra (multi-die) CPUs. I am reasonably certain it should work on those as well, as we use cluster-type (E or P), not cluster-id for classifying cores, but someone with those CPUs testing this would be very helpful.

Resolves #280

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -62,6 +62,18 @@ AC_USE_SYSTEM_EXTENSIONS
 
 # ----------------------------------------------------------------------
 
+# ----------------------------------------------------------------------
+# Checks for cpu.
+# ----------------------------------------------------------------------
+
+case "$host_cpu" in
+aarch64*)
+   my_htop_cpu=aarch64
+   AC_DEFINE([HTOP_AARCH64], [], [Building for aarch64.])
+   ;;
+esac
+
+# ----------------------------------------------------------------------
 
 # ----------------------------------------------------------------------
 # Checks for compiler.
@@ -679,6 +691,30 @@ fi
 
 
 # ----------------------------------------------------------------------
+# Checks for Darwin features and flags.
+# ----------------------------------------------------------------------
+
+if test "$my_htop_cpu" = aarch64; then
+   AC_ARG_ENABLE([libioreport],
+               [AS_HELP_STRING([--enable-libioreport],
+               [enable libIOReport support for getting cpu frequency on Apple Silicon machines @<:@default=yes@:>@])],
+               [],
+               [enable_libioreport=yes])
+   case "$enable_libioreport" in
+      no)
+         ;;
+      yes)
+         AC_CHECK_LIB([IOReport], [IOReportCreateSubscription])
+         ;;
+      *)
+         AC_MSG_ERROR([bad value '$enable_libioreport' for --enable_libioreport])
+         ;;
+   esac
+fi
+
+# ----------------------------------------------------------------------
+
+# ----------------------------------------------------------------------
 # Checks for compiler warnings.
 # ----------------------------------------------------------------------
 
@@ -768,6 +804,8 @@ AM_CONDITIONAL([HTOP_DARWIN], [test "$my_htop_platform" = darwin])
 AM_CONDITIONAL([HTOP_SOLARIS], [test "$my_htop_platform" = solaris])
 AM_CONDITIONAL([HTOP_PCP], [test "$my_htop_platform" = pcp])
 AM_CONDITIONAL([HTOP_UNSUPPORTED], [test "$my_htop_platform" = unsupported])
+
+AM_CONDITIONAL([HTOP_AARCH64], [test "$my_htop_cpu" = aarch64])
 
 AC_SUBST(my_htop_platform)
 AC_CONFIG_FILES([Makefile htop.1])
--- a/darwin/CpuFreq.c
+++ b/darwin/CpuFreq.c
@@ -1,0 +1,176 @@
+#include "darwin/CpuFreq.h"
+
+#ifdef CPUFREQ_SUPPORT
+
+#include <stdlib.h>
+#include <string.h>
+#include <stdint.h>
+
+/* Implementations */
+int CpuFreq_init(const Machine* machine, CpuFreqData* data) {
+    data->existingCPUs = machine->existingCPUs;
+    data->cluster_type_per_cpu = xCalloc(data->existingCPUs, sizeof(uint32_t));
+    data->frequencies = xCalloc(data->existingCPUs, sizeof(double));
+
+    /* Determine the cluster type for all CPUs */
+    char buf[128];
+
+    for (uint32_t num_cpus = 0U; num_cpus < data->existingCPUs; num_cpus++) {
+        snprintf(buf, sizeof(buf), "IODeviceTree:/cpus/cpu%u", num_cpus);
+        io_registry_entry_t cpu_io_registry_entry = IORegistryEntryFromPath(kIOMainPortDefault, buf);
+        if (cpu_io_registry_entry == MACH_PORT_NULL) {
+            return -1;
+        }
+
+        CFDataRef cluster_type_ref = (CFDataRef) IORegistryEntryCreateCFProperty(cpu_io_registry_entry, CFSTR("cluster-type"), kCFAllocatorDefault, 0U);
+        if (cluster_type_ref == NULL) {
+            IOObjectRelease(cpu_io_registry_entry);
+            return -1;
+        }
+        if (CFDataGetLength(cluster_type_ref) != 2) {
+            CFRelease(cluster_type_ref);
+            IOObjectRelease(cpu_io_registry_entry);
+            return -1;
+        }
+        UniChar cluster_type_char;
+        CFDataGetBytes(cluster_type_ref, CFRangeMake(0, 2), (uint8_t*) &cluster_type_char);
+        CFRelease(cluster_type_ref);
+
+        uint32_t cluster_type = 0U;
+        switch (cluster_type_char) {
+            case L'E':
+                cluster_type = 0U;
+                break;
+            case L'P':
+                cluster_type = 1U;
+                break;
+            default:
+                /* Unknown cluster type */
+                IOObjectRelease(cpu_io_registry_entry);
+                return -1;
+        }
+
+        data->cluster_type_per_cpu[num_cpus] = cluster_type;
+
+        IOObjectRelease(cpu_io_registry_entry);
+    }
+
+    /*
+     * Determine frequencies for per-cluster-type performance states
+     * Frequencies for the "E" cluster type are stored in voltage-states1,
+     * frequencies for the "P" cluster type are stored in voltage-states5.
+     * This seems to be hardcoded.
+     */
+    const CFStringRef voltage_states_key_per_cluster[CPUFREQ_NUM_CLUSTER_TYPES] = {CFSTR("voltage-states1"), CFSTR("voltage-states5")};
+
+    io_registry_entry_t pmgr_registry_entry = IORegistryEntryFromPath(kIOMainPortDefault, "IODeviceTree:/arm-io/pmgr");
+    if (pmgr_registry_entry == MACH_PORT_NULL) {
+        return -1;
+    }
+    for (size_t i = 0U; i < CPUFREQ_NUM_CLUSTER_TYPES; i++) {
+        CFDataRef voltage_states_ref =
+            (CFDataRef) IORegistryEntryCreateCFProperty(pmgr_registry_entry, voltage_states_key_per_cluster[i], kCFAllocatorDefault, 0U);
+        if (voltage_states_ref == NULL) {
+            IOObjectRelease(pmgr_registry_entry);
+            return -1;
+        }
+
+        CpuFreqPowerStateFrequencies* cluster_frequencies = &data->cpu_frequencies_per_cluster_type[i];
+        cluster_frequencies->num_frequencies = CFDataGetLength(voltage_states_ref) / 8;
+        cluster_frequencies->frequencies = xCalloc(cluster_frequencies->num_frequencies, sizeof(double));
+        const uint8_t* voltage_states_data = CFDataGetBytePtr(voltage_states_ref);
+        for (size_t j = 0U; j < cluster_frequencies->num_frequencies; j++) {
+            uint32_t freq_value;
+            memcpy(&freq_value, voltage_states_data + j * 8, 4);
+            cluster_frequencies->frequencies[j] = (65536000.0 / freq_value) * 1000000;
+        }
+        CFRelease(voltage_states_ref);
+    }
+    IOObjectRelease(pmgr_registry_entry);
+
+
+    /* Create subscription for CPU performance states */
+    CFMutableDictionaryRef channels = IOReportCopyChannelsInGroup(CFSTR("CPU Stats"), CFSTR("CPU Core Performance States"), NULL, NULL);
+    if (channels == NULL) {
+        return -1;
+    }
+
+    data->subscribed_channels = NULL;
+    data->subscription = IOReportCreateSubscription(NULL, channels, &data->subscribed_channels, 0U, NULL);
+
+    CFRelease(channels);
+
+    if (data->subscription == NULL) {
+        return -1;
+    }
+
+    data->prev_samples = NULL;
+
+    return 0;
+}
+
+void CpuFreq_update(CpuFreqData* data) {
+    CFDictionaryRef samples = IOReportCreateSamples(data->subscription, data->subscribed_channels, NULL);
+    if (samples == NULL) {
+        return;
+    }
+
+    if (data->prev_samples == NULL) {
+        data->prev_samples = samples;
+        return;
+    }
+
+    /* Residency is cumulative, we have to diff two samples to get a current view */
+    CFDictionaryRef samples_delta = IOReportCreateSamplesDelta(data->prev_samples, samples, NULL);
+
+    /* Iterate through performance state residencies. Index 0 is the idle residency, index 1-n is the per-performance state residency. */
+    __block uint32_t cpu_i = 0U;
+    IOReportIterate(samples_delta, ^(IOReportSampleRef ch) {
+        if (cpu_i >= data->existingCPUs) {
+            /* The report contains more CPUs than we know about. This should not happen. */
+            return kIOReportIterOk; // TODO: find way to possibly stop iteration early on error
+        }
+        const CpuFreqPowerStateFrequencies* cpu_frequencies = &data->cpu_frequencies_per_cluster_type[data->cluster_type_per_cpu[cpu_i]];
+        uint32_t state_count = IOReportStateGetCount(ch);
+        if (state_count != cpu_frequencies->num_frequencies + 1) {
+            /* The number of reported states does not match the number of previously queried frequencies. This should not happen. */
+            return kIOReportIterOk; // TODO: find way to possibly stop iteration early on error
+        }
+        /* Calculate average frequency based on residency and per-performance state frequency */
+        double average_freq = 0.0;
+        int64_t total_residency = 0U;
+        for (uint32_t i = 0U; i < state_count; i++) {
+            const int64_t residency = IOReportStateGetResidency(ch, i);
+            total_residency += residency;
+            /* We count idle as the smallest frequency */
+            average_freq += residency * cpu_frequencies->frequencies[(i == 0) ? 0 : (i - 1)];
+        }
+        average_freq /= total_residency;
+        data->frequencies[cpu_i] = average_freq / 1000000; // Convert to MHz
+
+        cpu_i++;
+        return kIOReportIterOk;
+    });
+    CFRelease(samples_delta);
+
+    CFRelease(data->prev_samples);
+    data->prev_samples = samples;
+}
+
+void CpuFreq_cleanup(CpuFreqData* data) {
+    if (data->subscription != NULL) {
+        CFRelease(data->subscription);
+    }
+    if (data->subscribed_channels != NULL) {
+        CFRelease(data->subscribed_channels);
+    }
+    if (data->prev_samples != NULL) {
+        CFRelease(data->prev_samples);
+    }
+    free(data->cluster_type_per_cpu);
+    free(data->frequencies);
+    for (size_t i = 0U; i < CPUFREQ_NUM_CLUSTER_TYPES; i++) {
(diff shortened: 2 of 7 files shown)


BenBE:
> This PR adds CPU frequency support on Darwin for Apple Silicon machines.

Nice.

> After doing research on this, it seems to me that the only way to get cpufreq data is to gather some information from the IODeviceTree using IORegistry and then get performance statistics using libIOReport.

I hope the source contains some documentation on the details of this.

> Unfortunately, the former's structure is not guaranteed to be stable, and the latter has a private API, so again, it can break at any update.

Calling me surprised by this, in the context if this being apple, would be an overstatement …

> That being said, based on my research on powermetrics and the available device trees of Apple Silicon SoCs, the solution presented here seems to be the current best option available.
> 
> I tested this on vanilla M1 and M2, but not on Max/Pro (different core count) or Ultra (multi-die) CPUs. I am reasonably certain it should work on those as well, as we use cluster-type (E or P), not cluster-id for classifying cores, but someone with those CPUs testing this would be very helpful.

I keep a firm distance from any fruit-based hardware, thus I cannot test this myself either.

darwin/CpuFreq.c
```diff
@@ -0,0 +1,176 @@
...
+    data->subscribed_channels = NULL;
+    data->subscription = IOReportCreateSubscription(NULL, channels, &data->subscribed_channels, 0U, NULL);
+
+    CFRelease(channels);
+
+    if (data->subscription == NULL) {
+        return -1;
+    }
+
+    data->prev_samples = NULL;
```
BenBE, line 107: IF this is for field initialization it shouldn't go that far to the end. Especially as it's not used anywhere else in this function.

darwin/CpuFreq.c
```diff
@@ -0,0 +1,176 @@
+#include "darwin/CpuFreq.h"
+
+#ifdef CPUFREQ_SUPPORT
+
+#include <stdlib.h>
+#include <string.h>
+#include <stdint.h>
+
+/* Implementations */
+int CpuFreq_init(const Machine* machine, CpuFreqData* data) {
```
BenBE, line 10: If you are just returning success, prefer `bool` from `stdbool.h`

darwin/CpuFreq.c
```diff
@@ -0,0 +1,176 @@
...
+    data->existingCPUs = machine->existingCPUs;
+    data->cluster_type_per_cpu = xCalloc(data->existingCPUs, sizeof(uint32_t));
+    data->frequencies = xCalloc(data->existingCPUs, sizeof(double));
+
+    /* Determine the cluster type for all CPUs */
+    char buf[128];
```
BenBE, line 16: Limit the scope of this stack buffer to the body of the loop.

darwin/CpuFreq.c
```diff
@@ -0,0 +1,176 @@
...
+     * Determine frequencies for per-cluster-type performance states
+     * Frequencies for the "E" cluster type are stored in voltage-states1,
+     * frequencies for the "P" cluster type are stored in voltage-states5.
+     * This seems to be hardcoded.
+     */
+    const CFStringRef voltage_states_key_per_cluster[CPUFREQ_NUM_CLUSTER_TYPES] = {CFSTR("voltage-states1"), CFSTR("voltage-states5")};
+
+    io_registry_entry_t pmgr_registry_entry = IORegistryEntryFromPath(kIOMainPortDefault, "IODeviceTree:/arm-io/pmgr");
+    if (pmgr_registry_entry == MACH_PORT_NULL) {
+        return -1;
```
BenBE, line 68: Memory leak.

darwin/CpuFreq.c
```diff
@@ -0,0 +1,176 @@
...
+    }
+    for (size_t i = 0U; i < CPUFREQ_NUM_CLUSTER_TYPES; i++) {
+        CFDataRef voltage_states_ref =
+            (CFDataRef) IORegistryEntryCreateCFProperty(pmgr_registry_entry, voltage_states_key_per_cluster[i], kCFAllocatorDefault, 0U);
+        if (voltage_states_ref == NULL) {
+            IOObjectRelease(pmgr_registry_entry);
+            return -1;
```
BenBE, line 75: Memory leak.

darwin/CpuFreq.c
```diff
@@ -0,0 +1,176 @@
...
+        }
+
+        CpuFreqPowerStateFrequencies* cluster_frequencies = &data->cpu_frequencies_per_cluster_type[i];
+        cluster_frequencies->num_frequencies = CFDataGetLength(voltage_states_ref) / 8;
+        cluster_frequencies->frequencies = xCalloc(cluster_frequencies->num_frequencies, sizeof(double));
+        const uint8_t* voltage_states_data = CFDataGetBytePtr(voltage_states_ref);
+        for (size_t j = 0U; j < cluster_frequencies->num_frequencies; j++) {
+            uint32_t freq_value;
+            memcpy(&freq_value, voltage_states_data + j * 8, 4);
```
BenBE, line 84: Where is that 8 coming from?

darwin/CpuFreq.c
```diff
@@ -0,0 +1,176 @@
...
+            cluster_frequencies->frequencies[j] = (65536000.0 / freq_value) * 1000000;
```
BenBE, line 85: Given that the `uint32_t` from `freq_value` is promoted to `double` for this division anyways, let's shorten this expression:

```suggestion
            cluster_frequencies->frequencies[j] = 65536e9 / freq_value;
```

darwin/CpuFreq.c
```diff
@@ -0,0 +1,176 @@
...
+        }
+        CFRelease(voltage_states_ref);
+    }
+    IOObjectRelease(pmgr_registry_entry);
+
+
+    /* Create subscription for CPU performance states */
+    CFMutableDictionaryRef channels = IOReportCopyChannelsInGroup(CFSTR("CPU Stats"), CFSTR("CPU Core Performance States"), NULL, NULL);
+    if (channels == NULL) {
+        return -1;
```
BenBE, line 95: Memory leak.

darwin/CpuFreq.c
```diff
@@ -0,0 +1,176 @@
...
+            cluster_frequencies->frequencies[j] = (65536000.0 / freq_value) * 1000000;
+        }
+        CFRelease(voltage_states_ref);
+    }
+    IOObjectRelease(pmgr_registry_entry);
+
+
+    /* Create subscription for CPU performance states */
+    CFMutableDictionaryRef channels = IOReportCopyChannelsInGroup(CFSTR("CPU Stats"), CFSTR("CPU Core Performance States"), NULL, NULL);
+    if (channels == NULL) {
```
BenBE, line 94:
```suggestion
    if (!channels) {
```

would be also fine …

darwin/CpuFreq.c
```diff
@@ -0,0 +1,176 @@
...
+        return -1;
+    }
+
+    data->subscribed_channels = NULL;
+    data->subscription = IOReportCreateSubscription(NULL, channels, &data->subscribed_channels, 0U, NULL);
+
+    CFRelease(channels);
+
+    if (data->subscription == NULL) {
```
BenBE, line 103:
```suggestion
    if (!data->subscription) {
```

darwin/CpuFreq.c
```diff
@@ -0,0 +1,176 @@
...
+        return -1;
```
BenBE, line 104: Memory leak.

darwin/CpuFreq.c
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    data->prev_samples = NULL;
+
+    return 0;
+}
+
+void CpuFreq_update(CpuFreqData* data) {
+    CFDictionaryRef samples = IOReportCreateSamples(data->subscription, data->subscribed_channels, NULL);
+    if (samples == NULL) {
```
BenBE, line 114:
```suggestion
    if (!samples) {
```

darwin/CpuFreq.c
```diff
@@ -0,0 +1,176 @@
...
+        return;
+    }
+
+    if (data->prev_samples == NULL) {
```
BenBE, line 118:
```suggestion
    if (!data->prev_samples) {
```

darwin/CpuFreq.c
```diff
@@ -0,0 +1,176 @@
...
+
+    data->prev_samples = NULL;
+
+    return 0;
+}
+
+void CpuFreq_update(CpuFreqData* data) {
+    CFDictionaryRef samples = IOReportCreateSamples(data->subscription, data->subscribed_channels, NULL);
+    if (samples == NULL) {
+        return;
```
BenBE, line 115: If no `samples` could be gathered, technically we should also update the `prev_samples` …

darwin/CpuFreq.c
```diff
@@ -0,0 +1,176 @@
...
+        data->prev_samples = samples;
+        return;
+    }
+
+    /* Residency is cumulative, we have to diff two samples to get a current view */
+    CFDictionaryRef samples_delta = IOReportCreateSamplesDelta(data->prev_samples, samples, NULL);
+
+    /* Iterate through performance state residencies. Index 0 is the idle residency, index 1-n is the per-performance state residency. */
+    __block uint32_t cpu_i = 0U;
+    IOReportIterate(samples_delta, ^(IOReportSampleRef ch) {
```
BenBE, line 128: As that's not strictly C99 conforming code, I assume that's intermixed ObjC stuff thrown in for good measure?

darwin/CpuFreq.c
```diff
@@ -0,0 +1,176 @@
...
+void CpuFreq_cleanup(CpuFreqData* data) {
+    if (data->subscription != NULL) {
+        CFRelease(data->subscription);
+    }
+    if (data->subscribed_channels != NULL) {
+        CFRelease(data->subscribed_channels);
+    }
+    if (data->prev_samples != NULL) {
+        CFRelease(data->prev_samples);
+    }
```
BenBE, line 169:
```suggestion
    if (data->subscription)
        CFRelease(data->subscription);
    if (data->subscribed_channels)
        CFRelease(data->subscribed_channels);
    if (data->prev_samples)
        CFRelease(data->prev_samples);
```

Normally you would even drop the NULL check, but given that `CFRelease` is [kinda picky in that regard](https://stackoverflow.com/a/1608425), these checks need to stay … :( But we can at least reduce the noise …

darwin/CpuFreq.c
```diff
@@ -0,0 +1,176 @@
...
+    free(data->cluster_type_per_cpu);
+    free(data->frequencies);
+    for (size_t i = 0U; i < CPUFREQ_NUM_CLUSTER_TYPES; i++) {
+        free(data->cpu_frequencies_per_cluster_type[i].frequencies);
+    }
```
BenBE, line 174: Isn't this missing a follow-up call of `free(data->cpu_frequencies_per_cluster_type);`???

darwin/DarwinMachine.c
```diff
@@ -76,6 +76,9 @@ void Machine_scan(Machine* super) {
    DarwinMachine_allocateCPULoadInfo(&host->curr_load);
    DarwinMachine_getVMStats(&host->vm_stats);
    openzfs_sysctl_updateArcStats(&host->zfs);
+#ifdef CPUFREQ_SUPPORT
+   CpuFreq_update(&host->cpu_freq);
```
BenBE, line 80: Shouldn't this check for `this->cpu_freq_ok` if it's stored from init earlier?

darwin/CpuFreq.h
```diff
@@ -0,0 +1,66 @@
+#ifndef HEADER_CpuFreq
+#define HEADER_CpuFreq
+
+#include "Machine.h"
+
+#if (defined(HAVE_LIBIOREPORT) && defined(HTOP_AARCH64))
```
BenBE, line 6:
```suggestion
#if defined(HAVE_LIBIOREPORT) && defined(HTOP_AARCH64)
```
We're not writing Lisp …

darwin/DarwinMachine.h
```diff
@@ -21,6 +21,10 @@ typedef struct DarwinMachine_ {
    vm_statistics_data_t vm_stats;
    processor_cpu_load_info_t prev_load;
    processor_cpu_load_info_t curr_load;
+#ifdef CPUFREQ_SUPPORT
+   CpuFreqData cpu_freq;
+   bool cpu_freq_ok;
```
BenBE, line 26: By refactoring `CpuFreq_init` to return `CpuFreqData*` on success and `NULL` on failure, the `bool` can be omitted.

darwin/Platform.c
```diff
@@ -279,7 +279,15 @@ double Platform_setCPUValues(Meter* mtr, unsigned int cpu) {
    /* Convert to percent and return */
    total = mtr->values[CPU_METER_NICE] + mtr->values[CPU_METER_NORMAL] + mtr->values[CPU_METER_KERNEL];
 
+#ifdef CPUFREQ_SUPPORT
+   if (dhost->cpu_freq_ok) {
+      mtr->values[CPU_METER_FREQUENCY] = dhost->cpu_freq.frequencies[cpu - 1];
+   } else {
+      mtr->values[CPU_METER_FREQUENCY] = NAN;
+   }
```
BenBE, line 287:
```suggestion
   mtr->values[CPU_METER_FREQUENCY] = dhost->cpu_freq_ok ? dhost->cpu_freq.frequencies[cpu - 1] : NAN;
```

BenBE (changes requested):
Please squash your changes into the commits of this PR to keep the patch set clean.

darwin/CpuFreq.c
bakaid: Not ObjC, rather an Apple extension for [Blocks](https://en.wikipedia.org/wiki/Blocks_(C_language_extension)#:~:text=Blocks%20are%20a%20non%2Dstandard,create%20closures%20within%20these%20languages.). We must use it, since that's what the API expects for aa a callback.

darwin/CpuFreq.c
```diff
@@ -0,0 +1,176 @@
...
+    data->frequencies = xCalloc(data->existingCPUs, sizeof(double));
+
+    /* Determine the cluster type for all CPUs */
+    char buf[128];
+
+    for (uint32_t num_cpus = 0U; num_cpus < data->existingCPUs; num_cpus++) {
+        snprintf(buf, sizeof(buf), "IODeviceTree:/cpus/cpu%u", num_cpus);
+        io_registry_entry_t cpu_io_registry_entry = IORegistryEntryFromPath(kIOMainPortDefault, buf);
+        if (cpu_io_registry_entry == MACH_PORT_NULL) {
+            return -1;
```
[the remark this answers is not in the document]
bakaid: For this and all similar subsequent comments: the intention was that even on failure, we should be in a consistent state, so CpuFreq_cleanup can (and should) be called (and it is, if you look at DarwinMachine.c). If that's not preferred, I can refactor, but it will introduce more boilerplate error handling.

darwin/CpuFreq.c
bakaid: This is how these values are stored, in an unstructured byte array, every 8 bytes storing one frequency value. There is no documentation for this, it had to be determined based on trial and error (and matching it with the frequency values displayed by `powermetrics`).

darwin/CpuFreq.h
```diff
@@ -0,0 +1,66 @@
...
+
+/*
+ * Seems to be hardcoded for now on all Apple Silicon platforms, no way to get it dynamically.
+ * Current cluster types are "E" for efficiency cores and "P" for performance cores.
+ */
+#define CPUFREQ_NUM_CLUSTER_TYPES 2
+
+typedef struct {
+    /* Number of CPUs */
+    unsigned int existingCPUs;
```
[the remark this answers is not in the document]
bakaid: I would prefer `size_t` too, but it is `unsigned int` in `Machine`, where I copy it from.

darwin/CpuFreq.h
```diff
@@ -0,0 +1,66 @@
...
+typedef CFDictionaryRef IOReportSampleRef;
+typedef CFDictionaryRef IOReportChannelRef;
+typedef int (^io_report_iterate_callback_t)(IOReportSampleRef ch);
+extern void IOReportIterate(CFDictionaryRef samples, io_report_iterate_callback_t callback);
+extern CFMutableDictionaryRef IOReportCopyChannelsInGroup(CFStringRef, CFStringRef, void*, void*);
+extern IOReportSubscriptionRef IOReportCreateSubscription(void *a, CFMutableDictionaryRef desiredChannels, CFMutableDictionaryRef *subbedChannels, uint64_t channel_id, CFTypeRef b);
+extern CFDictionaryRef IOReportCreateSamples(IOReportSubscriptionRef iorsub, CFMutableDictionaryRef subbedChannels, CFTypeRef a);
+extern uint32_t IOReportStateGetCount(IOReportChannelRef ch);
+extern uint64_t IOReportStateGetResidency(IOReportChannelRef ch, uint32_t index);
+extern CFDictionaryRef IOReportCreateSamplesDelta(CFDictionaryRef prev, CFDictionaryRef current, CFTypeRef a);
```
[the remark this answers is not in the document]
bakaid: Some of it is needed in the `CpuFreqData` struct, most of it I can move to the implementation.

darwin/CpuFreq.c
bakaid: That's why they are there :) If NULL equivalency check is not the preferred style, I will refactor those along with the rest.

darwin/CpuFreq.c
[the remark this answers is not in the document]
BenBE: Error handling is one of the rare cases where `goto err;` is actually a quite common pattern.

darwin/CpuFreq.c
BenBE: Would be nice to have some documentation of this somewhere. You could even do a packed structure with these values and reference those values that way.

configure.ac
```diff
@@ -769,6 +805,8 @@ AM_CONDITIONAL([HTOP_SOLARIS], [test "$my_htop_platform" = solaris])
 AM_CONDITIONAL([HTOP_PCP], [test "$my_htop_platform" = pcp])
 AM_CONDITIONAL([HTOP_UNSUPPORTED], [test "$my_htop_platform" = unsupported])
 
+AM_CONDITIONAL([HTOP_AARCH64], [test "$my_htop_cpu" = aarch64])
+
```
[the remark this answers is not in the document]
Explorer09: Can we just use the predefined macro `__aarch64__` to detect the CPU architecture, and combine with `HTOP_DARWIN` defined by the configure script to get what we need?

joske:
@bakaid how did you figure this out? Nothing is to be found about this API at all. Did you reverse engineer powermetrics?

I'm interested in this myself as I wrote the macos (and open/freeBSD) code for btop. I'd like to use parts of this code, but as htop is GPL and btop is Apache licensed, this is problematic. Would you be open to dual license the cpu frequency bit?

bakaid:
> @bakaid how did you figure this out? Nothing is to be found about this API at all. Did you reverse engineer powermetrics?
> 
> I'm interested in this myself as I wrote the macos (and open/freeBSD) code for btop. I'd like to use parts of this code, but as htop is GPL and btop is Apache licensed, this is problematic. Would you be open to dual license the cpu frequency bit?



> @bakaid how did you figure this out? Nothing is to be found about this API at all. Did you reverse engineer powermetrics?
> 
> I'm interested in this myself as I wrote the macos (and open/freeBSD) code for btop. I'd like to use parts of this code, but as htop is GPL and btop is Apache licensed, this is problematic. Would you be open to dual license the cpu frequency bit?

The only reference available for this is the behaviour of `powermetrics` indeed.

After this PR has been submitted, I finally had the opportunity to test this on an M2 Pro machine, and realized that unfortunately, this solution is incomplete.

With Pro, Max and Ultra Apple Silicon CPUs, there is more than one P-cluster, and in the case of Ultra, more than one die.
In these cases, the `CPU Core Performance States`, which seem to work fine with normal M1 and M2, start to become nonsensical, even in `powermetrics`, especially when CPU load is high.

It seems to me as if there is a mode, where individual cores have their individual scaled frequencies and a mode, where an entire cluster is driven with the same frequency, the latter especially prominent under larger loads.

This "cluster-wide same frequency mode" is reported in another channel, `CPU Complex Performance States`. When the load is high, and the cores of a cluster always in this cluster-wide state, the reported residency in `CPU Core Performance States` for per-core frequencies can be 0 samples, causing even `powermetrics` to display `NaN` for `CPU <N> active frequency` after a divison with 0, and only display a sensible value for `E/P<N>-Cluster HW active frequency` for the cluster-wise residency.
Even worse, under intermediate loads, there is a mixture of per-core and per-cpu complex states in a single measured interval, making the frequency output of `powermetrics` practically useless.

The solution, it seems to me, is to merge these two datasets, and synthesize an average frequency for each core, based on the the weighted average of its own individual `CPU Core Performance States` samples AND the `CPU Complex Performance States` samples of the cluster it belongs to.

This is what I plan to do, but haven't had the time (and reliable access to Pro/Max/Ultra machines) to do it yet.

As for licensing, once this is finalized, I am happy to dual-license the code.

joske:
@bakaid thanks for all the info! I only have an M1 and an M2 myself. I assumed this code would also work on x86, but apparently the IORegistry nodes `/cpus/cpux` don't exist there.

Great that you are open to dual license your work! 🎉  🙏

bakaid:
@joske of course, happy to do it.

Indeed, this solution is Apple Silicon-exclusive, `powermetrics` works entirely differently for x86.

Now that I see that there is interest in this, I'll try to make progress with the complex performance states.

joske:
Btw, btop can only display a single frequency, so I was only showing the highest one.

joske:
@bakaid any progress?

PR #1273, opened 2023-07-28
Remove VLA

BenBE:
This removes various instances of VLA in the code.

The two instances in `RichString.c` (`RichString_writeFromWide` & `RichString_appendnWideColumns`) have been kept intentionally, for performance reasons AND the alternatives (like `alloca`) would be less clear.

Those cases aside, I hope I caught all the other instances of VLA in the code.

### DIFF
--- a/darwin/PlatformHelpers.c
+++ b/darwin/PlatformHelpers.c
@@ -103,9 +103,8 @@ double Platform_calculateNanosecondsPerMachTick(void) {
     *    the "Apple M1" chip specifically when running under Rosetta 2.
     */
 
-   size_t cpuBrandStringSize = 1024;
-   char cpuBrandString[cpuBrandStringSize];
-   Platform_getCPUBrandString(cpuBrandString, cpuBrandStringSize);
+   char cpuBrandString[1024] = "";
+   Platform_getCPUBrandString(cpuBrandString, sizeof(cpuBrandString));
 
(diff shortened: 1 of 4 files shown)


darwin/PlatformHelpers.c
```diff
@@ -103,9 +103,8 @@ double Platform_calculateNanosecondsPerMachTick(void) {
     *    the "Apple M1" chip specifically when running under Rosetta 2.
     */
 
-   size_t cpuBrandStringSize = 1024;
-   char cpuBrandString[cpuBrandStringSize];
-   Platform_getCPUBrandString(cpuBrandString, cpuBrandStringSize);
+   char cpuBrandString[1024] = "";
+   Platform_getCPUBrandString(cpuBrandString, sizeof(cpuBrandString));
```
Explorer09, line 107: Technically not VLA but this change helps readability anyway :)

BenBE: Um, actually … ;-) Although this uses a constant number (which the compiler can deduce), syntactically it <s>is</s>was.

BTW: I did a `grep --color=always -nRP '\b(?!return)\w+ \w+\[[^\]0-9A-Z_]'|less -R` through the source and had a look at the remaining locations found by this. Quite a nice way to quickly get through the source without straight up working with the AST …

natoscott (approved):
LGTM.

fasterit (approved):
looks good, thanks for the housekeeping!

PR #1275, opened 2023-08-02
BSD platforms: Reject negative and zero KERN_FSCALE factors

Explorer09:
The "fscale" value, retrieved by `sysctl()` in BSD platforms, is used for computing CPU percentages of the processes. To prevent a division by zero, we should reject a zero "fscale" value. (A negative "fscale" value will not make sense either.)

For DragonFlyBSD and FreeBSD, this would fall back to the hard-coded default scale.
For NetBSD and OpenBSD, there is no hard-coded default value, so the zero or negative "fscale" is now a fatal error.


dragonflybsd/DragonFlyBSDMachine.c
```diff
@@ -105,7 +105,7 @@ Machine* Machine_new(UsersTable* usersTable, uid_t userId) {
    }
 
    len = sizeof(kernelFScale);
-   if (sysctlbyname("kern.fscale", &kernelFScale, &len, NULL, 0) == -1) {
+   if (sysctlbyname("kern.fscale", &kernelFScale, &len, NULL, 0) == -1 || kernelFScale <= 0) {
```
Explorer09, line 108: Note: I have no tools to test if this builds in DragonFlyBSD, but I suspect there is a `this->` specifier missing. It should have been `this->kernelFScale` to prevent compile error.

BenBE: Good catch. Based on Looking at the Code™ it's undeniable, this could never compile. Please go ahead and fix this.

BenBE (approved):
LGTM.

Issue #1187, opened 2023-02-12
Show Network Speeds in Kbps / Mbps instead of KBps / MBps


marcoczen:
Guys, is there any setting that we can change to show speeds in Kbps / Mbps ?

Thanks.

BenBE:
Currently there isn't any such setting.

marcoczen:
Thank u @BenBE. 

Is there anything that i can edit in the code itself ? Are u one of the developers ?

eworm-de:
To get this right... This issue is about having **bits** as the base, not **bytes**?

marcoczen:
Is this the right file -> https://github.com/htop-dev/htop/blob/main/NetworkIOMeter.c

BenBE:
> Is this the right file -> https://github.com/htop-dev/htop/blob/main/NetworkIOMeter.c Lines 81 , 91 ?

The file is correct, but a better approach would be just adjusting the display, which affects lines [115ff](https://github.com/htop-dev/htop/blob/main/NetworkIOMeter.c#L115-L117), [140ff](https://github.com/htop-dev/htop/blob/main/NetworkIOMeter.c#L140-L142) and [145ff](https://github.com/htop-dev/htop/blob/main/NetworkIOMeter.c#L145-L147) (If I didn't miss any right now).

> Is there anything that i can edit in the code itself ?

Have a look at the three places above. Basically would need to multiply the values passed to `Meter_humanUnit` (2nd argument) by 8 and change the unit string from `i/Bs` to `ibps`.

To have a full setting, there's also some more changes needed in [`Settings.c`](https://github.com/htop-dev/htop/blob/main/Settings.c) and [`DisplayOptionsPanel.c`](https://github.com/htop-dev/htop/blob/main/DisplayOptionsPanel.c), as can be seen in [this recent patch](https://github.com/htop-dev/htop/commit/623ee31cfe3a39c87e6066c3044c7c38ce676e6b) (The stuff in `Process.c`/`Process.h` are the actual implementation details of that feature).

> Are u one of the developers ?

Yes. You can find this [here](https://github.com/orgs/htop-dev/people).

> > To get this right... This issue is about having **bits** as the base, not **bytes**?
> 
[...]

PR #1279, opened 2023-08-08
Process_printBytes(): fix XX.XG and XX.XT byte fields.

xavierog:
htop was liable to display erroneous sizes. For instance, a program consuming 85.00 TiB (93.46 TB) of virtual memory used to be reported as consuming "87.0T".

### DIFF
--- a/Process.c
+++ b/Process.c
@@ -111,12 +111,11 @@ void Process_printBytes(RichString* str, unsigned long long number, bool colorin
       number %= 1000;
       len = xSnprintf(buffer, sizeof(buffer), "%03lluM ", number);
       RichString_appendnAscii(str, processMegabytesColor, buffer, len);
-   } else if (number < 100000 * ONE_K) {
+   } else if (number < 100 * ONE_M) {
       //2 digit GB, 1 digit MB
-      number /= 100 * ONE_K;
-      len = xSnprintf(buffer, sizeof(buffer), "%2llu", number / 10);
+      len = xSnprintf(buffer, sizeof(buffer), "%2llu", number / ONE_M);
       RichString_appendnAscii(str, processGigabytesColor, buffer, len);
-      number %= 10;
+      number = (number % ONE_M) * 10 / ONE_M;
       len = xSnprintf(buffer, sizeof(buffer), ".%1llu", number);
       RichString_appendnAscii(str, processMegabytesColor, buffer, len);
       RichString_appendAscii(str, processGigabytesColor, "G ");
@@ -133,12 +132,11 @@ void Process_printBytes(RichString* str, unsigned long long number, bool colorin
       number %= 1000;
       len = xSnprintf(buffer, sizeof(buffer), "%03lluG ", number);
       RichString_appendnAscii(str, processGigabytesColor, buffer, len);
-   } else if (number < 100000 * ONE_M) {
+   } else if (number < 100ULL * ONE_G) {
       //2 digit TB, 1 digit GB
-      number /= 100 * ONE_M;
-      len = xSnprintf(buffer, sizeof(buffer), "%2llu", number / 10);
+      len = xSnprintf(buffer, sizeof(buffer), "%2llu", number / ONE_G);
       RichString_appendnAscii(str, largeNumberColor, buffer, len);
-      number %= 10;
+      number = (number % ONE_G) * 10 / ONE_G;
       len = xSnprintf(buffer, sizeof(buffer), ".%1llu", number);
       RichString_appendnAscii(str, processGigabytesColor, buffer, len);
       RichString_appendAscii(str, largeNumberColor, "T ");


Explorer09:
Division by 102.4 would create unnecessary rounding errors. I know what you are trying to do, but I think there is way to avoid float divisions here.

Explorer09:
Update: I've thought of a better idea:
```c
   } else if (number < 100 * ONE_M) {
      //2 digit GB, 1 digit MB
      len = xSnprintf(buffer, sizeof(buffer), "%2llu", number / ONE_M);
      RichString_appendnAscii(str, processGigabytesColor, buffer, len);
      number = (number % ONE_M) * 10 / ONE_M;
      len = xSnprintf(buffer, sizeof(buffer), ".%1llu", number);
      RichString_appendnAscii(str, processMegabytesColor, buffer, len);
      RichString_appendAscii(str, processGigabytesColor, "G ");
   } // ...
```

xavierog:
Considering the small number of decimals we keep for formatting purposes and the restricted range of affected values, I doubt any rounding error could show up in htop, save for the explicit truncation that happens when we compute the decimals (e.g. 93897648766976 bytes is 85.3994140625 TiB and will end up as "85.3T" instead of "85.4T") -- this truncation does not strike me as a big deal though.

That being said, I like the idea of relying exclusively on integer operations -- does the htop project have a clear written policy about this somewhere? It probably deserves a word in docs/styleguide.md.
Anyway, I have applied your suggestion and am about to push the modifications.

Explorer09:
@xavierog What I saw is that 102.4 cannot be accurately represented in binary floating points, thus the rounding error would be present on the divisor constant, slowly propagating during the division operation.
Even `x = x * 10 / (1024 * ONE_K)` would work better than `x /= (102.4 * ONE_K)`, by my idea `x / ONE_K, (x % ONE_K) * 10 / ONE_K` could optimize a little bit better when the compiler can turn "divisions by a power of 2" to bit shifts.

xavierog:
I concur, 102.4 × 1024 = 104857.6 (i.e. the value that ends up in the executable's .rodata section) cannot be accurately represented (though the 64-bit approximation is quite decent):

- https://www.binaryconvert.com/result_float.html?decimal=049048052056053055046054
- https://www.binaryconvert.com/result_double.html?decimal=049048052056053055046054

Do you think the new code can be merged?

BenBE:
One goal when this implementation was revised was to avoid jumps in the numbers when values increase or decrease. The multiplication by ONE_K comes from the way the value is scaled internally (in KiB). Display of sizes in htop all follow IEC units, not SI (despite network people liking their SI units for bandwidths, cf. #1187).

xavierog:
> One goal when this implementation was revised was to avoid jumps in the numbers when values increase or decrease.

Can you elaborate with an example?

fasterit (approved):
Approved. The better rounding also makes `htop` report VIRT memory use closer to `top`.

BenBE:
> > One goal when this implementation was revised was to avoid jumps in the numbers when values increase or decrease.
> 
> Can you elaborate with an example?

9999K -> 10000K shouldn't jump to 8.9M (exaggerated).

It's not quite unavoidable with the IEC units, but it should be tried to minimize such effects.

Explorer09:
@BenBE
No, it won't happen. With IEC unit prefixes, there are only these "jumps" in htop:
100000 KiB (97.6 MiB, would display as `97M`)
10000 MiB (9.76 GiB, would display as `9.7G`)
10000 GiB (9.76 TiB, would display as `9.7T`)
And so on.

Maybe we can note this fact in the code comments.

Process.c
```diff
@@ -133,12 +132,11 @@ void Process_printBytes(RichString* str, unsigned long long number, bool colorin
       number %= 1000;
       len = xSnprintf(buffer, sizeof(buffer), "%03lluG ", number);
       RichString_appendnAscii(str, processGigabytesColor, buffer, len);
-   } else if (number < 100000 * ONE_M) {
+   } else if (number < 100 * ONE_G) {
```
Explorer09, line 135: Wait a minute. Do we need type casting here?
```suggestion
   } else if (number < 100ULL * ONE_G) {
```

xavierog: Likely, yes. I thought all ONE_x were ULL but it turns out ONE_K, ONE_M and ONE_G are UL whereas ONE_T and ONE_P are indeed ULL.

Explorer09:
I wonder if you guys are interested, but as a coding exercise, I made a simplified version of `Process_printBytes()` function.
https://github.com/Explorer09/htop-1/commits/print-bytes-improve
This branch is experimental.

fasterit:
> I wonder if you guys are interested, but as a coding exercise, I made a simplified version of `Process_printBytes()` function. https://github.com/Explorer09/htop-1/commits/print-bytes-improve This branch is experimental.

I like that. It looks much more readable.
/DLange

BenBE (approved):

xavierog:
I have integrated the missing "ULL" spotted by @Explorer09 .
His rewrite of the entire function is interesting. Imho, this calls for another, dedicated PR that could override this one.

Explorer09:
My code is intended to be applied after Xavier's patch. It's my attempt to shrink the code without change of the function's behavior. It's not done yet, as I have some more ideas to shrink the code further.

fasterit:
> My code is intended to be applied after Xavier's patch. It's my attempt to shrink the code without change of the function's behavior. It's not done yet, as I have some more ideas to shrink the code further.

Ok, I'll merge @xavierog's work. Please continue to prepare your cleanup and make this into a separate PR.

PR #1282, opened 2023-08-15
For supporting older than MAC OSX 12 system

jackzzjack:
Suppress "note: enclose 'IOMainPort' in a __builtin_available check to silence this warning" messages. The IOMainPort is not supported under older OS with version 12.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -21,7 +21,7 @@ AM_INIT_AUTOMAKE([-Wall std-options subdir-objects])
 # ----------------------------------------------------------------------
 # Checks for platform.
 # ----------------------------------------------------------------------
-
+darwin_version=""
 case "$host_os" in
 linux*|gnu*)
    my_htop_platform=linux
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -46,6 +46,7 @@ dragonfly*)
 darwin*)
    my_htop_platform=darwin
    AC_DEFINE([HTOP_DARWIN], [], [Building for Darwin.])
+   darwin_version=$(sw_vers | grep ProductVersion | awk '{ print $2 }' | awk -F . '{ print $1 }')
```
fasterit, line 49: darwin_version=$(sw_vers -productVersion | cut -d . -f 1)

Explorer09:
I don't get it. Why is adding a configure check able to silence a compiler warning?

jackzzjack:
Sorry for misunderstanding, let me explain more specific. IOMainPort is not only shown with warning messages, but also cannot be executed with compiled binary.
When I try to execute the binary with IOMainPort included, it shows  “dyld: lazy symbol binding failed” on runtime.

Explorer09:
Then, what's wrong with the current configure check for `IOMainPort`?
Checking the version number of macOS is a bad idea because the version number reported might not accurately indicate whether a particular API is available. Configure scripts are generally written to avoid checking version numbers but to test the API availability directly.

Issue #631, opened 2021-05-18
M1 Mac Ram Usage Incorrect for Application Memory (System Ram, and M1 GPU Ram)


sigkill:
Might not be exclusive to the M1.

htop version installed via homebrew:
 htop 3.0.5

I'm not sure of the ramifications of this, but it seems as if disabling the hardware acceleration in Google Chrome free'd up ram that was not being shown as in use by htop, but was shown in use by the Apple "Activity Monitor" program.
![image](https://user-images.githubusercontent.com/634740/118696692-696a1e00-b7c3-11eb-97e5-4fa5db6d2169.png)

Upon disabling hardware acceleration, I was able to get some ram back for my Application Memory problem, but then noticed that htop still seems to think the available ram is as follows:

![image](https://user-images.githubusercontent.com/634740/118697554-47bd6680-b7c4-11eb-81a0-6e0260e0c3df.png)

![image](https://user-images.githubusercontent.com/634740/118697643-5efc5400-b7c4-11eb-9992-d6d6947fa2cb.png)

The Application Memory killer popped up, and I looked at htop and found a nearly 4GB inconsistency from what htop had been reporting.

The weird part is that the memory was consumed, and I'm not sure on a hardware level if the GPU shares the system ram, but the Activity Monitor did show the memory being used at the time, I've replicated by opening a huge amount of tabs and re-enabling hardware acceleration here:

![image](https://user-images.githubusercontent.com/634740/118699252-2a899780-b7c6-11eb-9618-c83673f3c110.png)

I suspect this bug would, if it is a bug, show up on the new iMac systems as well.

Will compare to my Intel based system later today to see if the reporting is off there too.

sigkill:
Forgot to provide that I am running Big Sur 11.3 and 11.4 Beta (Both have the same issue), and have swap disabled.

cgzones:
The code in question is:
https://github.com/htop-dev/htop/blob/69cfaf238101c8d701176f734ad0371f4839972c/darwin/Platform.c#L240-L251

htop calculates the **used** memory as `active_count + wire_count` and the **cached** memory as `inactive_count`.
Probably the Activity Monitor treats part of the `inactive_count` as used memory, so its used memory is higher; cause the total amount of occupied memory (13.00GB + 2.98GB vs 1 space to the right of the yellow bar) seems equal.

sigkill:
[...]

PR #1285, opened 2023-08-24
Fix Remove M1 limitation

yngty:
Special handling is also required on M1Pro, M2, M2 Pro

### DIFF
--- a/darwin/PlatformHelpers.c
+++ b/darwin/PlatformHelpers.c
@@ -103,15 +103,12 @@ double Platform_calculateNanosecondsPerMachTick(void) {
     *    the "Apple M1" chip specifically when running under Rosetta 2.
     */
 
-   char cpuBrandString[1024] = "";
-   Platform_getCPUBrandString(cpuBrandString, sizeof(cpuBrandString));
-
    bool isRunningUnderRosetta2 = Platform_isRunningTranslated();
 
    // Kernel version 20.0.0 is macOS 11.0 (Big Sur)
    bool isBuggedVersion = Platform_KernelVersionIsBetween((KernelVersion) {20, 0, 0}, (KernelVersion) {999, 999, 999});
 
-   if (isRunningUnderRosetta2 && String_eq(cpuBrandString, "Apple M1") && isBuggedVersion) {
+   if (isRunningUnderRosetta2 && isBuggedVersion) {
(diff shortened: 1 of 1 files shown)


Explorer09:
I don't think this fix is correct, since you are not addressing whether Intel processor Macs need the special handling.

yngty:
> I don't think this fix is correct, since you are not addressing whether Intel processor Macs need the special handling.我不认为这个修复是正确的，因为您没有解决英特尔处理器Mac是否需要特殊处理。

yngty:
> I don't think this fix is correct, since you are not addressing whether Intel processor Macs need the special handling.我不认为这个修复是正确的，因为您没有解决英特尔处理器Mac是否需要特殊处理。

Intel Processor Mac will not enter isRunningUnderRosetta2 case

darwin/PlatformHelpers.c
```diff
@@ -95,23 +95,12 @@ double Platform_calculateNanosecondsPerMachTick(void) {
 #else
    mach_timebase_info_data_t info;
 
-   /* WORKAROUND for `mach_timebase_info` giving incorrect values on M1 under Rosetta 2.
-    *    rdar://FB9546856 https://openradar.appspot.com/radar?id=5055988478509056
-    *
-    *    We don't know exactly what feature/attribute of the M1 chip causes this mistake under Rosetta 2.
-    *    Until we have more Apple ARM chips to compare against, the best we can do is special-case
-    *    the "Apple M1" chip specifically when running under Rosetta 2.
-    */
```
Explorer09, line 104: I have one small advice. Don't remove the original code comment as it would help other programmers understand why the workaround is needed.
（一個小建議，不要隨便移除原程式碼的註解，因為它可以幫助其他程式員快速了解這個workaround出現的原因。）

yngty:
I've changed it.

SuCicada:
yes, I think this fix is necessary. This fix can be used to distinguish between **host_statistics** and **host_statistics64**.

**host_statistics64** is necessary at Apple Silicon to show right memory

There is a  related issue: https://github.com/htop-dev/htop/issues/631

https://github.com/htop-dev/htop/blob/599233ffdf1783011276145bd6933a1fd46fd932/darwin/DarwinMachine.c#L62-L68

PR #1286, opened 2023-08-26
Rework ZramMeter and remove MeterClass.comprisedValues

Explorer09:
The `comprisedValues` boolean property unnecessarily complicates the drawing algorithms of Bar meters and Graph meters. Since the only user of `comprisedValues` is ZramMeter, it is better to rework the meter so that it no longer needs `comprisedValues`.

The `values[ZRAM_METER_UNCOMPRESSED]` now stores the difference between uncompressed and compressed data size.

### DIFF
--- a/Meter.c
+++ b/Meter.c
@@ -11,9 +11,9 @@ in the source distribution for its full text.
 
 #include <assert.h>
 #include <math.h>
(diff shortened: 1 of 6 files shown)


BenBE (approved):
LGTM.

Issue #1274, opened 2023-08-01
Inconsistent behaviors of computing CPU% when time "delta" is zero


Explorer09:
I originally intended to address this issue in pull request #1271 before I found out the issue is too big to handle in a single commit.

The CPU percentages of the machine and processes are usually computed through a formula like this:
(float) (newUsedTime - oldUsedTime) / (newElapsedTime - oldElapsedTime) * 100.0
The denominator represents a difference between the current time and the timestamp of last measurement. However, when this time difference is zero, platform codes of htop handle them differently, and some of them are a bit buggy.

Here is a summary of the behaviors of different htop platform codes:
* DragonFlyBSD & FreeBSD
  * When delta time is zero, the machine's CPU percentage is computed as if delta time is one.
  * A process's CPU time is reported in percentage, but it requires a division by a fixed scale (`kernelFScale`) which is not checked for zero.
* NetBSD & OpenBSD
  * When delta time is zero, the machine's CPU percentage is computed as if delta time is one.
  * A process's CPU time is reported in percentage. If `fscale` is zero, the percentage is set to zero.
* Darwin
  * The machine's CPU percentage may be infinity or NaN (division by zero).
  * When delta time is zero, the process's CPU percentage is set to zero.
* Linux
  * When delta time is zero, the machine's CPU percentage is computed as if delta time is one.
  * When delta time is zero, the process's CPU percentage is set to zero.
* PCP
  * When delta time is zero, the machine's CPU percentage is computed as if delta time is one.
  * When delta time is zero, the process's CPU percentage may be either 100 or 0 due to clamping. Division by zero could happen, but the NaN value would be caught by the `isnan` check.
* Solaris
  * The machine's CPU percentage may be infinity or NaN (division by zero).
  * A process's CPU time is reported in percentage. The scale to divide is hard coded so no division by zero could happen.

Source code references (I use commit 2d4e5cba225439217464938bf7eaed277427e12d from `main` as a base; for convenience [this page](https://github.com/Explorer09/htop-1/commit/a65794c2d6f451c53f4e5a08f936f6c4222da560) marks all lines that are referenced):
```
dragonflybsd/DragonFlyBSDMachine.c line 204
dragonflybsd/DragonFlyBSDProcessList.c line 217
dragonflybsd/DragonFlyBSDMachine.c line 108
freebsd/FreeBSDMachine.c line 231
freebsd/FreeBSDProcessList.c line 238
freebsd/FreeBSDMachine.c line 139
[...]

PR #1288, opened 2023-08-29
Linux: add GPU meter and process column

cgzones:
Based on the DRM client usage stats[1] add statistics for GPU time spend
and percentage utilization.

[1]: https://www.kernel.org/doc/html/latest/gpu/drm-usage-stats.html

Tested with i915 and amdgpu.


BenBE:
Does this recognize/work with multi-GPU setups?

linux/GPU.c
```diff
@@ -0,0 +1,272 @@
...
+   struct client_id* next;
+};
+
+enum section_state {
+   SECST_UNKNOWN,
+   SECST_DUPLICATE,
+   SECST_NEW,
+};
+
+static bool null_str_eq(const char* s1, const char* s2) {
```
BenBE, line 36: Why not add this to `XUtils.h`?

linux/GPU.c
```diff
@@ -0,0 +1,272 @@
...
+#include <dirent.h>
+#include <errno.h>
+#include <sys/types.h>
+
+#include "XUtils.h"
+
+#include "linux/LinuxMachine.h"
+
+
+#define INVALID_CLIENT_ID ((unsigned long long int)-1)
```
BenBE, line 21: What about having a `typedef` for the client type?

linux/GPU.c
```diff
@@ -0,0 +1,272 @@
...
+      *newData = (GPUEngineData) {
+         .prevTime = 0,
+         .curTime  = 0,
+         .key      = engine,
+         .next     = NULL,
+      };
+
+      *engineData = newData;
+   } else
+      free(engine);
```
BenBE, line 81:
```suggestion
   } else {
      free(engine);
   }
```

linux/GPU.c
```diff
@@ -0,0 +1,272 @@
...
+      if (String_eq((*engineData)->key, engine))
+         break;
+
+      engineData = &((*engineData)->next);
+   }
+
+   if (!*engineData) {
+      GPUEngineData* newData;
+
+      newData = xMalloc(sizeof(*newData));
```
BenBE, line 71:
```suggestion
      GPUEngineData* newData = xMalloc(sizeof(*newData));
```

linux/GPU.c
```diff
@@ -0,0 +1,272 @@
...
+
+         if (String_startsWith(buf, "drm-client-id:")) {
+            char *endptr;
+
+            if (sstate == SECST_NEW) {
+               struct client_id* new;
+
+               assert(client_id != INVALID_CLIENT_ID);
+
+               new = xMalloc(sizeof(*new));
```
BenBE, line 152:
```suggestion
               assert(client_id != INVALID_CLIENT_ID);

               struct client_id* new = xMalloc(sizeof(*new));
```

linux/GPU.c
```diff
@@ -0,0 +1,272 @@
...
+      } /* finished parsing lines */
+
+      fclose(fp);
+
+      if (sstate == SECST_NEW) {
+         struct client_id* new;
+
+         assert(client_id != INVALID_CLIENT_ID);
+
+         new = xMalloc(sizeof(*new));
```
BenBE, line 225:
```suggestion
         assert(client_id != INVALID_CLIENT_ID);

         struct client_id* new = xMalloc(sizeof(*new));
```

linux/GPU.c
```diff
@@ -0,0 +1,272 @@
...
+      entry = readdir(fdinfoDir);
+      if (!entry)
+         break;
+      ename = entry->d_name;
+
+      if (String_eq(ename, ".") ||
+          String_eq(ename, "..") ||
+          String_eq(ename, "0") ||
+          String_eq(ename, "1") ||
+          String_eq(ename, "2"))
```
BenBE, line 129: Why exclude `stdin`, `stdout` and `stderr` explicitly?

linux/GPU.c
```diff
@@ -0,0 +1,272 @@
...
+#endif
+
+   while (true) {
+      const struct dirent* entry;
+      const char* ename;
+      FILE* fp;
+      char* pdev = NULL;
+      unsigned long long int client_id = INVALID_CLIENT_ID;
+      enum section_state sstate = SECST_UNKNOWN;
+      char buf[256];
```
BenBE, line 118: We're not kernel code. We _CAN_ declare our variables where we need them …

In particular: Please don't leave variables uninitialized. That's just causing PITA down the line.

linux/GPU.c
```diff
@@ -0,0 +1,272 @@
...
+
+#ifdef HAVE_OPENAT
+      fp = Compat_fopenat(dirfd(fdinfoDir), ename, "r");
+#else
+      fp = Compat_fopenat(fdinfoPathBuf, ename, "r");
+#endif
+      if (!fp)
+         continue;
+
+      while (fgets(buf, sizeof(buf), fp)) {
```
BenBE, line 140: I'd prefer to avoid the fopen style APIs in favour of reading a sensible chunk of the file at once (we've got some API for that).

linux/GPU.c
```diff
@@ -0,0 +1,272 @@
...
+            char* q;
+
+            p = buf + strlen("drm-pdev:");
+
+            while (isspace((unsigned char)*p))
+               p++;
+
+            q = strchr(p, '\n');
+            if (q)
+               *q = '\0';
```
BenBE, line 180:
```suggestion
            const char* p = buf + strlen("drm-pdev:");

            while (isspace((unsigned char)*p))
               p++;

            char* q = strchr(p, '\n');
            if (q)
               *q = '\0';
```

linux/GPU.c
```diff
@@ -0,0 +1,272 @@
...
+
+            assert(!pdev || String_eq(pdev, p));
+            if (!pdev)
+               pdev = xStrdup(p);
+         } else if (String_startsWith(buf, "drm-engine-")) {
+            const char* delim;
+            const char* engineStart;
+            char* endptr;
+            unsigned long long int value;
```
BenBE, line 189: Please initialize variables and in particular pointers …

linux/GPU.h
```diff
@@ -0,0 +1,16 @@
+#ifndef HEADER_GPU
+#define HEADER_GPU
+/*
+htop - GPU.h
+(C) 2023 htop dev team
+Released under the GNU GPLv2+, see the COPYING file
+in the source distribution for its full text.
+*/
+
+#include "linux/LinuxProcess.h"
```
BenBE, line 10: Should include `Compat.h` for `openat_arg_t`.

linux/GPUMeter.h
```diff
@@ -0,0 +1,17 @@
+#ifndef HEADER_GPUMeter
+#define HEADER_GPUMeter
+/*
+htop - GPUMeter.h
+(C) 2023 htop dev team
+Released under the GNU GPLv2+, see the COPYING file
+in the source distribution for its full text.
+*/
+
+#include "Meter.h"
```
BenBE, line 10: Missing `stdbool.h`

BenBE (changes requested):
Using the `fopen` style API is hard to transform into liburing usage later on, thus try to avoid it. We also have a function `xReadfile` and `xReadfileat` to read a whole file at once into a appropriately sized buffer. For `fdinfo` files this should be sufficient and avoid the overhead of the `fopen` API.

linux/GPU.c
```diff
@@ -0,0 +1,272 @@
...
+ */
+void GPU_readProcessData(LinuxProcessList* lpl, LinuxProcess* lp, openat_arg_t procFd) {
+   int fdinfoFd = -1;
+   DIR* fdinfoDir = NULL;
+   struct client_id* parsed_ids = NULL;
+   unsigned long long int new_gpu_time = 0;
+
+   fdinfoFd = Compat_openat(procFd, "fdinfo", O_RDONLY | O_NOFOLLOW | O_DIRECTORY | O_CLOEXEC);
+   if (fdinfoFd == -1)
+     goto out;
```
Explorer09, line 99:
```suggestion
      goto out;
```

linux/GPU.c
```diff
@@ -0,0 +1,272 @@
...
+      uint64_t monotonictimeDelta;
+
+      Process_updateFieldWidth(GPU_TIME, ceil(log10(new_gpu_time)));
+
+      gputimeDelta = saturatingSub(new_gpu_time, lp->gpu_time);
+      monotonictimeDelta = host->monotonicMs - host->prevMonotonicMs;
+      if (gputimeDelta == 0)
+         lp->gpu_percent = 0.0f;
+      else if (monotonictimeDelta == 0)
+         lp->gpu_percent = -1.0f;
```
Explorer09, line 251: Perhaps I should mention issue #1274 with this part of code.

linux/GPU.c
```diff
@@ -0,0 +1,272 @@
...
+      fp = Compat_fopenat(fdinfoPathBuf, ename, "r");
+#endif
+      if (!fp)
+         continue;
+
+      while (fgets(buf, sizeof(buf), fp)) {
+         if (!String_startsWith(buf, "drm-"))
+            continue;
+
+         if (String_startsWith(buf, "drm-client-id:")) {
```
Explorer09, line 144: A chain of `String_startsWith` conditionals doesn't look efficient.
Since the token strings you are matching all begin with "drm-", how about using a switch statement?

```c
switch (buf[strlen("drm-")]) {
case 'c':
   if (String_startsWith(buf, "drm-client-id:")) {
      ...
   }
   break;

case 'p':
   if (String_startsWith(buf, "drm-pdev:")) {
      ...
   }
   break;

case 'e':
   if (String_startsWith(buf, "drm-engine:")) {
      ...
   }
   break;
}
```

linux/GPU.c
Explorer09: Or better:
```suggestion
      *engineData = xMalloc(sizeof(**engineData));
```

linux/GPU.c
```diff
@@ -0,0 +1,272 @@
...
+   for (; parsed; parsed = parsed->next) {
+      if (id == parsed->id && null_str_eq(pdev, parsed->pdev)) {
+         return true;
+      }
+   }
+
+   return false;
+}
+
+static void update_system(LinuxProcessList* lpl, unsigned long long int time, char* engine) {
```
Explorer09, line 56: This API can be changed so you don't need to always duplicate the string when calling it
```c
// Current
static void update_system(LinuxProcessList* lpl, unsigned long long int time, char* engine);
// Suggestion
static void update_system(LinuxProcessList* lpl, unsigned long long int time, const char* engine, size_t engine_len);
```

Explorer09: By the way, the name `update_system` doesn't look like what this function does. Perhaps you mean `AddGPUEngineData`?

linux/GPU.c
natoscott: @cgzones monotonictimeDelta can never be zero here, so the 'else if' branch is dead code - please remove.

linux/GPU.c
natoscott: Yeah - had the same thought, I think this little optimization is not always valid unfortunately (e.g. a program closes one of these three at some point, as happens in some daemons, and then they later get reused).

natoscott (changes requested):

linux/GPU.c
Explorer09: @natoscott Place an assert statement there so that monotonictimeDelta can never be zero.

natoscott: The nature of time and space is what makes the value never zero, not an assert.  :)   But seriously, I'd not object to an assert if it helps others feel at ease, even though personally I think its just noise in the code.

Reminds me of the discussion a few years back about asserting on a non-NULL pointer right before we use it ... which some people like to do, but again personally I feel its just noise in the code since both debug and non-debug builds will inform you immediately and clearly if that ever happens (hence the assert adds no value IMO).

Explorer09: @natoscott The assert statement is to prevent bugs. 🙂
Even the variable is named `monotonicTime`, programmers can still retrieve the same cached value twice (usually done accidentally). There is no guarantee that the value would always be different unless they call `clock_gettime` or similar API every time.

natoscott: It doesn't prevent anything - the bug would still happen and if there's no assert, we divide by zero (in both debug *and* non-debug builds mind you) and stop on the exact same line of code.  As with the non-null-pointer-assert, its just line noise in the code IMO.

But as I said, I understand it makes some people feel better - so please go right ahead with an assert if you feel its useful.

natoscott: You're also kinda pointing out the problem with #1274 too - it would be hiding such bugs.  If they happen, we really want to know about them, not just carry on execution as if nothing happened.

Explorer09:
I wish the `autoTitleRightAlign` flag be split to a separate PR. The reason is simply I don't like the current design of the flag, and I think it deserves more discussion before introducing it into the codebase.

linux/GPU.c
cgzones: client ids are only used inside this one file, so It think a typedef is not needed

linux/GPU.c
cgzones: The current code is more readable for me, e.g. it avoids ` **engineData = ...`

linux/GPU.c
cgzones: But we are also using C, so we need to handle resource cleanup ourselves (no destructors).
Also modern compilers are quite good at spotting use of uninitialized variables at function scope.

linux/GPU.c
cgzones: added an assert in ./Machine.c

linux/GPU.c
cgzones: This reuse of standard file descriptors is dangerous, as any used library could write messages to stderr, corrupting any file opened at fileno 2. Do we really need to take care of such programs, which are mostly security related ones?

linux/LinuxProcess.c
```diff
@@ -231,6 +233,8 @@ static void LinuxProcess_rowWriteField(const Row* super, RichString* str, Proces
    switch (field) {
    case CMINFLT: Row_printCount(str, lp->cminflt, coloring); return;
    case CMAJFLT: Row_printCount(str, lp->cmajflt, coloring); return;
+   case GPU_PERCENT: Row_printPercentage(lp->gpu_percent, buffer, n, 5, &attr); break;
+   case GPU_TIME: xSnprintf(buffer, n, "%*llu", Row_fieldWidths[GPU_TIME], lp->gpu_time); break;
```
Explorer09, line 237: Do we really need to print the GPU time with that much precision?
I was thinking if the `humanTimeUnit()` format can also apply here. Maybe this printing format could become a new function called `Row_printNanosecends()`.

linux/GPU.c
```diff
@@ -0,0 +1,256 @@
...
+         parsed_ids = new;
+      }
+
+      free(pdev);
+   } /* finished parsing fdinfo entries */
+
+   if (new_gpu_time > 0) {
+      const Machine* host = lp->super.super.host;
+      unsigned long long int gputimeDelta;
+      uint64_t monotonictimeDelta;
```
Explorer09, line 231: A small nitpick on the naming: `monotonictimeDelta` or `monotonicTimeDelta`, which is more readable?

linux/GPU.c
natoscott: Any library assuming it can write to stderr is dangerous.  :)  Its more common for libraries to use return codes to indicate failure (along with a strerror-style interface for custom errors).

But, maybe we're lucky and perhaps it is indeed the case that there's little overlap between the set of programs that close standard file descriptors and the set of programs that use GPUs (I expect the latter is mostly games or other interactive visual programs and AI/ML programs).

BenBE: It's bad practice for any library to assume any fd it didn't open is available. In particular writing to stdout and stderr is frowned upon.

Also including these fd is a matter of correctnes, while excluding the doesn't provide enough advantage to warrant the downsides.

linux/GPU.c
BenBE: I find ulli hard to read and lacking of context, which a typedef, even local to this file, could provide.

linux/GPU.c
BenBE: Potential for UB vs. performance hit (dup init, which compilers are good at removing) vs. often less (more concise) code

linux/GPU.c
natoscott: | [...] doesn't provide enough advantage to warrant the downsides.

Hmm - I expect there's a noticeable performance hit from opening/reading/closing these three files for every running process on a system, on every sample, for ~zero value.  I certainly understand why @cgzones went this path and maybe we should see if we can get away with it, since I suspect its unlikely to cause issues in practice.

BenBE: What about a different approach entirely:

Instead of skipping individual files, just mark in the process object if there were no files at all with any references for the DRM open. If so, skip such processes for a certain time (like ~5-10 seconds) from further scans. That way you reduce the amount of work to usually about 3-5% (quick local test gave 13 of ~390 processes). That's much more savings right away, with the only down-side that first GPU activity might be detected with a short delay (if not seen on first scan of a process). To contrast this: When skipping FDs 0, 1 and 2 you end up not scanning 5-10% (~380 FDs of 6440). Or written differently: That's doing still about 90% of work, where-as skipping processes without DRM access has about 5% of the work left.

linux/LinuxProcess.c
```diff
@@ -407,6 +411,17 @@ static int LinuxProcess_compareByKey(const Process* v1, const Process* v2, Proce
       return SPACESHIP_NUMBER(p1->autogroup_id, p2->autogroup_id);
    case AUTOGROUP_NICE:
       return SPACESHIP_NUMBER(p1->autogroup_nice, p2->autogroup_nice);
+   case GPU_PERCENT: {
+      int r;
+
+      r = compareRealNumbers(p1->gpu_percent, p2->gpu_percent);
```
BenBE, line 417:
```suggestion
      int r = compareRealNumbers(p1->gpu_percent, p2->gpu_percent);
```

linux/LinuxProcessTable.c
```diff
@@ -1559,6 +1561,13 @@ static bool LinuxProcessTable_recurseProcTree(LinuxProcessTable* this, openat_ar
       }
       #endif
 
+      if (ss->flags & PROCESS_FLAG_LINUX_GPU || GPUMeter_active()) {
+         if (parent)
+            lp->gpu_time = ((const LinuxProcess*)parent)->gpu_time;
+         else
+            GPU_readProcessData(this, lp, procFd);
```
BenBE, line 1568:
```suggestion
         if (parent) {
            lp->gpu_time = ((const LinuxProcess*)parent)->gpu_time;
         } else {
            GPU_readProcessData(this, lp, procFd);
         }
```

linux/LinuxProcessTable.c
```diff
@@ -1630,6 +1639,19 @@ void ProcessTable_goThroughEntries(ProcessTable* super) {
...
+      GPUEngineData* engine;
+
+      lhost->prevGpuTime = lhost->curGpuTime;
+      lhost->curGpuTime = 0;
+
+      for (engine = lhost->gpuEngineData; engine; engine = engine->next) {
+         engine->prevTime = engine->curTime;
+         engine->curTime = 0;
+      }
+   }
```
BenBE, line 1653:
```suggestion
   {
      lhost->prevGpuTime = lhost->curGpuTime;
      lhost->curGpuTime = 0;

      for (GPUEngineData* engine = lhost->gpuEngineData; engine; engine = engine->next) {
         engine->prevTime = engine->curTime;
         engine->curTime = 0;
      }
   }
```

linux/GPU.c
```diff
@@ -0,0 +1,250 @@
...
+
+typedef unsigned long long int client_id_t;
+#define INVALID_CLIENT_ID ((client_id_t)-1)
+
+
+struct client_id {
+   char* pdev;
+   client_id_t id;
+   struct client_id* next;
+};
```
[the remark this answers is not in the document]
Explorer09: Uh. I'm not sure what's the context here. Why do we need a `typedef unsigned long long int client_id_t;` and a `struct client_id` definitions simultaneously?
Do these structures come from the kernel? If they don't come from the kernel, I would see the structures named like other structures in htop, i.e. in CamelCase.

linux/GPUMeter.c
```diff
@@ -0,0 +1,168 @@
...
+
+   value /= 60;
+
+   if (value < 96)
+      return snprintf(buffer, size, "%3lluh", value);
+
+   value /= 24;
+
+   return snprintf(buffer, size, "%3llud", value);
+}
```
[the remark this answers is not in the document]
Explorer09: I have some questions regarding this `humanTimeUnit()` function.
1. Is this meant for printing values in meters or data cells or both?
2. It seems that you are constraining the output to 5 characters (e.g. `999ns`). Is this intentional?
3. Why are there `600s`, `600m` and `96h` thresholds, in particular? As I rewrote the `Row_printTime()` function in PR #1325, I start to think that such thresholds in `humanTimeUnit()` don't make sense.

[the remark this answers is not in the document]
BenBE: As I had in the back of my mind, that there had been a PR regarding the time formatting code recently, I wrote that comment in #1321 so we can somehow focus these efforts.

For `600s` and `600m` the limits are kinda obvious with `10m` and `10h` respectively. For `96h` (AKA `4d`) the reason is likely to stay with whole days below a 3 digit value for hours.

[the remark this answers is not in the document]
BenBE: Possible alternative, mostly re-uses the ranges from your proposal.
```
  0ns - 999ns
1.0us - 9.9us
 10us - 999us
1.0ms - 999ms
 1.0s - 59.9s (Or 01.00 - 59.99 compared to the time column)
 1:00 - 59:59
 1h00 - 23h59
1d00h - 9d23h
10.0d - 99.9d
 100d -  364d
1y00m - 9y11m
10.0y - 99.9y
 100y - 9999y
  inf -   inf
```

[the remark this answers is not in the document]
Explorer09: @BenBE No, the colon is ambiguous where there are `hh:mm` and `mm:ss` notations.
`59m59` might be okay though.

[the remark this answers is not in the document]
BenBE: The colon would be consistent with existing formatting though. Yes, `59m59` would work, but it's not the highest option on my list …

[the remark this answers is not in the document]
cgzones: Used `xSnprintf()`.

Added formatting for X.Y(us|ms|s).

The input values are difference values to the prior scan so they should be lower than the scan interval (1.5s).

kryptobolt200528:
The GPU meter doesn't seem to work on my intel haswell series processor(Intel Celeron 2955u),at least in the way i expect it to,the gpu utilizationn is displayed to be zero for example when browsing the web whereas using intel's own 'intel-gpu-tools' one can notice the render engines to be in active use.

PR #1294, opened 2023-08-31
Rename ProcessList to ProcessTable throughout

natoscott:
Following up with some discusson from a few months back,
where it was proposed that ProcessTable is a better name.
This data structure is definitely not a list ... if it
was one-dimensional it'd be a set, but in practice it has
much more in common with a two-dimensional table.
 
The Process table is a familiar operating system concept
for many people too so it resonates a little in that way
as well.

In terms of functionality, this change is a no-op.


CommandLine.c
```diff
@@ -341,7 +341,7 @@ int CommandLine_run(int argc, char** argv) {
    Hashtable* ds = DynamicScreens_new();
 
    Machine* host = Machine_new(ut, flags.userId);
-   ProcessList* pl = ProcessList_new(host, flags.pidMatchList);
+   ProcessTable* pl = ProcessTable_new(host, flags.pidMatchList);
```
Explorer09, line 344: Maybe rename the local variable as well?
```suggestion
   ProcessTable* pt = ProcessTable_new(host, flags.pidMatchList);
```

natoscott: @Explorer09 good catch, thanks - will do.

PR #1295, opened 2023-08-31
RowField class / module improvements

Explorer09:


RowField.h
```diff
@@ -51,6 +51,6 @@ typedef enum ReservedFields_ {
 
 /* Follow ReservedField entries with dynamic fields defined at runtime */
 #define ROW_DYNAMIC_FIELDS LAST_RESERVED_FIELD
-typedef int32_t RowField;
+typedef int32_t FieldID;
```
natoscott, line 55: If we were going to do this, the filename here would have to change.

But (as I said before you wrote this patch) - IMO "Field" is too generic a term.  Conceptually, there are individual 'fields' within Meters for example.

So I don't think we should go this route.

Explorer09: Yes. And I plan the next commit would be to rename `RowField` to `Field` for the entire module.
Speaking of meters, the current code base refer to what you called 'fields' as 'items', so there is no confusion.
If you would complain that name being too generic, I would argue the same for the `Table` and `Row` class names.

natoscott: There is confusion here, for the reasons I've pointed out.

If we ask anyone what part of the htop UI relates to 'tables' and 'rows', they'll point out the exact right thing every time - trying to compare that to 'Fields' just doesn't work, sorry.

Its a NAK from me on this proposed naming change.

Explorer09: @natoscott I think the 'Field' would refer to the table cell. And it's not weird to call it a 'Field' anyway.

https://en.wikipedia.org/wiki/Column_(database)

natoscott: Sure, I understand where you're coming from - but it is quite ambiguous even within htop.  Definitely not "wierd" though - 'too generic' was the phrase used.

Explorer09:
1. I still don't like the 'Field' name to refer to the enum or identifier. And I think the type merge (`RowField` and `ProcessField`) is still necessary for ease of future maintanance.
2. Do you have a better name to suggest than `RowField` or `Field`? I was considering the `Column` name in place of `Field`. This will match the naming of `Table` and `Row` and might be easier to grasp.
3. This PR is still a draft, and I'll gather some comments before bringing in new commits. That are some more code mess that I will clean it up.

Explorer09:
My current plan:

- [ ] Rename `RowField` module to something like `Field` or `Column` that would make it more intuitive.
- [ ] Move the following to the 'Field' module: Row_printBytes, Row_printKBytes, Row_printCount, Row_printTime, Row_printRate, Row_printPercentage
- [ ] RowField_keyAt, RowField_alignedTitle

BenBE:
> My current plan:
> 
> * [ ]  Rename `RowField` module to something like `Field` or `Column` that would make it more intuitive.

Will likely get the PR rejected instantly. Both @natoscott and I object to this change.

> * [ ]  Move the following to the 'Field' module: Row_printBytes, Row_printKBytes, Row_printCount, Row_printTime, Row_printRate, Row_printPercentage
> * [ ]  RowField_keyAt, RowField_alignedTitle

Make that both `RowField` and we may get to talk about it.

Explorer09:
@BenBE How about the name `RowFieldID` in the `RowField` module?

Explorer09:
Closing this one temporarily as I don't know how to improve the naming with the current state of `main`...

PR #1297, opened 2023-09-03
Introduce autoTitleRightAlign column flag

cgzones:
Instead of handling PERCENT_CPU as a special case for whether to align the title of a dynamically sized column to the right or the left introduce a new flag, which can be reused by other columns.

Split from #1288 

/cc @Explorer09

### DIFF
--- a/Process.h
+++ b/Process.h
@@ -209,6 +209,9 @@ typedef struct ProcessFieldData_ {
 
    /* Whether the column width is dynamically adjusted (the minimum width is determined by the title length) */
    bool autoWidth;
+
+   /* Whether the title of a column with dynamically adjusted width is right aligned (default is left aligned) */
+   bool autoTitleRightAlign;
 } ProcessFieldData;
 
 #define LAST_PROCESSFIELD LAST_RESERVED_FIELD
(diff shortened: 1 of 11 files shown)


Process.h
```diff
@@ -209,6 +209,9 @@ typedef struct ProcessFieldData_ {
 
    /* Whether the column width is dynamically adjusted (the minimum width is determined by the title length) */
    bool autoWidth;
+
+   /* Whether the title of a column with dynamically adjusted width is right aligned (default is left aligned) */
+   bool autoTitleRightAlign;
```
natoscott, line 214: Is there anything automatic about the alignment?  (i.e. could the name be simplified to just 'titleRightAlign'?)

Explorer09:
My concerns about the `autoTitleRightAlign` flag are these:
1. There is only one user of the flag in this commit. Generally we shouldn't introduce a flag when there is only one user using it.
2. If a column needs to be right-aligned, the alignment would likely apply to not just the title, but the data cells as well. (It won't make sense to just right align the title and not the data.)

Process.h
BenBE: That `auto` refers to the automatic resize of certain columns.

cgzones:
Applied via https://github.com/htop-dev/htop/commit/cbcfc5f7d08e210a3477071607a3195fa31f06a3

PR #1305, opened 2023-09-23
Add recalculate to Ctrl-L refresh

### LINKED ISSUE #1304: feature request: key for rescan/recalculate
## Problem

Sometimes I need making rescan manually, for example when seeking for random popping processes, or when I know that some process currently doing really heavy job and I want to see how it's going more closely.

Previously I was abusing the fact that pressing `P` (sort by CPU usage) was doing rescan and I also feel comfortable with sorting by CPU usage. But starting from d83948ed09039864cdaf6066a1bd64476211ecaa this key stopped making full rescan.

(To be honest, I'm not sure this feature is really need. I mean, maybe I'm just got addicted to previous behavior. Hope someone will respond and we can discuss)

## Possible solutions

1. There is already key `Ctrl-L` which does force redraw (only). I suggest to make it full recalculate instead.
2. Add new key for recalculate, separate from `Ctrl-L`

fasterit:
Closes #1304

### DIFF
--- a/Action.c
+++ b/Action.c
@@ -619,7 +619,8 @@ static Htop_Reaction actionTag(State* st) {
 
 static Htop_Reaction actionRedraw(ATTR_UNUSED State* st) {
    clear();
-   return HTOP_REFRESH | HTOP_REDRAW_BAR;
+   // HTOP_RECALCULATE here to make Ctrl-L also refresh the data and not only redraw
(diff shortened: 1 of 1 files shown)


BenBE (approved):

natoscott (approved):
LGTM too.

PR #1306, opened 2023-09-24
Initialize user ID

cgzones:
Set the process user ID to an invalid value of -1 to trigger a name lookup for IDs with the value 0 ("root").  This initialization was dropped in commit 0f751e99 ("Introduce Row and Table classes for screens beyond top-processes").

Fixes: 0f751e99 ("Introduce Row and Table classes for screens beyond top-processes")


natoscott (approved):
LGTM, thanks @cgzones

Issue #1097, opened 2022-09-26
Can you supporte cygwin ? :)


1o1o1:
checking whether C compiler accepts -Wextra-semi-stmt... no
checking whether C compiler accepts -Wimplicit-int-conversion... yes
checking whether C compiler accepts -Wnull-dereference... no
checking that generated files are newer than configure... done
configure: creating ./config.status
config.status: creating Makefile
config.status: creating htop.1
config.status: creating config.h
config.status: config.h is unchanged
config.status: executing depfiles commands

****************************************************************
WARNING! This platform is not currently supported by htop.

The code will build, but it will produce a dummy version of htop
which shows no processes, using the files from the unsupported/
directory. This is meant to be a skeleton, to be used as a
starting point if you are porting htop to a new platform.
****************************************************************

  htop 3.2.1

  platform:                  unsupported
  os-release file:           /etc/os-release
  (Linux) proc directory:    /proc
  (Linux) openvz:            no
  (Linux) vserver:           no
  (Linux) ancient vserver:   no
  (Linux) delay accounting:  no
  (Linux) sensors:           no
  (Linux) capabilities:      no
  unicode:                   yes
  affinity:                  no
  unwind:                    no
  hwloc:                     no
  debug:                     no
  static:                    no

[22:05][admin@Windows2ooo:~/htop]#

BenBE:
It would be much better to make this compile native. The necessary code changes required to achieve this are manageable (but not trivial).

dnabre:
If all you want to should up are Cygwin processes, this is wouldn't take all that much work (ignoring all the Linux stuff like sensors and the like). 

If you wanted htop to display Windows process information, that's a whole another story all together. You'd have to operate as a Windows binary to start with, instead of going through Cygwin, and pull all the information about processes from Windows  API. That's a sizable amount of work, but the real issues come into play when you compare the Unix/Linux process model to the Windows one. Something stuff just doesn' map. Things like killing processes, process trees, and signals don't translate directly either.

BenBE:
> If all you want to should up are Cygwin processes, this is wouldn't take all that much work (ignoring all the Linux stuff like sensors and the like).

I don't think it would make much sense to have this run with Cygwin. The only sensible way IMHO is actual native support.

> If you wanted htop to display Windows process information, that's a whole another story all together.

[...]

PR #1308, opened 2023-09-27
Add experimental Cygwin/Msys support

ookiineko:
more details are in #1097


cygwin/CygwinMachine.c
```diff
@@ -0,0 +1,363 @@
...
+            tryRead("SReclaimable:", sreclaimableMem);  // zero
+            break;
+         }
+         break;
+      }
+
+      #undef tryRead
+   }
+
+   fclose(file);
```
BenBE, line 169: Have a look at [`GlobalMemoryStatusEx`](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-globalmemorystatusex). Further APIs should exist for even more details.

cygwin/CygwinMachine.c
```diff
@@ -0,0 +1,363 @@
...
+   host->usedSwap = swapTotalMem - swapFreeMem - swapCacheMem;
+   host->cachedSwap = swapCacheMem;
+}
+
+static void CygwinMachine_scanCPUTime(CygwinMachine* this) {
+   const Machine* super = &this->super;
+
+   CygwinMachine_updateCPUcount(this);
+
+   FILE* file = fopen(PROCSTATFILE, "r");
```
BenBE, line 196: HAve a look at https://stackoverflow.com/questions/63138932/calculating-cpu-usage-per-core-by-winapi

cygwin/CygwinMachine.c
```diff
@@ -0,0 +1,363 @@
...
+      cpuData->totalTime = totaltime;
+   }
+
+   this->period = (double)this->cpuData[0].totalPeriod / super->activeCPUs;
+
+   fclose(file);
+}
+
+static int scanCPUFrequencyFromSysCPUFreq(CygwinMachine* this) {
+   // TODO
```
BenBE, line 282: https://stackoverflow.com/questions/8351944/finding-out-the-cpu-clock-frequency-per-core-per-processor

Also note, that `QueryPerformanceCounter` gives false information, cf. [this video](https://www.youtu.be/pZ0MF1q_LUE) for details

cygwin/CygwinMachine.c
```diff
@@ -0,0 +1,363 @@
...
+   (void) this;
+
+   return -ENOSYS;
+}
+
+static void scanCPUFrequencyFromCPUinfo(CygwinMachine* this) {
+   const Machine* super = &this->super;
+
+   FILE* file = fopen(PROCCPUINFOFILE, "r");
```
BenBE, line 291: Preferably use Win32/WinNT API, cf. above. Also note [this video by Dave Plummer](https://www.youtu.be/oKQ1X_4JCn0) regarding per-CPU usage details.

cygwin/CygwinProcessTable.c
```diff
@@ -0,0 +1,799 @@
...
+*/
+
+#include "cygwin/CygwinProcessTable.h"
+
+#include <stdlib.h>
+#include <dirent.h>
+#include <stdio.h>
+#include <string.h>
+#include <sys/types.h>
+#include <unistd.h>
```
BenBE, line 16: Header order. Cf. style guide …

cygwin/CygwinProcessTable.c
```diff
@@ -0,0 +1,799 @@
...
+   int fd = Compat_openat(openatArg, pathname, O_RDONLY);
+   if (fd < 0)
+      return NULL;
+
+   FILE* stream = fdopen(fd, mode);
+   if (!stream)
+      close(fd);
+
+   return stream;
+}
```
BenBE, line 37: Please avoid the libc `FILE*` APIs and use `read`/`write` with the `fd` directly.

cygwin/CygwinProcessTable.c
```diff
@@ -0,0 +1,799 @@
...
+      nibble ^= 0x10; // switch letters and digits
+      nibble -= letter;
+      nibble &= 0x0f;
+      result <<= 4;
+      result += (uint64_t)nibble;
+      (*str)++;
+   }
+
+   return result;
+}
```
BenBE, line 79: If you are going to re-use these functions, please refactor them into some common code to avoid code duplication.

cygwin/CygwinProcessTable.c
```diff
@@ -0,0 +1,799 @@
...
+   /* Ignore further fields */
+
+   process->processor = 0;  // FIXME
+
+   process->time = cp->utime + cp->stime;
+
+   return true;
+}
+
+static bool CygwinProcessTable_updateUser(const Machine* host, Process* process, openat_arg_t procFd) {
```
BenBE, line 229: Should be made SID-aware

cygwin/CygwinProcessTable.c
```diff
@@ -0,0 +1,799 @@
...
+                  &dummy); /* unused; always 0 */
+   fclose(statmfile);
+
+   if (r == 7) {
+      process->super.m_virt *= host->pageSizeKB;
+      process->super.m_resident *= host->pageSizeKB;
+   }
+
+   return r == 7;
+}
```
BenBE, line 346: Have a look at [EnumProcessModulesEx](https://learn.microsoft.com/en-us/windows/win32/api/psapi/nf-psapi-enumprocessmodulesex).

cygwin/CygwinProcessTable.c
```diff
@@ -0,0 +1,799 @@
...
+
+         Process_updateExe(process, filename);
+      }
+   } else if (process->procExe) {
+      Process_updateExe(process, NULL);
+      process->procExeDeleted = false;
+   }
+
+   return true;
+}
```
BenBE, line 527: Have a look [here](https://stackoverflow.com/questions/6530565/getting-another-process-command-line-in-windows).
Also note the [comments by Reymond Chen on the subject](https://devblogs.microsoft.com/oldnewthing/20091125-00/?p=15923).

When rewriting the code, make sure the `mergedCommand` fields are properly set.
The fields `procExeDeleted` and `usesDeletedLib` can be assumed `false` on Windows, due to [$reasons](https://learn.microsoft.com/en-us/previous-versions/technet-magazine/cc983682(v=msdn.10)?redirectedfrom=MSDN) …

cygwin/CygwinProcessTable.c
```diff
@@ -0,0 +1,799 @@
...
+
+   int r = fscanf(winpidfile, "%u", &winpid);
+   fclose(winpidfile);
+
+   if (r == 1) {
+      process->winpid = winpid;
+   }
+
+   return r == 1;
+}
```
BenBE, line 568: Prefer `xReadfileat` instead.

cygwin/CygwinProcessTable.c
```diff
@@ -0,0 +1,799 @@
...
+   /* PROCDIR is an absolute path */
+   assert(PROCDIR[0] == '/');
+#ifdef HAVE_OPENAT
+   openat_arg_t rootFd = AT_FDCWD;
+#else
+   openat_arg_t rootFd = "";
+#endif
+
+   CygwinProcessTable_recurseProcTree(this, rootFd, chost, PROCDIR, NULL);
+}
```
BenBE, line 799: Cf. [EnumProcesses](https://learn.microsoft.com/en-us/windows/win32/api/psapi/nf-psapi-enumprocesses)

cygwin/Platform.c
```diff
@@ -0,0 +1,337 @@
...
+#include "Meter.h"
+#include "XUtils.h"
+#include "Settings.h"
+#include "SwapMeter.h"
+#include "SysArchMeter.h"
+#include "TasksMeter.h"
+#include "UptimeMeter.h"
+
+#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
```
BenBE, line 35: Please be consistent about how you handle the `WIN32_LEAN_AND_MEAN` flag.

cygwin/Platform.c
```diff
@@ -0,0 +1,337 @@
...
+#include "Macros.h"
+#include "MemoryMeter.h"
+#include "MemorySwapMeter.h"
+#include "Meter.h"
+#include "XUtils.h"
+#include "Settings.h"
+#include "SwapMeter.h"
+#include "SysArchMeter.h"
+#include "TasksMeter.h"
+#include "UptimeMeter.h"
```
BenBE, line 32: Header order, cf. style guide.

cygwin/Platform.c
```diff
@@ -0,0 +1,337 @@
...
+   FILE* fd = fopen(PROCDIR "/uptime", "r");
+   if (fd) {
+      int n = fscanf(fd, "%64lf", &uptime);
+      fclose(fd);
+      if (n <= 0) {
+         return 0;
+      }
+   }
+   return floor(uptime);
+}
```
BenBE, line 145: Cf. [GetTickCount64](https://learn.microsoft.com/de-de/windows/win32/api/sysinfoapi/nf-sysinfoapi-gettickcount64)

cygwin/Platform.c
```diff
@@ -0,0 +1,337 @@
...
+err:
+   *one = NAN;
+   *five = NAN;
+   *fifteen = NAN;
+}
+
+pid_t Platform_getMaxPid(void) {
+   // https://github.com/cygwin/cygwin/blob/a9e8e3d1cb8235f513f4d8434509acf287494fcf/winsup/cygwin/local_includes/pinfo.h#L223
+   return 65536;
+}
```
BenBE, line 172: Note the limitations regarding Cygwin here. On native Windows the story is [a bit different](https://learn.microsoft.com/en-us/answers/questions/70930/maximum-value-of-process-id) ([Reymond Chen](https://devblogs.microsoft.com/oldnewthing/20040223-00/?p=40503)) …

cygwin/Platform.h
```diff
@@ -0,0 +1,132 @@
...
+#include "Action.h"
+#include "BatteryMeter.h"
+#include "DiskIOMeter.h"
+#include "Hashtable.h"
+#include "Meter.h"
+#include "NetworkIOMeter.h"
+#include "Process.h"
+#include "ProcessLocksScreen.h"
+#include "SignalsPanel.h"
+#include "CommandLine.h"
```
BenBE, line 23: Header order …

BenBE (commented):
It would be much preferable if a WinNT port could avoid relying on any translation layers like Cygwin and read the API directly.

configure.ac
```diff
@@ -27,6 +27,10 @@ linux*|gnu*)
    my_htop_platform=linux
    AC_DEFINE([HTOP_LINUX], [], [Building for Linux.])
    ;;
+cygwin*|msys*)
```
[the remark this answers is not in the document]
ookiineko: > I'm not quite sure here, but aren't cygwin and MSys somewhat different in how they expose the low-level system API access?

read: https://www.msys2.org/wiki/How-does-MSYS2-differ-from-Cygwin/

for code changes: https://github.com/msys2/msys2-runtime/commits/msys2-3.4.9

cygwin/CygwinMachine.c
ookiineko: is there any difference from directly reading cygwin procfs? i thought cygwin already maps win32 api very well.

cygwin/CygwinMachine.c
ookiineko: > HAve a look at https://stackoverflow.com/questions/63138932/calculating-cpu-usage-per-core-by-winapi

same question as above

cygwin/CygwinMachine.c
ookiineko: > https://stackoverflow.com/questions/8351944/finding-out-the-cpu-clock-frequency-per-core-per-processor
> 
> Also note, that `QueryPerformanceCounter` gives false information

thanks. indeed, i will check it when i have time

cygwin/CygwinProcessTable.c
ookiineko: > Have a look at [EnumProcessModulesEx](https://learn.microsoft.com/en-us/windows/win32/api/psapi/nf-psapi-enumprocessmodulesex).

readMaps is only used to detect deleted libs, m_lrs is directly read from statm, not calculated.

for statm, cygwin implements it using kernel api directly: https://github.com/cygwin/cygwin/blob/a9e8e3d1cb8235f513f4d8434509acf287494fcf/winsup/cygwin/fhandler/process.cc#L1507

cygwin/CygwinProcessTable.c
ookiineko: i am not familiar with windows apis... so i dont have plan to replace these with native api calls when cygwin already provides wrappers for it.

cygwin/CygwinProcessTable.c
BenBE: I'd prefer for the htop port to take that route too (without the detour over cygwin's procfs layer).

cygwin/CygwinProcessTable.c
ookiineko: > Please avoid the libc `FILE*` APIs and use `read`/`write` with the `fd` directly.

this is copied from the existing Linux support: https://github.com/htop-dev/htop/blob/2978af01a79d1627d7fa90fda0efec9cf774d901/linux/LinuxProcessTable.c#L67-L79

cygwin/Platform.c
ookiineko: > Please be consistent about how you handle the `WIN32_LEAN_AND_MEAN` flag.

got it

cygwin/CygwinProcessTable.c
ookiineko: > If you are going to re-use these functions, please refactor them into some common code to avoid code duplication.

will move to generic later

cygwin/CygwinProcessTable.c
ookiineko: > Header order. Cf. style guide …

ok sorry, im still learning

cygwin/CygwinProcessTable.c
BenBE: Maybe a quick note re `m_lrs` on Linux for reference:
On linux this field has not been reported for quite a while, thus `htop` implemented calculating it on-the-fly when this column is shown.

Libraries on Windows can't be really deleted/replaced while a process is using them, thus my remark re the two `mergedCmdline` flags.

The overall code thus should reduce quite a bit if using native API here directly.

ookiineko: > Libraries on Windows can't be really deleted/replaced while a process is using them

i have tested it with a program which depends on a dll in cygwin, i am able to delete both the program executable and the library it loaded, and when the dll is deleted, its path showed in /proc/[pid]/maps starts with `C:\$Recycle.Bin` and some unique id in my case. in fact the deleted executable is also located there

cygwin/CygwinMachine.c
ookiineko: > Mostly performance and level of detail.
> 
> I'd prefer to avoid depending on both the cygwin and msys runtimes; and go fully native instead.

i see. but for now i will leave this work to someone who's more comfortable and familiar with windows apis.

about the performance actually it's not that bad, see https://github.com/htop-dev/htop/issues/1097#issuecomment-1737465996

cygwin/Platform.c
ookiineko: > Note the limitations regarding Cygwin here. On native Windows the story is a bit different

but, this is a cygwin port, not native windows port...

ookiineko:
i understand. but this is just very initial support for cygwin-based platform only

it's not meant to use native windows apis (unless necessary like the one used for fetching battery stats). since directly using windows api for cygwin wrapped stuffs may cause issues since it maintains some internal states besides windows stuffs

these code are mostly borrowed from elsewhere in the htop's existing codebase (like the platform support code for Linux, Solaris and BSDs), so proves making htop work for this platform is almost trivial


afterall, i plan to clean these up and move some code tomorrow

cygwin/CygwinProcessTable.c
ookiineko: > Prefer `xReadfileat` instead.

done

cygwin/Platform.c
ookiineko: > Please be consistent about how you handle the `WIN32_LEAN_AND_MEAN` flag.

please check it again, does it looks better now?

cygwin/Platform.c
ookiineko: > Header order, cf. style guide.

does it look better now?

cygwin/Platform.h
ookiineko: > Header order …

fixed

cygwin/CygwinProcessTable.c
ookiineko: > Header order. Cf. style guide …

corrected now

cygwin/CygwinProcessTable.c
BenBE: I know. And that code is on my todo list of deprecated things that need to be cleaned up at some point …

cygwin/CygwinProcessTable.c
BenBE: That's not really deleting the file, but moving (which Windows allows).

cygwin/CygwinMachine.c
```diff
@@ -0,0 +1,327 @@
...
+      if (String_startsWith(buffer, "btime ") == false)
+         continue;
+      if (sscanf(buffer, "btime %lld\n", &this->boottime) == 1)
+         break;
+      CRT_fatalError("Failed to parse btime from " PROCSTATFILE);
+   }
+   fclose(statfile);
+
+   if (this->boottime == -1)
+      CRT_fatalError("No btime in " PROCSTATFILE);
```
BenBE, line 89: Using [`GetTickCount64`](https://learn.microsoft.com/de-de/windows/win32/api/sysinfoapi/nf-sysinfoapi-gettickcount64?redirectedfrom=MSDN) in combination with [`clock_gettime`](https://linux.die.net/man/3/clock_gettime) and `CLOCK_REALTIME` is both faster, less code, and avoids most of the error checking.

cygwin/CygwinProcessTable.c
```diff
@@ -0,0 +1,747 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <sys/types.h>
+
+#include "Compat.h"
+#include "Object.h"
+#include "ProcessTable.h"
+#include "XUtils.h"
```
BenBE, line 26:
```suggestion
#include "cygwin/CygwinProcessTable.h"

#include <dirent.h>
#include <stdint.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <unistd.h>
#include <sys/types.h>

#include "Compat.h"
#include "Object.h"
#include "ProcessTable.h"
#include "XUtils.h"

#include "cygwin/CygwinMachine.h"
#include "cygwin/CygwinProcess.h"
#include "generic/fast_strtoull.h"
#include "generic/fdopenat.h"
```

cygwin/Platform.c
```diff
@@ -0,0 +1,339 @@
...
+#include "Macros.h"
+#include "MemoryMeter.h"
+#include "MemorySwapMeter.h"
+#include "Meter.h"
+#include "Settings.h"
+#include "SwapMeter.h"
+#include "SysArchMeter.h"
+#include "TasksMeter.h"
+#include "UptimeMeter.h"
+#include "XUtils.h"
```
BenBE, line 37:
```suggestion
#include "cygwin/Platform.h"

#include <math.h>
#include <stdio.h>

#include "ClockMeter.h"
#include "CPUMeter.h"
#include "DateMeter.h"
#include "DateTimeMeter.h"
#include "HostnameMeter.h"
#include "LoadAverageMeter.h"
#include "Macros.h"
#include "MemoryMeter.h"
#include "MemorySwapMeter.h"
#include "Meter.h"
#include "Settings.h"
#include "SwapMeter.h"
#include "SysArchMeter.h"
#include "TasksMeter.h"
#include "UptimeMeter.h"
#include "XUtils.h"

#include "cygwin/CygwinMachine.h"

#ifndef WIN32_LEAN_AND_MEAN
#define WIN32_LEAN_AND_MEAN
#endif /* WIN32_LEAN_AND_MEAN */
#include <windows.h>  // for DWORD
#undef WIN32_LEAN_AND_MEAN
```

cygwin/CygwinProcessTable.c
ookiineko: > That's not really deleting the file, but moving (which Windows allows).

ohh, then should i remove the related code?

generic/fdopenat.h
```diff
@@ -0,0 +1,32 @@
...
+/*
+htop - generic/fdopenat.h
+(C) 2014 Hisham H. Muhammad
+Released under the GNU GPLv2+, see the COPYING file
+in the source distribution for its full text.
+*/
+
+#include <stdio.h>
+#include <assert.h>
+#include <unistd.h>
```
BenBE, line 12:
```suggestion
#include <assert.h>
#include <stdio.h>
#include <unistd.h>
```

cygwin/CygwinProcessTable.c
ookiineko: >  [Suggested change]

thanks, fixed

cygwin/CygwinProcessTable.c
```diff
@@ -0,0 +1,747 @@
...
+         continue;
+
+      if (checkDeletedLib && map_execute && !proc->usesDeletedLib) {
+         while (*readptr == ' ')
+            readptr++;
+
+         if (*readptr != '/')
+            continue;
+
+         if (strstr(readptr, "$Recycle.Bin")) {
```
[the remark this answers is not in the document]
ookiineko: > Best to require `'/'` or `'\\'` at the start and end of this path segment.

ok, added

cygwin/Platform.c
ookiineko: > [Suggested change]

hum, i think `windows.h` is a system header

generic/fdopenat.h
ookiineko: > [Suggested change]

oops, fixed

cygwin/Platform.c
```diff
@@ -0,0 +1,339 @@
...
+
+FileLocks_ProcessData* Platform_getProcessLocks(pid_t pid) {
+   // TODO
+   (void)pid;
+   return NULL;
+}
+
+void Platform_getFileDescriptors(double* used, double* max) {
+   // TODO
+   *used = NAN;
```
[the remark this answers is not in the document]
ookiineko: > Could be accumulated via [GetProcessHandleCount](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getprocesshandlecount)

some fds can be wrapped by cygwin and not actual win32 handle, so i think using cygwin api is needed here

i think i can iterate /proc/[pid]/fd for this

cygwin/Platform.c
```diff
@@ -0,0 +1,339 @@
...
+   *max = NAN;
```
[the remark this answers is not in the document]
ookiineko: > TL;DR: 16,711,680 on x64 and 16,744,448 on x86; per process.

but this is a global limit of fds, we may need to multiply it by the maximum number of processes.

also, does it apply to cygwin? some fds can be wrapped by cygwin and are not actual win32 handles.

cygwin/CygwinProcessTable.c
ookiineko: > Should be made SID-aware

it's session ID, not security identifier...

see:
https://github.com/cygwin/cygwin/blob/main/winsup/cygwin/fhandler/process.cc#L1197
https://github.com/cygwin/cygwin/blob/a9e8e3d1cb8235f513f4d8434509acf287494fcf/winsup/cygwin/local_includes/pinfo.h#L87

cygwin/Platform.c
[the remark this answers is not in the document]
ookiineko: > i think i can iterate /proc/[pid]/fd for this

implemented

BenBE (commented):
LGTM.

cygwin/Platform.c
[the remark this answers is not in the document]
BenBE: Those limits are purely based on the implementation in the WinAPI and thus cygwin somewhat inherits them. But actual limits in cygwin may be lower or higher (e.g. when cygwin shares one handle for a resource). And yes, those numbers would need to be multiplied by the number of processes.

Though in all fairness: For the FD meter to be somewhat sensible, we can assume those as global limits and remember the max open if that was ever above the number of opened handles. Because we know at least that many handles can exist and tracking the actual number will be rather pointless for the actual display in the bar meter … (NB: You can report different numbers for the text representation and the bar graph).

@fasterit IIRC there was Solaris where things where more or less similarly dynamic?

cygwin/Platform.c
BenBE: Defining those `#ifdef`s inline is fine as is. Just a matter of ordering. Also, for things where there's not just the include itself, those are typically pulled down into their own block. It's not fully obvious, that's why this suggested patch.

cygwin/Platform.c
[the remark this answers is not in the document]
fasterit: > @fasterit IIRC there was Solaris where things where more or less similarly dynamic?

Yes, there is no global limit (RAM is the limit). But a hard and a soft limit per user and per process.
Which also change between minor versions. Because $reasons (https://support.oracle.com/knowledge/Sun%20Microsystems/2736930_1.html [paywalled]).
But the good news is ... none of that is implemented to be visualized in htop for Solaris :-).

ookiineko:
oops, just renamed branch to test ci (because i don't want CodeQL workflow to trigger)

ookiineko:
added CI for these on another branch, for more info please check:

https://github.com/htop-dev/htop/issues/1097#issuecomment-1740424506

I can send another PR if needed

BenBE:
Having looked at the CI part in your repo, I think this can go together.

Can you rebase your PR to integrate the fixups into the commits where the need for that fixup was introduced in? For example, adding the docs for the WINPID column in the same commit that column was introduced. Or reordering of includes in the commits where they were added initially.

I'm not a fan of cc4d180ca13124fa80711d8197c3f62d205fec42 as a compiler with active LTO will do this automatically.

The commit 6d518e99a9ce52fce2697fcd7a996a8aeced154c should be split. Also `_GNU_SOURCE` should be defined using `config.h` (cf. example in style guide regarding the required IWYU comment).

ookiineko:
> Having looked at the CI part in your repo, I think this can go together.

OK

> Can you rebase your PR to integrate the fixups into the commits where the need for that fixup was introduced in? For example, adding the docs for the WINPID column in the same commit that column was introduced. Or reordering of includes in the commits where they were added initially.

Yes I can do that

> I'm not a fan of [cc4d180](https://github.com/htop-dev/htop/commit/cc4d180ca13124fa80711d8197c3f62d205fec42) as a compiler with active LTO will do this automatically.

Oh, i see. i will drop it later

> The commit [6d518e9](https://github.com/htop-dev/htop/commit/6d518e99a9ce52fce2697fcd7a996a8aeced154c) should be split. Also `_GNU_SOURCE` should be defined using `config.h` (cf. example in style guide regarding the required IWYU comment).

i will rework that commit. thanks, got it.

ookiineko:
i have finished the required changes

also checked to ensure each commit (since the commit which adds skeleton for the new platform) is buildable and runable

ookiineko:
looks like the development team has no interest adding Cygwin/Msys support, and doesn't treat Cygwin as a standalone unix-like platform, so I'm closing this pr for now

(Cygwin/Msys support will still be maintained in [my fork](https://github.com/ookiineko-cygport/htop/tree/cygwin) )

PR #1309, opened 2023-09-28
Write configuration to temporary file first

cgzones:
Currently when the runtime configuration is saved to file the true configuration file is first truncated and then written to with each available option.

If for any reason htop dies (e.g. program crash, or system shutdown) while still performing the output the configuration file might end up empty or corrupted.

Write to a temporary file and rename at the end on success.

### DIFF
--- a/Settings.c
+++ b/Settings.c
@@ -603,11 +603,16 @@ static void writeMeterModes(const Settings* this, FILE* fd, char separator, unsi
 int Settings_write(const Settings* this, bool onCrash) {
    FILE* fd;
    char separator;
+   char *tmpFilename = NULL;
    if (onCrash) {
       fd = stderr;
       separator = ';';
    } else {
-      fd = fopen(this->filename, "w");
+      xAsprintf(&tmpFilename, "%s.tmp.XXXXXX", this->filename);
+      int fdtmp = mkstemp(tmpFilename);
+      if (fdtmp == -1)
+         return -errno;
+      fd = fdopen(fdtmp, "w");
       if (fd == NULL)
          return -errno;
       separator = '\n';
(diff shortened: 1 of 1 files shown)


Settings.c
```diff
@@ -603,11 +603,13 @@ static void writeMeterModes(const Settings* this, FILE* fd, char separator, unsi
 int Settings_write(const Settings* this, bool onCrash) {
    FILE* fd;
    char separator;
+   char *filename;
```
BenBE, line 606: Please initialize to `NULL` to avoid undefined/uninitialized values `filename` in case of a crash (even if not used). Better have a defined, unused value, than accidentally using an uninitialized one.

BenBE (commented):
LGTM.

Settings.c
```diff
@@ -603,11 +603,13 @@ static void writeMeterModes(const Settings* this, FILE* fd, char separator, unsi
...
    if (onCrash) {
       fd = stderr;
       separator = ';';
    } else {
-      fd = fopen(this->filename, "w");
+      xAsprintf(&filename, "%s.tmp", this->filename);
+      fd = fopen(filename, "w");
```
Explorer09, line 612: May I suggest using [mkstemp](https://man7.org/linux/man-pages/man3/mkstemp.3.html) for the purpose or is it an overkill?
```c
      xAsprintf(&filename, "%s.XXXXXX", this->filename);
      fd = mkstemp(filename);
```

BenBE: Not exactly overkill, but when multiple instances of htop are running and quit quickly after each other you would still run into an race of which configuration actually prevails. But avoiding the possible corruption of multiple instances writing the same temp file might be worth another look at this part (you then at least know the surviving configuration has not been corrupted while it was generated) …

PR #1312, opened 2023-10-08
strchrnul is a GNU libc extension, add a wrapper for other platforms

natoscott:


XUtils.h
```diff
@@ -56,6 +56,17 @@ void String_freeArray(char** s);
 
 char* String_readLine(FILE* fd) ATTR_MALLOC;
 
+static inline char *String_charNull(const char* s, int c) {
```
BenBE, line 59: Nitpick: The character is called `NUL` in the ASCII standard.

Any objections to calling this wrapper `String_strchrnul`?

Explorer09:
Even though `strchrnul` is a non-standard extension, there is an implementation in gnulib. Are we not considering using gnulib?

XUtils.h
```diff
@@ -56,6 +56,17 @@ void String_freeArray(char** s);
...
+#ifdef HAVE_STRCHRNUL
+   return strchrnul(s, c);
+#else
+   char *result;
+   if ((result = strchr(s, c)))
+      return result;
+   return strchr(s, '\0');
```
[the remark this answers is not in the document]
Explorer09: I have a naïve implementation here, because I don't think calling `strchr` twice is a good idea for performance:
```c
   char* p = (char*)s;
   while (*p && *p != c)
      p++;
   return p;
```

XUtils.h
Explorer09:
```suggestion
static inline char *String_charNull(const char* s, int c) ATTR_NONNULL {
```

BenBE:
> Even though `strchrnul` is a non-standard extension, there is an implementation in gnulib. Are we not considering using gnulib?

No. Just what gets auto-detected in `configure.ac` …

XUtils.h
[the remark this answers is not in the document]
natoscott: OTOH, strchr is highly optimised so chances are fair-to-middling that two calls are still more efficient than a naive re-implementation.  Think I'll stick with the strchr version unless measurements show compelling evidence to the contrary.

XUtils.h
natoscott: OK, no objections - will push these changes shortly, thanks.  Hmm, looks like the latter is invalid - results in:

```
XUtils.h:59:1: error: attributes should be specified before the declarator in a function definition
   59 | static inline char *String_strchrnul(const char* s, int c) ATTR_NONNULL {
      | ^~~~~~
```

I'll drop that second suggestion as a result.  If it was feasible, String_eq and String_startsWith in this header file could use the same treatment so I guess its not possible.

PR #1313, opened 2023-10-10
Add nonnull attribute to various String_* functions

Explorer09:
This is a follow-up to PR #1312. In particular `String_strchrnul` is better to have this attribute added.

Note that this not the best position to specify the function attributes in a function definition. I just wrote it to match the style of other function definitions in htop's headers.

The recommend placement of the function attributes, according to various documentation, is immediately before the function name, after the return type (and _before_ the function pointer asterisks, if any):
```c
static inline char* ATTR_NONNULL String_strchrnul(const char* s, int c) {
   ...
}
int (ATTR_NONNULL *sortCompare) (const void* a, const void* b);
```

This is the current style in htop (still works):
```c
ATTR_NONNULL
static inline char* String_strchrnul(const char* s, int c) {
   ...
}
```

The technical difference is minor, but according to gcc documentation, when multiple functions are declared in a comma-separated list, the attributes specified before the return type specifier would apply to all function declared in the list.

The [example in gcc documentation](https://gcc.gnu.org/onlinedocs/gcc/Attribute-Syntax.html) states that the `noreturn` attribute would apply to both `d1` and `d2`:
```c
__attribute__((noreturn)) void d0 (void),
    __attribute__((format(printf, 1, 2))) d1 (const char *, ...),
     d2 (void);
```

### DIFF
--- a/Macros.h
+++ b/Macros.h
@@ -42,7 +42,6 @@ in the source distribution for its full text.
 #ifdef  __GNUC__  // defined by GCC and Clang
 
 #define ATTR_FORMAT(type, index, check) __attribute__((format (type, index, check)))
-#define ATTR_NONNULL                    __attribute__((nonnull))
 #define ATTR_NORETURN                   __attribute__((noreturn))
 #define ATTR_UNUSED                     __attribute__((unused))
 #define ATTR_MALLOC                     __attribute__((malloc))
@@ -50,12 +49,23 @@ in the source distribution for its full text.
 #else /* __GNUC__ */
 
 #define ATTR_FORMAT(type, index, check)
-#define ATTR_NONNULL
 #define ATTR_NORETURN
 #define ATTR_UNUSED
 #define ATTR_MALLOC
 
 #endif /* __GNUC__ */
+
+#ifdef HAVE_ATTR_NONNULL
+
+#define ATTR_NONNULL                    __attribute__((nonnull))
+#define ATTR_NONNULL_N(...)             __attribute__((nonnull(__VA_ARGS__)))
+
+#else
+
+#define ATTR_NONNULL
+#define ATTR_NONNULL_N(...)
+
+#endif /* HAVE_ATTR_NONNULL */
 
 #ifdef HAVE_ATTR_ALLOC_SIZE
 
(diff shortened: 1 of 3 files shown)


natoscott:
Thanks for following up on this!  Great to see it is feasible with static inlines.

| ... when multiple functions are declared in a comma-separated list ...

and lets never, ever add code like that.  :)

natoscott (approved):

BenBE (commented):
LGTM.

Maybe also add for `String_cat`, `String_trim` and `String_readLine`.

For `String_split` you need the "first arg only" variant.
Other functions may need some-arg-only treatment too (like `xAsprintf` for `fmt`, but not `strp`).

Explorer09:
@BenBE I'm not sure which style htop would prefer if we are to introduce a one-arg `nonnull` attribute.

```c
// Current code

char** String_split(const char* s, char sep, size_t* n);

ATTR_FORMAT(printf, 2, 3)
int xAsprintf(char** strp, const char* fmt, ...);

// Style 1

#define ATTR_NONNULL1(a) __attribute__((nonnull (a)))
#define ATTR_NONNULL2(a, b) __attribute__((nonnull (a, b)))

ATTR_NONNULL1(1)
char** String_split(const char* s, char sep, size_t* n);

ATTR_FORMAT(printf, 2, 3) ATTR_NONNULL2(1, 2)
int xAsprintf(char** strp, const char* fmt, ...);

// Style 2

#define ATTR_NONNULL_N(...) __attribute__((nonnull (__VA_ARGS__)))

ATTR_NONNULL_N(1)
char** String_split(const char* s, char sep, size_t* n);

ATTR_FORMAT(printf, 2, 3) ATTR_NONNULL_N(1, 2)
int xAsprintf(char** strp, const char* fmt, ...);

// Style 3 (supported in Clang as an extension; might not work in GCC)

char** String_split(const char* ATTR_NONNULL s, char sep, size_t* n);

ATTR_FORMAT(printf, 2, 3)
int xAsprintf(char** ATTR_NONNULL strp, const char* ATTR_NONNULL fmt, ...);
```

Personally I think Clang made the right choice by allowing `nonnull` attributes specified directly on arguments rather than on functions. (This syntactically makes more sense.) But GCC is one complier that is unlikely to support a syntax like this.

BenBE:
Keep all args as `ATTR_NONNULL` and have individual args as varargs like `ATTR_NONNULL_N(1, 3, 12)`.

BenBE (approved):

natoscott (approved):
Nice work!

PR #1314, opened 2023-10-10
GraphMeterMode_draw() code cleanup 

Explorer09:
A follow up of PR #1024.

1. In `GraphMeterMode_draw()`, `w` can be less than zero when called from the draw routines of e.g. CPU meters, when the terminal window size is very small.
2. The last commit "Cap the maximum size of GraphData buffer" is optional. I can remove it upon request.

### DIFF
--- a/Meter.c
+++ b/Meter.c
@@ -299,14 +299,19 @@ static void GraphMeterMode_draw(Meter* this, int x, int y, int w) {
       return;
 
    GraphData* data = &this->drawData;
-   if ((size_t)w * 2 > data->nValues) {
+   if ((size_t)w * 2 > data->nValues && MAX_METER_GRAPHDATA_VALUES > data->nValues) {
       size_t oldNValues = data->nValues;
-      data->nValues = MAXIMUM(oldNValues + (oldNValues / 2), (size_t)w * 2);
+      data->nValues = MAXIMUM(oldNValues + oldNValues / 2, (size_t)w * 2);
+      data->nValues = MINIMUM(data->nValues, MAX_METER_GRAPHDATA_VALUES);
       data->values = xReallocArray(data->values, data->nValues, sizeof(*data->values));
       memmove(data->values + (data->nValues - oldNValues), data->values, oldNValues * sizeof(*data->values));
       memset(data->values, 0, (data->nValues - oldNValues) * sizeof(*data->values));
    }
    const size_t nValues = data->nValues;
+   if ((size_t)w * 2 > nValues) {
+      x += w - nValues / 2;
+      w = nValues / 2;
+   }
 
    const Machine* host = this->host;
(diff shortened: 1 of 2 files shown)


Meter.c
```diff
@@ -333,14 +333,9 @@ static void GraphMeterMode_draw(Meter* this, int x, int y, int w) {
...
-   ssize_t i = (ssize_t)nValues - (w * 2);
-   ssize_t k = 0;
-   if (i < 0) {
-      k = -i / 2;
-      i = 0;
-   }
-   for (; i < (ssize_t)nValues - 1; i += 2, k++) {
+   size_t i = nValues - (size_t)w * 2;
+   int k = 0;
+   for (; i < (size_t)nValues - 1; i += 2, k++) {
```
BenBE, line 343: With `i` and `nValues` being both `size_t`, no cast is needed here.

Meter.c
```diff
@@ -302,6 +308,18 @@ static void GraphMeterMode_draw(Meter* this, int x, int y, int w) {
...
+      int globalDelay = host->settings->delay;
+      struct timeval delay = { .tv_sec = globalDelay / 10, .tv_usec = (globalDelay % 10) * 100000L };
+      timeradd(&host->realtime, &delay, &(data->time));
+
+      for (size_t i = 0; i < nValues - 1; i++)
+         data->values[i] = data->values[i + 1];
+
+      data->values[nValues - 1] = sumPositiveValues(this->values, this->curItems);
+   }
+
```
BenBE, line 323: Maybe split this moved code segment as its own commit for clarity which of the changes go together.

Meter.c
```diff
@@ -333,6 +334,10 @@ static void GraphMeterMode_draw(Meter* this, int x, int y, int w) {
       GraphMeterMode_pixPerRow = PIXPERROW_ASCII;
    }
 
+   if ((size_t)w * 2 > nValues) {
+      x += w - (nValues / 2);
```
BenBE, line 338: Can you drop those superfluous parens?
There are some more places, where they aren't strictly necessary and may be removed.

Meter.c
```diff
@@ -333,6 +334,10 @@ static void GraphMeterMode_draw(Meter* this, int x, int y, int w) {
...
+      w = nValues / 2;
+   }
    size_t i = nValues - (size_t)w * 2;
    int k = 0;
```
BenBE, line 342: With `k` being a loop var, maybe pull it into the `for` initializer?

Meter.c
Explorer09: This one casts from a signed type (`w`) to unsigned (`nValues`). I'll get a compile warning if I don't make the explicit cast.

BenBE (approved):

PR #1315, opened 2023-10-11
Record new graph data even when no graph is drawn.

Explorer09:
Fixes a regression caused by 74b28849c4ee7234aa6018667ff6c8c9a6b41273.

When a graph meter is too narrow to draw the graph, but the graph data buffer is allocated, the meter should keep recording new values.

### DIFF
--- a/Meter.c
+++ b/Meter.c
@@ -10,6 +10,7 @@ in the source distribution for its full text.
 #include "Meter.h"
 
 #include <assert.h>
+#include <limits.h>
 #include <math.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


Meter.c
```diff
@@ -308,10 +308,8 @@ static void GraphMeterMode_draw(Meter* this, int x, int y, int w) {
       memset(data->values, 0, (data->nValues - oldNValues) * sizeof(*data->values));
    }
    const size_t nValues = data->nValues;
-   if ((size_t)w * 2 > nValues) {
-      x += w - nValues / 2;
-      w = nValues / 2;
-   }
+   if (nValues <= 0)
+      return;
```
BenBE, line 312: `size_t nValues` cannot be less than 0; only equal … I'd prefer `nValues < 1` for clarity instead.

Meter.c
```diff
@@ -325,6 +323,13 @@ static void GraphMeterMode_draw(Meter* this, int x, int y, int w) {
       data->values[nValues - 1] = sumPositiveValues(this->values, this->curItems);
    }
 
+   if (w <= 0)
+      return;
+   if ((size_t)w > nValues / 2) {
+      x += w - nValues / 2;
+      w = nValues / 2;
+   }
+
```
BenBE, line 332: Just some more space for the code to breathe … ;-)

```suggestion
   if (w <= 0)
      return;

   if ((size_t)w > nValues / 2) {
      x += w - nValues / 2;
      w = nValues / 2;
   }

```

Meter.c
```diff
@@ -308,10 +308,8 @@ static void GraphMeterMode_draw(Meter* this, int x, int y, int w) {
       memset(data->values, 0, (data->nValues - oldNValues) * sizeof(*data->values));
    }
    const size_t nValues = data->nValues;
```
BenBE, line 310:
```suggestion
   }

   const size_t nValues = data->nValues;
```

BenBE (commented):
Just some minor code style suggestions.

BenBE (approved):

Issue #1318, opened 2023-10-21
Bug: [F2] Setup: settings not saved/persistant


zWhdmB5T:
### my system:
* `OS: openSUSE Leap 15.5 x86_64`
* `htop 3.2.2`

### bug:
1. start htop
2. using [F2] Setup
3. making some changes
4. using [F10] Done
5. using htop
6. exit htop
7. (re-)start htop
*  settings not persistant

### notices:
* there is the directory `~/.config/htop/` with `drwx------` (user) present, but no config file inside

zWhdmB5T:
# addition:
I just noticed, this behaviour stated above only occurs when I start htop by the GUI menu (Xfce)
![Bildschirmfoto_2023-10-21_12-24-30](https://github.com/htop-dev/htop/assets/143814533/478f2746-8a87-4066-896a-59de2c557df1)
. This is the usual way here, so far. But when I start a terminal separately and then start htop explicitly, it works fine (change settings and storing settings). When I then start htop by the GUI, settings (like made before) are read and used. But I cannot persistently change them.

BenBE:
This looks strange.

Can you check the following:
- Use <kbd>Shift</kbd>+<kbd>S</kbd> instead of <kbd>F2</kbd>
- Use <kbd>Q</kbd> instead of <kbd>F10</kbd>
- What is the user shown for the `htop` process?
- What are the owner and permissions for `~/.config`, `~/.config/htop` and `~/.config/htop/htoprc`?
- What is your active (default) `umask`?

In addition:
- Which OS/Distro?
- If you start `htop` from a terminal, are the settings persisted (e.g. change sort order)? What does the created `~/.config/htop/htoprc` look like?

zWhdmB5T:
>     * Use Shift+S instead of F2
shows the screen of  `[F2]`
 
>     * Use Q instead of F10
closes the screen of `[F2]` — but: changes are not persistent!

>     * What is the user shown for the `htop` process?
I did `[F4]` and searched for `htop`: the relating user is my user `gunnersson`

>     * What are the owner and permissions for `~/.config`, `~/.config/htop` and `~/.config/htop/htoprc`?
`drwx------ 42 gunnersson users 4,0K 21. Okt 11:52 .config/`
`drwx------  2 gunnersson users 4,0K 21. Okt 12:09 htop/`
`-rw-------  1 gunnersson users 1,7K 21. Okt 12:17 htoprc`
 
>     * What is your active (default) `umask`?
[...]

PR #1319, opened 2023-10-24
CRT: print a warning when exiting on signal

eworm-de:
Every now and then users complain about settings not being persisted. Most of the time this turns out to be caused by the user: Exiting with any signal causes `htop` not to save settings to `htoprc`.

Make `htop` print a warning when exiting on signal.

### DIFF
--- a/CRT.c
+++ b/CRT.c
@@ -803,8 +803,16 @@ int CRT_scrollWheelVAmount = 10;
 ColorScheme CRT_colorScheme = COLORSCHEME_DEFAULT;
 
 ATTR_NORETURN
-static void CRT_handleSIGTERM(ATTR_UNUSED int sgn) {
+static void CRT_handleSIGTERM(int sgn) {
    CRT_done();
+
+   const char* signal_str = strsignal(sgn);
+   if (!signal_str)
+      signal_str = "unknown reason";
+
+   fprintf(stderr,
+           "A signal %d (%s) was received, exiting without persisting settings to htoprc.\n",
+           sgn, signal_str);
    _exit(0);
 }
 
--- a/CRT.c
+++ b/CRT.c
@@ -806,6 +806,9 @@ ATTR_NORETURN
 static void CRT_handleSIGTERM(int sgn) {
    CRT_done();
 
+   if (!CRT_crashSettings->changed)
+      _exit(0);
+
    const char* signal_str = strsignal(sgn);
    if (!signal_str)
       signal_str = "unknown reason";


BenBE (approved):
LGTM.

Explorer09:
Is it a good idea after all to let htop not save settings after a signal?
I know this is to address a usability issue mention in #1318, but in case the user is existing htop by pressing [X] of the terminal window (i.e. a SIGHUP), the user won't see the added message.

BenBE:
> Is it a good idea after all to let htop not save settings after a signal?

The reasoning behind this is that sending a signal to a process should be an exceptional situation, which might indicate a number of error conditions (SIGPIPE, SIGHUP) or user requests (SIGINT, SIGTERM, SIGKILL). The behaviour so far has been to only store the updated configuration upon normal program termination. Receiving a signal forcefully terminates the program flow and thus can be considered to fall outside that "normal" category (even though we handle those signals to terminate somewhat gracefully).

> I know this is to address a usability issue mention in #1318, but in case the user is existing htop by pressing [X] of the terminal window (i.e. a SIGHUP), the user won't see the added message.

Issue #1318 was just a recent example of people sending signals to `htop` and wondering why their configuration is lost. In 04c769c8 I added a few words to our manpage to clarify the situation. I also stand with the WontFix I added for #1318.

I'm aware that this warning won't be visible when hitting the [X] button on the GUI terminal.

eworm-de:
I have always seen this as a feature: You can play with configuration, and just exit with `Ctrl-c` if settings are messed. 😜 

Even when run from a graphical terminal window you can use a clean exit method, by pressing `q` or `F10`. Even clicking the `Quit` "*button*" works (unless mouse is disabled). And I guess most people try to run a program from (interactive) command line for more verbose output in case or (potential) issue.

I am all for keeping the current behavior of not saving pending settings on signals. The behavior is now documented in 04c769c8445e02c8bdaf52d9276332e565739d33, and the output from this pull request should help a lot as well.

fasterit:
It is wrong behavior of htop to exit e.g. on SIGUSR.
It should ignore these (until we need them for something useful).
SIGHUP should probably save the config and exit.
SIGKILL is for ... killing.

eworm-de:
Well, the changes are in `CRT_handleSIGTERM()`. 😉

Perhaps other signals should be handled different (from another handler).

BenBE:
Just for reference:

```c
static struct sigaction old_sig_handler[32];

static void CRT_installSignalHandlers(void) {
   struct sigaction act;
   sigemptyset(&act.sa_mask);
   act.sa_flags = (int)SA_RESETHAND | SA_NODEFER;
   act.sa_handler = CRT_handleSIGSEGV;
   sigaction(SIGSEGV, &act, &old_sig_handler[SIGSEGV]);
   sigaction(SIGFPE, &act, &old_sig_handler[SIGFPE]);
   sigaction(SIGILL, &act, &old_sig_handler[SIGILL]);
   sigaction(SIGBUS, &act, &old_sig_handler[SIGBUS]);
   sigaction(SIGPIPE, &act, &old_sig_handler[SIGPIPE]);
   sigaction(SIGSYS, &act, &old_sig_handler[SIGSYS]);
   sigaction(SIGABRT, &act, &old_sig_handler[SIGABRT]);

   signal(SIGCHLD, SIG_DFL);
   signal(SIGINT, CRT_handleSIGTERM);
   signal(SIGTERM, CRT_handleSIGTERM);
   signal(SIGQUIT, CRT_handleSIGTERM);
}
```

I.e. `SIGHUP` and `SIGUSR` aren't even considered ATM.

fasterit:
Send a SIGUSR1 to a running `htop` (you can do from inside with <kbd>k</kbd> ...

Explorer09:
@BenBE 
I would like SIGHUP be treated as a "normal" program termination. The reason I mentioned the [X] button on the terminal window is that, many distributions allow launching htop from a graphical start menu / launcher or whatever in the desktop environment, and through that flow, user won't notice there is a terminal window open and had no idea that closing the window would consider as an "abnormal" termination, and thus the issue report.

Allowing users to _not_ save changes on exit could be a feature, but I would like users to be informed and choose to save or discard the changes explicitly.
Telling users that pressing `q` or `F10` saves the changes and that `Ctrl-C` discards the changes would add cognitive load to users. As no terminal apps would have such "magic" behaviors.

BenBE:
I'm absolutely not happy with making SIGHUP an exception …

Explorer09:
> I'm absolutely not happy with making SIGHUP an exception …

And that's the problem. People can expect when the configuration is autosaved but in fact it's not. This can happen even without signals involved. I was thinking about a use case with multiple htop instances running:

1. Start instance 1 of htop. It loads settings A.
2. Go to Setup page, change settings to B, and leave the Setup page.
3. While instance 1 is running, start instance 2 of htop from the same user.

Which settings should instance 2 of htop get, A or B? Here we are not considering whether htop exits normally, just whether the new settings would be saved (so that other instances of htop can see the new settings). Depending on mindsets (or maybe "paradigms"?), the user could expect the settings updated or not updated at all.

BenBE:
> > I'm absolutely not happy with making SIGHUP an exception …
> 
> And that's the problem. People can expect when the configuration is autosaved but in fact it's not. This can happen even without signals involved. I was thinking about a use case with multiple htop instances running:
> 
> 1. Start instance 1 of htop. It loads settings A.
> 2. Go to Setup page, change settings to B, and leave the Setup page.
> 3. While instance 1 is running, start instance 2 of htop from the same user.
> 
> Which settings should instance 2 of htop get, A or B? 

Instance 2 will get Settings A, since instance 1 was not closed yet.

> Here we are not considering whether htop exits normally, just whether the new settings would be saved (so that other instances of htop can see the new settings). Depending on mindsets (or maybe "paradigms"?), the user could expect the settings updated or not updated at all.

With the current implementation (and its documentation), the expectation for the user is quite clear.

My problem with SIGHUP in particular is that in normal circumstances you can get this triggered when SSH'ing into a machine and your connection suddenly interrupted. If you happen to have multiple connections with htop open the stored settings now essentially amount to which connection was killed last. Great RNG you got there …

Explorer09:
> Instance 2 will get Settings A, since instance 1 was not closed yet.

The documented behavior may be different from what users expect based on the experience of other apps. So I wish a more thorough reconsideration of the UX here.

(I know making exceptions on certain signals should be the last resort, but my point is the whole UX being badly done.)

I said there are two mindsets:
1. Settings saved upon closing the Setup screen (or Settings dialog in other apps). This aligns the settings behavior of other GUI apps. In this case instance 2 of htop would get settings B. There would be no way to revert to settings A upon leaving the Setup screen, so users would be advised to backup the `htoprc` file in case things go wrong (htop may implement an automatic backup to save user from the trouble).
2. Settings saved upon closing the program or an explicit "Save" action: This mindset treats settings no different from files opened in an editor. User would save settings by either closing htop or an explicit "Save" action. In this case, instance 2 of htop would get settings A unless the Save action was done in instance 1 to get settings B. There would be a risk that unsaved changes be lost during an abnormal termination (no just signals, but cases like a sudden power cut of the machine), so it is better to keep a temporary file (`htoprc.tmpXXXXXX`) of unsaved changes until the settings are properly saved.

Explorer09:
Just a clarification: I'm not against the idea of adding a warning when `htoprc` is not saved. But the UX of the whole mechanism is not ideal, and a warning would not solve that.

BenBE:
> Just a clarification: I'm not against the idea of adding a warning when `htoprc` is not saved. But the UX of the whole mechanism is not ideal, and a warning would not solve that.

Same here … Even if this might have looked differently.

fasterit:
This code is not being executed.

Reproducer:

1. start `htop`
2. `killall -USR2 htop`

You'll get an output like:
>Benutzerdefiniertes Signal 2

from some other handler.

eworm-de:
Probably not handled at all, but just killed... For me the output regarding signal is from shell I think.

eworm-de:
We would have to add more signals around https://github.com/htop-dev/htop/blob/f9fb93aaf8712934ee0c78d5d887882fd29b6e36/CRT.c#L924 if we want them to be handled.

Explorer09:
@eworm-de Why are we bothered to ignore `SIGUSR1` and `SIGUSR2`, if they don't have a special purpose and can only be sent by the user?
There would be a lot of signals with user-defined purposes and with the default behavior of terminating the application, e.g. `SIGRT*`
And it would a waste of code to set to ignore all of them.

eworm-de:
No idea... I've never before sent `SIGUSR2` to `htop`. I am fine with what ever the consent is here.

One signal we should be aware of is `SIGHUP` as some terminal applications send it on closing terminal. So we consider this a clean exit and want to persist the configuration then?

Explorer09:
> No idea... I've never before sent `SIGUSR2` to `htop`. I am fine with what ever the consent is here.
> 
> One signal we should be aware of is `SIGHUP` as some terminal applications send it on closing terminal. So we consider this a clean exit and want to persist the configuration then?

That was my initial concern when I first see this PR. We cannot put the warning on evey exit, especially we cannot warn if htop exits on signal 9 (SIGKILL). The ideal solution to me is to have htop back up the configuration whenever it's changed, not just before htop exits, but that's another PR to address.

I think we should keep this patch simple: Just put warning on signals that htop would commonly exit on, and not care about the rest.

1 (SIGHUP), 2 (SIGINT, Ctrl-C), 3 (SIGQUIT, Ctrl-\), 15 (SIGTERM, default signal for `kill` command). Anything else?

cgzones:
> 1 (SIGHUP), 2 (SIGINT, Ctrl-C), 3 (SIGQUIT, Ctrl-\), 15 (SIGTERM, default signal for kill command). Anything else?

:+1:

> And that's the problem. People can expect when the configuration is autosaved but in fact it's not. This can happen even without signals involved. I was thinking about a use case with multiple htop instances running:
>
>    Start instance 1 of htop. It loads settings A.
>    Go to Setup page, change settings to B, and leave the Setup page.
>    While instance 1 is running, start instance 2 of htop from the same user.
>
>Which settings should instance 2 of htop get, A or B? Here we are not considering whether htop exits normally, just whether the new settings would be saved (so that other instances of htop can see the new settings). Depending on mindsets (or maybe "paradigms"?), the user could expect the settings updated or not updated at all.

This hit me also several times. Since autosave is the historic mode of htop, what about a new option toggling whether the configuration is auto-saved on exit or only on manual save (with added UI for that)?

fasterit:
DLange would not to have another TUI option, @cgzones. We should just do the right thing™.

@eworm-de: No need to output the message if settings are unchanged.
Otherwise works as intended now.

eworm-de:
Ah, sure... But is there a sane way to access `stettings->changed` inside the signal handler?

eworm-de:
Looks like I can use `CRT_crashSettings->changed`. Is that ok?

Explorer09:
1. I think this commit 29487c2e0d2154a3a2b5e7bcd566a68ec18f6374 ("CRT: handle (ignore) signals USR1 and USR2") should be dropped.
2. Squash the other two commits 401f4395acc74785d4586761c06ac724560f6c7e , 00f6ae494e82db0f3e743a8723f2e42ebd5be8e6 and perhaps reorder the code a little bit. `signal_str` would be unused if the message is not printed. It should be placed after the `CRT_done` line, for example.

eworm-de:
Reordered... But I did not squash and drop - looks like this is kind of controversial here. Who ever merges this: Feel free to squash and/or drop as preferred.

CRT.c
```diff
@@ -803,8 +803,21 @@ int CRT_scrollWheelVAmount = 10;
 ColorScheme CRT_colorScheme = COLORSCHEME_DEFAULT;
 
 ATTR_NORETURN
-static void CRT_handleSIGTERM(ATTR_UNUSED int sgn) {
+static void CRT_handleSIGTERM(int sgn) {
+   const char* signal_str;
+
    CRT_done();
+
+   if (CRT_crashSettings->changed == false)
```
BenBE, line 811:
```suggestion
   if (!CRT_crashSettings->changed)
```

CRT.c
```diff
@@ -921,9 +934,11 @@ static void CRT_installSignalHandlers(void) {
    sigaction(SIGABRT, &act, &old_sig_handler[SIGABRT]);
 
    signal(SIGCHLD, SIG_DFL);
-   signal(SIGINT, CRT_handleSIGTERM);
+   signal(SIGINT,  CRT_handleSIGTERM);
```
BenBE, line 937: No alignment here …

CRT.c
```diff
@@ -935,9 +950,11 @@ void CRT_resetSignalHandlers(void) {
    sigaction(SIGSYS, &old_sig_handler[SIGSYS], NULL);
    sigaction(SIGABRT, &old_sig_handler[SIGABRT], NULL);
 
-   signal(SIGINT, SIG_DFL);
+   signal(SIGINT,  SIG_DFL);
```
BenBE, line 953: Dito.

CRT.c
```diff
@@ -803,8 +803,21 @@ int CRT_scrollWheelVAmount = 10;
...
-static void CRT_handleSIGTERM(ATTR_UNUSED int sgn) {
+static void CRT_handleSIGTERM(int sgn) {
+   const char* signal_str;
+
    CRT_done();
+
+   if (!CRT_crashSettings->changed)
+      _exit(0);
+
+   signal_str = strsignal(sgn);
```
BenBE, line 814:
```suggestion
   CRT_done();

   if (!CRT_crashSettings->changed)
      _exit(0);

   const char* signal_str = strsignal(sgn);
```

eworm-de: Personally I prefer declarations at the top. We do not want this for `htop` code?

BenBE: For htop it is preferred to have declarations initialized with a value when sensibly possible. As C99 allows for declarations to be mixed in where they are needed it's usually preferred to place them where they are first needed. Thus when there's the choice between having uninitialized declarations for variables at the top and C99-style intermixed declarations with a valid assignment, then prefer C99. Overall the scope and lifetime of values should be minimized (sensibly) and uninitialized variables should be avoided.

BenBE (approved):

eworm-de:
Dropped the new signals from this pull request and moved the commit to  #1370.

The remaining commit are valid on their own - the first one adding the message, the second one showing it on pending changes only. For me this is a good reason to keep them separated. But what ever you want... Should I squash them?

cgzones (approved):

PR #1321, opened 2023-11-01
Meter_humanUnit() rewrite and *IOMeter code improvements

Explorer09:
1. Rewrite `Meter_humanUnit()` to make it work with `double` (floating point) values.
2. `DiskIOMeter` and `NetworkIOMeter` code improvements to work with the new `Meter_humanUnit()` function. See commit description for details.

### DIFF
--- a/DiskIOMeter.c
+++ b/DiskIOMeter.c
@@ -36,16 +36,13 @@ static void DiskIOMeter_updateValues(Meter* this) {
 
    static uint64_t cached_last_update;
    uint64_t passedTimeInMs = host->realtimeMs - cached_last_update;
+   bool hasNewData = false;
+   DiskIOData data;
 
    /* update only every 500ms to have a sane span for rate calculation */
    if (passedTimeInMs > 500) {
-      static uint64_t cached_read_total;
-      static uint64_t cached_write_total;
-      static uint64_t cached_msTimeSpend_total;
-      uint64_t diff;
-
-      DiskIOData data;
-      if (!Platform_getDiskIO(&data)) {
+      hasNewData = Platform_getDiskIO(&data);
+      if (!hasNewData) {
          status = RATESTATUS_NODATA;
       } else if (cached_last_update == 0) {
          status = RATESTATUS_INIT;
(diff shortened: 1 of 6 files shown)


pcp/PCPDynamicMeter.c
```diff
@@ -427,22 +427,22 @@ void PCPDynamicMeter_display(PCPDynamicMeter* this, ATTR_UNUSED const Meter* met
...
          case PM_TYPE_64:
             len = conv.dimSpace ?
-               Meter_humanUnit(buffer, atom.ll, sizeof(buffer)) :
+               Meter_humanUnit(buffer, (double)atom.ll, sizeof(buffer)) :
                xSnprintf(buffer, sizeof(buffer), "%lld", (long long) atom.ll);
             break;
          case PM_TYPE_U64:
             len = conv.dimSpace ?
-               Meter_humanUnit(buffer, atom.ull, sizeof(buffer)) :
+               Meter_humanUnit(buffer, (double)atom.ull, sizeof(buffer)) :
```
Explorer09, line 445: Does htop's coding style prefer implicit or explicit casts from integers to floating points?

BenBE: Implicit is sufficient … If no explicit cast is necessary nor does improve readability, it's preferred to go without. Reduce potential noise when reading the code.

natoscott: In this special case though (pmAtomValue), I'd prefer an explicit cast - otherwise it may look to the casual observer that atom.d should have been used here.

DiskIOMeter.c
```diff
@@ -56,41 +55,50 @@ static void DiskIOMeter_updateValues(Meter* this) {
...
 
       cached_last_update = host->realtimeMs;
+   }
 
-      if (status == RATESTATUS_NODATA) {
-         xSnprintf(this->txtBuffer, sizeof(this->txtBuffer), "no data");
-         return;
-      }
+   if (hasNewData) {
+      static uint64_t cached_read_total = 0;
```
natoscott, line 61: No need to set these variables to zero, the compiler is required to do this for static variables.

I know its not new with your code change here (though more static variables are added), but it would be better to move all these local static variables into the DiskIOMeter structure so they only take up space when that meter is in use (and dynamic space, rather than the global data segment).

NetworkIOMeter.c
```diff
@@ -58,20 +56,24 @@ static void NetworkIOMeter_updateValues(Meter* this) {
...
 
       cached_last_update = host->realtimeMs;
+   }
 
-      if (status == RATESTATUS_NODATA) {
-         xSnprintf(this->txtBuffer, sizeof(this->txtBuffer), "no data");
-         return;
-      }
+   if (hasNewData) {
+      static uint64_t cached_rxb_total = 0;
```
natoscott, line 62: Same comments as earlier (no init needed, move to struct) but this time for NetworkIOData.

DiskIOMeter.c
Explorer09: @natoscott
1. I would rather make the initialization explicit, for readability's sake. Not having an initial value implies me that the initial value is unimportant.
2. The commit like 1232e8274b5d584c6032d15c2af7ced212b9f300 ("Always update 'values' member of DiskIOMeter") is to address multiple instances of the meter. The cached values are class-wide, but the `values` members are per-instance, and the distinction has to be made, even if that mean a little waste of space. (Maybe we can introduce a `struct DiskIOMeterCache`) that allocates dynamically when in use, but I'm not working on it on this PR.)

```c
typedef struct {
   uint64_t lastUpdate;
   uint64_t totalBytesRead;
   uint64_t totalBytesWritten;
   uint64_t totalMsTimeSpent;
   uint64_t readDiff;
   uint64_t writeDiff;
   double utilisationDiff;
} DiskIOMeterCache;

static DiskIOMeterCache* cache = NULL;
```

natoscott: In terms of point 1, the initial value of a static is certainly not unimportant - its so important it has been part of the 
C language definition since day 1.  Please leave that code as-is.  Experienced C programmers know this sort of thing so there's no need to explicitly initialize to the already defined initial value.

For point 2, I agree - for now lets keep things static but a cache could be a neat addition IMO.  Again, no need to set the global cache variable to NULL - this is well defined and widely known.

NetworkIOMeter.c
```diff
@@ -28,8 +28,10 @@ static const int NetworkIOMeter_attributes[] = {
 
 static MeterRateStatus status = RATESTATUS_INIT;
 static uint32_t cached_rxb_diff;
+static char cached_rxb_diff_str[6];
 static uint32_t cached_rxp_diff;
 static uint32_t cached_txb_diff;
+static char cached_txb_diff_str[6];
```
BenBE: Alignment on these two cached strings is 8 bytes anyway, thus why not use them completely?

XUtils.h
```diff
@@ -111,4 +111,7 @@ int compareRealNumbers(double a, double b);
    nonnegative. */
 double sumPositiveValues(const double* array, size_t count);
 
+/* IEC unit prefixes */
+static const char unitPrefixes[] = { 'K', 'M', 'G', 'T', 'P', 'E', 'Z', 'Y', 'R', 'Q' };
```
BenBE, line 115:
```c
static const char* unitPrefixes = "KMGTPEZYRQ";
```

Ensures there's a NUL-byte that we can rely on in loops.

NetworkIOMeter.c
Explorer09: I thought 6 bytes are sufficient for Meter_humanUnit() output, since the function would never print more than 5 characters (or 6 characters if we include `\0`).

XUtils.h
Explorer09: Not having a NUL-byte was intentional by me as I use `ARRAYSIZE(unitPrefixes)` to get its length. If I have to pad a NUL byte to it (which wouldn't hurt much anyway), the length would have to change to `strlen(unitPrefixes)`. I just want to make sure you prefer using `strlen` to get its length.

Explorer09: Technical P.S. I define this array as a `static` constant in an header file in order to aid compiler optimization. My testing showed that GCC could hard-code 'K' and 'Q' characters (the first and last of the unit prefixes) in certain branches if the array contents are exposed fully in one translation unit. That's why the array is not an `extern`.

BenBE: Ah, k. In that case, that's fine as is.

DiskIOMeter.c
```diff
@@ -100,13 +110,7 @@ static void DiskIOMeter_updateValues(Meter* this) {
...
    }
 
-   this->values[0] = cached_utilisation_diff;
-   this->total = MAXIMUM(this->values[0], 100.0); /* fix total after (initial) spike */
-
-   char bufferRead[12], bufferWrite[12];
-   Meter_humanUnit(bufferRead, cached_read_diff, sizeof(bufferRead));
-   Meter_humanUnit(bufferWrite, cached_write_diff, sizeof(bufferWrite));
-   snprintf(this->txtBuffer, sizeof(this->txtBuffer), "r:%siB/s w:%siB/s %.1f%%", bufferRead, bufferWrite, cached_utilisation_diff);
+   snprintf(this->txtBuffer, sizeof(this->txtBuffer), "r:%siB/s w:%siB/s %.1f%%", cached_read_diff_str, cached_write_diff_str, cached_utilisation_diff);
```
BenBE, line 113: Any reason we are not using `xSnprintf` here?

Meter.c
```diff
@@ -51,32 +51,39 @@ Meter* Meter_new(const Machine* host, unsigned int param, const MeterClass* type
    return this;
 }
 
-int Meter_humanUnit(char* buffer, unsigned long int value, size_t size) {
-   const char* prefix = "KMGTPEZY";
-   unsigned long int powi = 1;
-   unsigned int powj = 1, precision = 2;
+/* Converts 'value' in kibibytes into a human readable string.
+   Example output strings: "0K", "1023K", "98.7M" and "1.23G" */
+int Meter_humanUnit(char* buffer, double value, size_t size) {
```
BenBE, line 56: What about `xSnprintf` usage here? Also I'd suggest adding braces for all `if` statements.

natoscott (approved):
@Explorer09 @BenBE LGTM, thanks.

BenBE (approved):
LGTM.

PR #1322, opened 2023-11-06
Permit zero 'total' in bar meter mode; remove assertion

Explorer09:
The `Meter.total` value may be 0 on BatteryMeter, SwapMeter, linux/HugePageMeter, linux/ZramMeter and others. In case of a zero total value, print an empty bar in bar meter mode.

Also, GraphMeterMode_draw() will no longer write back the `total` value
if it is zero.


natoscott (approved):
LGTM, thanks @Explorer09

Issue #1317, opened 2023-10-18
Feature request: formats for long cpu times


leahneukirchen:
Currently the `TIME+` columns stops at displaying hours.  When running VM for a long time, it could be useful to wrap over to displaying days after some time.  `74d20h` is more helpful output than `1778h45m`.

leahneukirchen:
Fixed in #1325.

PR #1323, opened 2023-11-08
Change time output formatting.

chestnykh:
In https://github.com/htop-dev/htop/issues/1317
there is the request to make these changes.
Now the output format becomes the same for
any `days` value in [1,365). The `hours > 0`
and `days >= 365` cases has left unchanged.

### DIFF
--- a/Row.c
+++ b/Row.c
@@ -351,7 +351,7 @@ void Row_printTime(RichString* str, unsigned long long totalHundredths, bool col
          len = xSnprintf(buffer, sizeof(buffer), "%1dd ", daysLeft);
          RichString_appendnAscii(str, dayColor, buffer, len);
       }
-   } else if (days >= 100) {
+   } else if (days >= 1) {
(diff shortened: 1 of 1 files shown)


BenBE:
I think the overall time display code should get a rework. What I don't quite like is how e.g. the allotted space for hours depends on the number of digits in the remaining hours, thus alternating the width of the day column throughout the day. Also considerations regarding a wider range of values (ms through s, m, h, d and y should be taken.

Thus while technically correct, I don't think we should go with this patch and cleanup things overall. @cgzones Didn't you recently work at something similar?

chestnykh:
Ok, tell me finally should i decline this patch or not

Explorer09:
@chestnykh Your code is correct, but looks like a quick hack. We would like to see a more thorough rework of the code.

I think I can try this one. I got the point @BenBE is talking about. Just wait until I start a branch for this.

PR #1325, opened 2023-11-09
Rewrite Row_printTime() with various improvements

Explorer09:
* Change variable types in the function from signed to unsigned.
* Change variable naming for consistency: `totalMinutes`, `totalHours`, `totalDays` for total values and `minutes`, `hours`, `days`, etc. for remainder values.
* Use unsigned long long type for `years` value to prevent a potential overflow.
* The time units are now evaluated from small to large, in the hopes that small values would print faster. (There is no performance test for this claim.)
* The days unit is now printed when total hours is at least 24. (Issue htop-dev#1317) There is room in the new code to revert to the old threshold (totalHours < 2400).

### DIFF
--- a/Row.c
+++ b/Row.c
@@ -314,79 +314,62 @@ void Row_printTime(RichString* str, unsigned long long totalHundredths, bool col
    char buffer[10];
    int len;
 
-   unsigned long long totalSeconds = totalHundredths / 100;
-   unsigned long long hours = totalSeconds / 3600;
-   unsigned long long days = totalSeconds / 86400;
-   int minutes = (totalSeconds / 60) % 60;
-   int seconds = totalSeconds % 60;
-   int hundredths = totalHundredths - (totalSeconds * 100);
-
    int yearColor = coloring ? CRT_colors[LARGE_NUMBER]      : CRT_colors[PROCESS];
    int dayColor  = coloring ? CRT_colors[PROCESS_GIGABYTES] : CRT_colors[PROCESS];
    int hourColor = coloring ? CRT_colors[PROCESS_MEGABYTES] : CRT_colors[PROCESS];
-   int baseColor  = CRT_colors[PROCESS];
+   int baseColor = CRT_colors[PROCESS];
 
-   if (days >= /* Ignore leap years */365) {
-      int years = days / 365;
-      int daysLeft = days - 365 * years;
+   unsigned long long totalSeconds = totalHundredths / 100;
+   unsigned long long totalMinutes = totalSeconds / 60;
+   unsigned long long totalHours = totalMinutes / 60;
+   unsigned int seconds = totalSeconds % 60;
+   unsigned int minutes = totalMinutes % 60;
 
-      if (years >= 10000000) {
-         RichString_appendnAscii(str, yearColor, "eternity ", 9);
-      } else if (years >= 1000) {
-         len = xSnprintf(buffer, sizeof(buffer), "%7dy ", years);
-         RichString_appendnAscii(str, yearColor, buffer, len);
-      } else if (daysLeft >= 100) {
-         len = xSnprintf(buffer, sizeof(buffer), "%3dy", years);
-         RichString_appendnAscii(str, yearColor, buffer, len);
-         len = xSnprintf(buffer, sizeof(buffer), "%3dd ", daysLeft);
-         RichString_appendnAscii(str, dayColor, buffer, len);
-      } else if (daysLeft >= 10) {
-         len = xSnprintf(buffer, sizeof(buffer), "%4dy", years);
-         RichString_appendnAscii(str, yearColor, buffer, len);
-         len = xSnprintf(buffer, sizeof(buffer), "%2dd ", daysLeft);
-         RichString_appendnAscii(str, dayColor, buffer, len);
-      } else {
-         len = xSnprintf(buffer, sizeof(buffer), "%5dy", years);
-         RichString_appendnAscii(str, yearColor, buffer, len);
-         len = xSnprintf(buffer, sizeof(buffer), "%1dd ", daysLeft);
-         RichString_appendnAscii(str, dayColor, buffer, len);
-      }
-   } else if (days >= 100) {
-      int hoursLeft = hours - days * 24;
-
-      if (hoursLeft >= 10) {
-         len = xSnprintf(buffer, sizeof(buffer), "%4llud", days);
-         RichString_appendnAscii(str, dayColor, buffer, len);
-         len = xSnprintf(buffer, sizeof(buffer), "%2dh ", hoursLeft);
-         RichString_appendnAscii(str, hourColor, buffer, len);
-      } else {
-         len = xSnprintf(buffer, sizeof(buffer), "%5llud", days);
-         RichString_appendnAscii(str, dayColor, buffer, len);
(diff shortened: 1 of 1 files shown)


Explorer09:
Examples of all formats:
```
 0:00.00 - 59:59.99
 1h00:00 - 23h59:59
1d 0h 0m - 9d23h59m
  10d 0h -  364d23h
  1y  0d - 999y364d
   1000y - 9999999y
eternity
```

Row.c
```diff
@@ -314,79 +314,62 @@ void Row_printTime(RichString* str, unsigned long long totalHundredths, bool col
...
-      }
-   } else if (hours > 0) {
-      len = xSnprintf(buffer, sizeof(buffer), "%2lluh", hours);
+   int baseColor = CRT_colors[PROCESS];
+
+   unsigned long long totalSeconds = totalHundredths / 100;
+   unsigned long long totalMinutes = totalSeconds / 60;
+   unsigned long long totalHours = totalMinutes / 60;
+   unsigned int seconds = totalSeconds - totalMinutes * 60;
+   unsigned int minutes = totalMinutes - totalHours * 60;
```
BenBE, line 326: Given that most CPUs calculate mod and div in one instruction (or have a function that does so), we should be able to just use the following without loosing any performance:
```suggestion
   unsigned int seconds = totalSeconds % 60;
   unsigned int minutes = totalMinutes % 60;
```
This comes, because we are doing the related division anyway, thus this mod should not introduce any overhead apart from not throwing away the result …

Row.c
```diff
@@ -314,79 +314,62 @@ void Row_printTime(RichString* str, unsigned long long totalHundredths, bool col
...
-      len = xSnprintf(buffer, sizeof(buffer), "%02d:%02d ", minutes, seconds);
+      len = xSnprintf(buffer, sizeof(buffer), "%02u:%02u ", minutes, seconds);
       RichString_appendnAscii(str, baseColor, buffer, len);
-   } else {
-      len = xSnprintf(buffer, sizeof(buffer), "%2d:%02d.%02d ", minutes, seconds, hundredths);
+      return;
+   }
+
+   unsigned long long totalDays = totalHours / 24;
+   unsigned int hours = totalHours - totalDays * 24;
```
BenBE, line 343: Same remark as above re mod/div …

Row.c
```diff
@@ -314,79 +314,62 @@ void Row_printTime(RichString* str, unsigned long long totalHundredths, bool col
...
+   if (totalDays < /* Ignore leap years */365) {
+      len = xSnprintf(buffer, sizeof(buffer), "%4llud", totalDays);
+      RichString_appendnAscii(str, dayColor, buffer, len);
+      len = xSnprintf(buffer, sizeof(buffer), "%2uh ", hours);
+      RichString_appendnAscii(str, hourColor, buffer, len);
+      return;
+   }
+
+   unsigned long long years = totalDays / 365;
+   unsigned int days = totalDays - years * 365;
```
BenBE, line 362: Same re mod/div…

Row.c
```diff
@@ -314,79 +314,62 @@ void Row_printTime(RichString* str, unsigned long long totalHundredths, bool col
...
+   }
+
+   unsigned long long totalDays = totalHours / 24;
+   unsigned int hours = totalHours - totalDays * 24;
+   if (totalDays < 10) {
+      len = xSnprintf(buffer, sizeof(buffer), "%1llud", totalDays);
+      RichString_appendnAscii(str, dayColor, buffer, len);
+      len = xSnprintf(buffer, sizeof(buffer), "%2uh", hours);
+      RichString_appendnAscii(str, hourColor, buffer, len);
+      len = xSnprintf(buffer, sizeof(buffer), "%2um ", minutes);
```
BenBE, line 349: Can we avoid spaces here?
```suggestion
      len = xSnprintf(buffer, sizeof(buffer), "%02uh", hours);
      RichString_appendnAscii(str, hourColor, buffer, len);
      len = xSnprintf(buffer, sizeof(buffer), "%02um ", minutes);
```

Row.c
```diff
@@ -314,79 +314,62 @@ void Row_printTime(RichString* str, unsigned long long totalHundredths, bool col
...
       RichString_appendnAscii(str, baseColor, buffer, len);
+      return;
+   }
+   if (totalDays < /* Ignore leap years */365) {
+      len = xSnprintf(buffer, sizeof(buffer), "%4llud", totalDays);
+      RichString_appendnAscii(str, dayColor, buffer, len);
+      len = xSnprintf(buffer, sizeof(buffer), "%2uh ", hours);
+      RichString_appendnAscii(str, hourColor, buffer, len);
```
BenBE, line 357: Avoid spaces in the middle …
```suggestion
      len = xSnprintf(buffer, sizeof(buffer), "%02uh ", hours);
      RichString_appendnAscii(str, hourColor, buffer, len);
```

Row.c
```diff
@@ -314,79 +314,62 @@ void Row_printTime(RichString* str, unsigned long long totalHundredths, bool col
...
+      return;
+   }
+
+   unsigned long long years = totalDays / 365;
+   unsigned int days = totalDays - years * 365;
+   if (years < 1000) {
+      len = xSnprintf(buffer, sizeof(buffer), "%3lluy", years);
+      RichString_appendnAscii(str, yearColor, buffer, len);
+      len = xSnprintf(buffer, sizeof(buffer), "%3ud ", days);
+      RichString_appendnAscii(str, dayColor, buffer, len);
```
BenBE, line 367: Avoid spaces in the middle …
```suggestion
      len = xSnprintf(buffer, sizeof(buffer), "%03ud ", days);
      RichString_appendnAscii(str, dayColor, buffer, len);
```

BenBE (commented):
Some suggestions. No perf testing done re mod/div vs. div & mul+sub.

How much of this code could we potentially re-use for #1288 time printing

Row.c
Explorer09: Actually I did test using the `%` operator instead of subtraction. Using `%` does produce smaller code in `gcc -Os` but it's not always the case for `-O2` mode. It's sad that GCC isn't smart enough to detect the two versions are equivalent.

I'm okay to use `%` here for the sake of readability. Just to note that GCC has some weird behaviors on the optimization.

BenBE: I'm okay with both versions. Just wanted to bring this to your attention in case you missed that alternative.

Explorer09:
With zero padding the formats will now look like this:
```
 0:00.00 - 59:59.99
 1h00:00 - 23h59:59
1d00h00m - 9d23h59m
  10d00h -  364d23h
  1y000d - 999y364d
   1000y - 9999999y
eternity - eternity
```

> How much of this code could we potentially re-use for #1288 time printing

Unfortunately I don't think there is potential for code reuse. The time factoring logic would be slightly different between two functions, but at least we can keep variable naming consistent.

BenBE (approved):

natoscott (approved):
Looks good, though I think that final branch and 'eternity' string is a bit over the top :P (could've just gone up to 9999999y and used that for anything above 1million years).

Explorer09:
@natoscott Just to say the "eternity" idea was not mine. It was there since commit ec809b7f71f21c93c96398caa552eab486c3423d. Credit @BenBE for that one.

Issue #1330, opened 2023-11-19
Name of container that contains a process


callumgare:
Most of my workload is dockerised and so it would be super useful to have a new column to show the name of the docker container that a process belongs to (if it does indeed belong to one).

Whenever there's a resource usage issue like ram or cpu having a higher utilisation than expected my first port of call is htop to see what process is responsible. Generally it'll be something in a docker container since most stuff on my machines run in docker containers. So once I know what process or processes seem to be the cause of the trouble I then have to trace that back to the docker container they live in (unless it happens to be obvious and unambiguous from the process name). 

I do that by copying the PIDs of the relevant processes from htop then running:
```shell
pid=3728712 && docker ps --filter "ID=$(cat /proc/$pid/cgroup | grep -o -E '[0-9a-f]{10,}')" --format "{{.Names}}"
```
to find the name of the docker container it belongs too.

It would be great to just have "container name" as an optional column in htop. As well as being able to troubleshoot quicker it would also just be nice to have when everything going okay as well. I'd be able to look though htop and just from a glance get a sense of what containers make up the bulk of resource usage. I know [ctop](https://github.com/bcicen/ctop) sort of does this but I'd like to be able to see that at the process level and also to compare it to processes not in docker containers.

I had a look for similar issues and found https://github.com/htop-dev/htop/issues/1328 which is great to have but doesn't solve this situation wanting to easily get the human readable container name for a process (as well as to be able to super easily glance though the list and quickly identify what processes are containerised and which containers the most resource heavy processes live in).

Thanks for your consideration!

BenBE:
As you noticed (re #1328 / #1329) there's a column called `CGROUP (compressed)` which reads the cgroup name for the process and does some heuristics on the cgroup name. This works for many container software quite nice (nspawn, lxc, with #1329 also podman), and a similar thing can be added for Docker too, as there's a [naming pattern](https://docs.docker.com/config/containers/runmetrics/). Now for the caveats: For cgroup v1, this will show Docker containers as LXC (which technically they are) and for cgroup v2 will only provide an container ID (which should be shortened to a sensible level for display). But that's just part of the issue at hand.

What this issue is actually about is a separate column for this, and the existing code could be updated to filter out any non-container parts of the cgroup path. This will create quite a bit of code duplication for that column though …

PR #1334, opened 2023-11-20
Guess container name from cgroup

BenBE:
This is a new column based on the heuristics employed by the CGroup compression logic.

This probably should be combined with the logic for determining Jails on *BSD and other virtualization technics we have available. This basically also means, this column might be better suited to be created as a platform independent info.

Leaving this here as a PoC for #1330. That code is by no means thought through; basically a quick shot. Run at your own risk.

### DIFF
--- a/linux/CGroupUtils.c
+++ b/linux/CGroupUtils.c
@@ -357,3 +357,144 @@ char* CGroup_filterName(const char* cgroup) {
    s.buf[s.size] = '\0';
    return s.buf;
 }
+
+static bool CGroup_filterContainer_internal(const char* cgroup, StrBuf_state* s, StrBuf_putc_t w) {
+   const char* str_lxc_payload_legacy = "lxc.payload";
+   const char* str_lxc_payload_prefix = "lxc.payload.";
+
+   const char* str_nspawn_scope_prefix = "machine-";
+   const char* str_nspawn_monitor_label = "/supervisor";
+   const char* str_nspawn_payload_label = "/payload";
+
+   const char* str_pod_scope_prefix = "libpod-";
+
+   const char* str_scope_suffix = ".scope";
+
+   while (*cgroup) {
+      if ('/' == *cgroup) {
+         while ('/' == *cgroup)
+            cgroup++;
+
+         continue;
+      }
+
+      const char* labelStart = cgroup;
+      const char* nextSlash = String_strchrnul(labelStart, '/');
+      const size_t labelLen = nextSlash - labelStart;
+
+      if (Label_checkPrefix(labelStart, labelLen, str_lxc_payload_prefix)) {
+         const size_t cgroupNameLen = labelLen - strlen(str_lxc_payload_prefix);
+
+         if (!StrBuf_putsz(s, w, "/lxc:"))
+            return false;
+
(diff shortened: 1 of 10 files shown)


natoscott:
> This is a new column based on the heuristics employed by the CGroup compression logic.
> 
> This probably should be combined with the logic for determining Jails on *BSD and other virtualization technics we have available. This basically also means, this column might be better suited to be created as a platform independent info.
> 
> Leaving this here as a PoC for #1330. That code is by no means thought through; basically a quick shot. Run at your own risk.

Looking at FreeBSDProcessTable_readJailName and related code, I don't think you need to worry about Jails here - it's likely to be completely separate, forever, totally different interfaces and no sign of need to squash those names.  

IOW, keep this confined to Linux-specific code IMO.  Other than that, conceptually seems fine to me.  I guess we could go for an object-oriented-ish function-pointer kind of interface so each container technology has set string filtering functions and callbacks get registers ... but, not sure, that may be overkill / uglier.

BenBE:
> Looking at FreeBSDProcessTable_readJailName and related code, I don't think you need to worry about Jails here - it's likely to be completely separate, forever, totally different interfaces and no sign of need to squash those names.

I'm aware re name-untangling. I thought more on the lines of creating the `CONTAINER` column from that value in a simular format as e.g. `/jail:fubar`. I'm not sure whether FreeBSD container nesting is visible the way it is on Linux (e.g. `/snc:foo/lxc:bar/pod:buzz` is totally fine and visible on Linux).

The reason I was asking about (Free)BSD, was more along the lines of: Do we need that container field in the platform-agnostic code or do we keep it with Linux (which will cause refactoring work later on).

> IOW, keep this confined to Linux-specific code IMO. Other than that, conceptually seems fine to me.

Okay. No `CONTAINER` column on BSD/Solaris/* then …

> I guess we could go for an object-oriented-ish function-pointer kind of interface so each container technology has set string filtering functions and callbacks get registers ... but, not sure, that may be overkill / uglier.

Currently, except some minor changes, the code for `CGroup_filterName_internal` and `CGroup_filterContainer_internal` are identical, sans their output format (and some unused constants): You can arrive at the new code basically by deleting lines from the first one – would be nice for maintenance to keep it this way.

Regarding filters for the various bits: As most container stuff seems to be at scope level (LXC and SNC need the next label to distinguish monitor/payload processes), with the format usually being something like `tech-ID.scope`, it should be easy to canonicalize that sort of check. A way that works for both namespace folding and container extraction would be great though.

linux/CGroupUtils.c
```diff
@@ -357,3 +357,144 @@ char* CGroup_filterName(const char* cgroup) {
...
+   const char* str_lxc_payload_legacy = "lxc.payload";
+   const char* str_lxc_payload_prefix = "lxc.payload.";
+
+   const char* str_nspawn_scope_prefix = "machine-";
+   const char* str_nspawn_monitor_label = "/supervisor";
+   const char* str_nspawn_payload_label = "/payload";
+
+   const char* str_pod_scope_prefix = "libpod-";
+
+   const char* str_scope_suffix = ".scope";
```
natoscott, line 371: @BenBE should these (above 10 lines) all be 'static const char*' so they're not on the stack?  I don't see them being modified but maybe I missed it.

natoscott (approved):
@BenBE LGTM, one minor suggestion.

PR #1336, opened 2023-11-24
Linux + PCP: Add private memory size column

BowDown097:
Private memory size is a metric supported by other tools like [System Monitoring Center](https://github.com/hakandundar34coding/system-monitoring-center) and [System Informer](https://systeminformer.sourceforge.io/). It is also a decently close measurement of the "real" memory usage of a process, like PSS.

Note that it is calculated by subtracting the shared memory size from the RSS, which won't always provide the exact real value, but it is a very close estimation an overwhelming majority of the time. Getting the real value appears to involve some querying of the kernel which is very slow for heavy processes, which is just unnecessary.

Would appreciate some testing for PCP, I can't do it because I'm not on systemd, which it requires. Not sure if what I did for it works properly, considering I have to multiply the shared memory by the page size in the Linux code.

### DIFF
--- a/linux/LinuxProcessTable.c
+++ b/linux/LinuxProcessTable.c
@@ -701,9 +701,9 @@ static bool LinuxProcessTable_readStatmFile(LinuxProcess* process, openat_arg_t 
    if (r == 7) {
       process->super.m_virt *= host->pageSizeKB;
       process->super.m_resident *= host->pageSizeKB;
-   }
 
-   process->m_priv = process->super.m_resident - (process->m_share * host->pageSizeKB);
+      process->m_priv = process->super.m_resident - (process->m_share * host->pageSizeKB);
+   }
 
    return r == 7;
 }
--- a/linux/LinuxProcessTable.c
+++ b/linux/LinuxProcessTable.c
@@ -703,6 +703,8 @@ static bool LinuxProcessTable_readStatmFile(LinuxProcess* process, openat_arg_t 
       process->super.m_resident *= host->pageSizeKB;
    }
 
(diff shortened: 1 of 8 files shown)


linux/LinuxProcessTable.c
```diff
@@ -703,6 +703,8 @@ static bool LinuxProcessTable_readStatmFile(LinuxProcess* process, openat_arg_t
       process->super.m_resident *= host->pageSizeKB;
    }
 
+   process->m_priv = process->super.m_resident - (process->m_share * host->pageSizeKB);
+
```
BenBE, line 707:
```suggestion

      process->m_priv = process->super.m_resident - (process->m_share * host->pageSizeKB);
   }

```

BenBE (changes requested):

natoscott (approved):
@BowDown097 looks good - I had to stare at the LinuxProcessTable.c units handling a little, but I think you've got it spot on.

PCP doesn't rely on systemd FWIW (predating it by at least 10 years) - its probably just the Debian packaging enforcing that nowadays for simplicity sake.  I'm sure that packaging could be made to work either way if anyone's keen.

I see from your diff that we're missing a CGROUP entry in pcp/ProcessField.h now - I'll follow up with that addition.

pcp/ProcessField.h
```diff
@@ -45,6 +45,7 @@ in the source distribution for its full text.
    SECATTR = 123,                \
    AUTOGROUP_ID = 127,           \
    AUTOGROUP_NICE = 128,         \
+   M_PRIV = 129,                 \
```
natoscott, line 48: Please update this to 130 now (see #1339 for context).

natoscott (changes requested):

PR #1337, opened 2023-11-24
Fix code style

BenBE:
Some code style updates …


Explorer09:
@fasterit Nope. That won't work. The `#define _GNU_SOURCE` line has to be before any include line of the standard header (`<string.h>` for this particular case), or else the header guard of glibc's `<string.h>` would prevent the `strchrnul()` from making available.

The two "solutions" I mentioned above are the proper ways to do it. If it's me I would pick option 1, as `config.h` is required for many .c files already in htop. It won't be a difficult policy to mandate every .c file include `config.h`.

fasterit:
Wtf.

I'd favor your option 1 then, too. The lesser evil.

Thank you very much for the analysis @Explorer09 !

/DLange

fasterit:
Seems to work for all but the PCP build...

```In file included from pcp/Metric.c:18:
./XUtils.h:13:2: error: #error "Must have #include \"config.h\" line at the top of the file that includes these XUtils helper functions"
   13 | #error "Must have #include \"config.h\" line at the top of the file that includes these XUtils helper functions"
      |  ^~~~
```

While it is included, of course.

BenBE:
What a fitting kind of issue to have as PR 1337 … ;-)

BenBE:
@fasterit The PCP failure comes from the brain-dead design of PCP that requires you to undefine `PACKAGE_NAME` because they are re-using that identifier in their own headers. I'll have a look to update the check to use an `HTOP_` specific define instead.

XUtils.h
```diff
@@ -3,13 +3,21 @@
...
 Released under the GNU GPLv2+, see the COPYING file
 in the source distribution for its full text.
 */
 
+#ifndef PACKAGE
+// strchrnul() needs _GNU_SOURCE defined, see PR #1337 for details
+#error "Must have #include \"config.h\" line at the top of the file that includes these XUtils helper functions"
+#endif
+
+// Workaround to make IWYU happy regarding the used symbols below; translates to noop.
```
Explorer09, line 16: I'm curious. What error would IWYU generate if we don't have this workaround and just have the `#include "config.h"` line removed?

And I wonder if this line works:
`// IWYU pragma: no_include "config.h"`

configure.ac
```diff
@@ -242,7 +243,7 @@ AC_LINK_IFELSE([
 #endif
          ]])],
          AC_MSG_RESULT([assume yes (cross compiling)]),
-	 flag_finite_math_only=yes)
+      flag_finite_math_only=yes)
```
Explorer09, line 246: This indent was not what I suggested before (it's one less level than I originally had)

linux/SystemdMeter.c
```diff
@@ -316,12 +318,12 @@ static int zeroDigitColor(unsigned int value) {
...
-      case INVALID_VALUE:
-         return CRT_colors[METER_VALUE_ERROR];
-      default:
-         return CRT_colors[METER_VALUE];
+   case 0:
+      return CRT_colors[METER_VALUE_NOTICE];
+   case INVALID_VALUE:
+      return CRT_colors[METER_VALUE_ERROR];
+   default:
+      return CRT_colors[METER_VALUE];
```
Explorer09, line 326: The indent style of `case` statements inside `switch` are not consistent even after this PR.
Do `case` statements indent one level after `switch`...
```c
switch (value) {
   case 0:
      // Do something
      break;
   case 1:
      // Do something else
      break;
   default:
      // Yet something else
}
```
Or do they stay the same level as the `switch`...
```c
switch (value) {
case 0:
   // Do something
   break;
case 1:
   // Do something else
   break;
default:
   // Yet something else
}
```

dragonflybsd/DragonFlyBSDProcessTable.c
```diff
@@ -249,44 +249,56 @@ void ProcessTable_goThroughEntries(ProcessTable* super) {
       // would be nice if we could store multiple states in proc->state (as enum) and have writeField render them
       /* Taken from: https://github.com/DragonFlyBSD/DragonFlyBSD/blob/c163a4d7ee9c6857ee4e04a3a2cbb50c3de29da1/sys/sys/proc_common.h */
       switch (kproc->kp_stat) {
```
Explorer09, line 251: Likewise. The indent style of `case` statements inside `switch` are not consistent even after this PR.

pcp/PCPProcessTable.c
```diff
@@ -389,10 +389,10 @@ static bool PCPProcessTable_updateProcesses(PCPProcessTable* this) {
 
       PCPProcessTable_updateMemory(pp, pid, offset);
 
-      if ((flags & PROCESS_FLAG_LINUX_SMAPS) &&
-          (Process_isKernelThread(proc) == false)) {
-         if (Metric_enabled(PCP_PROC_SMAPS_PSS))
+      if ((flags & PROCESS_FLAG_LINUX_SMAPS) && (Process_isKernelThread(proc) == false)) {
```
Explorer09, line 392: A separate commit to fix this one?

```suggestion
      if ((flags & PROCESS_FLAG_LINUX_SMAPS) && !Process_isKernelThread(proc)) {
```

XUtils.h
BenBE: For `XUtils.c`/`XUtils.h` it would generate "must include `config.h`" message due to the various defines we use from that file in the code below. Strictly speaking that include is covered by the `IWYU pragma: keep` in `XUtils.c`.

FWIW, I mostly added that explicit include for overall consistency; i.e. having the headers actually have an include for things they rely on. As noted in the comment: By the point that include is made it should generate a no-op, or the pre-processor is borked …

pcp/PCPProcessTable.c
BenBE: Can do …

configure.ac
BenBE: Was effectively `\t\x20` before, miscounted the level from the parens, should be one level deeper. Will update.

XUtils.h
BenBE: Used your suggested `IWYU pragma: no_include "config.h"` as a marker.

dragonflybsd/DragonFlyBSDProcessTable.c
BenBE: They should be now … ;-) Otherwise drop me a note and I'll re-check.

CRT.c
```diff
@@ -1113,8 +1113,8 @@ void CRT_setColors(int colorScheme) {
       for (short int j = 0; j < 8; j++) {
          if (ColorIndex(i, j) != ColorIndexGrayBlack && ColorIndex(i, j) != ColorIndexWhiteDefault) {
             short int bg = (colorScheme != COLORSCHEME_BLACKNIGHT)
-                     ? (j == 0 ? -1 : j)
-                     : j;
+               ? (j == 0 ? -1 : j)
+               : j;
```
Explorer09, line 1117: A little logic simplification can be done here (as a separate commit).
```c
short int bg = colorScheme != COLORSCHEME_BLACKNIGHT && j == 0 ? -1 : j;
```

ScreensPanel.c
```diff
@@ -196,91 +195,76 @@ static HandlerResult ScreensPanel_eventHandlerNormal(Panel* super, int ch) {
...
-      {
          Panel_moveSelectedDown(super);
          shouldRebuildArray = true;
          result = HANDLED;
          break;
-      }
-      case KEY_F(9):
-      //case KEY_DC:
-      {
+      case KEY_F(9): //case KEY_DC:
```
Explorer09, line 242: Should investigate why the `case KEY_DC:` line is commented out here.
That comment on the same line as `case KEY_F(9):` is surely wrong.

BenBE: FYI: This was introduced by commit 72ba20fa5 (committed by @natoscott) already in the original commit. Didn't investigate further in the upstream code on Hisham's old branch where this has been adapted from.

Explorer09:
Didn't see any problems now with this PR. Just a final note.

From the commit description:
> Also add a #error stanza to XUtils.h in case somebody forgets the beautiful mess GNU forces on us.

It's not just GNU, but all [feature test macros](https://man7.org/linux/man-pages/man7/feature_test_macros.7.html) coming from POSIX and any of its extensions are required to be defined this way. This "feature" is kind of annoying for developing in the C language (unlike C++, which comes with namespaces to clean up the duplicated APIs and symbol name conflicts). I hate it too, but we have to live with it.

Feature test macros was one of the reasons GNU autotools introduced `config.h`, although it only solves half of the problem. The use of `config.h` is incompatible with the IWYU's coding paradigm (that's the other half of the problem).

Explorer09:
@fasterit The real mess is not providing `strchrnul()` after the re-include of `<string.h>`. While the `_GNU_SOURCE` was there to avoid potential symbol name conflicts, the sin is putting the macro check under the include guard and only check it once.

I know the header can work like `<assert.h>` where the `NDEBUG` macro is checked on every `#include <assert.h>` line, thus making the function available after re-inclusion of the header.

Explorer09:
One more thing. Because of the feature test macro mess. I'm now thinking that the `#include "config.h"` lines might be removed from all "module" headers of htop. Interestingly, not all htop header files have this include line. I wonder how and why IWYU suggested including `config.h` for these headers in the first place (specifically, what tokens or symbols made IWYU suggest the `config.h`)

```text
./Object.h:11:#include "config.h" // IWYU pragma: keep
./Table.h:11:#include "config.h" // IWYU pragma: keep
./Machine.h:11:#include "config.h" // IWYU pragma: keep
./MainPanel.h:11:#include "config.h" // IWYU pragma: keep
./ProvideTerm.h:10:#include "config.h" // IWYU pragma: keep
./RichString.h:10:#include "config.h" // IWYU pragma: keep
./pcp/InDomTable.h:11:#include "config.h" // IWYU pragma: keep
./pcp/PCPProcessTable.h:10:#include "config.h" // IWYU pragma: keep
./pcp/PCPProcess.h:12:#include "config.h" // IWYU pragma: keep
./pcp/PCPMachine.h:10:#include "config.h" // IWYU pragma: keep
./Panel.h:10:#include "config.h" // IWYU pragma: keep
./ProvideCurses.h:11:#include "config.h" // IWYU pragma: keep
./HeaderLayout.h:10:#include "config.h" // IWYU pragma: keep
./linux/Platform.h:10:#include "config.h" // IWYU pragma: keep
./linux/LinuxProcess.h:11:#include "config.h" // IWYU pragma: keep
./linux/LinuxMachine.h:10:#include "config.h" // IWYU pragma: keep
./linux/LinuxProcessTable.h:10:#include "config.h" // IWYU pragma: keep
./Settings.h:10:#include "config.h" // IWYU pragma: keep
./Compat.h:10:#include "config.h" // IWYU pragma: keep
./Meter.h:10:#include "config.h" // IWYU pragma: keep
./ProcessTable.h:10:#include "config.h" // IWYU pragma: keep
./solaris/Platform.h:12:#include "config.h" // IWYU pragma: keep
./solaris/SolarisProcessTable.h:11:#include "config.h" // IWYU pragma: keep
./solaris/SolarisProcess.h:11:#include "config.h" // IWYU pragma: keep
./solaris/SolarisMachine.h:11:#include "config.h" // IWYU pragma: keep
./Scheduling.h:10:#include "config.h" // IWYU pragma: keep
./Action.h:10:#include "config.h" // IWYU pragma: keep
./Affinity.h:11:#include "config.h" // IWYU pragma: keep
./CRT.h:10:#include "config.h" // IWYU pragma: keep
./SignalsPanel.h:10:#include "config.h" // IWYU pragma: keep
```

BenBE:
@Explorer09 You can use the script `iwyu/run_iwyu.sh` to run `iwyu`. If you remove `-Xiwyu --no_comments` from the `CFLAGS` you get a short list of which symbols IWYU found from each header.

On the note of `config.h` in header files: Should probably also be a new note in `docs/styleguide.md`.

natoscott:
> EDIT: There are two solutions.
> 

There's a third simpler option which I see commonly used in OSS projects - pass -D_GNU_SOURCE=1 to the compiler (via the autoconf/automake setup) ensuring its always set before any header is included anywhere.

Explorer09:
> There's a third simpler option which I see commonly used in OSS projects - pass -D_GNU_SOURCE=1 to the compiler (via the autoconf/automake setup) ensuring its always set before any header is included anywhere.

I would not count that as an option as that defeats the purpose of autoheader/"config.h".

natoscott:
> @fasterit The PCP failure comes from the brain-dead design of PCP that requires you to undefine `PACKAGE_NAME` because they are re-using that identifier in their own headers. I'll have a look to update the check to use an `HTOP_` specific define instead.

Hmm - PCP uses these macros in it's own generated config.h in the same way htop does (both are autoconf projects).  So this #undef is just dealing with the fact that we have overlapping macros as a result of having two independent config.h files (the PCP build of pcp-htop uses the PCP config.h, which is a superset).

If this aspect is complicating things just bounce it over to me & I'll see if I can come up with something neater than we have to date @BenBE

natoscott:
> I would not count that as an option as that defeats the purpose of autoheader/"config.h".

"Perfect is the enemy of good".

Explorer09:
> > @fasterit The PCP failure comes from the brain-dead design of PCP that requires you to undefine `PACKAGE_NAME` because they are re-using that identifier in their own headers. I'll have a look to update the check to use an `HTOP_` specific define instead.
> 
> Hmm - PCP uses these macros in it's own generated config.h in the same way htop does (both are autoconf projects). So this #undef is just dealing with the fact that we have overlapping macros as a result of having two independent config.h files (the PCP build of pcp-htop uses the PCP config.h, which is a superset).
> 

Would it work by putting these lines in the PCP's `configure.ac` script:

```m4
AC_DEFINE([[PCP_PACKAGE_NAME]], [AC_PACKAGE_NAME])
AC_DEFINE([[PCP_PACKAGE_TARNAME]], [AC_PACKAGE_TARNAME])
AC_DEFINE([[PCP_PACKAGE_VERSION]], [AC_PACKAGE_VERSION])
AC_DEFINE([[PCP_PACKAGE_STRING]], [AC_PACKAGE_STRING])
AC_DEFINE([[PCP_PACKAGE_BUGREPORT]], [AC_PACKAGE_BUGREPORT])
AC_DEFINE([[PCP_PACKAGE_URL]], [AC_PACKAGE_URL])
```

And then these in PCP's internal header (e.g. [pcp/platform_defs.h](https://github.com/performancecopilot/pcp/blob/main/src/include/pcp/platform_defs.h.in)) right after the `#include "config.h"` line:
```c
#undef PACKAGE_NAME
#undef PACKAGE_TARNAME
#undef PACKAGE_VERSION
#undef PACKAGE_STRING
#undef PACKAGE_BUGREPORT
#undef PACKAGE_URL
```

Since macro renaming cannot be done in the C preprocessor, I got an idea of using `configure` to produce alternative macro tokens that would avoid name conflict.

natoscott:
> > > @fasterit The PCP failure comes from the brain-dead design of PCP that requires you to undefine `PACKAGE_NAME` because they are re-using that identifier in their own headers. I'll have a look to update the check to use an `HTOP_` specific define instead.
> > 
> > 
> > Hmm - PCP uses these macros in it's own generated config.h in the same way htop does (both are autoconf projects). So this #undef is just dealing with the fact that we have overlapping macros as a result of having two independent config.h files (the PCP build of pcp-htop uses the PCP config.h, which is a superset).
> 
> Would it work by putting these lines in the PCP's `configure.ac` script:
> 

Yep, possibly.  One problem though is that would only work for new versions of the PCP headers after that change, and all current distros wont have that for some time.  But, if this is the cleanest way forward and we accept that limitation (htop CI will be an issue requiring solving too for pcp-htop builds), that's OK.

Explorer09:
Just to drop a note that I edited my last comment to correct the code.

While htop can do similarly to avoid future name conflicts, I wish this should be done in the library's side first.

```m4
AC_DEFINE([[HTOP_PACKAGE_NAME]], ["AC_PACKAGE_NAME"], [Name of htop package])
```

BenBE:
For the meantime I moved the internal check over to a different define, thus we should be good on that front.

But we should fix this upstream with PCP such that PCP properly prefixes its config defines. Maybe we should recheck the same for htop too, but less preasure there due to lack of re-use as a library.

BenBE:
> One more thing. Because of the feature test macro mess. I'm now thinking that the `#include "config.h"` lines might be removed from all "module" headers of htop. Interestingly, not all htop header files have this include line. I wonder how and why IWYU suggested including `config.h` for these headers in the first place (specifically, what tokens or symbols made IWYU suggest the `config.h`)

I moved the includes of config.h from all header files (except `Provide*.h`) to the corresponding C module source. Fallout was about 6 more files that needed `config.h` as their first include. Hope I caught all by now …

Explorer09:
@fasterit Just curious. Would the current `main` branch (after the merge of this PR) cause any build error with your PCP configuration? While the CI server does build it fine, I think you may test building it with your environment just to be sure.

And there are two .c files in htop that have the `#undef PACKAGE_*` lines as workarounds. Are the workaround lines (like the following) still needed?

```c
#undef PACKAGE_NAME
#undef PACKAGE_TARNAME
#undef PACKAGE_VERSION
#undef PACKAGE_STRING
#undef PACKAGE_BUGREPORT
#undef PACKAGE_URL
#include <pcp/pmapi.h>
#undef PACKAGE_NAME
#undef PACKAGE_TARNAME
#undef PACKAGE_VERSION
#undef PACKAGE_STRING
#undef PACKAGE_BUGREPORT
#undef PACKAGE_URL
```

fasterit:
I think you want @natoscott for this @Explorer09.

BenBE:
Technically they are no longer needed once the variables for the corresponding defines in `Makefile.am`/`configure.ac` are renamed. But that's not an issue of this PR, but a separate issue for @natoscott to take care of.

Explorer09:
@BenBE The default C macro names like `PACKAGE_NAME` might cause conflicts if htop's "config.h" and PCP's internal "config.h" are included together. The computer might produce a "macro redefinition" warning. That is what I was talking about.

AFAIK, there is no way for autoconf to _not_ predefine these default `PACKAGE_*` C macros in "config.h".

(And just for the info, the C macros `PACKAGE` and `VERSION`, are predefined by automake. They are not to be confused with autoconf-defined `PACKAGE_NAME` and `PACKAGE_VERSION` macros. The two automake macros can be suppressed by the `no-define` automake option.)

AryanGitHub:
I wanted to know, what is the workaround for building PCP build?
I got the error and warning at the time of building, 
waring 
`configure.ac:385: warning: pkg.m4 is absent or older than version 0.16; this 'configure' would have incomplete pkg-config support
`
and error as
`checking for pcp/pmapi.h... no
configure: error: can not find PCP header file
`

natoscott:
@AryanGitHub if you want to build pcp-htop you just need to install the devel package for your distribution which includes the header file mentioned there.

PR #1339, opened 2023-11-26
PCP platform compressed cgroup column implementation

natoscott:


BenBE:
Anything to take care of for the implementation of #1334? AFAICS the container name stuff would need to be added for PCP too?

natoscott:
> Anything to take care of for the implementation of #1334? AFAICS the container name stuff would need to be added for PCP too?

Nothing to add - it looks straightforward to extend this for use in PCP.

PR #1349, opened 2023-12-22
HeaderLayout: add "3 columns - 40/30/30"

eworm-de:
Existing layouts with three columns do not match my needs... There is always a column too narrow. I think this would be a reasonable choice.

### DIFF
--- a/HeaderLayout.h
+++ b/HeaderLayout.h
@@ -26,6 +26,9 @@ typedef enum HeaderLayout_ {
    HF_THREE_25_25_50,
    HF_THREE_25_50_25,
    HF_THREE_50_25_25,
+   HF_THREE_40_30_30,
+   HF_THREE_30_40_30,
+   HF_THREE_30_30_40,
    HF_THREE_40_20_40,
    HF_FOUR_25_25_25_25,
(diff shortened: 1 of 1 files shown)


BenBE:
I think that one was left out for symmetry reasons in the first version of the patch. If you add this combination, you should also probably add 30/40/30 and 30/30/40.

eworm-de:
Hmm, not sure we want or need all these versions. I tend to put text meters to the left, and graphical meters to the right. The former take a little more space to be useful, as truncation just cuts information.
I think most people have it that way, no?

But anyway, if you prefer I can add all the others too. I just wanted to avoid cluttering this too much in first place.

BenBE:
I have text meters on the right usually … So it makes some sense to have at least the 50/30/30 and the 30/30/40 variants. And it doesn't hurt much to add the 30/40/30 variant too for people that want a symmetric largest column in the middle for some reason.

eworm-de:
Looks like you are the perfect proof I am wrong. 😜
Updated to include all three variants.

eworm-de:
Hmm, is the order ok? 🤔

BenBE (approved):

PR #1351, opened 2023-12-23
OpenFilesScreen: allocate 16 columns to the size

nabijaczleweli:
Previously this limited the size to single-digit-gigabytes(!). Now it can be single-digit-petabytes.

Fixes https://bugs.debian.org/1059369, see there for before/after snippets

### DIFF
--- a/OpenFilesScreen.c
+++ b/OpenFilesScreen.c
@@ -253,7 +253,7 @@ static void OpenFilesScreen_scan(InfoScreen* this) {
       while (fdata) {
          OpenFiles_Data* data = &fdata->data;
          char* entry = NULL;
-         xAsprintf(&entry, "%5.5s %-7.7s %-4.4s %-10.10s %10.10s %10.10s %10.10s  %s",
+         xAsprintf(&entry, "%5.5s %-7.7s %-4.4s %-4.4s %16s %10.10s %10.10s  %s",
                    getDataForType(data, 'f'),
                    getDataForType(data, 't'),
                    getDataForType(data, 'a'),


Explorer09:
Increasing the width to 16 terminal columns (characters) doesn't solve the problem. It's still possible to overflow.
Is it possible to adopt a formatting function like `Row_printBytes` to display a large number with unit suffixes?

BenBE:
I think with the new Table interface from @natoscott this should be easy. Doing so also allows for that column to be sized dynamically (in particular with the `inode` column which can be WIDE™).

OpenFilesScreen.c
```diff
@@ -253,7 +253,7 @@ static void OpenFilesScreen_scan(InfoScreen* this) {
       while (fdata) {
          OpenFiles_Data* data = &fdata->data;
          char* entry = NULL;
-         xAsprintf(&entry, "%5.5s %-7.7s %-4.4s %-10.10s %10.10s %10.10s %10.10s  %s",
+         xAsprintf(&entry, "%5.5s %-7.7s %-4.4s %-4.4s %16.16s %10.10s %10.10s  %s",
```
Explorer09, line 256: To prevent further truncation of the "SIZE" field, the precision specifier should be removed from the format. Like this:
```suggestion
         xAsprintf(&entry, "%5.5s %-7.7s %-4.4s %-4.4s %16s %10.10s %10.10s  %s",
```

nabijaczleweli: This runs it out of alignment for >10PB but that's fine I think. Applied.

BenBE:
Superseeded by #1356 …

Issue #1335, opened 2023-11-24
Temperature mapping is wrong on i7-1355U


leahneukirchen:
This bug is related to #879.

I have a 13th Gen Intel Raptor Lake i7-1355U CPU, which has 2 performance cores (with Hyperthreading), and 8 efficiency cores.

lscpu gives the respective frequencies:

```
% lscpu -e
CPU NODE SOCKET CORE L1d:L1i:L2:L3 ONLINE    MAXMHZ   MINMHZ       MHZ
  0    0      0    0 0:0:0:0          yes 5000.0000 400.0000 2198.6541
  1    0      0    0 0:0:0:0          yes 5000.0000 400.0000 2199.9771
  2    0      0    1 4:4:1:0          yes 5000.0000 400.0000  559.3190
  3    0      0    1 4:4:1:0          yes 5000.0000 400.0000 1644.4080
  4    0      0    2 8:8:2:0          yes 3700.0000 400.0000 1799.9900
  5    0      0    3 9:9:2:0          yes 3700.0000 400.0000  400.0000
  6    0      0    4 10:10:2:0        yes 3700.0000 400.0000 1658.6350
  7    0      0    5 11:11:2:0        yes 3700.0000 400.0000 1789.9850
  8    0      0    6 12:12:3:0        yes 3700.0000 400.0000 1728.6110
  9    0      0    7 13:13:3:0        yes 3700.0000 400.0000  743.2380
 10    0      0    8 14:14:3:0        yes 3700.0000 400.0000 1759.9810
 11    0      0    9 15:15:3:0        yes 3700.0000 400.0000 1405.3361
```

The `coretemp` sensor provides these temperatures:

```
coretemp-isa-0000
Adapter: ISA adapter
Package id 0:  +40.0°C  (high = +100.0°C, crit = +100.0°C)
Core 0:        +38.0°C  (high = +100.0°C, crit = +100.0°C)
Core 4:        +33.0°C  (high = +100.0°C, crit = +100.0°C)
Core 8:        +39.0°C  (high = +100.0°C, crit = +100.0°C)
Core 9:        +39.0°C  (high = +100.0°C, crit = +100.0°C)
Core 10:       +39.0°C  (high = +100.0°C, crit = +100.0°C)
Core 11:       +39.0°C  (high = +100.0°C, crit = +100.0°C)
Core 12:       +38.0°C  (high = +100.0°C, crit = +100.0°C)
[...]

Issue #1048, opened 2022-06-29
Missing temperature reports on Intel Alder Lake Processor cores


pjungkamp:
I have a `Lenovo Yoga 9 14IAP7` with an `Intel Core i5-1240P`. 

`htop` only shows temperatures for the 0, 4, 8 & 12 cores.
![image](https://user-images.githubusercontent.com/56401138/176481393-b7a99cb9-73a3-4d07-a249-546381f9708c.png)

The temperatures `tempX` from [`sensors -u coretemp-isa-0000`](https://github.com/htop-dev/htop/files/9012170/sensors.txt) seem to correspond to the `cpu id` rather than the `processor` of a core in [`/proc/cpuinfo`](https://github.com/htop-dev/htop/files/9012156/cpuinfo.txt).

Is this a new quirk of Alder Lake chips (e.g. due tho the big-little design)?

BenBE:
See #806 and #856 …

`htop` relies on `libsensors` to provide reasonable values.
Unfortunately `libsensors` is not reasonable enough …

dstahlke:
The issue might be that sensors indexes by core id, and these have a strange mapping to processors.  For some reason core id increments by 4, with two processors per core (hyperthreading).  That's for P-cores.  For E-cores there is one core id per processor.  These match between sensors and cpuinfo.  Should htop be tying sensor values to cpuinfo core id?

Here is an abbreviated demonstration:

$ sensors
Core 0:        +39.0°C  (high = +100.0°C, crit = +100.0°C)
Core 4:        +41.0°C  (high = +100.0°C, crit = +100.0°C)
Core 8:        +42.0°C  (high = +100.0°C, crit = +100.0°C)
... etc ...

$ grep 'core id' /proc/cpuinfo
core id         : 0
core id         : 0
core id         : 4
core id         : 4
core id         : 8
core id         : 8

PR #1352, opened 2023-12-25
linux: assign CPU temperatures by package/core or CCD

leahneukirchen:
Closes https://github.com/htop-dev/htop/issues/806.
Closes https://github.com/htop-dev/htop/issues/1048.
Closes https://github.com/htop-dev/htop/pull/1176.
Closes https://github.com/htop-dev/htop/issues/1335.
Addresses https://github.com/htop-dev/htop/issues/879 (on Linux).

### DIFF
--- a/linux/LibSensors.c
+++ b/linux/LibSensors.c
@@ -35,6 +35,7 @@ in the source distribution for its full text.
 #define sym_sensors_get_features sensors_get_features
 #define sym_sensors_get_subfeature sensors_get_subfeature
 #define sym_sensors_get_value sensors_get_value
+#define sym_sensors_get_label sensors_get_label
 
 #else
 
@@ -44,6 +45,7 @@ static const sensors_chip_name* (*sym_sensors_get_detected_chips)(const sensors_
 static const sensors_feature* (*sym_sensors_get_features)(const sensors_chip_name*, int*);
 static const sensors_subfeature* (*sym_sensors_get_subfeature)(const sensors_chip_name*, const sensors_feature*, sensors_subfeature_type);
 static int (*sym_sensors_get_value)(const sensors_chip_name*, int, double*);
+static char* (*sym_sensors_get_label)(const sensors_chip_name*, const sensors_feature *feature);
 
 static void* dlopenHandle = NULL;
 
@@ -82,6 +84,7 @@ int LibSensors_init(void) {
       resolve(sensors_get_features);
       resolve(sensors_get_subfeature);
       resolve(sensors_get_value);
+      resolve(sensors_get_label);
 
       #undef resolve
    }
--- a/linux/LibSensors.c
+++ b/linux/LibSensors.c
@@ -159,6 +159,38 @@ static int tempDriverPriority(const sensors_chip_name* chip) {
    return -1;
 }
 
+int LibSensors_countCCDs(void) {
+
+#ifndef BUILD_STATIC
+   if (!dlopenHandle)
+      return 0;
+#endif /* !BUILD_STATIC */
+
+   int ccds = 0;
+
+   int n = 0;
+   for (const sensors_chip_name* chip = sym_sensors_get_detected_chips(NULL, &n); chip; chip = sym_sensors_get_detected_chips(NULL, &n)) {
+      int m = 0;
+      for (const sensors_feature* feature = sym_sensors_get_features(chip, &m); feature; feature = sym_sensors_get_features(chip, &m)) {
+         if (feature->type != SENSORS_FEATURE_TEMP)
+            continue;
+
+         if (!feature->name || !String_startsWith(feature->name, "temp"))
+            continue;
+
+         char *label = sym_sensors_get_label(chip, feature);
+         if (label) {
+            if (String_startsWith(label, "Tccd")) {
+               ccds++;
+            }
+            free(label);
+         }
+      }
+   }
+
+   return ccds;
+}
+
 void LibSensors_getCPUTemperatures(CPUData* cpus, unsigned int existingCPUs, unsigned int activeCPUs) {
    assert(existingCPUs > 0 && existingCPUs < 16384);
 
--- a/linux/LibSensors.c
+++ b/linux/LibSensors.c
@@ -16,6 +16,7 @@ in the source distribution for its full text.
 #include <errno.h>
 #include <limits.h>
 #include <math.h>
+#include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -206,6 +207,8 @@ void LibSensors_getCPUTemperatures(CPUData* cpus, unsigned int existingCPUs, uns
    unsigned int coreTempCount = 0;
    int topPriority = 99;
 
+   int ccdID = 0;
+
    int n = 0;
    for (const sensors_chip_name* chip = sym_sensors_get_detected_chips(NULL, &n); chip; chip = sym_sensors_get_detected_chips(NULL, &n)) {
       const int priority = tempDriverPriority(chip);
@@ -223,6 +226,8 @@ void LibSensors_getCPUTemperatures(CPUData* cpus, unsigned int existingCPUs, uns
 
       topPriority = priority;
 
+      int physicalID = -1;
+
       int m = 0;
       for (const sensors_feature* feature = sym_sensors_get_features(chip, &m); feature; feature = sym_sensors_get_features(chip, &m)) {
          if (feature->type != SENSORS_FEATURE_TEMP)
@@ -237,9 +242,6 @@ void LibSensors_getCPUTemperatures(CPUData* cpus, unsigned int existingCPUs, uns
 
          /* Feature name IDs start at 1, adjust to start at 0 to match data indices */
          tempID--;
-
-         if (tempID > existingCPUs)
-            continue;
 
          const sensors_subfeature* subFeature = sym_sensors_get_subfeature(chip, feature, SENSORS_SUBFEATURE_TEMP_INPUT);
          if (!subFeature)
@@ -290,6 +292,46 @@ void LibSensors_getCPUTemperatures(CPUData* cpus, unsigned int existingCPUs, uns
                continue;
             }
          }
+
+         char *label = sym_sensors_get_label(chip, feature);
+         if (label) {
+            bool skip = true;
+            /* Intel coretemp names, labels mention package and phyiscal id */
+            if (String_startsWith(label, "Package id ")) {
+               physicalID = strtoul(label + strlen("Package id "), NULL, 10);
+            } else if (String_startsWith(label, "Physical id ")) {
+               physicalID = strtoul(label + strlen("Physical id "), NULL, 10);
+            } else if (String_startsWith(label, "Core ")) {
+               int coreID = strtoul(label + strlen("Core "), NULL, 10);
+               for (size_t i = 1; i < existingCPUs + 1; i++) {
+                  if (cpus[i].physicalID == physicalID && cpus[i].coreID == coreID) {
+                     data[i] = temp;
+                     coreTempCount++;
+                  }
+               }
+            }
+
+            /* AMD k10temp/zenpower names, only CCD is known */
+            else if (String_startsWith(label, "Tccd")) {
+               for (size_t i = 1; i <= existingCPUs; i++) {
+                  if (cpus[i].ccdID == ccdID) {
+                     data[i] = temp;
+                     coreTempCount++;
+                  }
+               }
+               ccdID++;
+            } else {
+               skip = false;
+            }
+
+            free(label);
+
+            if (skip)
+               continue;
+         }
+
+         if (tempID > existingCPUs)
+            continue;
 
          /* If already set, e.g. Ryzen reporting platform temperature for each die, use the bigger one */
          if (isNaN(data[tempID])) {
--- a/linux/LinuxMachine.c
+++ b/linux/LinuxMachine.c
@@ -669,6 +669,47 @@ static void LinuxMachine_fetchCPUTopologyFromCPUinfo(LinuxMachine* this) {
 
    fclose(file);
 }
+
+static void LinuxMachine_assignCCDs(LinuxMachine* this, int ccds) {
+   /* For AMD k10temp/zenpower, temperatures are provided for CCDs only,
+      which is an aggregate of multiple cores.
+      There's no obvious mapping between hwmon sensors and sockets and CCDs.
+      Assume both are iterated in order.
+      Hypothesis: Each CCD has same size N = #Cores/#CCD
+      and is assigned N coreID in sequence.
+      Also assume all CPUs have same number of CCDs. */
+
+   const Machine* super = &this->super;
+   CPUData *cpus = this->cpuData;
+
+   if (ccds == 0) {
+      for (size_t i = 0; i < super->existingCPUs + 1; i++) {
+         cpus[i].ccdID = -1;
+      }
+      return;
+   }
+
+   int coresPerCCD = super->existingCPUs / ccds;
+
+   int ccd = 0;
+   int nc = coresPerCCD;
+   for (int p = 0; p <= (int)this->maxPhysicalID; p++) {
+      for (int c = 0; c <= (int)this->maxCoreID; c++) {
+         for (size_t i = 1; i <= super->existingCPUs; i++) {
+            if (cpus[i].physicalID != p || cpus[i].coreID != c)
+               continue;
+
+            cpus[i].ccdID = ccd;
+
+            if (--nc <= 0) {
+               nc = coresPerCCD;
+               ccd++;
+            }
+         }
+      }
+   }
+}
+
 #endif
 
 static void LinuxMachine_scanCPUFrequency(LinuxMachine* this) {
@@ -749,6 +790,8 @@ Machine* Machine_new(UsersTable* usersTable, uid_t userId) {
    #ifdef HAVE_SENSORS_SENSORS_H
    // Fetch CPU topology
    LinuxMachine_fetchCPUTopologyFromCPUinfo(this);
+   int ccds = LibSensors_countCCDs();
+   LinuxMachine_assignCCDs(this, ccds);
    #endif
 
    return super;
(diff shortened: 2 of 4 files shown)


leahneukirchen:
Fixed build without libsensors.

Fixed assignment on RPi.

linux/LibSensors.c
```diff
@@ -149,6 +152,62 @@ static int tempDriverPriority(const sensors_chip_name* chip) {
...
+                  cpus[i].ccdID = ccd;
+                  if (--nc == 0) {
+                     nc = ccdsPerCore;
+                     ccd++;
+                  }
+               }
+            }
+         }
+      }
+   }
```
BenBE, line 208: Try to avoid deep nesting by early returns or negating conditions to use `continue` in strategic places.

linux/LibSensors.c
```diff
@@ -208,6 +271,42 @@ void LibSensors_getCPUTemperatures(CPUData* cpus, unsigned int existingCPUs, uns
          if (r != 0)
             continue;
 
+         char *label = sym_sensors_get_label(chip, feature);
+         if (label) {
+            int skip = 1;
```
BenBE, line 276: Use `bool` instead if integers with just `0`/`1` as their value. Include `stdbool.h` as necessary.

linux/LinuxMachine.c
```diff
@@ -603,6 +603,48 @@ static void scanCPUFrequencyFromCPUinfo(LinuxMachine* this) {
...
+
+   int cpuid = -1;
+   int coreid = -1;
+   int physicalid = -1;
+
+   while (!feof(file)) {
+      char buffer[PROC_LINE_LENGTH];
+
+      if (fgets(buffer, PROC_LINE_LENGTH, file) == NULL)
+         break;
```
BenBE, line 622: Beware of excessively long lines; do proper EOL handling.

linux/LinuxMachine.c
```diff
@@ -603,6 +603,48 @@ static void scanCPUFrequencyFromCPUinfo(LinuxMachine* this) {
...
+
+      if (sscanf(buffer, "processor : %d", &cpuid) == 1) {
+         continue;
+      } else if (sscanf(buffer, "physical id : %d", &physicalid) == 1) {
+         continue;
+      } else if (sscanf(buffer, "core id : %d", &coreid) == 1) {
+         continue;
+      } else if (buffer[0] == '\n') {
```
BenBE, line 630: Should be checked first, as `sscanf` is quite slow.

linux/LinuxMachine.c
```diff
@@ -603,6 +603,48 @@ static void scanCPUFrequencyFromCPUinfo(LinuxMachine* this) {
...
+
+      if (fgets(buffer, PROC_LINE_LENGTH, file) == NULL)
+         break;
+
+      if (sscanf(buffer, "processor : %d", &cpuid) == 1) {
+         continue;
+      } else if (sscanf(buffer, "physical id : %d", &physicalid) == 1) {
+         continue;
+      } else if (sscanf(buffer, "core id : %d", &coreid) == 1) {
+         continue;
```
BenBE, line 629: Consider using prefix check (`String_startsWith` from `XUtils.h`) with `fast_strtoull`.

BenBE (changes requested):

linux/LinuxMachine.c
leahneukirchen: All code in that file does it that way, 4096 should be plenty?  Also we only care about short lines.

linux/LinuxMachine.c
leahneukirchen: I want to preserve the whitespace skipping of sscanf for robustness, gating it with a prefix check now.

leahneukirchen:
Note that the cpu are only parsed once, this takes <0.1ms on my machine.

linux/LinuxMachine.c
BenBE: There'S `String_readline` in `XUtils.h` that's fully capable of very long lines and does proper re-assembly.

linux/LibSensors.c
BenBE:
```suggestion
   if (ccds < 1)
      return;

   int ccdsPerCore = existingCPUs / ccds;

   int ccd = 0;
   int nc = ccdsPerCore;

   for (int p = 0; p < (int)existingCPUs; p++) {
      for (int c = 0; c < (int)existingCPUs; c++) {
         for (size_t i = 1; i <= existingCPUs; i++) {
            if (cpus[i].physicalID != p || cpus[i].coreID != c)
               continue;

            cpus[i].ccdID = ccd;
            if (--nc == 0) {
               nc = ccdsPerCore;
               ccd++;
            }
         }
      }
   }
```

Any reason for this last loop to be O(n³)?

linux/LinuxMachine.c
```diff
@@ -603,6 +603,48 @@ static void scanCPUFrequencyFromCPUinfo(LinuxMachine* this) {
...
+         cpuData->physicalID = physicalid;
+
+         cpuid = -1;
+         coreid = -1;
+      } else if (buffer[0] == 'p' && buffer[1] == 'r' && sscanf(buffer, "processor : %d", &cpuid) == 1) {
+         continue;
+      } else if (buffer[0] == 'p' && buffer[1] == 'h' && sscanf(buffer, "physical id : %d", &physicalid) == 1) {
+         continue;
+      } else if (buffer[0] == 'c' && buffer[1] == 'o' && sscanf(buffer, "core id : %d", &coreid) == 1) {
+         continue;
```
BenBE, line 640:
```suggestion
      } else if (String_startsWith(buffer, "processor") && sscanf(buffer, "processor : %d", &cpuid) == 1) {
         continue;
      } else if (String_startsWith(buffer, "physical id") && sscanf(buffer, "physical id : %d", &physicalid) == 1) {
         continue;
      } else if (String_startsWith(buffer, "core id") && sscanf(buffer, "core id : %d", &coreid) == 1) {
         continue;
```

linux/LibSensors.c
leahneukirchen: This was the most straightforward way to write this, but perhaps I can find a better solution...

BenBE: I think a O(n) option could be:

1. Alloc memory for each CCD to store plus a 2D array of pointers to map the CCD ID. Initialize each entry to -1 (unassigned); index by physical id/core id.
2. Iterate over all CPU cores (loop `i`) and check if `ccd[pid][cid]` has been set: If so, assign the current CPU to the given CCD, otherwise allocate a new one and use that one instead.
3. Free the temporary memory.

leahneukirchen: One could also do one iteration over all CPU to find max core_id and phyiscal_id and then keep the O(n^3) loop with vastly smaller n.

BenBE: Technically that's O(n*m), but yes. Also this check could be included with the reading of the physical and core id values in the first place, making the cost effectively zero. Having these limits also reduces the memory cost of the suggested algorithm from my previous comment.

linux/LinuxMachine.c
```diff
@@ -615,34 +615,84 @@ static void LinuxMachine_fetchCPUTopologyFromCPUinfo(LinuxMachine* this) {
...
+            cpuData->coreID = coreid;
+            cpuData->physicalID = physicalid;
 
-         CPUData* cpuData = &(this->cpuData[cpuid + 1]);
-         cpuData->coreID = coreid;
-         cpuData->physicalID = physicalid;
+	    if (coreid > max_coreid)
+               max_coreid = coreid;
+	    if (physicalid > max_physicalid)
+               max_physicalid = physicalid;
```
BenBE, line 635: Indentation.

linux/LinuxMachine.c
```diff
@@ -615,34 +615,84 @@ static void LinuxMachine_fetchCPUTopologyFromCPUinfo(LinuxMachine* this) {
...
+   for (int p = 0; p <= (int)this->maxPhysicalID; p++) {
+      for (int c = 0; c <= (int)this->maxCoreID; c++) {
+         for (size_t i = 1; i <= super->existingCPUs; i++) {
+            if (cpus[i].physicalID == p && cpus[i].coreID == c) {
+               cpus[i].ccdID = ccd;
+               if (--nc == 0) {
+                  nc = ccdsPerCore;
+                  ccd++;
+               }
+            }
```
BenBE, line 690:
```suggestion
            if (cpus[i].physicalID != p || cpus[i].coreID != c)
               continue;

            cpus[i].ccdID = ccd;

            if (--nc == 0) {
               nc = ccdsPerCore;
               ccd++;
            }
```

linux/LibSensors.c
```diff
@@ -149,6 +153,32 @@ static int tempDriverPriority(const sensors_chip_name* chip) {
    return -1;
 }
 
+int LibSensors_countCCDs() {
```
BenBE, line 156:
```
linux/LibSensors.c:156:25: error: a function declaration without a prototype is deprecated in all versions of C [-Werror,-Wstrict-prototypes]
int LibSensors_countCCDs() {
                        ^
                         void
```

leahneukirchen:
I think this is ready to go, should I squash it together sensibly?

BenBE:
> I think this is ready to go, should I squash it together sensibly?

As you like. Just took a quick peek at the list of commits and some suggest that they mostly rectify things changed in previous commits of the PR: Those are the prime subjects to squash in the PR right now. So there's not really much to squash together right now AFAICS.

BenBE:
@leahneukirchen Anything left to do from your side? If not please mark the PR ready for review. TIA.

BenBE (approved):

linux/LinuxMachine.c
```diff
@@ -603,6 +603,100 @@ static void scanCPUFrequencyFromCPUinfo(LinuxMachine* this) {
...
+   CPUData *cpus = this->cpuData;
+
+   if (ccds == 0) {
+      for (size_t i = 0; i < super->existingCPUs + 1; i++) {
+         cpus[i].ccdID = -1;
+      }
+      return;
+   }
+
+   int coresPerCCD = super->existingCPUs / ccds;
```
cgzones, line 677: maybe add `assert(super->existingCPUs % ccds == 0)` to detect remnant cpus

linux/LinuxMachine.c
```diff
@@ -674,6 +772,13 @@ Machine* Machine_new(UsersTable* usersTable, uid_t userId) {
    // Initialize CPU count
    LinuxMachine_updateCPUcount(this);
 
+   #ifdef HAVE_SENSORS_SENSORS_H
+   // Fetch CPU topology
+   LinuxMachine_fetchCPUTopologyFromCPUinfo(this);
```
cgzones, line 777: maybe `LinuxMachine_fetchCPUTopologyFromCPUinfo()` can return whether the topology changed and then `LibSensors_countCCDs()` and `LinuxMachine_assignCCDs()` get only called on change?

fasterit:
@leahneukirchen Do you want to implement the three review comments from @cgzones ?

linux/LinuxMachine.c
```diff
@@ -603,6 +603,100 @@ static void scanCPUFrequencyFromCPUinfo(LinuxMachine* this) {
    }
 }
 
+#ifdef HAVE_SENSORS_SENSORS_H
+static void LinuxMachine_fetchCPUTopologyFromCPUinfo(LinuxMachine* this) {
```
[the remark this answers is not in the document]
leahneukirchen: Then we need to move the `ifdef HAVE_SENSORS_SENSORS_H` into that function, which makes it harder to read I think.

linux/LinuxMachine.c
leahneukirchen: I'm not sure failing hard is reasonable in this case, but there seems to be no warning mechanism in htop.

linux/LinuxMachine.c
leahneukirchen: I don't know what this means, this code runs only once already?

linux/LinuxMachine.c
cgzones: True, an assert would be wrong.

cgzones:
I somehow dislike that the sensors are iterated twice: once in `LibSensors_countCCDs()` and once in `LibSensors_getCPUTemperatures()`.
Do we need to compute the number of ccds beforehand, what about adjusting the CPU data after iterating all sensors  (and counting all CCD ones) in `LibSensors_getCPUTemperatures()` ?

fasterit:
> I somehow dislike that the sensors are iterated twice: once in `LibSensors_countCCDs()` and once in `LibSensors_getCPUTemperatures()`. Do we need to compute the number of ccds beforehand, what about adjusting the CPU data after iterating all sensors (and counting all CCD ones) in `LibSensors_getCPUTemperatures()` ?

Seconded, libsensors is slow

eworm-de:
Tested on my notebook with Alder Lake CPU... It shows temperatures for all cores at least.

ulope:
Tested on an AMD EPYC 7F52 (16 core, 16x1 CCX), seems to work. At least plausible looking temperatures are now shown for all cores.

BenBE:
AFAICS this is waiting for some refactoring work, but that work is yet not fully specified/explained. @cgzones can you please elaborate the open discussion threads?.

leahneukirchen:
Note that sensors are iterated twice on first start only, later only to read the temperatures.  (Patches welcome, I find it hard to rewrite this to use one pass.)

eworm-de:
If this is a one-time event it sounds like a reasonable tradeoff, compared to (probably) more complex code. I have not looked at it myself, though.

eworm-de:
Oh, this crashes with a segmentation fault if `libsensors.so` is not available...

leahneukirchen:
Sigh. With Kernel 6.9 there seems to be a bug in the logic now, my i7-1355U isn't mapped correctly anymore...

leahneukirchen:
With kernel 6.9.5 I get:

```
% pcat /sys/class/hwmon/hwmon2/*label
/sys/class/hwmon/hwmon2/temp1_label	Package id 0
/sys/class/hwmon/hwmon2/temp2_label	Core 0
/sys/class/hwmon/hwmon2/temp6_label	Core 4
/sys/class/hwmon/hwmon2/temp10_label	Core 8
/sys/class/hwmon/hwmon2/temp11_label	Core 9
/sys/class/hwmon/hwmon2/temp12_label	Core 10
/sys/class/hwmon/hwmon2/temp13_label	Core 11
/sys/class/hwmon/hwmon2/temp14_label	Core 12
/sys/class/hwmon/hwmon2/temp15_label	Core 13
/sys/class/hwmon/hwmon2/temp16_label	Core 14
/sys/class/hwmon/hwmon2/temp17_label	Core 15
```

Note that the tempID are not sequential.

eworm-de:
Thanks a lot for the quick fix!

linux/LibSensors.c
```diff
@@ -208,6 +245,45 @@ void LibSensors_getCPUTemperatures(CPUData* cpus, unsigned int existingCPUs, uns
...
+
+            /* AMD k10temp/zenpower names, only CCD is known */
+            else if (String_startsWith(label, "Tccd")) {
+               for (size_t i = 1; i <= existingCPUs; i++) {
+                  if (cpus[i].ccdID == ccdID) {
+                     data[i] = temp;
+                     coreTempCount++;
+                  }
+               }
+               ccdID++;
```
Vogtinator, line 274: Why not use the number from the label here, e.g. `Tccd5`? The features might not be sorted or even monotonically increasing by 1.

leahneukirchen: We need a total count of CCD however, too...

Vogtinator:
On my laptop with a Ryzen 5500U, there's only a single `Tctl` measurement for the CPU:

```
k10temp-pci-00c3
Adapter: PCI adapter
Tctl:         +47.4°C
```

According to the kernel doc's quote of the AMD manual:

> It does _not_ represent an actual physical temperature like die or case temperature.

But it appears somewhat representative, at least here, so it might make sense to include it.

Vogtinator:
I just checked on a server with AMD EPYC 7502, which supposedly (no idea whether it's correct!) has 2 sockets x 4 CCD x 2 CCX x 4 cores = 64 cores. The sockets are exposed as separate k10temp instances:

```
k10temp-pci-00c3
Adapter: PCI adapter
Tctl:         +38.4°C  
Tccd1:        +35.0°C  
Tccd3:        +36.8°C  
Tccd5:        +35.8°C  
Tccd7:        +35.8°C

k10temp-pci-00cb
Adapter: PCI adapter
Tctl:         +39.8°C  
Tccd1:        +38.0°C  
Tccd3:        +39.0°C  
Tccd5:        +37.5°C  
Tccd7:        +40.0°C
```

leahneukirchen:
Just having one sensor works already, no?

Vogtinator:
> Just having one sensor works already, no?

Depends on what you mean by "works". It shows a temperature value for each core, but I have no idea whether it matches. FWICT both sensors and its CCDs are read in whatever order sensors provides, which might or might not be related to the CPU core numbers...

leahneukirchen:
I tested on a `AMD EPYC 7443` that when i lock processes to some cores, only these heat up.

eworm-de:
This need a rebase on top of 6de810fa55936ad2ab8ae86db6b5f8b6d97a26c2...

BenBE (commented):
LGTM.

@leahneukirchen : What's still pending with the comment from @Vogtinator ?

rharish101:
Can confirm that this solves #806 for me on Arch Linux, with an AMD Ryzen 7 7800X3D.

linux/LinuxMachine.c
```diff
@@ -618,6 +618,100 @@ static void scanCPUFrequencyFromCPUinfo(LinuxMachine* this) {
...
+   int nc = coresPerCCD;
+   for (int p = 0; p <= (int)this->maxPhysicalID; p++) {
+      for (int c = 0; c <= (int)this->maxCoreID; c++) {
+         for (size_t i = 1; i <= super->existingCPUs; i++) {
+            if (cpus[i].physicalID != p || cpus[i].coreID != c)
+               continue;
+
+            cpus[i].ccdID = ccd;
+
+            if (--nc == 0) {
```
BenBE, line 704: This has `nc < 0` for `coresPerCCD==0`; which may happen with `super->existingCPUs < ccds`. This should not happen normally, but for the sake of robustness a check doesn't hurt. The easiest way I can think of to mitigate this would be changing this check like this:
```suggestion
            if (--nc <= 0) {
```

leahneukirchen: sgtm

PR #1353, opened 2023-12-25
Row_printKBytes() improvements

Explorer09:
* Allow the maximum value of "15.9Z" ((2^64 - 2) KiB) to be printed. (The previous version supports the maximum of "99.9P" (99.9 PiB) before the printing format goes weird.)
* Print up to 2 fraction digits for values between "9.76G" (10000 MiB) and "9.99G" (< 10 GiB), "9.76T" and "9.99T", and so on.
* `Row_printTime()` code shrink

### DIFF
--- a/Row.c
+++ b/Row.c
@@ -350,12 +350,12 @@ void Row_printTime(RichString* str, unsigned long long totalHundredths, bool col
 
    if (totalMinutes < 60) {
       unsigned int hundredths = totalHundredths % 100;
-      len = xSnprintf(buffer, sizeof(buffer), "%2llu:%02u.%02u ", totalMinutes, seconds, hundredths);
+      len = xSnprintf(buffer, sizeof(buffer), "%2u:%02u.%02u ", (unsigned int)totalMinutes, seconds, hundredths);
       RichString_appendnAscii(str, baseColor, buffer, len);
       return;
    }
    if (totalHours < 24) {
-      len = xSnprintf(buffer, sizeof(buffer), "%2lluh", totalHours);
+      len = xSnprintf(buffer, sizeof(buffer), "%2uh", (unsigned int)totalHours);
       RichString_appendnAscii(str, hourColor, buffer, len);
       len = xSnprintf(buffer, sizeof(buffer), "%02u:%02u ", minutes, seconds);
       RichString_appendnAscii(str, baseColor, buffer, len);
@@ -365,7 +365,7 @@ void Row_printTime(RichString* str, unsigned long long totalHundredths, bool col
    unsigned long long totalDays = totalHours / 24;
    unsigned int hours = totalHours % 24;
    if (totalDays < 10) {
-      len = xSnprintf(buffer, sizeof(buffer), "%1llud", totalDays);
+      len = xSnprintf(buffer, sizeof(buffer), "%1ud", (unsigned int)totalDays);
       RichString_appendnAscii(str, dayColor, buffer, len);
       len = xSnprintf(buffer, sizeof(buffer), "%02uh", hours);
       RichString_appendnAscii(str, hourColor, buffer, len);
@@ -374,7 +374,7 @@ void Row_printTime(RichString* str, unsigned long long totalHundredths, bool col
       return;
    }
    if (totalDays < /* Ignore leap years */365) {
-      len = xSnprintf(buffer, sizeof(buffer), "%4llud", totalDays);
+      len = xSnprintf(buffer, sizeof(buffer), "%4ud", (unsigned int)totalDays);
       RichString_appendnAscii(str, dayColor, buffer, len);
       len = xSnprintf(buffer, sizeof(buffer), "%02uh ", hours);
       RichString_appendnAscii(str, hourColor, buffer, len);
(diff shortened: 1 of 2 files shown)


BenBE (approved):
LGTM. What's missing in this draft in your PoV?

Explorer09:
The code is ready now. (I was performing minor optimizations to the code trying to make the binary smaller.)

Row.c
```diff
@@ -190,94 +190,112 @@ RowField RowField_keyAt(const Settings* settings, int at) {
...
 
-   number /= ONE_K;
+   // 100000 KiB (97.6 MiB) or greater. A unit prefix would be added.
+   const size_t maxUnitIndex = (sizeof(number) * CHAR_BIT - 1) / 10 + 1;
+   const bool canOverflow = maxUnitIndex >= ARRAYSIZE(unitPrefixes);
+
+   size_t i = 1;
+   int prevUnitColor;
+   // Convert KiB to (1/100) of MiB
+   unsigned long long hundredths = (number / 256) * 25 + (number % 256) * 25 / 256;
```
BenBE, line 237: I'd prefer `uint64_t` over those clunky `unsigned long long int`s (type-ref-wise) … ;-)

Row.c
```diff
@@ -190,94 +190,112 @@ RowField RowField_keyAt(const Settings* settings, int at) {
...
-   int gigabytesColor = coloring ? CRT_colors[PROCESS_GIGABYTES] : CRT_colors[PROCESS];
-   int shadowColor = coloring ? CRT_colors[PROCESS_SHADOW] : CRT_colors[PROCESS];
-   int baseColor = CRT_colors[PROCESS];
+   int color = CRT_colors[PROCESS];
+   int nextUnitColor = CRT_colors[PROCESS];
 
-   if (number == ULLONG_MAX) {
-      //Invalid number
-      RichString_appendAscii(str, shadowColor, "  N/A ");
+   int colors[4] = {
```
BenBE, line 200: Consider making this `const`.

Row.c
```diff
@@ -190,94 +190,112 @@ RowField RowField_keyAt(const Settings* settings, int at) {
...
+      nextUnitColor = colors[1];
+   }
+
+   if (number < 1000) {
+      // Plain number, no markings
+      len = xSnprintf(buffer, sizeof(buffer), "%5u ", (unsigned int)number);
+      RichString_appendnAscii(str, color, buffer, len);
+      return;
+   }
+   if (number < 100000) {
```
BenBE, line 221: Blank line before

Row.c
```diff
@@ -190,94 +190,112 @@ RowField RowField_keyAt(const Settings* settings, int at) {
...
+      prevUnitColor = color;
+      color = nextUnitColor;
+      if (coloring && i + 1 < ARRAYSIZE(colors))
+         nextUnitColor = colors[i + 1];
+
+      if (hundredths < 1000000)
+         break;
+      hundredths /= ONE_K;
+      i++;
+   }
```
BenBE, line 251: Add some blanks …
```suggestion
   while (true) {
      if (canOverflow && i >= ARRAYSIZE(unitPrefixes))
         goto invalidNumber;

      prevUnitColor = color;
      color = nextUnitColor;

      if (coloring && i + 1 < ARRAYSIZE(colors))
         nextUnitColor = colors[i + 1];

      if (hundredths < 1000000)
         break;

      hundredths /= ONE_K;
      i++;
   }
```

Row.c
```diff
@@ -190,94 +190,112 @@ RowField RowField_keyAt(const Settings* settings, int at) {
...
+      RichString_appendnAscii(str, nextUnitColor, buffer, len);
+      len = xSnprintf(buffer, sizeof(buffer), "%03u%c ", (unsigned int)number % 1000, unitPrefixes[i]);
    }
+   RichString_appendnAscii(str, color, buffer, len);
+   return;
+
+invalidNumber:
+   if (coloring)
+      color = CRT_colors[PROCESS_SHADOW];
+   RichString_appendAscii(str, color, "  N/A ");
```
BenBE, line 290: Blank line above …

Row.c
```diff
@@ -190,94 +190,112 @@ RowField RowField_keyAt(const Settings* settings, int at) {
...
+   return;
 }
 
-void Row_printKBytes(RichString* str, unsigned long long number, bool coloring) {
+void Row_printBytes(RichString* str, unsigned long long number, bool coloring) {
```
BenBE, line 294: Things below 1K don't really show as bytes with this despite the naming. We could extend the colors to use `PROCESS_SHADOW` for such small values though (only available if called through this wrapper) …

Row.h
```diff
@@ -142,12 +142,12 @@ void Row_setPidColumnWidth(pid_t maxPid);
...
 
+/* Takes number in kibibytes (base 1024). Prints 6 columns. */
+void Row_printKBytes(RichString* str, unsigned long long number, bool coloring);
+
 /* Takes number in bytes (base 1024). Prints 6 columns. */
 void Row_printBytes(RichString* str, unsigned long long number, bool coloring);
 
-/* Takes number in kilo bytes (base 1024). Prints 6 columns. */
-void Row_printKBytes(RichString* str, unsigned long long number, bool coloring);
-
```
BenBE, line 150: Why switch the order of these two declarations?

Row.c
Explorer09: I was keeping the original behavior of `Row_printBytes` while extending the function of `Row_printKBytes`. The original `Row_printBytes` prints 0 for everying below 1K (1024). My patch didn't change any of that behavior. It would be another PR for improving `Row_printBytes` after this...

Row.h
Explorer09: The .c file swaps the definitions of two functions as I now make `Row_printBytes` call `Row_printKBytes` rather than the other way around.

Row.c
Explorer09: That would change the API/prototype of `Row_printKBytes`.
(I know `uint64_t` is a synonym of `unsigned long long` in most 64-bit OSes, but in the standard's sense, `unsigned long long` is "64 bits or more" and a cast to `uint64_t` may serve as a type demotion.)

BenBE: As `uint64_t` is an alias for `unsigned long long int`, that's mostly cosmetic. I can live with that …

PR #1356, opened 2023-12-29
Dynamic column width for list of open files

BenBE:
Superseeds #1351

### DIFF
--- a/OpenFilesScreen.c
+++ b/OpenFilesScreen.c
@@ -251,15 +251,15 @@ static void OpenFiles_Data_clear(OpenFiles_Data* data) {
       free(data->data[i]);
 }
 
-static void OpenFilesScreen_scan(InfoScreen* this) {
-   Panel* panel = this->display;
+static void OpenFilesScreen_scan(InfoScreen* super) {
+   Panel* panel = super->display;
    int idx = Panel_getSelectedIndex(panel);
    Panel_prune(panel);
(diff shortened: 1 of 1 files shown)


OpenFilesScreen.c
```diff
@@ -164,6 +168,9 @@ static OpenFiles_ProcessData* OpenFilesScreen_getProcessData(pid_t pid) {
          {
             size_t index = getIndexForType(cmd);
             free_and_xStrdup(&item->data[index], line + 1);
+            if (strlen(item->data[index]) > (size_t)pdata->cols[index]) {
+               pdata->cols[index] = (int)CLAMP(strlen(item->data[index]), 0, INT16_MAX);
+            }
```
Explorer09, line 173:
```c
            int len = (int)CLAMP(strlen(item->data[index]), 0, INT16_MAX);
            assert(index < ARRAYSIZE(pdata->cols));
            if (len > pdata->cols[index]) {
               pdata->cols[index] = len;
            }
```

BenBE: Not necessary, as the `cols` array just mirrors the number of items from the `data` array, thus `index` always holds the correct value based on `size_t index = getIndexForType(cmd);`

Explorer09: @BenBE I just edited the post and corrected that. About the "optimization" I refer to using `strnlen` or `memchr` instead of `strlen` and then `CLAMP`.

BenBE: I applied the `strlen` part of that optimization; neither of the other two suggestions are fully standards compliant, because they imply the buffer is larger than it actually is.

PR #1360, opened 2024-01-01
Fix build errors and warnings in NetBSD.

fraggerfox:
- Fix build in NetBSD from the renaming done in b74673fe.
- Simplify the `ProcessTable_goThroughEntries()` and also apply the fixes for the compiler warning.
  ```
  netbsd/NetBSDProcessTable.c: In function 'NetBSDProcessTable_scanProcs':
  netbsd/NetBSDProcessTable.c:185:53: warning: comparison is always true due to limited range of data type [-Wtype-limits]
           const char* name = ((int64_t)kproc->p_tdev != (int64_t)KERN_PROC_TTY_NODEV) ? devname(kproc->p_tdev, S_IFCHR): NULL;
  ```
  The fix came in as part of the discussion [here](https://github.com/htop-dev/htop/pull/1360#discussion_r1439257379).

### DIFF
--- a/netbsd/NetBSDProcessTable.c
+++ b/netbsd/NetBSDProcessTable.c
@@ -150,8 +150,8 @@ static double getpcpu(const NetBSDMachine* nhost, const struct kinfo_proc2* kp) 
    return 100.0 * (double)kp->p_pctcpu / nhost->fscale;
 }
 
-static void NetBSDProcessTable_scanProcs(NetBSDProcessTable* this) {
-   const Machine* host = this->super.super.host;
+void ProcessTable_goThroughEntries(ProcessTable* super) {
+   const Machine* host = super->super.host;
    const NetBSDMachine* nhost = (const NetBSDMachine*) host;
    const Settings* settings = host->settings;
    bool hideKernelThreads = settings->hideKernelThreads;
@@ -164,7 +164,7 @@ static void NetBSDProcessTable_scanProcs(NetBSDProcessTable* this) {
       const struct kinfo_proc2* kproc = &kprocs[i];
 
       bool preExisting = false;
-      Process* proc = ProcessTable_getProcess(&this->super, kproc->p_pid, &preExisting, NetBSDProcess_new);
+      Process* proc = ProcessTable_getProcess(super, kproc->p_pid, &preExisting, NetBSDProcess_new);
 
       proc->super.show = ! ((hideKernelThreads && Process_isKernelThread(proc)) || (hideUserlandThreads && Process_isUserlandThread(proc)));
 
@@ -179,10 +179,12 @@ static void NetBSDProcessTable_scanProcs(NetBSDProcessTable* this) {
          proc->isUserlandThread = Process_getPid(proc) != Process_getThreadGroup(proc); // eh?
          proc->starttime_ctime = kproc->p_ustart_sec;
          Process_fillStarttimeBuffer(proc);
-         ProcessTable_add(&this->super, proc);
+         ProcessTable_add(super, proc);
 
          proc->tty_nr = kproc->p_tdev;
-         const char* name = ((dev_t)kproc->p_tdev != KERN_PROC_TTY_NODEV) ? devname(kproc->p_tdev, S_IFCHR) : NULL;
+         // KERN_PROC_TTY_NODEV is a negative constant but the type of
+         // kproc->p_tdev may be unsigned.
+         const char* name = ((dev_t)~kproc->p_tdev != (dev_t)~(KERN_PROC_TTY_NODEV)) ? devname(kproc->p_tdev, S_IFCHR) : NULL;
          if (!name) {
             free(proc->tty_name);
             proc->tty_name = NULL;
@@ -257,21 +259,15 @@ static void NetBSDProcessTable_scanProcs(NetBSDProcessTable* this) {
       }
 
       if (Process_isKernelThread(proc)) {
-         this->super.kernelThreads++;
+         super->kernelThreads++;
       } else if (Process_isUserlandThread(proc)) {
-         this->super.userlandThreads++;
+         super->userlandThreads++;
       }
 
-      this->super.totalTasks++;
+      super->totalTasks++;
       if (proc->state == RUNNING) {
-         this->super.runningTasks++;
+         super->runningTasks++;
       }
       proc->super.updated = true;
    }
-}
-
-void ProcessTable_goThroughEntries(ProcessTable* super) {
-   NetBSDProcessTable* npt = (NetBSDProcessTable*) super;
-
-   NetBSDProcessTable_scanProcs(npt);
 }
--- a/netbsd/NetBSDProcessTable.c
+++ b/netbsd/NetBSDProcessTable.c
@@ -151,7 +151,7 @@ static double getpcpu(const NetBSDMachine* nhost, const struct kinfo_proc2* kp) 
 }
 
 static void NetBSDProcessTable_scanProcs(NetBSDProcessTable* this) {
-   const Machine* host = this->super.host;
+   const Machine* host = this->super.super.host;
    const NetBSDMachine* nhost = (const NetBSDMachine*) host;
    const Settings* settings = host->settings;
    bool hideKernelThreads = settings->hideKernelThreads;
--- a/netbsd/NetBSDMachine.c
+++ b/netbsd/NetBSDMachine.c
@@ -81,7 +81,7 @@ static void NetBSDMachine_updateCPUcount(NetBSDMachine* this) {
    }
 
    if (value != super->existingCPUs) {
-      opl->cpuData = xReallocArray(this->cpuData, value + 1, sizeof(CPUData));
+      this->cpuData = xReallocArray(this->cpuData, value + 1, sizeof(CPUData));
       super->existingCPUs = value;
       change = true;
    }
@@ -127,7 +127,7 @@ Machine* Machine_new(UsersTable* usersTable, uid_t userId) {
       CRT_fatalError("kvm_openfiles() failed");
    }
 
-   return this;
+   return super;
 }
 
 void Machine_delete(Machine* super) {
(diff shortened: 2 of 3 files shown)


Explorer09:
Do you have a build log, i.e. a list of compiler warnings and errors of your build from the original code? That would save us time reviewing the code.

fraggerfox:
> Do you have a build log, i.e. a list of compiler warnings and errors of your build from the original code? That would save us time reviewing the code.

Yes, this makes sense, should have added it when I was making PR.

fraggerfox:
```
netbsd/NetBSDProcess.c: In function 'NetBSDProcess_rowWriteField':
netbsd/NetBSDProcess.c:240:26: error: incompatible type for argument 1 of 'Process_writeField'
       Process_writeField(np->super, str, field);
                          ^~
In file included from ./netbsd/NetBSDProcess.h:17:0,
                 from netbsd/NetBSDProcess.c:13:
./Process.h:219:6: note: expected 'const Process * {aka const struct Process_ *}' but argument is of type 'Process {aka const struct Process_}'
 void Process_writeField(const Process* row, RichString* str, ProcessField field);
      ^~~~~~~~~~~~~~~~~~
*** [netbsd/NetBSDProcess.o] Error code 1

make[1]: stopped in /usr/pkgsrc-current/wip/htop-git/work/htop-c3b4a770c02c28662980014b5a893e01451ec2b2
--- netbsd/NetBSDMachine.o ---
netbsd/NetBSDMachine.c: In function 'NetBSDMachine_updateCPUcount':
netbsd/NetBSDMachine.c:84:7: error: 'opl' undeclared (first use in this function); did you mean 'open'?
       opl->cpuData = xReallocArray(this->cpuData, value + 1, sizeof(CPUData));
       ^~~
       open
netbsd/NetBSDMachine.c:84:7: note: each undeclared identifier is reported only once for each function it appears in
netbsd/NetBSDMachine.c: In function 'Machine_new':
netbsd/NetBSDMachine.c:130:11: warning: return from incompatible pointer type [-Wincompatible-pointer-types]
    return this;
           ^~~~
netbsd/NetBSDMachine.c: In function 'Machine_scan':
netbsd/NetBSDMachine.c:271:4: warning: implicit declaration of function 'NetBSDProcessTable_scanMemoryInfo'; did you mean 'NetBSDMachine_scanMemoryInfo'? [-Wimplicit-function-declaration]
    NetBSDProcessTable_scanMemoryInfo(this);
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    NetBSDMachine_scanMemoryInfo
netbsd/NetBSDMachine.c:272:4: warning: implicit declaration of function 'NetBSDProcessTable_scanCPUTime'; did you mean 'NetBSDMachine_scanCPUTime'? [-Wimplicit-function-declaration]
    NetBSDProcessTable_scanCPUTime(this);
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    NetBSDMachine_scanCPUTime
netbsd/NetBSDMachine.c:275:7: warning: implicit declaration of function 'NetBSDProcessTable_scanCPUFrequency'; did you mean 'NetBSDMachine_scanCPUFrequency'? [-Wimplicit-function-declaration]
       NetBSDProcessTable_scanCPUFrequency(npl);
       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
       NetBSDMachine_scanCPUFrequency
netbsd/NetBSDMachine.c:275:43: error: 'npl' undeclared (first use in this function); did you mean 'nl'?
       NetBSDProcessTable_scanCPUFrequency(npl);
                                           ^~~
                                           nl
At top level:
netbsd/NetBSDMachine.c:222:13: warning: 'NetBSDMachine_scanCPUFrequency' defined but not used [-Wunused-function]
 static void NetBSDMachine_scanCPUFrequency(NetBSDMachine* this) {
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
netbsd/NetBSDMachine.c:197:13: warning: 'NetBSDMachine_scanCPUTime' defined but not used [-Wunused-function]
 static void NetBSDMachine_scanCPUTime(NetBSDMachine* this) {
             ^~~~~~~~~~~~~~~~~~~~~~~~~
netbsd/NetBSDMachine.c:145:13: warning: 'NetBSDMachine_scanMemoryInfo' defined but not used [-Wunused-function]
 static void NetBSDMachine_scanMemoryInfo(NetBSDMachine* this) {
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
*** [netbsd/NetBSDMachine.o] Error code 1
```
Relevant section of the build log showing the failure.

netbsd/NetBSDProcessTable.c
```diff
@@ -182,7 +182,7 @@ static void NetBSDProcessTable_scanProcs(NetBSDProcessTable* this) {
          ProcessTable_add(&this->super, proc);
 
          proc->tty_nr = kproc->p_tdev;
-         const char* name = ((dev_t)kproc->p_tdev != KERN_PROC_TTY_NODEV) ? devname(kproc->p_tdev, S_IFCHR) : NULL;
```
fraggerfox, line 185:
```c
(dev_t)kproc->p_tdev != KERN_PROC_TTY_NODEV)
```

There was a compiler warning stating this expression always evaluates to true. so removed it. The new build seems to be working fine after this removal.

Explorer09: I dug the NetBSD header files, and this looks like a bug in the conditional. It probably meant to work like this:

```c
         const char* name = ((int64_t)kproc->p_tdev != (int64_t)KERN_PROC_TTY_NODEV) ? devname(kproc->p_tdev, S_IFCHR) : NULL;
```

Better make this change a separate commit as this one is a bug fix rather than a build error fix.

Explorer09: All relevant definitions in NetBSD headers:

```c
// NetBSD <sys/types.h>
typedef uint64_t dev_t;
// NetBSD <sys/param.h>
#define NODEV (dev_t)(-1)
// NetBSD <sys/sysctl.h>
#define KERN_PROC_TTY_NODEV NODEV	
struct kinfo_proc2 {
...
uint32_t p_tdev;
...
};
```

fraggerfox: I had attempted this change when I was fixing the build and it still did not resolve the issue

```
netbsd/NetBSDProcessTable.c: In function 'NetBSDProcessTable_scanProcs':
netbsd/NetBSDProcessTable.c:185:53: warning: comparison is always true due to limited range of data type [-Wtype-limits]
          const char* name = ((int64_t)kproc->p_tdev != (int64_t)KERN_PROC_TTY_NODEV) ? devname(kproc->p_tdev, S_IFCHR): NULL;
```

I have reverted the change for now, if it causes trouble in the future will revisit it.

Explorer09: @fraggerfox I made a typo. `uint32_t` cast to `int64_t` always results in a positive value.

Now I'm troubled. While the downcast like the following code would work, I really don't want to depend on the type of 
`(struct kinfo_proc2).p_tdev` for portability reason.

```c
         const char* name = (kproc->p_tdev != (uint32_t)KERN_PROC_TTY_NODEV) ? devname(kproc->p_tdev, S_IFCHR) : NULL;
```

Since I am not the author or maintainer of htop NetBSD port. I would stop here and let someone who is more familiar with NetBSD code to look at this.

fraggerfox: @Explorer09 : I agree with you, I think I need to look a bit deeper into the meaning of this check and why I put there in the first place. For now I see no harm in leaving it as it is.

P.S: I authored the [NetBSD](https://github.com/htop-dev/htop/pull/603) implementation a while back and I currently maintain it.

Explorer09: @fraggerfox 
I finally figured out what to do. :)
The purpose of the code is to not allocate the name of the controlling TTY for processes that don't have one. (NetBSD source code shows that they do use `NODEV` placeholder value for `kproc->p_tdev`.) Although it might be safe to call [devname(3)](https://man.netbsd.org/devname.3) for a "no TTY device" placeholder. The `devname` function would return a placeholder string "??" which would not be ideal for displaying in htop. Hence the filter.

As `KERN_PROC_TTY_NODEV` is a negative constant. Compare with `~(KERN_PROC_TTY_NODEV)` would get rid of the signed and unsigned problem.

```c
         // KERN_PROC_TTY_NODEV is a negative constant but the type of 
         // kproc->p_tdev may be unsigned. 
         const char* name = ((dev_t)~kproc->p_tdev != (dev_t)~(KERN_PROC_TTY_NODEV)) ? devname(kproc->p_tdev, S_IFCHR) : NULL;
```

fraggerfox: @Explorer09 : Thank you for doing that investigation, shall I include your suggestion into this PR? (Now I remember about the `??` characters when I was doing the port originally)

I did a build locally and it seems to have worked!

Explorer09: @fraggerfox You can include it, but as a separate commit. This way if another BSD port of htop has the problem people can adapt the patch easily.

fraggerfox: Agreed, I shall do a follow up PR once this is merged.

netbsd/NetBSDProcessTable.c
```diff
@@ -151,7 +151,7 @@ static double getpcpu(const NetBSDMachine* nhost, const struct kinfo_proc2* kp)
 }
 
 static void NetBSDProcessTable_scanProcs(NetBSDProcessTable* this) {
-   const Machine* host = this->super.host;
+   const Machine* host = this->super.super.host;
```
BenBE, line 154: As `this->super` is handed to `ProcessTable_getProcess` below it might be worth splitting this assignment and extract `super` into it's own variable as is already done with most other platform implementations of `*ProcessTable_scanProcs`.

BenBE (approved):
Just a very small idea. Apart from this the changes look reasonable.

IIRC there was a similar PR done for FreeBSD/OpenBSD already. Maybe compare there too for changes.

netbsd/NetBSDProcessTable.c
fraggerfox: I had a look at this and simplified it based on what I saw in the `darwin` port.

- https://github.com/htop-dev/htop/blob/main/darwin/DarwinProcessTable.c#L72

I created a follow up PR against my branch to show the changes.

@BenBE : Would you like me to commit that also into this PR and merge both the commits in this single PR?

BenBE: Go ahead and cherry-pick. Changes look easy enough.

fraggerfox: Done.

BenBE (approved):
LGTM.

PR #1362, opened 2024-01-07
Clarify isNonnegative() behavior in comments

Explorer09:
"With -0.0 considered nonnegative". (The function is not intended to check the sign bit.)

No code changes.


BenBE:
The added comment as is does add more confusion than it resolves IMHO.

Explorer09:
@BenBE Do you have a better idea for the wording? Or I might just close this PR as unnecessary.

Macros.h
```diff
@@ -126,7 +126,8 @@ static inline bool isNaN(double x) {
    return !isgreaterequal(x, x);
 }
 
-/* Checks if x is nonnegative. Returns false if x is NaN. */
+/* Checks if x is nonnegative (with -0.0 considered nonnegative).
+   Returns false if x is NaN. */
```
BenBE, line 130: Possible wording?

```suggestion
/* Checks if x is nonnegative by checking x >= 0.0.
   Due to IEEE754 having a duplicate zero -0.0 is considered as nonnegative.
   Returns false if x is NaN. */
```

Explorer09: I've shortened your suggestion a bit. Let's see if the new description looks better.

Macros.h
```diff
@@ -119,19 +119,20 @@ in the source distribution for its full text.
...
    (ISO/IEC TS 18661-1 and the C23 standard stated that isnan() throws no
-   exceptions even with a "signaling" NaN) */
+   exceptions even with a "signaling NaN") */
 static inline bool isNaN(double x) {
    return !isgreaterequal(x, x);
 }
 
-/* Checks if x is nonnegative. Returns false if x is NaN. */
+/* Checks if x >= 0.0 but returns false if x is NaN. The -0.0 in IEEE 754 is
+   considered nonnegative by this function (-0.0 == 0.0). */
```
BenBE, line 130:
```suggestion
/* Checks if x >= 0.0 but returns false if x is NaN.
   Due to IEEE 754 -0.0 == 0.0, thus non-negative. */
```

Explorer09: Reworded the comment slightly.

Issue #1363, opened 2024-01-08
htop orange pi 


renewedbullet:
Does htop 3.X not include kernel and OS version and amount of users logged in like 2.2 did?

![image](https://github.com/htop-dev/htop/assets/140523069/ab7fd53b-15e7-4fb3-b303-63a35290b189)

renewedbullet:
similiar to this, 
![image](https://github.com/htop-dev/htop/assets/140523069/acd0653f-8a32-4030-a744-c133ce862d6e)

BenBE:
There's a Meter for the System in 3.3, which can be seen in [this screenshot](https://social.chaotikum.org/deck/@benbe/111303264986839094):
![image](https://github.com/htop-dev/htop/assets/111196/71097f0c-bb92-4ba7-b46d-788987b360fe)

The number of logged in users is not currently available as a meter in standard htop; this is a modification of some derivative versions; if you can look for those changes and point them here, we might be able to integrate them.

renewedbullet:
I apologize for being dumb, is this the files in which your looking for? 
[Downloads.zip](https://github.com/htop-dev/htop/files/13873931/Downloads.zip)

renewedbullet:
OS_Meter.c OS_meter.h kernel_Meter.c Kernel_meter.h

BenBE:
Yes, do you have a repository you can point to for these files?

I noticed they might likely be from:
https://github.com/avafinger/htop_2.2.2/

Given there's [some history with that fork of htop](https://github.com/htop-dev/htop/issues/523#issuecomment-821986130), I'm quite reluctant to import patches from there directly (also that fork is based on old htop 2.2). That being said, all the mentioned meters are easy enough to re-implement from scratch.

renewedbullet:
Yes it is, i wasn't aware of any conflicts I apologize just native htop  i installed was missing those figured it would be a nice QoL upgrade then I saw them on 2.2 was curious why an older version had it. If its to much no worries.

BenBE:
Writing those meters is not too difficult. For the OS/Kernel/Distro meters we already have the code to gather the information, so that's basically some boilerplate to implement.

For the logged-in user count the function to get that information is not to difficult either; it's just a matter of implementing it in the platform layer, as that meter is relevant for more than just Linux.

PR #1364, opened 2024-01-09
Linux: add meter showing the current active user sessions

cgzones:
Display the active user sessions either via systemd-logind or utmp.

Suggested in #1363.


UserSessionsMeter.c
```diff
@@ -0,0 +1,162 @@
...
+Released under the GNU GPLv2+, see the COPYING file
+in the source distribution for its full text.
+*/
+
+#include "config.h" // IWYU pragma: keep
+
+#include "UserSessionsMeter.h"
+
+#include <dlfcn.h>
+#include <utmpx.h>
```
BenBE, line 13: This should be moved to some platform implementation as this header does not necessarily exist on all platforms (e.g. WinNT). Putting the actual query part of the code in the `generic` directory is fine though, so most platforms may pull their implementation from there.

Similar things go for the systemd support in this, which are specific to Linux.

UserSessionsMeter.c
```diff
@@ -0,0 +1,162 @@
...
+static void UserSessionsMeter_done(ATTR_UNUSED Meter* this) {
+#ifndef BUILD_STATIC
+   if (dlopenHandle) {
+      dlclose(dlopenHandle);
+      dlopenHandle = NULL;
+   }
+#endif /* !BUILD_STATIC */
+}
+
+#if !defined(BUILD_STATIC) || defined(HAVE_LIBSYSTEMD)
```
BenBE, line 48: Doesn't this duplicate code we already have elsewhere for systemd?

UserSessionsMeter.c
```diff
@@ -0,0 +1,162 @@
...
+            }
+
+            if (sessions[i] == -1) {
+               ret++;
+               sessions[i] = ut->ut_session;
+               break;
+            }
+         }
+#else
+      ret++;
```
BenBE, line 110:
```suggestion
         ret++;
```

BenBE (changes requested):
What re OpenBSD/NetBSD/Solaris?

@fraggerfox Can you help here?

fraggerfox:
> What re OpenBSD/NetBSD/Solaris?
> 
> @fraggerfox Can you help here?

I can check this out in the coming weekend and see if I can make the changes work in NetBSD.

UserSessionsMeter.c
```diff
@@ -0,0 +1,162 @@
...
+
+static void UserSessionsMeter_updateValues(Meter* this) {
+   int ret;
+#if !defined(BUILD_STATIC) || defined(HAVE_LIBSYSTEMD)
+   ret = update_via_sd_login();
+   if (ret < 0)
+      ret = updateViaUtmp();
+#else
+   ret = updateViaUtmp();
+#endif /* !BUILD_STATIC || HAVE_LIBSYSTEMD */
```
Explorer09, line 127: 1. Can the systemd code depend only on `HAVE_LIBSYSTEMD`? People who build htop without systemd support would likely not want the systemd code even when it's dynamically loaded and linked.
2. Code simplification suggestion:


```c
   int ret = -1; // Or maybe -ENOSYS for better diagnostic message
#if defined(HAVE_LIBSYSTEMD)
   ret = update_via_sd_login();
#endif /* HAVE_LIBSYSTEMD */
   if (ret < 0)
      ret = updateViaUtmp();
```

BenBE: 1. There is a systemd meter. Linking should be done accordingly as it's already done there.

2. This looks a bit cleaner, but has a dead store in `ret`.

UserSessionsMeter.c
```diff
@@ -0,0 +1,162 @@
...
+#endif /* BUILD_STATIC */
+
+
+static void UserSessionsMeter_done(ATTR_UNUSED Meter* this) {
+#ifndef BUILD_STATIC
+   if (dlopenHandle) {
+      dlclose(dlopenHandle);
+      dlopenHandle = NULL;
+   }
+#endif /* !BUILD_STATIC */
```
Explorer09, line 45: Make the unload code a separate function, so that you won't need to duplicate the code for destroying `dlopenHandle`

```c
static void closeDlHandle(void) {
#ifndef BUILD_STATIC
   if (dlopenHandle) {
      dlclose(dlopenHandle);
      dlopenHandle = NULL;
   }
#else
   return; // No-op
#endif /* !BUILD_STATIC */
}

Comment (Explorer09) [reply UserSessionsMeter.c:45]: Sorry. I take this suggestion back. It would be way better if the handle, `dlopen` and `dlclose` code be shared with the SystemdMeter. But that's a code refactoring to be done...

Comment (BenBE) [reply UserSessionsMeter.c:45]: That's also why I suggested to move the actual reading of values into the platform code. Systemd is Linux-specific after all and each platform needs things to be done slightly different anyway; though the utmp stuff can mostly be shared as `generic` code, as is e.g. already done with the uname stuff for the System meter to read OS/kernel info.

File: UserSessionsMeter.c (hunk 2 above)
Comment (cgzones) [reply UserSessionsMeter.c]: Refactored the runtime loading of libsystemd.

File: UserSessionsMeter.c (hunk 1 above)
Comment (cgzones) [reply UserSessionsMeter.c]: Moved to *generic/utmpx.(c|h)*.

**Comment by cgzones** [comment]
> What re OpenBSD/NetBSD/Solaris?

According to manual pages only OpenBSD does have not support for it.

File: generic/utmpx.c [hunk 6]
```diff
@@ -0,0 +1,64 @@
...
+         for (size_t i = 0; i < sessions_size; i++) {
+            if (sessions[i] == ut->ut_session) {
+               found = true;
+               break;
+            }
+         }
+         if (!found) {
+            ret++;
+
+            if (sessions_size +1 >= sessions_cap) {
```
BenBE, line 43:
```suggestion
            if (sessions_size + 1 >= sessions_cap) {
```

UserSessionsMeter.c
```diff
@@ -0,0 +1,157 @@
...
+#include "generic/utmpx.h"
+
+#if defined(BUILD_STATIC) && defined(HAVE_LIBSYSTEMD)
+#include <systemd/sd-login.h>
+#endif
+
+#ifndef BUILD_STATIC
+#include <dlfcn.h>
+#include "linux/Platform.h"
+#endif
```
BenBE, line 29: There should not be a platform-specific reference in the platform-independent code section.

UserSessionsMeter.c
```diff
@@ -0,0 +1,157 @@
...
+   int ret;
+#if (!defined(BUILD_STATIC) && defined(HTOP_LINUX)) || defined(HAVE_LIBSYSTEMD)
+   ret = update_via_sd_login();
+   if (ret < 0)
+      ret = Generic_utmpx_get_users();
+#else
+   ret = Generic_utmpx_get_users();
+#endif /* (!BUILD_STATIC && HTOP_LINUX) || HAVE_LIBSYSTEMD */
+
+   users = ret;
```
BenBE, line 116: Move this code into a platform-specific function. For everything aside from Linux this can be a direct forward to `Generic_utmpx_get_users`; on Linux that's mostly the above block of code.

UserSessionsMeter.c
```diff
@@ -0,0 +1,157 @@
...
+
+   if (users >= 0) {
+      xSnprintf(this->txtBuffer, sizeof(this->txtBuffer), "%d", ret);
```
BenBE, line 119: Why distinguish here between `ret` and `users`?

UserSessionsMeter.c
```diff
@@ -0,0 +1,157 @@
...
+}
+
+static void UserSessionsMeter_display(ATTR_UNUSED const Object* cast, RichString* out) {
+   char buffer[16];
+
+   if (users >= 0) {
+      int len = xSnprintf(buffer, sizeof(buffer), "%d", users);
+      RichString_appendnAscii(out, CRT_colors[METER_VALUE], buffer, len);
+   } else {
+      RichString_appendAscii(out, CRT_colors[METER_VALUE], "N/A");
```
BenBE, line 132: Maybe shadow the value if not available?

BenBE (changes requested):

UserSessionsMeter.c
Explorer09: For meter text, I don't think shadowing is a good idea. IMHO.

linux/Platform.c
```diff
@@ -1085,3 +1091,34 @@ void Platform_done(void) {
    LibSensors_cleanup();
 #endif
 }
+
+#ifndef BUILD_STATIC
+
+static void* libsystemd_handle;
+static size_t libsystemd_loaded;
+
+void* Platform_load_libsystemd(void) {
```
Explorer09, line 1100: I write some ideas just in case you are planning to refactor this further:
1. The LibSystemd init code can become its own C files, `LibSystemd.c` and `LibSystemd.h`. Within it the `LibSystemd_init()` function would perform what you have `Platform_load_libsystemd()` here. Look at the existing `linux/LibSensors.c` file for what I mean. Dynamically loaded libraries should have the init and unload functions like that.
2. When a libsystemd handle has been `dlopen`'d, all the symbols htop requires should resolve all at once. No need to have the `resolve()` code in different places for the same library.
3. Make the resolved function pointers publicly assessible. That is, put these in `linux/LibSystemd.h` when needed:

```c
// In "linux/LibSystemd.h"
#ifdef BUILD_STATIC

#define sym_sd_bus_open_system sd_bus_open_system
#define sym_sd_bus_open_user sd_bus_open_user
#define sym_sd_bus_get_property_string sd_bus_get_property_string
#define sym_sd_bus_get_property_trivial sd_bus_get_prope
rty_trivial
#define sym_sd_bus_unref sd_bus_unref
#define sym_sd_get_sessions sd_get_sessions

#else

extern int (*sym_sd_bus_open_system)(sd_bus**);
extern int (*sym_sd_bus_open_user)(sd_bus**);
extern int (*sym_sd_bus_get_property_string)(sd_bus*, const char*, const char*, const char*, const char*, sd_bus_error*, char**);
extern int (*sym_sd_bus_get_property_trivial)(sd_bus*, const char*, const char*, const char*, const char*, sd_bus_error*, char, void*);
extern sd_bus* (*sym_sd_bus_unref)(sd_bus*);
extern int (*sym_sd_get_sessions)(char***);

#endif

PR #1370, opened 2024-01-11
CRT: handle/ignore signals

eworm-de:
The signals `SIGUSR1` and `SIGUSR2` should not kill the application.

This is split off from #1319 to keep the topics separated...


cgzones (approved):

BenBE (approved):

cgzones:
Thanks, merged.

PR #1371, opened 2024-01-11
fixes for illumos/Solaris platforms

hadfl:
This fixes build errors/warnings on illumos that were introduced in the most recent `htop` release (`3.3.0`), as well as a segmentation fault due to a non-allocated kstat structure.


fasterit (approved):
Confirmed that 3.3.0 is broken on Solaris and builds & runs after these fixes.
Not reviewed for correctness.

BenBE (approved):
Reviewed changes to make sense in the context of the recent refactorings of internal structures. LGTM.

Issue #1369, opened 2024-01-11
Running option is 0


FerranLlobetC:
Hi, 
In the standard htop version that can be installed via apt in Ubuntu (2.2.0) you have the "running" counter that counts the number of threads that are active, however it is quite old and I was interested in keeping an eye on Temperatures.

Therefore, I builded it from source and it functions properly, however this "running" counter does not work anymore.

I would appreicate any light in how to makre the "running" counter to work again. Do I have to enable other libraries for it to work? Is it an option that I need to toggle on?

Best,
Ferran.

BenBE:
A quick bisect showed this to be a regression introduced in 72235d8e098d9d79029dca65122605741e1aafad by @natoscott.

PR #1372, opened 2024-01-11
Fix the display of running tasks

BenBE:
Broken by: 72235d8e
Fixes: #1369

### DIFF
--- a/linux/LinuxMachine.c
+++ b/linux/LinuxMachine.c
@@ -489,8 +489,7 @@ static void LinuxMachine_scanCPUTime(LinuxMachine* this) {
    char buffer[PROC_LINE_LENGTH + 1];
    while (fgets(buffer, sizeof(buffer), file)) {
       if (String_startsWith(buffer, "procs_running")) {
-         ProcessTable* pt = (ProcessTable*) super->processTable;
-         pt->runningTasks = strtoul(buffer + strlen("procs_running"), NULL, 10);
+         this->runningTasks = strtoul(buffer + strlen("procs_running"), NULL, 10);
          break;
       }
    }


natoscott:
@BenBE thanks for triaging and fixing.  Can we do this though...?

```
--- /tmp/git-blob-ZYfkIM/LinuxMachine.h 2024-01-12 10:08:32.757491693 +1100
+++ linux/LinuxMachine.h        2024-01-12 10:05:06.305232398 +1100
@@ -61,7 +61,6 @@ typedef struct LinuxMachine_ {
    int pageSizeKB;
 
    /* see Linux kernel source for further detail, fs/proc/stat.c */
-   unsigned int runningTasks;   /* procs_running from /proc/stat */
    long long boottime;   /* btime field from /proc/stat */
 
    double period;
--- /tmp/git-blob-Tf77n4/LinuxProcessTable.c    2024-01-12 10:08:32.879491785 +1100
+++ linux/LinuxProcessTable.c   2024-01-12 10:08:28.100488151 +1100
@@ -1345,9 +1345,6 @@ static bool LinuxProcessTable_recursePro
    const struct dirent* entry;
 
-   /* set runningTasks from /proc/stat (from Machine_scanCPUTime) */
-   pt->runningTasks = lhost->runningTasks;
-
 #ifdef HAVE_OPENAT
    int dirFd = openat(parentFd, dirname, O_RDONLY | O_DIRECTORY | O_NOFOLLOW);
    if (dirFd < 0)
```

natoscott:
To answer my own question :) - no, we cannot - the field is zeroed in ProcessTable_prepareEntries so we need to keep the cached value from the /proc/stat scan, I remember now.  Your fix looks good, thanks.  I wonder if this code might be a little less confusing if these fields had different names though ... but that's unrelated to your fix.

natoscott (approved):
LGTM as per previous comment.

Issue #1373, opened 2024-01-12
Invalid encoding in cmdline leaves Command column empty


BenBE:
If there's a bad UTF-8/multibyte encoding in `cmdline` (or another column that is used to render the `Command` column), the resulting text stays empty.

Found while investigating #1346 with #1369 applied. Likely to happen on unpatched HEAD too.

Reproducer in `LinuxProcessTable_readCmdlineFile` to replace the contents of `command`:
```c
   const char foo[] = "test\0\xaa\0invisible\0";
   memcpy(command, foo, sizeof(foo));
   amtRead = sizeof(foo);
```

When such an invalid multi-byte sequence is found, the return in `RichString_writeFromWide` is taken due to `len==-1`:

```c
static inline int RichString_writeFromWide(RichString* this, int attrs, const char* data_c, int from, int len) {
   wchar_t data[len + 1];
   len = mbstowcs(data, data_c, len);
   if (len <= 0)
      return 0;
   // …
```

PR #1377, opened 2024-01-15
Fallback on printing invalid wide string

cgzones:
In case the string to be printed contains invalid wide characters, fall
back to print as ASCII string.

Closes: https://github.com/htop-dev/htop/issues/1373

### DIFF
--- a/RichString.c
+++ b/RichString.c
@@ -54,9 +54,42 @@ void RichString_rewind(RichString* this, int count) {
 
 #ifdef HAVE_LIBNCURSESW
 
+static size_t mbstowcs_nonfatal(wchar_t* dest, const char* src, size_t n) {
+   size_t written = 0;
+   mbstate_t ps = { 0 };
+   bool broken = false;
+
+   while (n > 0) {
+      size_t ret = mbrtowc(dest, src, n, &ps);
+      if (ret == (size_t)-1 || ret == (size_t)-2) {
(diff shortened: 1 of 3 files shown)


BenBE:
Patch itself LGTM, but: Isn't there an multi-byte conversion function that uses an replacement character in case of invalid sequences? I.e. use U+FFFD for the erroneous sequence?

Instead of moving things around, we'd probably forward-declare the static functions at the top of the file.

cgzones:
Used a different approach, retrying to interpret the remaining bytes still as multi-byte string.

RichString.c
```diff
@@ -54,9 +54,45 @@ void RichString_rewind(RichString* this, int count) {
 
 #ifdef HAVE_LIBNCURSESW
 
+static size_t mbstowcs_nonfatal(wchar_t* dest, const char* src, size_t n) {
+   wchar_t *d = dest;
```
Explorer09, line 58: 1. Style
```suggestion
   wchar_t* d = dest;
```

2. I didn't get why you are duplicating two pointers `src` and `dest` with `s` and `d` respectively where you can just increment `src` and `dest` directly.

BenBE: Same goes for `remaining` vs `n`.

`s` and `d` are not very descriptive.

Also arguments are by-value, thus you can change the pointers locally without issues.

PR #1380, opened 2024-01-17
Undo too-aggressive code collapsing from tree mode refactoring

### LINKED ISSUE #1378: Sorting key and direction change nothing in tree mode
Processes seem to always be sorted by PID in ascending order while in tree mode.

Screenshot_2024-01-16-225441
Screenshot_2024-01-16-225250

The sorting direction is changed in the screenshots, but the order of the processes is the same. Selecting other sorting keys also changes nothing.

The issue wasn't present in 3.2.2 and appeared in 3.3.0. I've bisected it down to 0f751e991d399769fb8d7800f7c4bccec2ca7f60. However, I couldn't figure out the exact reason since it seems to be a huge commit.

natoscott:
While abstracting processes from row display, code relating to row parent sorting was moved to row-level that in fact must be done at process-level (using the Row_compareByParent_Base call circumvented this unintentially).

Resolves #1378

### DIFF
--- a/Process.c
+++ b/Process.c
@@ -902,7 +902,10 @@ int Process_compare(const void* v1, const void* v2) {
 }
 
 int Process_compareByParent(const Row* r1, const Row* r2) {
-   int result = Row_compareByParent_Base(r1, r2);
+   int result = SPACESHIP_NUMBER(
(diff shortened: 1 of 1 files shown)


BenBE (approved):
LGTM. Took me a bit to process the subtle code flow change though.

PR #1381, opened 2024-01-18
darwin/Platform.c: do not include non-existing headers which break build

barracuda156:
Commit ed7eac5dfe8239f7942102f9d1a4f9ef5210f4d7 unconditionally includes a header which is not present on earlier macOS. This unnecessarily breaks the build. Use includes correctly.

### DIFF
--- a/darwin/Platform.c
+++ b/darwin/Platform.c
@@ -18,7 +18,7 @@ in the source distribution for its full text.
 #include <net/if_types.h>
 #include <net/route.h>
 #include <sys/socket.h>
-#include <sys/_types/_mach_port_t.h>
+#include <mach/port.h>
 
 #include <CoreFoundation/CFBase.h>
 #include <CoreFoundation/CFDictionary.h>


barracuda156:
Everything looks good now: 
htop_ppc

darwin/Platform.h
```diff
@@ -10,6 +10,7 @@ in the source distribution for its full text.
 
 #include <stdbool.h>
 #include <sys/types.h>
+#include <AvailabilityMacros.h>
```
BenBE, line 13: Isn't this only used in `darwin/Platform.c`? If so, it should go there instead as this header is not required to form the public interface of `darwin/Platform.h`.

darwin/Platform.c
```diff
@@ -18,7 +18,11 @@ in the source distribution for its full text.
 #include <net/if_types.h>
 #include <net/route.h>
 #include <sys/socket.h>
+#if MAC_OS_X_VERSION_MIN_REQUIRED < 1090
+#include <mach/mach_port.h>
+#else
 #include <sys/_types/_mach_port_t.h>
+#endif
```
BenBE, line 25: This should go into the list further down with the rest of conditional defines.

Also compare re `HAVE_MACH_MACH_TIME_H` defined via `configure.ac`/`config.h`.

If checking this via the availability macros, a check could be placed like this:
```suggestion
#include <AvailabilityMacros.h>
#if MAC_OS_X_VERSION_MIN_REQUIRED < 1090
#include <mach/mach_port.h>
#else
#include <sys/_types/_mach_port_t.h>
#endif
```

or

```suggestion
#if HAVE_MACH_MACH_PORT_H
#include <mach/mach_port.h>
#else
#include <sys/_types/_mach_port_t.h>
#endif
```

BenBE (changes requested):

barracuda156:
@BenBE Thank you, I have dropped AvailabilityMacros header from Platform.h now, the fallback now is in Platform.c. 

(Perhaps there is little point to make a specific check here, we know that prior to 10.9 there is no folder for types in /usr/include/sys, SDKs are public, and the files are specifically for Darwin, no surprises here.)

Explorer09:
@barracuda156 htop coding style prefers not to have a version number check in order to determine whether a header is available. You may mention the version number in code comments, but we would like the header availability be checked in `configure.ac` instead.

barracuda156:
@Explorer09 Could you please review?

darwin/Platform.c
```diff
@@ -18,7 +18,11 @@ in the source distribution for its full text.
 #include <net/if_types.h>
 #include <net/route.h>
 #include <sys/socket.h>
+#ifdef HAVE_SYS__TYPES__MACH_PORT_T_H
 #include <sys/_types/_mach_port_t.h>
+#else
+#include <mach/mach_port.h>
```
Explorer09, line 24: 1. Conditional include lines should be moved to below the mandatory include lines.
2. When I search the code on the web, the type `mach_port_t` seems to be defined in `<mach/port.h>`. I want to make sure you did not confuse `<mach/mach_port.h>` with `<mach/port.h>`.

barracuda156: Let me verify and fix if needed.

barracuda156: @Explorer09 Ok, so `mach_port_t` is in `mach/port.h`, however `mach/mach_port.h` [includes](https://github.com/alexey-lysiuk/macos-sdk/blob/a79de83d781f7fb76d329aee456bab9f72f935d1/MacOSX10.6.sdk/usr/include/mach/mach_port.h#L14) the former, so it works.
I do not know which is preferable to include.
What do you think? Here are the headers from 10.6, for example:
https://github.com/alexey-lysiuk/macos-sdk/blob/main/MacOSX10.6.sdk/usr/include/mach/mach_port.h
https://github.com/alexey-lysiuk/macos-sdk/blob/main/MacOSX10.6.sdk/usr/include/mach/port.h

Explorer09: Does htop build on your Mac if you simply include `<mach/port.h>`? If it does build, I don't think you need to include `<mach/mach_port.h>`. There is no indication that `<mach/port.h>` is a private header and that we should not include directly.

barracuda156: Apparently it does. I have changed the header, as you advised, and moved the chunk below.

darwin/Platform.c
```diff
@@ -63,6 +62,12 @@ in the source distribution for its full text.
 #include <mach/mach_time.h>
 #endif
 
+#ifdef HAVE_SYS__TYPES__MACH_PORT_T_H
+#include <sys/_types/_mach_port_t.h>
```
ryandesign, line 66: I'm surprised by the inclusion of a system header whose name begins with an underscore. I always assumed that meant it was a private header, not intended to be included directly and subject to change in the future.

barracuda156: @ryandesign This is not my addition though. It was added in https://github.com/htop-dev/htop/commit/ed7eac5dfe8239f7942102f9d1a4f9ef5210f4d7 and has broken everything which does not have this header. I am merely trying to fix it for those systems without changing the code for the new ones.

ryandesign: @barracuda156 Exactly, you correctly noted that including `<sys/_types/_mach_port_t.h>` causes a problem, but your proposal makes the code include that header sometimes and a different header at other times. My question is whether it wouldn't be better never to include `<sys/_types/_mach_port_t.h>` and to instead always include whatever header is documented to provide whatever the functionality is that is needed here.

I assume that the wanted functionality is the definition of `mach_port_t` since that's the only thing in `<sys/_types/_mach_port_t.h>` in the macOS 11 SDK. I've never found the Mach stuff to be documented very well. [Apple's site documents](https://developer.apple.com/documentation/kernel/mach_port_t) that `mach_port_t` has existed since Mac OS X 10.0 but nothing else about it, such as what header to include. Maybe it used to be better documented and Apple removed that documentation at some point.

grepping through headers, I see that `_mach_port_t.h` is included by `<pthread/pthread.h>` and `<mach/port.h>`, so I'm guessing the latter is the one that should always be included.

barracuda156: @ryandesign Could you confirm it works for Intel? If yes, then we should just go with that, and drop all fancy checks.

ryandesign: I can't test it at the moment but it seems like it should work.

Oh, and since `<mach/port.h>` is documented to exist since the beginning of time you won't need to check for its existence.

ryandesign: It looks like htop has a GitHub Actions workflow set up that builds on macOS, so hopefully that is sufficient to verify any proposed change does not break the build on current macOS.

barracuda156: Ok let me rebase.

barracuda156:
@Explorer09 @BenBE I have switched to a simpler version which just replaces the included header with the universally supported one, as suggested by @ryandesign above.

BenBE (approved):

Issue #1382, opened 2024-01-18
[macOS] Fix CPU type detection for PowerPC


barracuda156:
The current code in `PlatformHelpers.c` uses `machdep.cpu.brand_string` to get CPU type: on my MacBook Air it returns Apple M1, for example.
However this is not supported on PowerPC Macs, so instead we get a warning that CPU is unknown.

In fact, there is a way to get CPU, which works across the board: via `hostinfo`. On a PowerMac I get:
```
36-246% hostinfo
Mach kernel version:
	 Darwin Kernel Version 10.0.0d2: Fri Oct 10 19:37:52 PDT 2008; root:xnu-1346.7~1/RELEASE_PPC
Kernel configured for up to 4 processors.
4 processors are physically available.
4 processors are logically available.
Processor type: ppc970 (PowerPC 970)
Processors active: 0 1 2 3
Primary memory available: 16.00 gigabytes
Default processor set: 65 tasks, 347 threads, 4 processors
Load average: 1.24, Mach factor: 2.76
```
What is needed is `Processor type` line.

In the terminal I can extract CPU type in a number of ways:
```
36-246% hostinfo | grep type | sed 's/Processor type\: //'
ppc970 (PowerPC 970)
36-246% hostinfo | awk -F'[()]' '{print $2}'





PowerPC 970
```

Alternatively, `sysctl` can be used, but output values have to be mapped to readable ones:
```
36-246% sysctl hw.cputype hw.cpusubtype
hw.cputype: 18
hw.cpusubtype: 100
```
18 stands for any PowerPC.
100 = ppc970
11 = ppc7450
10 = ppc7400

Can either of these be used in the code here?

barracuda156:
P. S. Or maybe just use `sysctl -n hw.cpusubtype`:

```
36-246% sysctl -n hw.cpusubtype
100
```

barracuda156:
Refs: https://github.molgen.mpg.de/pages/bs/macOSnotes/mac/mac_procs_cpu.html
https://cr.yp.to/hardware/ppc.html

fasterit:
Closing a per the discussion in #1383

PR #1383, opened 2024-01-18
darwin/PlatformHelpers: fix CPU detection for PowerPC

barracuda156:
Closes: https://github.com/htop-dev/htop/issues/1382

Notice, that hw.cpusubtype outputs a numerical code of a CPU. For PowerPC, 100 = G5, 11 = ppc7450, 10 = ppc7400.

P. S. As far as I can see, the resulting string is only used on Apple Silicon, and in other cases it is either something (cpu is known) or unknown. It is desirable to avoid a bogus warning on PowerPC – we know those cpus very well :)

### DIFF
--- a/darwin/PlatformHelpers.c
+++ b/darwin/PlatformHelpers.c
@@ -60,6 +60,10 @@ bool Platform_KernelVersionIsBetween(KernelVersion lowerBound, KernelVersion upp
 
 void Platform_getCPUBrandString(char* cpuBrandString, size_t cpuBrandStringSize) {
    if (sysctlbyname("machdep.cpu.brand_string", cpuBrandString, &cpuBrandStringSize, NULL, 0) == -1) {
+   #ifdef __POWERPC__
+      if (sysctlbyname("hw.cpusubtype", cpuBrandString, &cpuBrandStringSize, NULL, 0) != -1)
+         return;
+   #endif
       fprintf(stderr,
          "WARN: Unable to determine the CPU brand string.\n"
          "errno: %i, %s\n", errno, strerror(errno));


darwin/PlatformHelpers.c
```diff
@@ -59,7 +59,11 @@ bool Platform_KernelVersionIsBetween(KernelVersion lowerBound, KernelVersion upp
 }
 
 void Platform_getCPUBrandString(char* cpuBrandString, size_t cpuBrandStringSize) {
+#ifdef __POWERPC__
+   if (sysctlbyname("hw.cpusubtype", cpuBrandString, &cpuBrandStringSize, NULL, 0) == -1) {
+#else
    if (sysctlbyname("machdep.cpu.brand_string", cpuBrandString, &cpuBrandStringSize, NULL, 0) == -1) {
+#endif
```
BenBE, line 66: I think this one might be a bit more obvious in its intention?
```suggestion
   if (sysctlbyname("machdep.cpu.brand_string", cpuBrandString, &cpuBrandStringSize, NULL, 0) == -1) {
#ifdef __POWERPC__
      if (sysctlbyname("hw.cpusubtype", cpuBrandString, &cpuBrandStringSize, NULL, 0) != -1)
         return;
#endif
```

barracuda156:
@BenBE Please have a look.

BenBE:
So effectively with this patch, instead of say "Intel i7-7590K" you now have "100" as the brand string? I think that might look a bit surprising and I think from an UX perspective having real strings there would be better.

barracuda156:
Is this value actually used anywhere? I have only found it being used in a condition to detect arm64.
If cpu name is used elsewhere, then of course we should either map numerical codes to cpu names or parse the output of hostinfo instead, which gives out cpu name as a string among other info.

cgzones:
The return value of `Platform_getCPUBrandString()` is currently only used in a single check:

https://github.com/htop-dev/htop/blob/240dd4ab50e1a753905c711f99c57408e18d08a3/darwin/PlatformHelpers.c#L114

What about removing the warning or only show it in debug mode?

cgzones:
The only usage might even be removed, see #1285.

barracuda156:
@cgzones Thanks. Then I guess we close this for now?

Once there will be a practical case for detecting PowerPC on macOS, this can be reopened and adjusted accordingly.

PR #1386, opened 2024-01-20
Explicitly check sscanf(3) and fscanf(3) return values

cgzones:
Compare the return value of sscanf(3) and fscanf(3) explicitly against the expected number of parsed items to avoid matching the return value EOF.

Reported by CodeQL.

### DIFF
--- a/Header.c
+++ b/Header.c
@@ -86,10 +86,9 @@ static void Header_addMeterByName(Header* this, const char* name, MeterModeId mo
    unsigned int param = 0;
    size_t nameLen;
    if (paren) {
-      int ok = sscanf(paren, "(%10u)", &param); // CPUMeter
-      if (!ok) {
+      if (sscanf(paren, "(%10u)", &param) != 1) { // not CPUMeter
          char dynamic[32] = {0};
-         if (sscanf(paren, "(%30s)", dynamic)) { // DynamicMeter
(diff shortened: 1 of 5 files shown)


Explorer09:
"To avoid the matching the return value EOF"? That doesn't look like what this patch is doing. It's more about making the logic more robust by rejecting partially scanned results (that is, return value less than expected number of items but greater than the `EOF` constant).

Explorer09:
"To avoid matching the return value EOF"? That doesn't look like what this patch is doing. It's more about making the logic more robust by rejecting partially scanned results (that is, return value less than expected number of items but greater than the EOF constant).

Except for the commit descriptions, the other parts of the patch looks good to me.

Header.c
```diff
@@ -86,10 +86,9 @@ static void Header_addMeterByName(Header* this, const char* name, MeterModeId mo
    unsigned int param = 0;
    size_t nameLen;
    if (paren) {
-      int ok = sscanf(paren, "(%10u)", &param); // CPUMeter
-      if (!ok) {
+      if (sscanf(paren, "(%10u)", &param) != 1) { // CPUMeter
```
BenBE, line 89: Shouldn't this be `==` here instead?

cgzones: No, we want to enter the block and scan for a dynamic Meter if the parameter does not match a number.

PR #1342, opened 2023-12-03
Use sysconf() to determine active and existing CPUs on Linux

glaubitz:
Instead of enumerating the number of active and existing CPUs on Linux by iterating over the pseudo-files below /sys/devices/system/cpu/, it is better to retrieve the number of active and existing CPUs using the sysconf() function using _SC_NPROCESSORS_ONLN and _SC_NPROCESSORS_CONF which will always report the correct numbers.

Fixes #757

Issue #1198, opened 2023-02-25
Dynamic display of online CPUs in CPU meter


BenBE:
This is a follow up to #993 / #995 / #1195 / #1197 aimed at systems with dynamic power management of CPU cores.

When a system has many cores and most of them are offline (e.g. due to power saving mode or because of containerization) these offline CPUs should be hidden while offline, but shown once first active (or recently active). An additional color coding similar to #74 / #241 might be added to indicate cores that recently became active or were recently disabled.

Basically, what is needed is:

1. Track which cores have been seen online (and when)
2. Hide cores that have not been online (recently)
3. Implement some highlighting similar to #74 / #241 for CPU cores that have recently changed their state (possibly with flapping detection when the state changes regularly)
4. Create some configuration for this additional behavior as either an additional type of CPU meter OR as part of the normal CPU meters
5. Better display CPU node hierarchy information (physical id, core id, apic id, core type re SMT/BIG.little)

/cc @fasterit

glaubitz:
As explained in my comment [here](https://github.com/htop-dev/htop/issues/757#issuecomment-1807210492), I think the main problem is that the Linux version of `updateCPUcount()` checks for available processors instead (equivalent to `sysconf(_SC_NPROCESSORS_CONF)`) while the Solaris version of that function checks for online processors by correctly using `sysconf(_SC_NPROCESSORS_ONLN)`.

I would simply switch to using `sysconf(_SC_NPROCESSORS_ONLN)` on Linux as well instead of parsing through `/sys`.

C0rn3j:
@glaubitz That won't work as that will just reintroduce https://github.com/htop-dev/htop/issues/1195 again, but worse, as commented on the PR draft that https://github.com/htop-dev/htop/pull/1394 supersedes.  
Either of the two PRs is not related to this one as neither actually *hides* the view.

While this is being plumbed it would be nice to fix the terminology - CPU, socket, cores, threads, … - at least the linux Machine file has "cpu" used everywhere for threads.

RSully:
I suspect however this would be implemented would solve my problem: running htop in an LXC with only, for example, 4 cores on a 32 core system, the meters are the top are impossible to parse quickly. It would be great if only the 4 "online" cores showed up so I could at-a-glance see my utilization within the container

glaubitz:
I think the current design is already much better than before as it marks unavailable CPUs as offline.

Would be nice though if showing offline CPUs could be toggled with a hotkey.

[...]

PR #1394, opened 2024-01-30
Linux: Correctly mark trailing offline CPU threads

C0rn3j:
htop checks cpu IDs in /proc/stat, and marks any in-between IDs as missing.  
This check was counting with leading offline threads(somewhat accidentally, as cpu0 is total in the code), but not trailing.

Left is 3.3.0, Right is 3.3.0 with this PR.
image

This supersedes #1342 which attempts to fix the same thing, but ignores thread IDs and just marks them active sequentially from 0.

It also fixes one of the two problems in #757 and the remaining issue is already tracked in #1198, so that can be closed after merge.

This was introduced in https://github.com/htop-dev/htop/commit/41af31be7ffbd34518b27aad56a4f54af6a8adf3#diff-7b4ee00635e2355bc3b83ada6e57991d2a79ba37828dab7128c461c14645580aR1865

### DIFF
--- a/linux/LinuxMachine.c
+++ b/linux/LinuxMachine.c
@@ -408,7 +408,9 @@ static void LinuxMachine_scanCPUTime(LinuxMachine* this) {
    if (!file)
       CRT_fatalError("Cannot open " PROCSTATFILE);
 
-   unsigned int lastAdjCpuId = 0;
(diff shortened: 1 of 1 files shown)


cgzones (approved):

cgzones:
Thanks, merged.

PR #1395, opened 2024-02-05
Minor bug fixes for Coverity issues

kmcdonell:
Over in the PCP project, Coverity has spotted a couple of minor issues in the htop code.

This commit addresses them:

Settings.c
    need to take control of the umask before calling mkstemp()

pcp/Platform.c
    Although Platform_getMaxPid() returns a pid_t (which
    strictly speaking is signed), the value is used to call
    Row_setPidColumnWidth() and a negative value here is not good
    ... using INT_MAX instead of UNIT_MAX on the (unlikely) error
    path makes no practical difference but may keep Coverity quiet

### DIFF
--- a/Settings.c
+++ b/Settings.c
@@ -622,8 +622,11 @@ int Settings_write(const Settings* this, bool onCrash) {
       fd = stderr;
       separator = ';';
    } else {
+      /* create tempfile with mode 0600 */
+      mode_t	cur_umask = umask(S_IXUSR | S_IRWXG | S_IRWXO);
       xAsprintf(&tmpFilename, "%s.tmp.XXXXXX", this->filename);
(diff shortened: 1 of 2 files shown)


Settings.c
```diff
@@ -622,8 +622,11 @@ int Settings_write(const Settings* this, bool onCrash) {
       fd = stderr;
       separator = ';';
    } else {
+      /* create tempfile with mode 0600 */
+      mode_t	cur_umask = umask(S_IXUSR | S_IRWXG | S_IRWXO);
```
BenBE, line 626: Indentation
```suggestion
      mode_t cur_umask = umask(S_IXUSR | S_IRWXG | S_IRWXO);
```

PR #1400, opened 2024-02-24
Add *BSD / Solaris to the CI

fraggerfox:
Add support for building `htop` in DragonFlyBSD /  FreeBSD / NetBSD / OpenBSD / Solaris environments in the CI workflow.

> NOTE: In future, this should help catch any changes that can potentially break the BSD / Solaris builds.

Workflows adapted from [btop's GHAs](https://github.com/aristocratos/btop/tree/main/.github/workflows).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -218,6 +218,31 @@ jobs:
               ./configure --enable-unicode --enable-werror
               gmake -k
 
+  build-netbsd-latest-gcc:
+    runs-on: ubuntu-22.04
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+      - name: Compile
+        uses: vmactions/netbsd-vm@v1
+        with:
+          release: '9.3'
+          usesh: true
+          prepare: |
+            PATH="/usr/pkg/sbin:/usr/pkg/bin:$PATH"
+            PKG_PATH="https://ftp.netbsd.org/pub/pkgsrc/packages/NetBSD/amd64/9.3/All/"
+            export PATH PKG_PATH
+            /usr/sbin/pkg_add pkgin
+            pkgin -y install autoconf automake libtool ncurses ncursesw gmake git
+            git config --global --add safe.directory /home/runner/work/htop/htop
+          run: |
+            set -e
+            ./autogen.sh
+            CPPFLAGS="-I/usr/pkg/include" ./configure --enable-unicode --enable-werror
+            gmake -k
+
   whitespace_check:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -243,6 +243,29 @@ jobs:
             CPPFLAGS="-I/usr/pkg/include" ./configure --enable-unicode --enable-werror
             gmake -k
 
+  build-openbsd-latest-gcc:
+      runs-on: ubuntu-22.04
+      timeout-minutes: 20
+      steps:
+        - uses: actions/checkout@v4
+          with:
+            submodules: recursive
+        - name: Compile
+          uses: vmactions/openbsd-vm@v1
+          with:
+            release: '7.4'
+            usesh: true
+            prepare: |
+              pkg_add gmake autoconf-2.71 automake-1.16.5 git
+              git config --global --add safe.directory /home/runner/work/htop/htop
+            run: |
+              set -e
+              export AUTOCONF_VERSION=2.71
+              export AUTOMAKE_VERSION=1.16
+              ./autogen.sh
+              ./configure --enable-unicode --enable-werror
+              gmake -k
+
   whitespace_check:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -197,6 +197,27 @@ jobs:
     - name: Distcheck
       run: make distcheck DISTCHECK_CONFIGURE_FLAGS="--enable-werror"
 
+  build-freebsd-latest-clang:
+      runs-on: ubuntu-22.04
+      timeout-minutes: 20
+      steps:
+        - uses: actions/checkout@v4
+          with:
+            submodules: recursive
+        - name: Compile
+          uses: vmactions/freebsd-vm@v1
+          with:
+            release: '14.0'
+            usesh: true
+            prepare: |
+              pkg install -y gmake autoconf-2.71 automake git
+              git config --global --add safe.directory /home/runner/work/htop/htop
+            run: |
+              set -e
+              ./autogen.sh
+              ./configure --enable-unicode --enable-werror
+              gmake -k
+
   whitespace_check:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -197,6 +197,29 @@ jobs:
     - name: Distcheck
       run: make distcheck DISTCHECK_CONFIGURE_FLAGS="--enable-werror"
 
+  build-dragonflybsd-latest-gcc:
+      runs-on: ubuntu-22.04
+      timeout-minutes: 20
+      steps:
+        - uses: actions/checkout@v4
+          with:
+            submodules: recursive
+        - name: Compile
+          uses: vmactions/dragonflybsd-vm@v1
+          with:
+            release: '6.4.0'
+            usesh: true
+            prepare: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -197,6 +197,28 @@ jobs:
     - name: Distcheck
       run: make distcheck DISTCHECK_CONFIGURE_FLAGS="--enable-werror"
 
+  build-freebsd-latest-clang:
```
Explorer09, line 200: I would say this is a "freebsd-vm", and not "freebsd", and thus I would name this build `build-freebsd-vm-14-clang`.

Explorer09: Likewise, `build-netbsd-vm-9.3-gcc` `build-openbsd-vm-7.4-gcc`

fraggerfox: I do not mind making that change, does this mean the GitHub hosted runners are all bare metal (the ubuntu builds) and not running "vm"s?

Explorer09: No. My idea was to distinguish OSes running inside a "container" and OSes that are fully virtualized. Since what you are doing is a VM running inside another VM. And I was initially confused when there is a `runs-on: ubuntu` line there.

Explorer09:
Your latest DragonFlyBSD commit (00bca18f3415effbc71e8a9608993192a0ca98b6) caused a build failure, and I noticed one thing in the build log: The configure scripts detected gcc and tried to use it instead of clang. Did you notice that?

Also, the line `mv htop htop-clang-"$GIT_HASH"` got run even when `make` produced errors. (What's the purpose of this rename, by the way?)

fraggerfox:
> Your latest DragonFlyBSD commit ([00bca18](https://github.com/htop-dev/htop/commit/00bca18f3415effbc71e8a9608993192a0ca98b6)) caused a build failure, and I noticed one thing in the build log: The configure scripts detected gcc and tried to use it instead of clang. Did you notice that?
> 

Yes, I was just testing out how DragonFlyBSD works, I was assuming it would be using `clang` instead of `gcc` (like in FreeBSD), but that is a minor change to do. I was more curious on the missing `ncurses` library. I am trying to get a copy of DragonFlyBSD to boot locally in my system so that I can check this out.

> Also, the line `mv htop htop-clang-"$GIT_HASH"` got run even when `make` produced errors. (What's the purpose of this rename, by the way?)

The rename if it fails (i.e. the binary was not built) would cause the workflow to fail, I did that since it is the easiest way I could check if the build failed or not.

.github/workflows/ci.yml
BenBE: As if all the other builds are done on real hardware … ;-) The important part IMHO for those tests is that the kernel side of things is the real deal™, which is satisfied if you don't just containerize, but run a VM (para-virtualized or full).

Explorer09:
> An example of a build failure but not a failed step can be found here
> 
>     * https://github.com/htop-dev/htop/actions/runs/8029660529/job/21936201103#step:3:1724

Does a "`set -e`" line work within a CI script file?
That could save you some `|| exit 1` code there.

fraggerfox:
> > An example of a build failure but not a failed step can be found here
> > ```
> > * https://github.com/htop-dev/htop/actions/runs/8029660529/job/21936201103#step:3:1724
> > ```
> 
> Does a "`set -e`" line work within a CI script file? That could save you some `|| exit 1` code there.

I tested out this suggestion and it seem to have worked. Thank you.

- https://github.com/fraggerfox/htop-dev/actions/runs/8049339870/job/21982494851

I have updated the file with the changes.

BenBE (approved):
LGTM.

You could fixup the `set -e` patch into the individual platforms, but that's optional.

.github/workflows/ci.yml
```diff
@@ -197,6 +197,119 @@ jobs:
...
+            prepare: |
+              pkg install -y gmake autoconf-2.71 automake ncurses git
+              git config --global --add safe.directory /home/runner/work/htop/htop
+            run: |
+              set -e
+              export LDFLAGS="-L/usr/local/lib"
+              export CFLAGS="-I/usr/local/include"
+              ./autogen.sh
+              ./configure --enable-unicode
+              gmake
```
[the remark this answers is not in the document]
fraggerfox: Good idea, I have updated the command.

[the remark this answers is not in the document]
BenBE: Same should be done for regular make too … Good point.

BenBE:
Some ideas, no need to implement them right now:
* What about ZFS support on FreeBSD/NetBSD? Shouldn't we try to build on the CI with such support enabled?
* What about PCP builds? @natoscott AFAIR PCP should build on all platforms¹, right?
* Other optional libs we should test in the CI?

¹given the right dependencies are installed

fasterit (approved):

fasterit:
> What about ZFS support on FreeBSD/NetBSD? Shouldn't we try to build on the CI with such support enabled?

This is not gated by a config-flag. So it is build in automatically (on FreeBSD).
Now, having test-cases is another thing...

/DLange

.github/workflows/ci.yml
```diff
@@ -197,6 +197,119 @@ jobs:
...
+            usesh: true
+            prepare: |
+              pkg install -y gmake autoconf-2.71 automake ncurses git
+              git config --global --add safe.directory /home/runner/work/htop/htop
+            run: |
+              set -e
+              export LDFLAGS="-L/usr/local/lib"
+              export CFLAGS="-I/usr/local/include"
+              ./autogen.sh
+              ./configure --enable-unicode
```
Explorer09, line 220: I see other platforms in this .yml file have `--enable-werror` option specified in `./configure`. Is it a good idea to have `--emable-werror` here as well? (I'm not sure; if it could pop out a lot of compiler warnings then it's probably not a good idea.)

natoscott:
> Some ideas, no need to implement them right now:
> [...]
>     * What about PCP builds? @natoscott AFAIR PCP should build on all platforms¹, right?
> ¹given the right dependencies are installed

Yes, many platforms are supported by PCP - the issue will be ensuring the PCP bits are readily accessible on non-Linux platforms.

If possible, it'd be great to get PCP builds for non-Linux platforms here in htop.dev repo, however in practice I think it will be more practical for this to happen via the actual PCP build (which builds a vendored htop git repo via git-subtree).  Perhaps best will be if we automate the sync between PCP and htop (currently its manual) so that commits here automatically trigger a PCP build.

https://github.com/performancecopilot/pcp/tree/main/vendor/github.com/htop-dev/htop

.github/workflows/ci.yml
BenBE: We should aim to build without warnings for a "default" build. Skimming over the build logs it doesn't look like we have too many warnings.

Two options:
1. Go ahead without `-enable-werror` for now and enable it as a new PR
2. Enable it now and see what breaks.

If simply enabling it now works without causing compilation errors for any of the new platforms, I'd suggest to enable it now. Otherwise it's best to have it split.

BenBE:
@natoscott Thus it's best to skip the PCP stuff in this PR.

.github/workflows/ci.yml
fraggerfox: I have added `--enable-werror` for *BSD / Solaris, let us see what happens.

fraggerfox: There was a minor change needed for OpenBSD, I have pushed in that change to see if it works.

@BenBE: The change is a one liner, would you like me to split the PR? (enable `--enable-werror` for OpenBSD alone along with the fix).

BenBE: Updated the PR accordingly.

Explorer09:
May I bump this PR and mention an issue of it?

The "`build-openbsd-latest-gcc`" CI job in commit 1fa7d6c73b54460125aa5ac989131921bfd93bf3 does not actually run GCC, but Clang under the program name "`cc`" ! I discovered this when it build-failed in one of my commits and threw me a compile warning. Please either correct the job name or do something to let GCC run in there.

fraggerfox:
> May I bump this PR and mention an issue of it?
> 
> The "`build-openbsd-latest-gcc`" CI job in commit [1fa7d6c](https://github.com/htop-dev/htop/commit/1fa7d6c73b54460125aa5ac989131921bfd93bf3) does not actually run GCC, but Clang under the program name "`cc`" ! I discovered this when it build-failed in one of my commits and threw me a compile warning. Please either correct the job name or do something to let GCC run in there.

You are right, I also over looked this, I can raise a PR to fix this and update the name to `clang` instead of `gcc`.

Issue #1408, opened 2024-03-06
htop crashed


matthiaskrgr:
I found this, unfortunately I did not see it crash life, I think the htop had been running for a couple of days:

````
at 12:19:38 ❯ htop


FATAL PROGRAM ERROR DETECTED
============================
Please check at https://htop.dev/issues whether this issue has already been reported.
If no similar issue has been reported before, please create a new issue with the following information:
  - Your htop version: '3.3.0'
  - Your OS and kernel version (uname -a)
  - Your distribution and release (lsb_release -a)
  - Likely steps to reproduce (How did it happen?)
  - Backtrace of the issue (see below)

Error information:
------------------
A signal 7 (Bus error) was received.

Setting information:
--------------------
[...]

PR #1409, opened 2024-03-08
Avoid glibc FILE voodoo

BenBE:
Addresses #1408

Background: Using `xReadfileat` avoids the internal buffer handling from glibc, which should improve robustness in low-memory situations as detailled in #1408. Avoiding any dynamic memory allocation also improves on the performance side of things as no dynamic allocations need to be managed (Except for the changed code formatting this IS even shorter ;-)).

### DIFF
--- a/linux/LinuxProcessTable.c
+++ b/linux/LinuxProcessTable.c
@@ -682,13 +682,15 @@ static void LinuxProcessTable_readMaps(LinuxProcess* process, openat_arg_t procF
 }
 
 static bool LinuxProcessTable_readStatmFile(LinuxProcess* process, openat_arg_t procFd, const LinuxMachine* host) {
-   FILE* statmfile = fopenat(procFd, "statm", "r");
-   if (!statmfile)
+   char statmdata[128] = {0};
+
+   if (xReadfileat(procFd, "statm", statmdata, sizeof(statmdata)) < 1) {
       return false;
+   }
 
    long int dummy, dummy2;
 
-   int r = fscanf(statmfile, "%ld %ld %ld %ld %ld %ld %ld",
+   int r = sscanf(statmdata, "%ld %ld %ld %ld %ld %ld %ld",
                   &process->super.m_virt,
                   &process->super.m_resident,
                   &process->m_share,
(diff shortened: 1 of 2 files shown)


fasterit (approved):

linux/LinuxProcessTable.c
```diff
@@ -682,21 +682,22 @@ static void LinuxProcessTable_readMaps(LinuxProcess* process, openat_arg_t procF
...
+   char statmdata[128] = {0};
+
+   if (xReadfileat(procFd, "statm", statmdata, sizeof(statmdata)) < 1) {
       return false;
+   }
 
    long int dummy, dummy2;
 
-   int r = fscanf(statmfile, "%ld %ld %ld %ld %ld %ld %ld",
+   int r = sscanf(statmdata, "%ld %ld %ld %ld %ld %ld %ld",
```
Explorer09, line 693: I wonder if it is a good idea to skip the two dummy arguments by using the format `"%ld %ld %ld %ld %*ld %ld %*ld"` ?

BenBE: This is kinda unrelated to what this PR is aiming at (further patches incoming).

Also, given this is all stack memory, it's not a memory issue (focus of this PR), but a performance one. And if we were to go for the performance aspect (no plans as of now) we'd still have my `fast_strtoull` implementation here …

Explorer09: I'm okay if the `LinuxProcessTable` performance improvements are addresses in a separate commit or PR.
Just to note that there are rooms to improve the code overall.
(BTW, I think I can slightly tweak the `fast_strtoull_hex` so that it uses one less instruction. Mind if I submit a draft for that?)

BenBE: As the `fast_strtoull_hex` is used in some time-critical places, submitting a draft for a performance improvement for it would be good. Please include measurements with valgrind/callgrind or some other performance measurement tool to compare.

cgzones (approved):

linux/LinuxProcessTable.c
BenBE: Yes, the `isxdigit` was written by me. The constant stems from masking `0x1F`, after that, you only care about things `0x00` through `0x1F`, which corresponds to `0x30` through `0x4F`. The bits in that constant are basically `0x00..0x0F` -> hex digits `'A'..'F'` and `0x10..0x1F` -> `'0'..'9'`. Hope this explains that constant.

Explorer09: I do understand how the constant works, and it was clever. I try to extend your idea to filter upper bits of a byte, too. `(0x58 & (1L << (ch >> 4)))` Can't say if this would be faster, but I run into some compiler problems that it somehow missed the optimizations that I expected.

BenBE: Mostly noted the construction here for reference, so people less versed in the code can get a feel for what this constant actually represents.

Issue #1404, opened 2024-03-01
Turing RK1: missing temperature values


segabor:
Please add temperature values for Turing RK1 cores.
This is a Rockchip RK3588 based SoC. I managed to retrieve temperatures with `sensors` utility bundled with `lm-sensors` / `libsensors`.

```
npu_thermal-virtual-0
Adapter: Virtual device
temp1:        +35.2 C  

center_thermal-virtual-0
Adapter: Virtual device
temp1:        +35.2 C  

bigcore1_thermal-virtual-0
Adapter: Virtual device
temp1:        +36.1 C  

soc_thermal-virtual-0
Adapter: Virtual device
temp1:        +36.1 C  (crit = +115.0 C)

nvme-pci-0100
Adapter: PCI adapter
Composite:    +40.9 C  (low  =  -0.1 C, high = +71.8 C)
                       (crit = +89.8 C)

gpu_thermal-virtual-0
Adapter: Virtual device
temp1:        +35.2 C  

littlecore_thermal-virtual-0
Adapter: Virtual device
temp1:        +36.1 C  

bigcore0_thermal-virtual-0
Adapter: Virtual device
temp1:        +36.1 C  
```

<img width="1318" alt="Screenshot 2024-03-01 at 16 51 37" src="https://github.com/htop-dev/htop/assets/23614/cdf0806a-b621-43fa-97c3-50a79effd3b3">

PR #1411, opened 2024-03-11
Add core temperatures for Rockchip RK3588 SoC

segabor:
My attempt to map SoC core temperatures to virtual cores. Rather a hack than a well reasoned change. There must be a proper way to determine chipset so temperature mapping can be done if condition is met (a refactor will be needed).
Also, `data[0]` is not set. We have two candidates here, `center_thermal` and `soc_thermal` values.

## SoC Thermal Drivers

`sensors` utility bundled in `lm-sensors` package detects the following thermal drivers

```
npu_thermal-virtual-0
Adapter: Virtual device
temp1:        +35.2 C  

center_thermal-virtual-0
Adapter: Virtual device
temp1:        +35.2 C  

bigcore1_thermal-virtual-0
Adapter: Virtual device
temp1:        +36.1 C  

soc_thermal-virtual-0
Adapter: Virtual device
temp1:        +36.1 C  (crit = +115.0 C)

nvme-pci-0100
Adapter: PCI adapter
Composite:    +41.9 C  (low  =  -0.1 C, high = +71.8 C)
                       (crit = +89.8 C)

gpu_thermal-virtual-0
Adapter: Virtual device
temp1:        +35.2 C  

littlecore_thermal-virtual-0
Adapter: Virtual device
temp1:        +36.1 C  

bigcore0_thermal-virtual-0
Adapter: Virtual device
temp1:        +36.1 C  
```

## Future Additions

- [ ] NPU temperature
- [ ] GPU temperature

## References

- [Rockchip RK3588 Datasheet](https://www.cnx-software.com/pdf/Rockchip%C2%A0RK3588%C2%A0Datasheet%C2%A0V0.1-20210727.pdf)
- [Core temperature mapping in Linux](https://forum.armsom.org/t/topic/51)


segabor:
This PR addresses #1404

BenBE:
I think, instead of hacks on top of other hacks, the overall libsensors-to-cpucore mapping should be addressed properly.

linux/LibSensors.c
```diff
@@ -208,6 +213,33 @@ void LibSensors_getCPUTemperatures(CPUData* cpus, unsigned int existingCPUs, uns
...
+            data[6] = temp;
+            coreTempCount+=2;
+            continue;
+         }
+         if (existingCPUs == 8 && String_eq(chip->prefix, "bigcore1_thermal")) {
+            data[7] = temp;
+            data[8] = temp;
+            coreTempCount+=2;
+            continue;
+         }
```
BenBE, line 241: A better solution here, instead of hard-coding CPU counts would be to determine the cores associated with each of the groups and assigning the values from there.

Also: There's some indentation issue …

Explorer09:
> I think, instead of hacks on top of other hacks, the overall libsensors-to-cpucore mapping should be addressed properly.

I believe the libsensors-to-cpu-core mapping shouldn't be handled in htop's side. Perhaps we can ask this question in libsensors upstream to see if there is a portable way to describe the mappings.

My impression is that [libsensors.conf](https://linux.die.net/man/5/sensors.conf) is supposed to do the job. Or, maybe, ACPI thermal zone (even though I personally hate ACPI).
Once we have a "protocol" to read the mapping from, we would no longer need these chipset-specific hacks.

linux/LibSensors.c
segabor: I'll fix the indentation. You're right about the code, it's just a POC. I like @Explorer09 's idea about syncing with the team behind libsensors.

BenBE:
@Explorer09 Mind to open a discussion with the libsensors people and link it here for reference? It's not just this bug, but quite a few more, that basically boil down to how to perform this mapping. TIA.

Wer-Wolf:
Maybe it would be better to create some sort of naming convention for the temperature channels used for the cpu cores.

I am strongly against adding driver-specific code to libsensors, it was done once and proved to be an maintainance nightmare.

With a proper naming convention in place, sensors.conf can be used to do the necessary adjustments.

BenBE:
We'd also strongly prefer not to deal with any of the driver-specific stuff in htop. And honestly, there are two places things could go: Either into sysfs alongside each sensor providing the core/thread information we need (can figure the rest from procfs -> cpuinfo). Or we get this information from the library we use for fetching the temperature information.

Both have their drawbacks, but the alternative would be, that each user of lmsensors would have to write their own code to figure this out instead, which is effectively worse …

BenBE:
@segabor Please squash your commits in this PR. Cf. the styleguide for details.

BenBE:
The latest rebase looks strange …

segabor:
Sorry for the mess @BenBE , hope it's fixed now

linux/LibSensors.c
```diff
@@ -167,6 +171,7 @@ void LibSensors_getCPUTemperatures(CPUData* cpus, unsigned int existingCPUs, uns
    int n = 0;
    for (const sensors_chip_name* chip = sym_sensors_get_detected_chips(NULL, &n); chip; chip = sym_sensors_get_detected_chips(NULL, &n)) {
       const int priority = tempDriverPriority(chip);
+      // printf("chip prefix: %s, path: %s, prio: %d\n", chip->prefix, chip->path, priority);
```
BenBE, line 174: No debug code

BenBE (changes requested):
The mess seems fixed now …

linux/LibSensors.c
segabor: I somehow forgot to remove that line. Sorry about that. Fixed.

linux/LibSensors.c
```diff
@@ -208,6 +212,33 @@ void LibSensors_getCPUTemperatures(CPUData* cpus, unsigned int existingCPUs, uns
...
+          *   littlecore -> cores 1..4
+          *   bigcore0   -> cores 5,6
+          *   bigcore1   -> cores 7,8
+          */
+         if (existingCPUs == 8 && String_eq(chip->prefix, "littlecore_thermal")) {
+            data[1] = temp;
+            data[2] = temp;
+            data[3] = temp;
+            data[4] = temp;
+            coreTempCount+=4;
```
BenBE, line 226:
```suggestion
            coreTempCount += 4;
```

linux/LibSensors.c
```diff
@@ -208,6 +212,33 @@ void LibSensors_getCPUTemperatures(CPUData* cpus, unsigned int existingCPUs, uns
...
+            continue;
+         }
+         if (existingCPUs == 8 && String_eq(chip->prefix, "bigcore0_thermal")) {
+            data[5] = temp;
+            data[6] = temp;
+            coreTempCount+=2;
```
BenBE, line 232:
```suggestion
            coreTempCount += 2;
```

linux/LibSensors.c
```diff
@@ -208,6 +212,33 @@ void LibSensors_getCPUTemperatures(CPUData* cpus, unsigned int existingCPUs, uns
...
+            continue;
+         }
+         if (existingCPUs == 8 && String_eq(chip->prefix, "bigcore1_thermal")) {
+            data[7] = temp;
+            data[8] = temp;
+            coreTempCount+=2;
```
BenBE, line 238:
```suggestion
            coreTempCount += 2;
```

linux/LibSensors.c
```diff
@@ -208,6 +212,33 @@ void LibSensors_getCPUTemperatures(CPUData* cpus, unsigned int existingCPUs, uns
          if (r != 0)
             continue;
 
+         /* Map temperature values to Rockchip cores
+          *
+          *   littlecore -> cores 1..4
+          *   bigcore0   -> cores 5,6
+          *   bigcore1   -> cores 7,8
+          */
+         if (existingCPUs == 8 && String_eq(chip->prefix, "littlecore_thermal")) {
```
BenBE, line 221: If the first condition is checking for 8 CPUs anyway, why not pull this as an outer layer. That way this automatically groups the three blocks belonging to the Rockchip cores as a side effect.

BenBE (commented):
Minor code style stuff and a small suggestion.

linux/LibSensors.c
```diff
@@ -208,6 +212,33 @@ void LibSensors_getCPUTemperatures(CPUData* cpus, unsigned int existingCPUs, uns
...
+            data[6] = temp;
+            coreTempCount += 2;
+            continue;
+         }
+         if (existingCPUs == 8 && String_eq(chip->prefix, "bigcore1_thermal")) {
+            data[7] = temp;
+            data[8] = temp;
+            coreTempCount += 2;
+            continue;
+         }
```
BenBE, line 240:
```suggestion
         if (existingCPUs == 8) {
            if (String_eq(chip->prefix, "littlecore_thermal")) {
               data[1] = temp;
               data[2] = temp;
               data[3] = temp;
               data[4] = temp;
               coreTempCount += 4;
               continue;
            }
            if (String_eq(chip->prefix, "bigcore0_thermal")) {
               data[5] = temp;
               data[6] = temp;
               coreTempCount += 2;
               continue;
            }
            if (String_eq(chip->prefix, "bigcore1_thermal")) {
               data[7] = temp;
               data[8] = temp;
               coreTempCount += 2;
               continue;
            }
         }
```

isarrider:
@segabor highly appreacted as I have many rk3588...

linux/LibSensors.c
segabor: Done.

BenBE: I fixed the remaining whitespace issue from your latest revision.

BenBE (approved):

isarrider:
wohoo - so happy that my thank you triggered a merge ;)

farlongsignal:
thank you for this, i easily adapted this for Qualcomm SDM845 (Oneplus 6 enchilada running PostmarketOS with mainline kernel), it reports temperatures in similar way.
Although i could imagine this getting really redundant if you wanted to support all the other possible Qualcomm chips.
[Here](https://github.com/farlongsignal/htop) in case someone wants it.

BenBE:
> thank you for this, i easily adapted this for Qualcomm SDM845 (Oneplus 6 enchilada running PostmarketOS with mainline kernel), it reports temperatures in similar way. Although i could imagine this getting really redundant if you wanted to support all the other possible Qualcomm chips. [Here](https://github.com/farlongsignal/htop) in case someone wants it.

Could you please link the exact commit for future reference? That way you allow people even when your repo at some point contains more commits to easily find the exact change you are referring to in your comment. TIA.

Issue #1410, opened 2024-03-09
Allow disabling the sorting timeout


afdw:
Currently, sorting only takes effect after some time has passed: https://github.com/htop-dev/htop/blob/4feac8e72f76c7d50ac4e61689645796cd5fdc31/ScreenManager.c#L240-L241

Also, the timeout is reset every time an unknown key is pressed: https://github.com/htop-dev/htop/blob/4feac8e72f76c7d50ac4e61689645796cd5fdc31/ScreenManager.c#L420

Meaning, if an unbound key such as `o` is pressed, sorting is paused for a few seconds. For me, this also happens when the window of a terminal emulator where htop is running gains of looses focus. So I have to wait that time after switching to the htop window to see an up-to-date information about what is consuming CPU, or find another way to trigger sorting to start updating, such as reversing the sorting direction twice or pressing `Z` twice.

As far as I understand, the purpose of the sorting timeout is to give the user time to read the process list before it starts flashing and changing order. But I do not see why the timeout is always reset in cases described above.

I find all that a bit confusing and I would prefer the sorting to be always consistently applied without any timeouts. Would it be possible to add an option to complete disable the sorting timeout? Would a simple on/off option suffice?

Explorer09:
"Disabling" the sorting timeout is not an easy task as it requires adding a real-time algorithm to insert and sort the entries at the same time. (This is computer science knowledge.)

Is there any real reason you want this real-time update and sorting? I thought a 500 ms sorting period should be enough for humans to not observe any sorting delay.

afdw:
> "Disabling" the sorting timeout is not an easy task as it requires adding a real-time algorithm to insert and sort the entries at the same time. (This is computer science knowledge.)

It seems that I did not guess correctly the purpose of the timeout. Can you explain it? Why would it be an issue for `sortTimeout` to always be `0`, what would that break? And why a special algorithm would be needed?

> Is there any real reason you want this real-time update and sorting? I thought a 500 ms sorting period should be enough for humans to not observe any sorting delay.

For me currently it is more than 3 seconds. But even then, I do think 0.5 seconds would be noticeable: this is enough time for a human to look at the first row in the table and get to a wrong conclusion about which process consumes the most resources, if the table is not properly updated.

And again, why is there need to reset this timeout every time an unknown key is pressed or the htop window changes focus?

Explorer09:
[...]

PR #1413, opened 2024-03-13
Ignore FOCUS_IN and FOCUS_OUT events if these have been activated within xterm

fasterit:
Closes: #1410

### DIFF
--- a/ScreenManager.c
+++ b/ScreenManager.c
@@ -363,6 +363,9 @@ void ScreenManager_run(ScreenManager* this, Panel** lastFocus, int* lastKey, con
          ScreenManager_resize(this);
          continue;
       }
+      case KEY_FOCUS_IN:
+      case KEY_FOCUS_OUT:
+         break;
       case KEY_LEFT:
       case KEY_CTRL('B'):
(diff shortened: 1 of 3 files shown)


ScreenManager.c
```diff
@@ -363,6 +363,11 @@ void ScreenManager_run(ScreenManager* this, Panel** lastFocus, int* lastKey, con
          ScreenManager_resize(this);
          continue;
       }
+      #ifdef KEY_FOCUS_IN
```
BenBE, line 366: Why have this in `#ifdef` here when it's without conditional checks at the other location where it's used?

fasterit: That's a leftover from making it more easy to test. We are quite inconsistent in what we have #ifdef'd in the keyboard / mouse handling and what not. But that's for another time. I'm happy to remove this #ifdef before merge.
Would be interesting if checking for these "keys" has any collateral damage in non-xterm use case. Unfortunately we typically only get wider testing when releases happen so ... YOLO.

fasterit: removed the #ifdef

BenBE (approved):

PR #1414, opened 2024-03-15
ci: bump to LLVM 18 and drop workaround

cgzones:
The ASAN issue should be fixed with LLVM 17: https://github.com/llvm/llvm-project/commit/fb77ca05ffb4f8e666878f2f6718a9fb4d686839


BenBE (commented):
Given this runs atop a Ubuntu 22.04, shouldn't we also use the jammy sources for the repos?

cgzones:
> Given this runs atop a Ubuntu 22.04, shouldn't we also use the jammy sources for the repos?

True; added.

Issue #1420, opened 2024-03-20
RISC-V crash on launch


jdek:
## Info

Not sure if this is my environment or htop itself, but submitting it anyway since maybe it'll be useful.

**Device**: CanMV K230 RISC-V 64bit SBC
**uname**: `Linux riscv 5.10.4+ #1 SMP Fri Mar 1 21:48:00 CET 2024 riscv64 GNU/Linux`
**os**: Debian Unstable

Reproduce: `htop` -> crashes on launch at a high rate.
[htop.objdump.tar.gz](https://github.com/htop-dev/htop/files/14665017/htop.objdump.tar.gz)

### Crash: 
```
Error information:
------------------
A signal 6 (Aborted) was received.

Setting information:
--------------------
[...]

PR #1421, opened 2024-03-20
Fix reachable assert for CPU percentage of NAN

cgzones:
* Avoid too large CPU field widths

  In case the current CPU usage percentage is NAN the width calculation is invalid.  Use the default with of 4, which is appropriate for "N/A ".

* Check width in Row_printPercentage()
  The passed width should always be at least 4, otherwise printing will always truncate and lead to an abort().

  The passed should not be greater or equal to the available buffer size, otherwise printing will always truncate and lead to an abort().

  Add fallback for non debug builds.

Fixes: #1420

### DIFF
--- a/Row.c
+++ b/Row.c
@@ -442,6 +442,10 @@ void Row_printLeftAlignedField(RichString* str, int attr, const char* content, u
 }
 
 int Row_printPercentage(float val, char* buffer, size_t n, uint8_t width, int* attr) {
+   assert(width >= 4 && width < n && "Invalid width in Row_printPercentage()");
+   // truncate in favour of abort in xSnprintf()
+   width = (uint8_t)CLAMP(width, 4, n - 1);
+
    if (isNonnegative(val)) {
       if (val < 0.05F)
          *attr = CRT_colors[PROCESS_SHADOW];
--- a/Process.c
+++ b/Process.c
@@ -1071,7 +1071,7 @@ void Process_updateExe(Process* this, const char* exe) {
 }
 
 void Process_updateCPUFieldWidths(float percentage) {
-   if (percentage < 99.9F) {
+   if (percentage < 99.9F || isNaN(percentage)) {
       Row_updateFieldWidth(PERCENT_CPU, 4);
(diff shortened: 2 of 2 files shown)


Row.c
```diff
@@ -442,6 +442,11 @@ void Row_printLeftAlignedField(RichString* str, int attr, const char* content, u
 }
 
 int Row_printPercentage(float val, char* buffer, size_t n, uint8_t width, int* attr) {
+   assert(width >= 4 && width < n && "Invalid width in Row_printPercentage()");
+   // truncate in favour of abort in xSnprintf()
+   if (width >= n)
+      width = (uint8_t)n - 1;
```
BenBE, line 448: Prefer `CLAMP` to avoid integer truncation.

BenBE (changes requested):

BenBE (approved):

Row.c
```diff
@@ -442,6 +442,10 @@ void Row_printLeftAlignedField(RichString* str, int attr, const char* content, u
 }
 
 int Row_printPercentage(float val, char* buffer, size_t n, uint8_t width, int* attr) {
+   assert(width >= 4 && width < n && "Invalid width in Row_printPercentage()");
+   // truncate in favour of abort in xSnprintf()
+   width = (uint8_t)CLAMP(width, 4, n - 1);
```
Explorer09, line 447: @cgzones The maximum width should be `n - 2`. Because the format string of `xSnprintf` below has a trailing space.
Or alternatively just use `snprintf` instead of `xSnprintf`?

BenBE: I'd prefer `n-2` for the fix here. The `xSnprintf` is there to catch such issues.

Explorer09: And what if `n` is too small? `n - 2 < 4` for the `CLAMP` macro

BenBE: I noticed there to be an assert (`n > width`) after that clamping to be still missing. Will push some additional fix for that part later; hopefully resolving this entirely.

PR #1424, opened 2024-03-27
Simplify string formatting

cgzones:
Just call RichString_appendAscii() with the fixed string and avoid an superfluous snprintf() formatting.

Suggested-by: Explorer09

/cc @Explorer09


natoscott (approved):
LGTM @Explorer09 @cgzones

PR #1425, opened 2024-03-27
New "nanoseconds" format for GPU time fields.

Explorer09:
This is a new time format that I just came up with, because I just don't want to see extra time precision gets truncated.

The new formats are:
```text
     0ns - 999999ns
.001000s - .999999s (1 ms or above)
1.00000s - 9.99999s (1 second or above)
10.0000s - 59.9999s (10 seconds or above)
1:00.000 - 9.59.999 (1 minute or above)
10:00.00 - 59.59.99 (uses Row_printTime() to avoid duplicate code)
```

### DIFF
--- a/Row.c
+++ b/Row.c
@@ -403,6 +403,57 @@ void Row_printTime(RichString* str, unsigned long long totalHundredths, bool col
    }
 }
 
+void Row_printNanoseconds(RichString* str, unsigned long long totalNanoseconds, bool coloring) {
+   if (totalNanoseconds == 0) {
+      int shadowColor = coloring ? CRT_colors[PROCESS_SHADOW] : CRT_colors[PROCESS];
+
+      RichString_appendAscii(str, shadowColor, "     0ns ");
+      return;
+   }
+
+   char buffer[10];
+   int len;
+   int baseColor = CRT_colors[PROCESS];
+
+   if (totalNanoseconds < 1000000) {
+      len = xSnprintf(buffer, sizeof(buffer), "%6luns ", (unsigned long)totalNanoseconds);
+      RichString_appendnAscii(str, baseColor, buffer, len);
+      return;
+   }
+
+   unsigned long long totalMicroseconds = totalNanoseconds / 1000;
+   if (totalMicroseconds < 1000000) {
+      len = xSnprintf(buffer, sizeof(buffer), ".%06lus ", (unsigned long)totalMicroseconds);
+      RichString_appendnAscii(str, baseColor, buffer, len);
+   }
+
+   unsigned long long totalSeconds = totalMicroseconds / 1000000;
+   unsigned long microseconds = totalMicroseconds % 1000000;
+   if (totalSeconds < 60) {
+      int width = 5;
+      unsigned long fraction = microseconds;
+      if (totalSeconds >= 10) {
+         width--;
+         fraction /= 10;
+      }
+      len = xSnprintf(buffer, sizeof(buffer), "%u.%0*lus ", (unsigned int)totalSeconds, width, fraction);
+      RichString_appendnAscii(str, baseColor, buffer, len);
+      return;
+   }
+
+   if (totalSeconds < 600) {
+      unsigned int minutes = totalSeconds / 60;
+      unsigned int seconds = totalSeconds % 60;
+      unsigned int milliseconds = microseconds / 1000;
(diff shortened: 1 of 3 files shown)


Explorer09:
Now I am thinking: Does the jump between `"9.999999"` and `" 0:10.00"` seem too large and is it good to add another format in between, that is `"10.00000"` - `"59.99999"`?

Row.c
```diff
@@ -404,6 +404,37 @@ void Row_printTime(RichString* str, unsigned long long totalHundredths, bool col
...
+      len = xSnprintf(buffer, sizeof(buffer), "%6luns ", (unsigned long)totalNanoseconds);
+      RichString_appendnAscii(str, baseColor, buffer, len);
+      return;
+   }
+
+   unsigned long long totalMicroseconds = totalNanoseconds / 1000;
+   if (totalMicroseconds < 10000000) {
+      unsigned int seconds = totalMicroseconds / 1000000;
+      unsigned long microseconds = totalMicroseconds % 1000000;
+      len = xSnprintf(buffer, sizeof(buffer), "%u.%06lu ", seconds, microseconds);
```
cgzones, line 429: Maybe add an additional case for micro seconds to avoid outputs like `0.000042` and also avoid output without a unit name.

Explorer09: I chose this format without a unit name because it helps transitioning to the minute unit and above. The existing Row_printTime() would be used for those units.

As for microseconds, the format like "`0.999999`" already has microsecond precision; thus displaying like "`999999us`" would be redundant.

There is one exceptional case, "`1000.0us`" to "`9999.9us`", where the "us" suffix could be useful and it shows one more digit of precision than "`0.009999`".

What's your advice then? Should I add "`9999.9us`" into it?

Explorer09: P.S.
This is the current behavior:
```text
     0ns - 999999ns (0ns – 999999ns)
0.001000 - 0.009999 (1ms – 9.9999ms or 1000μs – 9999μs)
0.010000 - 9.999999 (10ms – 9.999999s)
 0:10.00 -  0:59.99
 1:00.00 - 59:59.99
 1h00:00 - 23h59:59
```
This is the format I can change to with the most precision possible for the 8 character limit:
```text
     0ns - 999999ns (0ns – 999999ns)
1000.0us - 9999.9us (1ms – 9.9999ms or 1000μs – 9999.9μs)
0.010000 - 9.999999 (10ms – 9.999999s)
10.00000 - 59.99999 (seconds)
 1:00.00 - 59:59.99
 1h00:00 - 23h59:59
```

BenBE: The jump from us to s looks kinda harsh to me (first glance). Apart from that I prefer the unit to be always visible for clarity.

Explorer09: @BenBE I would rather not have a "us" unit shown at all, if the "seconds" unit with fractions is enough for the microseconds.

By the way. There are two other format ideas that I haven't tried:

```text
.010000s - .999999s (no leading zero before decimal point; shows same precision as "us")
0s010000 - 0s999999 (the RKM-like notation; looks ugly IMHO)
```

If people are okay with the `.999999s` format, maybe we don't need the "μs" format at all. Just "ns" and then jumping to "seconds with fractions".

BenBE: Of those two `.123456` looks best. I don't like the `0s123456` variant either …

Explorer09:
Updated with the new formats:
```text
     0ns - 999999ns
.001000s - .999999s (1 ms or above)
1.00000s - 9.99999s (1 second or above)
10.0000s - 59.9999s (10 seconds or above)
1:00.000 - 9.59.999 (1 minute or above)
10:00.00 - 59.59.99 (uses Row_printTime() to avoid duplicate code)
```

Row.c
```diff
@@ -403,6 +403,57 @@ void Row_printTime(RichString* str, unsigned long long totalHundredths, bool col
...
+   }
+
+   unsigned long long totalSeconds = totalMicroseconds / 1000000;
+   unsigned long microseconds = totalMicroseconds % 1000000;
+   if (totalSeconds < 60) {
+      int width = 5;
+      unsigned long fraction = microseconds;
+      if (totalSeconds >= 10) {
+         width--;
+	 fraction /= 10;
```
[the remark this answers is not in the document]
Explorer09: @cgzones It's an accidental tab character left within.

PR #1427, opened 2024-03-30
linux: use dlopen for libnl3 instead of dynamic linking

cgzones:
Instead of the current behavior of dynamic linking against *libnl3* and *libnl-genl-3* when configured with `--enable-delayacct`, load the shared libraries on request, if any delay accounting related process field is active, via `dlopen(3)`, similar to *libsensors* and *libsystemd*.

Distribution, who currently build htop with `--enable-delayacct`, need to explicitly add *libnl3* and *libnl-genl-3* as runtime dependencies to continue supporting delay accounting out-of-the-box.

Tested on Debian sid and Fedora 39.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -692,6 +692,7 @@ case "$enable_delayacct" in
       AC_MSG_ERROR([bad value '$enable_delayacct' for --enable-delayacct])
       ;;
 esac
+AM_CONDITIONAL([HAVE_DELAYACCT], [test "$enable_delayacct" = yes])
 
 
 AC_ARG_ENABLE([sensors],
--- a/configure.ac
+++ b/configure.ac
@@ -658,40 +658,26 @@ case "$enable_delayacct" in
       elif test "$enable_static" = yes; then
          enable_delayacct=no
       else
-         m4_ifdef([PKG_PROG_PKG_CONFIG], [
-            enable_delayacct=yes
-            PKG_PROG_PKG_CONFIG()
-            PKG_CHECK_MODULES(LIBNL3, libnl-3.0, [], [enable_delayacct=no])
-            PKG_CHECK_MODULES(LIBNL3GENL, libnl-genl-3.0, [], [enable_delayacct=no])
-            if test "$enable_delayacct" = yes; then
-               AM_CFLAGS="$AM_CFLAGS $LIBNL3_CFLAGS $LIBNL3GENL_CFLAGS"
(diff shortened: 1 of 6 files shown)


linux/LibNl.c
```diff
@@ -0,0 +1,220 @@
...
+
+   resolve(libnlGenlHandle, genl_ctrl_resolve);
+   resolve(libnlGenlHandle, genlmsg_parse);
+   resolve(libnlGenlHandle, genlmsg_put);
+
+   #undef resolve
+
+   return 0;
+
+dlfailure:
```
BenBE, line 100: I wonder if we should set all pointers to `NULL` on failure (and when free'ing the lib..

BenBE (approved):
LGTM.

configure.ac
```diff
@@ -658,40 +658,27 @@ case "$enable_delayacct" in
...
+      AC_CHECK_HEADERS([netlink/attr.h netlink/handlers.h netlink/msg.h], [], [AC_MSG_ERROR([can not find required header files netlink/attr.h, netlink/handlers.h, netlink/msg.h])])
+      CFLAGS="$old_CFLAGS"
       ;;
    *)
       AC_MSG_ERROR([bad value '$enable_delayacct' for --enable-delayacct])
       ;;
 esac
+if test "$enable_delayacct" = yes; then
+  AC_DEFINE([HAVE_DELAYACCT], [1], [Define if delay accounting support should be enabled.])
+  AM_CFLAGS="$AM_CFLAGS -I/usr/include/libnl3"
```
fpletz, line 679: Why was discovery via pkg-config removed? Some distributions like NixOS are not shipping headers and libs in `/usr`.

fasterit: Because libnl-3 is not linked into the htop executable any more but assessed to be available or not at runtime. I'm sure NixOS packagers can make the library available at build time and symlink the headers to the default location.

SuperSandro2000: Not all OS's have libraries even at that location even if installed. NixOS cannot symlink any libraries to that either. We now patch things to make things work. 

This change was kinda short sighted.

BenBE: > I'm happy to open a PR to fix this but only if there is a chance that it will be accepted. 🙂

Go ahead. Any proper PR will at least be considered, and given there's a valid use case (building against user-compiled libraries), there is a good chance for it to go through.

Anyways, this discussion should have been a separate issue, instead of a code review discussion on an already merged PR …

Issue #1426, opened 2024-03-30
HTOPRC handling too aggressive?


ossimoi:
Wanted to run htop with default settings. `HTOPRC=/dev/null htop` as root was a bad idea since htop replaced /dev/null char device with default config. This obviously caused problems. Should there be an extra check to not touch the file if it exists but does not contain actual config?

BenBE:
At least things should not be touched, if the original file is not a regular file (or symlink).

Explorer09:
Surely is was a bad idea. But keep in mind the root user is already able to break the system with `HTOPRC=/etc/passwd htop` (/etc/passwd is surely a regular file).

We've handled the symlinks already in #1412. Maybe we can add exception that any file type that's not regular file should not be touched. This would skip not just device files but also directories, FIFOs, and sockets.

PR #1428, opened 2024-03-30
Avoid saving configuration to non regular files

cgzones:
Check on startup whether the resolved configuration filename points to a
regular file, and only then write the settings on exit.

Fixes: #1426

### DIFF
--- a/Settings.c
+++ b/Settings.c
@@ -622,6 +622,8 @@ int Settings_write(const Settings* this, bool onCrash) {
    if (onCrash) {
       fd = stderr;
       separator = ';';
+   } else if (!this->writeConfig) {
+      return 0;
    } else {
       /* create tempfile with mode 0600 */
       mode_t cur_umask = umask(S_IXUSR | S_IRWXG | S_IRWXO);
@@ -753,6 +755,8 @@ int Settings_write(const Settings* this, bool onCrash) {
 Settings* Settings_new(unsigned int initialCpuCount, Hashtable* dynamicMeters, Hashtable* dynamicColumns, Hashtable* dynamicScreens) {
    Settings* this = xCalloc(1, sizeof(Settings));
 
+   this->writeConfig = true;
+
    this->dynamicScreens = dynamicScreens;
    this->dynamicColumns = dynamicColumns;
    this->dynamicMeters = dynamicMeters;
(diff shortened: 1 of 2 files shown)


Explorer09:
My advice is to check the file type on read, not on write.

After resolving symlinks, check the file type of the resolved path to see if It's a regular file. If it's not a regular file (this case might include a newly created symlink after resolving symlinks, remember TOUTOU is possible) start htop with no write to the configuration at all.

Don't bother to check the file type on write. No matter how we check at that time, TOCTOU is possible.

cgzones:
Changed to check on startup rather than exit.

Settings.c
```diff
@@ -832,6 +836,14 @@ Settings* Settings_new(unsigned int initialCpuCount, Hashtable* dynamicMeters, H
    if (!realpath(this->initialFilename, this->filename))
       free_and_xStrdup(&this->filename, this->initialFilename);
 
+   /* check whether filename points to a regular file */
+   {
+      struct stat sb;
+      int err = lstat(this->filename, &sb);
+      if ((err == -1 && errno != ENOENT) || (err == 0 && !S_ISREG(sb.st_mode)))
```
Explorer09, line 843: I try to align with the code style a few lines above.
```suggestion
      if (err && !S_ISREG(sb.st_mode) || (!err && errno != ENOENT))
```

Explorer09:
While we have a `writeConfig` flag to tell htop whether to write the configuration, there is one check I wish to add into the code as well.

It's the (obvious) check of whether the file has a write permission. I guess you know the reason why -- user might remove the write permission on the file to tell htop not to write to it.

The code is better checked in the `fopen()` part of the `Settings_read()` function. Something like this (the code is untested):

```c
static bool Settings_read(Settings* this, const char* fileName, unsigned int initialCpuCount) {
   FILE* fd = fopen(fileName, "r+");
   if (!fd) {
      this->writeConfig = false;
      FILE* fd = fopen(fileName, "r");
      if (!fd) {
         return false;
      }
   }
   ...
}
```

Settings.c
cgzones: I think you mixed up the two OR operands.
Also we need to handle ENOENT differently here.

Explorer09: Sorry. Probably this would work:
```suggestion
      if ((err && errno != ENOENT) || (!err && !S_ISREG(sb.st_mode)))
```

cgzones:
Added a writable check.
It cannot be added in `Settings_read()` though, because it is called with different file names, e.g. */etc/htoprc*.

Explorer09:
> Added a writable check. It cannot be added in `Settings_read()` though, because it is called with different file names, e.g. _/etc/htoprc_.

The `access()` API is what I _wished_ to avoid because there is a potential of TOCTOU. I've read the `Settings_read()` code somehow and think the function can add an argument "`bool checkWritability`" to achieve what I want.

* `checkWritability` is true -> try [open(2)](https://man7.org/linux/man-pages/man2/open.2.html) with write permission (maybe `O_RDWR | O_NOFOLLOW`), then check whether it's a regular file with [fstat(2)](https://man7.org/linux/man-pages/man2/fstat.2.html)
* `checkWritability` is false -> open(2) in read only (`O_RDONLY`), and maybe allow following symlinks (because the path won't be the final location where we write the configuration on exit.)

Explorer09:
@cgzones Please check this one out. This is my version of the patch:

https://github.com/Explorer09/htop-1/tree/htoprc-write-check

cgzones:
@Explorer09 seem good (two comments), feel free to propose

cgzones:
Closing in favor of #1437

PR #1429, opened 2024-04-02
Use designated initializer more

cgzones:
Designated initializers are supported in C99 and simplify struct initializations.  All members not explicitly mentioned are default initialized to 0, and also modern compilers can warn on of those missing ones.

### DIFF
--- a/Hashtable.c
+++ b/Hashtable.c
@@ -109,13 +109,15 @@ static size_t nextPrime(size_t n) {
 }
 
 Hashtable* Hashtable_new(size_t size, bool owner) {
-   Hashtable* this;
+   size = size ? nextPrime(size) : 13;
 
-   this = xMalloc(sizeof(Hashtable));
-   this->items = 0;
-   this->size = size ? nextPrime(size) : 13;
-   this->buckets = (HashtableItem*) xCalloc(this->size, sizeof(HashtableItem));
-   this->owner = owner;
+   Hashtable* this = xMalloc(sizeof(Hashtable));
+   *this = (Hashtable) {
+      .items = 0,
+      .size = size,
+      .buckets = xCalloc(size, sizeof(HashtableItem)),
+      .owner = owner,
+   };
 
    assert(Hashtable_isConsistent(this));
    return this;
--- a/Settings.c
+++ b/Settings.c
@@ -790,7 +790,7 @@ Settings* Settings_new(unsigned int initialCpuCount, Hashtable* dynamicMeters, H
    this->topologyAffinity = false;
    #endif
 
-   this->screens = xCalloc(Platform_numberOfDefaultScreens * sizeof(ScreenSettings*), 1);
+   this->screens = xCalloc(Platform_numberOfDefaultScreens, sizeof(ScreenSettings*));
    this->nScreens = 0;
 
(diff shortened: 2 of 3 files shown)


Explorer09:
For the `Settings` object, it seems that not every member is initialized when you declare a struct initializer there. This could hurt readability and may increase code size for _not_ using `xCalloc` to zero-initialize other struct members.

For structures where you initialize all members at once, I see no disadvantage of using a compound literal there. But for objects whose members are not all initialized at once, I see no advantage of using the syntax like this.

Vector.c
```diff
@@ -25,14 +25,16 @@ Vector* Vector_new(const ObjectClass* type, bool owner, int size) {
...
-   this->array = (Object**) xCalloc(size, sizeof(Object*));
-   this->arraySize = size;
-   this->items = 0;
-   this->type = type;
-   this->owner = owner;
-   this->dirty_index = -1;
-   this->dirty_count = 0;
+   *this = (Vector) {
+      .growthRate = size,
+      .array = (Object**) xCalloc(size, sizeof(Object*)),
```
BenBE, line 30: This cast strictly necessary?

BenBE (approved):
LGTM. Mostly some code style stuff to consider …

Hashtable.c
```diff
@@ -109,13 +109,15 @@ static size_t nextPrime(size_t n) {
...
-   this->size = size ? nextPrime(size) : 13;
-   this->buckets = (HashtableItem*) xCalloc(this->size, sizeof(HashtableItem));
-   this->owner = owner;
+   size = size ? nextPrime(size) : 13;
+
+   Hashtable* this = xMalloc(sizeof(Hashtable));
+   *this = (Hashtable) {
+      .items = 0,
+      .size = size,
+      .buckets = (HashtableItem*) xCalloc(size, sizeof(HashtableItem)),
```
Explorer09, line 118: I personally don't like this writing style, especially for object initialization code.

It might look ok for htop codebase (since htop doesn't check for allocation failures and just crashes if that happens), but for other projects, it is a better practice to allocate member buffers first, before the main object buffer. That way it would be easier to rewind the process (freeing the members) in case of a failure.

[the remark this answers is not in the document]
cgzones: dropped

Vector.c
cgzones: dropped

Settings.c
```diff
@@ -751,47 +751,55 @@ int Settings_write(const Settings* this, bool onCrash) {
...
+      .highlightDelaySecs = DEFAULT_HIGHLIGHT_SECS,
+      .findCommInCmdline = true,
+      .stripExeFromCmdline = true,
+      .showMergedCommand = false,
+      .hideFunctionBar = 0,
+      .headerMargin = true,
+      #ifdef HAVE_LIBHWLOC
+      .topologyAffinity = false,
+      #endif
+      .screens = xCalloc(Platform_numberOfDefaultScreens * sizeof(ScreenSettings*), 1),
```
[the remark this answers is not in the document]
cgzones: true; fixed

cgzones:
Dropped the Settings struct changes.

BenBE (approved):
LGTM.

PR #1431, opened 2024-04-02
Check for absolute paths in environment variables

cgzones:
Only use the environment variables HOME and XDG_CONFIG_HOME, or the home directory from getpwuid(3) if they are absolute paths. Avoid different behavior depending on the current working directory.

### DIFF
--- a/Settings.c
+++ b/Settings.c
@@ -799,14 +799,14 @@ Settings* Settings_new(unsigned int initialCpuCount, Hashtable* dynamicMeters, H
       this->initialFilename = xStrdup(rcfile);
    } else {
       const char* home = getenv("HOME");
-      if (!home) {
+      if (!home || home[0] != '/') {
          const struct passwd* pw = getpwuid(getuid());
-         home = pw ? pw->pw_dir : "";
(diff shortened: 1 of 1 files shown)


BenBE (approved):
Note, that this tries to store as `/.config/htop/htoprc`, i.e. i the filesystem root, if something goes wrong, instead of running without config (which might be the better option in that case).

Explorer09:
> Note, that this tries to store as `/.config/htop/htoprc`, i.e. i the filesystem root, if something goes wrong, instead of running without config (which might be the better option in that case).

Speaking of the root directory, our config file path is concatenated in the naive manner, if the user sets the `HOME` directory to `/` (i.e. the root directory), then the config file path would have two leading slashes. Not sure if this would cause other problems when opening or saving the configuration file.

BenBE:
Duplicate slashes are canonized by the kernel when parsing the path. Thus no issue on that side. But you likely don't want to allow path traversal and other artifacts their either …

cgzones:
I think protecting against path traversal is out of scope, since there is no security boundary. Also users can just use `HTOPRC=/etc/shadow` instead of `HTOPRC=/home/alice/../../etc/shadow`.

PR #1432, opened 2024-04-02
Read htoprc in modern location first before old one

Explorer09:
Read htop configuration file in the modern location first ("~/.config/htop/htoprc") before trying the legacy location ("~/.htoprc"). This would prevent a case where configuration files exist in both locations and the new configuration gets replaced by the old one.


BenBE (approved):
LGTM.

cgzones (approved):

PR #1435, opened 2024-04-06
Avoid fprintf in signal handlers

cgzones:
fprintf(3) is not safe to call ins signal handlers due to file buffering and memory allocations.
Format messages via snprintf(3) and output them via write(2).


CRT.c
```diff
@@ -1240,74 +1247,79 @@ void CRT_handleSIGSEGV(int signal) {
...
    );
+   full_write_str(STDERR_FILENO, err_buf);
 
-   fprintf(stderr,
+   full_write_str(STDERR_FILENO,
       "Setting information:\n"
       "--------------------\n");
    Settings_write(CRT_crashSettings, true);
-   fprintf(stderr, "\n\n");
+    full_write_str(STDERR_FILENO, "\n\n");
```
BenBE, line 1277: Indentation.

CRT.c
```diff
@@ -1240,74 +1247,79 @@ void CRT_handleSIGSEGV(int signal) {
...
 
 #ifdef PRINT_BACKTRACE
-   fprintf(stderr,
+    full_write_str(STDERR_FILENO,
```
BenBE, line 1280: Indentation.

CRT.c
```diff
@@ -1185,6 +1187,8 @@ static void print_backtrace(void) {
 
    unsigned int item = 0;
 
+   char err_buf[512];
```
BenBE, line 1190: Given symbol and file names may get quite large, consider more space here …

BenBE (commented):
LGTM.

CRT.c
cgzones: fixed

CRT.c
cgzones: Doubled the buffer

cgzones:
Adopted `Settings_write()` accordingly.

BenBE:
Please split the changes for `Settings.c` into its own commit.

Also, use a typedef for callback prototypes.

Apart from that: 'm not sure if all write operations succeed with only 512 bytes; in particular thinking about the Meters / columns entries.

Settings.c
```diff
@@ -563,65 +563,88 @@ static bool Settings_read(Settings* this, const char* fileName, unsigned int ini
...
+      return n;
+
+   return full_write_str(fileno(stream), buf);
 }
 
 int Settings_write(const Settings* this, bool onCrash) {
    FILE* fd;
    char separator;
    char* tmpFilename = NULL;
+   OutputFunc of;
```
BenBE, line 643: Why not have it initialized?

```suggestion
   OutputFunc of = fprintf;
```

And skip it in the else branch …

BenBE (approved):

PR #1436, opened 2024-04-06
Misc Clang tidy hints

cgzones:


BenBE:
What's the reasoning behind the uppercase type indicator for `float` on those literals (i.e. `0.0F` vs `0.0f`)?

Rest looks good to me.

cgzones:
> What's the reasoning behind the uppercase type indicator for `float` on those literals (i.e. `0.0F` vs `0.0f`)?
> 
> Rest looks good to me.

https://clang.llvm.org/extra/clang-tidy/checks/readability/uppercase-literal-suffix.html, probably extended from https://wiki.sei.cmu.edu/confluence/pages/viewpage.action?pageId=87152241.

Also we use capitalized F already.

PR #1437, opened 2024-04-06
Avoid writing configuration to non-regular file

Explorer09:
(This pull request supersedes #1428)

Check whether the configuration file (after symlink resolution) is writable and is a regular file on htop startup, and only then, write the configuration on exit.

Thanks to @cgzones for providing the initial version of the patch.

Fixes: #1426

### DIFF
--- a/Settings.c
+++ b/Settings.c
@@ -11,6 +11,7 @@ in the source distribution for its full text.
 
 #include <ctype.h>
 #include <errno.h>
+#include <fcntl.h>
 #include <limits.h>
 #include <pwd.h>
 #include <stdio.h>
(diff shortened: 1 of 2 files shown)


Settings.c
```diff
@@ -838,10 +877,12 @@ Settings* Settings_new(unsigned int initialCpuCount, Hashtable* dynamicMeters, H
 #endif
    this->changed = false;
    this->delay = DEFAULT_DELAY;
-   bool ok = Settings_read(this, this->filename, initialCpuCount);
+   bool checkWritability = true;
```
cgzones, line 880: What's the purpose of this variable?
Why not call `Settings_read()` with `true`, `false` and `false`?

cgzones (approved):

Settings.c
Explorer09: It's for code readability. Because generally functions with boolean parameters are a bad idea.
But for `Settings_read()` we unfortunately have to add one and it just helps code readability if we use a named boolean variable rather than pass `true`/`false` directly to the functions.

BenBE: `/*param*/true` is always an option …

Explorer09: Not a fan of the "`/*param*/ true`" style, but it wouldn't hurt, I guess.

Explorer09:
Now I changed one behavior with this patch. Previously htop would ignore the legacy configuration file (`~/.htoprc`) if it's a symlink. Now I permit following the link and read from it. However, it won't write to the new location (`~/.config/htop/htoprc`) when the legacy is read only or not a regular file.

PR #1439, opened 2024-04-08
darwin: Fix memory metrics support for Apple Silicon (ARM64)

SuCicada:
Fix memory metrics of Apple Silicon (ARM64)
Detailed description: https://github.com/htop-dev/htop/issues/631

The calculation of memory refers to [exelban/stats Modules/RAM/readers.swift](
https://github.com/exelban/stats/blob/c78319bdecdf56a73b8edfd791f727a3c1c1451c/Modules/RAM/readers.swift#L56)
```
let used = active + inactive + speculative + wired + compressed - purgeable - external
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -330,6 +330,10 @@ if test "$my_htop_platform" = darwin; then
    AC_CHECK_FUNCS([mach_timebase_info])
    AC_CHECK_DECLS([IOMainPort], [], [], [[#include <IOKit/IOKitLib.h>]])
    AC_CHECK_DECLS([IOMasterPort], [], [], [[#include <IOKit/IOKitLib.h>]])
+
+   AC_CHECK_FUNCS([host_statistics64], [
+      AC_CHECK_TYPES([struct vm_statistics64], [], [], [[#include <mach/vm_statistics.h>]])
+   ], [])
 fi
 
 if test "$my_htop_platform" = pcp; then
--- a/darwin/DarwinMachine.c
+++ b/darwin/DarwinMachine.c
@@ -59,12 +59,20 @@ static unsigned DarwinMachine_allocateCPULoadInfo(processor_cpu_load_info_t* p) 
    return cpu_count;
 }
 
-static void DarwinMachine_getVMStats(vm_statistics_t p) {
+static void DarwinMachine_getVMStats(DarwinMachine* this) {
+#ifdef HAVE_STRUCT_VM_STATISTICS64
+   mach_msg_type_number_t info_size = HOST_VM_INFO64_COUNT;
+
+   if (host_statistics64(mach_host_self(), HOST_VM_INFO64, (host_info_t)&this->vm_stats, &info_size) != 0) {
+      CRT_fatalError("Unable to retrieve VM statistics64");
+   }
+#else
    mach_msg_type_number_t info_size = HOST_VM_INFO_COUNT;
 
-   if (host_statistics(mach_host_self(), HOST_VM_INFO, (host_info_t)p, &info_size) != 0) {
+   if (host_statistics(mach_host_self(), HOST_VM_INFO, (host_info_t)&this->vm_stats, &info_size) != 0) {
       CRT_fatalError("Unable to retrieve VM statistics");
    }
+#endif
 }
 
 void Machine_scan(Machine* super) {
@@ -74,7 +82,7 @@ void Machine_scan(Machine* super) {
    DarwinMachine_freeCPULoadInfo(&host->prev_load);
    host->prev_load = host->curr_load;
    DarwinMachine_allocateCPULoadInfo(&host->curr_load);
-   DarwinMachine_getVMStats(&host->vm_stats);
+   DarwinMachine_getVMStats(host);
    openzfs_sysctl_updateArcStats(&host->zfs);
 }
 
@@ -91,7 +99,7 @@ Machine* Machine_new(UsersTable* usersTable, uid_t userId) {
    DarwinMachine_allocateCPULoadInfo(&this->curr_load);
 
    /* Initialize the VM statistics */
-   DarwinMachine_getVMStats(&this->vm_stats);
+   DarwinMachine_getVMStats(this);
 
    /* Initialize the ZFS kstats, if zfs.kext loaded */
    openzfs_sysctl_init(&this->zfs);
--- a/darwin/DarwinMachine.h
+++ b/darwin/DarwinMachine.h
@@ -18,7 +18,11 @@ typedef struct DarwinMachine_ {
    Machine super;
 
    host_basic_info_data_t host_info;
+#ifdef HAVE_STRUCT_VM_STATISTICS64
+   vm_statistics64_data_t vm_stats;
+#else
    vm_statistics_data_t vm_stats;
+#endif
    processor_cpu_load_info_t prev_load;
    processor_cpu_load_info_t curr_load;
 
--- a/darwin/Platform.c
+++ b/darwin/Platform.c
@@ -292,13 +292,28 @@ double Platform_setCPUValues(Meter* mtr, unsigned int cpu) {
 
 void Platform_setMemoryValues(Meter* mtr) {
    const DarwinMachine* dhost = (const DarwinMachine*) mtr->host;
+#ifdef HAVE_STRUCT_VM_STATISTICS64
+   const struct vm_statistics64* vm = &dhost->vm_stats;
+#else
    const struct vm_statistics* vm = &dhost->vm_stats;
+#endif
    double page_K = (double)vm_page_size / (double)1024;
 
    mtr->total = dhost->host_info.max_mem / 1024;
+#ifdef HAVE_STRUCT_VM_STATISTICS64
+   natural_t used = vm->active_count + vm->inactive_count +
+              vm->speculative_count + vm->wire_count +
+              vm->compressor_page_count - vm->purgeable_count - vm->external_page_count;
+   mtr->values[MEMORY_METER_USED] = (double)(used - vm->compressor_page_count) * page_K;
+#else
    mtr->values[MEMORY_METER_USED] = (double)(vm->active_count + vm->wire_count) * page_K;
+#endif
    // mtr->values[MEMORY_METER_SHARED] = "shared memory, like tmpfs and shm"
+#ifdef HAVE_STRUCT_VM_STATISTICS64
+   mtr->values[MEMORY_METER_COMPRESSED] = (double)vm->compressor_page_count * page_K;
+#else
    // mtr->values[MEMORY_METER_COMPRESSED] = "compressed memory, like zswap on linux"
+#endif
    mtr->values[MEMORY_METER_BUFFERS] = (double)vm->purgeable_count * page_K;
    mtr->values[MEMORY_METER_CACHE] = (double)vm->inactive_count * page_K;
    // mtr->values[MEMORY_METER_AVAILABLE] = "available memory"


darwin/DarwinMachine.h
```diff
@@ -19,6 +19,9 @@ typedef struct DarwinMachine_ {
 
    host_basic_info_data_t host_info;
    vm_statistics_data_t vm_stats;
+#if defined(__arm64__)
+   vm_statistics64_data_t vm_stats64;
```
Explorer09, line 23: This structure does not seem to be exclusive to ARM64. Would it be used in Intel Macs, too? A reference to Apple documentation would be useful.

darwin/DarwinMachine.c
```diff
@@ -75,6 +85,9 @@ void Machine_scan(Machine* super) {
    host->prev_load = host->curr_load;
    DarwinMachine_allocateCPULoadInfo(&host->curr_load);
    DarwinMachine_getVMStats(&host->vm_stats);
+#if defined(__arm64__)
+   DarwinMachine_getVMStats64(&host->vm_stats64);
```
Explorer09, line 89: Any reason to duplicate the effort of retrieving both `vm_stats` structures?

darwin/DarwinMachine.c
```diff
@@ -67,6 +67,16 @@ static void DarwinMachine_getVMStats(vm_statistics_t p) {
    }
 }
 
+#if defined(__arm64__)
+static void DarwinMachine_getVMStats64(vm_statistics64_t p) {
```
Explorer09, line 71: Consider merging this function with `DarwinMachine_getVMStats` above, so as to avoid duplicate code.

You can use "`DarwinMachine* this`" as the input argument.

darwin/Platform.c
```diff
@@ -290,6 +290,24 @@ double Platform_setCPUValues(Meter* mtr, unsigned int cpu) {
    return CLAMP(total, 0.0, 100.0);
 }
 
+#if defined(__arm64__)
+void Platform_setMemoryValues(Meter *mtr) {
```
Explorer09, line 294: Consider merging two `Platform_setMemoryValues` functions together.

darwin/Platform.c
```diff
@@ -290,6 +290,24 @@ double Platform_setCPUValues(Meter* mtr, unsigned int cpu) {
...
+   const DarwinMachine *dhost = (const DarwinMachine *) mtr->host;
+   const struct vm_statistics64 *vm = &dhost->vm_stats64;
+   double page_K = (double) vm_page_size / (double) 1024;
+
+   mtr->total = dhost->host_info.max_mem / 1024;
+   int used = vm->active_count + vm->inactive_count +
```
Explorer09, line 300: Is it intentional to use signed `int` type here? Note that `vm->active_count` and similar members are in "`natural_t`" type, which is unsigned.

darwin/DarwinMachine.h
SuCicada: You are right. According to the documentation:
https://developer.apple.com/documentation/kernel/vm_statistics64_data_t
https://developer.apple.com/documentation/kernel/vm_statistics_data_t

**vm_statistics64_data_t** can be used after macOS 10.6+
**vm_statistics_data_t** can be used after macOS 10.0+

Explorer09: @SuCicada  I prefer keeping a fallback mechanism when `vm_statistics64_data_t` is not available. There is no reason to remove 32-bit Mac support (htop documentation doesn't state a minimum required version of macOS).

BenBE: Haven't looked at the details of the implementation yet, but would it be feasible to check for `vm_statistics64_data_t` and `vm_statistics_data_t` in `configure` and use the 64 bit version everywhere when available? And if none of both is available, we either fall back entirely or bail out at compile time.

Explorer09: @BenBE Since it is said that `vm_statistics_data_t` is available in OS X 10.0 (the very first Mac OS X), these is no need to check this one in particular in `configure`. Just checking for `vm_statistics64_data_t` is enough.

darwin/DarwinMachine.c
SuCicada: Just to prevent affecting existing functionality, but I checked the code again and it didn't affect anywhere else.
Same as above. I fix that.

BenBE:
Two general notes:

1. Squash the two commits, as commits should show a direct way from A to B without all the steps inside the maze … Mostly split commits, when there are independent changes that work without each other or there's something preparing for a later change

2. There should be no blank after casts, like this `(foo)expression` …

Also note the remarks earlier re implementation for both 32 and 64 bit VM structures using conditional compilation.

darwin/DarwinMachine.h
SuCicada: So I will restore support for vm state 32.

BenBE: Exactly: Use `vm_statistics64_data_t` when available, otherwise fall back to `vm_statistics_data_t`.

BenBE:
LGTM.

Any particular reason for checking an architecture setting (64 bit support of the architecture) over explicit testing for availability of a concrete type/structure in `configure.ac`?

SuCicada:
> LGTM.
> 
> Any particular reason for checking an architecture setting (64 bit support of the architecture) over explicit testing for availability of a concrete type/structure in `configure.ac`?

Oh, You are right. It is more explicit to test the availability of a concrete type/structure in `configure.ac`.
I will try

SuCicada:
Thank you for your guidance, I have resubmitted.

1. I added checking of vm_statistics64 at `configure.ac`.
2. I use `HAVE_VM_STATISTICS64` for conditional compilation instead.

What do you think

configure.ac
```diff
@@ -330,6 +330,21 @@ if test "$my_htop_platform" = darwin; then
...
+#include <mach/vm_statistics.h>
+      ]], [[
+            mach_msg_type_number_t info_size = HOST_VM_INFO64_COUNT;
+            vm_statistics64_data_t vm_stat;
+            host_statistics64(mach_host_self(), HOST_VM_INFO64, (host_info64_t)&vm_stat, &info_size);
+      ]]
+   )],
+   AC_DEFINE([HAVE_VM_STATISTICS64], 1, [The vm_statistics64 features is supported.])
+   AC_MSG_RESULT(yes),
+   AC_MSG_RESULT(no))
```
BenBE, line 347: Why not use things like this (untested)?
```suggestion
   AC_CHECK_FUNCS([host_statistics64], [], [], [[#include <mach/vm_statistics.h>]])
   AC_CHECK_DECLS([vm_statistics64_data_t], [], [], [[#include <mach/vm_statistics.h>]])
```

configure.ac
```diff
@@ -330,6 +330,21 @@ if test "$my_htop_platform" = darwin; then
    AC_CHECK_FUNCS([mach_timebase_info])
    AC_CHECK_DECLS([IOMainPort], [], [], [[#include <IOKit/IOKitLib.h>]])
    AC_CHECK_DECLS([IOMasterPort], [], [], [[#include <IOKit/IOKitLib.h>]])
+
+   AC_MSG_CHECKING([for vm_statistics64 supported])
```
Explorer09, line 334: I thought you can use `AC_CHECK_TYPES` macro to simplify the check.
Maybe this is sufficient:
```shell
AC_CHECK_TYPES([struct vm_statistics64], [], [], [[#include <mach/vm_statistics.h>]]
dnl This defines HAVE_STRUCT_VM_STATISTICS64
```

configure.ac
Explorer09: @BenBE You and I suggested the same thing at the same time...

By the way, AC_CHECK_DECLS is not for checking type declarations, use AC_CHECK_TYPES instead. See my other comment.

BenBE: I had a look at the AutoConf docs, but didn't see it there somehow …

SuCicada:
Makes sense, but this will produce 2 macros. If use 
If someone's computer cannot meet host_statistics64 and vm_statistics64_data_t at the same time, (very unlikely)
Then it will happen:

| host_statistics64 support | vm_statistics64_data_t support |  result | 
| - | - | - |
| yes | no|  can compile and run normally, but the memory value is incorrect |
| no | yes |  can compile normally, but fails to run:<br> `[1] 16517 segmentation fault ./htop` |

I am not sure if there is a better solution, or if we can ignore this very unlikely situation.

BenBE:
AutoConf has the ability to nest those macros, i.e. only check the second if the first succeeded. Likewise for the second macro: Only define the feature flag once the second macro succeeded too.

Untested:

```autoconf
   AC_CHECK_FUNCS([host_statistics64], [
      AC_CHECK_TYPES([struct vm_statistics64], [
         AC_DEFINE([HAVE_VM_STATISTICS64], 1, [The vm_statistics64 features is supported.])
      ], [], [[#include <mach/vm_statistics.h>]])
   ], [], [[#include <mach/vm_statistics.h>]])
```

Check for `HAVE_VM_STATISTICS64` in the code …

Explorer09:
@BenBE 
Both `AC_CHECK_FUNCS` and `AC_CHECK_TYPES` would define the "HAVE_" C macro names for you. There's no need to use `AC_DEFINE` manually.

There would be `HAVE_STRUCT_VM_STATISTICS64` and `HAVE_HOST_STATISTICS64` defined automatically. Just pick a name that sounds better in the C code. And put that macro check as an inner call in `configure.ac`.

```m4
AC_CHECK_TYPES([struct vm_statistics64], [
   AC_CHECK_FUNCS([host_statistics64])
   ], [], [[#include <mach/vm_statistics.h>]])
dnl Then use HAVE_HOST_STATISTICS64 in the C code
```

BenBE:
I know. The `AC_DEFINE` was there to use the other code AS-IS and have the implicit AND …

SuCicada:
Thank you for your patient guidance, I have resubmitted.

(In fact, I had previously attempted this method but mistakenly believed it didn't work🫣. After your confirmation, I have re-verified and got the correct compile result.👍)

SuCicada:
Thank you for your patient guidance, I have resubmitted.

(In fact, I had previously attempted this method but mistakenly believed it didn't work🫣. After your confirmation, I have re-verified and got the correct compile result.👍)

darwin/Platform.c
```diff
@@ -292,13 +292,28 @@ double Platform_setCPUValues(Meter* mtr, unsigned int cpu) {
...
 
    mtr->total = dhost->host_info.max_mem / 1024;
+#ifdef HAVE_STRUCT_VM_STATISTICS64
+   natural_t used = vm->active_count + vm->inactive_count +
+              vm->speculative_count + vm->wire_count +
+              vm->compressor_page_count - vm->purgeable_count - vm->external_page_count;
+   mtr->values[MEMORY_METER_USED] = (double)(used - vm->compressor_page_count) * page_K;
+#else
    mtr->values[MEMORY_METER_USED] = (double)(vm->active_count + vm->wire_count) * page_K;
+#endif
```
BenBE, line 310: Any reasons for this calculation to differ between those two places, besides missing fields from the old struct?

Also, `vm->compressor_page_count` is substracted twice AFAICS.

BenBE (changes requested):
Not sure about the changed memory calculation. There seems something off …

configure.ac
```diff
@@ -330,6 +330,10 @@ if test "$my_htop_platform" = darwin; then
    AC_CHECK_FUNCS([mach_timebase_info])
    AC_CHECK_DECLS([IOMainPort], [], [], [[#include <IOKit/IOKitLib.h>]])
    AC_CHECK_DECLS([IOMasterPort], [], [], [[#include <IOKit/IOKitLib.h>]])
+
+   AC_CHECK_FUNCS([host_statistics64], [
```
Explorer09, line 334: AC_CHECK_FUNCS does not take the fourth argument. The `#include <mach/mach.h>` is unnecessary and would be ignored by Autoconf.

darwin/Platform.c
SuCicada: As https://github.com/htop-dev/htop/issues/631 and https://github.com/htop-dev/htop/issues/955 said, the current memory calculation may have a bit problem. 

I referenced the code of  [exelban/stats](https://github.com/exelban/stats/blob/c78319bdecdf56a73b8edfd791f727a3c1c1451c/Modules/RAM/readers.swift#L56) and [GuillaumeGomez/sysinfo](
https://github.com/GuillaumeGomez/sysinfo/blob/9705c681a61f9443b63115b4bc7720e86a3d63ee/src/unix/apple/system.rs#L151
)(used by `ClementTsang/bottom` mentioned by https://github.com/htop-dev/htop/issues/955 ) 

There are some differences in details, but it seems that the main thing missing is adding `vm->compressor_page_count`

It seems that many people have raised the same issue of memory parameter display. 
So I just want macOS to display the correct memory parameters. But there may be something I haven't considered.

time-killer-games:
I would suggest doing what Apple officially does in their top(1) program, as I suggested in my recent issue: 

https://github.com/htop-dev/htop/issues/1573

BenBE (approved):

Explorer09:
Well, I didn't expect this PR (pull request) to be merged so soon, since there are discussions going in #1573, and there are people who disagrees with the formula of calculating the used memory.

Note that I take a neutral position here and won't suggest which way of calculating used memory is "correct", but I think @time-killer-games can now make a PR based on the current `main` and suggest a better formula, and let the discussions continue.

BenBE:
I'm also mostly neutral on the specifics. Also as far as I checked before merging there was nothing actually critical open, that was a blocker. Mostly cosmetic stuff.

time-killer-games:
> I'm also mostly neutral on the specifics. Also as far as I checked before merging there was nothing actually critical open, that was a blocker. Mostly cosmetic stuff.

It's not very pressing, if you need time to talk it over or if you like this solution better than what top does, no worries. I found out in my discussion with the fastfetch lead dev this solution matches neofetch.

time-killer-games:
My pull request has been submitted: https://github.com/htop-dev/htop/pull/1439#issuecomment-2566728125

However please note I'm away from home and won't be able to test on my mac mini until tomorrow at the earliest. I'm hoping the code change I made does actually match what top(1) shows now.

BenBE:
The time around and after X-mas is always kinda busy for everyone. Have been busy the past few days myself and will likely be for some time. No need to hurry things along. I'll take a look at it when I've got a moment.

PR #1446, opened 2024-04-13
CI: Remove Autoconf version number from *BSD jobs

Explorer09:
Recently FreeBSD updated its Autoconf package to 2.72 and removed the old Autoconf-2.71 package. This caused build error in our CI jobs.

Remove the Autoconf version number from FreeBSD and DragonFlyBSD jobs.

For OpenBSD, add shell commands to fetch the latest versions of Autoconf and Automake packages.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -279,12 +279,24 @@ jobs:
             release: '7.4'
             usesh: true
             prepare: |
-              pkg_add gmake autoconf-2.71 automake-1.16.5 git
+              pkg_add gmake git
               git config --global --add safe.directory /home/runner/work/htop/htop
             run: |
               set -e
-              export AUTOCONF_VERSION=2.71
-              export AUTOMAKE_VERSION=1.16
+              autoconf_version_full=$(pkg_info -Q autoconf |
+                LC_ALL=C sed 's/^autoconf-\([0-9]*\.[0-9]*\)\(p[.0-9]*\)\{0,1\}$/\1\2/p; d' |
+                LC_ALL=C sort -n -r -t . -k 1,1 -k 2,2 |
+                sed '1 q')
+              automake_version_full=$(pkg_info -Q automake |
+                LC_ALL=C sed 's/^automake-\([0-9]*\.[0-9]*\)\(\.[0-9]*\)\{0,1\}\(p[0-9]*\)\{0,1\}$/\1\2\3/p; d' |
+                LC_ALL=C sort -n -r -t . -k 1,1 -k 2,2 -k 3,3 |
+                sed '1 q')
+              pkg_add -v autoconf-${autoconf_version_full} automake-${automake_version_full}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -233,7 +233,7 @@ jobs:
             release: '14.0'
             usesh: true
             prepare: |
-              pkg install -y gmake autoconf-2.71 automake git
+              pkg install -y gmake autoconf automake git
```
BenBE, line 236: Also, I noticed that `pkg` complains about wanting to be upgraded first. Please do an explicit upgrade of `pkg` first

Explorer09: @BenBE I know nothing about FreeBSD package management and all my personal attempts of silencing that "New version of pkg detected; it needs to be installed first" have failed. That message would pop up no matter what I do.

Maybe I can let @fraggerfox look at this issue?

fraggerfox: @BenBE / @Explorer09 : The warning from `pkg` is normal since the version coming in the base image `14.0-RELEASE` has changed and it will do a self update before attempting to update / install the packages.

The removal of the version number is fine, I guess when I set up the CI I should have checked if `2.71` was referring to the latest package, since there are a other versions of `autoconf` package being maintained in FreeBSD ports.

Explorer09: @fraggerfox For the OpenBSD job I think you do need a function to detect the latest version of Autoconf. By the time I made this fix, I only know that it's fine to remove the version number as FreeBSD never keeps two versions of Autoconf package simultaneously.

I only made the fix for FreeBSD and DragonFlyBSD as they use the same package manager. I would leave it to you to fix the OpenBSD one. (It's non-trivial, but I have a rough idea on how to make it.)

```shell
autoconf_latest_version=$( "command that lists all versions of autoconf packages" | sort -r -t '.-' -k 2n | sed 's/^autoconf-//p; q;'
)
```

fraggerfox: I also seem to be facing an issue building in NetBSD `10.0` (the `gcc` version has been updated to `10.5.0`), I think it might be due to the compiler upgrade, 

I am looking into that over here https://github.com/htop-dev/htop/pull/1447

Once that is done, will look into the OpenBSD configuration, I am also not too familiar with how OpenBSD does its `autoconf` detect, but I think your approach is in the right direction.

BenBE: Okay, will await your feedback before going ahead with this, so we can kinda group those CI changes …

Explorer09: @fraggerfox I made the OpenBSD one as well just now:
2eaf731ad9469d1b18170bf2a3bf2e259c55c4c1

fraggerfox: @Explorer09 : The changes look fine, I think we can go with that.

@BenBE : This PR can go in first, since the build in NetBSD `9.3` still works as expected,I will need a bit more time to investigate the failure in `10.0`.

PR #1448, opened 2024-04-14
'FILE*' and 'struct stat' variable rename. No code changes.

Explorer09:
Rename local `FILE*` and `struct stat` variables to make the naming consistent throughout htop codebase.

### DIFF
--- a/Compat.c
+++ b/Compat.c
@@ -50,11 +50,11 @@ int Compat_faccessat(int dirfd,
    }
 
    // Fallback to stat(2)/lstat(2) depending on flags
-   struct stat statinfo;
+   struct stat sb;
    if (flags) {
-      ret = lstat(pathname, &statinfo);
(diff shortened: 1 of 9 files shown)


BenBE:
Why `sb` over `st`?

Explorer09:
> Why `sb` over `st`?

Already written in the commit description: I didn't see any documentation uses the `st` naming. This  buffer is usually named `sb` or `statbuf`.
* [Example code in POSIX uses "statbuf" or "buf"](https://pubs.opengroup.org/onlinepubs/009696799/functions/stat.html)
* [FreeBSD names it "sb" in its man page](https://man.freebsd.org/cgi/man.cgi?stat(2))
* [Linux man page uses both "statbuf" and "sb" naming](https://man7.org/linux/man-pages/man2/stat.2.html)

Issue #1449, opened 2024-04-15
Memory leak at Settings.c:867 (legacyDotfile)


LuMingYinDetect:
The detailed path of the defect:

1.At line 796 in /htop/Settings.c, a pointer variable named legacyDotfile is defined. This pointer variable allocates a dynamic memory area through the function String_cat at line 818, as shown in the figure below:
https://github.com/LuMingYinDetect/htop_defects/blob/main/htop_1.png

2.The function String_cat allocates a new dynamic memory area and returns it. At line 130 of String_cat, a pointer variable named out is defined and a dynamic memory area is allocated through the function xMalloc. This pointer out is returned at line 134, as shown in the figure below:
https://github.com/LuMingYinDetect/htop_defects/blob/main/htop_2.png

3.After the execution of String_cat, if the if statement at line 825 returns false, the program will not execute the release operation of the dynamic memory area pointed to by legacyDotfile at line 826. Similarly, if the if statement at line 842 also returns false, the program will not execute the release operation of the dynamic memory area pointed to by legacyDotfile at line 850. In this scenario, the program will return at line 867. During this process, the dynamic memory area pointed to by legacyDotfile remains unreleased, thus resulting in a memory leak defect, as illustrated in the figure below:
https://github.com/LuMingYinDetect/htop_defects/blob/main/htop_3.png

BenBE:
Linking the proper places in the code (note the commit hash in the links):
1. https://github.com/htop-dev/htop/blob/8122fc394e05507c514425b1f96637668dcefe65/Settings.c#L796
2. https://github.com/htop-dev/htop/blob/8122fc394e05507c514425b1f96637668dcefe65/Settings.c#L818
3. https://github.com/htop-dev/htop/blob/8122fc394e05507c514425b1f96637668dcefe65/Settings.c#L850

Usually works best to reference specific code locations as images are kinda unwieldy for that purpose.

PR #1450, opened 2024-04-15
Memory leak when a legacy config file exists

BenBE:
Fixes: #1449


Explorer09:
Oops indeed. I didn't catch this when I wrote 15b4bc45b2b0ccf2d93164b2b1b05fb94cc89220. Ideally the variable scope of `legacyDotfile` can be shrunken further.

cgzones (approved):

PR #1452, opened 2024-04-15
Prevent a memory leak when legacy config file exists

Explorer09:
(This supercedes pull request #1450)

Fix a regression caused by 15b4bc45b2b0ccf2d93164b2b1b05fb94cc89220

Resolves: #1449

### DIFF
--- a/Settings.c
+++ b/Settings.c
@@ -815,11 +815,12 @@ Settings* Settings_new(unsigned int initialCpuCount, Hashtable* dynamicMeters, H
          configDir = String_cat(home, "/.config");
          htopDir = String_cat(home, "/.config/htop");
       }
-      legacyDotfile = String_cat(home, "/.htoprc");
       (void) mkdir(configDir, 0700);
       (void) mkdir(htopDir, 0700);
       free(htopDir);
       free(configDir);
+
+      legacyDotfile = String_cat(home, "/.htoprc");
       struct stat st;
       int err = lstat(legacyDotfile, &st);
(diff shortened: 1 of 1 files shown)


Settings.c
```diff
@@ -847,8 +849,9 @@ Settings* Settings_new(unsigned int initialCpuCount, Hashtable* dynamicMeters, H
             unlink(legacyDotfile);
          }
       }
-      free(legacyDotfile);
    }
+   free(legacyDotfile);
+
```
BenBE, line 854: Moving it down to the end of the function in my PR was intentional (to avoid splitting the various steps for reading the config with this cascade of `if (!ok …)` statements …

Settings.c
```diff
@@ -815,11 +815,12 @@ Settings* Settings_new(unsigned int initialCpuCount, Hashtable* dynamicMeters, H
          configDir = String_cat(home, "/.config");
          htopDir = String_cat(home, "/.config/htop");
       }
-      legacyDotfile = String_cat(home, "/.htoprc");
       (void) mkdir(configDir, 0700);
       (void) mkdir(htopDir, 0700);
       free(htopDir);
       free(configDir);
+
+      legacyDotfile = String_cat(home, "/.htoprc");
```
BenBE, line 823: Ported over to my PR. Thx for the idea.

Explorer09:
I originally thought I can rework the logic so that `legacyDotfile` string can be allocated later, but I failed. Closing this request as redundant.

PR #1459, opened 2024-04-17
Add the timezone in clock meter

eworm-de:
I connect to servers with non-local timezones every now and then... Let's add the timezone in clock meter to minimize confusion there.

### DIFF
--- a/ClockMeter.c
+++ b/ClockMeter.c
@@ -26,7 +26,7 @@ static void ClockMeter_updateValues(Meter* this) {
 
    struct tm result;
    const struct tm* lt = localtime_r(&host->realtime.tv_sec, &result);
-   strftime(this->txtBuffer, sizeof(this->txtBuffer), "%H:%M:%S", lt);
+   strftime(this->txtBuffer, sizeof(this->txtBuffer), "%H:%M:%S %Z", lt);
 }
 
 const MeterClass ClockMeter_class = {
--- a/DateTimeMeter.c
+++ b/DateTimeMeter.c
@@ -26,7 +26,7 @@ static void DateTimeMeter_updateValues(Meter* this) {
 
    struct tm result;
    const struct tm* lt = localtime_r(&host->realtime.tv_sec, &result);
-   strftime(this->txtBuffer, sizeof(this->txtBuffer), "%F %H:%M:%S", lt);
+   strftime(this->txtBuffer, sizeof(this->txtBuffer), "%F %H:%M:%S %Z", lt);
 }
 
 const MeterClass DateTimeMeter_class = {


fasterit:
I would at least make this optional (config option) as (... how do I say that politely ...) it is not a sign of following best practices to run distributed servers with clocks in local time zones.
/DLange

Explorer09:
> I would at least make this optional (config option) as (... how do I say that politely ...) it is not a sign of following best practices to run distributed servers with clocks in local time zones. /DLange

Nah. It's about the _display_ of the system time in UTC or in local time. Note the colck meter uses `localtime_r` already and not `gmtime_r`. But without the UTC keyword it can cause the user confusion.

The switch between local time and UTC can be done as a separate feature.

eworm-de:
Whether or not the timezone being local on a server makes sense... Depends.

But that's the point: My workstation has local timezone `CEST`, and the server has `UTC`. Looking at the clock from server's `htop` shocked me for a little moment. 🙃

Will have a look to make it configurable.

fasterit:
I am confused how sysadmins would be confused about the `TZ` they set on their own user / their root user.
And why in `ClockMeter.c`  but not in `DateTimeMeter.c` ?

fasterit:
> But that's the point: My workstation has local timezone `CEST`, and the server has `UTC`. Looking at the clock from server's `htop` shocked me for a little moment. 🙃

Bring your `TZ` to the server. `TZ="Europe/Berlin" htop` works.
(so does fixing your ssh env :smile:)

Explorer09:
```text
grep -r localtime
ClockMeter.c:   const struct tm* lt = localtime_r(&host->realtime.tv_sec, &result);
DateMeter.c:   const struct tm* lt = localtime_r(&host->realtime.tv_sec, &result);
DateTimeMeter.c:   const struct tm* lt = localtime_r(&host->realtime.tv_sec, &result);
Process.c:   (void) localtime_r(&this->starttime_ctime, &date);
```
That's all the code that would be affected by time zones. And yes, if you show the timezone offset in ClockMeter, it should also show in DateTimeMeter.

eworm-de:
> I am confused how sysadmins would be confused about the `TZ` they set on their own user / their root user.

Well, my confusion was not about the settings of the server, but more about what terminal window I was looking at. 😜

> And why in `ClockMeter.c` but not in `DateTimeMeter.c` ?

Because that's what I use. 😁 Added it to date/time meter in another commit, but still without any configuration option.

> Bring your `TZ` to the server. `TZ="Europe/Berlin" htop` works. (so does fixing your ssh env 😄)

I would have to add that in user configuration of any server... Using environment variables via `ssh` is a nice idea, but does not work out: This requires the server to accept these variables (see `AcceptEnv` and `PermitUserEnvironment`  in `sshd_config` man page), and the default is to not do so.

fasterit:
We have discussed this PR on IRC and nobody likes it™.

You could make it configurable, a meter mode or a separate date-time-meter-with-TZ but we will not force the change on the existing users as our scientific poll (N=2) found this proposed change to be ugly.

eworm-de:
> [...] as our scientific poll (N=2) found this proposed change to be ugly.

🤣

Never mind, fine with me. Adding just another configuration option is not worth it I think.

PR #1454, opened 2024-04-15
Meter mode code refactor

Explorer09:
The main goal of this set of patches is to remove the use of `CUSTOM_METERMODE` in `defaultMode` values, and to adjust `Meter_setMode()` code in preparation of the `supportedModes` feature (PR #1387).

PR #1460, opened 2024-04-19
Add forward declarations in some headers

Explorer09:
I'm not sure this code quality issue is worth fixing in htop, but I hope getting some discussions.

In #1454 I mentioned a "header dependency hell" that motivated me to split the `MeterModeId` type into a separate header. It is this one.

When I added an `#include "Meter.h"` line into `Settings.h`, a _circular inclusion_ of headers would happen and that results in build errors in many `.c` files.

`Meter.h` includes `Machine.h` which includes `Settings.h` which would then include `Meter.h`, but due to the header guard, the latter `Meter.h` include line resolves to no-op. Sometimes a module might include just `Machine.h` or just `Settings.h` but the circular inclusion can happen either case.

When a C header depends on a `typedef` of another type, it has to include the header defining that type as the type redefinition (`typedef` defining the same type twice) is an error in C. Forward declarations only work with `struct`, `union`, `enum` and `class` (C++) names, but not `typedef`s, but htops uses `typedef`s extensively even for `struct`s in the coding style.

When a structure like `Meter` only include a _reference_ of another class such as `Machine`, there's no need to fully define the `Machine` type nor include the `Machine.h` in order to use it (in the header; the implementation part, i.e. `Meter.c` would have to include `Machine.h` of course). A forward declaration could work equally well for a header.

The commit I presented is not a complete fix. It's a part of header changes that, when using forward declarations, can fix build errors I encountered. Is it a good idea to apply such forward declarations whenever possible in htop headers?

### DIFF
--- a/CRT.h
+++ b/CRT.h
@@ -14,6 +14,8 @@ in the source distribution for its full text.
 #include "Settings.h"
 
 
+struct Settings_; // IWYU pragma: keep
+
 typedef enum TreeStr_ {
    TREE_STR_VERT,
    TREE_STR_RTEE,
@@ -202,7 +204,7 @@ void CRT_setMouse(bool enabled);
 #define CRT_setMouse(enabled)
 #endif
 
-void CRT_init(const Settings* settings, bool allowUnicode, bool retainScreenOnExit);
+void CRT_init(const struct Settings_* settings, bool allowUnicode, bool retainScreenOnExit);
 
 void CRT_done(void);
 
--- a/Machine.h
+++ b/Machine.h
@@ -32,6 +32,9 @@ in the source distribution for its full text.
 #define MAX_READ 2048
 #endif
 
+struct Panel_; // IWYU pragma: keep
+struct Settings_; // IWYU pragma: keep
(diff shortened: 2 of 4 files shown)


BenBE:
IMO we should keep forward declarations to a minimum. As htop already uses IWYU for minimizing includes, it's mostly a matter of keeping those recursive dependencies at bay. Apart from some core types, there should – in the best case – be no further such loops.

Regarding this concrete PR: Not a fan of it at all … Personally I find that repeated `struct Foo_` instead of just `Foo` distracting and thus causing the code to be harder to read. Thus those long-form declarations should be kept at a bare minimum, where there's no real other chance to avoiding them.

Also, the code style (although IIRC not explicitly written) encourages to scope includes of headers to be as minimal as possible. If you need a type only for an implementation, only include it with the `.c` file, instead of making it part of the public interface. Usually IWYU will hint you at this whenever it is possible directly.

Explorer09:
Is there anyone else wanting to comment on this issue? I am considering closing this PR.

PR #1461, opened 2024-04-20
Fix and improve NICE on FreeBSD

herrhotzenplotz:
Previously the niceness value on FreeBSD was utterly broken (to my and others annoyance):

 - it displayed out-of-range values for idletime and realtime scheduled processes
 - there was no indicator for such scheduled processes

This fixes:

 - The calculation of the niceness value (see https://github.com/freebsd/freebsd-src/blob/main/usr.bin/top/machine.c#L1182)
 - Add a column (SCHEDCLASS) that tells whether a process is idletime, realtime, timeshared or otherwise scheduled

I have tested this with other people on a few architectures and different versions of FreeBSD and it seems to be working just fine.

Please tell me if I can improve on this or if something is utterly wrong and I should fix.

### DIFF
--- a/Process.h
+++ b/Process.h
@@ -8,6 +8,7 @@ Released under the GNU GPLv2+, see the COPYING file
 in the source distribution for its full text.
 */
 
+#include <limits.h>
 #include <stdbool.h>
 #include <stdint.h>
 #include <sys/types.h>
@@ -23,6 +24,9 @@ in the source distribution for its full text.
 #define PROCESS_FLAG_SCHEDPOL        0x00000004
 
 #define DEFAULT_HIGHLIGHT_SECS 5
+
+/* Sentinel value for an unknown niceness in Process.nice */
+#define PROCESS_NICE_UNKNOWN (-LONG_MAX)
 
 typedef enum Tristate_ {
    TRI_INITIAL = 0,
--- a/freebsd/FreeBSDProcess.c
+++ b/freebsd/FreeBSDProcess.c
@@ -55,6 +55,7 @@ const ProcessFieldData Process_fields[LAST_PROCESSFIELD] = {
 #endif
    [JID] = { .name = "JID", .title = "JID", .description = "Jail prison ID", .flags = 0, .pidColumn = true, },
    [JAIL] = { .name = "JAIL", .title = "JAIL        ", .description = "Jail prison name", .flags = 0, },
+   [SCHEDCLASS] = { .name = "SCHEDCLASS", .title = "SC", .description = "Scheduling Class (Timesharing, Realtime, Idletime)", .flags = 0, },
    [EMULATION] = { .name = "EMULATION", .title = "EMULATION        ", .description = "System call emulation environment (ABI)", .flags = 0, },
 };
 
@@ -73,22 +74,41 @@ void Process_delete(Object* cast) {
    free(this);
 }
 
+static const char FreeBSD_schedclassChars[MAX_SCHEDCLASS] = {
+   [SCHEDCLASS_UNKNOWN] = '?',     // Something went wrong or the base system has a new scheduling class
+   [SCHEDCLASS_INTR_THREAD] = '-', // interrupt thread, these have special handling of priority
+   [SCHEDCLASS_IDLE] = 'i',        // idletime scheduling
+   [SCHEDCLASS_TIMESHARE] = ' ',   // timesharing process scheduling (regular processes are timeshared)
+   [SCHEDCLASS_REALTIME] = 'r',    // realtime scheduling
+};
+
 static void FreeBSDProcess_rowWriteField(const Row* super, RichString* str, ProcessField field) {
    const FreeBSDProcess* fp = (const FreeBSDProcess*) super;
 
    char buffer[256]; buffer[255] = '\0';
+   char sched_class;
    int attr = CRT_colors[DEFAULT_COLOR];
    size_t n = sizeof(buffer) - 1;
 
    switch (field) {
    // add FreeBSD-specific fields here
    case JID: xSnprintf(buffer, n, "%*d ", Process_pidDigits, fp->jid); break;
+
    case JAIL:
       Row_printLeftAlignedField(str, attr, fp->jname ? fp->jname : "N/A", 11);
       return;
+
    case EMULATION:
       Row_printLeftAlignedField(str, attr, fp->emul ? fp->emul : "N/A", 16);
       return;
+
+   case SCHEDCLASS:
+      assert(0 <= fp->sched_class && fp->sched_class < ARRAYSIZE(FreeBSD_schedclassChars));
+      sched_class = FreeBSD_schedclassChars[fp->sched_class];
+      assert(sched_class);
+      xSnprintf(buffer, n, " %c", sched_class);
+      break;
+
    default:
       Process_writeField(&fp->super, str, field);
       return;
@@ -109,6 +129,8 @@ static int FreeBSDProcess_compareByKey(const Process* v1, const Process* v2, Pro
       return SPACESHIP_NULLSTR(p1->jname, p2->jname);
    case EMULATION:
       return SPACESHIP_NULLSTR(p1->emul, p2->emul);
+   case SCHEDCLASS:
+      return SPACESHIP_NUMBER(p1->sched_class, p2->sched_class);
    default:
       return Process_compareByKey_Base(v1, v2, key);
    }
--- a/freebsd/FreeBSDProcess.h
+++ b/freebsd/FreeBSDProcess.h
@@ -13,12 +13,23 @@ in the source distribution for its full text.
 #include "Process.h"
 #include "Machine.h"
 
+typedef enum {
+   SCHEDCLASS_UNKNOWN = 0,
+
+   SCHEDCLASS_INTR_THREAD, /* interrupt thread */
+   SCHEDCLASS_REALTIME,
+   SCHEDCLASS_TIMESHARE, /* Regular scheduling */
+   SCHEDCLASS_IDLE,
+
+   MAX_SCHEDCLASS,
+} FreeBSDSchedClass;
 
 typedef struct FreeBSDProcess_ {
    Process super;
    int   jid;
    char* jname;
    char* emul;
+   FreeBSDSchedClass sched_class;
 } FreeBSDProcess;
 
 extern const ProcessClass FreeBSDProcess_class;
--- a/freebsd/FreeBSDProcessTable.c
+++ b/freebsd/FreeBSDProcessTable.c
@@ -248,14 +248,45 @@ void ProcessTable_goThroughEntries(ProcessTable* super) {
 
       proc->priority = kproc->ki_pri.pri_level - PZERO;
 
-      if (String_eq("intr", kproc->ki_comm) && (kproc->ki_flag & P_SYSTEM)) {
-         proc->nice = 0; //@etosan: intr kernel process (not thread) has weird nice value
-      } else if (kproc->ki_pri.pri_class == PRI_TIMESHARE) {
-         proc->nice = kproc->ki_nice - NZERO;
-      } else if (PRI_IS_REALTIME(kproc->ki_pri.pri_class)) {
-         proc->nice = PRIO_MIN - 1 - (PRI_MAX_REALTIME - kproc->ki_pri.pri_level);
-      } else {
-         proc->nice = PRIO_MAX + 1 + kproc->ki_pri.pri_level - PRI_MIN_IDLE;
+      switch (PRI_BASE(kproc->ki_pri.pri_class)) {
+         /* Handling of the below is explained in the FreeBSD base system in:
+          * /usr/src/usr.bin/top/machine.c (function format_nice) */
+         case PRI_ITHD:
+            fp->sched_class = SCHEDCLASS_INTR_THREAD;
+            proc->nice = 0;
+            break;
+
+         case PRI_REALTIME:
+            fp->sched_class = SCHEDCLASS_REALTIME;
+
+            /* Different for KPROCs and user procs */
+            if (kproc->ki_flag & P_KPROC) {
+               proc->nice = kproc->ki_pri.pri_native - PRI_MIN_REALTIME;
+            } else {
+               proc->nice = kproc->ki_pri.pri_user - PRI_MIN_REALTIME;
+            }
+            break;
+
+         case PRI_IDLE:
+            fp->sched_class = SCHEDCLASS_IDLE;
+
+            /* Different for KPROCs and user procs */
+            if (kproc->ki_flag & P_KPROC) {
+               proc->nice = kproc->ki_pri.pri_native - PRI_MIN_IDLE;
+            } else {
+               proc->nice = kproc->ki_pri.pri_user - PRI_MIN_IDLE;
+            }
+            break;
+
+         case PRI_TIMESHARE:
+            fp->sched_class = SCHEDCLASS_TIMESHARE;
+            proc->nice = kproc->ki_nice - NZERO;
+            break;
+
+         default:
+            fp->sched_class = SCHEDCLASS_UNKNOWN;
+            proc->nice = PROCESS_NICE_UNKNOWN;
+            break;
       }
 
       /* Taken from: https://github.com/freebsd/freebsd-src/blob/1ad2d87778970582854082bcedd2df0394fd4933/sys/sys/proc.h#L851 */
--- a/Process.c
+++ b/Process.c
@@ -649,10 +649,15 @@ void Process_writeField(const Process* this, RichString* str, RowField field) {
    case M_RESIDENT: Row_printKBytes(str, this->m_resident, coloring); return;
    case M_VIRT: Row_printKBytes(str, this->m_virt, coloring); return;
    case NICE:
-      xSnprintf(buffer, n, "%3ld ", this->nice);
-      attr = this->nice < 0 ? CRT_colors[PROCESS_HIGH_PRIORITY]
-         : this->nice > 0 ? CRT_colors[PROCESS_LOW_PRIORITY]
-         : CRT_colors[PROCESS_SHADOW];
+      if (this->nice == PROCESS_NICE_UNKNOWN) {
+         xSnprintf(buffer, n, "N/A ");
+         attr = CRT_colors[PROCESS_SHADOW];
+      } else {
+         xSnprintf(buffer, n, "%3ld ", this->nice);
+         attr = this->nice < 0 ? CRT_colors[PROCESS_HIGH_PRIORITY]
+            : this->nice > 0 ? CRT_colors[PROCESS_LOW_PRIORITY]
+            : CRT_colors[PROCESS_SHADOW];
+      }
       break;
    case NLWP:
       if (this->nlwp == 1)
--- a/freebsd/ProcessField.h
+++ b/freebsd/ProcessField.h
@@ -12,6 +12,7 @@ in the source distribution for its full text.
    JID = 100,                    \
    JAIL = 101,                   \
    EMULATION = 102,              \
+   SCHEDCLASS = 103,             \
                                  \
    DUMMY_BUMP_FIELD = CWD,       \
    // End of list


swills:
Screenshot_20240420_140527_cropped
Here's a before and after screenshot

clausecker:
Very cool!  Does this also fix that rtprio tasks were not shown in blue on the CPU usage bar? 
 @BenBE, what do you think of this?  Is there a way to show for the scheduling class what level the process has within the scheduling class?

herrhotzenplotz:
> Very cool! Does this also fix that rtprio tasks were not shown in blue on the CPU usage bar?

No

> Is there a way to show for the scheduling class what level the process has within the scheduling class?

Doesn't the niceness indicate that now?

clausecker:
> Is there a way to show for the scheduling class what level the process has within the scheduling class?

I think niceness and priority class are orthogonal process.  E.g. I can use `idprio 5` to put a process into idle priority class 5, which means it only runs if no process of lower idle priority class is runnable.

freebsd/FreeBSDProcessTable.c
```diff
@@ -250,12 +250,39 @@ void ProcessTable_goThroughEntries(ProcessTable* super) {
...
       if (String_eq("intr", kproc->ki_comm) && (kproc->ki_flag & P_SYSTEM)) {
          proc->nice = 0; //@etosan: intr kernel process (not thread) has weird nice value
-      } else if (kproc->ki_pri.pri_class == PRI_TIMESHARE) {
-         proc->nice = kproc->ki_nice - NZERO;
-      } else if (PRI_IS_REALTIME(kproc->ki_pri.pri_class)) {
-         proc->nice = PRIO_MIN - 1 - (PRI_MAX_REALTIME - kproc->ki_pri.pri_level);
-      } else {
-         proc->nice = PRIO_MAX + 1 + kproc->ki_pri.pri_level - PRI_MIN_IDLE;
+         fp->sched_class = ' ';
+      } else switch (PRI_BASE(kproc->ki_pri.pri_class)) {
```
BenBE, line 254: Explicit block for the `else` …

BenBE (changes requested):
Where does this touch the displayed `NICE` value?

herrhotzenplotz:
> Where does this touch the displayed `NICE` value?

I don't quite understand what you mean by that. It computes the (now correct) niceness in the switch statement now.

freebsd/FreeBSDProcessTable.c
herrhotzenplotz: Fixed

BenBE:
> > Where does this touch the displayed `NICE` value?
> 
> I don't quite understand what you mean by that. It computes the (now correct) niceness in the switch statement now.

Missed that when I had a first quick look at the code. Did some further changes. Apart from this LGTM; didn't test though. Maybe @fraggerfox + @clausecker can take a look&test?

IIRC we already have a more or less "global" column for the priority class that could/should be re-used here. Also, is that value different from the one shown by `SCHEDULERPOLICY`?

@clausecker For the processes to be shown in blue for the CPU time, they need to have "lower than normal" priority IIRC. If they previously didn't due to this bug, they now should be accounted accordingly.

clausecker:
@BenBE On FreeBSD, everything in the "idle priority" scheduler class has "lower than normal" priority.  Maybe new colours for "idle priority" and "real time priority" could be helpful instead.

BenBE:
> @BenBE On FreeBSD, everything in the "idle priority" scheduler class has "lower than normal" priority. Maybe new colours for "idle priority" and "real time priority" could be helpful instead.

Okay, this would thus require a second change (commit) to update the accounting on the CPU meter. Will need to look up how that code works exactly though …

herrhotzenplotz:
> IIRC we already have a more or less "global" column for the priority class that could/should be re-used here. Also, is that value different from the one shown by `SCHEDULERPOLICY`?

Yes that value is different.

BenBE:
> > @BenBE On FreeBSD, everything in the "idle priority" scheduler class has "lower than normal" priority. Maybe new colours for "idle priority" and "real time priority" could be helpful instead.
> 
> Okay, this would thus require a second change (commit) to update the accounting on the CPU meter. Will need to look up how that code works exactly though …

Okay, quick update:

The CPU meter takes the `NICE` time for detailled CPU usage statistics from the values set in [`Platform_setCPUValues`](https://github.com/htop-dev/htop/blob/6f142695cbf5281dd2a5f67b7233d6b93111fa99/freebsd/Platform.c#L205). These are set via [FreeBSDMachine_scanCPU](https://github.com/htop-dev/htop/blob/6f142695cbf5281dd2a5f67b7233d6b93111fa99/freebsd/FreeBSDMachine.c#L241-L248) AFAICS.

freebsd/FreeBSDProcessTable.c
```diff
@@ -250,12 +250,44 @@ void ProcessTable_goThroughEntries(ProcessTable* super) {
...
-         proc->nice = kproc->ki_nice - NZERO;
-      } else if (PRI_IS_REALTIME(kproc->ki_pri.pri_class)) {
-         proc->nice = PRIO_MIN - 1 - (PRI_MAX_REALTIME - kproc->ki_pri.pri_level);
+         fp->sched_class = ' ';
       } else {
-         proc->nice = PRIO_MAX + 1 + kproc->ki_pri.pri_level - PRI_MIN_IDLE;
+         switch (PRI_BASE(kproc->ki_pri.pri_class)) {
+            /* Handling of the below is explained in the FreeBSD base system in:
+             *
+             * /usr/src/usr.bin/top/machine.c (fucntion format_nice) */
```
Explorer09, line 258: Typo. ("function")

freebsd/FreeBSDProcess.c
```diff
@@ -109,6 +111,8 @@ static int FreeBSDProcess_compareByKey(const Process* v1, const Process* v2, Pro
       return SPACESHIP_NULLSTR(p1->jname, p2->jname);
    case EMULATION:
       return SPACESHIP_NULLSTR(p1->emul, p2->emul);
+   case SCHEDCLASS:
+      return SPACESHIP_NUMBER(p1->sched_class, p2->sched_class);
```
Explorer09, line 115: Alphabetical sorting is probably not desireble for this SCHEDCLASS field. Better define `sched_class` as an enum and convert the field to character when displaying.

freebsd/FreeBSDProcessTable.c
herrhotzenplotz: Fixed

freebsd/FreeBSDProcess.c
herrhotzenplotz: I like this idea! :+1:

herrhotzenplotz:
schedclass_new
Example - now with a `SCHEDCLASS_` enum.

freebsd/FreeBSDProcess.c
```diff
@@ -73,22 +74,40 @@ void Process_delete(Object* cast) {
...
+};
+
 static void FreeBSDProcess_rowWriteField(const Row* super, RichString* str, ProcessField field) {
    const FreeBSDProcess* fp = (const FreeBSDProcess*) super;
 
    char buffer[256]; buffer[255] = '\0';
    int attr = CRT_colors[DEFAULT_COLOR];
    size_t n = sizeof(buffer) - 1;
 
+
```
Explorer09, line 93: Unnecessary extra newline.

freebsd/FreeBSDProcess.c
```diff
@@ -73,22 +74,40 @@ void Process_delete(Object* cast) {
    free(this);
 }
 
+static_assert(MAX_SCHEDCLASS == 5, "The lookup table below isn't in sync with the SCHEDCLASS_ enum. Please update.");
+static const char FreeBSD_schedclassChars[MAX_SCHEDCLASS] = {
```
Explorer09, line 78: 1. The static assert is probably unnecessary.
2. It's good to give some rationale on how the characters are chosen this way, because this thing is likely added into htop documentation. E.g. is the blank character (`SCHEDCLASS_TIMESHARE`) considered normal for a FreeBSD process, and why is `SCHEDCLASS_ITHD` represented as a hyphen?

freebsd/FreeBSDProcess.h
```diff
@@ -19,6 +19,16 @@ typedef struct FreeBSDProcess_ {
...
+   enum {
+      SCHEDCLASS_UNKNOWN = 0,
+
+      SCHEDCLASS_ITHD, /* interrupt thread */
+      SCHEDCLASS_REALTIME,
+      SCHEDCLASS_TIMESHARE,
+      SCHEDCLASS_IDLE,
+
+      MAX_SCHEDCLASS,
+   } sched_class;
```
Explorer09, line 31: Move the enum definitions into its own type in `FreeBSDProcess.h`

```c
typedef enum {
   SCHEDCLASS_UNKNOWN = 0,
   SCHEDCLASS_ITHD, /* interrupt thread */
   SCHEDCLASS_REALTIME,
   SCHEDCLASS_TIMESHARE,
   SCHEDCLASS_IDLE,
   MAX_SCHEDCLASS,
} FreeBSDSchedClass;
```

freebsd/FreeBSDProcess.c
herrhotzenplotz: I've put in the static assert for a few reasons:
- leave a compiler hint whenever someone changes the schedclass enum
- and thus prevent access to uninitialised data when such changes happen

Yes, it is unlikely this enum will ever change - this is just being less hostile to whoever works on this code and to prevent potential bugs.

Explorer09: @herrhotzenplotz There are other mapping tables from an enum in htop codebase like this, and none of them needs a static assert. It's unlikely you need one here either.
When an array is initialized with a "named index" initializer like this, unspecified indices would have zero value. If you need an assert, do so on the mapping side so that the mapped character isn't `'\0'`.

herrhotzenplotz: > There are other mapping tables from an enum in htop codebase like this, and none of them needs a static assert.

Well I'd say this would be a problem with those "other mapping tables". Having the static_assert in place does no harm and only improves things. I don't see why anyone would willingly remove this simple assurance.

> If you need an assert, do so on the mapping side so that the mapped character isn't '\0'.

I can certainly add this runtime assertion.

Explorer09: > Well I'd say this would be a problem with those "other mapping tables". Having the static_assert in place does no harm and only improves things. I don't see why anyone would willingly remove this simple assurance.

First, it's not my decision, I only request this for the sake of code consistency. Second, having

Explorer09: > Well I'd say this would be a problem with those "other mapping tables". Having the static_assert in place does no harm and only improves things. I don't see why anyone would willingly remove this simple assurance.

First, it's not my decision, I only request this for the sake of code consistency. Second, having an assertion of `MAX_SCHEDCLASS == 5` isn't helpful and it partly defeats the purpose of defining the `MAX_SCHEDCLASS` enum value (which is meant to increase over time).

herrhotzenplotz: > First, it's not my decision, I only request this for the sake of code consistency.

Sacrificing being prone to bugs for code consistency seems like a very bad thing to me.

> Second, having an assertion of MAX_SCHEDCLASS == 5 isn't helpful

Yes it is helpful. The point is to stop compilation, let the developer update the lookup table, correct the static assertion and continue. This is what the assertion message tells. This is a pretty common way to guide someone making changes in large C codebases.

>  and it partly defeats the purpose of defining the MAX_SCHEDCLASS enum value (which is meant to increase over time).

The reason to define `MAX_SCHEDCLASS` is to define the static array of the correct size and to range-check incoming enum values.

Explorer09: @herrhotzenplotz Would you raise this issue to a separate discussion?
IMO, a long term solution is for compiler to introduce an array/struct attribute to reject any implicit initialization of array/struct members.

herrhotzenplotz: Sure!

BenBE: While I personally don't have any objections against this `static_assert` there are two reasons against this in this particular case:
- The `htop` source is targeted to be compile-able by any compliant C99 compiler. As [`static_assert`](https://en.cppreference.com/w/c/language/_Static_assert) was not introduced before C11 (as `_Static_assert`) or C23 proper this violates the language standard we are aiming for.
- The rest of the code base merely assumes for indexed initializers that if there's a `MAX_FOO` or `NUM_FOO` constant in an enum it's the last one (i.e. the one with the highest value). Thus having entries defined implicitly below that maximum value is totally fine and expected. The most prominent example is the set of internal IDs for columns.

If you really want to check, I'm totally fine with the runtime check as suggested by @Explorer09.

freebsd/FreeBSDProcess.h
```diff
@@ -13,12 +13,23 @@ in the source distribution for its full text.
 #include "Process.h"
 #include "Machine.h"
 
+typedef enum {
+   SCHEDCLASS_UNKNOWN = 0,
+
+   SCHEDCLASS_ITHD, /* interrupt thread */
```
BenBE, line 19: Not the most expressive name. Unless forced by some reference documentation I'd prefer something like `SCHEDCLASS_IRQTHREAD`.

BenBE (changes requested):

freebsd/FreeBSDProcess.h
Explorer09: @BenBE The original constant is named `PRI_ITHD`, but I do agree it can be named better here in htop. My personal suggestion is `SCHEDCLASS_INTR_THREAD`.

freebsd/FreeBSDProcess.c
herrhotzenplotz: @BenBE I am fully aware of the fact that C99 does *not* have `static_assert`. However, your code base has got compatibility code: https://github.com/htop-dev/htop/blob/4c0e96592d5f5f884aab9b29a3300ae5e1b6009b/Compat.h#L71-L73

In fact, this static_assert macro was used during my testing.

freebsd/FreeBSDProcessTable.c
```diff
@@ -248,14 +248,46 @@ void ProcessTable_goThroughEntries(ProcessTable* super) {
...
-         proc->nice = 0; //@etosan: intr kernel process (not thread) has weird nice value
-      } else if (kproc->ki_pri.pri_class == PRI_TIMESHARE) {
-         proc->nice = kproc->ki_nice - NZERO;
-      } else if (PRI_IS_REALTIME(kproc->ki_pri.pri_class)) {
-         proc->nice = PRIO_MIN - 1 - (PRI_MAX_REALTIME - kproc->ki_pri.pri_level);
-      } else {
-         proc->nice = PRIO_MAX + 1 + kproc->ki_pri.pri_level - PRI_MIN_IDLE;
+      switch (PRI_BASE(kproc->ki_pri.pri_class)) {
+         /* Handling of the below is explained in the FreeBSD base system in:
+          *
```
Explorer09, line 253: This blank line might be left by accident.

freebsd/FreeBSDProcess.c
```diff
@@ -73,22 +74,41 @@ void Process_delete(Object* cast) {
...
    case JAIL:
       Row_printLeftAlignedField(str, attr, fp->jname ? fp->jname : "N/A", 11);
       return;
+
    case EMULATION:
       Row_printLeftAlignedField(str, attr, fp->emul ? fp->emul : "N/A", 16);
       return;
+
+   case SCHEDCLASS:
+      assert(SCHEDCLASS_UNKNOWN <= fp->sched_class && fp->sched_class < MAX_SCHEDCLASS);
```
Explorer09, line 106:
```suggestion
      assert(0 <= fp->sched_class && fp->sched_class < ARRAYSIZE(FreeBSD_schedclassChars));
```

freebsd/FreeBSDProcessTable.c
```diff
@@ -248,14 +248,46 @@ void ProcessTable_goThroughEntries(ProcessTable* super) {
...
+            break;
+
+         case PRI_TIMESHARE:
+            fp->sched_class = SCHEDCLASS_TIMESHARE;
+            proc->nice = kproc->ki_nice - NZERO;
+            break;
+
+         default:
+            fp->sched_class = SCHEDCLASS_UNKNOWN;
+            proc->nice = kproc->ki_nice - NZERO;
```
Explorer09, line 289: I see implementation differences between yours in this PR and the [FreeBSD top source code](https://github.com/freebsd/freebsd-src/blob/main/usr.bin/top/machine.c#L1182) you cited.

Specifically, when the `sched_class` is unknown, FreeBSD `top` simply display "`?`" for the nice value. It doesn't use the `kproc->ki_nice - NZERO` as if it were a regular ("time sharing") process.

Is this behavior difference intentional?

Explorer09:
While I have no more problem with the code qualities of this PR, there is one potential issue that I am not sure if it's worth addressing.

Since it becomes apparent that FreeBSD doesn't use a simple integer nice value for process priority, the sorting of `NICE` field in htop should no longer be simple integer sorting.

Then I saw this "FreeBSD process priority to nice value" conversation table that I can't tell of it's accurate (from here https://git.quacker.org/d/freebsd-nq/commit/de916c8b747ae5d875788555f6d39adb3fc91e4a):

```c
	    /*
	     * normal time      -> nice value -20 - +20
	     * real time 0 - 31 -> nice value -52 - -21
	     * idle time 0 - 31 -> nice value +21 - +52
	     */
```

For example, "real time" 31 and "idle time" 31 should definitely not be ordered close together when sorted by `NICE`. Similarly, "normal time" 20 and "real time" 20 are definitely different.

freebsd/FreeBSDProcessTable.c
herrhotzenplotz: I don't see how to represent an unknown nice value since `proc->nice` is an integer in htop. Setting it to zero would be wrong.

freebsd/FreeBSDProcessTable.c
herrhotzenplotz: Not really, it just looks a bit crammed together without the newline. I will remove it.

herrhotzenplotz:
> Since it becomes apparent that FreeBSD doesn't use a simple integer nice value for process priority, the sorting of `NICE` field in htop should no longer be simple integer sorting.

That is a good point.

> Then I saw this "FreeBSD process priority to nice value" conversation table that I can't tell of it's accurate (from here https://git.quacker.org/d/freebsd-nq/commit/de916c8b747ae5d875788555f6d39adb3fc91e4a):

This link does not seem to be an official src mirror and it also references code from 2006 which is very much not up-to-date. The current top code looks very different (e.g.: https://codeberg.org/FreeBSD/freebsd-src/src/branch/main/usr.bin/top/machine.c)

Explorer09:
@herrhotzenplotz 

> This link does not seem to be an official src mirror and it also references code from 2006 which is very much not up-to-date. The current top code looks very different (e.g.: https://codeberg.org/FreeBSD/freebsd-src/src/branch/main/usr.bin/top/machine.c)

Good reference. I have two questions to ask, if you can help answering those for me.

From what I've seen, there are `PRI` and `NICE` fields in FreeBSD `top`; and they are formatted like the following
```c
                sbuf_printf(procbuf, "%3d ", pp->ki_pri.pri_level - PZERO);
		sbuf_printf(procbuf, "%4s", format_nice(pp));
```
According to the `compare_prio()` function in the lower part of the source code, only the `PRI` field is used for sorting purpose, and the `NICE` field never used for sorting.
1. Is the `PRI` field ever used in the FreeBSD code of htop (this project)?
2. Is any of the priority fields modifiable in `top`, by users or administrators? I wish to have an idea on how "`Nice-`" and "`Nice+`" commands in htop can work there in FreeBSD.

freebsd/FreeBSDProcessTable.c
BenBE: If you have a sentinel value would it be sensible to display `N/A` in that case?

herrhotzenplotz: I guess that makes sense.

herrhotzenplotz:
> @herrhotzenplotz
> 
> > This link does not seem to be an official src mirror and it also references code from 2006 which is very much not up-to-date. The current top code looks very different (e.g.: https://codeberg.org/FreeBSD/freebsd-src/src/branch/main/usr.bin/top/machine.c)
> 
> Good reference. I have two questions to ask, if you can help answering those for me.
> 
> From what I've seen, there are `PRI` and `NICE` fields in FreeBSD `top`; and they are formatted like the following
> 
> ```c
>                 sbuf_printf(procbuf, "%3d ", pp->ki_pri.pri_level - PZERO);
> 		sbuf_printf(procbuf, "%4s", format_nice(pp));
> ```
> 
> According to the `compare_prio()` function in the lower part of the source code, only the `PRI` field is used for sorting purpose, and the `NICE` field never used for sorting.
> 
>     1. Is the `PRI` field ever used in the FreeBSD code of htop (this project)?

Yes, the `proc->priority` field is exactly this value. In fact if you want to sort by "real scheduling priority" you can sort by this value.

>     2. Is any of the priority fields modifiable in `top`, by users or administrators? I wish to have an idea on how "`Nice-`" and "`Nice+`" commands in htop can work there in FreeBSD.

It is possible, though I typically use the `renice`, `rtprio` and `idprio` commands to accomplish this: https://man.freebsd.org/cgi/man.cgi?rtprio . I don't think you can alter the priority of idletime or realtime scheduled processes from within top (please correct me if I am wrong).

Also: calling renice on an idletime- or realtime scheduled process has no effect.

Relevant syscalls are: 
- [rtprio(2)](https://man.freebsd.org/cgi/man.cgi?query=rtprio&sektion=2&apropos=0&manpath=FreeBSD+14.0-RELEASE+and+Ports)
- [(get|set)priority(2)](https://man.freebsd.org/cgi/man.cgi?query=setpriority&sektion=2&apropos=0&manpath=FreeBSD+14.0-RELEASE+and+Ports)

herrhotzenplotz:
@BenBE @Explorer09 any news on this?

BenBE:
> @BenBE @Explorer09 any news on this?

There are still two code comments open and need fixing. The first one is about the static assert and the table indices, which should be handled similar to how it is done in other places in the htop code; second there's the handling of niceness values and sorting AFAICS. Lastly, there's the issue of rebasing the code to be up-to-date with the latest main branch. Once these have been resolved, leave a quick comment and I'll take a closer look at merging. As I read the discussions at the moment, there's still some aspects open. Not sure if @Explorer09 has further notes to add.

freebsd/FreeBSDProcess.c
BenBE: Could you update this please to align with the rest of the code base?

freebsd/FreeBSDProcessTable.c
BenBE: Anything missing here? Any code changes outstanding?

BenBE (commented):
As I was contacted off-site re this PR I'll mark again the two places where I think things seem not ready yet. Please update the PR accordingly or provide appropriate feedback.

freebsd/FreeBSDProcess.c
herrhotzenplotz: > Could you update this please to align with the rest of the code base?

I have done this weeks ago and have no idea why github still shows the static assertion here. It's gone in the full diff and I have converted them to runtime assertions instead as you suggested.

freebsd/FreeBSDProcessTable.c
herrhotzenplotz: I have added a sentinel value for unknown process niceness.

Process.h
```diff
@@ -160,7 +161,10 @@ typedef struct Process_ {
    /* Scheduling priority */
    long int priority;
 
-   /* Nice value */
+   /* Nice value
+    *
+    * PROCESS_NICE_UNKNOWN is a sentinel value for unknown niceness */
+#define PROCESS_NICE_UNKNOWN (-LONG_MAX)
```
BenBE, line 167: Minor style thing: I think we should avoid having `#define`s intermixed with `struct` declarations. I see two possible places where this could go: Near the top before the define for `DEFAULT_HIGHLIGHT_SECS` or down below this structure; I'd probably prefer the first location.

BenBE (approved):
One minor change left, the remainder of the PR LGTM.

PR #1472, opened 2024-05-06
Linux: show newlines in commands as ? instead of space

cgzones:
Since `\n` is used internally by htop to split command lines, replace `\n` with `\r` in command lines to not display them as space.

Merging the command string will not work, but newlines in commands should be rather the exception.

### DIFF
--- a/linux/LinuxProcessTable.c
+++ b/linux/LinuxProcessTable.c
@@ -1172,6 +1172,12 @@ static bool LinuxProcessTable_readCmdlineFile(Process* process, openat_arg_t pro
 
       /* newline used as delimiter - when forming the mergedCommand, newline is
        * converted to space by Process_makeCommandStr */
+      if (argChar == '\n') {
+         /* Set to some other non-printable character */
+         command[i] = '\r';
+         continue;
+      }
+
       if (argChar == '\0') {
          command[i] = '\n';
 


linux/LinuxProcessTable.c
```diff
@@ -1172,6 +1172,12 @@ static bool LinuxProcessTable_readCmdlineFile(Process* process, openat_arg_t pro
 
       /* newline used as delimiter - when forming the mergedCommand, newline is
        * converted to space by Process_makeCommandStr */
+      if (argChar == '\n') {
+         /* Set to some other non-pritnable character */
```
Explorer09, line 1176: Typo.

Explorer09:
Why choose `\r` in particular? Is this meant to be a graphical representation of the LF?
I wonder if U+240A is useful at the moment.

BenBE:
The problem with U+240A at this particular code location is, that the string is still in UTF-8 at this location and thus doing this replacement would require additional memory to be allocated.

Explorer09:
> The problem with U+240A at this particular code location is, that the string is still in UTF-8 at this location and thus doing this replacement would require additional memory to be allocated.

I just wish this problem to have a long term solution. There should be ideally be a uniform way to sanitize and escape the characters from the `/proc/<pid>/cmdline` inputs.

BenBE:
We could allocate the `cmdline` buffer twice: Once for the original and a second copy for display with special characters replaced by their visual replacements. Though we'd need to introduce some markup in there to distinguish a normal U+240A in the command line vs. the replacement U+240A from displaying the LF. Any ideas? Opinions?

cgzones:
This fix is just so that newlines in commands are not displayed as spaces, but as unknown characters (in `RichString_writeFrom(Wide|Ascii)()`).

PR #1473, opened 2024-05-06
Perf

cgzones:
* Annotate mbstowcs_nonfatal() with restrict

  The arguments to our mbrtowc(3) wrapper should not alias, since they also must not for mbrtowc(3).

  Might help some compilers optimizing the code.

* Linux: use fast integer parsing for /proc/<pid>/stat

  /proc/<pid>/stat is parsed for every displayed process on every iteration containing several numbers to parse.
  Use out own implementations since strto(u)l(3) came up during profiling.

  @BenBE what was the reason for adding the parameter `maxlen` to those functions?

### DIFF
--- a/linux/LinuxProcessTable.c
+++ b/linux/LinuxProcessTable.c
@@ -79,10 +79,10 @@ static FILE* fopenat(openat_arg_t openatArg, const char* pathname, const char* m
 }
 
 static inline uint64_t fast_strtoull_dec(char** str, int maxlen) {
-   register uint64_t result = 0;
+   uint64_t result = 0;
 
    if (!maxlen)
-      --maxlen;
+      maxlen = 20; // length of maximum value of 18446744073709551615
 
    while (maxlen-- && **str >= '0' && **str <= '9') {
       result *= 10;
@@ -91,6 +91,34 @@ static inline uint64_t fast_strtoull_dec(char** str, int maxlen) {
    }
 
    return result;
+}
+
+static long long fast_strtoll_dec(char** str, int maxlen) {
+   bool neg = false;
+
+   if (**str == '-') {
+      neg = true;
+      (*str)++;
+   }
+
+   unsigned long long res = fast_strtoull_dec(str, maxlen);
+   assert(res <= LLONG_MAX);
+   long long result = (long long)res;
+
+   return neg ? -result : result;
+}
+
+static long fast_strtol_dec(char** str, int maxlen) {
+   long long result = fast_strtoll_dec(str, maxlen);
+   assert(result <= LONG_MAX);
+   assert(result >= LONG_MIN);
+   return (long)result;
+}
+
+static unsigned long fast_strtoul_dec(char** str, int maxlen) {
+   unsigned long long result = fast_strtoull_dec(str, maxlen);
+   assert(result <= ULONG_MAX);
+   return (unsigned long)result;
 }
 
 static inline uint64_t fast_strtoull_hex(char** str, int maxlen) {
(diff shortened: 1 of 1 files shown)


BenBE:
The reason for the `maxlen` argument had mostly two aspects:
- Defense in depth regarding avoiding buffer overruns
- Unless the input is all zero, every string above a certain length will overflow.

linux/LinuxProcessTable.c
```diff
@@ -78,11 +78,68 @@ static FILE* fopenat(openat_arg_t openatArg, const char* pathname, const char* m
    return fp;
 }
 
+static long fast_strtol_dec(char** str, int maxlen) {
```
Explorer09, line 81: How about just calling fast_stroull_dec() and downcast the result for this one?
```c
static long fast_strtol_dec(char** str, int maxlen) {
   long long result = fast_strtoll_dec(str, maxlen);
   assert(result <= LONG_MAX);
   assert(result >= LONG_MIN);
   return (long)result;
}
```

linux/LinuxProcessTable.c
```diff
@@ -78,11 +78,68 @@ static FILE* fopenat(openat_arg_t openatArg, const char* pathname, const char* m
...
+   }
+
+   if (!maxlen)
+      maxlen = 20; // −9223372036854775808
+
+   while (maxlen-- && **str >= '0' && **str <= '9') {
+      result *= 10;
+      result += **str - '0';
+      (*str)++;
+   }
```
Explorer09, line 133: This "while" loop can be converted to a `fast_strtoull_dec(str, maxlen)` and then cast the result from unsigned to signed.

linux/LinuxProcessTable.c
```diff
@@ -78,11 +78,68 @@ static FILE* fopenat(openat_arg_t openatArg, const char* pathname, const char* m
...
+
+   return neg ? -result : result;
+}
+
 static inline uint64_t fast_strtoull_dec(char** str, int maxlen) {
    register uint64_t result = 0;
 
    if (!maxlen)
-      --maxlen;
+      maxlen = 20; // 18446744073709551615
```
BenBE, line 142: Documenting the previous default makes no sense for this line. If at all we want documentation it would explain, that this is "assume max reasonable length if none given". I'm for dropping this comment though …

linux/LinuxProcessTable.c
```diff
@@ -78,11 +78,68 @@ static FILE* fopenat(openat_arg_t openatArg, const char* pathname, const char* m
...
+   long result = 0;
+   bool neg = false;
+
+   if (**str == '-') {
+      neg = true;
+      (*str)++;
+   }
+
+   if (!maxlen)
+      maxlen = 20; // −9223372036854775808
```
BenBE, line 91: If at all, document the why not the magic value …

And the why is basically "because C does not have default arguments" …

linux/LinuxProcessTable.c
```diff
@@ -78,11 +78,68 @@ static FILE* fopenat(openat_arg_t openatArg, const char* pathname, const char* m
...
+   long long result = 0;
+   bool neg = false;
+
+   if (**str == '-') {
+      neg = true;
+      (*str)++;
+   }
+
+   if (!maxlen)
+      maxlen = 20; // −9223372036854775808
```
BenBE, line 127: Similar here: Drop that comment.

linux/LinuxProcessTable.c
```diff
@@ -78,11 +78,68 @@ static FILE* fopenat(openat_arg_t openatArg, const char* pathname, const char* m
...
+   }
+
+   return neg ? -result : result;
+}
+
+static unsigned long fast_strtoul_dec(char** str, int maxlen) {
+   unsigned long result = 0;
+
+   if (!maxlen)
+      maxlen = 20; // 18446744073709551615
```
BenBE, line 106: Same here: Drop this comment …

linux/LinuxProcessTable.c
```diff
@@ -78,11 +78,68 @@ static FILE* fopenat(openat_arg_t openatArg, const char* pathname, const char* m
...
+
+   while (maxlen-- && **str >= '0' && **str <= '9') {
+      result *= 10;
+      result += **str - '0';
+      (*str)++;
+   }
```
BenBE, line 112: Same replacement of this `while` loop could be done here too …

linux/LinuxProcessTable.c
Explorer09: Just FYI, the standard `strtoull()` permits almost unlimited number of leading zeros in the string. While it won't certainly happen in parsing any Linux `procfs` file, having an arbitrary length limit in a `strtoull()` implementation would technically violate ISO C.

BenBE: The issue here is less about ISO C compliance, but more about squeeze out cycles. `fast_strtoull` is not meant to comply to ISO C, but to parse the data we actually encounter in realistic situations. Otherwise we'd have to support locales and a full load of other stuff (different radix prefixes) we don't need …

linux/LinuxProcessTable.c
```diff
@@ -93,6 +93,34 @@ static inline uint64_t fast_strtoull_dec(char** str, int maxlen) {
...
+static long long fast_strtoll_dec(char** str, int maxlen) {
+   bool neg = false;
+
+   if (**str == '-') {
+      neg = true;
+      (*str)++;
+   }
+
+   unsigned long long res = fast_strtoull_dec(str, maxlen);
+   assert(res <= LONG_LONG_MAX);
```
Explorer09, line 105:
```suggestion
   assert(res <= LLONG_MAX);
```

And I'm not sure what you are going to do if the input is exactly `-9223372036854775808`, since this input become an undefined behavior in the standard's sense.

BenBE: This won't matter, as this function would `assert` in that case.

PR #1475, opened 2024-05-07
Update configure.ac to include a check for KEY_RESIZE

### LINKED ISSUE #1474: htop fails to compile with gcc-14.1 unless ncurses has been built with --enable-sigwinch
The below line causes a failure to compile htop with gcc-14.1 when ncurses has not been built with --enable-sigwinch.
https://github.com/htop-dev/htop/blob/4b3dfa246e9fa0fe3ce7d97e7a37239251ada42b/ScreenManager.c#L361-L365

```
../ScreenManager.c: In function 'ScreenManager_run':
../ScreenManager.c:361:12: error: 'KEY_RESIZE' undeclared (first use in this function); did you mean 'KEY_RESUME'?
  361 |       case KEY_RESIZE:
      |            ^~~~~~~~~~
      |            KEY_RESUME
../ScreenManager.c:361:12: note: each undeclared identifier is reported only once for each function it appears in
```

Ref:

```
/*
 * KEY_RESIZE is an extended feature that relies upon the SIGWINCH handler
 * in ncurses.
 */
#if 1
#ifndef NCURSES_SIGWINCH
#define NCURSES_SIGWINCH 1
#endif
#else
#undef NCURSES_SIGWINCH
#define NCURSES_SIGWINCH 0
#endif

...

#if NCURSES_SIGWINCH
#define KEY_RESIZE      0632            /* Terminal resize event */
#endif
```

fasterit:
i.e. ncurses(w) compiled with the default --enable-sigwinch Closes #1474. Thank you very much Rudi Heitbaum.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -448,6 +448,11 @@ else
    AC_SEARCH_LIBS([keypad], [tinfo])
 fi
 
+# test for KEY_RESIZE being defined in one of the (n)curses header files
+AC_CHECK_DECLS([KEY_RESIZE], [],
+   [AC_MSG_ERROR([can not find required ncurses KEY_RESIZE definition, compile ncurses with --enable-sigwinch])],
+      [#include "ProvideCurses.h"])
+
 if test "$enable_static" = yes; then
    AC_SEARCH_LIBS([Gpm_GetEvent], [gpm])
 fi


fasterit:
ok, so that fails as we need `config.h` to make `ProvideCurses.h` useful.
Not really looking forward to duplicate the whole logic :cry:

PR #1482, opened 2024-05-18
Embed ELF metadata about dynamic runtime dependencies

cgzones:
Embed metadata about runtime dependencies of dynamic libraries used via dlopen(3) in the binary so that distributions can automatically generate package dependencies.

Inspired-by: https://github.com/systemd/systemd/pull/32234


cgzones:
See https://salsa.debian.org/debian/htop/-/merge_requests/5 for the packaging update.

Macros.h
```diff
@@ -142,4 +151,32 @@ static inline unsigned long long saturatingSub(unsigned long long a, unsigned lo
...
+         .n_namesz = sizeof(ELF_NOTE_DLOPEN_OWNER),                                                \
+         .n_descsz = sizeof(content),                                                              \
+         .n_type   = ELF_NOTE_DLOPEN_TYPE,                                                         \
+      },                                                                                           \
+      .name = ELF_NOTE_DLOPEN_OWNER,                                                               \
+      .dlopen_content = content,                                                                   \
+   };                                                                                              \
+   IGNORE_W11EXTENSIONS_END
+#else
+#define DECLARE_ELF_NOTE_DLOPEN()
```
BenBE, line 179: Should name the `content` parameter for consistency.

Macros.h
```diff
@@ -142,4 +151,32 @@ static inline unsigned long long saturatingSub(unsigned long long a, unsigned lo
...
 
+#ifdef HAVE_ALIGNAS
+#include <stdalign.h>
+
+#define ELF_NOTE_DLOPEN_OWNER "FDO"
+#define ELF_NOTE_DLOPEN_TYPE  UINT32_C(0x407c0c0a)
+
+#define DECLARE_ELF_NOTE_DLOPEN(content) \
+   IGNORE_W11EXTENSIONS_BEGIN                                                                      \
+   __attribute__((used, section(".note.dlopen"))) alignas(sizeof(uint32_t)) static const struct {  \
```
BenBE, line 162:
```suggestion
   __attribute__((packed, used, section(".note.dlopen"))) static const struct {                    \
```

Just byte-align ourselves and round up with some little magic later

Macros.h
```diff
@@ -142,4 +151,32 @@ static inline unsigned long long saturatingSub(unsigned long long a, unsigned lo
...
+      struct {                                                                                     \
+         uint32_t n_namesz, n_descsz, n_type;                                                      \
+      } nhdr;                                                                                      \
+      char name[sizeof(ELF_NOTE_DLOPEN_OWNER)];                                                    \
```
BenBE, line 166:
```suggestion
      char name[(sizeof(ELF_NOTE_DLOPEN_OWNER)+3)&(~3)];                                           \
```

Macros.h
```diff
@@ -142,4 +151,32 @@ static inline unsigned long long saturatingSub(unsigned long long a, unsigned lo
...
+      alignas(sizeof(uint32_t)) char dlopen_content[sizeof(content)];                              \
```
BenBE, line 167:
```suggestion
      char dlopen_content[(sizeof(content)+3)&(~3)];                                               \
```

Although one could argue if this last expansion is still required at the end of the attribute section.

configure.ac
```diff
@@ -273,6 +273,18 @@ AC_LINK_IFELSE([
...
+   AC_LANG_SOURCE(
+      [[
+         #include <stdalign.h>
+         alignas(128) char buffer[128];
+      ]]
+   )],
+   AC_DEFINE([HAVE_ALIGNAS], 1, [The alignas C11 feature is supported.])
+   AC_MSG_RESULT(yes),
+   AC_MSG_RESULT(no))
+
```
BenBE, line 287: Cf. `Macros.h` for a way on how to do this without C11 support. If we throw around GCC/Clang compiler extensions, we also could just use `packed`, which at least is even kinda portable to other compilers too …

linux/LibNl.c
```diff
@@ -75,6 +75,12 @@ static void unload_libnl(void) {
 }
 
 static int load_libnl(void) {
+
+   DECLARE_ELF_NOTE_DLOPEN("[{\"soname\":[\"libnl-3.so\",\"libnl-3.so.200\"],\"feature\":\"delay-accounting\"," \
+      "\"description:\":\"Enables delay accounting support\",\"priority\":\"recommended\"},{\"soname\":[" \
+      "\"libnl-genl-3.so\",\"libnl-genl-3.so.200\"],\"feature\":\"delay-accounting\",\"description:\":"\
+      "\"Enables delay accounting support\",\"priority\":\"recommended\"}]")
```
BenBE, line 82: Any objections to have this literal span multiple lines for better maintainability?

Please add proper meta-commands for `astyle` to leave this block alone afterwards.

Also, this note probably should only contain the optional dependencies we actually have enabled; i.e. if configured without delay accounting, there's no need to declare this as an optional runtime dependency if we would never try to load that library anyway.

```suggestion
   DECLARE_ELF_NOTE_DLOPEN( "["\
      "{"\
         "\"soname\":[\"libnl-3.so\",\"libnl-3.so.200\"],"\
         "\"feature\":\"delay-accounting\"," \
         "\"description:\":\"Enables delay accounting support\","\
         "\"priority\":\"recommended\""\
      "},"\
      "{"\
         "\"soname\":[\"libnl-genl-3.so\",\"libnl-genl-3.so.200\"],"\
         "\"feature\":\"delay-accounting\","\
         "\"description:\":\"Enables delay accounting support\","\
         "\"priority\":\"recommended\""\
      "}"\
   "]" )
```

Macros.h
```diff
@@ -142,4 +151,32 @@ static inline unsigned long long saturatingSub(unsigned long long a, unsigned lo
...
+   } variable_name = {                                                                             \
```
BenBE, line 168: Shouldn't be `variable_name` be dynamic?

linux/LibSensors.c
```diff
@@ -56,6 +56,9 @@ int LibSensors_init(void) {
 
 #else
 
+   DECLARE_ELF_NOTE_DLOPEN("[{\"soname\":[\"libsensors.so\",\"libsensors.so.5\",\"libsensors.so.4\"],\"feature\":"\
+      "\"sensors\",\"description:\":\"Enables hardware sensor support\",\"priority\":\"recommended\"}]")
```
BenBE, line 60: Multiple lines, example above …

linux/SystemdMeter.c
```diff
@@ -98,6 +98,9 @@ static void SystemdMeter_done(ATTR_UNUSED Meter* this) {
 static int updateViaLib(bool user) {
    SystemdMeterContext_t* ctx = user ? &ctx_user : &ctx_system;
 #ifndef BUILD_STATIC
+   DECLARE_ELF_NOTE_DLOPEN("[{\"soname\":[\"libsystemd.so.0\"],\"feature\":\"systemd\",\"description:\":"\
+      "\"Enables systemd support\",\"priority\":\"suggested\"}]")
```
BenBE, line 102: Multiple lines, example above …

Macros.h
Explorer09: Another problem: C99 doesn't support anonymous struct yet.

linux/LibNl.c
Explorer09: I'm not a fan of having a full JSON syntax in a macro caller. Can we rework the macros so that only important fields will be uses as parameters?

I expect the usage would ideally work like this:
```c
#define ELF_NOTE_DLOPEN_DEP(sonames, feature_tag, priority) ...
// Example
ELF_NOTE_DLOPEN_DEP(
   "[\"libnl-3.so\",\"libnl-3.so.200\"]", "delay-accounting", "recommended")
ELF_NOTE_DLOPEN_DEP(
   "[\"libnl-genl-3.so\",\"libnl-genl-3.so.200\"]", "delay-accounting", "recommended")
```

How do you think?

BenBE: While I also thought about this, including making generating the JSON be split into different macro calls (definitively possible AFAICS), there's one minor issue with the pre-processor IMHO, that prevents this from being really nice syntax. The issue is, that you can't properly auto-terminate lists of arguments to insert the semicolon in list contexts, thus you'd most likely have to resort to something like `"[" ELF_NOTE_DEP(foo…) "," ELF_NOTE_DEP(bar…) "]"`; which doesn't look too terrible, but in a really great world could work without the explicit "," token in between.

Also from a practcal PoV, I'd prefer such macros to go ELF_NOTE_DLOPEN_DEP(freature, level, description, solist)`, but comparibly that's a minor change to implement.

Macros.h
BenBE: > Another problem: C99 doesn't support anonymous struct yet.

The workaround for that is kinda easy though, too …

BenBE:
Any objections to having the code for generating these annotations be its own source code file?

NB: There's also `__attribute__((aligned(4))))` available to play around, which seems to have been supported by GCC for a long time (no idea re Clang) …

Explorer09:
Just my general comments:
I am personally not a fan of the specification of embedding the dlopen dependency information in ELF binaries. For me it look like increasing the binary size with little benefit, and the same information can be provided by (distros) package managers already.

Besides, this information is ELF-specific, which could potentially discriminate other OSes such as BSDs.
I wonder if the same information can be provided through a separate file, not embedding into binaries? The manifest file for Windows executable binaries can work either way - provided separately with ".manifest" extension, or embedding into resource sections of ".exe" files.

PR #1483, opened 2024-05-20
PCP: Resolve nested derived metric definitions for devices

natoscott:
Previously these expressions were (incorrectly) silently failing to evaluate because a derived metric expression was geing nested which isn't valid.  Now libpcp generates several warnings like: "Semantic error: derived metric htop.screen.disks.read_await: operand disk.dev.r_await: Illegal nested derived metric"

Fortunately they're all simple expressions so just unwinding the nesting is the simplest fix.  Add in help text for good measure.


BenBE (approved):
LGTM.

PR #1486, opened 2024-05-30
linux: support for monitoring syscall

JohnSanpe:


linux/LinuxProcess.c
```diff
@@ -112,6 +112,7 @@ const ProcessFieldData Process_fields[LAST_PROCESSFIELD] = {
 #endif
    [GPU_TIME] = { .name = "GPU_TIME", .title = "GPU_TIME ", .description = "Total GPU time", .flags = PROCESS_FLAG_LINUX_GPU, .defaultSortDesc = true, },
    [GPU_PERCENT] = { .name = "GPU_PERCENT", .title = " GPU% ", .description = "Percentage of the GPU time the process used in the last sampling", .flags = PROCESS_FLAG_LINUX_GPU, .defaultSortDesc = true, },
+   [SYSCALL] = { .name ="SYSCALL", .title = "SYSCALL", .description = "Current syscall of the process", .flags = PROCESS_FLAG_LINUX_SYSCALL, .autoWidth = true},
```
natoscott, line 115: very minor whitespace botch after 'name =' here.

JohnSanpe: Fixed!

linux/LinuxProcess.h
```diff
@@ -118,6 +119,8 @@ typedef struct LinuxProcess_ {
    /* Autogroup scheduling (CFS) information */
    long int autogroup_id;
    int autogroup_nice;
+
+   char *syscall;
```
cgzones, line 123: This needs to be free'd in `linux/LinuxProcess.c:Process_delete()`.

Also please align pointers to the left.

linux/LinuxProcessTable.c
```diff
@@ -1098,6 +1098,33 @@ static void LinuxProcessTable_readCwd(LinuxProcess* process, openat_arg_t procFd
    free_and_xStrdup(&process->super.procCwd, pathBuffer);
 }
 
+/*
+ * Read /proc/<pid>/syscall (process-shared data)
```
cgzones, line 1102: I don't think this is process shared data, AFAIC a thread does not stall all threads, see https://lwn.net/Articles/799557/

linux/LinuxProcessTable.c
```diff
@@ -1098,6 +1098,33 @@ static void LinuxProcessTable_readCwd(LinuxProcess* process, openat_arg_t procFd
...
+   }
+
+   buffer[r - 1] = '\0';
+
+   char* arg = strchr(buffer, ' ');
+   if (arg) {
+      *arg = '\0';
+   }
+
+   if (process->syscall && String_eq(buffer, process->syscall))
```
cgzones, line 1122: This optimization is already handled by `free_and_xStrdup()`

linux/LinuxProcessTable.c
```diff
@@ -1098,6 +1098,33 @@ static void LinuxProcessTable_readCwd(LinuxProcess* process, openat_arg_t procFd
...
+   ssize_t r;
+
+   r = xReadfileat(procFd, "syscall", buffer, sizeof(buffer));
+   if (r <= 0) {
+      free(process->syscall);
+      process->syscall = NULL;
+      return;
+   }
+
+   buffer[r - 1] = '\0';
```
cgzones, line 1115: `xReadfileat()` already nul-terminates on success.

linux/LinuxProcess.c
```diff
@@ -362,6 +363,17 @@ static void LinuxProcess_rowWriteField(const Row* super, RichString* str, Proces
          xSnprintf(buffer, n, "N/A  ");
       }
       break;
+   case SYSCALL: {
+      const char* syscall;
+      if (lp->syscall) {
+         syscall = lp->syscall;
```
cgzones, line 369: Would it make sense to translate the number into a name, e.g. via by creating a lookup table at compile time, see https://stackoverflow.com/a/23249568?
Also maybe handle the special value `-1` (*otherwise blocked*)?

cgzones (changes requested):

linux/LinuxProcess.c
JohnSanpe: Firstly, I agree with your opinion, using a lookup table is correct, but maintaining lookup tables for different architectures can be cumbersome (the x64 platform also needs to be backward compatible with x86), so this submission is not prepared. If anyone is interested, they can work for it, otherwise when I have spare time, I'll take care of it myself.

JohnSanpe:
> Any chance this kind of information is available on other platforms too?

Theoretically possible, but currently only intended to support Linux.

linux/LinuxProcessTable.c
```diff
@@ -1098,6 +1098,33 @@ static void LinuxProcessTable_readCwd(LinuxProcess* process, openat_arg_t procFd
    free_and_xStrdup(&process->super.procCwd, pathBuffer);
 }
 
+/*
+ * Read /proc/<pid>/syscall (thread-specific data)
+ */
+static void LinuxProcessTable_readSyscall(LinuxProcess* process, openat_arg_t procFd) {
+   char buffer[PATH_MAX];
```
Explorer09, line 1105: Doesn't make sense to use PATH_MAX here.

linux/LinuxProcessTable.c
```diff
@@ -1098,6 +1098,33 @@ static void LinuxProcessTable_readCwd(LinuxProcess* process, openat_arg_t procFd
...
+   ssize_t r;
+
+   r = xReadfileat(procFd, "syscall", buffer, sizeof(buffer));
```
BenBE, line 1108:
```suggestion
   ssize_t r = xReadfileat(procFd, "syscall", buffer, sizeof(buffer));
```

linux/LinuxProcessTable.c
```diff
@@ -1098,6 +1098,33 @@ static void LinuxProcessTable_readCwd(LinuxProcess* process, openat_arg_t procFd
...
+   if (r <= 0) {
+      free(process->syscall);
+      process->syscall = NULL;
+      return;
+   }
+
+   /* Remove '\n' */
+   buffer[r - 1] = '\0';
+
+   char* arg = strchr(buffer, ' ');
```
BenBE, line 1118: Using `strchrnul` would allow to skip the following check. There's a proper implementation in `XUtils.h` you can use.

linux/LinuxProcessTable.c
```diff
@@ -1098,6 +1098,33 @@ static void LinuxProcessTable_readCwd(LinuxProcess* process, openat_arg_t procFd
...
+
+   r = xReadfileat(procFd, "syscall", buffer, sizeof(buffer));
+   if (r <= 0) {
+      free(process->syscall);
+      process->syscall = NULL;
+      return;
+   }
+
+   /* Remove '\n' */
+   buffer[r - 1] = '\0';
```
BenBE, line 1116: This should use `strchrnul` to remove the `\n` optionally, in case it doesn't exist.

Also no need to trim `\n` if we already split at any other whitespace. What about using `fast_strtoull` and check for proper conversion? As we skip arguments anyway this should suffice and would additionally allow proper display as string (syscall name) later.

linux/LinuxProcessTable.c
JohnSanpe: If you think the kernel may remove `'\n'` in the future, and also for avoid duplicate scanning, I think use `strcspn` is more elegant.

linux/LinuxProcessTable.c
JohnSanpe: Task might not have doing any syscall, which is "running" status in the file, so I think currently we using strings is the most appropriate.

Explorer09: @JohnSanpe You should handle the `running` string as special and assign a placeholder number for it. The reason is it allows us to display `running` string with colors. This status should be special in sorting/collating anyway.

Explorer09:
One more thing. We should allow the user to disable this "syscall" monitoring at compile time time. Since the `/proc/<pid>/syscall` file is only present when the kernel is compiled with `CONFIG_HAVE_ARCH_TRACEHOOK`, it may be useless for users without such kernel. Another reason for having this compile option is we can save code size by not compiling the syscall number-to-name mapping table into htop binary, in case we implement one in the future.

BenBE:
@Explorer09 Or disable the column, when the feature is not available at runtime.

BenBE:
> > @ Explorer09 Or disable the column, when the feature is not available at runtime.
> 
> I wish a compile time option to disable it as well.

Like so?

```autoconf
# Check for syscall feature option
AC_ARG_ENABLE(
   [syscall],
   [
      AS_HELP_STRING(
         [--enable-syscall],
         [Enable 'syscall' feature. Possible values: off, buildtime-check, runtime-check, always-on. Default is 'off'.]
      )
   ],
   [case "${enableval}" in
      off) syscall_feature=0 ;;
      buildtime-check) syscall_feature=1 ;;
      runtime-check) syscall_feature=2 ;;
      always-on) syscall_feature=3 ;;
      *) AC_MSG_ERROR([Invalid value for --enable-syscall]) ;;
   esac],
   [syscall_feature=0]
)

# If buildtime-check is enabled, check if /proc/self/syscall exists and is non-empty
AS_IF(
   [test "x$syscall_feature" = "x1"],
   [
      AC_CHECK_FILE(
         [/proc/self/syscall],
         [syscall_exists=yes],
         [syscall_exists=no]
      )
      AS_IF(
         [test "x$syscall_exists" = "xyes" && test -s /proc/self/syscall],
         [],
         [syscall_feature=0]
      )
   ]
)

# Set the result of the check to a preprocessor directive
AC_DEFINE_UNQUOTED([SYSCALL_FEATURE], [$syscall_feature], [Define to the selected syscall feature state.])
```

Explorer09:
@BenBE 

* I merged your "runtime-check" and "always-on" into one option, because checking the thing at runtime is trivial.
* When cross-compiling, checking `/proc/self/syscall` won't help. It's better just enable it in that case.

```m4
AC_ARG_ENABLE([syscall],
   [AS_HELP_STRING([--enable-syscall],
   [enable 'syscall' monitoring. @<:@default=check@:>@])],
   [],
   [enable_syscall=check]
)

if test "x$enable_syscall" = xcheck; then
   if "$cross_compiling" != no; then
      enable_syscall=yes
   elif test -f /proc/self/syscall && test -s /proc/self/syscall; then
      enable_syscall=yes
   else
      enable_syscall=no
   fi
fi
if test "x$enable_syscall" = xyes; then
   AC_DEFINE([HAVE_SYSCALL], [1], [Define to 1 is syscall monitoring is enabled.])
fi
```

JohnSanpe:
> @BenBE
> 
>     * I merged your "runtime-check" and "always-on" into one option, because checking the thing at runtime is trivial.
> 
>     * When cross-compiling, checking `/proc/self/syscall` won't help. It's better just enable it in that case.
> 
> 
> ```m4
> AC_ARG_ENABLE([syscall],
>    [AS_HELP_STRING([--enable-syscall],
>    [enable 'syscall' monitoring. @<:@default=check@:>@])],
>    [],
>    [enable_syscall=check]
> )
> 
> if test "x$enable_syscall" = xcheck; then
>    if "$cross_compiling" != no; then
>       enable_syscall=yes
>    elif test -f /proc/self/syscall && test -s /proc/self/syscall; then
>       enable_syscall=yes
>    else
>       enable_syscall=no
>    fi
> fi
> if test "x$enable_syscall" = xyes; then
>    AC_DEFINE([HAVE_SYSCALL], [1], [Define to 1 is syscall monitoring is enabled.])
> fi
> ```



> @BenBE
> 
>     * I merged your "runtime-check" and "always-on" into one option, because checking the thing at runtime is trivial.
> 
>     * When cross-compiling, checking `/proc/self/syscall` won't help. It's better just enable it in that case.
> 
> 
> ```m4
> AC_ARG_ENABLE([syscall],
>    [AS_HELP_STRING([--enable-syscall],
>    [enable 'syscall' monitoring. @<:@default=check@:>@])],
>    [],
>    [enable_syscall=check]
> )
> 
> if test "x$enable_syscall" = xcheck; then
>    if "$cross_compiling" != no; then
>       enable_syscall=yes
>    elif test -f /proc/self/syscall && test -s /proc/self/syscall; then
>       enable_syscall=yes
>    else
>       enable_syscall=no
>    fi
> fi
> if test "x$enable_syscall" = xyes; then
>    AC_DEFINE([HAVE_SYSCALL], [1], [Define to 1 is syscall monitoring is enabled.])
> fi
> ```

Very good job, I will merge it

JohnSanpe:
@Explorer09

> @JohnSanpe You should handle the `running` string as special and assign a placeholder number for it. The reason is it allows us to display `running` string with colors. This status should be special in sorting/collating anyway.

Using special placeholder number may conflict with future syscalls, so storing possible states separately may be a better choice? This is also more convenient for us to use lookup tables in the future.

Explorer09:
> > You should handle the `running` string as special and assign a placeholder number for it. The reason is it allows us to display `running` string with colors. This status should be special in sorting/collating anyway.
> 
> Using special placeholder number may conflict with future syscalls, so storing possible states separately may be a better choice? This is also more convenient for us to use lookup tables in the future.

Use a negative number. E.g. `-2`, or maybe `-1000` which has an even less chance to cause conflict.

JohnSanpe:
> > > You should handle the `running` string as special and assign a placeholder number for it. The reason is it allows us to display `running` string with colors. This status should be special in sorting/collating anyway.
> > 
> > 
> > Using special placeholder number may conflict with future syscalls, so storing possible states separately may be a better choice? This is also more convenient for us to use lookup tables in the future.
> 
> Use a negative number. E.g. `-2`, or maybe `-1000` which has an even less chance to cause conflict.

I have already fixed it 😄

linux/LinuxProcess.h
```diff
@@ -31,6 +31,13 @@ in the source distribution for its full text.
 #define PROCESS_FLAG_LINUX_AUTOGROUP 0x00080000
 #define PROCESS_FLAG_LINUX_GPU       0x00100000
 #define PROCESS_FLAG_LINUX_CONTAINER 0x00200000
+#define PROCESS_FLAG_LINUX_SYSCALL   0x00400000
+
+typedef enum SyscallState_ {
+   SYSCALL_STATE_RUNNING,
+   SYSCALL_STATE_NORMAL,
+   SYSCALL_STATE_NA,
+} SyscallState;
```
Explorer09, line 40: I think you are doing something wrongly here. Normally a process won't be waiting for a syscall, so the presence of a syscall number would be more of an exception rather than rule.

I won't say you can't use an enum here. But the better way is to merge this status with the syscall number, so that you don't need two member variables for recording the same thing.

Here is what I mean:

```c
typedef enum LinuxSyscallState_ {
   SYSCALL_STATE_NO_DATA = INT_MIN,
   SYSCALL_STATE_RUNNING = INT_MIN + 1,
   SYSCALL_STATE_BLOCKED = -1
} LinuxSyscallState;
```

By the way, Linux syscall table for x86-64 is in [`syscall_64.tbl`](https://github.com/torvalds/linux/blob/master/arch/x86/entry/syscalls/syscall_64.tbl). The file is located in different directories in Linux source code throughout Linux versions.

JohnSanpe: You should also know that the x32 system call starts from 0x40000000, and the system call allows negative numbers to exist. Using one variable is always unsafe, but using two can completely eliminate this kind of thing and is beneficial for sorting

Explorer09: > You should also know that the x32 system call starts from 0x40000000, and the system call allows negative numbers to exist. Using one variable is always unsafe, but using two can completely eliminate this kind of thing and is beneficial for sorting

Nah. 0x40000000 is positive. The negative should always be unused or otherwise they won't reserve -1 for the documented reason.

JohnSanpe: Firstly, 0x40000000 is certainly not a negative number, but it is a special case. With one special case, there may be more. Also, if you do this, it will cause all N/A to be ranked at the top, which is not user-friendly. 😄

Explorer09: @JohnSanpe With a mapping table implemented, you are not likely order the syscall by its number, but by the mapped name instead.
Linux assigns syscall number in a rather arbitrary way, so, with the table implemented, ordering by name would make more sense.

ffashion: First, i think use negative number to show a syscall state. it's unsafe. Although Linux may not currently use negative numbers as syscall ID. 
Second, The state of syscall and the syscall ID themselves are actually two different things, and merging them would reduce readability.
Third, i think we can use a struct to wrapper syscall id and syscall state,  like 
```c
typedef enum LinuxSyscallState_ {
   SYSCALL_STATE_NO_DATA = INT_MIN,
   SYSCALL_STATE_RUNNING = INT_MIN + 1,
   SYSCALL_STATE_BLOCKED = -1
} LinuxSyscallState;

typedef enum LinuxSyscall_ {
LinuxSyscallState state;
int syscall_id;
} LinuxSyscall;

```

configure.ac
```diff
@@ -659,6 +659,27 @@ case "$enable_capabilities" in
...
+   if "$cross_compiling" != no; then
+      enable_syscall=yes
+   elif test -f /proc/self/syscall && test -s /proc/self/syscall; then
+      enable_syscall=yes
+   else
+      enable_syscall=no
+   fi
+fi
+if test "x$enable_syscall" = xyes; then
+   AC_DEFINE([HAVE_SYSCALL], [1], [Define to 1 is syscall monitoring is enabled.])
```
BenBE, line 679:
```suggestion
   AC_DEFINE([HAVE_SYSCALL], [1], [Define if syscall monitoring enabled.])
```

linux/LinuxProcess.c
```diff
@@ -362,6 +365,23 @@ static void LinuxProcess_rowWriteField(const Row* super, RichString* str, Proces
          xSnprintf(buffer, n, "N/A  ");
       }
       break;
+   #ifdef HAVE_SYSCALL
+   case SYSCALL: {
+      switch (lp->syscall_state) {
+      case SYSCALL_STATE_NORMAL:
+         xSnprintf(buffer, n, "%-8d", lp->syscall_num);
```
BenBE, line 372: Using the number here may not be very useful nor intuitive for the user. Also 8 characters is kinda short to print a readable name. Maybe use a dynamically resizeable column for display?

linux/LinuxProcessTable.c
```diff
@@ -1098,6 +1098,38 @@ static void LinuxProcessTable_readCwd(LinuxProcess* process, openat_arg_t procFd
...
+   process->syscall_num = fast_strtoull_dec(&numPtr, r);
+   if (*numPtr == ' ') {
+      process->syscall_state = SYSCALL_STATE_NORMAL;
+      return;
+   }
+
+   char* eol = String_strchrnul(buffer, '\n');
+   *eol = '\0';
+
+   if (!strcmp(buffer, "running")) {
```
BenBE, line 1123:
```suggestion
   if (String_eq(buffer, "running")) {
```

BenBE (changes requested):

linux/LinuxProcessTable.c
```diff
@@ -1098,6 +1098,41 @@ static void LinuxProcessTable_readCwd(LinuxProcess* process, openat_arg_t procFd
...
+   if (*numPtr == ' ') {
+      process->syscall_state = SYSCALL_STATE_CALLING;
+      Row_updateFieldWidth(SYSCALL, snprintf(NULL, 0, "%d", process->syscall_num));
+      return;
+   }
+
+   char* eol = String_strchrnul(buffer, '\n');
+   *eol = '\0';
+
+   if (String_eq(buffer, "running")) {
```
Explorer09, line 1124: You can use `String_startsWith(buffer, "running")` and free the need of assigning `*eol = '\0'` on the line above.

JohnSanpe:
Does CI seem to have some issues?

BenBE:
> Does CI seem to have some issues?

No. CI is waiting for approval by one of the maintainers as this is your first pull request for this project.

linux/LinuxProcess.c
```diff
@@ -362,6 +365,26 @@ static void LinuxProcess_rowWriteField(const Row* super, RichString* str, Proces
          xSnprintf(buffer, n, "N/A  ");
       }
       break;
+   #ifdef HAVE_SYSCALL
+   case SYSCALL: {
+      char syscall[32];
+      switch (lp->syscall_state) {
+      case SYSCALL_STATE_CALLING:
+         xSnprintf(syscall, sizeof(syscall), "%d", lp->syscall_num);
```
BenBE, line 373: Shouldn't this be right-aligned?

PR #1490, opened 2024-06-22
Define keys for other $TERM values

### LINKED ISSUE #1397: Alt-[hjkl] doesn't work when $TERM is not xterm
<kbd>Alt</kbd> + <kbd>[hjkl]</kbd> does not work when `$TERM` is not set to `xterm-256-color`. This occurs when using programs that set `$TERM` such as tmux or terminal emulators that set a different `$TERM` value such as simple terminal.

In my case, experiencing this issue using Suckless' Simple Terminal. ST's value is set to `st-256color`.

See the issue [filed on the old repository](https://github.com/hishamhm/htop/issues/942).
Related: https://github.com/htop-dev/htop/issues/1316

unrealapex:
Allow keymaps to work in terminals that have other `$TERM` values set.

Closes #1397

### DIFF
--- a/CRT.c
+++ b/CRT.c
@@ -1007,6 +1007,24 @@ static bool terminalSupportsDefinedKeys(const char* termType) {
    }
 
    switch (termType[0]) {
+   case 'a':
+      if (String_eq(termType, "alacritty")) {
+         return true;
+      }
+      break;
+   case 's':
+      if (termType[1] == 't' && (termType[2] == '-' || !termType[2])) {
+         return true;
+      }
+      if (String_eq(termType, "screen")) {
+         return true;
+      }
+      break;
+   case 't':
+      if (String_eq(termType, "tmux")) {
+         return true;
+      }
+      break;
    case 'v':
       if (String_eq(termType, "vt220")) {
          return true;
--- a/CRT.c
+++ b/CRT.c
@@ -1001,6 +1001,27 @@ void CRT_setMouse(bool enabled) {
 }
 #endif
 
+static bool terminalSupportsDefinedKeys(const char* termType) {
+   if (!termType) {
+      return false;
+   }
+
+   switch (termType[0]) {
+   case 'v':
+      if (String_eq(termType, "vt220")) {
+         return true;
+      }
+      break;
+   case 'x':
+      if (String_eq(termType, "xterm")) {
+         return true;
+      }
+      break;
+   }
+
+   return false;
+}
+
 void CRT_init(const Settings* settings, bool allowUnicode, bool retainScreenOnExit) {
    initscr();
 
@@ -1046,7 +1067,7 @@ void CRT_init(const Settings* settings, bool allowUnicode, bool retainScreenOnEx
       CRT_scrollHAmount = 5;
    }
 
-   if (termType && (String_startsWith(termType, "xterm") || String_eq(termType, "vt220"))) {
+   if (terminalSupportsDefinedKeys(termType)) {
 #ifdef HTOP_NETBSD
 #define define_key(s_, k_) define_key((char*)s_, k_)
 IGNORE_WCASTQUAL_BEGIN


CRT.c
```diff
@@ -1046,7 +1046,7 @@ void CRT_init(const Settings* settings, bool allowUnicode, bool retainScreenOnEx
       CRT_scrollHAmount = 5;
    }
 
-   if (termType && (String_startsWith(termType, "xterm") || String_eq(termType, "vt220"))) {
+   if (termType && (String_startsWith(termType, "xterm")  || String_eq(termType, "vt220") || String_startsWith(termType, "screen") || String_startsWith(termType, "tmux") || String_startsWith(termType, "st"))) {
```
unrealapex, line 1049: If there are other `$TERM` values that could also benefit from being here, I would appreciate it if they were pointed out.

CRT.c
```diff
@@ -1046,7 +1046,7 @@ void CRT_init(const Settings* settings, bool allowUnicode, bool retainScreenOnEx
       CRT_scrollHAmount = 5;
    }
 
-   if (termType && (String_startsWith(termType, "xterm") || String_eq(termType, "vt220"))) {
+   if (termType && (String_startsWith(termType, "xterm") || String_eq(termType, "vt220") || String_startsWith(termType, "screen") || String_startsWith(termType, "tmux") || String_startsWith(termType, "st"))) {
```
BenBE, line 1049:
```suggestion
   if (termType && (
      String_startsWith(termType, "xterm") ||
      String_eq(termType, "vt220") ||
      String_startsWith(termType, "screen") ||
      String_startsWith(termType, "tmux") ||
      String_startsWith(termType, "st")
   )) {
```

Isn't `st` a bit generic on its own?

unrealapex: I agree `st` is pretty generic. On my system, the termtypes that are erroneously matched are as follows
|TERM|description|
|---|--|
|`st52`|Atari ST|
|`st52-old`|Atari ST with VT52 emulation|
|`stv52`|MiNT virtual console|
|`stv52pc`|MiNT virtual console with PC charset|

The default TERM for ST is set to `st-256color` so we could use that instead.

BenBE: I'd prefer that as `st` (Suckless Terminal) and `st52` (Atari ST) are quite different and I doubt the Atari ST implements most of the features relevant in that part of the code. Not sure about `stv52` though, but from what I could gather this seems kinda limited in its feature set too. `String_eq(termType, "st") || String_beginsWith(termType, "st-")` might be a compromise here to capture both `st` and `st-256color` (or variants) …

Also a quick note about grouping checks, as this might be of interest here: We could do something like:

```c
      (termType[0] == 's' && (
         String_beginsWith(…, 'screen') ||
         String_eq(…, 'st') ||
         String_beginsWith(…, 'st-')
      )) ||
```

For groups with at least 3 items as this allows the compiler to short-circuit on large portions of the condition with similar first letter.

CRT.c
```diff
@@ -1046,7 +1046,14 @@ void CRT_init(const Settings* settings, bool allowUnicode, bool retainScreenOnEx
       CRT_scrollHAmount = 5;
    }
 
-   if (termType && (String_startsWith(termType, "xterm") || String_eq(termType, "vt220"))) {
+   if (termType && (
```
Explorer09, line 1049: Can we add a section comment or factor this into a function on what this list of `termType` matches for?

CRT.c
unrealapex: Great idea! Short circuiting is more efficient, but the former you provided is a simper implementation. Which one do you prefer we use?

Explorer09: I suggest you to factor this into its own function first, since the list of string would be eventually growing anyway, that would make the list more readable and manageable.

unrealapex: I think that would be a good idea. I'll refactor the code.

BenBE:
Please keep your commit history linear (no merges) and simplify it to have mere fixups of previous commits squashed.
E.g. factor out the original check into its own function, have further commits add on that. The short-circuit optimization can go in last or as part of adding in additional terminals.

CRT.c
```diff
@@ -1001,6 +1001,26 @@ void CRT_setMouse(bool enabled) {
 }
 #endif
 
+static bool termKeySupport(const char* termType) {
```
Explorer09, line 1004:
```c
static bool termSupportsDefinedKeys(const char* termType) {
   if (!termType)
      return false;

   if (termType[0] == 's') {
      if (termType[1] == 't' && (termType[2] == '-' || !termType[2])) {
         // Suckless Terminal. Regex "st(-)?".
         // Don't match other terminals that might start with "st".
         return true;
      }
      if (String_startsWith(termType, "screen") {
         return true;
      }
   }
   if (
      String_startsWith(termType, "xterm") ||
      String_eq(termType, "vt220") ||
      String_startsWith(termType, "tmux") ||
      String_eq(termType, "alacritty")
   ) {
      return true;
   }
   return false;
}
```

BenBE: Consider …

```c
   switch(termType[0]) {
      case 's':
         if (
            String_eq(…)
         ) {
            return true;
         }
         break;
      case 't':
         if (
            String_eq(…)
         ) {
            return true;
         }
         break;
      case 'x':
         if (
            String_eq(…)
         ) {
            return true;
         }
         break;
   }

   return false;
```

Explorer09: @BenBE I would say it's okay either way.
Here's the suggested edit:

```c
static bool termSupportsDefinedKeys(const char* termType) {
   if (!termType)
      return false;

   switch (termType[0]) {
   case 'a':
      if (String_eq(termType, "alacritty"))
         return true;
      break;
   case 's':
      if (termType[1] == 't' && (termType[2] == '-' || !termType[2])) {
         // Suckless Terminal. Regex "st(-)?".
         // Don't match other terminals that might start with "st".
         return true;
      }
      if (String_startsWith(termType, "screen") {
         return true;
      }
      break;
   case 't':
      if (String_startsWith(termType, "tmux"))
         return true;
      break;
   case 'v':
      if (String_eq(termType, "vt220"))
         return true;
      break;
   case 'x':
      if (String_startsWith(termType, "xterm"))
         return true;
      break;
   }
   return false;
}
```

CRT.c
```diff
@@ -12,8 +12,10 @@ in the source distribution for its full text.
 #include <errno.h>
 #include <fcntl.h>
 #include <langinfo.h>
+#include <ncursesw/term.h>
```
BenBE, line 15: Should be covered by "ProvideCurses.h" …

CRT.c
```diff
@@ -1068,7 +1088,7 @@ void CRT_init(const Settings* settings, bool allowUnicode, bool retainScreenOnEx
 
    /*check termTypes to determine if term supports escape sequences for defined
     * keys */
-   if (termType && termKeySupport(termType)) {
+   if (termType && termSupportsDefinedKeys(termType)) {
```
BenBE, line 1091: Duplicated NULL check; skip here and just leave the one inside the function.

CRT.c
```diff
@@ -1001,23 +1003,41 @@ void CRT_setMouse(bool enabled) {
...
-      )) ||
-      String_eq(termType, "alacritty")
-   ) {
-      return true;
-   } else {
+static bool termSupportsDefinedKeys(const char* termType) {
+   if (!termType) {
       return false;
    }
+   switch (termType[0]) {
```
BenBE, line 1010: Blank line before.

Also the `break;` should be indented … Compare with other `switch` statements or the style guide …

BenBE (changes requested):

BenBE:
> > Please keep your commit history linear (no merges) and simplify it to have mere fixups of previous commits squashed.
> 
> How would I achieve this? I haven't been in a situation where I needed to make the commit history linear. Do I rebase my branch or is there some other Git magic I'm missing?

Yes, you rebase your branch. Have a look at `git rebase -i --autosquash main` (while on your branch).

CRT.c
```diff
@@ -1046,7 +1086,9 @@ void CRT_init(const Settings* settings, bool allowUnicode, bool retainScreenOnEx
       CRT_scrollHAmount = 5;
    }
 
-   if (termType && (String_startsWith(termType, "xterm") || String_eq(termType, "vt220"))) {
+   /*check termTypes to determine if term supports escape sequences for defined
+    * keys */
```
BenBE, line 1090: Keep this on one line.
```suggestion
   /* check termTypes to determine if term supports escape sequences for defined keys */
```

CRT.c
Explorer09: @unrealapex An informative note: the `break` statement is part of the `case` block. A `case` block doesn't need to be terminated by a `break` statement, and there might be multiple `break` statements if the block has multiple exit points. That's the reason the `break` statement is indented one level below the `case` label. It looks like an elementary mistake to me if someone doesn't indent the `break` statement in that usual way.

CRT.c
Explorer09: I think this comment is outdated and we may just remove it. I suggested adding that comment before the function `termSupportsDefinedKeys` is added so that people can know what the conditional is about. Now the `termSupportsDefinedKeys` name is self-documenting that we can remove this comment line.

unrealapex:
LGTM!

BenBE (approved):

BenBE:
@unrealapex That's what I meant with "refactor into separate function first, add on additional checks later". This way the history looks much more natural to read and changes aren't dragged across several commits (like the additional term checks which were in at least 2 commits where they needed to be touched). A good commit history tells the story of what needs to be done, not how you got there. ;-)

unrealapex:
> That's what I meant with "refactor into separate function first, add on additional checks later".

That makes sense. Initially I was a bit unsure on what exactly you envisioned. Thank you for clarification.

> A good commit history tells the story of what needs to be done, not how you got there. ;-)

 Great analogy, I'll keep this in mind :)

PR #1496, opened 2024-07-10
Additional function attributes for XUtils (+analyzer fixes)

BenBE:
This does 3 things:
- Fix a nasty diagnostic from GCC 14's static code analyzer
- Introduce `ATTR_RETNONNULL` to indicate that a function always returns a non-NULL result
- Mark several functions in `XUtils.h` with applicable attributes

The third commit depends on the second one to land. ;-)

Also a note on the first commit: We previously already asserted that `size` is non-zero for `xRealloc`, thus we'd never intentionally free. Thus if assertions were compiled in you already couldn't have passed `xRealloc(p, 0)`. Thus although technically correct, previously the function would have to have an additional `if (!size) { free(p); return NULL; }` path or simply bail on `size==0` too. Interestingly [util-linux](https://sources.debian.org/src/util-linux/2.40.2-1/include/xalloc.h/?hl=42#L42-L49) did basically the same thing, but given our explicit constraint on `size` we can simplify. Not simplifying triggers GCC 14.1, while simply checking for `!data` after `realloc` makes GCC 14 happy.

### DIFF
--- a/Macros.h
+++ b/Macros.h
@@ -69,6 +69,16 @@ in the source distribution for its full text.
 
 #endif /* HAVE_ATTR_NONNULL */
 
+#ifdef HAVE_ATTR_RETNONNULL
+
+#define ATTR_RETNONNULL                 __attribute__((returns_nonnull))
+
+#else
+
+#define ATTR_RETNONNULL
+
+#endif /* HAVE_ATTR_RETNONNULL */
+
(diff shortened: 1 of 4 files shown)


fasterit (approved):
LGTM
Thank you for improving the housekeeping!

Compiles clean, runs clean
```bash
$ ./htop --version
htop 3.4.0-dev-3.3.0-148-g07f98c5
```

PR #1499, opened 2024-07-17
PCP configuration file linting

BenBE:
This adds linting for the PCP configuration files shipped with each release.

Currently this abomination of an AWK monster checks for:
- Duplicate sections (per file)
- Missing keys per section
- Duplicate keys per section
- Missing keys per group (meters, screens)
- Blank line at EOF of files

The linting is kinda dumb, but gets the job done. It even found a mistake in the MySQL meters … ;-)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -322,9 +322,16 @@ jobs:
               ./configure --enable-unicode --enable-werror
               gmake -k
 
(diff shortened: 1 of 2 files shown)


natoscott (approved):
LGTM @BenBE - I've tackled the MySQL config issue within #1500

Issue #952, opened 2022-03-06
Feature Request: Disk IO per disk


cfitz25:
Hello, i recently found out htop can show DIsk IO read and write, though it seems to be a combination of all things it considers a "drive". 
Would it be possible to have the Disk IO per drive as a "Meter"? 
This would just mean a lot more if you have drives with very different read/write speeds, such as an SSD for the main system and HDD for long term storage

BenBE:
This idea has been brought up before as part of PR #200. Unfortunately that PR did not receive much traction.

tjk213:
+1 to this request! The CPUs have one meter per core, why not one meter per disk? And for that matter, one meter per network interface would be great too.

BenBE:
This mostly hinges on a proper implementation of #200 on the GUI side of things. Unfortunately the selector for possibly varying lists of items is not really intuitive as far as I remember the state of that PR.

natoscott:
If it is of use to anyone here, the htop PCP backend (pcp-htop) provides a Disks tab in the htop UI.

This reports quite clearly on the state of each disk, with individual columns for reads, writes, discards, IOPs, queue lengths, throughput, utilization, etc for each device and with each column being individually sortable - you may find its a bit more like the htop UI you're used to.

I'm certainly not against the Meter Choice idea from #200 though.

ChillarAnand:
I am running htop with pcp but still I am not able to see the disks tab in htop UI?

@natoscott Can you please share how to achieve this?

natoscott:
If you have a htoprc file from previously running htop with the regular Linux platform backend (i.e. not pcp-htop), pcp-htop will respect that configuration - including the use of the Tab configuration from Linux.  This is usually the cause of this confusion (probably we should separate these config files to avoid this).  Alternatively, you may be using an older version of pcp-htop - in that case, upgrade PCP - we track htop very closely with the builds provided in PCP itself.

PR #1501, opened 2024-07-21
Separate htoprc for htop and pcp-htop to prevent overwrite

natoscott:
The original intention with pcp-htop was to share the same htoprc configuration file with regular htop.  However this also stores tab configuration which causes confusion since the default pcp-htop tab set is unintentionally reduced to match the defaults on Linux (when htop is invoked prior to pcp-htop).

Relates to issue #952.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -599,6 +599,19 @@ if test -n "$with_os_release" && test ! -f "$with_os_release"; then
 fi
 AC_DEFINE_UNQUOTED([OSRELEASEFILE], ["$with_os_release"], [File with OS release details.])
 
+AC_ARG_WITH([config],
+            [AS_HELP_STRING([--with-config=DIR],
+			    [configuration path @<:@default=/.config@:>@])],
+            [],
+            [with_config="/.config"])
+dnl Performance Co-Pilot configuration location to prevent overwrite
+if test "$my_htop_platform" = pcp -a "$with_config" = /.config; then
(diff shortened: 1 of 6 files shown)


BenBE:
Why not pass this via `configure.ac`/`Makefile.am`?

natoscott:
> Why not pass this via `configure.ac`/`Makefile.am`?

No reason, good idea - let me code that up quickly.

configure.ac
```diff
@@ -599,6 +599,19 @@ if test -n "$with_os_release" && test ! -f "$with_os_release"; then
...
+			    [configuration path @<:@default=/.config@:>@])],
+            [],
+            [with_config="/.config"])
+dnl Performance Co-Pilot configuration location to prevent overwrite
+if test "$my_htop_platform" = pcp -a "$with_config" = /.config; then
+    with_config="/.pcp"
+elif test -z "$with_config"; then
+   AC_MSG_ERROR([bad empty value for --with-config option])
+fi
+AC_DEFINE_UNQUOTED([CONFIGDIR], ["$with_config"], [Configuration path.])
```
BenBE, line 613: Kinda confusing to see the initial `/` in the path, but that's correct given the context where it's used. May cause some confusion from users downstream.

natoscott: Agreed, its a bit of a tradeoff but I leaned towards keeping things simpler.

PR #1506, opened 2024-07-24
Implement a meta-check for ncurses detection

BenBE:
This was inspired by a patch for OpenEmbedded:
https://git.openembedded.org/meta-openembedded/tree/meta-oe/recipes-support/htop/files/0001-Use-pkg-config.patch?id=110667ac3798b5b64552cb4b8dc706aad8fbcdfe

Using this patch, detection can be switched as appropriate.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -407,6 +407,50 @@ m4_define([HTOP_CHECK_SCRIPT],
    fi
 ])
 
+# Some meta programming for fun and profit
+
+_check_nc() {
+   tool="$1-config"
+   shift 1
+   "$tool" "$@" 2>/dev/null
+}
+
+AC_ARG_WITH([check-ncurses],
+   [AS_HELP_STRING([--with-check-ncurses={auto|pkg-config|nc-config|script-name}],
+   [Specify how to check for ncurses (default: auto)])],
+   [check_ncurses="$withval"],
+   [check_ncurses="auto"]
+)
+
+AC_MSG_CHECKING([for a way to check for ncurses])
+case "$check_ncurses" in
+   pkg-config)
+      check_nc="pkg-config"
+      AC_MSG_RESULT([pkg-config])
+      ;;
+   nc-config)
+      check_nc="_check_nc"
+      AC_MSG_RESULT([ncurses*-config helpers])
+      ;;
+   auto)
+      if command -v pkg-config >/dev/null 2>&1; then
+         check_nc="pkg-config"
+         AC_MSG_RESULT([pkg-config (auto-detected)])
+      else
+         check_nc="_check_nc"
+         AC_MSG_RESULT([ncurses*-config (auto-detected)])
+      fi
+      ;;
+   *)
+      if command -v "$check_ncurses" >/dev/null 2>&1; then
+         check_nc="$check_ncurses"
+         AC_MSG_RESULT([$check_ncurses (custom script)])
+      else
+         AC_MSG_ERROR([The specified helper script "$check_ncurses" is not an executable command])
+      fi
+      ;;
+esac
+
 # HTOP_CHECK_LIB(LIBNAME, FUNCTION, DEFINE, ELSE_PART)
 m4_define([HTOP_CHECK_LIB],
 [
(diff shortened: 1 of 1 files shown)


BenBE:
Something seems off with the Ubuntu build. Any ideas?

fasterit:
> Something seems off with the Ubuntu build. Any ideas?

I get the same on a local build with `--enable-static`` (on Debian bookworm) after your change:

```
...
checking for a way to check for ncurses... pkg-config (auto-detected)
checking for addnwstr in -lncursesw6... no
checking for addnwstr in -lncursesw... no
checking for addnwstr in -lncurses... no
configure: error: can not find required library libncursesw; you may want to use --disable-unicode
```

previously it compiles fine to a static executable and the configure output snippet is:

```
....
checking for waddwstr in -lncursesw6... no
checking for waddwstr in -lncursesw... yes
checking for ncursesw/curses.h... yes
checking for ncursesw/term.h... yes
```

BenBE:
Does it build with `./configure --enable-static --with-check-ncurses=nc-config`? Currently the patch changes the default when auto-detecting the setting.

fasterit:
> Does it build with `./configure --enable-static --with-check-ncurses=nc-config`? Currently the patch changes the default when auto-detecting the setting.

Yes, it does.

```
...
checking for a way to check for ncurses... ncurses*-config helpers
checking for waddwstr in -lncursesw6... no
checking for waddwstr in -lncursesw... yes
checking for ncursesw/curses.h... yes
checking for ncursesw/term.h... yes
checking for library containing keypad... none required
```

Explorer09:
The problem in my opinion is that it doesn't make sense to let user specify how to _check_ ncurses. The user should only specify whether ncurses exists, and optionally how to link to it.
So the option should look like this: `--with-ncurses=<libname>` while `<libname>` could be `ncursesw6`, `ncursesw5`, `ncurses6`, `ncurses`, etc.
Perhaps we can also introduce two variables `NCURSES_CFLAGS` and `NCURSES_LDFLAGS` just to allow user to override the flags when the values retrieved from "`*-config`" has errors.

Now, when it comes to automatic detection of `ncurses`, I can suggest this order:

1. `NCURSES_CFLAGS` and `NCURSES_LDFLAGS` take highest priority if both are specified.
2. `--with-ncurses=<libname>` if user specifies it. We can retrieve `NCURSES_CFLAGS` and `NCURSES_LDFLAGS` with the specified library name. First through `pkg-config`, and then `<libname>-config`. The `pkg-config` is preferred as it's a more generic library information utility. If `pkg-config` is not available, look for `<libname>-config` instead (which is specific to ncurses information).
3. If user does not specify `--with-ncurses`, then we guess all possible names of ncurses installed in the user's system. First through `pkg-config`, and then `<libname>-config`.

BenBE:
@Explorer09 Please see the linked patch from OpenEmbedded for context.

Explorer09:
@BenBE I did read [the linked patch](https://git.openembedded.org/meta-openembedded/tree/meta-oe/recipes-support/htop/files/0001-Use-pkg-config.patch?id=110667ac3798b5b64552cb4b8dc706aad8fbcdfe), and thus I get the conclusion we are doing the wrong way.

The `--with-check-ncurses` variable should be internal to `configure`. These is no sense for user to specify that. Because either `pkg-config` works or it doesn't, and if user has to specify a awkward script name (say, `ncursesfoo-config2`) just to get the linker flags for ncurses, then their ncurses installation is broken. We don't need to find any script name other than `<libname>-config` because that's what the ncurses config information utility would be installed as. Thus, `--with-ncurses=<libname>` would be enough.

Explorer09:
> The problem in my opinion is that it doesn't make sense to let user specify how to _check_ ncurses. The user should only specify whether ncurses exists, and optionally how to link to it. So the option should look like this: `--with-ncurses=<libname>` while `<libname>` could be `ncursesw6`, `ncursesw5`, `ncurses6`, `ncurses`, etc. Perhaps we can also introduce two variables `NCURSES_CFLAGS` and `NCURSES_LDFLAGS` just to allow user to override the flags when the values retrieved from "`*-config`" has errors.
> 

I implemented the idea in #1512. Please take a look and make some comments.

PR #1509, opened 2024-08-04
CI: Fix NetBSD build.

fraggerfox:
~Add `--disable-unicode` for now so that the CI passes the build without failing the `ncurses` check.~

Fix the build with proper `ncurses` support for NetBSD 10.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -241,34 +241,30 @@ jobs:
               ./configure --enable-unicode --enable-werror
               gmake -k
 
-# This is broken since the ncursesw merge in NetBSD, cf.
-#  https://github.com/NetBSD/pkgsrc/commit/4f72c2c3a2081165a195b2667bbd5e2030397442
-# DL240722, disabling this until somebody with NetBSD can fix the build
-#
-#  build-netbsd-latest-gcc:
-#    runs-on: ubuntu-22.04
-#    timeout-minutes: 20
-#    steps:
-#      - uses: actions/checkout@v4
-#        with:
-#          submodules: recursive
-#      - name: Compile
-#        uses: vmactions/netbsd-vm@v1
-#        with:
-#          release: '10.0'
-#          usesh: true
-#          prepare: |
-#            PATH="/usr/pkg/sbin:/usr/pkg/bin:$PATH"
-#            PKG_PATH="https://ftp.netbsd.org/pub/pkgsrc/packages/NetBSD/amd64/10.0/All/"
-#            export PATH PKG_PATH
-#            /usr/sbin/pkg_add pkgin
-#            pkgin -y install autoconf automake libtool ncurses ncursesw gmake git
-#            git config --global --add safe.directory /home/runner/work/htop/htop
-#          run: |
-#            set -e
-#            ./autogen.sh
-#            CPPFLAGS="-I/usr/pkg/include" ./configure --enable-unicode --enable-werror
-#            gmake -k
+  build-netbsd-latest-gcc:
+      runs-on: ubuntu-22.04
+      timeout-minutes: 20
+      steps:
+        - uses: actions/checkout@v4
+          with:
+            submodules: recursive
+        - name: Compile
+          uses: vmactions/netbsd-vm@v1
+          with:
+            release: '10.0'
+            usesh: true
+            prepare: |
+              PATH="/usr/pkg/sbin:/usr/pkg/bin:$PATH"
+              PKG_PATH="https://ftp.netbsd.org/pub/pkgsrc/packages/NetBSD/amd64/10.0/All/"
+              export PATH PKG_PATH
+              /usr/sbin/pkg_add pkgin
+              pkgin -y install autoconf automake libtool ncurses gmake git
+              git config --global --add safe.directory /home/runner/work/htop/htop
+            run: |
+              set -e
+              ./autogen.sh
+              CPPFLAGS="-I/usr/pkg/include -I/usr/pkg/include/ncurses" LDFLAGS="-L/usr/pkg/lib" ./configure --enable-unicode --enable-werror
+              gmake -k
 
   build-openbsd-latest-clang:
       runs-on: ubuntu-22.04
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -258,12 +258,12 @@ jobs:
               PKG_PATH="https://ftp.netbsd.org/pub/pkgsrc/packages/NetBSD/amd64/10.0/All/"
               export PATH PKG_PATH
               /usr/sbin/pkg_add pkgin
-              pkgin -y install autoconf automake libtool ncurses gmake git
+              pkgin -y install autoconf automake libtool ncurses pkg-config gmake git
               git config --global --add safe.directory /home/runner/work/htop/htop
             run: |
               set -e
               ./autogen.sh
-              CPPFLAGS="-I/usr/pkg/include -I/usr/pkg/include/ncurses" LDFLAGS="-L/usr/pkg/lib" ./configure --enable-unicode --enable-werror
+              CPPFLAGS=$(pkg-config --cflags ncurses) LDFLAGS=$(pkg-config --libs ncurses) ./configure --enable-unicode --enable-werror
               gmake -k
 
   build-openbsd-latest-clang:


Explorer09:
How about we let NetBSD teams figure out what happened that breaks the build before we implement more workarounds to the problem?

fraggerfox:
@Explorer09 : Do you have some specific person or team within NetBSD to figure out what happened? If not, I am a developer associated with NetBSD, who maintains this package.

The build within NetBSD 10 and pkgsrc currently does not enable unicode at least according to the configure options, but I could be wrong.

If you would like someone else from the NetBSD team to have a look, happy to close the PR.

Explorer09:
@fraggerfox The htop used to succeed building in NetBSD with `--enable-unicode` until upstream removed the obsolete `ncursesw` package. The upstream claimed that `ncursesw` is merged with `ncurses` (see this: https://github.com/NetBSD/pkgsrc/commit/af6a20fc3d516416e5ff5bde25accf1d5eeb3523), but somehow our configure script cannot detect the wide character functionality from the new `ncurses`. So there is a compatibility issue, either on our side or on NetBSD's side. I hope you can figure out which side has the issue.

By the way, the detection logic of `ncurses` is pending for a rework (see #1506). Maybe the two issues can be solved together.

fraggerfox:
@Explorer09 : Yup I am aware of the `ncurses` merger, as of now the NetBSD's pkgsrc links it against the base `libcurses`

```
/usr/pkg/bin/htop:
        -lexecinfo.0 => /usr/lib/libexecinfo.so.0
        -lelf.2 => /usr/lib/libelf.so.2
        -lc.12 => /usr/lib/libc.so.12
        -lgcc_s.1 => /usr/lib/libgcc_s.so.1
        -lcurses.9 => /usr/lib/libcurses.so.9
        -lterminfo.2 => /usr/lib/libterminfo.so.2
        -lprop.1 => /usr/lib/libprop.so.1
        -lkvm.6 => /usr/lib/libkvm.so.6
        -lm.0 => /usr/lib/libm.so.0
```

And the build was configured to use the built in `curses`,  as mentioned in the [Readme](https://github.com/htop-dev/htop/tree/main/netbsd)

I guess when I setup the CI I should have used the default values as pkgsrc was using https://github.com/NetBSD/pkgsrc/blob/trunk/sysutils/htop/options.mk#L18 since `ncurses` is marked as the optional one with `curses` as the default.

But if you would like to wait until the `ncurses` issue is resolved, I can close the PR for now and investigate on how to re-enable the `ncurses` logic.

Explorer09:
@fraggerfox Wait a second. So NetBSD has `ncurses` as well as NetBSD's own `curses` implementation?

Can we let our CI select `ncurses` as default with `--enable-unicode`, and `curses` with `--disable-unicode`? This way we can satisfy both build cases, I mean users.

BenBE:
Seems even more like a reason to implement a mechanism like #1506 to allow users to easily switch the detection. But that's just an aside. We can discuss the details for it over there.

Regarding this issue: I think, even if the proposed workaround somewhat "limits" the coverage on NetBSD, it's still better than no coverage at all. For the time being, I can live with the reduced feature set for now, as it also introduces a system with disabled Unicode support (which we do not have many AFAIR).

In the long run we should though, even if this PR is merged, aim to get the underlying issue on NetBSD fixed to restore full Unicode support.

.github/workflows/ci.yml
```diff
@@ -241,34 +241,30 @@ jobs:
...
+             PATH="/usr/pkg/sbin:/usr/pkg/bin:$PATH"
+             PKG_PATH="https://ftp.netbsd.org/pub/pkgsrc/packages/NetBSD/amd64/10.0/All/"
+             export PATH PKG_PATH
+             /usr/sbin/pkg_add pkgin
+             pkgin -y install autoconf automake libtool ncurses ncursesw gmake git
+             git config --global --add safe.directory /home/runner/work/htop/htop
+           run: |
+             set -e
+             ./autogen.sh
+             CPPFLAGS="-I/usr/pkg/include -I/usr/pkg/include/ncurses" LDFLAGS="-L/usr/pkg/lib -Wl,-R/usr/pkg/lib" ./configure --enable-unicode --enable-werror
```
fraggerfox, line 266: Restore the build with `ncurses` support.

fraggerfox:
```
  htop 3.4.0-dev

  platform:                  netbsd
  os-release file:           /etc/os-release
  (Linux) proc directory:    /proc
  (Linux) openvz:            no
  (Linux) vserver:           no
  (Linux) ancient vserver:   no
  (Linux) delay accounting:  no
  (Linux) sensors:           no
  (Linux) capabilities:      no
  unicode:                   yes
  affinity:                  no
  unwind:                    no
  hwloc:                     no
  debug:                     no
  static:                    no
```

Output of `./configure`

fraggerfox:
```
gcc -DHAVE_CONFIG_H -I.  -DNDEBUG -I/usr/pkg/include -I/usr/pkg/include/ncurses -std=c99 -pedantic -Wall -Wcast-align -Wcast-qual -Wextra -Wfloat-equal -Wformat=2 -Winit-self -Wmissing-format-attribute -Wmissing-noreturn -Wmissing-prototypes -Wpointer-arith -Wshadow -Wstrict-prototypes -Wundef -Wunused -Wwrite-strings -Wnull-dereference -Werror -D_XOPEN_SOURCE_EXTENDED -DSYSCONFDIR="\"/usr/local/etc\"" -I"./netbsd" -g -O2 -MT RichString.o -MD -MP -MF $depbase.Tpo -c -o RichString.o RichString.c &&\
In file included from /usr/include/ctype.h:100,
                 from RichString.c:12:
RichString.c: In function ‘RichString_writeFromAscii’:
RichString.c:147:87: error: array subscript has type ‘char’ [-Werror=char-subscripts]
  147 |       this->chptr[i] = (CharType) { .attr = attrs & 0xffffff, .chars = { (isprint(data[j]) ? data[j] : L'\xFFFD') } };
      |                                                                                       ^
```

and

```
gcc -DHAVE_CONFIG_H -I.  -DNDEBUG -I/usr/pkg/include -I/usr/pkg/include/ncurses -std=c99 -pedantic -Wall -Wcast-align -Wcast-qual -Wextra -Wfloat-equal -Wformat=2 -Winit-self -Wmissing-format-attribute -Wmissing-noreturn -Wmissing-prototypes -Wpointer-arith -Wshadow -Wstrict-prototypes -Wundef -Wunused -Wwrite-strings -Wnull-dereference -D_XOPEN_SOURCE_EXTENDED -DSYSCONFDIR="\"/usr/local/etc\"" -I"./netbsd" -g -O2 -MT Settings.o -MD -MP -MF $depbase.Tpo -c -o Settings.o Settings.c &&\
In file included from /usr/include/ctype.h:100,
                 from Settings.c:12:
Settings.c: In function ‘toFieldIndex’:
Settings.c:235:19: warning: array subscript has type ‘char’ [-Wchar-subscripts]
  235 |    if (isdigit(str[0])) {
      |        
```

We need to fix those before merging if we have `--enable-werror`, I am looking into this.

BenBE:
I think with some similar cases we mostly just added some explicit typecasts IIRC.

fraggerfox:
> Due to the following warnings
> 
> ```
> gcc -DHAVE_CONFIG_H -I.  -DNDEBUG -I/usr/pkg/include -I/usr/pkg/include/ncurses -std=c99 -pedantic -Wall -Wcast-align -Wcast-qual -Wextra -Wfloat-equal -Wformat=2 -Winit-self -Wmissing-format-attribute -Wmissing-noreturn -Wmissing-prototypes -Wpointer-arith -Wshadow -Wstrict-prototypes -Wundef -Wunused -Wwrite-strings -Wnull-dereference -Werror -D_XOPEN_SOURCE_EXTENDED -DSYSCONFDIR="\"/usr/local/etc\"" -I"./netbsd" -g -O2 -MT RichString.o -MD -MP -MF $depbase.Tpo -c -o RichString.o RichString.c &&\
> mv -f $depbase.Tpo $depbase.Po
> In file included from /usr/include/ctype.h:100,
>                  from RichString.c:12:
> RichString.c: In function ‘RichString_writeFromAscii’:
> RichString.c:147:87: error: array subscript has type ‘char’ [-Werror=char-subscripts]
>   147 |       this->chptr[i] = (CharType) { .attr = attrs & 0xffffff, .chars = { (isprint(data[j]) ? data[j] : L'\xFFFD') } };
>       |                                                                                       ^
> cc1: all warnings being treated as errors
> ```
> 
> and
> 
> ```
> gcc -DHAVE_CONFIG_H -I.  -DNDEBUG -I/usr/pkg/include -I/usr/pkg/include/ncurses -std=c99 -pedantic -Wall -Wcast-align -Wcast-qual -Wextra -Wfloat-equal -Wformat=2 -Winit-self -Wmissing-format-attribute -Wmissing-noreturn -Wmissing-prototypes -Wpointer-arith -Wshadow -Wstrict-prototypes -Wundef -Wunused -Wwrite-strings -Wnull-dereference -D_XOPEN_SOURCE_EXTENDED -DSYSCONFDIR="\"/usr/local/etc\"" -I"./netbsd" -g -O2 -MT Settings.o -MD -MP -MF $depbase.Tpo -c -o Settings.o Settings.c &&\
> mv -f $depbase.Tpo $depbase.Po
> In file included from /usr/include/ctype.h:100,
>                  from Settings.c:12:
> Settings.c: In function ‘toFieldIndex’:
> Settings.c:235:19: warning: array subscript has type ‘char’ [-Wchar-subscripts]
>   235 |    if (isdigit(str[0])) {
>       |        
> ```
> 
> the build would still fail and we need to fix those before merging if we have `--enable-werror`, I am looking into this.

These were an oversight from my end, they have been fixed in main.

fraggerfox:
@BenBE / @Explorer09 : I think we are good to merge this into main and see if the CI works.

.github/workflows/ci.yml
Explorer09: One issue here I see is that you need explicit `CFLAGS` and `LDFLAGS` in order to link against `ncurces` and not `curses`. This is not convenient for builders, and is the issue I wish to see it resolved in #1506.

fraggerfox: When building via pkgsrc with the `ncurses` option these flags are set before doing `./configure`
- https://github.com/NetBSD/pkgsrc/blob/trunk/sysutils/htop/Makefile#L21

The `${MAKE_ENV}` sets them in this case.

I agree it is not clean but with #1506 may be we can tweak the build to look cleaner outside the pkgsrc build script.

fraggerfox: I have removed the `LDFLAGS` and I think it should work now.

Explorer09: @fraggerfox Is it possible to obtain the library include path via a `pkg-config` command? That's the way I expect to see it done.

```text
LDFLAGS=$($PKG_CONFIG --libs ncurses)
```

Similarly for CPPFLAGS:

```text
CPPFLAGS=$($PKG_CONFIG --cflags ncurses)
```

fraggerfox: @Explorer09 : Should be done in the latest commit.

PR #1512, opened 2024-08-06
New "configure" ncurses detection code

Explorer09:
This PR is intended to supersede #1506

New detection code for `curses` library for `configure`.

* Both `pkg-config` and `ncurses*-config` will be used to find the compiler flags and linker flags for `curses`/`ncurses`. Fall back to simple `-l<libname>` linking if neither utilities are available.
* Environment variables `CURSES_CFLAGS` and `CURSES_LIBS` may be used to override the `--cflags` and `-libs` value.
* New configure option `--with-curses=<libname>` to specify the `curses`/`ncurses` module name. Should be used if the default detect list is not sufficient to find the library.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -263,7 +263,7 @@ jobs:
             run: |
               set -e
               ./autogen.sh
-              CPPFLAGS=$(pkg-config --cflags ncurses) LDFLAGS=$(pkg-config --libs ncurses) ./configure --enable-unicode --enable-werror
+              ./configure --enable-unicode --enable-werror
               gmake -k
 
   build-openbsd-latest-clang:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -233,7 +233,7 @@ jobs:
             release: '14.1'
             usesh: true
             prepare: |
-              pkg install -y gmake autoconf automake git
+              pkg install -y gmake autoconf automake pkgconf git
               git config --global --add safe.directory /home/runner/work/htop/htop
             run: |
               set -e
--- a/configure.ac
+++ b/configure.ac
@@ -450,19 +450,38 @@ htop_check_curses_capability () {
    # At this point we have not checked the name of curses header, so
    # use forward declaration for the linking tests below.
 
+   # htop uses keypad() and "stdscr", but for ncurses implementation,
+   # the symbols are in "-ltinfo" and not "-lncurses".
+   # Check "-ltinfo" symbols first, as libncurses might require
+   # explicit "-ltinfo" to link (for internal dependency).
+   AC_MSG_CHECKING([for keypad in $htop_msg_linker_flags])
+   AC_LINK_IFELSE([AC_LANG_PROGRAM([[
+/* extern WINDOW* stdscr; */
+/* int keypad(WINDOW* win, bool enable); */
+extern void* stdscr;
+int keypad(void* win, int enable);
+      ]], [[
+keypad(stdscr, 0);
+      ]])],
+      [AC_MSG_RESULT(yes)],
+      [AC_MSG_RESULT(no)
+      htop_curses_status=1])
+
    # htop calls refresh(), which might be implemented as a macro.
    # It is more reliable to test linking with doupdate(), which
    # refresh() would call internally.
-   AC_MSG_CHECKING([for doupdate in $htop_msg_linker_flags])
-   AC_LINK_IFELSE([AC_LANG_PROGRAM([[
+   if test "$htop_curses_status" -eq 0; then
+      AC_MSG_CHECKING([for doupdate in $htop_msg_linker_flags])
+      AC_LINK_IFELSE([AC_LANG_PROGRAM([[
 int doupdate(void);
-      ]], [[
+         ]], [[
 doupdate();
-      ]])],
-      [AC_MSG_RESULT(yes)
-      htop_curses_capability=nonwide],
-      [AC_MSG_RESULT(no)
-      htop_curses_status=1])
+         ]])],
+         [AC_MSG_RESULT(yes)
+         htop_curses_capability=nonwide],
+         [AC_MSG_RESULT(no)
+         htop_curses_status=1])
+   fi
 
    if test "x$htop_curses_status$enable_unicode" = x0yes; then
       AC_MSG_CHECKING([for mvadd_wchnstr in $htop_msg_linker_flags])
@@ -553,6 +572,15 @@ none-*|nonwide-yes)
       if htop_check_curses_capability "" "-l$curses_name"; then
          break
       fi
+      # For ncurses implementation, an extra "-ltinfo" or "-ltinfow"
+      # flag might be needed to link.
+      if test "x$enable_unicode" = xyes &&
+         htop_check_curses_capability "" "-l$curses_name -ltinfow"; then
+         break
+      fi
+      if htop_check_curses_capability "" "-l$curses_name -ltinfo"; then
+         break
+      fi
    done
    ;;
 esac
@@ -579,10 +607,6 @@ if test "x$enable_unicode" = xyes; then
       [AC_CHECK_HEADERS([ncurses/term.h], [],
          [AC_CHECK_HEADERS([term.h], [],
             [AC_MSG_ERROR([can not find required term header file])])])])
-
-   # check if additional linker flags are needed for keypad(3)
-   # (at this point we already link against a working ncurses library with wide character support)
-   AC_SEARCH_LIBS([keypad], [tinfow tinfo])
 else
    AC_CHECK_HEADERS([curses.h], [],
       [AC_CHECK_HEADERS([ncurses/curses.h], [],
@@ -593,10 +617,6 @@ else
    AC_CHECK_HEADERS([ncurses/term.h], [],
       [AC_CHECK_HEADERS([term.h], [],
          [AC_MSG_ERROR([can not find required term header file])])])
-
-   # check if additional linker flags are needed for keypad(3)
-   # (at this point we already link against a working ncurses library)
-   AC_SEARCH_LIBS([keypad], [tinfo])
 fi
 CFLAGS=$htop_save_CFLAGS
 
--- a/configure.ac
+++ b/configure.ac
@@ -384,6 +384,11 @@ dnl If the macro is not called, some pkg-config checks might be skipped
 dnl and $PKG_CONFIG might be unset.
 m4_ifdef([PKG_PROG_PKG_CONFIG], [
    PKG_PROG_PKG_CONFIG()
+], [
+   pkg_m4_absent=1 # Makefile might grep this keyword. Don't remove.
+   m4_warn(
+      [syntax],
+      [pkg.m4 is absent or older than version 0.16; this 'configure' would have incomplete pkg-config support])
 ])
 
 # HTOP_CHECK_SCRIPT(LIBNAME, FUNCTION, DEFINE, CONFIG_SCRIPT, ELSE_PART)
--- a/configure.ac
+++ b/configure.ac
@@ -614,7 +614,7 @@ AC_DEFINE_UNQUOTED([OSRELEASEFILE], ["$with_os_release"], [File with OS release 
 
 AC_ARG_WITH([config],
             [AS_HELP_STRING([--with-config=DIR],
-			    [configuration path @<:@default=/.config@:>@])],
+                            [configuration path @<:@default=/.config@:>@])],
             [],
             [with_config="/.config"])
 dnl Performance Co-Pilot configuration location to prevent overwrite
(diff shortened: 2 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -263,7 +263,7 @@ jobs:
             run: |
               set -e
               ./autogen.sh
-              CPPFLAGS=$(pkg-config --cflags ncurses) LDFLAGS=$(pkg-config --libs ncurses) ./configure --enable-unicode --enable-werror
+              ./configure --enable-unicode --enable-werror
```
fraggerfox, line 266: Looks a lot cleaner! Thanks @Explorer09

Explorer09: One thing left to do is to update the netbsd/README.md to mention that configure would now favor libncurses over libcurses when the former is found. To use NetBSD's curses, you now need `--disable-unicode --with-curses=curses` two configure options.

Explorer09:
I found one possible room for improvement:
In the static build of htop (see [this CI build log](https://github.com/htop-dev/htop/actions/runs/10258721630/job/28381996376)), `-ltinfo` is sometimes required (to be specified after `-lncursesw`) to link to ncurses. But our configure script detects `tinfo` after `curses`, which is probably of the cause of the configure error as seen in [this build log](https://github.com/htop-dev/htop/actions/runs/10245592060/job/28340914848) in PR #1506.

Apparently the `--libs` output from `pkg-config` should be self-contained (with all the dependency libraries listed), but the detection would matter when `pkg-config` is not available.

configure.ac
```diff
@@ -377,61 +377,199 @@ fi
...
+   ])
+
+   if test "$htop_config_libs_retrieved" = no; then
+      echo running "${curses_name}-config" for "$curses_name" >&AS_MESSAGE_LOG_FD
+      if "${curses_name}-config" --version >/dev/null 2>&AS_MESSAGE_LOG_FD; then
+         htop_config_cflags=`"${curses_name}-config" --cflags 2>/dev/null`
+         if test "$?" -eq 0; then
+            htop_config_libs=`"${curses_name}-config" --libs 2>/dev/null`
+            if test "$?" -eq 0; then
+               AC_MSG_NOTICE([found $curses_name configuration through ${curses_name}-config])
```
BenBE, line 534: Shouldn't this be `AC_MSG_RESULT`?

Explorer09: I was experimenting the output message when configuring. The idea was to show where we found the linker flags (CFLAGS and LIBS, in particular) and to leave a copy of the message in `config.log`. If I use `AC_MSG_RESULT`, then the message in `config.log` would have the `result:` prefix. The wording would look weird as in this example:

```text
result: found ncursesw6 configuration through ncursesw6-config
```

Do you have a better wording suggestion?

Or alternatively I can just use `echo text >&AS_MESSAGE_FD` so it prints the same line in standard output and in `config.log`.

BenBE: I did `AC_MSG_RESULT("found $LIBNAME ($METHOD)")` in #1506 and I think that's completely okay. Similar to what some of the other messages do too.

Explorer09: I just got the idea. The message could look like this in `config.log`...

```text
retrieving $curses_name configuration through $PKG_CONFIG
[pkg-config output]
result: found $curses_name configuration through $PKG_CONFIG
```

BenBE:
@Explorer: Please don't use `draft` in a commit message. Commits should aim to be "the final product". If the patch series overall is not ready for merging yet, that's what the "draft" status of PRs is for.

On the patch series: LGTM. Maybe switch the order of the last two commits. If there's no objection from the OpenEmbedded people, I think we can go ahead with this one.

Explorer09:
> Please don't use `draft` in a commit message. Commits should aim to be "the final product". If the patch series overall is not ready for merging yet, that's what the "draft" status of PRs is for.

I say "draft" in the commit because I don't think I'm done yet, and to prevent accidental pulling. Don't worry, I'll remove that tag when I think I'm ready.

configure.ac
```diff
@@ -383,61 +383,167 @@ m4_ifdef([PKG_PROG_PKG_CONFIG], [
...
-      AM_CFLAGS="$AM_CFLAGS $htop_config_script_cflags"
-      AC_CHECK_LIB([$1], [$2], [
-         AC_DEFINE([$3], 1, [The library is present.])
-         LIBS="$htop_config_script_libs $LIBS "
-         htop_script_success=yes
+], [with_curses=check])
+
+# $1: "CFLAGS" for curses
+# $2: linker flags for curses
+htop_check_libcurses_capability () {
```
Explorer09, line 420: This is perhaps the first time I introduce a shell function in the configure script, and I'm not sure the function naming and comment style id good enough.

1. [The Autoconf documentation](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.72/html_node/Shell-Functions.html) states that shell function might share the same namespace as shell variables, and to prevent name clashing and maximize portability, they use `as_fn_` prefix for shell functions. I'm not sure if we need this, by using `htop_fn_` prefix when naming this function, or I can stick to the current name and hope it won't collide with other symbols in the future.
2. The style of documenting function arguments. Should I change any of this?
```sh
# $1: "CFLAGS" for curses
# $2: linker flags for curses
```

configure.ac
```diff
@@ -383,61 +383,167 @@ m4_ifdef([PKG_PROG_PKG_CONFIG], [
...
+   for curses_name in $curses_pkg_names; do
+      echo trying simple linking with "-l$curses_name" >&AS_MESSAGE_LOG_FD
+      if htop_check_libcurses_capability "" "-l$curses_name"; then
+         break
+      fi
+   done
+fi
+
+if test "$htop_curses_pkg_found" = no; then
+   AC_MSG_ERROR([cannot link to curses library])
```
Explorer09, line 571: In the current `main` of `configure`, it might suggest users to specify `--disable-unicode` in the error message. My PR didn't yet have this usability feature. Should I add it?

configure.ac
BenBE: I think using `htop_` as prefix is fine.

For the documentation, I'd probably use either both full description (compiler flags / linker flags) or variables (CFLAGS / LDFLAGS).

configure.ac
BenBE: How much work is it to re-run the detection code with `--disable-unicode` if not provided explicitly in either direction? But as a first step, providing this hint might suffice.

Explorer09: @BenBE Not sure what you mean by "direction".

Since htop documentation already states that `--enable-unicode` would be the default if unspecified (it's `yes`, not `check`). I cannot make automatic fallback to `--disable-unicode` but to tell the user to specify such. Since `htop_check_libcurses_capability` checks _libncurses_ features first before _libncursesw_ features. Adding the `--disable-unicode` suggestion message is as easy as adding a shell variables noting that _libncurses_ is found but not _libncursesw_.

BenBE: Yes, I was refering to `check` vs `yes`/`no`.

Also, shouldn't it be `libncursesw` before `libncurses` (based on the old code)?

Explorer09: @BenBE The order swapping in `htop_check_libcurses_capability` is necessary to tell that, "oh, I found a working ncurses library, but this not doesn't support wide character functions".

BenBE: Ah, k.

Explorer09:
The configure script is revised and ready. However I got an unexpected configure error in OpenBSD build job in CI. It seems that OpenBSD build system in CI has `pkg-config` and `libncursesw`, but there is no `ncurses.pc` file describing the configuration. I can't tell if there is a package missing (to be installed) or if the system is designed that way.

BenBE:
There's also a failure for Solaris AFAICS. Any intel on that one?

Explorer09:
> There's also a failure for Solaris AFAICS. Any intel on that one?

@BenBE I was testing whether htop can build on all BSD systems and Solaris with `pkg-config` installed. Ideally I want the ncurses compile and linker flags all grabbed from `pkg-config`, but some systems don't seem to provide the `ncurses.pc` file for us to use.

.github/workflows/ci.yml
BenBE: @Explorer09 Can you add this note for NetBSD and other platforms where it applies?

Explorer09: Updated netbsd/README to mention the `--with-curses` option.

BenBE (approved):

BenBE:
Some minor notes on the current state of the patch:

```diff
diff --git a/configure.ac b/configure.ac
index 6347859f..7c9a745b 100644
--- a/configure.ac
+++ b/configure.ac
@@ -383,11 +383,13 @@ dnl PKG_PROG_PKG_CONFIG initializes $PKG_CONFIG and related variables.
 dnl If the macro is not called, some pkg-config checks might be skipped
 dnl and $PKG_CONFIG might be unset.
 m4_ifdef([PKG_PROG_PKG_CONFIG], [
-PKG_PROG_PKG_CONFIG
+   PKG_PROG_PKG_CONFIG()
 ], [
-pkg_m4_absent=1
-m4_warn([syntax],
-   [pkg.m4 is absent or older than version 0.16; this 'configure' would have incomplete pkg-config support])
+   pkg_m4_absent=1
+   m4_warn(
+      [syntax],
+      [pkg.m4 is absent or older than version 0.16; this 'configure' will have incomplete pkg-config support]
+   )
 ])
 
 AC_ARG_ENABLE([unicode],
@@ -565,14 +567,14 @@ esac
 
 case ${htop_curses_capability}-$enable_unicode in
 none-*|nonwide-yes)
-   if test "x$PKG_CONFIG" = x; then (
+   if test "x$PKG_CONFIG" = x; then
       # Friendly warning to advise user to install pkg-config.
       # The local variables get discarded when done.
       list=""
       echo "" >conftest.c
       if test "$cross_compiling" != yes; then
          # "-print-multi-directory" supported in GCC 3.0 or later
-        name=""
+         name=""
          if $CC $CPPFLAGS $CFLAGS -print-multi-directory -E conftest.c >/dev/null 2>&1; then
             name=`$CC $CPPFLAGS $CFLAGS -print-multi-directory -E conftest.c 2>/dev/null |
                sed '/^ *#/ d; s/^\.$//; q'`
@@ -610,7 +612,7 @@ none-*|nonwide-yes)
             break
          fi
       done
-   ) fi
+   fi
 
    # OpenBSD and Solaris are known to not provide '*curses*.pc' files.
    AC_MSG_RESULT([no curses information found through '*-config' utilities; will guess the linker flags])
@@ -796,7 +798,7 @@ AC_DEFINE_UNQUOTED([OSRELEASEFILE], ["$with_os_release"], [File with OS release
 
 AC_ARG_WITH([config],
             [AS_HELP_STRING([--with-config=DIR],
-                           [configuration path @<:@default=/.config@:>@])],
+                            [configuration path @<:@default=/.config@:>@])],
             [],
             [with_config="/.config"])
 dnl Performance Co-Pilot configuration location to prevent overwrite
diff --git a/netbsd/README.md b/netbsd/README.md
index c319c199..30b344a1 100644
--- a/netbsd/README.md
+++ b/netbsd/README.md
@@ -13,8 +13,8 @@ NetBSD is one of the last operating systems to use and maintain its own
 implementation of Curses.
 
 htop(1) can be compiled against either ncurses or NetBSD's curses(3).
-htop(1) would use ncurses by default when it is found, as NetBSD's curses
-support in htop is limited.
+By default, htop(1) will use ncurses when it is found, as support for
+NetBSD's curses in htop is limited.
 
 To use NetBSD's libcurses, htop(1) must be configured with `--disable-unicode`.
 Starting with htop 3.4.0, a new option `--with-curses=curses` may be specified
```

configure.ac
```diff
@@ -379,61 +379,270 @@ fi
...
+      if htop_check_curses_capability "$htop_config_cflags" "$htop_config_libs"; then
+         break
+      fi
+   done
+   ;;
+esac
 
+case ${htop_curses_capability}-$enable_unicode in
+none-*|nonwide-yes)
+   if test "x$PKG_CONFIG" = x; then (
```
Explorer09, line 568: Replying the [suggestion](https://github.com/htop-dev/htop/pull/1512#issuecomment-2359445533) from @BenBE.

The parentheses on that `if` block was intentional by me. This way I can isolate the local variables used within since all this block does is printing warning messages and it can all be done in a subshell.

BenBE: Reasonable. The way that scope is introduced just looks kinda strange without any annotation.

I'm also not sure if starting a subshell just for isolation is the right choice here or if  braces (`{}`) might do the trick too …

Explorer09: The `{}` won't isolate local variables like I had originally intended.
If it was braces `{}` block, then it's no different from having no braces at all.

I wonder if I can make this code more obvious by turning into a shell function.

PR #1513, opened 2024-08-09
LLVM 19 related updates

cgzones:


AffinityPanel.c
```diff
@@ -389,7 +389,7 @@ Panel* AffinityPanel_new(Machine* host, const Affinity* affinity, int* width) {
 
       char number[16];
       xSnprintf(number, 9, "CPU %d", Settings_cpuId(host->settings, i));
-      unsigned cpu_width = 4 + strlen(number);
+      unsigned cpu_width = 4 + (unsigned) strlen(number);
```
BenBE, line 392:
```suggestion
      unsigned cpu_width = 4 + (unsigned)strlen(number);
```

CommandLine.c
```diff
@@ -204,12 +205,15 @@ static CommandLineStatus parseArguments(int argc, char** argv, CommandLineSettin
             if (!username) {
                flags->userId = geteuid();
             } else if (!Action_setUserOnly(username, &(flags->userId))) {
-               for (const char* itr = username; *itr; ++itr)
-                  if (!isdigit((unsigned char)*itr)) {
+               char *endptr;
+               errno = 0;
+               unsigned long res = strtoul(username, &endptr, 10);
+               unsigned castRes = (unsigned) res;
```
BenBE, line 211:
```suggestion
               unsigned castRes = (unsigned)res;
```

CommandLine.c
```diff
@@ -204,12 +205,15 @@ static CommandLineStatus parseArguments(int argc, char** argv, CommandLineSettin
...
+               if (*endptr != '\0' || res == ULONG_MAX || errno != 0 || castRes != res) {
```
BenBE, line 212:
```suggestion
               if (*endptr != '\0' || res == ULONG_MAX || errno != 0 || res > UINT_MAX) {
```
or
```suggestion
               if (*endptr != '\0' || res > UINT_MAX || errno != 0) {
```

DynamicMeter.c
```diff
@@ -102,10 +102,11 @@ static void DynamicMeter_getUiName(const Meter* this, char* name, size_t length)
    if (meter) {
       const char* uiName = meter->caption;
       if (uiName) {
-         int len = strlen(uiName);
+         size_t len = strlen(uiName);
+         assert(len < 32);
```
BenBE, line 106: Given this is user-controlled via a config file, this should be a runtime check (DiD).

Process.c
```diff
@@ -85,8 +85,8 @@ static bool findCommInCmdline(const char* comm, const char* cmdline, int cmdline
 
       if ((tokenLen == commLen || (tokenLen > commLen && commLen == (TASK_COMM_LEN - 1))) &&
           strncmp(tokenBase, comm, commLen) == 0) {
-         *pCommStart = tokenBase - cmdline;
-         *pCommEnd = token - cmdline;
+         *pCommStart = (int)(tokenBase - cmdline);
+         *pCommEnd = (int)(token - cmdline);
```
BenBE, line 89: Given that these both refer to string offsets, we actually should switch to `size_t*` eventually (DiD).

Process.c
```diff
@@ -350,7 +350,7 @@ void Process_makeCommandStr(Process* this, const Settings* settings) {
       return;
    }
 
-   int exeLen = strlen(this->procExe);
+   int exeLen = (int)MINIMUM(strlen(this->procExe), 65536);
    int exeBasenameOffset = this->procExeBasenameOffset;
    int exeBasenameLen = exeLen - exeBasenameOffset;
```
BenBE, line 355: Switch to `size_t` would be better (DiD).

Process.c
```diff
@@ -867,16 +867,16 @@ static bool Process_setPriority(Process* this, int priority) {
...
-
 bool Process_rowChangePriorityBy(Row* super, Arg delta) {
    Process* this = (Process*) super;
    assert(Object_isA((const Object*) this, (const ObjectClass*) &Process_class));
-   return Process_setPriority(this, this->nice + delta.i);
+
+   long value = this->nice + delta.i;
+   int prio = (int)value;
+   if (prio != value)
+      return false;
```
BenBE, line 877: Isn't this overflow check UB?

ScreenTabsPanel.c
```diff
@@ -210,7 +210,7 @@ static HandlerResult ScreenNamesPanel_eventHandlerRenaming(Panel* super, int ch)
       if (this->cursor < SCREEN_NAME_LEN - 1) {
          this->buffer[this->cursor] = (char)ch;
          this->cursor++;
-         super->selectedLen = strlen(this->buffer);
+         super->selectedLen = (int) strlen(this->buffer);
```
BenBE, line 213:
```suggestion
         super->selectedLen = (int)strlen(this->buffer);
```

ScreenTabsPanel.c
```diff
@@ -223,7 +223,7 @@ static HandlerResult ScreenNamesPanel_eventHandlerRenaming(Panel* super, int ch)
          if (this->cursor > 0) {
             this->cursor--;
             this->buffer[this->cursor] = '\0';
-            super->selectedLen = strlen(this->buffer);
+            super->selectedLen = (int) strlen(this->buffer);
```
BenBE, line 226:
```suggestion
            super->selectedLen = (int)strlen(this->buffer);
```

ScreenTabsPanel.c
```diff
@@ -271,10 +271,10 @@ static void startRenaming(Panel* super) {
    this->saved = name;
    strncpy(this->buffer, name, SCREEN_NAME_LEN);
    this->buffer[SCREEN_NAME_LEN] = '\0';
-   this->cursor = strlen(this->buffer);
+   this->cursor = (int) strlen(this->buffer);
```
BenBE, line 274:
```suggestion
   this->cursor = (int)strlen(this->buffer);
```

ScreenTabsPanel.c
```diff
@@ -271,10 +271,10 @@ static void startRenaming(Panel* super) {
...
    item->value = this->buffer;
    Panel_setSelectionColor(super, PANEL_EDIT);
-   super->selectedLen = strlen(this->buffer);
+   super->selectedLen = (int) strlen(this->buffer);
```
BenBE, line 277:
```suggestion
   super->selectedLen = (int)strlen(this->buffer);
```

ScreensPanel.c
```diff
@@ -69,7 +69,7 @@ static HandlerResult ScreensPanel_eventHandlerRenaming(Panel* super, int ch) {
       if (this->cursor < SCREEN_NAME_LEN - 1) {
          this->buffer[this->cursor] = (char)ch;
          this->cursor++;
-         super->selectedLen = strlen(this->buffer);
+         super->selectedLen = (int) strlen(this->buffer);
```
BenBE, line 72:
```suggestion
         super->selectedLen = (int)strlen(this->buffer);
```

ScreensPanel.c
```diff
@@ -82,7 +82,7 @@ static HandlerResult ScreensPanel_eventHandlerRenaming(Panel* super, int ch) {
          if (this->cursor > 0) {
             this->cursor--;
             this->buffer[this->cursor] = '\0';
-            super->selectedLen = strlen(this->buffer);
+            super->selectedLen = (int) strlen(this->buffer);
```
BenBE, line 85:
```suggestion
            super->selectedLen = (int)strlen(this->buffer);
```

ScreensPanel.c
```diff
@@ -129,10 +129,10 @@ static void startRenaming(Panel* super) {
    this->saved = name;
    strncpy(this->buffer, name, SCREEN_NAME_LEN);
    this->buffer[SCREEN_NAME_LEN] = '\0';
-   this->cursor = strlen(this->buffer);
+   this->cursor = (int) strlen(this->buffer);
```
BenBE, line 132:
```suggestion
   this->cursor = (int)strlen(this->buffer);
```

ScreensPanel.c
```diff
@@ -129,10 +129,10 @@ static void startRenaming(Panel* super) {
...
    item->value = this->buffer;
    Panel_setSelectionColor(super, PANEL_EDIT);
-   super->selectedLen = strlen(this->buffer);
+   super->selectedLen = (int) strlen(this->buffer);
```
BenBE, line 135:
```suggestion
   super->selectedLen = (int)strlen(this->buffer);
```

Settings.c
```diff
@@ -668,7 +668,7 @@ static int signal_safe_fprintf(FILE* stream, const char* fmt, ...) {
    if (n <= 0)
       return n;
 
-   return full_write_str(fileno(stream), buf);
+   return (int) full_write_str(fileno(stream), buf);
```
BenBE, line 671:
```suggestion
   return (int)full_write_str(fileno(stream), buf);
```

linux/LinuxMachine.c
```diff
@@ -654,7 +656,7 @@ Machine* Machine_new(UsersTable* usersTable, uid_t userId) {
    Machine_init(super, usersTable, userId);
 
    // Initialize page size
-   if ((this->pageSize = sysconf(_SC_PAGESIZE)) == -1)
+   if ((this->pageSize = (int) sysconf(_SC_PAGESIZE)) == -1)
```
BenBE, line 659:
```suggestion
   if ((this->pageSize = (int)sysconf(_SC_PAGESIZE)) == -1)
```

linux/LinuxProcess.c
```diff
@@ -162,7 +162,7 @@ IOPriority LinuxProcess_updateIOPriority(Process* p) {
    IOPriority ioprio = 0;
 // Other OSes masquerading as Linux (NetBSD?) don't have this syscall
 #ifdef SYS_ioprio_get
-   ioprio = syscall(SYS_ioprio_get, IOPRIO_WHO_PROCESS, Process_getPid(p));
+   ioprio = (int) syscall(SYS_ioprio_get, IOPRIO_WHO_PROCESS, Process_getPid(p));
```
BenBE, line 165:
```suggestion
   ioprio = (int)syscall(SYS_ioprio_get, IOPRIO_WHO_PROCESS, Process_getPid(p));
```

linux/LinuxProcessTable.c
```diff
@@ -334,23 +334,23 @@ static bool LinuxProcessTable_readStatFile(LinuxProcess* lp, openat_arg_t procFd
    location += 2;
 
    /* (4) ppid  -  %d */
-   Process_setParent(process, fast_strtol_dec(&location, 0));
+   Process_setParent(process, (pid_t) fast_strtol_dec(&location, 0));
```
BenBE, line 337:
```suggestion
   Process_setParent(process, (pid_t)fast_strtol_dec(&location, 0));
```

linux/LinuxProcessTable.c
```diff
@@ -334,23 +334,23 @@ static bool LinuxProcessTable_readStatFile(LinuxProcess* lp, openat_arg_t procFd
...
    location += 1;
 
    /* (5) pgrp  -  %d */
-   process->pgrp = fast_strtol_dec(&location, 0);
+   process->pgrp = (int) fast_strtol_dec(&location, 0);
```
BenBE, line 341:
```suggestion
   process->pgrp = (int)fast_strtol_dec(&location, 0);
```

linux/LinuxProcessTable.c
```diff
@@ -334,23 +334,23 @@ static bool LinuxProcessTable_readStatFile(LinuxProcess* lp, openat_arg_t procFd
...
    location += 1;
 
    /* (6) session  -  %d */
-   process->session = fast_strtol_dec(&location, 0);
+   process->session = (int) fast_strtol_dec(&location, 0);
```
BenBE, line 345:
```suggestion
   process->session = (int)fast_strtol_dec(&location, 0);
```

linux/LinuxProcessTable.c
```diff
@@ -334,23 +334,23 @@ static bool LinuxProcessTable_readStatFile(LinuxProcess* lp, openat_arg_t procFd
...
    location += 1;
 
    /* (7) tty_nr  -  %d */
    process->tty_nr = fast_strtoul_dec(&location, 0);
    location += 1;
 
    /* (8) tpgid  -  %d */
-   process->tpgid = fast_strtol_dec(&location, 0);
+   process->tpgid = (int) fast_strtol_dec(&location, 0);
```
BenBE, line 353:
```suggestion
   process->tpgid = (int)fast_strtol_dec(&location, 0);
```

linux/LinuxProcessTable.c
```diff
@@ -420,7 +420,7 @@ static bool LinuxProcessTable_readStatFile(LinuxProcess* lp, openat_arg_t procFd
    assert(location != NULL);
 
    /* (39) processor  -  %d */
-   process->processor = fast_strtol_dec(&location, 0);
+   process->processor = (int) fast_strtol_dec(&location, 0);
```
BenBE, line 423:
```suggestion
   process->processor = (int)fast_strtol_dec(&location, 0);
```

linux/LinuxProcessTable.c
```diff
@@ -648,7 +648,7 @@ static void LinuxProcessTable_readMaps(LinuxProcess* process, openat_arg_t procF
       bool map_execute;
       unsigned int map_devmaj;
       unsigned int map_devmin;
-      uint64_t map_inode;
+      unsigned int map_inode;
```
BenBE, line 651: Why reduce the width of the inode ID? Docs?

linux/LinuxProcessTable.c
```diff
@@ -678,19 +678,19 @@ static void LinuxProcessTable_readMaps(LinuxProcess* process, openat_arg_t procF
       if (' ' != *readptr++)
          continue;
 
-      map_devmaj = fast_strtoull_hex(&readptr, 4);
+      map_devmaj = (unsigned int) fast_strtoull_hex(&readptr, 4);
```
BenBE, line 681:
```suggestion
      map_devmaj = (unsigned int)fast_strtoull_hex(&readptr, 4);
```

linux/LinuxProcessTable.c
```diff
@@ -678,19 +678,19 @@ static void LinuxProcessTable_readMaps(LinuxProcess* process, openat_arg_t procF
...
       if (':' != *readptr++)
          continue;
 
-      map_devmin = fast_strtoull_hex(&readptr, 4);
+      map_devmin = (unsigned int) fast_strtoull_hex(&readptr, 4);
```
BenBE, line 685:
```suggestion
      map_devmin = (unsigned int)fast_strtoull_hex(&readptr, 4);
```

linux/LinuxProcessTable.c
```diff
@@ -678,19 +678,19 @@ static void LinuxProcessTable_readMaps(LinuxProcess* process, openat_arg_t procF
...
       if (' ' != *readptr++)
          continue;
 
-      //Minor shortcut: Once we know there's no file for this region, we skip
+      // Minor shortcut: Once we know there's no file for this region, we skip
       if (!map_devmaj && !map_devmin)
          continue;
 
-      map_inode = fast_strtoull_dec(&readptr, 0);
+      map_inode = (unsigned int) fast_strtoull_dec(&readptr, 0);
```
BenBE, line 693:
```suggestion
      map_inode = (unsigned int)fast_strtoull_dec(&readptr, 0);
```

linux/LinuxProcessTable.c
```diff
@@ -795,12 +795,11 @@ static bool LinuxProcessTable_readSmapsFile(LinuxProcess* process, openat_arg_t
...
-         while (fgets(buffer, sizeof(buffer), fp)) {
-            if (strchr(buffer, '\n')) {
-               break;
-            }
+         if (!skipEndOfLine(fp)) {
+            fclose(fp);
+            return false;
          }
-         continue;
+
```
BenBE, line 802: No blank line here …

linux/LinuxProcessTable.c
```diff
@@ -1015,13 +1011,13 @@ static void LinuxProcessTable_readOomData(LinuxProcess* process, openat_arg_t pr
 
    char buffer[PROC_LINE_LENGTH + 1] = {0};
 
-   ssize_t oomRead = xReadfileat(procFd, "oom_score", buffer, sizeof(buffer));
+   int oomRead = (int) xReadfileat(procFd, "oom_score", buffer, sizeof(buffer));
```
BenBE, line 1014: Use `ssize_t` instead.

linux/LinuxProcessTable.c
```diff
@@ -1250,10 +1246,10 @@ static bool LinuxProcessTable_readCmdlineFile(Process* process, openat_arg_t pro
       tokenStart = -1;
       tokenEnd = -1;
 
-      size_t exeLen = process->procExe ? strlen(process->procExe) : 0;
+      int exeLen = process->procExe ? (int) strlen(process->procExe) : 0;
```
BenBE, line 1249: What's wrong with `size_t` here?

Vector.c
```diff
@@ -69,8 +69,8 @@ static bool Vector_isConsistent(const Vector* this) {
    return true;
 }
 
-bool Vector_countEquals(const Vector* this, unsigned int expectedCount) {
-   unsigned int n = 0;
+bool Vector_countEquals(const Vector* this, size_t expectedCount) {
+   size_t n = 0;
    for (int i = 0; i < this->items; i++) {
```
BenBE, line 74: Shouldn't this be updated to `size_t` here too (I know this would add a cast to `this->items`).

BenBE (changes requested):
Not really happy with some of the changes. Can you provide a set of compiler messages describing why certain changes were made?

CommandLine.c
Explorer09, line 212: Before I can suggest a better code for this, I need one behavior clarified:
Is `username` being an empty string acceptable here?
Because `strtoul` would not produce an error for the empty string. It would happily return 0 as the result.

Process.c
Explorer09, line 89: Is there any reason not to use `ptrdiff_t` or `ssize_t` here?

Row.c
```diff
@@ -495,7 +495,8 @@ void Row_printRate(RichString* str, double rate, bool coloring) {
 
 void Row_printLeftAlignedField(RichString* str, int attr, const char* content, unsigned int width) {
    int columns = width;
-   RichString_appendnWideColumns(str, attr, content, strlen(content), &columns);
+   size_t contentLen = strlen(content);
+   RichString_appendnWideColumns(str, attr, content, (int)MINIMUM(contentLen, 1000), &columns);
```
Explorer09, line 499: I'm not sure I like this construct. If you want the cap the maximum length for a `strlen` output, then `strnlen` is there for use.

CommandLine.c
Explorer09: @BenBE  I would say `res >= UINT_MAX` because, AFAIK, (uid_t)-1 is invalid.

Process.c
Explorer09: The type after the substraction is technically `ptrdiff_t`. Casting to `int` would lose upper bits. Also, it's a signed type.

Process.c
Explorer09: Just to mention I don't like arbitrary limit here. If this `procExe` was for display, we can cut the length to as short as `PATH_MAX`, or is there a case where `PATH_MAX` is not enough?

Meter.h
```diff
@@ -111,7 +111,7 @@ struct Meter_ {
    unsigned int param;
    GraphData drawData;
    int h;
-   int columnWidthCount;      /**< only used internally by the Header */
+   size_t columnWidthCount;      /**< only used internally by the Header */
```
Explorer09, line 114: Can't understand the reason why expanding the bit width of this `columnWidthCount` to 64. Would it save code size by saving some cast instructions?

Row.c
```diff
@@ -428,22 +428,22 @@ void Row_printNanoseconds(RichString* str, unsigned long long totalNanoseconds,
    }
 
    unsigned long long totalSeconds = totalMicroseconds / 1000000;
-   unsigned long microseconds = totalMicroseconds % 1000000;
+   unsigned int microseconds = totalMicroseconds % 1000000;
```
Explorer09, line 431: When I wrote this piece of `Row_printNanoseconds` code, I _did not_ assume the `unsigned int` type would have 32 bits for you (for stricter standard compliance). I know what you are doing, but I would like `uint32_t` be used here instead of `unsigned int`.

Row.c
```diff
@@ -428,22 +428,22 @@ void Row_printNanoseconds(RichString* str, unsigned long long totalNanoseconds,
...
    if (totalSeconds < 60) {
       int width = 5;
-      unsigned long fraction = microseconds / 10;
+      unsigned int fraction = microseconds / 10;
       if (totalSeconds >= 10) {
          width--;
          fraction /= 10;
       }
-      len = xSnprintf(buffer, sizeof(buffer), "%u.%0*lus ", (unsigned int)totalSeconds, width, fraction);
+      len = xSnprintf(buffer, sizeof(buffer), "%u.%0*us ", (unsigned int)totalSeconds, width, fraction);
```
Explorer09, line 439: I would like this `(unsigned long)` cast in `snprintf` function be preserved unless you use `PRIu32` format instead.

linux/LinuxMachine.c
```diff
@@ -74,7 +74,7 @@ static void LinuxMachine_updateCPUcount(LinuxMachine* this) {
 
       char* endp;
       unsigned long int id = strtoul(entry->d_name + 3, &endp, 10);
-      if (id == ULONG_MAX || endp == entry->d_name + 3 || *endp != '\0')
+      if (id == UINT_MAX || endp == entry->d_name + 3 || *endp != '\0')
```
Explorer09, line 77:
```suggestion
      if (id >= UINT_MAX || endp == entry->d_name + 3 || *endp != '\0')
Comment

```

linux/LinuxProcessTable.c
```diff
@@ -1015,13 +1015,13 @@ static void LinuxProcessTable_readOomData(LinuxProcess* process, openat_arg_t pr
...
 
-   ssize_t oomRead = xReadfileat(procFd, "oom_score", buffer, sizeof(buffer));
+   int oomRead = (int) xReadfileat(procFd, "oom_score", buffer, sizeof(buffer));
    if (oomRead < 1) {
       return;
    }
 
    char* oomPtr = buffer;
-   uint64_t oom = fast_strtoull_dec(&oomPtr, oomRead);
+   unsigned long oom = fast_strtoul_dec(&oomPtr, oomRead);
```
Explorer09, line 1020: We shouldn't use `fast_strou*_dec` here, as these "fast" versions come with no overflow check.
By looking at the lines below, it seems like the overflow check is needed.

linux/LinuxProcessTable.c
```diff
@@ -1485,9 +1485,9 @@ static bool LinuxProcessTable_recurseProcTree(LinuxProcessTable* this, openat_ar
       {
          char* endptr;
          unsigned long parsedPid = strtoul(name, &endptr, 10);
-         if (parsedPid == 0 || parsedPid == ULONG_MAX || *endptr != '\0')
+         if (parsedPid == 0 || parsedPid >= INT_MAX || *endptr != '\0')
```
Explorer09, line 1484: The logic doesn't look right here. Why would `INT_MAX` become an invalid PID anyway? If we want to check for overflow, it could be `parsedPid > INT_MAX || !errno`.

linux/LinuxProcessTable.c
Explorer09: I would say I am a little tired of these `fast_strtol_dec` calls with `(int)` downcasts. Can we introduce a `fast_atoi_dec` and call it a day? The function can be modelled after `fast_strtol_dec`:

```c
static int fast_atoi_dec(char** str, int maxlen) {
   long result = fast_strtol_dec(str, maxlen);
   assert(result <= INT_MAX);
   assert(result >= INT_MIN);
   return (int)result;
}
```

CommandLine.c
BenBE: That's something the current version didn't check for …

Process.c
BenBE: Technically yes, but since `tokenBase > cmdline` and `token > cmdline`, both are positive and a cast to `size_t` wouldn't cause trouble with the sign bit. I'd have to check re sentinel values, but that's another concern.

CRT.c
```diff
@@ -1260,8 +1260,7 @@ static void print_backtrace(void) {
 #elif defined(HAVE_EXECINFO_H)
    void* backtraceArray[256];
 
-   size_t size = backtrace(backtraceArray, ARRAYSIZE(backtraceArray));
```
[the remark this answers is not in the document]
BenBE: Whenever possible and unconstrained by external API, I'd actually prefer switching towards `size_t` whenever reasonable.

Row.c
BenBE: Also in favour of explicit bit width spec.

linux/LinuxProcessTable.c
BenBE: `strtoul` returns `unsigned long`, but the result is later assigned to a field of `unsigned int`. Thus anything greater than `UINT_MAX` causes an integer overflow. Everything equal to `UINT_MAX` is the sentinel value for "invalid user". Thus `parsedPid >= UINT_MAX` to catch both cases at once.

CRT.c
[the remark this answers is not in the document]
Explorer09: @BenBE Well, for this `backtrace()` case, it returns `int`, and thus it's meaningless to upcast to `size_t`.

But eliminating a local variable might be an overkill here. That's why I suggest keeping that `size` var but just define it as `int` type.

linux/LinuxProcessTable.c
Explorer09: @BenBE This is PID, not UID. Please get the right context. POSIX didn't have a clear limit on the maximum value of PID, or that PID == INT_MAX is always invalid. So I want to play safe here.

Looking at [this Stack Exchange question](https://unix.stackexchange.com/questions/16883/), it look like Linux currently has a hard limit of 4194303 (the maximum PID is *one less* than the value in `/proc/sys/kernel/pid_max`). And that no other OS had been using INT_MAX as maximum PID yet, so treating INT_MAX as invalid value may be fine, for now.

Explorer09: By the way, in Unix, `pid_t` is always a signed type, unlike `uid_t`. The `pid_t` is required to be signed as [fork(2)](https://man7.org/linux/man-pages/man2/fork.2.html) returns value in this type and that negative values are reserved for special uses.

Explorer09:
Since there are changes in this PR that look not trivial. Would you guys mind if I pick some of the easier changes and file a new PR for them? I just hope some of the changes can be merged early.

Action.c
```diff
@@ -401,15 +401,19 @@ static Htop_Reaction actionPrevScreen(State* st) {
 }
 
 Htop_Reaction Action_setScreenTab(State* st, int x) {
+   assert(x >= 0);
+   if (x < 0)
+      return 0;
+   unsigned int pos = x;
    Settings* settings = st->host->settings;
-   int s = 2;
+   unsigned int s = 2;
```
Explorer09, line 409: I think for this algorithm, it's better for `s` to be in `size_t`, and use `(size_t)x` (upcast) for comparisons below. I think I would submit my propsed change in a separate PR.

DynamicMeter.c
Explorer09: What does the colon in this `uiName` string do? And may I replace this `strlen` call with `(size_t)(String_strchrnul(uiName, ':') - uiName)` ?

BenBE:
Most of the non-trivial changes in this PR boil down to structural refactorings where many fields that denote size values still use int instead of the more appropriate `size_t`. Unfortunately, if you start this, this goes through the codebase at least once. Having this as a separate PR would be good. Also splitting other non-trivial changes out is much appreciated.

CRT.c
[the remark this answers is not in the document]
Explorer09: This particular one is resolved by #1516 and is cherry-picked to `main`.

Row.c
Explorer09: Update: I've changed my mind on the use of `strnlen` here. While `strnlen` can guarantee it never reads past the specified `maxlen`, `strnlen` is (theoretically) less performant than `strlen` when the input string is well formed already.

That doesn't mean I deny the possibility of using `strnlen` in htop, but the length clamping, if needed, should be done as early as possible, e.g. in `Settings.c` when reading the strings from config and not long after the config file was read.

Just using `size_t` for `strlen` return values would be slightly cheaper than usijng strnlen`

Process.c
Explorer09: This Process.c specific changes for LLVM 19 compatibility may be superseded by #1588.

Action.c
Explorer09: Oops. When I re-review this function code just now, I discovered that the use of `strlen` here was wrong at the start. This should have been "terminal width" instead and use a function similar to `wcswidth` but working with multi-byte string instead.

BenBE:
I moved this PR and #1588 to 3.5.0 because 3.4.1 is targeted to be a minor bugfix release with little refactoring where possible. As these two PRs both carry quite some code changes that are hard to justify as minor fixes, these will have to wait a bit. Also I'm not really happy with the amount of change that is introduced in these two PRs. Please check if we can manage these fixes with a little bit more structured approach.

@Explorer09: Since @cgzones is kinda inactive you may continue the changes of this PR in a superseded implementation, unless there's an objection. If you do, please be mindful to the one who has to review the refactoring. ;-)

PR #1516, opened 2024-08-12
Partial code improvements with respect to Clang 19 warnings

Explorer09:
This branch contains my proposal of the code quality changes with respect to Clang 19 warnings. (Notably, `-Wshorten-64-to-32` is enabled by default in Clang 19.) I would try to break these into small changes so that these commits may be cherry-picked into `main` early.

(This pull request is related to #1513)

### DIFF
--- a/CategoriesPanel.c
+++ b/CategoriesPanel.c
@@ -42,13 +42,13 @@ static void CategoriesPanel_delete(Object* object) {
 }
 
 static void CategoriesPanel_makeMetersPage(CategoriesPanel* this) {
-   size_t columns = HeaderLayout_getColumns(this->scr->header->headerLayout);
+   unsigned int columns = HeaderLayout_getColumns(this->scr->header->headerLayout);
    MetersPanel** meterPanels = xMallocArray(columns, sizeof(MetersPanel*));
    Settings* settings = this->host->settings;
 
-   for (size_t i = 0; i < columns; i++) {
+   for (unsigned int i = 0; i < columns; i++) {
       char titleBuffer[32];
-      xSnprintf(titleBuffer, sizeof(titleBuffer), "Column %zu", i + 1);
+      xSnprintf(titleBuffer, sizeof(titleBuffer), "Column %u", i + 1);
       meterPanels[i] = MetersPanel_new(settings, titleBuffer, this->header->columns[i], this->scr);
 
       if (i != 0) {
--- a/Meter.c
+++ b/Meter.c
@@ -49,16 +49,19 @@ static inline void Meter_displayBuffer(const Meter* this, RichString* out) {
 /* ---------- TextMeterMode ---------- */
 
 static void TextMeterMode_draw(Meter* this, int x, int y, int w) {
+   if (w <= 0)
+      return;
+
    const char* caption = Meter_getCaption(this);
    attrset(CRT_colors[METER_TEXT]);
    mvaddnstr(y, x, caption, w);
    attrset(CRT_colors[RESET_COLOR]);
 
-   int captionLen = strlen(caption);
-   x += captionLen;
-   w -= captionLen;
-   if (w <= 0)
+   size_t captionLen = strlen(caption);
+   if ((size_t)w <= captionLen)
       return;
+   x += (int)captionLen;
+   w -= (int)captionLen;
 
    RichString_begin(out);
    Meter_displayBuffer(this, &out);
@@ -301,6 +304,10 @@ static void LEDMeterMode_drawDigit(int x, int y, int n) {
 }
 
 static void LEDMeterMode_draw(Meter* this, int x, int y, int w) {
+   assert(x >= 0);
+   if (w <= 0)
+      return;
+
 #ifdef HAVE_LIBNCURSESW
    if (CRT_utf8)
       LEDMeterMode_digits = LEDMeterMode_digitsUtf8;
@@ -318,25 +325,25 @@ static void LEDMeterMode_draw(Meter* this, int x, int y, int w) {
       y + 2;
    attrset(CRT_colors[LED_COLOR]);
    const char* caption = Meter_getCaption(this);
-   mvaddstr(yText, x, caption);
-   int xx = x + strlen(caption);
+   mvaddnstr(yText, x, caption, w);
+   size_t xx = (size_t)x + strlen(caption);
    int len = RichString_sizeVal(out);
-   for (int i = 0; i < len; i++) {
+   for (size_t i = 0; i < (size_t)len; i++) {
       int c = RichString_getCharVal(out, i);
       if (c >= '0' && c <= '9') {
-         if (xx - x + 4 > w)
+         if (xx + 4 > (unsigned int)x + (unsigned int)w)
             break;
 
-         LEDMeterMode_drawDigit(xx, y, c - '0');
+         LEDMeterMode_drawDigit((int)xx, y, c - '0');
          xx += 4;
       } else {
-         if (xx - x + 1 > w)
+         if (xx + 1 > (unsigned int)x + (unsigned int)w)
             break;
 #ifdef HAVE_LIBNCURSESW
          const cchar_t wc = { .chars = { c, '\0' }, .attr = 0 }; /* use LED_COLOR from attrset() */
-         mvadd_wch(yText, xx, &wc);
+         mvadd_wch(yText, (int)xx, &wc);
 #else
-         mvaddch(yText, xx, c);
+         mvaddch(yText, (int)xx, c);
 #endif
          xx += 1;
       }
--- a/Row.c
+++ b/Row.c
@@ -428,23 +428,23 @@ void Row_printNanoseconds(RichString* str, unsigned long long totalNanoseconds, 
    }
 
    unsigned long long totalSeconds = totalMicroseconds / 1000000;
-   unsigned long microseconds = totalMicroseconds % 1000000;
+   uint32_t microseconds = totalMicroseconds % 1000000;
    if (totalSeconds < 60) {
       int width = 5;
-      unsigned long fraction = microseconds / 10;
+      uint32_t fraction = microseconds / 10;
       if (totalSeconds >= 10) {
          width--;
          fraction /= 10;
       }
-      len = xSnprintf(buffer, sizeof(buffer), "%u.%0*lus ", (unsigned int)totalSeconds, width, fraction);
+      len = xSnprintf(buffer, sizeof(buffer), "%u.%0*lus ", (unsigned int)totalSeconds, width, (unsigned long)fraction);
       RichString_appendnAscii(str, baseColor, buffer, len);
       return;
    }
 
    if (totalSeconds < 600) {
-      unsigned int minutes = totalSeconds / 60;
-      unsigned int seconds = totalSeconds % 60;
-      unsigned int milliseconds = microseconds / 1000;
+      unsigned int minutes = (unsigned int)totalSeconds / 60;
+      unsigned int seconds = (unsigned int)totalSeconds % 60;
+      unsigned int milliseconds = (unsigned int)(microseconds / 1000);
       len = xSnprintf(buffer, sizeof(buffer), "%u:%02u.%03u ", minutes, seconds, milliseconds);
       RichString_appendnAscii(str, baseColor, buffer, len);
       return;
--- a/Header.c
+++ b/Header.c
@@ -51,8 +51,8 @@ void Header_delete(Header* this) {
 }
 
 void Header_setLayout(Header* this, HeaderLayout hLayout) {
-   size_t oldColumns = HeaderLayout_getColumns(this->headerLayout);
-   size_t newColumns = HeaderLayout_getColumns(hLayout);
+   unsigned int oldColumns = HeaderLayout_getColumns(this->headerLayout);
+   unsigned int newColumns = HeaderLayout_getColumns(hLayout);
 
    this->headerLayout = hLayout;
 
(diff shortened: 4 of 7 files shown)


CRT.c
```diff
@@ -1260,8 +1260,9 @@ static void print_backtrace(void) {
 #elif defined(HAVE_EXECINFO_H)
    void* backtraceArray[256];
 
-   size_t size = backtrace(backtraceArray, ARRAYSIZE(backtraceArray));
-   backtrace_symbols_fd(backtraceArray, size, STDERR_FILENO);
+   int nptrs = backtrace(backtraceArray, ARRAYSIZE(backtraceArray));
+   assert(nptrs >= 0);
```
BenBE, line 1264: If you check this assert, do proper error handling instead by writing "Backtrace failed" or alike.

CPUMeter.c
```diff
@@ -55,6 +55,8 @@ static void CPUMeter_init(Meter* this) {
 
 // Custom uiName runtime logic to include the param (processor)
 static void CPUMeter_getUiName(const Meter* this, char* buffer, size_t length) {
+   assert(length > 0);
+
```
BenBE, line 59: Using `ATTR_NONNULL_ALL ATTR_ACCESS_RW(2,3)` would have the same effect at compile-time.

Similar with the other `*_UiName` functions.

DynamicMeter.c
```diff
@@ -97,15 +98,18 @@ static const char* DynamicMeter_getCaption(const Meter* this) {
...
    if (meter) {
       const char* uiName = meter->caption;
       if (uiName) {
-         int len = strlen(uiName);
+         size_t len = strlen(uiName);
          if (len > 2 && uiName[len - 2] == ':')
             len -= 2;
-         xSnprintf(name, length, "%.*s", len, uiName);
+
+         xSnprintf(name, MINIMUM(length, len + 1), "%s", uiName);
```
BenBE, line 112: Why this change?

Explorer09: @BenBE The "precision" part of the format specifier takes type `int` only and not `size_t`. Since I am processing string length now in `size_t`, changing the length argument to this avoids an `int` downcast and a `len <= INT_MAX` assumption.

CRT.c
Explorer09: @BenBE  What do you mean? `backtrace(3)` never documents it can return a negative value, so I won't know what error it can produce. `nptrs == 0` might be a good case to handle by letting it print something rather than nothing.

DynamicMeter.c
BenBE: Couldn't this be written as a `xStrncpy` in that case instead?

Explorer09: @BenBE You mean `String_safeStrncpy`, but good call anyway.

DynamicMeter.c
```diff
@@ -97,17 +98,20 @@ static const char* DynamicMeter_getCaption(const Meter* this) {
...
    if (meter) {
       const char* uiName = meter->caption;
       if (uiName) {
-         int len = strlen(uiName);
+         size_t len = strlen(uiName);
          if (len > 2 && uiName[len - 2] == ':')
             len -= 2;
-         xSnprintf(name, length, "%.*s", len, uiName);
+
+         String_safeStrncpy(name, uiName, MINIMUM(length, len + 1));
```
BenBE, line 112: The final `+1` doesn't seem necessary as `String_safeStrncpy` enforces NUL-termination of the destination buffer.

BenBE: Also, for `len==0` (AKA empty caption) this is UB.

Explorer09: @BenBE The length in `String_safeStrncpy` includes the terminating NUL byte while `strlen` return value does not include it. Hence, the `+1` is intentional and required.

And yes, `len==0` means you get an empty string copied (i.e. just one NUL byte). No UB.

CPUMeter.c
Explorer09: @BenBE Since `getUiName` functions are called through pointers, the function attribute would be best specified on pointers and not on individual functions.

And this could be the first time in htop codebase that we specify attributes on typedefs. I need a style to be decided, while in build tests all three styles work:

```c
// A
ATTR_NONNULL ATTR_ACCESS3_W(2, 3) typedef void(*Meter_GetUiName)(const Meter*, char*, size_t);
// B
typedef ATTR_NONNULL ATTR_ACCESS3_W(2, 3) void(*Meter_GetUiName)(const Meter*, char*, size_t);
// C
typedef void(*Meter_GetUiName)(const Meter*, char*, size_t) ATTR_NONNULL ATTR_ACCESS3_W(2, 3);
```

I would personally go with B, but I'd like to hear other people's opinions.

DynamicMeter.c
BenBE: Confusion on my side. I misread the implementation of `String_safeStrncpy` regarding the termination for how much to copy from the source buffer, thus the "no need for the +1"; the `len==0` was a follow-up on that mistake. My bad.

CPUMeter.c
BenBE: I'd also suggest B, which is mot consistent with what we do for normal functions.

CategoriesPanel.c
```diff
@@ -42,13 +42,13 @@ static void CategoriesPanel_delete(Object* object) {
...
-   size_t columns = HeaderLayout_getColumns(this->scr->header->headerLayout);
+   unsigned int columns = HeaderLayout_getColumns(this->scr->header->headerLayout);
    MetersPanel** meterPanels = xMallocArray(columns, sizeof(MetersPanel*));
    Settings* settings = this->host->settings;
 
-   for (size_t i = 0; i < columns; i++) {
+   for (unsigned int i = 0; i < columns; i++) {
       char titleBuffer[32];
-      xSnprintf(titleBuffer, sizeof(titleBuffer), "Column %zu", i + 1);
+      xSnprintf(titleBuffer, sizeof(titleBuffer), "Column %u", i + 1);
```
BenBE, line 51: I think this one is a step backwards. Stick witz `size_t`.

Explorer09: Can't understand what you mean. What's the reason of keeping the `size_t`?

BenBE: Two-fold:

1. It's a number of items (columns)
2. `unsigned int` looks kinda unwieldy IMHO.
3. For the format string `%zu` is much more portable than `uint32_t` which would require you to use `%"PRIu32"` instead of just `%u`.

BenBE: Personal opinion following:

> @BenBE Well, I have a different view on this:
> 
> 1. Object sizes with no likelihood of overflowing an `int` type would not need `size_t` (that is, `unsigned long`) for storing.

Nonetheless `size_t` is less about the number of entities itself, but that you are **counting** entities.

Otherwise you could argue that "this hashtable implementation only needs `short` for its size, because I never store more than 65535 elements in it".

I know, that example is hyperbole, but follows that same logic. When I ask for `size_t` it's really little about "how many", but the "what" (AKA semantics).

> 2. In x86-64, incrementing and comparing `int` types are slightly cheaper than comparing `long` or `size_t` (in terms of code size). When `int` types can suffice, using `long` would make the code one-byte-per-operation larger.

That might be true for x86_64, but that's not the only platform we need to support. There's at least ARM, ARM64, MIPS, and probably several others, where htop is runing. Unless performance really IS a bottleneck for these functions (AFAICS it's not), priority should be on correctness, not code size or performance.

> 3. Header column count is unlikely to exceed terminal column count (`COLS`), and the latter is bound to `signed int` type already.

See arguments from the article you linked.

That limit is implied by the current interface from ncurses, but a future version could just switch to size_t for its column and row counts (potentially breaking BC at that point); there are good reasons like huge screens where terminals COULD have that many columns/rows.

> 4. And no, we don't need `PRIu32` here. It's `unsigned int`, not `uint32_t`.

k, ACK on that. But that's only a minor note regarding the chosen type signature.

> This case is slightly different from string lengths. My view is string lengths in `size_t` can serve as a defence in depth, where the user is possible to feed a 4 GiB large string in a config file. But if the object size would be bound by any other factor, then `size_t` might be a waste.

Allocations are typically backed by 4 KiB pages at minimum, with the usual memory manager implementation pre-allocating some memory to handle sudden spikes. Thus unless the savings are major or we have massive amounts of objects (where storing UTF-8 instead of UTF-16 saves several hundreds of megabytes (like with some browsers) there's usually little need to save every last byte. Furthermore, by storing smaller data types you sometimes may make the alignment of the data worse, thus actually reducing performance.

From my experience I can tell, that I've seen real-life cases of where `uint16_t` vs. `uint32_t` was actually harmful due to alignment and memory access patterns. For 32 bit values on a 64 bit architecture, similar considerations apply.

> (By the way, I did read [this article from EWONTFIX](http://ewontfix.com/9/) when I write my changes.)

FWIW, I personally side with the author.

> I hope htop developers can reconsider.

@cgzones @fasterit @ginggs @natoscott : Your opinion on this matter of `size_t` vs `unsigned int`?

natoscott: I'm happy to stick with size_t here.

fasterit: Seconded
/DLange

Explorer09: @BenBE @natoscott @fasterit 
This commit was to fix the implicit downcast warnings from `HeaderLayout_getColumns()` by Clang 19 (the messages was: "`implicit conversion loses integer precision: 'size_t' (aka 'unsigned long') to 'unsigned int' [-Wshorten-64-to-32]`"). `HeaderLayout_getColumns()` was an inline function that returns a `size_t` that, in certain caller codes, has to downcast to `int` because terminal column variables are in `int` type to follow `curses` API.

With this background, I would like some clarifications on which decision to make:
(A) Keep `HeaderLayout_getColumns()` return value in `size_t`. Use `size_t` for local variables that hold the return values of `HeaderLayout_getColumns()`, and add explicit downcasts when needed (e.g. when used in the context of terminal column widths).
(B) Let `HeaderLayout_getColumns()` return an `unsigned int`. Keep using `int` type for places like terminal columns. But when using in array iterators (like in this example code you guys are commented on), upcast to `size_t`.
(C) Let `HeaderLayout_getColumns()` return an `unsigned int`. Also use `unsigned int` for array iterators that would never cross the bound of `HeaderLayout_getColumns()`. This is my original proposal.

BenBE: A.

Row.c
```diff
@@ -428,23 +428,23 @@ void Row_printNanoseconds(RichString* str, unsigned long long totalNanoseconds,
...
+      len = xSnprintf(buffer, sizeof(buffer), "%u.%0*lus ", (unsigned int)totalSeconds, width, (unsigned long)fraction);
       RichString_appendnAscii(str, baseColor, buffer, len);
       return;
    }
 
    if (totalSeconds < 600) {
-      unsigned int minutes = totalSeconds / 60;
-      unsigned int seconds = totalSeconds % 60;
-      unsigned int milliseconds = microseconds / 1000;
+      unsigned int minutes = (unsigned int)totalSeconds / 60;
```
cgzones, line 445: Shouldn't the cast here happen after the division?

Explorer09: Not necessary because of the `(totalSeconds < 600)` conditional. Although I argue this is more of a style issue...
(Compiler should catch that `(unsigned int)(totalSeconds / 60)` can be optimized to `((unsigned int)totalSeconds) / 60`, no?)

BenBE: Not necessarily, as `(unsigned int)((1UL<<32) / 60)` != `(unsigned int)(1UL<<32) / 60`.

Won't happen in this case based on the previous check, but the correctness should take priority here.

Meter.c
```diff
@@ -318,25 +325,25 @@ static void LEDMeterMode_draw(Meter* this, int x, int y, int w) {
       y + 2;
    attrset(CRT_colors[LED_COLOR]);
    const char* caption = Meter_getCaption(this);
-   mvaddstr(yText, x, caption);
-   int xx = x + strlen(caption);
+   mvaddnstr(yText, x, caption, w);
```
Explorer09, line 328: This one is a little beyond my knowledge. According to man pages, the `n` in `mvaddnstr` refers to number of "characters" to print. However, what I really want to limit is the number of terminal columns, not the number of Unicode characters (this distinction is needed if East Asian Wide characters are in the strings).

Perhaps a good way is to test how this function call would output, would it be "`１２`" or just "`１`":

```c
/* "１２３４ fullwidth" */
addnstr("\uFF11\uFF12\uFF13\uFF14 fullwidth", 2);
```

BenBE: What about [`wcswidth(3)`](https://man7.org/linux/man-pages/man3/wcswidth.3.html)?

Explorer09: What I wish is a function that does an inverse of [`wcswidth(3)`](https://man7.org/linux/man-pages/man3/wcswidth.3.html). `wcswidth()` returns the terminal column width given a string and a number of (Unicode) characters limit. What I wished is retrieving the number of characters (or number of bytes) given a string and a column width limit. `RichString_appendnWideColumns()` might give the closest I need, but I want a function that operates on a `const char*` rather than a `RichString` object. Maybe I would write one from scratch?

BenBE: Unfortunately that is easier said than actual done, but based on the fact that `wcswidth` returns >=0 you can probably build something like this by calling `wcwidth(3)` for each character in turn until you overshoot the limit, in which case you know the last character was too much (this accounts for zero-width-combiners and other special cases). The down-side is, that you will need to transform the `const char*` into `wchar_t*` first, thus having this functionality live with the other RichString functions might be reasonable.

NB: There's actually an issue that might benefit from this functionality in #462 IIRC.

Explorer09: @BenBE Yes, I'm aware of PR #462. And perhaps PR #1231 can benefit, too, as I wish this function would work as a word wrapper that is also CJK-aware.

Explorer09:
Closing this one in favor of #1669.

Issue #1517, opened 2024-08-14
Segmentation fault when toggling threads visibility while paused


SpaghettiBorgar:
Making process threads visible while process updates are paused causes a crash if tree view is enabled.
Other permutations of the steps seems to result in the same issue.

Version: `3.4.0-dev-3.3.0-179-g59578af`
`Linux arch 6.10.2-zen1-1-zen #1 ZEN SMP PREEMPT_DYNAMIC Sat, 27 Jul 2024 16:49:32 +0000 x86_64 GNU/Linux`

Steps to reproduce:
1. make sure process threads are hidden
2. Open tree view (F5)
3. Pause process updates (Shift+Z)
4. Show process threads (Shift+H)

```
Backtrace information:
----------------------
 0: 0x59ae5f4116a8  ./htop  (CRT_handleSIGSEGV+0xf6)  [0x59ae5f4116a8]
 1: 0x708ab81691d0  /usr/lib/libc.so.6  (+0)  [0x708ab81691cf]  {signal frame}
 2: 0x59ae5f418f86  ./htop  (Object_isA+0x12)  [0x59ae5f418f86]
 3: 0x59ae5f428aec  ./htop  (Vector_get+0x2a)  [0x59ae5f428aec]
 4: 0x59ae5f427838  ./htop  (Table_rebuildPanel+0xc2)  [0x59ae5f427838]
 5: 0x59ae5f421ee9  ./htop  (checkRecalculation+0x138)  [0x59ae5f421ee9]
 6: 0x59ae5f42265f  ./htop  (ScreenManager_run+0xba)  [0x59ae5f42265f]
 7: 0x59ae5f40f74e  ./htop  (CommandLine_run+0x2e3)  [0x59ae5f40f74e]
 8: 0x59ae5f40ad42  ./htop  (main+0x9)  [0x59ae5f40ad42]
 9: 0x708ab8151e08  /usr/lib/libc.so.6  (+0)  [0x708ab8151d90]
10: 0x708ab8151ecc  /usr/lib/libc.so.6  (__libc_start_main+0x8c)  [0x708ab8151ecc]
11: 0x59ae5f40ac65  ./htop  (_start+0x25)  [0x59ae5f40ac65]
```

natoscott:
I poked at it for awhile but haven't been able to get to the root cause so far.  Valgrind suggests it relates to a process/thread that has been deleted (exited) and is removed from the Table hash but still exists in the displayList somehow.
```
[...]

PR #1519, opened 2024-08-14
Force rebuild of display table after removals

cgzones:
Table_compact() is called after Table_cleanupRow(), which might remove
rows, which are still referenced by the display table.  Set needsSort to
true so that Table_updateDisplayList() always performs a rebuild and
does access removed rows.

Fixes: https://github.com/htop-dev/htop/issues/1517


BenBE (approved):
LGTM.

natoscott (approved):
LGTM too.

PR #1526, opened 2024-08-20
configure script improvements regarding pkg-config support

Explorer09:
* Move the macro call of PKG_PROG_PKG_CONFIG.
* Bring back the warning when configure is generated without pkg.m4
* CI FreeBSD job now installs pkg-config.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -233,7 +233,7 @@ jobs:
             release: '14.1'
             usesh: true
             prepare: |
-              pkg install -y gmake autoconf automake git
+              pkg install -y gmake autoconf automake pkgconf git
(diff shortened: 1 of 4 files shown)


Explorer09:
Minor note: Solaris 11.4 in our CI build jobs supports pkg-config, but the system doesn't seem to come with pkg.m4. So the warning would always exist if you generate htop's configure script there.

BenBE (approved):
LGTM.

Explorer09:
I added a change in the README to mention the use of `pkg-config`. Please review whether the wording is good enough.

BenBE (approved):

BenBE:
Merged as part of #1512 …

PR #1531, opened 2024-09-03
Minor Settings.c cleanup

natoscott:


Settings.c
```diff
@@ -73,20 +73,27 @@ static void writeQuotedList(FILE* fp, char** list) {
...
-      }
       free(this->screens);
    }
+}
+
+void Settings_delete(Settings* this) {
+   free(this->filename);
+   free(this->initialFilename);
+   Settings_deleteColumns(this);
+   Settings_deleteScreens(this);
```
BenBE, line 96:
```suggestion
   Settings_deleteColumns(this);
   Settings_deleteScreens(this);
   free(this->filename);
   free(this->initialFilename);
```

Remove sub-structures first. If (I know that's not the case here) the cleanup routines needed any access on the main structure, this would provide a more complete environment, thus reducing the dependencies for what may not be accessed.

Settings.c
```diff
@@ -73,20 +73,27 @@ static void writeQuotedList(FILE* fp, char** list) {
...
-   free(this->filename);
-   free(this->initialFilename);
-   for (unsigned int i = 0; i < HeaderLayout_getColumns(this->hLayout); i++) {
+static void Settings_deleteColumns(Settings* this) {
+   for (size_t i = 0; i < HeaderLayout_getColumns(this->hLayout); i++) {
       String_freeArray(this->hColumns[i].names);
       free(this->hColumns[i].modes);
    }
    free(this->hColumns);
+}
```
BenBE, line 82: This should set `this->hColumns = NULL;` explicitly, to avoid potential UAF issues.

Settings.c
```diff
@@ -73,20 +73,27 @@ static void writeQuotedList(FILE* fp, char** list) {
...
+
+static void Settings_deleteScreens(Settings* this) {
    if (this->screens) {
-      for (unsigned int i = 0; this->screens[i]; i++) {
+      for (size_t i = 0; this->screens[i]; i++)
          ScreenSettings_delete(this->screens[i]);
-      }
       free(this->screens);
```
BenBE, line 88: This should set `this->screens = NULL;` explicitly, to avoid potential UAF issues.

Settings.c
```diff
@@ -73,20 +73,27 @@ static void writeQuotedList(FILE* fp, char** list) {
...
-void Settings_delete(Settings* this) {
-   free(this->filename);
-   free(this->initialFilename);
-   for (unsigned int i = 0; i < HeaderLayout_getColumns(this->hLayout); i++) {
+static void Settings_deleteColumns(Settings* this) {
+   for (size_t i = 0; i < HeaderLayout_getColumns(this->hLayout); i++) {
       String_freeArray(this->hColumns[i].names);
       free(this->hColumns[i].modes);
    }
    free(this->hColumns);
```
BenBE, line 81: Should check for `this->hColumns` being non-NULL. For consistency with `Settings_deleteScreens`.

Settings.c
```diff
@@ -146,19 +153,16 @@ static bool Settings_validateMeters(Settings* this) {
    return anyMeter;
 }
 
-static void Settings_defaultMeters(Settings* this, unsigned int initialCpuCount) {
+static void Settings_defaultMeters(Settings* this, const Machine* host) {
+   unsigned int initialCpuCount = host->activeCPUs;
```
BenBE, line 157: Any reason for not using `size_t` here?
```suggestion
   size_t initialCpuCount = host->activeCPUs;
```

BenBE (commented):
LGTM. Some small suggestions for DiD and consistent code style.

BenBE:
Further small refactor:

```diff
diff --git a/Settings.c b/Settings.c
index 6ea15d02..555df42f 100644
--- a/Settings.c
+++ b/Settings.c
@@ -190,14 +190,9 @@ static const char* toFieldName(Hashtable* columns, int id, bool* enabled) {
    }
    if (id >= ROW_DYNAMIC_FIELDS) {
       const DynamicColumn* column = DynamicColumn_lookup(columns, id);
-      if (!column) {
-         if (enabled)
-            *enabled = false;
-         return NULL;
-      }
       if (enabled)
-         *enabled = column->enabled;
-      return column->name;
+         *enabled = column ? column->enabled : false;
+      return column ? column->name : NULL;
    }
    if (enabled)
       *enabled = true;
```

Settings.c
natoscott: That is not a pattern followed by the rest of the code base (in every _delete routine I just quickly audited at least), and I'm not particularly keen to start introducing it here TBH.  Better to use tools like valgrind IMO.

Settings.c
natoscott: Nak, as above.  That kinda change would warrant more general buy-in IMO, it affects the entire code base.

Settings.c
natoscott: Again, I'm not making functional changes here.  IIRC there is a legit case where NULL screens is used so I'm keeping the code as-is and not changing any logic like this.

Settings.c
natoscott: Sure thing, will do.

Settings.c
natoscott: Since this was a simple cleanup, I preferred not to change the logic or ordering at all.

natoscott:
OK, I've reviewed those changes - I merged the size_t changes into one commit, dropped those things that I felt were closer to "code churn" than "code cleanup", and added a commit for that final suggestion.

PR #1533, opened 2024-09-06
Normalize DiskIOMeter bar by number of disks

### LINKED ISSUE #1374: Allow Disk IO meter to be above 100%
The latest htop version has capped the Disk IO meter at 100%, which is quite annoying as there is currently no per-disk utilization meter (#952) and doesn't seem like there will be one for a while. Allowing the Disk IO meter to be above 100% is useful if you have a disk that is always running at 100%, allowing you to see if other disks are running. This functionality was removed with 89b90c4

Explorer09:
The maximum utilization should be "100% * number of disks". Set the bar and graph of the meter to draw with that maximum. Also remove the 100% cap on the utilization percentage display.

Resolves #1374.

### DIFF
--- a/DiskIOMeter.c
+++ b/DiskIOMeter.c
@@ -31,6 +31,7 @@ static MeterRateStatus status = RATESTATUS_INIT;
 static char cached_read_diff_str[6];
 static char cached_write_diff_str[6];
 static double cached_utilisation_diff;
+static double cached_utilisation_norm;
 
 static void DiskIOMeter_updateValues(Meter* this) {
    const Machine* host = this->host;
@@ -82,12 +83,15 @@ static void DiskIOMeter_updateValues(Meter* this) {
          }
          Meter_humanUnit(cached_write_diff_str, diff, sizeof(cached_write_diff_str));
 
+         cached_utilisation_diff = 0.0;
+         cached_utilisation_norm = 0.0;
          if (data.totalMsTimeSpend > cached_msTimeSpend_total) {
             diff = data.totalMsTimeSpend - cached_msTimeSpend_total;
             cached_utilisation_diff = 100.0 * (double)diff / passedTimeInMs;
-            cached_utilisation_diff = MINIMUM(cached_utilisation_diff, 100.0);
-         } else {
-            cached_utilisation_diff = 0.0;
+            if (data.numDisks > 0) {
+               cached_utilisation_norm = (double)diff / (passedTimeInMs * data.numDisks);
+               cached_utilisation_norm = MINIMUM(cached_utilisation_norm, 1.0);
+            }
          }
       }
 
@@ -96,7 +100,7 @@ static void DiskIOMeter_updateValues(Meter* this) {
       cached_msTimeSpend_total = data.totalMsTimeSpend;
    }
 
-   this->values[0] = cached_utilisation_diff;
+   this->values[0] = cached_utilisation_norm;
 
    if (status == RATESTATUS_NODATA) {
       xSnprintf(this->txtBuffer, sizeof(this->txtBuffer), "no data");
@@ -154,7 +158,7 @@ const MeterClass DiskIOMeter_class = {
    .defaultMode = TEXT_METERMODE,
    .supportedModes = METERMODE_DEFAULT_SUPPORTED,
    .maxItems = 1,
-   .total = 100.0,
+   .total = 1.0,
    .attributes = DiskIOMeter_attributes,
    .name = "DiskIO",
    .uiName = "Disk IO",
--- a/DiskIOMeter.h
+++ b/DiskIOMeter.h
@@ -16,6 +16,7 @@ typedef struct DiskIOData_ {
    uint64_t totalBytesRead;
    uint64_t totalBytesWritten;
    uint64_t totalMsTimeSpend;
+   uint64_t numDisks;
 } DiskIOData;
 
 extern const MeterClass DiskIOMeter_class;
--- a/darwin/Platform.c
+++ b/darwin/Platform.c
@@ -404,7 +404,7 @@ bool Platform_getDiskIO(DiskIOData* data) {
    if (IOServiceGetMatchingServices(iokit_port, IOServiceMatching("IOBlockStorageDriver"), &drive_list))
       return false;
 
-   unsigned long long int read_sum = 0, write_sum = 0, timeSpend_sum = 0;
+   uint64_t read_sum = 0, write_sum = 0, timeSpend_sum = 0;
    uint64_t numDisks = 0;
 
    io_registry_entry_t drive;
@@ -437,7 +437,7 @@ bool Platform_getDiskIO(DiskIOData* data) {
       numDisks++;
 
       CFNumberRef number;
-      unsigned long long int value;
+      uint64_t value;
 
       /* Get bytes read */
       number = (CFNumberRef) CFDictionaryGetValue(statistics, CFSTR(kIOBlockStorageDriverStatisticsBytesReadKey));
--- a/darwin/Platform.c
+++ b/darwin/Platform.c
@@ -405,6 +405,7 @@ bool Platform_getDiskIO(DiskIOData* data) {
       return false;
 
    unsigned long long int read_sum = 0, write_sum = 0, timeSpend_sum = 0;
+   uint64_t numDisks = 0;
 
    io_registry_entry_t drive;
    while ((drive = IOIteratorNext(drive_list)) != 0) {
@@ -432,6 +433,8 @@ bool Platform_getDiskIO(DiskIOData* data) {
          IOObjectRelease(drive);
          continue;
       }
+
+      numDisks++;
 
       CFNumberRef number;
       unsigned long long int value;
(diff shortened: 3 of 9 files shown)


natoscott:
Something like this patch will provide the PCP value you seek.  You might want to consider a similar change for the NetworkIOMeter, in which case there is a "hinv.ninterface" metric that can be used similarly.

```
pcp/Metric.h
--- /tmp/git-blob-5igpu5/Metric.h	2024-09-06 15:32:13.776523149 +1000
+++ pcp/Metric.h	2024-09-06 15:26:46.052772959 +1000
@@ -26,6 +26,7 @@ typedef enum Metric_ {
    PCP_CONTROL_THREADS,         /* proc.control.perclient.threads */
 
    PCP_HINV_NCPU,               /* hinv.ncpu */
+   PCP_HINV_NDISK,              /* hinv.ndisk */
    PCP_HINV_CPUCLOCK,           /* hinv.cpu.clock */
    PCP_UNAME_SYSNAME,           /* kernel.uname.sysname */
    PCP_UNAME_RELEASE,           /* kernel.uname.release */
pcp/Platform.c
--- /tmp/git-blob-JCaR3p/Platform.c	2024-09-06 15:32:13.956523525 +1000
+++ pcp/Platform.c	2024-09-06 15:30:56.940361942 +1000
@@ -127,6 +127,7 @@ static const char* Platform_metricNames[
    [PCP_CONTROL_THREADS] = "proc.control.perclient.threads",
 
    [PCP_HINV_NCPU] = "hinv.ncpu",
+   [PCP_HINV_NDISK] = "hinv.ndisk",
    [PCP_HINV_CPUCLOCK] = "hinv.cpu.clock",
    [PCP_UNAME_SYSNAME] = "kernel.uname.sysname",
    [PCP_UNAME_RELEASE] = "kernel.uname.release",
@@ -385,6 +386,7 @@ bool Platform_init(void) {
    Metric_enable(PCP_PID_MAX, true);
    Metric_enable(PCP_BOOTTIME, true);
    Metric_enable(PCP_HINV_NCPU, true);
+   Metric_enable(PCP_HINV_NDISK, true);
    Metric_enable(PCP_PERCPU_SYSTEM, true);
    Metric_enable(PCP_UNAME_SYSNAME, true);
    Metric_enable(PCP_UNAME_RELEASE, true);
@@ -753,6 +755,8 @@ bool Platform_getDiskIO(DiskIOData* data
       data->totalBytesWritten = value.ull;
    if (Metric_values(PCP_DISK_ACTIVE, &value, 1, PM_TYPE_U64) != NULL)
       data->totalMsTimeSpend = value.ull;
+   if (Metric_values(PCP_HINV_NDISK, &value, 1, PM_TYPE_U64) != NULL)
+      data->numDisks = value.ull;
    return true;
 }
 
```

Explorer09:
@natoscott There is no "utilization percentage" for NetworkIOMeter at the moment. So I would consider showing the number of interfaces a feature separate from this pull request.

natoscott:
Understood & agreed.  FWIW, I think this approach is not ideal (shares the same problem of the original 100% limiting, as mentioned by the reporter of #1374 - the code was fine the way it was originally).  We need to be aware here of properties of some hardware described in the iostat(1) man page ...

```
              %util  Percentage  of elapsed time during which I/O requests were
                     issued to the device (bandwidth utilization  for  the  de‐
                     vice).  Device  saturation occurs when this value is close
                     to 100% for devices serving requests  serially.   But  for
                     devices  serving requests in parallel, such as RAID arrays
                     and modern SSDs, this number does not reflect  their  per‐
                     formance limits.
```

Explorer09:
@natoscott What was the ideal approach anyway?

natoscott:
Just reverting to the original htop behaviour as the issue reporter suggested.

Explorer09:
@natoscott Look in my patch carefully. I separated the utilization value into two. `cached_utilisation_diff` would revert to original behavior of allowing more than 100% to be shown; the bar would be drawn with `cached_utilisation_norm` which is the normalized value.

DiskIOMeter.h
```diff
@@ -16,6 +16,7 @@ typedef struct DiskIOData_ {
    uint64_t totalBytesRead;
    uint64_t totalBytesWritten;
    uint64_t totalMsTimeSpend;
+   uint64_t numDisks;
```
Explorer09, line 19: Should the data type be `unsigned int`, `size_t`, `uint64_t` or `unsigned long long`?
Should the property be named "numDisks" or "nDisks"?

Other than those questions, this patch is ready.

natoscott:
> @natoscott Look in my patch carefully. I separated the utilization value into two. `cached_utilisation_diff` would revert to original behavior of allowing more than 100% to be shown; the bar would be drawn with `cached_utilisation_norm` which is the normalized value.

Two thumbs up - good stuff, thanks.  Regarding your data type question, I think I know what @BenBE is going to suggest: size_t FTW.  :)  However, systems with more than 4 billion disks are few and far between so I'm sure a 32 bit unsigned value would suffice.  If noone has strong opinions, I suggest sticking with what you have now since that's consistent with the surrounding code.

natoscott (approved):
LGTM.

DiskIOMeter.c
```diff
@@ -85,9 +86,11 @@ static void DiskIOMeter_updateValues(Meter* this) {
          if (data.totalMsTimeSpend > cached_msTimeSpend_total) {
             diff = data.totalMsTimeSpend - cached_msTimeSpend_total;
             cached_utilisation_diff = 100.0 * (double)diff / passedTimeInMs;
-            cached_utilisation_diff = MINIMUM(cached_utilisation_diff, 100.0);
+            cached_utilisation_norm = (double)diff / (passedTimeInMs * data.numDisks);
```
BenBE, line 89: Risk of DivByZero?

Explorer09:
> Regarding your data type question, I think I know what @BenBE is going to suggest: size_t FTW. :)

For PCP this would require a downcast from `uint64_t` and an `(numDisks <= SIZE_MAX)` check that I don't think it's worth it. That's why I chose a larger data type than `size_t` for now.

DiskIOMeter.c
Explorer09: Good point. Fixed.

natoscott:
> > Regarding your data type question, I think I know what @BenBE is going to suggest: size_t FTW. :)
> 
> For PCP this would require a downcast from `uint64_t` and...

FWIW, for PCP to use 32-bit types we would change this section:
```
   if (Metric_values(PCP_HINV_NDISK, &value, 1, PM_TYPE_U64) != NULL)
      data->numDisks = value.ull;
```
to
```
   if (Metric_values(PCP_HINV_NDISK, &value, 1, PM_TYPE_U32) != NULL)
      data->numDisks = value.ul;
```

but, keep it as you have it, it's fine IMO.

ShaiMagal:
When this PR will be merged? It's "game changer" for overall about disk activity :)

fasterit:
@ShaiMagal: Have you tested the PR?

ShaiMagal:
@natoscott When this PR will be merged? It's "game changer" for overall about disk activity :)

DiskIOMeter.h
BenBE: I think, both `uint64_t` and `size_t` would be fine. No need to change this.

darwin/Platform.c
```diff
@@ -405,6 +405,7 @@ bool Platform_getDiskIO(DiskIOData* data) {
       return false;
 
    unsigned long long int read_sum = 0, write_sum = 0, timeSpend_sum = 0;
+   unsigned long long int numDisks = 0;
```
BenBE, line 408: Possible cleanup: Be consistent with later usage of this variable and use `uint64_t` here too …

freebsd/Platform.c
```diff
@@ -313,6 +313,7 @@ bool Platform_getDiskIO(DiskIOData* data) {
    int count = current.dinfo->numdevs;
 
    unsigned long long int bytesReadSum = 0, bytesWriteSum = 0, timeSpendSum = 0;
+   unsigned long long int numDisks = 0;
```
BenBE, line 317: Possible cleanup: Be consistent with later usage of this variable and use `uint64_t` here too …

linux/Platform.c
```diff
@@ -614,6 +614,7 @@ bool Platform_getDiskIO(DiskIOData* data) {
    char lastTopDisk[32] = { '\0' };
 
    unsigned long long int read_sum = 0, write_sum = 0, timeSpend_sum = 0;
+   unsigned long long int numDisks = 0;
```
BenBE, line 617: Possible cleanup: Be consistent with later usage of this variable and use `uint64_t` here too …

Also, for consistency with other platforms, a blank line after these two lines would be good.

BenBE (approved):
Aside from some minor cleanup stuff the PR LGTM.

@ShaiMagal Please confirm whether you have tested these changes.

@Explorer09 Mind to include these minor cleanups? Just throw in an extra patch with the consistency cleanup stuff and I think we are fine.

darwin/Platform.c
BenBE: I meant both lines (just as marked ;-)), as `*_sum` maps to `total*`, which all are `uint64_t` …

Explorer09: That one is better done in a separate commit.

Issue #1541, opened 2024-09-28
Migrate away from deprecated ncurses functions `attrset`, `attron` and `attroff`


C0rn3j:
https://invisible-island.net/ncurses/man/curs_attr.3x.html

Perhaps even just use wide characters by default with `wattr_on`/`wattr_off`/`wattr_set`?  
In which case other things also have to be migrated, see https://invisible-island.net/ncurses/man/curs_attr.3x.html#h2-PORTABILITY

Ran into this while trying to fix https://github.com/htop-dev/htop/issues/243

attron -> attr_on
```yaml
% grep -R attron 2>/dev/null
CRT.c:   attron(resetColor);
```

attroff -> attr_off
```yaml
% grep -R attroff 2>/dev/null
CRT.c:   attroff(resetColor);
```

attrset -> attr_set
```yaml
% grep -R attrset 2>/dev/null
Action.c:   attrset(attr);
Action.c:   attrset(CRT_colors[HELP_BOLD]);
Action.c:   attrset(CRT_colors[DEFAULT_COLOR]);
Action.c:   attrset(CRT_colors[DEFAULT_COLOR]);
Action.c:   attrset(CRT_colors[DEFAULT_COLOR]);
Action.c:   attrset(CRT_colors[DEFAULT_COLOR]);
Action.c:   attrset(CRT_colors[DEFAULT_COLOR]);
Action.c:      attrset((helpLeft[item].roInactive && readonly) ? CRT_colors[HELP_SHADOW] : CRT_colors[DEFAULT_COLOR]);
Action.c:      attrset((helpLeft[item].roInactive && readonly) ? CRT_colors[HELP_SHADOW] : CRT_colors[HELP_BOLD]);
Action.c:         attrset((helpLeft[item].roInactive && readonly) ? CRT_colors[HELP_SHADOW] : CRT_colors[PROCESS_THREAD]);
Action.c:         attrset((helpLeft[item].roInactive && readonly) ? CRT_colors[HELP_SHADOW] : CRT_colors[PROCESS_THREAD]);
Action.c:      attrset((helpRight[item].roInactive && readonly) ? CRT_colors[HELP_SHADOW] : CRT_colors[HELP_BOLD]);
Action.c:      attrset((helpRight[item].roInactive && readonly) ? CRT_colors[HELP_SHADOW] : CRT_colors[DEFAULT_COLOR]);
Action.c:   attrset(CRT_colors[HELP_BOLD]);
[...]

PR #1542, opened 2024-09-29
Stopgap solution to preserve colors and attributes on highlighting

C0rn3j:
[Screencast_20240929_172318.webm](https://github.com/user-attachments/assets/0418af32-457f-4a61-bdbc-c8e890682166)

This is a decent workaround for https://github.com/htop-dev/htop/issues/243 solely by using text attributes taking https://github.com/htop-dev/htop/pull/434 and adapting it.  
This is a stopgap solution until the issue is properly resolved by actually being able to define the desired text colors while keeping the highlight background, which I assume will take a while.

There is a couple problems:
  * Function is from original PR needs renaming `RichString_setAttrn_preserveBold` -> `RichString_setAttrn_preserve` ?
  * There are debug statements left in the current code, will be removed upon finalization
  * I am not a C/C++ dev, someone please verify my actual functional code makes sense
    * Currently I check if text color is not default (ColorPair 0) or if text is Bolded, is there even a situation where text is bolded but not colored?
  * The entire column gets lit up - is this desired? I presume not. 
    * Should I solve this in my code by ensuring the char is not a space, or should this be fixed in other code that ensures it's not setting params and text colors for spaces?  
image
  * Biggest thorn is the fact that we can be setting highlight color that is identical to text color, which means we are not highlighting anything. This is detected and text attribute is set to italics to at least differentiate it, but it is not an amazing solution, we would have to be able to italicize the entire column for it to not look like ass:  
image

Mildly related to #1541

### DIFF
--- a/Panel.c
+++ b/Panel.c
@@ -289,7 +289,7 @@ void Panel_draw(Panel* this, bool force_redraw, bool focus, bool highlightSelect
          }
          if (item.highlightAttr) {
             attrset(item.highlightAttr);
-            RichString_setAttr(&item, item.highlightAttr);
+            RichString_setAttr_preserveBold(&item, item.highlightAttr);
             this->selectedLen = itemLen;
          }
          mvhline(y + line, x, ' ', this->w);
--- a/RichString.c
+++ b/RichString.c
@@ -169,6 +169,28 @@ void RichString_appendChr(RichString* this, int attrs, char c, int count) {
    }
 }
 
+inline void RichString_setAttrn_preserveBold(RichString* this, int attrs, int start, int finish) {
+   cchar_t* ch = this->chptr + start;
+   finish = CLAMP(finish, 0, this->chlen - 1);
+   for (int i = start; i <= finish; i++) {
+      ch->attr = (ch->attr & A_BOLD) ? (attrs | A_BOLD) : attrs;
+      ch++;
+   }
+}
+
+void RichString_setAttrn_preserveBold(RichString* this, int attrs, int start, int finish) {
+   chtype* ch = this->chptr + start;
+   finish = CLAMP(finish, 0, this->chlen - 1);
+   for (int i = start; i <= finish; i++) {
+      *ch = (*ch & 0xff) | attrs | (*ch & A_BOLD);
+      ch++;
+   }
+}
+
+void RichString_setAttr_preserveBold(RichString* this, int attrs) {
+   RichString_setAttrn_preserveBold(this, attrs, 0, this->chlen - 1);
+}
+
 int RichString_findChar(const RichString* this, char c, int start) {
    const wchar_t wc = btowc(c);
    const cchar_t* ch = this->chptr + start;
--- a/RichString.c
+++ b/RichString.c
@@ -169,20 +169,11 @@ void RichString_appendChr(RichString* this, int attrs, char c, int count) {
    }
 }
 
-inline void RichString_setAttrn_preserveBold(RichString* this, int attrs, int start, int finish) {
+void RichString_setAttrn_preserveBold(RichString* this, int attrs, int start, int finish) {
    cchar_t* ch = this->chptr + start;
    finish = CLAMP(finish, 0, this->chlen - 1);
    for (int i = start; i <= finish; i++) {
-      ch->attr = (ch->attr & A_BOLD) ? (attrs | A_BOLD) : attrs;
-      ch++;
-   }
-}
-
-void RichString_setAttrn_preserveBold(RichString* this, int attrs, int start, int finish) {
-   chtype* ch = this->chptr + start;
-   finish = CLAMP(finish, 0, this->chlen - 1);
-   for (int i = start; i <= finish; i++) {
-      *ch = (*ch & 0xff) | attrs | (*ch & A_BOLD);
+      ch->attr = (ch->attr & A_BOLD) ? ((unsigned int)attrs | A_BOLD) : (unsigned int)attrs;
       ch++;
    }
 }
--- a/RichString.c
+++ b/RichString.c
@@ -8,6 +8,8 @@ in the source distribution for its full text.
 #include "config.h" // IWYU pragma: keep
 
 #include "RichString.h"
+#include "ProvideCurses.h"
+#include "CRT.h"
 
 #include <assert.h>
 #include <ctype.h>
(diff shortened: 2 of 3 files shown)


BenBE:
I converted this PR to a draft right now, as there are several issues I see with this current implementation.

1. Visually this does not look right on several levels and I think we should discuss some of the workings for the themes to find a good workaround for those. E.g. for the highlighting to the right (red+yellow for basenames) the coloring inversed just looks ugly and much too offending to the eye.

2. Please keep the git history short and linear with corrections squashed/fixup'd into the erroneous commit.

3. Several build failures on the CI. If you need help with addressing these, feel free to ask.

Regarding highlighting there are basically 5 options right now:
- Normal
- Selected
- Followed
- New Process
- Terminated Process

On your question with A_BOLD without color: AFAIR there's no place that uses this and if it is used, the most likely place to look would be with the (text-styled) meters.

C0rn3j:
> On your question with A_BOLD without color: AFAIR there's no place that uses this and if it is used, the most likely place to look would be with the (text-styled) meters.

Thanks, threw that check out.

> Please keep the git history short and linear with corrections squashed/fixup'd into the erroneous commit.

Edits are enabled, I assumed that whoever pulls it squashes it into a single commit.

> Several build failures on the CI. If you need help with addressing these, feel free to ask.

Fixed all but one, help welcome..

It was pain to actually find the issues in the walls of text.  
If the tooling can be forced to use colors in the CI (CI itself can handle colors), it would be nice if it did.  
Looks like [it should be possible](https://stackoverflow.com/a/73349744/8962143)?

> Visually this does not look right

I know, though I do not intend it to look perfect, just good enough for normal use until proper colors are implemented.

If this is not acceptable, suggestion that looks better but retains less information would be to

A) A_STANDOUT + retain A_BOLD for all colored text:
image

B) A_STANDOUT + retain A_BOLD for all A_BOLD text:
image

C) Just don't touch BOLD text:
image

D) Just don't touch color text:
image


My main goal here is to retain the yellow/red highlighting as I rely on it often and when I filter out instead of search for the process I wish to know is misbehaving because of outdated libs/binary, I get no info about it, this helps:

image

BenBE:
> > On your question with A_BOLD without color: AFAIR there's no place that uses this and if it is used, the most likely place to look would be with the (text-styled) meters.
> 
> Thanks, threw that check out.

NP.

> > Please keep the git history short and linear with corrections squashed/fixup'd into the erroneous commit.
> 
> Edits are enabled, I assumed that whoever pulls it squashes it into a single commit.

Usually we aim to touch the PRs as little as possible. The PR itself should already tell the changes; the story how you got there is of little interest. Have a look at the [style guide](https://github.com/htop-dev/htop/blob/main/docs/styleguide.md) for some details.

Also, with your code, please mind the indentation: We don't do hanging indentations, but instead do either not break (reasonably long lines allowed) or indent exactly one more level. In the two implementations for `RichString_setAttrn_preserveWithStandout` a long line on the `ch = … ?:` line would be acceptable.

> > Several build failures on the CI. If you need help with addressing these, feel free to ask.
> 
> Fixed all but one, help welcome.

Will take a look …

> It was pain to actually find the issues in the walls of text. If the tooling can be forced to use colors in the CI (CI itself can handle colors), it would be nice if it did. Looks like [it should be possible](https://stackoverflow.com/a/73349744/8962143)?

Will take a look …

> > Visually this does not look right
> 
> I know, though I do not intend it to look perfect, just good enough for normal use until proper colors are implemented.
> 
> If the suggested variant is not acceptable, some others would be:
> 
> A) A_STANDOUT + retain A_BOLD for all colored text: image
> 
> B) A_STANDOUT + retain A_BOLD for all A_BOLD text: image
> 
> C) Just don't touch BOLD text: image
> 
> D) Just don't touch color text: image
> 
> My main goal here is to retain the yellow/red highlighting as I rely on it often and when I filter out instead of search for the process I wish to know is misbehaving because of outdated libs/binary, I get no info about it, this helps:
> 
> image

The black background for places where colors were preserved visually makes this kinda noisy. So, for most of the columns (e.g. sizes) mapping the color to e.g. bold would usually already suffice and would be much less distracting. Yes, that's not very helpful with your usecase for the outdated binary highlighting while following, but this at least would be a good first step.

For the second step, we may extend the color stuff in a way to allow for the (internal) attributes mapped to the theme colors contextually (e.g. specify bold black with red background for a removed process with outdated executable, but bright red with orange background if you followed the same process. But this will likely need some updates to the theme data and thus should be postponed for a bit.

BenBE:
The remaining error on OpenBSD is this one:

```
cc -DHAVE_CONFIG_H -I.  -DNDEBUG  -std=c99 -pedantic  -Wall -Wcast-align -Wcast-qual -Wextra -Wfloat-equal -Wformat=2 -Winit-self -Wmissing-format-attribute -Wmissing-noreturn -Wmissing-prototypes -Wpointer-arith -Wshadow -Wstrict-prototypes -Wundef -Wunused -Wwrite-strings -Wextra-semi-stmt -Wimplicit-int-conversion -Wnull-dereference -Werror -D_XOPEN_SOURCE_EXTENDED -DSYSCONFDIR="\"/usr/local/etc\"" -I"./openbsd" -g -O2 -MT RichString.o -MD -MP -MF $depbase.Tpo -c -o RichString.o RichString.c &&\
RichString.c:190:24: error: use of undeclared identifier 'A_ITALIC'
         attrToPass |= A_ITALIC;
                       ^
1 error generated.
```

Issue seems to be, that `A_ITALIC` is not supported in the ncurses version of OpenBSD or needs to be treated differently as `A_ITALIC` is an extension.

C0rn3j:
> So, for most of the columns (e.g. sizes) mapping the color to e.g. bold would usually already suffice and would be much less distracting

Bolding things will have terrible contrast in some cases, which was done and was refused in the last PR.

> Issue seems to be, that A_ITALIC is not supported in the ncurses version of OpenBSD or needs to be treated differently as A_ITALIC is an extension.

I'll ifdef it somehow if it will even be kept in the end, and just not have it on OpenBSD I suppose.

---

Here is my least intrusive idea so far - just A_STANDOUT everything that's not colored:


[Screencast_20240929_234724.webm](https://github.com/user-attachments/assets/0a9672b8-b926-423b-8f03-9f667afec0f2)

image

None of this really looks *great*.

Issue #647, opened 2021-06-06
Custom character for meter (instead of "|")


BenBE:
It would be nice if the character that makes the meter could be changed from "|" to the one the user prefers. Is there ANY way to change that ugly default pipe for a better visual character with more body defined by us? Thanks in advance.

_Originally posted by @noctumsempra in https://github.com/htop-dev/htop/issues/526#issuecomment-855311230_

Noctumsempra:
Hello again. Is there any new about this "new feature"?

Can these bars' char be some character being READ from ~/.config/htoprc via one of its config pairs for example?
I wish I could do this myself, it doesn't look difficult to implement, and it would visually help any of us with burden in the eyes to be able to chose a solid █ (U+2588) instead of the THIN UGLY PIPE.

Thanks.

BenBE:
There has no work been done on this, but in general I see this as a nice enhancement.

The new part here is mostly with the additional setting that has to be devised and the UI for its setup. We aim for all settings to be reachable in the UI directly. Thus just having the setting available in the config file won't be sufficient (that would be just part of the necessary changes).

eworm-de:
This could even go with fractional blocks to increase precision... Not sure if it would be worth the trouble.

BenBE:
> This could even go with fractional blocks to increase precision... Not sure if it would be worth the trouble.

I though of something like this too and if we'd like to go that route we'd design the configuration UI accordingly.

In a really trimmed-down version this might be as simple as providing a set of defaults with something like the following sets as defaults:

```
 ||||||||
 ########
 ▏▎▍▌▋▊▉█
 ▁▂▃▄▅▆▇█
 ░░▒▒▓▓██
 ▌▌▌▌████
⠀⡀⡄⡆⡇⣇⣧⣷⣿
```

This basically would produce subpixel rendering for ⅛th of the cell (full cells using index 8, empty cells using cell 0 from the 9 bytes in each profile).

Noctumsempra:
Love the different pattern ideas for the rendering.

0cjs:
This would be really nice to have because what's particularly hard to read about the current meters using just the `|` character is the colours. In small font sizes on LCD monitors using anti-aliasing it can be nearly impossible to distinguish a green `|` from a yellow `|`, or even a red `|`.

rustedusted:
If this feature is not added yet

 i am ready to take this feature request

derVedro:
@rustedusted Any chance you keep on it?

[...]

PR #1545, opened 2024-10-03
adding bar_label parameter in htoprc adressing (feature request) #647

rustedusted:
just add
bar_label=#

to your htoprc and we see the change from | to #


rustedusted:
This pr addresses issue #647

rustedusted:
The PR only has support for normal ascii characters

i can extend this feature to add UTF-8 characters but the CONTRIBUTING.md suggested that newer features shouldn't unnecessarily increase the footprint of the program
(this is my first commit btw so i don't know much)

also Please look if the change in the file Settings.c is correct 

options[1][0] doesn't seem a good way to get the character on right btw

PR #1546, opened 2024-10-03
adding bar_label parameter to htoprc

rustedusted:
adding bar_label feature to change the pipe in htop to any other ascii character user wants

this is addressing feature request #647 

The PR only has support for normal ascii characters

i can extend this feature to add UTF-8 characters but the CONTRIBUTING.md suggested that newer features shouldn't unnecessarily increase the footprint of the program
(this is my first commit btw so i don't know much)

also Please look if the change in the file Settings.c is correct
options[1][0] doesn't seem a good way to get the character on right of = sign

### DIFF
--- a/DisplayOptionsPanel.c
+++ b/DisplayOptionsPanel.c
@@ -156,6 +156,7 @@ DisplayOptionsPanel* DisplayOptionsPanel_new(Settings* settings, ScreenManager* 
    Panel_add(super, (Object*) CheckItem_newByRef("Highlight new and old processes", &(settings->highlightChanges)));
    Panel_add(super, (Object*) NumberItem_newByRef("- Highlight time (in seconds)", &(settings->highlightDelaySecs), 0, 1, 24 * 60 * 60));
    Panel_add(super, (Object*) NumberItem_newByRef("Hide main function bar (0 - off, 1 - on ESC until next input, 2 - permanently)", &(settings->hideFunctionBar), 0, 0, 2));
+   Panel_add(super, (Object*) NumberItem_newByRef("Bar Type (0-7)", &(settings->barType), 0, 0, 7));
    #ifdef HAVE_LIBHWLOC
    Panel_add(super, (Object*) CheckItem_newByRef("Show topology when selecting affinity by default", &(settings->topologyAffinity)));
    #endif
--- a/Meter.c
+++ b/Meter.c
@@ -13,6 +13,7 @@ in the source distribution for its full text.
 #include <math.h>
 #include <stdlib.h>
 #include <string.h>
+#include <wchar.h>
 
 #include "CRT.h"
 #include "Macros.h"
@@ -70,6 +71,17 @@ static void TextMeterMode_draw(Meter* this, int x, int y, int w) {
 
 static const char BarMeterMode_characters[] = "|#*@$%&.";
 
+static const wchar_t* bars[8] = {
+   L" ||||||||",
+   L" ########",
+   L"⠀⡀⡄⡆⡇⣇⣧⣷⣿",
+   L" ░░▒▒▓▓██",
+   L" ▏▎▍▌▋▊▉█",
+   L" ▁▂▃▄▅▆▇█",
+   L" ▌▌▌▌████",
+   L" ▔🮂🮃▀🮄🮅🮆█"
+};
+
 static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
    // Draw the caption
    const char* caption = Meter_getCaption(this);
@@ -124,26 +136,44 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
 
    // First draw in the bar[] buffer...
    int offset = 0;
+
+   const Settings* settings = this->host->settings;
+   const wchar_t* barChars = &bars[settings->barType][1];
+   const size_t barLen = wcslen(barChars);
+   const size_t wsub = w * barLen;
+
    for (uint8_t i = 0; i < this->curItems; i++) {
       double value = this->values[i];
-      if (isPositive(value) && this->total > 0.0) {
+      int actualWidth = 0;
+
+      if (isPositive(value) && this->total > 0.0 && (value * wsub > 0.5 * this->total)) {
          value = MINIMUM(value, this->total);
+         actualWidth = ceil((value / this->total) * wsub);
          blockSizes[i] = ceil((value / this->total) * w);
       } else {
          blockSizes[i] = 0;
+         continue;
       }
+
       int nextOffset = offset + blockSizes[i];
       // (Control against invalid values)
       nextOffset = CLAMP(nextOffset, 0, w);
-      for (int j = offset; j < nextOffset; j++)
+
+      for (int j = offset; j < nextOffset; j++) {
          if (RichString_getCharVal(bar, startPos + j) == ' ') {
             if (CRT_colorScheme == COLORSCHEME_MONOCHROME) {
                assert(i < strlen(BarMeterMode_characters));
                RichString_setChar(&bar, startPos + j, BarMeterMode_characters[i]);
+            } else if (settings->barType) {
+               RichString_setChar(&bar, startPos + j, bars[settings->barType][barLen]);
             } else {
                RichString_setChar(&bar, startPos + j, '|');
             }
          }
+      }
+
+      RichString_setChar(&bar, startPos + nextOffset - 1, barChars[actualWidth % barLen]);
+
       offset = nextOffset;
    }
 
--- a/Settings.h
+++ b/Settings.h
@@ -111,6 +111,8 @@ typedef struct Settings_ {
 
    bool changed;
    uint64_t lastUpdate;
+
+   int barType;
 } Settings;
 
 #define Settings_cpuId(settings, cpu) ((settings)->countCPUsFromOne ? (cpu)+1 : (cpu))


rustedusted:
i will right away commit all the things that might be necessary

BenBE:
While the PR itself looks fine from a code PoV adding new settings is always a bit tricky as future backwards compatibility has to be taken into account. In the mentioned issue, where this new feature is somewhat discussed, the aim was more for something extensible, that may be added upon later if necessary.

As the PR looks now, this is kinda fixed just on this one usecase of just a single character and somewhat neglects the extensibility for e.g. "sub-pixel" highlighting also discussed in the linked issue.

Another aspect is that all settings should be available in the UI. This is currently not the case with this PR.

@fasterit @natoscott : Further thoughts?

rustedusted:
Oh thanks for the input, :D
I can add sub pixel rendering 

maybe an integer called bar_type={1-6}  where each number represents the type of bar to use 

is this fine or do you guys have any other ideas?


also you said the settings should be available in UI 
should I add a shortcut key for it?? or when the bar is clicked the pipe cycles through different character-set?

BenBE:
I think for the UI settings it should suffice to have it selectable in the setup screen using <kbd>F2</kbd> similar to how the `Hide main function bar` setting is handled. Have a look at how number inputs for the settings are done; you can do a new one for the style. This setting can, as you suggested, be an integer in the config file.

Possible future extension: If we really wanted to include a custom option (fully hypothetical), that one could add 18 code points (9 for each bar level) after the initial "custom" designator. But please leave this out for now.

rustedusted:
oh will get into that right away

Explorer09:
@rustedusted 
There are two main things to address regarding the whole issue of Bar meter characters:
1. The "subpixel" rendering. (Technically it's character cell not pixel, but anyway.) It would require an algorithm rewrite, and while I had some ideas on the algorithm (because I wrote the wonderful #714 feature that's still waiting to be merged), I am still unsure about how useful this "subpixel" rendering of bars can provide comparing to the code complexity that will add. (Also, once implemented, the set of characters to be used would likely be hard-coded, U+2588 - U+258F, as customization on that is rarely useful.)
2. The set of characters to be used in a monochrome terminal. It is defined in `BarMeterMode_characters`. If we want customizability, it should be customizing that set instead of just one character. But still, I can't see how that customization can be useful.

BenBE:
> @rustedusted There are two main things to address regarding the whole issue of Bar meter characters:
> 
> 1. The "subpixel" rendering. (Technically it's character cell not pixel, but anyway.) It would require an algorithm rewrite, and while I had some ideas on the algorithm (because I wrote the wonderful [Dynamic scaling & Graph meter coloring (new) #714](https://github.com/htop-dev/htop/pull/714) feature that's still waiting to be merged), I am still unsure about how useful this "subpixel" rendering of bars can provide comparing to the code complexity that will add. (Also, once implemented, the set of characters to be used would likely be hard-coded, U+2588 - U+258F, as customization on that is rarely useful.)

From a technical PoV the complexity is basically just: Treat the bar meter as having 8 times more space and use the remainder when dividing by 8 as the index into the character table.

One reason for #714 not being merged yet is the heap of complexity it introduces despite what it tries to achieve. That's also why I pushed the "per meter" flags for rendering, as this allows different meter renderings to be used where necessary, thus reducing the complexity in the meter rendering (only has to care that its exact rendering works properly for the data provided).

> 2. The set of characters to be used in a monochrome terminal. It is defined in `BarMeterMode_characters`. If we want customizability, it should be customizing that set instead of just one character. But still, I can't see how that customization can be useful.

That set of characters is meant for a different feature and should be handled as such. That other feature is not subject of this PR. Please don't mix up different things.

rustedusted:
Hey there sorry for the delay was getting myself acquainted with codebase so i could solve take on more issues

added the sub pixel rendering, however as you might see the bars are distancing themselves between two differently colored bars

should i add sub pixel only for the last bar and keep other bars non subpixel so that it seems consistent instead of disjoint??
or should i make the background of the ending characters of bars to be the foreground color of next bar?

if you have any other idea please let me know i feel like the next commits will be much faster now that i know significantly more about the code ig :smile:

Meter.c
```diff
@@ -124,29 +125,51 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
...
       const Settings* settings = this->host->settings;
+      const char* bars[7][8] = {
+         {"|","|","|","|","|","|","|","|"},
+         {"#","#","#","#","#","#","#","#"},
+         {"⡀","⡄","⡆","⡇","⣇","⣧","⣷","⣿"},
+         {"░","░","▒","▒","▓","▓","█","█"},
+         {"▏","▎","▍","▌","▋","▊","▉","█"},
+         {"▁","▂","▃","▄","▅","▆","▇","█"},
+         {"▌","▌","▌","▌","█","█","█","█"}
+      };
```
BenBE, line 151: Include the whitespace character for the various sets, as for e.g. braile you want to use the "zero dots" codepoint, instead of the normal blank character for consistency.

Also you can use `" ||||||||"`, which while using one byte more is more compact to write. Furthermore, your type should be:

```suggestion
      static const char* bars[7] = {
         " ||||||||",
         " ########",
         " ⡀⡄⡆⡇⣇⣧⣷⣿",
         " ░░▒▒▓▓██",
         " ▏▎▍▌▋▊▉█",
         " ▁▂▃▄▅▆▇█",
         " ▌▌▌▌████",
      };
```
(Used blank for all of them for now, please update the blanks for the graphical ones)

Meter.c
```diff
@@ -124,29 +125,51 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
...
                assert(i < strlen(BarMeterMode_characters));
                RichString_setChar(&bar, startPos + j, BarMeterMode_characters[i]);
             } else {
-               RichString_setChar(&bar, startPos + j, '|');
+               // barType 1 and 2 both are ascii characters | and # hence will be handled here
+               RichString_setChar(&bar, startPos + j, bars[settings->barType][0][0] );
             }
          }
+
+      int subPixel = floor(barLen*8);
```
BenBE, line 168: Possibly multiply by `strlen(bars[settings->barType])-1` to allow styles with different number of characters.

Meter.c
```diff
@@ -124,29 +125,51 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
...
+      wchar_t a;
+      mbstowcs(&a,bars[settings->barType][subPixel%8] ,1);
```
BenBE, line 170: Lookup with similar `%(strlen(…) - 1)`

DisplayOptionsPanel.c
```diff
@@ -156,6 +156,7 @@ DisplayOptionsPanel* DisplayOptionsPanel_new(Settings* settings, ScreenManager*
    Panel_add(super, (Object*) CheckItem_newByRef("Highlight new and old processes", &(settings->highlightChanges)));
    Panel_add(super, (Object*) NumberItem_newByRef("- Highlight time (in seconds)", &(settings->highlightDelaySecs), 0, 1, 24 * 60 * 60));
    Panel_add(super, (Object*) NumberItem_newByRef("Hide main function bar (0 - off, 1 - on ESC until next input, 2 - permanently)", &(settings->hideFunctionBar), 0, 0, 2));
+   Panel_add(super, (Object*) NumberItem_newByRef("Bar Type (0-7)", &(settings->barType), 0, 0, 6));
```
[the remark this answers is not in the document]
rustedusted: oh yeah forgot to do that Thanks for spotting that one out

Meter.c
rustedusted: Will commit everything bundled in one commit 

i'll change the data type from instead of char to wchar so that we won't have to worry about converting char to wchar on everytime requested

rustedusted: thanks for the changes i was quite uncertain while writing this :D

also i had to ask why don't we use normal space ?
and which is the utf8 character that you're referring to instead of the space.

BenBE: For Braille you can use this one:
https://www.compart.com/en/unicode/U+2800

BenBE: Reason for using other ones is for consistency, because sometimes the spacehas a different width in fonts compared to the default blank …

rustedusted: > Reason for using other ones is for consistency, because sometimes the spacehas a different width in fonts compared to the default blank …

damn i didn't even think about this you're a genius :exploding_head:

rustedusted: i will check the website for other characters don't worry about it :D
btw i will have to keep spaces for | and # right?

BenBE: Yes. Only for the grapical sets you may want to change things if necessary.

Explorer09: @BenBE 

> Reason for using other ones is for consistency, because sometimes the spacehas a different width in fonts compared to the default blank …

This is not an issue in monospace terminals. The real issue you should worry, however, is that certain block characters will occupy two character width in East Asian locales. It's true for these blocks `▏▎▍▌▋▊▉█`.

This is known as [East Asian Width](https://www.unicode.org/reports/tr11/) property. And you use `wcwidth(3)` to get this property in C.

Data from Unicode:
```text
2580..258F     ; A  # So    [16] UPPER HALF BLOCK..LEFT ONE EIGHTH BLOCK
2590..2591     ; N  # So     [2] RIGHT HALF BLOCK..LIGHT SHADE
2592..2595     ; A  # So     [4] MEDIUM SHADE..RIGHT ONE EIGHTH BLOCK
2596..259F     ; N  # So    [10] QUADRANT LOWER LEFT..QUADRANT UPPER RIGHT AND LOWER LEFT AND LOWER RIGHT
```

rustedusted: hey there tried finding other characters but couldn't find others
i thought it would be easy 

what do i do for the remaining characters??
people online suggested to use spaces

Explorer09: > hey there tried finding other characters but couldn't find others i thought it would be easy
> 
> what do i do for the remaining characters?? people online suggested to use spaces

For block characters, Unicode recommendeds U+00A0 No-break Space for empty block.

Explorer09: For completeness here is another set: The upper filled blocks. Requires Unicode 13 or later.

`\u00A0\u2594\U0001FB82\U0001FB83\u2580\U0001FB84\U0001FB85\U0001FB86\u2588;`
`▔🮂🮃▀🮄🮅🮆█`

rustedusted: oh hey there thanks exlorer will add the no break space for all other places

and will add the other character set you recommended

i had to ask should i add it in reverse order??

rustedusted: or would the bars be starting from upside down?

Meter.c
rustedusted: this change is done :D

thanks for this one  it looks much more cleaner now

Meter.c
rustedusted: this has been done as well :)

Meter.c
rustedusted: Just made the latest commit see how it feels would love to know suggestions
@BenBE @Explorer09 
Thanks for helping me out

Explorer09: I might be crazy suggesting this, but with Unicode 16 another set could become available.
This is like the braille set `⡀⡄⡆⡇⣇⣧⣷⣿` but with dots replaced with blocks.

`\u00A0\U0001CEA3\u2596\U0001CD48\u258C\U0001CDC0\u2599\U0001CDE4\u2588`
`𜺣▖𜵈▌𜷀▙𜷤█`

This is pretty new, so not many fonts support rendering these. [Cascadia Code](https://github.com/microsoft/cascadia-code) is one of the fonts supporting these new characters.

rustedusted: @BenBE what do you think about this ??

should i add it ?
if yes,
should i replace it instead of the currently used characters for braille?

or should we add utf16 later ??

rustedusted: btw adding utf16 is as easy as just adding new line in the array

BenBE: I think, Unicode 13 (March 2020) is fine. For Unicode 16 (September 2024) we can add this as a commented out inclusion, but should not compile it in by default (people shall edit their code if they wish to use it). Regarding the Unicode 16 variant I also have my doubts, if this set should be abbreviated or if there is a slightly extended version. Also for braille you could even do 0 to 8 dots starting at the bottom and adding the bottom-left most empty dot.

P.S.: There's a tool called `charmap` that can be useful to look up these characters.

Meter.c
```diff
@@ -124,26 +125,48 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
...
+         L" ########",
+         L"⠀⡀⡄⡆⡇⣇⣧⣷⣿",
+         L" ░░▒▒▓▓██",
+         L" ▏▎▍▌▋▊▉█",
+         L" ▁▂▃▄▅▆▇█",
+         L" ▌▌▌▌████",
+         L" ▔🮂🮃▀🮄🮅🮆█"
+      };
+
       for (int j = offset; j < nextOffset; j++)
```
BenBE, line 154: Can you add braces for this `for` loop? TIA.

Meter.c
```diff
@@ -124,26 +125,48 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
...
                assert(i < strlen(BarMeterMode_characters));
                RichString_setChar(&bar, startPos + j, BarMeterMode_characters[i]);
             } else {
                RichString_setChar(&bar, startPos + j, '|');
             }
          }
+
+      int barsLen=wcslen(bars[settings->barType]);
+      int subPixel = floor(actualBarWidth*(barsLen-1));
+      RichString_setChar(&bar, startPos + nextOffset - 1, bars[settings->barType][subPixel%(barsLen-1)]);
```
BenBE, line 168: Code style:

```suggestion
      const wchar_t *barChars = &bars[settings->barType][1];
      int barsLen = wcslen(barChars);
      int subPixel = floor(actualBarWidth * barsLen);
      RichString_setChar(&bar, startPos + nextOffset - 1, barChars[subPixel % barsLen]);
```

Storing the pointer to `bars[settings->barType][1]` simplifies the offset calculations below as the space does not need to be skipped. Empty cells still should use `bars[settings->barType][0]` and full ones `bars[settings->barType][barsLen]`.

Looking at the code, the floor at that position does not look right. Please re-check.

Meter.c
rustedusted: will do that :D

Meter.c
rustedusted: so is this done??

rustedusted:
i have added everything in the following commits that was suggested

if there is anything more please let me know

Meter.c
rustedusted: done :D

Meter.c
rustedusted: if there is anything yet to add please do let me know

Meter.c
```diff
@@ -124,26 +125,50 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
 
    // First draw in the bar[] buffer...
    int offset = 0;
+   double actualBarWidth=0;
```
BenBE, line 128: You should be able to avoid this floating point code by defining `int wsub = w * barsLen;` (rearranging code may apply) and using `wsub` instead of `w` whenever referring to the length of the bar, instead of actual cell position on the screen.

Meter.c
```diff
@@ -124,26 +125,50 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
...
    for (uint8_t i = 0; i < this->curItems; i++) {
       double value = this->values[i];
       if (isPositive(value) && this->total > 0.0) {
          value = MINIMUM(value, this->total);
-         blockSizes[i] = ceil((value / this->total) * w);
+         actualBarWidth = ((value / this->total) * w);
```
BenBE, line 133: The outer parens aren't necessary and should be removed.

BenBE (changes requested):
Please rebase your PR on top of main instead of merging. Also fixes to existing commits should be squashed into the appropriate commits instead of cluttering the history with trial and error: The commit history of a PR should show, what needs to be done to implement a feature, not every turn you took on the way to get there.

Adding the braces can (and should) stay a separate commit, as it only does code style changes and doesn't touch any functionality.

Meter.c
rustedusted: done

rustedusted:
Oh hey there you're right about squashing I didn't know a lot about git just learned few things will make a new pull request to do this 

I'm really very sorry for the inconvenience

Explorer09:
@rustedusted What are you doing? Try [git rebase](https://git-scm.com/docs/git-rebase) with `-i` option and it can help you with many things. No need to make another pull request.

rustedusted:
@Explorer09 i was trying to merge main i did not create a new branch

rustedusted:
in fact i didn't know i had to create a new branch for that matter
this was my first commit tbh 

is there a way to recover this?!

rustedusted:
image


this is the entire history of commits all the commits should be squashed ig

Explorer09:
@rustedusted I don't recommend breaking existing discussions so I would suggest you to keep this branch `main` for now unless you really want the PR to be closed.

You can set up a new branch `upstream-main` (using `git branch` command) and have `upstream-main` follow the `main` of this repository, and you won't get conflicts.

rustedusted:
hey there just an  update i'm working on this issue couldn't work on this yesterday will definitely try putting a commit tomorrow

rustedusted:
i've made a branch with appropriate changes named new-stream 

do i make a new pull request??
@Explorer09

rustedusted:
hey also wanted to try out solving other issues can i take a new issue as i feel like this one is about to be solved ig idk tbh:|

rustedusted:
can you tell me how to do it i asked on irc and stuff but couldn't get much answers

the branch that i made pr with is main 
and the branch that contains updated and correct changes is called new-stream

how do i make the origin/main of my repo follow new-stream in my local 

i tried doing git push -u origin main 
while checked out in new-stream

most people are just telling me to make new pull request

BenBE:
> how do i make the origin/main of my repo follow new-stream in my local

```
git checkout new-stream # Checkout the branch you want to modify
git branch --set-upstream-to=origin/main # Update the upstream marker
git push --force-with-lease # Send the (new) commits to the server after rewriting history
```

rustedusted:
hey there finally done that thank you so much for the gb --set-upstream-to command i didn't even think it was possible

i have done the appropriate changes tell me any more changes if there are any

BenBE:
Rearranged some lines to better group the changes. Also fixed some code style issues.

rustedusted:
what are the issues here?!! it says 15 failing did i do something wrong?!

rustedusted:
Thanks for helping me out @BenBE 
also i had to ask if there's any way to solve all the checks

BenBE:
> Thanks for helping me out @BenBE also i had to ask if there's any way to solve all the checks

Have you tried compiling this locally?

From the build log:
```
depbase=`echo DisplayOptionsPanel.o | sed 's|[^/]*$|.deps/&|;s|\.o$||'`;\
gcc -DHAVE_CONFIG_H -I.  -DNDEBUG -DGCC_PRINTF -DGCC_SCANF -D_FORTIFY_SOURCE=2 -std=c99 -pedantic -D_DEFAULT_SOURCE -D_XOPEN_SOURCE=600  -Wall -Wcast-align -Wcast-qual -Wextra -Wfloat-equal -Wformat=2 -Winit-self -Wmissing-format-attribute -Wmissing-noreturn -Wmissing-prototypes -Wpointer-arith -Wshadow -Wstrict-prototypes -Wundef -Wunused -Wwrite-strings -Wnull-dereference -Werror -D_XOPEN_SOURCE_EXTENDED -DSYSCONFDIR="\"/usr/local/etc\"" -I"./linux" -g -O2 -I/usr/include/libunwind -MT DisplayOptionsPanel.o -MD -MP -MF $depbase.Tpo -c -o DisplayOptionsPanel.o DisplayOptionsPanel.c &&\
DisplayOptionsPanel.c: In function ‘DisplayOptionsPanel_new’:
DisplayOptionsPanel.c:159:[69](https://github.com/htop-dev/htop/actions/runs/11323101825/job/31485077594?pr=1546#step:7:70): error: passing argument 2 of ‘NumberItem_newByRef’ from incompatible pointer type [-Werror=incompatible-pointer-types]
  159 |    Panel_add(super, (Object*) NumberItem_newByRef("Bar Type (0-7)", &(settings->barType), 0, 0, 7));
      |                                                                     ^~~~~~~~~~~~~~~~~~~~
      |                                                                     |
      |                                                                     uint8_t * {aka unsigned char *}
In file included from DisplayOptionsPanel.c:20:
OptionItem.h:[73](https://github.com/htop-dev/htop/actions/runs/11323101825/job/31485077594?pr=1546#step:7:74):56: note: expected ‘int *’ but argument is of type ‘uint8_t *’ {aka ‘unsigned char *’}
   73 | NumberItem* NumberItem_newByRef(const char* text, int* ref, int scale, int min, int max);
      |                                                   ~~~~~^~~
```

rustedusted:
yeah i did compile it on my machine i did that using 
`./autogen.sh && ./configure --enable-debug --enable-unicode && make && ./htop
`

the above command 

it worked at least when i've tried it on my computer

i'm using ubuntu 22.04 right now

rustedusted:
oh hey there very sorry i did not notice the warning i thought we were compiling with Werror and Wall flag

rustedusted:
done i don't think there should be any warnings now also i saw the snippet you sent 

this was a message in the snippet you sent i don't get such messages in fact it just showed warnings and ran without any issue

BenBE:
See: https://github.com/htop-dev/htop/blob/4102862d12695cdf003e2d51ef6ce5984b7136d7/.github/workflows/ci.yml#L17-L29

Pass `--enable-werror` to `./configure` …

Meter.c
```diff
@@ -120,40 +132,61 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
    assert(startPos <= w);
    assert(startPos + w <= RichString_sizeVal(bar));
 
+   const Settings* settings = this->host->settings;
    int blockSizes[10];
 
    // First draw in the bar[] buffer...
    int offset = 0;
+   const wchar_t* barChar = &bars[settings->barType][1];
+   const uint8_t barLen = wcslen(barChar);
```
Explorer09, line 141: No sense to downcast this local variable to `uint8_t`. Just use `size_t` for `barLen`.

Meter.c
```diff
@@ -120,40 +132,61 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
...
    // ...then print the buffer.
    offset = 0;
    for (uint8_t i = 0; i < this->curItems; i++) {
       int attr = this->curAttributes ? this->curAttributes[i] : Meter_attributes(this)[i];
-      RichString_setAttrn(&bar, CRT_colors[attr], startPos + offset, blockSizes[i]);
-      RichString_printoffnVal(bar, y, x + offset, startPos + offset, MINIMUM(blockSizes[i], w - offset));
-      offset += blockSizes[i];
+      RichString_setAttrn(&bar, CRT_colors[attr], startPos + offset, ceil(blockSizes[i]));
+      RichString_printoffnVal(bar, y, x + offset, startPos + offset, MINIMUM(ceil(blockSizes[i]), w - offset));
+      offset += ceil((double)blockSizes[i]);
```
Explorer09, line 189: I see `ceil((double)blockSizes[i])` and `ceil(blockSizes[i])` and they look like two different calculations. `blocksizes` was defined as `int[10]` according to the code above, so the `ceil` function won't do what you want.

Meter.c
```diff
@@ -120,40 +132,61 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
...
          blockSizes[i] = ceil((value / this->total) * w);
       } else {
          blockSizes[i] = 0;
       }
       int nextOffset = offset + blockSizes[i];
       // (Control against invalid values)
       nextOffset = CLAMP(nextOffset, 0, w);
-      for (int j = offset; j < nextOffset; j++)
+
+
```
Explorer09, line 164: One unneeded newline here.

Meter.c
```diff
@@ -120,40 +132,61 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
...
+   const wchar_t* barChar = &bars[settings->barType][1];
+   const uint8_t barLen = wcslen(barChar);
+   const uint8_t wsub = w * barLen;
+
    for (uint8_t i = 0; i < this->curItems; i++) {
       double value = this->values[i];
+      int actualWidth = 0;
+
+      // ignore extremely small values
+      if((value / this->total) * wsub < 0.5){
```
Explorer09, line 149: Spacing
```suggestion
      if ((value / this->total) * wsub < 0.5) {
```

Meter.c
```diff
@@ -120,40 +132,61 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
    assert(startPos <= w);
    assert(startPos + w <= RichString_sizeVal(bar));
 
+   const Settings* settings = this->host->settings;
    int blockSizes[10];
 
    // First draw in the bar[] buffer...
    int offset = 0;
+   const wchar_t* barChar = &bars[settings->barType][1];
```
Explorer09, line 140: Since this is still an array, you should name it with plural noun, that is, `barChars`.

Meter.c
rustedusted: oh yes you're right i had tried some other stuff thinking it would've been more readable forgot to remove that :+1: thanks for noticing :D

Meter.c
rustedusted: oh will do that

Meter.c
rustedusted: oh will do that

Meter.c
rustedusted: oh will do that

Meter.c
rustedusted: done

rustedusted:
all suggested changes have been made :D

BenBE:
And some of the updates I did were dropped …

```diff
diff --git a/Meter.c b/Meter.c
index 00311a57..21e2afd4 100644
--- a/Meter.c
+++ b/Meter.c
@@ -132,14 +132,15 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
    assert(startPos <= w);
    assert(startPos + w <= RichString_sizeVal(bar));
 
-   const Settings* settings = this->host->settings;
    int blockSizes[10];
 
    // First draw in the bar[] buffer...
    int offset = 0;
+
+   const Settings* settings = this->host->settings;
    const wchar_t* barChars = &bars[settings->barType][1];
    const size_t barLen = wcslen(barChars);
-   const uint8_t wsub = w * barLen;
+   const size_t wsub = w * barLen;
 
    for (uint8_t i = 0; i < this->curItems; i++) {
       double value = this->values[i];
@@ -150,6 +151,7 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
          blockSizes[i] = 0;
          continue;
       }
+
       if (isPositive(value) && this->total > 0.0) {
          value = MINIMUM(value, this->total);
          actualWidth = ceil((value / this->total) * wsub);
@@ -161,7 +163,7 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
       // (Control against invalid values)
       nextOffset = CLAMP(nextOffset, 0, w);
 
-      for (int j = offset; j < nextOffset; j++){
+      for (int j = offset; j < nextOffset; j++) {
          if (RichString_getCharVal(bar, startPos + j) == ' ') {
             if (CRT_colorScheme == COLORSCHEME_MONOCHROME) {
                assert(i < strlen(BarMeterMode_characters));
@@ -174,7 +176,7 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
          }
       }
 
-      RichString_setChar(&bar, startPos + nextOffset-1, barChars[actualWidth % barLen]);
+      RichString_setChar(&bar, startPos + nextOffset - 1, barChars[actualWidth % barLen]);
 
       offset = nextOffset;
    }
```

rustedusted:
oh sorry did not notice that i guess the changes are made now but 

image


this is what i get

rustedusted:
is there anything remaining

rustedusted:
it was division with zero solved that issue :D was slightly confused for some time

rustedusted:
solved all the checks

rustedusted:
now i don't know what can go wrong

rustedusted:
if there's anything do let me know :D

BenBE:
If you reword that comparison you can avoid the division completely … Although I'm not fully sure if that short-hand is needed.

rustedusted:
oh are you referring to line 157 in Meter.c

well the thing is that sometimes value is extremely small so just to avoid a single level it felt correct

most times the value is extremely small so it just occupies a single character space for a character containing mostly empty space something like ⡀
in braille

you can try removing it and if you feel that is correct then we can use it

Meter.c
```diff
@@ -124,26 +136,48 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
...
+
       if (isPositive(value) && this->total > 0.0) {
          value = MINIMUM(value, this->total);
+         actualWidth = ceil((value / this->total) * wsub);
          blockSizes[i] = ceil((value / this->total) * w);
       } else {
          blockSizes[i] = 0;
       }
+
+      if ((value / this->total) * wsub < 0.5) {
```
BenBE, line 157: Avoid division by zero …
```suggestion
      if (value * wsub < 0.5 * this->total) {
```

rustedusted: oh right sorry for that made the same mistake

rustedusted: oh wait i think making just one if condition is possible

see the new changes and see if they seem fine

because we need to continue even when total is zero

rustedusted:
it works as expected in this as well removing if might actually be a good method

rustedusted:
i don't know though i'm a newbie

rustedusted:
hey there any updates to this ?!

PR #1547, opened 2024-10-06
README: clarification on kernel requirements for I/O

dilyanpalauzov:
After reducing the kernel features, I found that htop showed for DISK R/W, DISK READ, DISK WRITE `N/A`.  Then I figured out when the file `/proc/$PID/task/$PID/io` was maintained and here comes the outcome.


natoscott (changes requested):
This relates to Linux specifically, so please move this 5 lines down from where you have it (into the markdown section with "Linux" header).  There's no need for the comma in your sentence there either.

natoscott (approved):
LGTM, thanks.

PR #1551, opened 2024-10-11
Adding different characters than default pipe for bar |

rustedusted:
this pull request addresses issue #647


rustedusted:
did not squash commits yet please tell me suggested changes

rustedusted:
will squash at the end when everything seems correct

PR #1554, opened 2024-10-12
Remove duplicate function declaration in Row.h

### LINKED ISSUE #1553: Row.h has a duplicate function declaration.
This function declaration was made twice
`void Row_display(const Object* cast, RichString* out);`
on line [125](https://github.com/htop-dev/htop/blob/261df1c685b93b3f21c7c26cbffd882997f351dc/Row.h#L125) and line [165](https://github.com/htop-dev/htop/blob/261df1c685b93b3f21c7c26cbffd882997f351dc/Row.h#L165).

While reading the source code to understand the project, I found this.
I'm sorry if this was not such an important issue. I'm just at my learning phase and new to open-source contributions and to this project.

Thanks.

AryanGitHub:
just removed the line 125th line. 

`void Row_display(const Object* cast, RichString* out);`
Fixes #1553

Thanks.

### DIFF
--- a/Row.h
+++ b/Row.h
@@ -162,8 +162,6 @@ void Row_printRate(RichString* str, double rate, bool coloring);
 
 int Row_printPercentage(float val, char* buffer, size_t n, uint8_t width, int* attr);
 
-void Row_display(const Object* cast, RichString* out);
-
 static inline int Row_idEqualCompare(const void* v1, const void* v2) {
    const int p1 = ((const Row*)v1)->id;
    const int p2 = ((const Row*)v2)->id;


BenBE:
Thank you for your PR.

The duplicate declaration of a function in C isn't usually a bug, as you can have as many of them as you like as long as they all agree with each other.

In the context of `Row.h` there is a small difference between functions called `Row_display` (present current state) and `Row_print…` (prepare this visual state). Thus keeping the first one in favour of the second one below makes semantically a bit more sense, as it keeps the actions an user want's to perform with the row (create, free, display its contents) together, while having the operations needed for maintenance (filling its presentation with data to display) in a group below.

Issue #991, opened 2022-05-03
`DISK READ`/`DISK WRITE` column changes


BenBE:
The columns `DISK READ` and `DISK WRITE` currently show as `%5.2f %sB/s` like e.g. `0.00 B/s`. This makes those columns quite long compared to other columns. Also showing whole bytes with two significant digits is somewhat unnecessary precision.

What about shortening the column labels to `RDISK`/`WDISK` and the display of the actual column value as follows:

0. Values below 1KiB/s are shaded
1. Values below 1KiB/s displayed in bytes
2. Starting with 1KiB/s values receive 3-4 significant digits and the usual unit suffices.
3. Shading of MiB/s, GiB/s and TiB/s done similar to normal byte counts.

Currently each of the two column takes up 10 characters, when similar columns like `RCHAR`/`WCHAR` (Number of bytes read/written) use 5 chars.

NikhilReddy-2:
Hey, would love to take this up if there hasn't been any work done on it yet as this looks like a good starting point to get into this repo.

Has there been any further discussion on this on any other PRs?

BenBE:
There has not been any further discussion I'm aware of. Usually issues are heavily cross-referenced, thus if nothing shows here, it's unlikely there was essential discussion.

Thus feel free to take this one up. There's some documentation on the code style available in the docs folder. Also don't hesitate to ask if you've got any questions.

AryanGitHub:
Can we also reduce the column size of `DISK R/W` (Total I/O rate in bytes per second)? Its column title size is also 12 characters ("   DISK R/W ") along with `DISK READ/DISK WRITE`. All three of these process fields are using "Row_printRate" in Rox.c.

If we consider fixing all three issues, we can directly change the function "Row_printRate" in Rox.c to address this.

I also wanted to know if, while changing the total width of "Row_printRate," we need to modify this line in `./pcp/PCPDynamicColumn.c`:
`column->super.width = 11; // Row_printRate` [here](https://github.com/htop-dev/htop/blob/4102862d12695cdf003e2d51ef6ce5984b7136d7/pcp/PCPDynamicColumn.c#L298). 

I am asking this because I was unable to build the PCP version and test it myself [here](https://github.com/htop-dev/htop/pull/1337#issuecomment-2418485547).

PR #1557, opened 2024-10-19
Enhance Readability of DISK READ/DISK WRITE Columns by Shortening Labels and Values

AryanGitHub:
### Observations before addressing GitHub issue #991:

- The column titles for **RCHAR** & **WCHAR** take [6 characters](https://github.com/htop-dev/htop/blob/4102862d12695cdf003e2d51ef6ce5984b7136d7/linux/LinuxProcess.c#L79), while the titles for **DISK READ**, **DISK WRITE**, and **DISK R/W** (representing total I/O rate in bytes per second) take [12 characters](https://github.com/htop-dev/htop/blob/4102862d12695cdf003e2d51ef6ce5984b7136d7/linux/LinuxProcess.c#L86). This PR reduces them to 9 characters.

- Why not make the column titles for DISK operations consistent with **RCHAR** & **WCHAR**?

- The **RCHAR** & **WCHAR** columns use an `unsigned long long` data type, while **DISK READ** & **DISK WRITE** use `double`, which adds an extra character for the decimal point when printing.

- The **DISK READ** & **DISK WRITE** columns represent rates, which include a time dimension, thus requiring 2 additional characters compared to **RCHAR** & **WCHAR**.

- In total, we need 3 extra characters for **DISK READ**, **DISK WRITE**, and **DISK R/W** compared to **RCHAR** & **WCHAR**.

### What has this PR changed?
This PR changed the column titles as required and
introduced a dynamic precision variable, adjusting based on the value to be printed. The total number of numerical characters allowed for printing is 4. The precision adjusts as follows:

- If the value is between **0-9**, the precision is 3, e.g., `3.141`.
- If the value is between **10-99**, the precision is 2, e.g., `31.41`.
- If the value is greater than **99**, the precision is 1, e.g., `314.1`.
- If the value is greater than **999**, the precision is 0, e.g., `3141`.


AryanGitHub:
I also wanted to know when, while changing the total width of "`Row_printRate`", do we need to modify this line in ./pcp/PCPDynamicColumn.c:
 https://github.com/htop-dev/htop/blob/4102862d12695cdf003e2d51ef6ce5984b7136d7/pcp/PCPDynamicColumn.c#L298
 

This is because I don't know, what this does. please inform me about this, I will try to make the changes accordingly.

Row.c
```diff
@@ -468,27 +475,34 @@ void Row_printRate(RichString* str, double rate, bool coloring) {
...
+      RichString_appendAscii(str, shadowColor, "     N/A ");
    } else if (rate < ONE_K) {
-      int len = snprintf(buffer, sizeof(buffer), "%7.2f B/s ", rate);
+      int len = snprintf(buffer, sizeof(buffer), "%5.0fB/s ", rate);
       RichString_appendnAscii(str, baseColor, buffer, len);
    } else if (rate < ONE_M) {
-      int len = snprintf(buffer, sizeof(buffer), "%7.2f K/s ", rate / ONE_K);
+      double rateToPrint =  rate / ONE_K;
+      unsigned char precision = Row_getPrecision (rateToPrint);
+      int len = snprintf(buffer, sizeof(buffer), "%5.*fK/s ", precision, rateToPrint);
```
Explorer09: 1. `precision` takes the type `int`. You should set the data type properly.
2. There is a potential of converting this into a loop. You can see the code of `Meter_humanUnit()` function for how it can be done. There is a global `unitPrefixes` array that you can utilize.

Row.c
```diff
@@ -454,6 +454,13 @@ void Row_printNanoseconds(RichString* str, unsigned long long totalNanoseconds,
    Row_printTime(str, totalHundredths, coloring);
 }
 
+static inline unsigned char Row_getPrecision (double displayRate){
+   if (displayRate < 10) return 3;
+   else if (displayRate < 100) return 2;
+   else if (displayRate < 1000) return 1;
```
Explorer09, line 460: Consider there is a rounding issue with this approach. Look at `Meter_humanUnit()` function to see how I addressed that.

BenBE:
The spacing for `RDISK`, `WDISK` and `RW DISK` is inconsistent.

Further more, `DIOR`, `DIOW` and `DIORW` would be even shorter and even consistent.

From the display side of things: There's a list of IEC unit prefixes for a reason. Please use that one to avoid code duplication.

Also, the space before `B/s` is intentional to ease reading `8B/s`. Which can be hard to discern with some console fonts.

AryanGitHub:
```c
Row.c:511:13: warning: 6th function call argument is an uninitialized value [core.CallAndMessage]
  511 |       len = snprintf(buffer, sizeof(buffer), "%c/s ", unitPrefixes[unitPrefixIndex]);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/x86_64-linux-gnu/bits/stdio2.h:77:3: note: expanded from macro 'snprintf'
   77 |   __builtin___snprintf_chk (str, len, __USE_FORTIFY_LEVEL - 1,                \
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   78 |                             __glibc_objsize (str), __VA_ARGS__)
      |                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  ```
  
  What is left uninitialized? I am unable to understand.

**Comment by Explorer09** [comment]
> ```c
> Row.c:511:13: warning: 6th function call argument is an uninitialized value [core.CallAndMessage]
>   511 |       len = snprintf(buffer, sizeof(buffer), "%c/s ", unitPrefixes[unitPrefixIndex]);
>       |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> /usr/include/x86_64-linux-gnu/bits/stdio2.h:77:3: note: expanded from macro 'snprintf'
>    77 |   __builtin___snprintf_chk (str, len, __USE_FORTIFY_LEVEL - 1,                \
>       |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
>    78 |                             __glibc_objsize (str), __VA_ARGS__)
>       |                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> ```
> 
> What is left uninitialized? I am unable to understand.

Potential reading out of bounds. `unitPrefixes[-1]` unhandled (you probably want it to print `B` instead.)

File: Row.c [hunk 3]
```diff
@@ -468,28 +468,49 @@ void Row_printRate(RichString* str, double rate, bool coloring) {
...
-      int len = snprintf(buffer, sizeof(buffer), "%7.2f M/s ", rate / ONE_M);
-      RichString_appendnAscii(str, megabytesColor, buffer, len);
-   } else if (rate < ONE_T) {
-      int len = snprintf(buffer, sizeof(buffer), "%7.2f G/s ", rate / ONE_G);
-      RichString_appendnAscii(str, largeNumberColor, buffer, len);
-   } else if (rate < ONE_P) {
-      int len = snprintf(buffer, sizeof(buffer), "%7.2f T/s ", rate / ONE_T);
-      RichString_appendnAscii(str, largeNumberColor, buffer, len);
+      int len = snprintf(buffer, sizeof(buffer), "%4.0f B/s ", rate);
+      RichString_appendnAscii(str, shadowColor, buffer, len);
```
Explorer09, line 474: Why using `shadowColor`? Ideally nonzero values should not use that text color -- the visual of that would let users skip that entry.

Row.c
```diff
@@ -468,28 +468,49 @@ void Row_printRate(RichString* str, double rate, bool coloring) {
...
    } else {
-      int len = snprintf(buffer, sizeof(buffer), "%7.2f P/s ", rate / ONE_P);
-      RichString_appendnAscii(str, largeNumberColor, buffer, len);
+      size_t unitPrefixIndex = 0;
+      do {
+         if (unitPrefixIndex > ARRAYSIZE(unitPrefixes)-1) {
+            int len = snprintf(buffer, sizeof(buffer), "     INF ");
```
Explorer09, line 479: I think you can spell the word "infinity" out. There is enough space.

```suggestion
            int len = snprintf(buffer, sizeof(buffer), "infinity ");
```

```

File: Row.c (hunk 3 above)
Comment (AryanGitHub) [reply Row.c:474]: I used it because it was mentioned in the Issue.
https://github.com/htop-dev/htop/issues/991#issue-1223784136

File: Row.c [hunk 5]
```diff
@@ -468,28 +468,49 @@ void Row_printRate(RichString* str, double rate, bool coloring) {
...
+      }
+
+      if (precision < 3) {
+         double upper_limit = (precision < 2) ? ((precision < 1) ? 1000 : 100) : 10;
+         if (rate < upper_limit) {
+            rate = upper_limit;
+         }
+      }
+
+      int len = snprintf(buffer, sizeof(buffer), "%5.*f", precision, rate);
```
Explorer09, line 508: I thought you would want `"%4.*f"` as the format. That is, one digit less.

Output example A
```text
   0 B/s - 9999 B/s
9.77 K/s - 9.99 K/s
10.0 K/s - 99.9 K/s
 100 K/s - 9999 K/s
9.77 M/s - 9.99 M/s
     ...
 100 Q/s - 9999 Q/s
infinity
```

Output example B
```text
   0 B/s - 1023 B/s
1.00 K/s - 9.99 K/s
10.0 K/s - 99.9 K/s
 100 K/s - 1023 K/s
1.00 M/s - 9.99 M/s
     ...
 100 Q/s - 9999 Q/s
infinity
```

Row.c
Explorer09: Well. I would personally disagree with @BenBE unless these small values would practically make a lot of visual noise. But I am not the one to decide on this (I just put my suggestion).

Row.c
AryanGitHub: we actually want 4 digits.
By using `"%5.*f"`, it meant the total width of the output field is 5, including the decimal point and digits.

Explorer09: Then you would get `999.9 K/s`. I bet you didn't test that one out.

AryanGitHub:
> The spacing for `RDISK`, `WDISK` and `RW DISK` is inconsistent.
> 
> Further more, `DIOR`, `DIOW` and `DIORW` would be even shorter and even consistent.
> 
> From the display side of things: There's a list of IEC unit prefixes for a reason. Please use that one to avoid code duplication.
> 
> Also, the space before `B/s` is intentional to ease reading `8B/s`. Which can be hard to discern with some console fonts.

Should there be space for other values too, like when it is `K/s` or `M/s`, like `31.41 K/s`? Rn I have space only for `B/s`.

Explorer09:
> > Also, the space before `B/s` is intentional to ease reading `8B/s`. Which can be hard to discern with some console fonts.
> 
> Should there be space for other values too, like when it is `K/s` or `M/s`, like `31.41 K/s`? Rn I have space only for `B/s`.

I can help answer that. Yes. Because otherwise the numbers can look misaligned, when there is an entire column of the numbers.

Row.c
AryanGitHub: As its yes, should I shift to `4.*f` now, or use `5.*f`. 
I am asking this because for `4.*f` I have to change the precision logic and if it stays `5.*f` I have to add a column width, currently its 9 (for any kind of Disk IO rates), it will be then 10.

AryanGitHub: > Then you would get `999.9 K/s`. I bet you didn't test that one out.

I wanted to understand, is there any issue with printing `999.9 K/s`? I am unable to get it...

Explorer09: For size units we usually just use 3 significant digits as extra precisions are less useful for UI display purpose. Add a decimal point and it becomes 4 characters of width. When the decimal points are not needed, i.e. for numbers in range 100 - 9999, four digits can be shown for the same width, but that's more of a special case.

The numbers jump from one unit to the next in power of 1024, and that was the reason of 3 significant digits.

AryanGitHub:
I made the rate to `4.*f` and changed `INF` to `infinity`, as discussed.

Row.c
BenBE: IIRC, the current code shadows values below 1KiB/s. Fine either way IMO, but the note from @Explorer09 regarding accidentally visually skipping small values is valid.

AryanGitHub:
Thank you, everyone, for the previous review. It helped me understand the project better and improved the application overall!
I've made the requested changes, as suggested, and would appreciate it if you could review them again. :)

AryanGitHub:
> Judging from the screenshot, I think marking the entries with exactly `0 B/s` (probably `<0.5` to account for floating point stuff) as shadowed might better emphasize the non-zero values.
> 
> Additionally, what would you think of highlighting the unit in the value the same as its value? With other columns the unit marker has the same color as the most-significant unit used.

@BenBE 
- Yes, I think providing a shadow colour for `0 B/s` is good, as it emphasizes the non-zero values. because a good number of rows will be `0 B/s`, at least for Fedora Linux-based OS, which I tested.


I believe the disk read speed in htop on Linux is never between 1 and 1024 bits/s. Instead, the speed was observed to be between 0 and 1 bits/s, but never between 1 and 1024. I think the same applies to disk write speeds as well.
To test this, I created a simple if block in linux/LinuxProcessTable.c and set a breakpoint inside the if block. I let gdb run for 4 hours, but it never caught the breakpoint. 
```c
// linux/LinuxProcessTable.c
// Debug
if (lp->io_rate_read_bps > 1 && lp->io_rate_read_bps <= 1024) {
printf("breakpoint");
}
```


- I also believe that implementing colour coding for `M/s`, `G/s`, and `T/s` would be beneficial. In this image, the column is sorted based on these rate values, but that doesn't happen often.
To help the human eye observe sudden spikes in data rates across a significant number of rows, this would be helpful. Additionally, since the rate fluctuates very quickly and the read/write lasts for only a few seconds, this enhancement would be particularly useful.

Explorer09:
@AryanGitHub 

> I believe the disk read speed in htop on Linux is never between 1 and 1024 bits/s. Instead, the speed was observed to be between 0 and 1 bits/s, but never between 1 and 1024. I think the same applies to disk write speeds as well. To test this, I created a simple if block in linux/LinuxProcessTable.c and set a breakpoint inside the if block. I let gdb run for 4 hours, but it never caught the breakpoint.

I think less than 1 KiB/s may be observed if you set a long refresh rate for htop. For example one minute, so that the value would be "averaged" to less than 1 KiB.

Issue #1548, opened 2024-10-06
configure fails on curses checks missing stdscr


seidler:
The configure script on the main branch fails while it checks for curses libraries. 
It creates a `conftest.c` program which refers to a curses `stdscr` variable with
`extern void* stdscr;`
I suppose the proper way of using `stdscr` would be to use the `initscr` function. For a configuration test declaring the variable should be fair enough by removing the `extern` keyword:
`void* stdscr;`
This shall be done in `configure.ac`

I am refering to commit 9c316cc14cafce1097583247710b6d74c1feb505

seidler:
For me the following patch worked:

`diff --git a/configure.ac b/configure.ac
index 40a83063..54f6e0e2 100644
--- a/configure.ac
+++ b/configure.ac
@@ -458,7 +458,7 @@ htop_check_curses_capability () {
    AC_LINK_IFELSE([AC_LANG_PROGRAM([[
 /* extern WINDOW* stdscr; */
 /* int keypad(WINDOW* win, bool enable); */
-void* stdscr;
+extern void* stdscr;
 int keypad(void* win, int enable);
       ]], [[
 keypad(stdscr, 0);
`

BenBE:
Please provide the full configure.log and the necessary information about your system.

Declaring `extern void* stdscr;` is intetional in the test, as `configure` tries to find if the curses library provides a symbol called `stdscr`.

seidler:
The system is an Opensuse Leap Version 15.5 on an amd_64 processor.

The config.log.gz file is attached. 

[config.log.gz](https://github.com/user-attachments/files/17337471/config.log.gz)

Explorer09:
From the attached `config.log.gz`

```text
retrieving ncursesw information through /usr/bin/pkg-config
configure:7976: result: found ncursesw information through /usr/bin/pkg-config
curses cflags: -D_GNU_SOURCE -D_DEFAULT_SOURCE -I/usr/include/ncursesw
curses libs: -lncursesw -ltinfo
configure:7837: checking for keypad in -lncursesw -ltinfo
configure:7857: gcc -o conftest -std=c99 -pedantic -D_GNU_SOURCE -D_DEFAULT_SOURCE -I/usr/include/ncursesw -g -O2   conftest.c -lncursesw -ltinfo -ldl -lm  >&5
[...]

PR #1558, opened 2024-10-19
Fix "stdscr" configure failure & ncursest support

Explorer09:
This fixes the issue of #1548, while adding minor improvements to the ncurses detecttion code since #1512.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -450,18 +450,17 @@ htop_check_curses_capability () {
    # At this point we have not checked the name of curses header, so
    # use forward declaration for the linking tests below.
 
-   # htop uses keypad() and "stdscr", but for ncurses implementation,
-   # the symbols are in "-ltinfo" and not "-lncurses".
+   # htop uses keypad(), but for ncurses implementation, the symbol is
+   # in "-ltinfo" and not "-lncurses".
    # Check "-ltinfo" symbols first, as libncurses might require
    # explicit "-ltinfo" to link (for internal dependency).
    AC_MSG_CHECKING([for keypad in $htop_msg_linker_flags])
    AC_LINK_IFELSE([AC_LANG_PROGRAM([[
-/* extern WINDOW* stdscr; */
 /* int keypad(WINDOW* win, bool enable); */
-extern void* stdscr;
 int keypad(void* win, int enable);
       ]], [[
-keypad(stdscr, 0);
+static char dummy;
+keypad((void*)&dummy, 0);
       ]])],
       [AC_MSG_RESULT(yes)],
       [AC_MSG_RESULT(no)
--- a/configure.ac
+++ b/configure.ac
@@ -482,13 +482,17 @@ doupdate();
          htop_curses_status=1])
    fi
 
+   # htop calls mvadd_wchnstr(), which might be implemented as a macro.
+   # It is more reliable to test linking with wadd_wchnstr().
    if test "x$htop_curses_status$enable_unicode" = x0yes; then
-      AC_MSG_CHECKING([for mvadd_wchnstr in $htop_msg_linker_flags])
+      AC_MSG_CHECKING([for wadd_wchnstr in $htop_msg_linker_flags])
       AC_LINK_IFELSE([AC_LANG_PROGRAM([[
-/* int mvadd_wchnstr(int y, int x, const cchar_t* wchstr, int n); */
-int mvadd_wchnstr(int y, int x, const void* wchstr, int n);
+/* int wadd_wchnstr(WINDOW* win, const cchar_t* wchstr, int n); */
+int wadd_wchnstr(void* win, const void* wchstr, int n);
          ]], [[
-mvadd_wchnstr(0, 0, (void*)0, 0);
+static char dummy1;
+static char dummy2;
+wadd_wchnstr((void*)&dummy1, (void*)&dummy2, 0);
          ]])],
          [AC_MSG_RESULT(yes)
          htop_curses_capability=wide],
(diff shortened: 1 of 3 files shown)


BenBE (changes requested):
LGTM.

Document the `#if defined(HAVE_CONFIG_H)` to be used for reference inside `configure.ac`.

Also, if not done, have the ncursest specific changes re macros and the `ProvideCurses` changes live inside separate commits (though that already seems to be the case AFAICS).

configure.ac
```diff
@@ -68,6 +68,15 @@ esac
 # Enable extensions, required by hwloc scripts
 AC_USE_SYSTEM_EXTENSIONS
 
+dnl May be needed for htop Unicode support
+AH_VERBATIM([_XOPEN_SOURCE_EXTENDED],
+[/* Enables XPG4v2 (SUSv1) interfaces if they are not enabled already
+   with _XOPEN_SOURCE */
+#ifndef _XOPEN_SOURCE_EXTENDED
+#undef _XOPEN_SOURCE_EXTENDED
+#endif
```
BenBE, line 77: I think you got the check inverted by accident …

```suggestion
#ifdef _XOPEN_SOURCE_EXTENDED
#undef _XOPEN_SOURCE_EXTENDED
#endif
```

configure.ac
```diff
@@ -665,6 +684,28 @@ else
...
+#if defined(HAVE_LIBNCURSESW)
+{
+   static cchar_t dummy;
+   mvadd_wchnstr(0, 0, &dummy, 0);
+}
+#endif
+   ]])],
+   [AC_MSG_RESULT(yes)],
+   [AC_MSG_RESULT(no)
+   AC_MSG_FAILURE([there are problems with the curses header])])
```
BenBE, line 708:
```suggestion
   AC_MSG_FAILURE([there are problems with the curses header])
])
```

configure.ac
BenBE: TIL. Please add a small note about this to document this somewhat. TIA.

configure.ac
```diff
@@ -68,6 +68,23 @@ esac
...
+# _XOPEN_SOURCE_EXTENDED is defined or _XOPEN_SOURCE defined to be
+# >= 500. In DragonFly BSD, FreeBSD and OpenBSD, defining
+# _XOPEN_SOURCE to any value *hides* interfaces that would be useful
+# for htop.
+dnl
+dnl Note: The "#undef" in AH_VERBATIM will be replaced with "#define"
+dnl when config.h is generated.
+AH_VERBATIM([_XOPEN_SOURCE_EXTENDED],
+[/* Enables XPG4v2 (SUSv1) interfaces if they are not enabled already
+   with _XOPEN_SOURCE */
```
BenBE, line 81: Can be one line IMO.

```suggestion
[/* Enables XPG4v2 (SUSv1) interfaces if they are not enabled already with _XOPEN_SOURCE */
```

Explorer09:
@BenBE

I have a idea when trying to improve configure in detecting ncurses header, but am not sure whether such a change is acceptable:

```diff
--- a/ProvideCurses.h
+++ b/ProvideCurses.h
@@ -16,16 +16,13 @@ in the source distribution for its full text.
 
 // IWYU pragma: begin_exports
 
-#if defined(HAVE_NCURSESW_CURSES_H)
-#include <ncursesw/curses.h>
-#elif defined(HAVE_NCURSES_NCURSES_H)
-#include <ncurses/ncurses.h>
-#elif defined(HAVE_NCURSES_CURSES_H)
-#include <ncurses/curses.h>
-#elif defined(HAVE_NCURSES_H)
-#include <ncurses.h>
-#elif defined(HAVE_CURSES_H)
-#include <curses.h>
+// The CURSES_HEADER_NAME macro may be defined by configure script to any of these:
+// <ncursesw/curses.h>
+// <ncurses/curses.h>
+// <ncurses.h>
+// <curses.h>
+#if defined(CURSES_HEADER_NAME)
+#include CURSES_HEADER_NAME
 #endif
 
 #ifdef HAVE_LIBNCURSESW
```

Instead of having a lot of boolean `HAVE_*CURSES_H` macro tokens, I think it would be better to have one token for curses/ncurses header name. This form of include directive is supposed to work with any C89-compliant compiler, but programmers might be unfamiliar with this form of include.

BenBE:
I considered it briefly when that header was created, but decided against it, as we hate the ProvideCurses.h anyway to deal with all the other aspects of curses that come bundled with terminal control stuff. And since you get those booleans for free™ from autoconf there's no real reason not using them in favour of some little-used feature of the pre-processor. So keep it as-is.

Explorer09:
> I considered it briefly when that header was created, but decided against it, as we hate the ProvideCurses.h anyway to deal with all the other aspects of curses that come bundled with terminal control stuff. And since you get those booleans for free™ from autoconf there's no real reason not using them in favour of some little-used feature of the pre-processor. So keep it as-is.

The problem is, the list of possible header names for ncurses is non-exhaustive. We  only check header names that are common in distributions (`<ncursesw/curses.h>`, `<ncurses.h>`, etc.), but I've experimented with ncurses' configure (upstream), and it technically allows builders to specify any suffix for the library and header directory. So the header could be `ncurseswfoo/curses.h` (if ncurses is configured with `--with-extra-suffix=foo`), `ncurseswbar/curses.h` (if `--with-extra-suffix=bar`) and so on.

I know these are custom builds and niche use cases, but as I introduced `--with-curses=XXX` in htop's configure, I think I should allow builders to specify custom header for ncurses (to free them from needing to hack the ProvideCurses.h). It won't be a configure option, but a `CPPFLAGS='-DCURSES_HEADER_NAME=<ncurseswfoo/curses.h>'` override.

BenBE:
People who mess around with the header names may also just mess around with the code at that point. I doubt most programs using (n)curses at that point would even compile properly with these custom header names. That's some additional complexity added that doesn't provide reasonable benefit for the average case or is common enough to warrant going that extra kilometer.

People messing with their header names can feel lucky, that it's basically just one place they have to mess with in our code …

Explorer09:
I think I would defer the header detection changes to another PR. The rest of the changes are ready.

PR #1570, opened 2024-12-16
Avoid calling snprintf in SIGTERM handler

### LINKED ISSUE #1563: CWE-479 in CRT.c
While compiling I noticed the following warning for [CWE-479](https://cwe.mitre.org/data/definitions/479.html) from GCC 14:

```
depbase=`echo CRT.o | sed 's|[^/]*$|.deps/&|;s|\.o$||'`;\
gcc-14 -DHAVE_CONFIG_H -I.  -DNDEBUG  -std=c99 -pedantic -D_DEFAULT_SOURCE -D_XOPEN_SOURCE=600  -I/usr/include/libnl3 -Wall -Wcast-align -Wcast-qual -Wextra -Wfloat-equal -Wformat=2 -Winit-self -Wmissing-format-attribute -Wmissing-noreturn -Wmissing-prototypes -Wpointer-arith -Wshadow -Wstrict-prototypes -Wundef -Wunused -Wwrite-strings -Wnull-dereference -D_XOPEN_SOURCE_EXTENDED -DSYSCONFDIR="\"/usr/local/etc\"" -I"./linux" -fanalyzer -MT CRT.o -MD -MP -MF $depbase.Tpo -c -o CRT.o CRT.c &&\
CRT.c: In function ‘CRT_handleSIGTERM’:
CRT.c:846:4: warning: call to ‘snprintf’ from within signal handler [CWE-479] [-Wanalyzer-unsafe-call-within-signal-handler]
  846 |    snprintf(err_buf, sizeof(err_buf),
      |    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  847 |            "A signal %d (%s) was received, exiting without persisting settings to htoprc.\n",
      |            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  848 |            sgn, signal_str);
      |            ~~~~~~~~~~~~~~~~
  ‘CRT_installSignalHandlers’: events 1-2
    |
    |  952 | static void CRT_installSignalHandlers(void) {
    |      |             ^~~~~~~~~~~~~~~~~~~~~~~~~
    |      |             |
    |      |             (1) entry to ‘CRT_installSignalHandlers’
    |......
    |  966 |    signal(SIGINT, CRT_handleSIGTERM);
    |      |    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    |      |    |
    |      |    (2) registering ‘CRT_handleSIGTERM’ as signal handler
    |
  event 3
    |
    |cc1:
    | (3): later on, when the signal is delivered to the process
    |
    +--> ‘CRT_handleSIGTERM’: events 4-7
           |
           |  835 | static void CRT_handleSIGTERM(int sgn) {
           |      |             ^~~~~~~~~~~~~~~~~
           |      |             |
           |      |             (4) entry to ‘CRT_handleSIGTERM’
           |......
           |  838 |    if (!CRT_settings->changed)
           |      |       ~      
           |      |       |
           |      |       (5) following ‘false’ branch...
           |......
           |  841 |    const char* signal_str = strsignal(sgn);
           |      |                             ~~~~~~~~~~~~~~
           |      |                             |
           |      |                             (6) ...to here
           |......
           |  846 |    snprintf(err_buf, sizeof(err_buf),
           |      |    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
           |      |    |
           |      |    (7) call to ‘snprintf’ from within signal handler
           |  847 |            "A signal %d (%s) was received, exiting without persisting settings to htoprc.\n",
           |      |            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
           |  848 |            sgn, signal_str);
           |      |            ~~~~~~~~~~~~~~~~
           |
```

Explorer09:
By Implementing our own routine for formatting the number.

This could resolve #1563 but I'm sure people would say this solution is ugly.

### DIFF
--- a/CRT.c
+++ b/CRT.c
@@ -842,11 +842,51 @@ static void CRT_handleSIGTERM(int sgn) {
    if (!signal_str)
       signal_str = "unknown reason";
 
-   char err_buf[512];
-   snprintf(err_buf, sizeof(err_buf),
-           "A signal %d (%s) was received, exiting without persisting settings to htoprc.\n",
-           sgn, signal_str);
+   // snprintf(3) is not async signal safe.
+   char err_buf[80] = "Receives signal ";
+   size_t offset = strlen(err_buf);
+   do {
+      size_t spaces_left = sizeof(err_buf) - offset - 2;
+
+      size_t token_size = countDigits(sgn, 10);
+      if (spaces_left < token_size)
+         break;
+      spaces_left -= token_size;
+
+      for (; token_size > 0; token_size--) {
+         err_buf[offset + token_size - 1] = (sgn % 10) + '0';
+         sgn /= 10;
+      }
+      offset += token_size;
+
+      token_size = 1;
+      if (spaces_left < token_size)
+         break;
+      spaces_left -= token_size;
+
+      err_buf[offset] = '(';
+      offset += token_size;
+
+      token_size = strlen(signal_str);
+      token_size = MINIMUM(token_size, spaces_left);
+      spaces_left -= token_size;
+
+      memcpy(&err_buf[offset], signal_str, token_size);
+      offset += token_size;
+
+      token_size = 1;
+      if (spaces_left < token_size)
+         break;
+      spaces_left -= token_size;
+
+      err_buf[offset] = ')';
+      offset += token_size;
+   } while (0);
+   err_buf[offset] = '\n';
+   err_buf[offset + 1] = '\0';
+
    full_write_str(STDERR_FILENO, err_buf);
+   full_write_str(STDERR_FILENO, "Exiting without persisting settings to htoprc.\n");
    _exit(0);
 }
 


fasterit:
Please just use write(). The 25 lines of formatting magic are a bit over-the-top imo.

BenBE:
I agree with @fasterit here. Please shorten that formatting logic. What can (and probably should) be done is reordering things to work like:

1. Short static buffer for the signal ID num2string conversion
2. strcat the three parts into a full buffer
3. write the final resulting string.

Please don't inline the num2string conversion, but use a function which is provided with the target buffer+size. Use `"???"` if the conversion doesn't fit that buffer.

Explorer09:
I am thinking there could be another way to solve the CWE-479 problem, as that warning wasn't present in htop's fatal error reporting code, and yet the fatal error code also reports what signal it has received.

fasterit:
We could vendor a stripped down version of https://github.com/idning/safe_snprintf . The license there should be GPLv2 as the upstream MySQL code is.

fasterit:
Another implementation: https://chromium.googlesource.com/chromium/src/base/+/master/strings/safe_sprintf.h (BSD3 license)

Explorer09:
@fasterit What I said is that maybe we don't need another implementation of snprintf(3) at all.

As mentioned, GCC 14 doesn't warn about CWE-479 for `CRT_handleSIGSEGV` (maybe this is a bug - a missed warning), but warns for `CRT_handleSIGTERM`. If the signal safety of snprintf does matter for both signal handler routines, then the CWE-479 should apply to _both_ and not just one of them. And we should fix `CRT_handleSIGSEGV` as well since snprintf is also used extensibly there.

The problem of CWE-479 happens when another signal is delivered during the handling of `CRT_handleSIGSEGV` or `CRT_handleSIGTERM`, causing either function or `snprintf` within to "reenter" in an asynchronous way. Since both handlers in htop are only there to report messages, the simplest and the most proper way (I think) is to block signals when any of the messages is being constructed.

BenBE:
I think that GCC14 only warns about one of those is a bug due to a missed warning. So when we are fixing this it will need to be both. And as mentioned, I can fully live with a minimalistic `itoa`+5x`strncat` solution, if the custom-rolled `itoa` is not inlined.

Basic requirements:
1. Gets the job done
2. Signal-safe
3. Single atomic `write` to `stderr`

If we can re-use this custom-rolled `itoa` in further places, where it also improves performance I'd be the last to reject such an implementation. But that's not a priority.

Explorer09:
@BenBE I was thinking deeper that, when another signal arrives during a handler, what kind of message htop's output should look like.

To prevent the situation where we implemented our own `itoa`+`strncat` combinations only to find out we've done it wrong and the message is still corrupt like this:

```text
FATAL PROGRAM ERROR DETECTED
FATAL PROGRAM ERROR DETECTED
============================
FATAL PROGRAM ERROR DETECTED
============================
```

BenBE:
I can live with that. And for nested signals we could just go ahead and ignore the second signal.

```
// Imagine this being all atomically sound …
atomic_int8_t signal_nested = 0;
void sighandler() {
   signal_nested++;

   if(signal_nested > 1) return;

   // Actual handler code
}
```

Explorer09:
@BenBE Oops. I was confused with the CWE-479 issue here because there was actually two sub-issues:

* The use of `snprintf` in signal handlers was unsafe - It's nothing to do with what messages would be printed on error, but [Glibc documentation](https://www.gnu.org/software/libc/manual/html_node/Formatted-Output-Functions.html) says that `snprintf` may internally call `malloc` ("AS-Unsafe heap" [and "AS-Unsafe heap" implies "AS-Unsafe lock"](https://www.gnu.org/software/libc/manual/html_node/Unsafe-Features.html)), so it could deadlock internally when malloc is "reentered" due to signals.
* The error message could be garbled when htop signal handers are "reentered".

I was too focused on the latter and yet forget the former issue at hand.

BenBE:
> > I don't think we should start putting C++ code in htop. The C implementation you linked also has some quirks which aren't quite right (aside from missing the format attributes, it's not fully validating the format subset that it supports).
> 
> I was not implying to use C++ in htop, I still think looking at how other people solved the problem is inspirational.

Didn't interpret it as such. Just remarked that I think it would be overkill to start doing C++, just to be able to handle printing some error message right before we emergency exit our process … ;-)

> So why not write a minimal good `safe_snprintf` for our use case?

All for it. No objections. Just the requirement for correctness (i.e. implemented features should be correct, unsupported features should be properly rejected in the implementation). Needed features:

- string output
- integer output (signed/unsigned) for different sizes (int32, int64)
- hex display
- no [bf]loat support
- optional width/padding support
- signal-safe

Explorer09:
> 
> So why not write a minimal good `safe_snprintf` for our use case?

I believe there should be an existing `safe_snprintf` solution for us already, that can be embedded in htop code, or we could go for a "`itoa` + various `strncat`" solution.

However I wish to give up on this PR and let someone else who is motivated to work on this. It's not my interest in implementing a `snprintf` or `itoa` just for this, and I personally consider this issue in htop a low priority.

PR #1571, opened 2024-12-17
FlexMeter: Add FlexMeter functionality

bogdanovs:
FlexMeter provides functionality which will allow
users to make custom meters without need of rebuilding every time htop binary and adding source to the project. It can be used to print some device status, free disk space CPU or other specific temeraturer, fan RPM and many more. Everything that can be fetched from linux shell
with one line result can be printer. For fething information can be used anything from shell, python, precompiled binary or simply reading file located somewhere in file system.

New meter will appear uppon restart of htop in list with available meters.

Configuration folder location where metes should be placed:
- /home/$USER/.config/htop/FlexMeter/ On start folder will be created if does not exist, together with template file .Template in same folder.
Note: Files starting with '.' (.Template for examlpe) are ignored

Meter Example:
File name : Template

name=<NAME SHOWN IN AvailableMeters>
command=<COMMAND WHICH WILL BE EXECUTED>
type=<METER TYPE FOR NO ONLY "TEXT_METER">
caption="CAPTION TEXT SHOWN IN THE BEGGINING OF THE METER"

According to this implementation 30 Flex meter can be added Currently they have hardcoded limit of 30 meter in addition to all that already exist.

I am using this functionality for about an years maybe, so far had no issues. It might not be most optimal implementation by try to follow project stile while developed it. I am open for suggestion to improvements.

htop-flexmeter


BenBE:
Please have a very close look at your implementation again as I noticed several trivial buffer overflows in the file iteration/handling code. Furthermore I'd like to point you to our styleguide which gives additional guidance on how the code should be set up.

Also when integrating this meter, we have to take care of privilege escalations when running the specified commands. This is in particular true when running htop as root via sudo, when the home directory is still set to the logged-in user's HOME directory. In that situation a command in `/home/user/.config/htop/FlexMeter/malicious` would now potentially be run as root, instead of the user this meter belongs to. A minimum safety check would be limiting meter execution to files that belong to the current process's EUID IFF that config file is also `chmod 644` or less. Also note, that htop tries to drop capabilities: With the FlexMeters I'd suggest to enforce this.

bogdanovs:
@BenBE I looking in codding stile which I might not follow strictly , and thanks for remark on overflow issue I am aware of it but totally forget to fix it 

@Explorer09 FlexMeter was chosen because you can select name of the Meter created from configuration file, I thought it was good. I was looking for easy and simple way to extend my htop with some stats, which would require development of bunch of specific meters for simple one line shell for example. 

Regarding PCPDynamicMeter - I was looking for something simpler. This was my idea. I was looking to report some custom stuff from my system like peripherals battery status or UPS work temperature.

> pcp-htop is a version of htop built using the Performance Co-Pilot (PCP) Metrics API (see [PCPIntro(1)]
> https://man.archlinux.org/man/PCPIntro.1.en), [PMAPI(3)](https://man.archlinux.org/man/PMAPI.3.en)),
> allowing to extend htop to display values from arbitrary metrics. See the section below titled CONFIG FILES for further details.

I will fix security issue too as far it is possible.

BenBE:
> @BenBE I looking in codding stile which I might not follow strictly , and thanks for remark on overflow issue I am aware of it but totally forget to fix it

The buffer overflow was just one thing in the implementation. What initially tipped me off was the extensive use of static buffers all over the place. `htop` has `xMalloc` and `xCalloc` available for strings and these should also be used. Also when using the standard C library string functions: If there is an error-checked version in `XUtils.h` that one should be used (or if missing created). Also, if you ensure, all invalid/unset pointers are NULL (by using `xCalloc` when requesting memory and assigning NULL after `free`), you can simply call `free` (we follow the C99 standard, which defines `free(NULL);` to be a no-op); checks for `!=NULL` before accessing should be placed though (unless marked by a function attribute).

Overall, `memset`/`memmove` should appear sparingly and memory buffers on the stack should be a rarity (even no VLA at all, anywhere). Usually instead of `memcpy` you usually want something like `xStrdup` or `xSnprintf`/`xAsprintf` instead.

> @Explorer09 FlexMeter was chosen because you can select name of the Meter created from configuration file, I thought it was good. I was looking for easy and simple way to extend my htop with some stats, which would require development of bunch of specific meters for simple one line shell for example.

I think naming-wise I'm fine with both: FlexMeter or DynamicMeter. Depending on how much infrastructure can be shared with the PCP implementation, calling it DynamicMeter might be an option; but that might be a source of confusion.

> Regarding PCPDynamicMeter - I was looking for something simpler. This was my idea. I was looking to report some custom stuff from my system like peripherals battery status or UPS work temperature.

AFAICS the current implementation only implements text mode? Maybe we should limit it to that too; thinking re #1387  …

> > pcp-htop is a version of htop built using the Performance Co-Pilot (PCP) Metrics API (see [PCPIntro(1)]
> > https://man.archlinux.org/man/PCPIntro.1.en), [PMAPI(3)](https://man.archlinux.org/man/PMAPI.3.en)),
> > allowing to extend htop to display values from arbitrary metrics. See the section below titled CONFIG FILES for further details.
> 
> I will fix security issue too as far it is possible.

Both the buffer overflows ([CWE-787](https://cwe.mitre.org/data/definitions/787.html), [CWE-121](https://cwe.mitre.org/data/definitions/121.html), and [CWE-122](https://cwe.mitre.org/data/definitions/122.html)) and the privilege escalation ([CWE-250](https://cwe.mitre.org/data/definitions/250.html), [CWE-265](https://cwe.mitre.org/data/definitions/265.html), [CWE-269](https://cwe.mitre.org/data/definitions/269.html), and [CWE-270](https://cwe.mitre.org/data/definitions/270.html)) are all security issues; the privilege escalation is just the more obvious architectural one, which needs some more thorough thoughts to avoid.

A good rule of thumb is to assume that every bug your code has will wipe your system. Now write your code like (if) you value your data …

bogdanovs:
I have been working to make FlexMeter more compliant with all comments above. Used xStrdup in place of previous implementation. I followed `docs/styleguide.md` with next iteration will clean some more leftover things.

If we agree lest stick with FlexMeter, since AFAIK it does not share functionality and implementation with PCP and might be confusing.

> AFAICS the current implementation only implements text mode? Maybe we should limit it to that too; thinking re https://github.com/htop-dev/htop/pull/1387 …

Yes it implements only text mode at the moment. I kept type in implementation since I thought it might be cool to implements and others. So far it look like it is not needed.
```
type=TEXT_METERMODE
```
>Both the buffer overflows ([CWE-787](https://cwe.mitre.org/data/definitions/787.html), [CWE-121](https://cwe.mitre.org/data/definitions/121.html), and [CWE-122](https://cwe.mitre.org/data/definitions/122.html)) and the privilege escalation ([CWE-250](https://cwe.mitre.org/data/definitions/250.html), [CWE-265](https://cwe.mitre.org/data/definitions/265.html), [CWE-266](https://cwe.mitre.org/data/definitions/266.html),, [CWE-269](https://cwe.mitre.org/data/definitions/269.html), [CWE-270](https://cwe.mitre.org/data/definitions/270.html), and [CWE-273](https://cwe.mitre.org/data/definitions/273.html),) are all security issues; the privilege escalation is just the more obvious architectural one, which needs some more thorough thoughts to mitigate.

Most of this are addressed. I need to handle privileges for files located in  `~/.config/htop/FlexMeter`

FlexMeter.c
```diff
@@ -0,0 +1,226 @@
...
+#include "FlexMeter.h"
+#include <time.h>
+#include <sys/time.h>
+#include <dirent.h>
+#include "CRT.h"
+#include "Object.h"
+
+#include <unistd.h>
+#include <sys/types.h>
+#include <pwd.h>
```
Explorer09, line 18: Header order please.

FlexMeter.c
```diff
@@ -0,0 +1,226 @@
...
+{
+   for (int i = strlen(line);i>0;i--)
+   {
+      if ((line[i]=='\r') || (line[i]=='\n'))
+      {
+         line[i]='\0';
+      }
+   }
+
+   if (!strncmp(line,"name=",5))
```
Explorer09, line 53: Please use `String_startsWith` when possible. It saves your time manually updating the string length.

By the way, if we are forbidding whitespace between `name` and equal sign, better state that explicitly in the documentation.

FlexMeter.c
```diff
@@ -0,0 +1,226 @@
...
+   char path[400]; // full path
+   DIR *d;
+   struct dirent *dir;
+   struct passwd *pw = getpwuid(getuid());
+   const char *homedir = pw->pw_dir;
+
+   // path to home folder 1 for zero 1 for slash
+   char * home = (char * ) xCalloc(1,(strlen(homedir) + strlen(FLEX_CFG_FOLDER) + 2));
+
+   sprintf(home,"%s/%s",homedir,FLEX_CFG_FOLDER);
```
Explorer09, line 104: Use `xSnprintf`. `sprintf` is insecure.

FlexMeter.c
```diff
@@ -0,0 +1,226 @@
...
+         memset(buff,0,256);
+         if ( meter_list[i].command[0] != 0 )
+         {
+            FILE* fd = popen(meter_list[i].command, "r");
+            if (fd)
+            {
+               if (fgets(buff, 256, fd) == NULL)
+                  ret = pclose(fd);
+               else
+                  ret = pclose(fd);
```
Explorer09, line 177: Redundant conditional. Why?

FlexMeter.c
```diff
@@ -0,0 +1,226 @@
...
+            }
+
+            if( (buff[0] != 0) && (ret == 0) )
+            {
+               int l = strlen(buff);
```
Explorer09, line 182: Please use `size_t` for `strlen` return value, or else there will be unnecessary downcast (machine instructions).

FlexMeter.c
```diff
@@ -0,0 +1,226 @@
...
+
+   // path to home folder 1 for zero 1 for slash
+   char * home = (char * ) xCalloc(1,(strlen(homedir) + strlen(FLEX_CFG_FOLDER) + 2));
+
+   sprintf(home,"%s/%s",homedir,FLEX_CFG_FOLDER);
+
+   d = opendir(home);
+   if (d) {
+      while ((dir = readdir(d)) != NULL) {
+         if ( dir->d_name[0]!='.' && strcmp(".",dir->d_name) && strcmp("..",dir->d_name))
```
Explorer09, line 109: Maybe there is a logic error here? You likely mean this:

```suggestion
         if (dir->d_name[0] != '.' || (!String_eq(".", dir->d_name) && !String_eq("..", dir->d_name)))
```

FlexMeter.c
```diff
@@ -0,0 +1,226 @@
...
+   }
+
+   return 0;
+}
+
+static int load_config(char * file)
+{
+   int ret=0;
+   char buff[2048];
+   memset(buff,0,2048);
```
Explorer09, line 81:
```suggestion
   char buff[2048] = {0};
```

FlexMeter.c
```diff
@@ -0,0 +1,226 @@
...
+   FILE *fp = fopen( file , "r");
+   while( NULL != fgets(buff, 2048, fp) )
```
Explorer09, line 83: You will get an undefined behavior for `fgets` if `fp` is NULL.

FlexMeter.c
```diff
@@ -0,0 +1,226 @@
...
+
+static int parse_input(char * line)
+{
+   for (int i = strlen(line);i>0;i--)
+   {
+      if ((line[i]=='\r') || (line[i]=='\n'))
+      {
+         line[i]='\0';
+      }
+   }
```
Explorer09, line 51: We have a `String_readLine` function. Consider using that.

FlexMeter.c
```diff
@@ -0,0 +1,226 @@
...
+            {
+               break; // go out we reach the limit
+            }
+         }
+      }
+      closedir(d);
+   }
+   else
+   {
+      if (0 == mkdir(home,0777))
```
Explorer09, line 133: I would not permit someone create a directory with world-writable permission. Definitely bad for security. 
```suggestion
      if (0 == mkdir(home,0775))
```

FlexMeter.c
```diff
@@ -0,0 +1,226 @@
...
+         char  template_filename[500];
+         sprintf(template_filename,"%s/%s",home,".Template");
+         FILE * fp = fopen(template_filename,"w");
+         if (fp != NULL )
+         {
+            for (int i=0; i<4; i++)
+            {
+               fprintf(fp,"%s\n",template[i]);
+            }
+         }
```
Explorer09, line 152: I suggest not to create any sample file here. When the directory doesn't exist, it is likely that user does not want any flex meter at all. Besides, creating a file that runs an external command is always risky. You can't tell whether the `uptime` program from user's device is always a safe one. If that had been compromised already, you would aid an attack.

FlexMeter.c
```diff
@@ -0,0 +1,226 @@
...
+      }
+   }
+
+   free(home);
+   return meters_count;
+}
+
+static void FlexMeter_updateValues(Meter* this)
+{
+   for (int i =0 ; i < meters_count; i++)
```
Explorer09, line 162: 1. Use `size_t` for iterator when possible.
2. Opening brace on the same line of `for` keyword, please. (Our style is closer to K&R rather than Allman here.)

Explorer09:
Before I review more about the code style and quality of the meter, I have some general comments:

1. **How do we indicate the user has permissions and intent to execute the commands** in the FlexMeter template?

Honestly, I don't like the current design of the template file format. It mixes static data and executable code, and that's a bad idea. Also, it does not allow me to indicate the shell to use (i.e. it contains no shebang). I know you are using `popen(3)` to launch commands in a shell, but I expect also that the shell to use be customizable.

In other words, I wish the FlexMeter template files be shell scripts themselves, with appropriate shebang and execute permission (mode) bits. This would help the security in the long run. (The last thing I wish to see is a systemd-style, INI-like format for managing executable sub-processes. I was referring to [this "comparison of init systems" page](https://wiki.gentoo.org/wiki/Comparison_of_init_systems).)

2. **What to do if the command produces no output**, or hangs?

AFAIK, the `fgets(3)` function can block if the file descriptor wasn't opened with [O_NONBLOCK](https://www.gnu.org/software/libc/manual/html_node/Operating-Modes.html) flag. If `fgets` blocks, then htop would look frozen, which is bad. Also I believe there should be a timeout of some sort for htop waiting for child process to output. We should probably handle both situations, where (a.) the child process outputs nothing or needs too much time to output (so that htop can skip it and go to the next one), and (b.) the child process outputs in a faster rate than htop can consume. In the latter situation I expect htop only displays the _last_ line, rather than the first line waiting in the pipe, so that the data can be closer to real time when possible.

Explorer09:
> In other words, I wish the FlexMeter template files be shell scripts themselves, with appropriate shebang and execute permission (mode) bits. This would help the security in the long run.

P.S. I have a rough logic of determining whether the user can or has intent to run the commands. But it needs execute permission on the template files themselves:

```text
If the file's owner is the same as EUID of the running htop process:
   If the file does NOT have S_IXUSR permission:
      Reject. (This trumps other execute permission bits.)
   If the file has S_IXUSR permission:
      Accept.
Else, if the file belongs to a group of which the EUID is a member:
   If the file does NOT have S_IXGRP permission:
      Reject.
   If the file has S_IXGRP permission:
      Accept.
Else, if the file's owner is UID 0 (that is, root), and the file has S_IXOTH permission:
      Accept.
Else
      Reject.
```

I'm not sure if it's worth it to implements a "drop privilege" mechanism or switch back to UID on this. But if there is a way to drop privilege, then the logic of checking whether the commands should run with the UID would be same as when checking with the EUID.

FlexMeter.c
bogdanovs: > Please use `String_startsWith` when possible. It saves your time manually updating the string length.

Replacing it with String_startsWith is causing - `Segmentation fault`.

> By the way, if we are forbidding whitespace between `name` and equal sign, better state that explicitly in the documentation.

Is man page enough ?

Explorer09: > Is man page enough ?

As I have comments about the current format of the template file, I think we can think of updating the man page later. Yes, the format is supposed to be documented, but we can think of documentation later, until the code is stable enough.

FlexMeter.c
Explorer09: Oops. I missed your commit message that you intended to skip all filenames beginning with a dot `"."`. In this case, you can save your checking against `"."` and `".."` directories.

Anyway. You should note a reason why you ignore files with a leading dot, in the code comments. I can guess good reasons, but it's you who can tell which reason is the right one.

```suggestion
         // Ignore filenames beginning with a dot, so as not to
         // touch metadata like ".gitignore" or ".DS_Store".
         // This also ignores "." and ".." entries.
         if (dir->d_name[0] == '.') {
            continue;
         }
```

FlexMeter.c
Explorer09: I forgot to mention this: htop complies with [XDG Base Directory specification](https://specifications.freedesktop.org/basedir-spec/latest/) and that means we fetch user's config directory with the variable `XDG_CONFIG_HOME`. Specifically the flex meter directories should go in `$XDG_CONFIG_HOME/htop/FlexMeter`. Don't hard-code the `.config` directory part and don't use `$HOME/.config/htop/FlexMeter`. Look in the [Settings.c](https://github.com/htop-dev/htop/blob/main/Settings.c) file in htop source code to see how we fetch the `XDG_CONFIG_HOME` variable and use the fallback default if it's unset.

FlexMeter.c
bogdanovs: It was more for convenience this part, I can drop it and just document it or even place .Template file in `doc` folder of the project

FlexMeter.c
bogdanovs: I will check that, thanks

FlexMeter.c
Explorer09: @bogdanovs I'm okay with sample files. But just don't create them in the user's folders without their consent. They can read the manual and make one if they want. And yes, go ahead and put a sample in the `doc` folder of this source repo.

FlexMeter.c
Explorer09: Update: XDG recommends that folders be created with mode `0700`. Thus updating the suggestion:

```suggestion
      // Directory mode 0700 is recommended by XDG
      if (0 == mkdir(home, S_IRWXU))
```

FlexMeter.c
BenBE: Also, the style guide strongly suggests to use braces around code blocks here …

FlexMeter.c
BenBE: Why not use `const char* template[] = {…, NULL}` for the definition?

FlexMeter.c
BenBE: Also note the placement of blanks/spaces according to the styleguide.

BenBE:
@Explorer09 What do you think about handling group and user sticky bits on the files? How to handle symlinks?

Explorer09:
> @Explorer09 What do you think about handling group and user sticky bits on the files? How to handle symlinks?

* Sticky bit for executables are largely irrelevant for our case. (Linux ignores it currently, and historically it controlled whether the executable's text segments should be unloaded from swap space after exit - nothing to do with permissions here.)
* SUID and SGID - I would suggest rejecting files with SUID or SGID (or both) for now. I can't think of a good use case of starting a meter, executing a command/script, with a privilege of someone else. Until there is a good use case for that, we can change that position.

Note that my proposed logic only checks for execute permission of the _meter (template) files themselves_, and not on the executability of the _commands_. My imagined case is when the meter command was `rm -rf /`. `rm(1)` is world-executable or course, but it's the "command and arguments" combination that makes it dangerous to be run as root. So I go for checking the permissions on the meter files instead.

* Symlinks - I don't think there can be a problem on this. Just follow the link and check for execute permission of the target. (We did this with `htoprc` already.)

BenBE:
@Explorer09 Thanks for linking that old issue.

@bogdanovs What do you think of the ideas/protocol mentioned in that issue #526?

bogdanovs:
> @bogdanovs What do you think of the ideas/protocol mentioned in that issue #526?

This issue is similar to my proposal but more complex to be used in first glance. In next iteration of FlexMeter could be implemented different than TEXT_MODE, but this will require scripting on background to provide more specific data. This is not bad but more demanding from user.

FlexMeter is some how simple , not so tight to anything else. You can show whatever data you want. Issue described is more restrictive and might be harder for most average used to jump in directly.

bogdanovs:
@BenBE @Explorer09
1. Regarding the permissions of the Meters - I think we can stick with permissions which used for SSH
2. For executable scripts instead of configuration files: this file could be changed and something malicious could be injected in it. Currently FlexMeter - just load the command and while htop is running we dont care for the templates. I personally dont like the idea. We could advice user for more complex operation to write script which will be called.

FlexMeter.c
bogdanovs: I cant recall the exact reason to have all 3 checks. Thanks, I will leave only "." check

FlexMeter.c
bogdanovs: I will not create .Template from htop. I will put it sample in doc/FlexMeter and document in man page. This should be enough

FlexMeter.c
bogdanovs: I will start applying `astyle -xb -s3 -p -xg -c -k1 -W1 -H FlexMeter.c` after every change this should be fine, right ?

FlexMeter.c
bogdanovs: Perfect I am using it now, solve new line issues :) Thanks for pointing it !

FlexMeter.c
bogdanovs: Done

Explorer09:
> 1. Regarding the permissions of the Meters - I think we can stick with permissions which used for SSH

I didn't get this. Would you explain?

> 2. For executable scripts instead of configuration files: this file could be changed and something malicious could be injected in it. Currently FlexMeter - just load the command and while htop is running we dont care for the templates. I personally dont like the idea. We could advice user for more complex operation to write script which will be called.

My vision is that this file would work both as a configuration and executable script. The file format would be a [polyglot](https://en.wikipedia.org/wiki/Polyglot_(computing)). When htop starts, it loads the whole file into memory (there's a good reason for this - if the file is changed in the middle we don't want the htop instance to reflect the change until it restarts), parsing the first few lines for the meter information. It will not execute the command. When the users adds the meter, htop runs the command and grab its standard output.

This is the format of my vision:

```shell
#!/bin/sh
#htop_meter_name=Meter name
#htop_meter_modes=text
#htop_meter_caption=Caption

# Commands follow...
uptime
```

bogdanovs:
> > 1. Regarding the permissions of the Meters - I think we can stick with permissions which used for SSH
> 
> I didn't get this. Would you explain?

Permissions for .ssh folder and keys
```
chmod 700 ~/.ssh/ # same for FlexMeter folder like you suggested
chmod 700 ~/.config/htop/FlexMeter

chmod 600 ~/.ssh/authorized_keys # same for FlexMeter meter templates
chmod 600 ~/.config/htop/FlexMeter/SampleMeter

```

> > 2. For executable scripts instead of configuration files: this file could be changed and something malicious could be injected in it. Currently FlexMeter - just load the command and while htop is running we dont care for the templates. I personally dont like the idea. We could advice user for more complex operation to write script which will be called.
> 
> My vision is that this file would work both as a configuration and executable script. The file format would be a [polyglot](https://en.wikipedia.org/wiki/Polyglot_(computing)). When htop starts, it loads the whole file into memory (there's a good reason for this - if the file is changed in the middle we don't want the htop instance to reflect the change until it restarts), parsing the first few lines for the meter information. It will not execute the command. When the users adds the meter, htop runs the command and grab its standard output.
> 
> This is the format of my vision:
> 
> ```shell
> #!/bin/sh
> #htop_meter_name=Meter name
> #htop_meter_modes=text
> #htop_meter_caption=Caption
> 
> # Commands follow...
> uptime
> ```

That is ok like format, but still can get the point of having `#!/bin/bash` and file have execute permissions if this it is not supposed to be used like script and executed and we load it to htop instance like now and not touch it again until next re-spawn. Maybe I am missing some point..

Explorer09:
> Permissions for .ssh folder and keys
> 
> ```
> chmod 700 ~/.ssh/ # same for FlexMeter folder like you suggested
> chmod 700 ~/.config/htop/FlexMeter
> 
> chmod 600 ~/.ssh/authorized_keys # same for FlexMeter meter templates
> chmod 600 ~/.config/htop/FlexMeter/SampleMeter
> ```
> 

Imagine you are root and you accidentally put a `rm -ri /` command in the SampleMeter file. The SampleMeter has mode 600 but htop is happy to run the command and destroy your computer...

bogdanovs:
> > Permissions for .ssh folder and keys
> > ```
> > chmod 700 ~/.ssh/ # same for FlexMeter folder like you suggested
> > chmod 700 ~/.config/htop/FlexMeter
> > 
> > chmod 600 ~/.ssh/authorized_keys # same for FlexMeter meter templates
> > chmod 600 ~/.config/htop/FlexMeter/SampleMeter
> > ```
> 
> Imagine you are root and you accidentally put a `rm -ri /` command in the SampleMeter file. The SampleMeter has mode 600 but htop is happy to run the command and destroy your computer...

You mentioned that earlier, but this is always possibility if you are running like root. In case we use XDG_CONFIG_HOME, if run like root htop will look for meters in /root/.config/htop/FlexMeter.

You think this is too open, I am confused?

We could try to not allow execution or rm with simple regex while loading meters?

Explorer09:
> 
> You mentioned that earlier, but this is always possibility if you are running like root. In case we use XDG_CONFIG_HOME, if run like root htop will look for meters in /root/.config/htop/FlexMeter.

XDG_CONFIG_HOME can point to any directory. And we cannot assume the directory is always safe.
 
> We could try to not allow execution or rm with simple regex while loading meters?

Any program that can call unlink(2) can come with a risk of ruining your computer. Besides, there can be `alias uptime=rm` that would get around your regex check.

bogdanovs:
@Explorer09 all that you stated is true, I agree.

But If we think in that direction even htop binary could be security risk. Every single program running on the system could be malicious and could harm your system. Every systemd service or other daemon or even `ls` or `cd` could be linked to `rm` or other harmful software.

Do you have an idea how this risk could be mitigated?

I understand you concerns so we will not provide automatically generated sample on the system.
Lets allow users to write their own Meters, if user system in compromised or hacked this another topic.

Explorer09:
> But If we think in that direction even htop binary could be security risk. Every single program running on the system could be malicious and could harm your system. Every systemd service or other daemon or even `ls` or `cd` could be linked to `rm` or other harmful software.
> 
> Do you have an idea how this risk could be mitigated?

The Unix practice is to avoid running things as root whenever possible. Unix gives so much power to the root user, and it assumes that root is competent and responsible.

How responsible can a root account be for machines that aren't ours? Well, we can't tell. But what we can do is to avoid _accidentally_ running things that a user (root or not) would not intend to run. And there is one feature designed for this purpose: The **execute permission bit**.

That was why I suggest flex meter files should have `chmod 700` and not `chmod 600`. If a user is not comfortable turning on the execute bit and run that meter file like an ordinary command, then the meter shouldn't be run by htop either. That's the idea.

I won't prevent a root user from running `rm` in htop, but if they want to do so, they have to `chmod u+x` to tell they mean it. Basic defense for preventing accidentally running a program.

bogdanovs:
> The Unix practice is to avoid running things as root whenever possible. Unix gives so much power to the root user, and it assumes that root is competent and responsible.
> 
> How responsible can a root account be for machines that aren't ours? Well, we can't tell. But what we can do is to avoid _accidentally_ running things that a user (root or not) would not intend to run. And there is one feature designed for this purpose: The **execute permission bit**.
> 
> That was why I suggest flex meter files should have `chmod 700` and not `chmod 600`. If a user is not comfortable turning on the execute bit and run that meter file like an ordinary command, then the meter shouldn't be run by htop either. That's the idea.
> 
> I won't prevent a root user from running `rm` in htop, but if they want to do so, they have to `chmod u+x` to tell they mean it. Basic defense for preventing accidentally running a program.

Now is clean what you mean with 700. `chmod u+x` should act like sign that Meter is kind of "safe" or atleast user is aware of what meter is doing, kind of consent. In case mode is just 600 we will not list this Meter at all.

Meter configuration file will not be executed and it is not meant to be executed like standalone script, just loaded from htop at start.

Now we are on the same page, I will implement it that way.

Explorer09:
> Now is clean what you mean with 700. `chmod u+x` should act like sign that Meter is kind of "safe" or atleast user is aware of what meter is doing, kind of consent. In case mode is just 600 we will not list this Meter at all.

Correct. If the mode is 600, the user can still read and edit the file but would work as a sign that "this meter wasn't ready / it's a draft".

> Meter configuration file will not be executed and it is not meant to be executed like standalone script, just loaded from htop at start.

I _could_ make it work like a standalone script and use [watch(1)](https://man7.org/linux/man-pages/man1/watch.1.html) command to monitor the same thing. `watch -n 1 MyFlexMeter` Does this example make sense? Same script, but monitor without htop.

bogdanovs:
> I _could_ make it work like a standalone script and use [watch(1)](https://man7.org/linux/man-pages/man1/watch.1.html) command to monitor the same thing. `watch -n 1 MyFlexMeter` Does this example make sense? Same script, but monitor without htop.

Yes it make sens to use `watch` but not very useful tbh, just to test MyFlexMeter maybe is fine.

But I still think `#!/bin/sh` is not needed, having script with which is not supposed to be executed execpt for testing is kind of confusing without further explanation.

Explorer09:
> But I still think `#!/bin/sh` is not needed, having script with which is not supposed to be executed execpt for testing is kind of confusing without further explanation.

Your [popen(3)](https://linux.die.net/man/3/popen) call will launch the command under the `/bin/sh` shell anyway. I just make that fact explicit in the format.
I won't say we need to support other shebangs (Perl or Python?) but I think users would be able to workaround them quite easily. (Such as `python -c my_command`)

bogdanovs:
> Your [popen(3)](https://linux.die.net/man/3/popen) call will launch the command under the `/bin/sh` shell anyway. I just make that fact explicit in the format. I won't say we need to support other shebangs (Perl or Python?) but I think users would be able to workaround them quite easily. (Such as `python -c my_command`)

Correct `popen` is using `/bin/sh -c <some_commands>`. You can call whatever you want.

Well it should not affect FlexMeter configuration if there is other shebang, I can generally ignore it while loading and care only for whatever keys  I am supporting.

bogdanovs:
Summary on latest update : 
* check `~/.config/htop/FlexMeter` folder mod bits , if not 700 ignore it
* check every meter under FlexMeter directory for access bits
* Dry run every command on load. If fail for some reason fail meter is listed and when selected print '[ERR] Check command'

Explorer09:
> * Dry run every command on load. If fail for some reason fail meter is listed and when selected print '[ERR] Check command'

I would choose to run the meter command only when the meter is added. Not an earlier. Think of when the command is `echo something >some_log_file` and now the log file would be written when htop starts no matter you have the meter or not. (I can imagine some other more destructive commands here, but you should get the idea.)

I will have comments about the permission check code later.

FlexMeter.c
```diff
@@ -0,0 +1,216 @@
...
+   return ret;
+}
+
+int check_for_meters(void)
+{
+   char path[400]; // full path
+   DIR* d;
+   struct dirent* dir;
+   struct passwd* pw = getpwuid(getuid());
+   const char* homedir = pw->pw_dir;
```
BenBE, line 108: Note the comment by @Explorer09 regarding the `XDG_HOME_DIR` variable and fallbacks …

FlexMeter.c
```diff
@@ -0,0 +1,216 @@
...
+                     meters_count++;
+                  } else {
+                     break; // go out we reach the limit
+                  }
+               }
+            }
+         }
+         closedir(d);
+      } else {
+         mkdir(home, 0700);
```
BenBE, line 148: Don't create folders when opening them fails.

BenBE (changes requested):
Possibly split the unrelated changes into a separate commit.

Also, the types within `FlexMeter.c` don't follow the usual naming convention?

Any reason for the hard limit on supported flex meters?

FlexMeter.c
```diff
@@ -0,0 +1,216 @@
...
+      int ret = -1;
+      memset(buff, 0, 256);
+      FILE* fd = popen(meter_list[meters_count].command, "r");
+      if (fd) {
+         if (fgets(buff, 256, fd) == NULL) {
+            ret = pclose(fd);
+         } else {
+            ret = pclose(fd);
+         }
+      }
```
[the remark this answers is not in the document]
Explorer09: I don't even consider "dry" running the meter to check whether the command works is a good idea. It's not technically dry at all when the command(s) can modify something. I would choose to run the commands only when the meter is actually added. If the command would fail, then show the message that it fails.

FlexMeter.c
```diff
@@ -0,0 +1,216 @@
...
+   } else if (!strncmp(line, "type=", 5)) {
+      meter_list[meters_count].type = xStrdup(line + 6);
+   } else {
+      return -1;
+   }
+
+   return 0;
+}
+
+static int load_config(char* file)
```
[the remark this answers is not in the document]
Explorer09: Some coding styles do use `int` to return a status of an operation (e.g. Linux kernel). It might be simply that they don't understand htop uses `bool` for operation statuses and I would personally tolerate that.

But, if there is no need to indicate statuses other than success and failure, then do stick with existing code and use `bool`.

FlexMeter.c
Explorer09: I am neutral on whether to create the folder anyway. If creating an empty folder would result in extra code then I would suggest to not bother with creating the folder.

FlexMeter.c
```diff
@@ -0,0 +1,216 @@
...
+   struct stat fileStat;
+
+   if (stat(home, &fileStat) < 0) {
+      return -1;
+   }
+
+   if (S_ISDIR(fileStat.st_mode) &&
+         (fileStat.st_mode & S_IRUSR) &&
+         (fileStat.st_mode & S_IWUSR) &&
+         (fileStat.st_mode & S_IXUSR) ) {
```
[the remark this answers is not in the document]
Explorer09: May I ask @bogdanovs about what purpose it is when you check the permission of the _folder_? I don't think there need to be write permission in order for htop to load the meters, and checking for read permission is redundant as you would loop over the files in the folder later on (which implies read permission already).

Execute permission is technically irrelevant - we won't use it as a working directory (chdir) for anything.

FlexMeter.c
```diff
@@ -0,0 +1,216 @@
...
+   .uiName = NULL,
+   .caption = NULL,
+};
+
+int load_flex_modules(void)
+{
+   uint8_t meters_num = check_for_meters();
+   if (FlexMeter_class == NULL && meters_num != 0) {
+      FlexMeter_class = (MeterClass*) xCalloc(meters_num, sizeof(MeterClass));
+      for (uint8_t i = 0 ; i < meters_num; i++) {
```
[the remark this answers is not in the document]
Explorer09: Just a reminder that `uint8_t` for iterators here won't bring any speed or size advantage. Compilers would emit code to clear upper bits (`i = ((i + 1) & 0xFF)`) to keep the values in bounds of the `uint8_t` type. It could be worse when it masks an overflow, turning a loop intended to be finite to an infinite one... You get the idea.

FlexMeter.c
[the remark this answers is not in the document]
BenBE: ACK.

Mostly noted regarding the code changes, as they affect another location later in the PR too.

FlexMeter.c
BenBE: This is extra code, that is not needed when the folder exists. And if opening the folder fails this might be for other reasons than the folder being amiss. Thus the `mkdir` would need to be error-checked, which is code, we really don't need for the feature to work.

FlexMeter.c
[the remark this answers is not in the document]
bogdanovs: We commented that  FlexMeter folder and Meters should have 0700 permissions.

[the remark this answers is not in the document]
BenBE: Yes, but that check currently only covers the folder, not the individual files …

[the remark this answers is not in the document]
Explorer09: I really have no idea why you (bogdanovs) check the folder. Even you confirmed it has `u+rwx` permissions, it might be not owned by the same user as the home directory's owner (e.g. `/home/alice/.config/htop/FlexMeter` owned by `bob`, mode `777`), making your permission check moot.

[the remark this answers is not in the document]
bogdanovs: Yes , actually this is right. I will add check and for user ownership. Thanks !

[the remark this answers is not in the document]
Explorer09: @bogdanovs I'm still confused about why you check the _folder_ ownership and permissions. That seems actually redundant considering we are going to check for _file_ permissions already (and perhaps also ownership, for symlinks).

Besides, what's wrong when the folder is owned by someone else?

Like, when Alice or someone trusts a `FlexMeter` directory owned by root, and she makes a symlink to a system-wide `FlexMeter` directory like this:

```shell
cd ~alice/.config/htop
ln -s /etc/htop/FlexMeter FlexMeter
# FlexMeter -> /etc/htop/FlexMeter symlink owned by alice
# drwxr-xr-x 2 root root <size> <date> /etc/htop/FlexMeter
```

FlexMeter.c
```diff
@@ -0,0 +1,216 @@
...
+         memcpy(&FlexMeter_class[i], &FlexMeter_class_template, sizeof(MeterClass));
+         FlexMeter_class[i].name = (const char*) xStrdup(meter_list[i].name);
+         FlexMeter_class[i].uiName = (const char*) xStrdup(meter_list[i].uiName);
+         FlexMeter_class[i].caption = (const char*) xStrdup(meter_list[i].caption);
```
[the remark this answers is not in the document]
bogdanovs: @BenBE can you please elaborate ?

FlexMeter.c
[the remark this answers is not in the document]
bogdanovs: I personally prefer `int` tbh

FlexMeter.c
```diff
@@ -0,0 +1,216 @@
...
+
+typedef struct {
+   char* name;
+   char* command;
+   char* type;
+   char* caption;
+   char* uiName;
+} _flex_meter;
+
+_flex_meter meter_list[30];
```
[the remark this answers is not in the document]
bogdanovs: yes technical mistake , fixed now

bogdanovs:
@BenBE Is it mandatory to user `bool` instead of `int`. If not I would prefer to keep function with `int`

FlexMeter.c
bogdanovs: this part is stripped now , if user need it, will create it

BenBE:
> @BenBE Is it mandatory to user `bool` instead of `int`. If not I would prefer to keep function with `int`

If semantics are pure truth values: yes. If these functions provide full error reasons as their return value, `int` is fine.

If using `int`, use `<0` (e.g. `-errno`) for error, and a sensible meaning for the success case.

FlexMeter.c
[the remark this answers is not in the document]
bogdanovs: dry run part is removed. If meter is selected and `popen` return error on first run ERR message is printed and command is `free()` and not executed anymore

FlexMeter.c
```diff
@@ -0,0 +1,216 @@
...
+static const int DateMeter_attributes[] = {
+   FLEX
+};
+
+MeterClass* FlexMeter_class = NULL;
+
+int check_for_meters(void);
+
+static int parse_input(char* line)
+{
```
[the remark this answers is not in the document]
bogdanovs: What exactly you mean by repeated string comparison ?

FlexMeter.c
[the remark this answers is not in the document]
bogdanovs: Currently `~/.config/htop/FlexMeter` should be folder in this specific place and simlinks are not allowed.
All this make sens, but make things more open.

solaris/Platform.c
```diff
@@ -101,7 +101,9 @@ const SignalItem Platform_signals[] = {
 
 const unsigned int Platform_numberOfSignals = ARRAYSIZE(Platform_signals);
 
-const MeterClass* const Platform_meterTypes[] = {
+const ProcessField Platform_defaultFields[] = { PID, LWPID, USER, PRIORITY, NICE, M_VIRT, M_RESIDENT, STATE, PERCENT_CPU, PERCENT_MEM, TIME, COMM, 0 };
+
```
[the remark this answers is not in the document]
bogdanovs: Result of bad rebase

unsupported/Platform.c
```diff
@@ -44,7 +44,9 @@ const SignalItem Platform_signals[] = {
 
 const unsigned int Platform_numberOfSignals = ARRAYSIZE(Platform_signals);
 
-const MeterClass* const Platform_meterTypes[] = {
+const ProcessField Platform_defaultFields[] = { PID, USER, PRIORITY, NICE, M_VIRT, M_RESIDENT, STATE, PERCENT_CPU, PERCENT_MEM, TIME, COMM, 0 };
+
```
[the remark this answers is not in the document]
bogdanovs: Result of bad rebase

bogdanovs:
FYI: Just pushed latest version which is without some comments which are already under some kind of discussion.

FlexMeter.c
```diff
@@ -0,0 +1,191 @@
...
+#include <sys/time.h>
+#include <sys/types.h>
+#include <unistd.h>
+#include <dirent.h>
+#include <time.h>
+#include <pwd.h>
+#include "FlexMeter.h"
+#include "Object.h"
+#include "config.h"
+#include "CRT.h"
```
BenBE, line 17: Header order; cf. styleguide …

FlexMeter.c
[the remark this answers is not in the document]
BenBE: You have multiple `strncmp` calls that all check for `line` to start with a certain string. You can drastically reduce the number of these calls that need to be performed for each check if you switch case on the first letter, and only check for strings which start a given way. This is e.g. used on Linux when parsing certain files in the procfs to reduce the amount of work needed to find all the data from the processed files.

FlexMeter.c
```diff
@@ -0,0 +1,191 @@
...
+               /* We are ignoring all files starting with . like ".Template"
+                * and "." ".." directories
+                */
+               continue;
+            }
+
+            meter_list[meters_count].name = xStrdup(dir->d_name);
+            xAsprintf(&path, "%s/%s", home, dir->d_name);
+
+            if (access(path, R_OK | W_OK | X_OK) == 0) {
```
BenBE, line 112: Do we need write access here? RX should be fine.

FlexMeter.c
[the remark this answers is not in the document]
BenBE: Given the unrestricted nature of how the `uiName` and `caption` are set, a FlexMeter could just configure to use the same strings as another (built-in) meter and thus impersonate another (e.g. built-in) meter. This is not a security issue directly, but more a point for discussion, if we want to allow the user to create FlexMeters that are indistinguishable from built-in meters.

FlexMeter.c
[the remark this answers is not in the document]
BenBE: My preference is to avoid plain `int` and `long` as much as possible, especially if there is a proper type that carries the intended semantics (`bool`, `size_t`, `uintptr_t`, …) or provides an indication of what range of numbers (`uint32_t`) the author expected a given variable to carry.

FlexMeter.c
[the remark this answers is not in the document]
Explorer09: Good point from @BenBE about meter impersonation. But since we have PCPDynamicMeter already, what would PCPDynamicMeter handle this situation?

I think @natoscott can come in and look at this whole PR and give a review.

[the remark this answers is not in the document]
bogdanovs: I think could be overcome with prefix or suffix to the `uiName` and for the `caption` it is matter of user decision what caption should be selected. It could be literally everything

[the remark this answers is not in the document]
BenBE: TBH, I don't see this as a critical security issue, as the impact – lacking a threat model – is potentially minimal and due to lack of formatting options it's hardly realistic to confuse the user with text meters (most of them have some sort of emphasis which FlexMeters would lack). So this is mostly posted here as a note, less about sth that needs urgent fixing.

Though for the selection screen I'd still suggest to force a `Flex:` prefix like `Flex: Evil Meter Impersonation` …

Also, when storing the config, Flex meters should receive a similar prefix AND have their name filtered for invalid characters. When loading flex meters, similar filtering should occur too.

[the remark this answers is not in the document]
Explorer09: 1. Good suggestion about a meter prefix. And it could avoid name collisions with any meter added into htop built-in in the future. (BTW, did this issue ever come up with PCPDynamicMeter too?)
2. For "captions" I would suggest a laxer filter. Honestly I would like to see Unicode support for custom meter captions, because why not. The only kind of characters worth filtering out is non-printable characters (including control characters). Otherwise I don't like the things to be too strict when it comes to custom meter captions.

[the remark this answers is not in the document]
BenBE: > 1. Good suggestion about a meter prefix. And it could avoid name collisions with any meter added into htop built-in in the future. (BTW, did this issue ever come up with PCPDynamicMeter too?)

No, AFAIR not. But there is some code in `PCPDynamicColumn_validateColumnName` that does minimalistic filtering. Also there are checks to detect duplicate names.

> 2. For "captions" I would suggest a laxer filter. Honestly I would like to see Unicode support for custom meter captions, because why not. The only kind of characters worth filtering out is non-printable characters (including control characters). Otherwise I don't like the things to be too strict when it comes to custom meter captions.

ACK.

FlexMeter.c
bogdanovs: no we dont need write access , but this match what we discuss for `700`

BenBE: I think it'd be best to allow both 0500 and 0700 so the user can disable write access as that is not needed for the feature to work. The check should be for the minimum we require, with 0700 being the maximum we allow (technically even 0755 would still be fine).

Explorer09: @BenBE I think even 0400 would suffice. Or maybe just remove this check altogether as it is redundant.

Explorer09: (Sorry for the last comment about 0400 as I misread the code. I deleted that comment.)

By the way, do we have a better API than access(2) here? access(2) is vulnerable to TOCTOU and I wish to avoid it when possible.

BenBE: An option may be to use `open(2)` with `O_EXCL` and `fstat(2)` the `fd`, before passing it on to `fdopen(3)` in `load_config`, which would need to receive the `fd` (or a `FILE*`, if we did the `fdopen(3)` in the caller).

There is even the possibility to use `faccessat(2)` with `AT_EMPTY_FILE` to emulate `access(2)` using an already open `fd`, but this is Linux-specific (read: non-portable) and requires a recent (>=5.8) kernel to work.

Header.c
```diff
@@ -26,7 +26,7 @@ in the source distribution for its full text.
 #include "ProvideCurses.h"
 #include "Settings.h"
 #include "XUtils.h"
-
+#include "FlexMeter.h"
```
BenBE, line 29: Include order

FlexMeter.c
```diff
@@ -0,0 +1,216 @@
...
+   char* command;
+   char* type;
+   char* caption;
+   char* uiName;
+} _flex_meter;
+
+_flex_meter meter_list[30];
+
+int meter_list_idx = 0;
+static int meters_count = 0;
```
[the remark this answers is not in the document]
bogdanovs: agree , done

FlexMeter.c
```diff
@@ -0,0 +1,216 @@
...
+
+static const int DateMeter_attributes[] = {
+   FLEX
+};
+
+MeterClass* FlexMeter_class = NULL;
+
+int check_for_meters(void);
```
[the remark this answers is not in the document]
bogdanovs: function return meters count, I will keep it as int and make it static.

FlexMeter.c
bogdanovs: what about replacing `access(2)` with `stat(2)` like for folder ?

FlexMeter.c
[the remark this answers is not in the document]
bogdanovs: 1. `Flex:` prefix - ok
2. lexer is good but too powerful tool for that task. I think simple implementation will do the job. Valid range from 26 to 126, for caption is open enough and without non printable chars

FlexMeter.c
Explorer09: @BenBE @bogdanovs 
How about doing it in the same way that we check for `htoprc`?
`open(2)` the file with the option (`O_RDONLY | O_NOFOLLOW`); after it succeeds, `fstat(2)` the file descriptor and check its execute bit `(sb.st_mode & S_IXUSR) != 0`; after both of these are done, `fdopen(3)`.

The [`stat` structure](https://man7.org/linux/man-pages/man3/stat.3type.html) should give us enough information to check not only for modes but also owner ID and group ID. If you are able, you may try implementing [the whole logic I made in a previous comment](https://github.com/htop-dev/htop/pull/1571#issuecomment-2565338735).

FlexMeter.c
```diff
@@ -0,0 +1,203 @@
...
+            }
+         }
+         closedir(d);
+      }
+   }
+
+   if (home != NULL) {
+      free(home);
+      home = NULL;
+   }
```
Explorer09, line 143: You can assume `free(NULL)` will do nothing. This behavior is required by ANSI C standard. (Operating systems that do not have this behavior are too old for htop to support.)

FlexMeter.c
bogdanovs: > @BenBE @bogdanovs How about doing it in the same way that we check for `htoprc`? `open(2)` the file with the option (`O_RDONLY | O_NOFOLLOW`); after it succeeds, `fstat(2)` the file descriptor and check its execute bit `(sb.st_mode & S_IXUSR) != 0`; after both of these are done, `fdopen(3)`.
> 
> The [`stat` structure](https://man7.org/linux/man-pages/man3/stat.3type.html) should give us enough information to check not only for modes but also owner ID and group ID. If you are able, you may try implementing [the whole logic I made in a previous comment](https://github.com/htop-dev/htop/pull/1571#issuecomment-2565338735).

if we agree on `stat` approach I will implement it that way

BenBE: @Explorer09 What's the difference in your suggestion compared to mine? Mine was mostly focused on the actual files, but would work similar for the directory too.

FlexMeter.c
BenBE: To be precise: A libc where `free(NULL);` is anything but a no-op does not comply to the ISO C99 standard (section 7.20.3.2) and is thus outside the scope of what we support.

FlexMeter.c
Explorer09: @BenBE Probably no difference, except that the `O_EXCL` flag you mentioned won't do what you want. (`O_EXCL` is for creating the file only - it forces `open(2)` to error out if a file with the same name exists).

FlexMeter.c
bogdanovs: Just to confirm which approach is preferred here with `if (ptr != NULL)` of without ?

BenBE: Without …

bogdanovs: My fault , initially I understood that this check is required.
I will remove them

bogdanovs:
> > This is the format of my vision:
> > ```shell
> > #!/bin/sh
> > #htop_meter_name=Meter name
> > #htop_meter_modes=text
> > #htop_meter_caption=Caption
> > 
> > # Commands follow...
> > uptime
> > ```
> 
> That is ok like format, but still cant get the point of having `#!/bin/bash` and file have execute permissions if this it is not supposed to be used like script and executed and we load it to htop instance like now and not touch it again until next re-spawn. Maybe I am missing some point..

I think this format will encourage users to have more complex multi line commands which will cause issues.

I would like to keep current format:
```shell
name=Meter name
type=text
caption=Caption
command=<one line command or call script>
```

bogdanovs:
In last update
* add `XDG_CONFIG_HOME` with fallback
* Remove checks for NULL
* use `stat` instead of `access` for folder

Explorer09:
> I think this format will encourage users to have more complex multi line commands which will cause issues.

What issues? It was intentional for me to support multiline commands, because there's no reason not to. Also, how was it any different if I can squeeze multiple commands already into one line?

BenBE:
The only way I think supporting multi-line commands makes sense is if we were to run the config files as a shell script (in which case the shabang+commented-out settings make sense). But this opens up a whole other can of worms. In the end, either way will handle arbitrary scripts run with some set of privileges (TBD), where you will be able to spawn shells and sub-processes. TBH; there's no sane way to make this anywhere "safe".

But, going the "shell script route" encourages to have more complex behaviours packaged in such meters, which is IMO nothing we should encourage, even though it's possible (and there's no real way in stopping the user from doing). Thus I somewhat prefer the single-line approach as it tends to make the config files simpler and cause the nitty-gritty details out into external scripts that can be written in whatever language fits better than a basic shell script used to bootstrap things.

Explorer09:
> 
> But, going the "shell script route" encourages to have more complex behaviours packaged in such meters, which is IMO nothing we should encourage, even though it's possible (and there's no real way in stopping the user from doing). Thus I somewhat prefer the single-line approach as it tends to make the config files simpler and cause the nitty-gritty details out into external scripts that can be written in whatever language fits better than a basic shell script used to bootstrap things.

There is nothing going simpler when you limit the command to one line. The real thing that could make the script simpler is limiting the _total length_ of the commands htop could load per file. For example 4 KiB.

Also we can limit the interpreter shebang to `/bin/sh` only. Any other language will require launching the interpreter indirectly.

The point of putting multiple commands in a meter file is that htop can keep the commands _in memory_ so that any modification to the meter file after htop started would not affect that htop instance of starting commands. If the commands are launched through another script (be it shell script or Perl or Python), then they lose this advantage.

Explorer09:
Additional comments...

> The only way I think supporting multi-line commands makes sense is if we were to run the config files as a shell script (in which case the shabang+commented-out settings make sense).

The reality is that we _are_ running it as a shell script due to how [popen(3)](https://man7.org/linux/man-pages/man3/popen.3.html) command works.

If we have been forking and then using [exec(3)](https://man7.org/linux/man-pages/man3/exec.3.html) family of functions, without `p` in function name, then we can model the meter file format as a simpler, one-command directive file.

As I mentioned in [a comment before](https://github.com/htop-dev/htop/pull/1571#issuecomment-2565317586), the last thing I wished to see is an executable script file modeled like an INI that looked harmless in a glance. (Many INI-like files in Microsoft Windows can contain executable code, mind you, the most notable of which are driver installation `.inf` files. I've worked with such files before and thus I know them.)

> In the end, either way will handle arbitrary scripts run with some set of privileges (TBD), where you will be able to spawn shells and sub-processes. TBH; there's no sane way to make this anywhere "safe".

Acknowledged. Although my vision is to makes things simple and utilize Unix mechanism of user privileges whenever possible.

That is to say, (1) don't be afraid of running arbitrary scripts with elevated privileges if there are enough permission bits indicating the users have intent to do so, and (2) we can restrict the execution of scripts that are owned by someone else, except when the owner is `root`, who we can assume trust with.

By the way, I've changed some mind about the SUID bit on meter files. I think we can allow them. Even though that could result in an elevation of privilege, ultimately it's not the fault of us or the user who execute it, it is the file owner who is responsible.

BenBE:
> Additional comments...
> 
> > The only way I think supporting multi-line commands makes sense is if we were to run the config files as a shell script (in which case the shabang+commented-out settings make sense).
> 
> The reality is that we _are_ running it as a shell script due to how [popen(3)](https://man7.org/linux/man-pages/man3/popen.3.html) command works.
> 
> If we have been forking and then using [exec(3)](https://man7.org/linux/man-pages/man3/exec.3.html) family of functions, without `p` in function name, then we can model the meter file format as a simpler, one-command directive file.
> 
> As I mentioned in [a comment before](https://github.com/htop-dev/htop/pull/1571#issuecomment-2565317586), the last thing I wished to see is an executable script file modeled like an INI that looked harmless in a glance. (Many INI-like files in Microsoft Windows can contain executable code, mind you, the most notable of which are driver installation `.inf` files. I've worked with such files before and thus I know them.)
> 

I'm afraid of neither and the point for discussion basically boils down to whether we want soem INI like format which emphasizes/prioritizes single-line commands or if we would prefer to go the route of shell scripts, with the meta-information placed in a header at the top of the file.

I'm fine with either way, but somewhat prefer the INI format as it makes parsing out the information we need to know for the meter much simpler. In the end, you can't run code without somewhere placing the code you want to run.

> > In the end, either way will handle arbitrary scripts run with some set of privileges (TBD), where you will be able to spawn shells and sub-processes. TBH; there's no sane way to make this anywhere "safe".
> 
> Acknowledged. Although my vision is to makes things simple and utilize Unix mechanism of user privileges whenever possible.

No objection here.

> That is to say, (1) don't be afraid of running arbitrary scripts with elevated privileges if there are enough permission bits indicating the users have intent to do so,

I just tend to be very careful whenever there is some "arbitrary code execution" on a code path. Occupational hazard …

> and (2) we can restrict the execution of scripts that are owned by someone else, except when the owner is `root`, who we can assume trust with.

ACK. But coming from an IT sec perspective, I'd prefer to look at it that way: How much do we have to allow for things to work properly. Ideally, scripts would run with no more than the privileges of the owner, but also no higher privileges than the user who executes them. Which brings us to the next point:

> By the way, I've changed some mind about the SUID bit on meter files. I think we can allow them. Even though that could result in an elevation of privilege, ultimately it's not the fault of us or the user who execute it, it is the file owner who is responsible.

We could use the SUID bit in the sense it is used on regular executables: Run as the owner of that file. That way, by default meters would be executed as the user running htop, except when a meter is SUID, in which case it would inherit that user's privileges when executing the meter. This would leave us with 10 cases basically:

1. htop as root, meter by root, no SUID: runs as root (htop user)
2. htop as root, meter by non-root, no SUID: does not run (prevent privesc)
3. htop as root, meter by root, SUID: runs as root (SUID)
4. htop as root, meter by non-root, SUID: run as non-root (SUID) -> enforce priv downgrade(!!!)
5. htop as non-root, meter by root, no SUID: runs as non-root (htop user)
6. htop as non-root, meter by non-root (same), no SUID: run as non-root (htop user)
7. htop as non-root, meter by non-root (different), no SUID: does not run (prevent privesc)
8. htop as non-root, meter by root, SUID: run as root (SUID)
9. htop as non-root, meter by non-root (same), SUID: run as non-root (SUID)
10. htop as non-root, meter by non-root (different), SUID: does not run (prevent privesc)

Alternatively, you could ignore SUID if the meter was not owned by root, which gives the following cases:

1. htop as root, meter by root, no SUID: runs as root (htop user)
2. htop as root, meter by non-root, no SUID: does not run (prevent privesc)
3. htop as root, meter by root, SUID: runs as root (SUID)
4. htop as root, meter by non-root, SUID: does not run (prevent privesc)
5. htop as non-root, meter by root, no SUID: runs as non-root (htop user)
6. htop as non-root, meter by non-root (same), no SUID: run as non-root (htop user)
7. htop as non-root, meter by non-root (different), no SUID: does not run (prevent privesc)
8. htop as non-root, meter by root, SUID: run as root (SUID)
9. htop as non-root, meter by non-root (same), SUID: does not run (prevent privesc)
10. htop as non-root, meter by non-root (different), SUID: does not run (prevent privesc)

FlexMeter.c
```diff
@@ -0,0 +1,217 @@
...
+
+   return true;
+}
+
+static bool load_config(_flex_meter *meter, char* file)
+{
+   bool ret = false;
+   FILE* fp = fopen(file, "r");
+
+   if (fp != NULL) {
```
BenBE, line 76: Note, that `fclose(NULL);` is undefined behaviour, usually characterized to crash your program. ;-)

```suggestion
   if (!fp) {
      return false;
   }

```

Doing it that way around also allows to lower the indentation level.

FlexMeter.c
```diff
@@ -0,0 +1,217 @@
...
+      char* buff;
+      while ((buff = String_readLine(fp)) != NULL) {
+         ret = parse_input(meter, buff);
+         if (!ret) {
+            break;
+         }
+      }
+      free(buff);
+      buff = NULL;
```
BenBE, line 85: Not strictly necessary. should best be removed.

For stack variables, calling `free` suffices. Only when the variable or its value is exposed to other scopes it's necessary to explicitly set the value to `NULL`.

FlexMeter.c
BenBE: `getpwuid` has some bad side effects (calls to libpam and potentially causes network traffic; thus can potentially hang). Furthermore, when building htop with `--enable-static` this may cause issues, thus call sites should be minimized.

Not to mention the call can fail and is lacking error checks.

Please compare with `Settings_new` in `Settings.c`. Possibly, we should refactor this code block to be a separate function. @Explorer09?

FlexMeter.c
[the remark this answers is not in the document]
BenBE: For `caption=` all* `>=32` is fine. With just `26<=c && c<=126` you cause two issues:
1. You allow for inserting ECMA48 escape sequences (which _can_ rebind keys -> security issue)
2. You break Unicode support for things like `caption=運勢` (when executing `command=LANG=ja_JA fortune`)

*Unicode filtering has its whole other can of worms … But that would apply to the output of command too …

FlexMeter.h
```diff
@@ -0,0 +1,22 @@
...
+in the source distribution for its full text.
+*/
+#include <stdbool.h>
+
+#include "Meter.h"
+
+
+#define METERS_LIST_SIZE 30
+
+#define MAX_METERS_COUNT METERS_LIST_SIZE-1
```
BenBE, line 16: Only one of these (`METERS_LIST_SIZE`, `MAX_METERS_COUNT`) should exist.

Also, I'd suggest a slight rename:
```suggestion
#define MAX_FLEXMETERS_COUNT 30
```

Header.c
```diff
@@ -20,14 +20,14 @@ in the source distribution for its full text.
...
 #include "DynamicMeter.h"
+#include "FlexMeter.h"
 #include "Macros.h"
 #include "Object.h"
 #include "Platform.h"
 #include "ProvideCurses.h"
 #include "Settings.h"
 #include "XUtils.h"
 
-
```
BenBE, line 30: Keep the second blank after the list of includes.

Header.c
```diff
@@ -121,6 +121,14 @@ void Header_populateFromSettings(Header* this) {
    const Settings* settings = this->host->settings;
    Header_setLayout(this, settings->hLayout);
 
+   int num = load_flex_modules();
+   int platform_size = 0;
+
+   for (platform_size = 0; Platform_meterTypes[platform_size] != NULL; platform_size++);
+   for (int i = 0; i < num; i++) Platform_meterTypes[platform_size+i]=FlexMeter_class+i;
+
+   Platform_meterTypes[platform_size+num]=NULL;
```
BenBE, line 130: Space around binary operators:

```suggestion
   for (platform_size = 0; Platform_meterTypes[platform_size] != NULL; platform_size++) {
      // Count meter types for the platform
   }
   for (int i = 0; i < num; i++) {
      Platform_meterTypes[platform_size + i] = FlexMeter_class + i;
   }

   Platform_meterTypes[platform_size + num] = NULL;
```

Furthermore, `for` loops always should have a body; and be it empty.

Meter.h
```diff
@@ -67,9 +67,9 @@ typedef struct MeterClass_ {
    const MeterModeId defaultMode;
    const double total;
    const int* const attributes;
-   const char* const name;                 /* internal name of the meter, must not contain any space */
-   const char* const uiName;               /* display name in header setup menu */
-   const char* const caption;              /* prefix in the actual header */
+   const char* name;                 /* internal name of the meter, must not contain any space */
+   const char* uiName;               /* display name in header setup menu */
+   const char* caption;              /* prefix in the actual header */
```
BenBE, line 72:
```suggestion
   const char* name;                       /* internal name of the meter, must not contain any space */
   const char* uiName;                     /* display name in header setup menu */
   const char* caption;                    /* prefix in the actual header */
```

dragonflybsd/Platform.c
```diff
@@ -43,7 +44,6 @@ in the source distribution for its full text.
 #include "dragonflybsd/DragonFlyBSDProcessTable.h"
 #include "generic/fdstat_sysctl.h"
 
-
```
BenBE, line 46: Second blank line after includes.

openbsd/Platform.c
```diff
@@ -47,7 +48,6 @@ in the source distribution for its full text.
 #include "openbsd/OpenBSDMachine.h"
 #include "openbsd/OpenBSDProcess.h"
 
-
```
BenBE, line 50: Second blank line after includes.

pcp/Platform.c
```diff
@@ -55,7 +56,6 @@ in the source distribution for its full text.
 #include "zfs/ZfsArcStats.h"
 #include "zfs/ZfsCompressedArcMeter.h"
 
-
```
BenBE, line 58: Second blank line after includes.

solaris/Platform.c
```diff
@@ -43,7 +44,6 @@ in the source distribution for its full text.
 #include "zfs/ZfsArcMeter.h"
 #include "zfs/ZfsCompressedArcMeter.h"
 
-
```
BenBE, line 46: Second blank line after includes.

unsupported/Platform.c
```diff
@@ -27,7 +28,6 @@ in the source distribution for its full text.
 #include "TasksMeter.h"
 #include "UptimeMeter.h"
 
-
```
BenBE, line 30: Second blank line after includes.

Explorer09:
> I'm afraid of neither and the point for discussion basically boils down to whether we want soem INI like format which emphasizes/prioritizes single-line commands or if we would prefer to go the route of shell scripts, with the meta-information placed in a header at the top of the file.
> 
> I'm fine with either way, but somewhat prefer the INI format as it makes parsing out the information we need to know for the meter much simpler. In the end, you can't run code without somewhere placing the code you want to run.

The INI-like format is foreign to a Unix like environment when it comes to shell commands. That's the primary reason I suggest against that.

Really. You need to handle the quoting and character escape mechanisms, and perhaps also distinguishing `;` (a semicolon) as a command delimiter and `;` as a literal argument (to a command such as find(1)). Even though we are unlikely to
get a security vulnerability when misinterpreting a user command, such issue can block any future extensibility to the meter format, and that's what I wished to avoid.

```text
rm -f foo ; rm -f bar
# (Are we supposed to delete the files named ";", "rm" and "-f" ?)
```

Just for another information, I know that an INI-like format _could_ work if we are executing a program directly without a shell. The closest example we can take as reference is the [desktop entry format](https://wiki.archlinux.org/title/Desktop_entries) (`.desktop` extension, this is the Unix/Linux counterpart to the `.lnk` files in Windows). 

> 
> We could use the SUID bit in the sense it is used on regular executables: Run as the owner of that file. That way, by default meters would be executed as the user running htop, except when a meter is SUID, in which case it would inherit that user's privileges when executing the meter. This would leave us with 10 cases basically:
> 
> 1. htop as root, meter by root, no SUID: runs as root (htop user)
> 2. htop as root, meter by non-root, no SUID: does not run (prevent privesc)
> 3. htop as root, meter by root, SUID: runs as root (SUID)
> 4. htop as root, meter by non-root, SUID: run as non-root (SUID) -> enforce priv downgrade(!!!)
> 5. htop as non-root, meter by root, no SUID: runs as non-root (htop user)
> 6. htop as non-root, meter by non-root (same), no SUID: run as non-root (htop user)
> 7. htop as non-root, meter by non-root (different), no SUID: does not run (prevent privesc)
> 8. htop as non-root, meter by root, SUID: run as root (SUID)
> 9. htop as non-root, meter by non-root (same), SUID: run as non-root (SUID)
> 10. htop as non-root, meter by non-root (different), SUID: does not run (prevent privesc)

I was about to think of all possible cases before you presented this list to me. Thanks.

For case 2., this is a case I also wish to avoid, and I should mention that htop should also ignore the "others execute" bit in this situation.
Case 5 would require the non-root user to have execute permission on that meter (either "group execute" or "other execute"), but I think everyone got the idea already.
For case 7, I think we can loosen the requirement by allowing the meter to be run _if and only if_ the user and meter file belong to the same group, and it has "group execute" permission.  The "others execute" would be ignored, in particular.
Case 10 is like case 7 that we can loosen the requirement to allow the meter to run with "group execute" permission. Here I think even the "others execute" would be safe here, since the privilege is limited to whoever owns the meter as SUID is in effect.

How to combine these points into a privilege evaluation logic/code that we can make it into htop, by the way?

FlexMeter.c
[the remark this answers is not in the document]
Explorer09: I didn't expect I would comment on Unicode filtering here. Do you guys plan to revise or improve the Unicode filtering code in `RichString` class? 
[Here is a draft code of mine, just for reference.](https://github.com/htop-dev/htop/pull/1586#issuecomment-2585914903)

BenBE:
> How to combine these points into a privilege evaluation logic/code that we can make it into htop, by the way?

```
file_uid = stat.uid;
file_mask = stat.mode;
file_suid = file_mask & 04000 != 0;
file_rx = file_mask & 0500 == 0500;
htop_uid = getuid();

if not file_rx:
   // do not exec
   return

if file_uid == htop_uid:
   // Execute as normal
elif file_uid == 0:
   if file_suid:
      // run SUID
   else:
      // run normally
elif htop_uid == 0:
   // Do not execute
else:
   assert(false && "Should not reach here");
```

If I didn't mess up …

Also, I'm deliberately ignoring any bits apart from the owner bits and SUID.

fasterit:
We had a quick discussion about this feature on IRC.
This is good as a test-bed and to try out the "FlexMeter" functionality. But for a merge that we can support long-term we need to see an better architected version with the comments from this PR and #526 incorporated.
Sorry, I know it is frustrating when people tell you that your work will not make mainline. But please consider it either a motivation to do a v2 _after_ discussing the architecture / implementation plan or be happy that you have a local version that scratches "your itch".
@BenBE has moved this to 3.5.0 to give you a chance to consider what you want to do with it.
/DLange

bogdanovs:
@fasterit Do you have IRC or Discord where I can join the discussions ?

BenBE:
> @fasterit Do you have IRC or Discord where I can join the discussions ?

We don't have Discord. You can find us on [libera.chat in channel #htop](https://web.libera.chat/#htop) …

FlexMeter.c
Explorer09: Minor update. I recently made an [experimental branch](https://github.com/Explorer09/htop-1/tree/refs/heads/user-config-dir) that would consolidate the operation of grabbing the config directory into a new API named `Settings_getUserHtopConfigDir`. It's not in a good quality yet, but I upload here for you to get an early preview.

PR #1572, opened 2024-12-19
CI: Update libsensors-dev package name

Explorer09:
Ubuntu has removed the transitional dummy package "libsensors4-dev", which caused failures in our CI build jobs. Change the package name to "libsensors-dev" to make them work again.


fasterit:
Removed after [jammy (22.04LTS)](https://packages.ubuntu.com/jammy/libsensors4-dev)

PR #1575, opened 2024-12-28
Simplify printing of memory values for more intuitive comprehension

eMPee584:
(please excuse the slight loss of inutile precision exhibited)

### DIFF
--- a/Row.c
+++ b/Row.c
@@ -212,23 +212,14 @@ void Row_printKBytes(RichString* str, unsigned long long number, bool coloring) 
       nextUnitColor = colors[1];
    }
 
-   if (number < 1000) {
-      // Plain number, no markings
-      len = xSnprintf(buffer, sizeof(buffer), "%5u ", (unsigned int)number);
+   if (number < ONE_K) {
+      // Below one MiB
+      len = xSnprintf(buffer, sizeof(buffer), "%4uK ", (unsigned int)number);
       RichString_appendnAscii(str, color, buffer, len);
       return;
    }
 
-   if (number < 100000) {
-      // 2 digits for M, 3 digits for K
-      len = xSnprintf(buffer, sizeof(buffer), "%2u", (unsigned int)(number / 1000));
-      RichString_appendnAscii(str, nextUnitColor, buffer, len);
-      len = xSnprintf(buffer, sizeof(buffer), "%03u ", (unsigned int)(number % 1000));
-      RichString_appendnAscii(str, color, buffer, len);
-      return;
-   }
-
-   // 100000 KiB (97.6 MiB) or greater. A unit prefix would be added.
+   // Add unit prefix.
    const size_t maxUnitIndex = (sizeof(number) * CHAR_BIT - 1) / 10 + 1;
    const bool canOverflow = maxUnitIndex >= ARRAYSIZE(unitPrefixes);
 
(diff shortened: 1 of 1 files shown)


eMPee584:
htop display of memory sizes before: huh? 🤔
> htop-before_2024-12-28_01-16-59

htop display of memory sizes after: aha! 🤓
> htop-after_2024-12-28_01-17-13

Explorer09:
May I ask, why? What's the point of removing precision or letting both 1024 GiB and 2047 GiB display as "1T"?

eMPee584:
(Sorry, this notification (and many other github mails) got buried in GMX spam folder..)

Well the high "precision" is arbitrary and confusing, it adds no value ( and is just unintuitive. I never understood what these numbers mean, didn't care to think about it and have been using `htop` for two decades (ok shame on me for that, would have been trivial to figure out.)

Look at the two screenshots and consider which is easier and clearer to understand.
Does it matter to an admin whether an app uses `94.325` or `94.493` megs of RAM? absolutely not.
**The loss of display "precision" is only at the sub-megabyte level and only for programs from `1MB` to `99MB`**.. For the example  you gave, it should still give `1024G` and `2047G`. Don't have programs that large.. but should look just like this:

image

Actually, seeing that time column, that could also need a little touch-up... :thinking:

BenBE:
The high precision is exactly what you'd want from a system monitor when evaluating the resources available on your system. And, although extreme, but there's a real difference between 1024G and 2047G of used memory, so putting it both as 1T would be of little help. Furthermore, the information are available on 4K granularity, thus only showing MiB or GiB already looses precision (BTW: always rounded down).

The current display is optimized to be compact, thus unless shaving another digit off the column width without sacrificing precision, this ain't gonna fly. ;-)

Explorer09:
@eMPee584 

> Does it matter to an admin whether an app uses `94.325` or `94.493` megs of RAM? absolutely not. 

Sorry, but it matters to me. When there are multiple threads doing similar tasks, and when I'm sorting them by memory usage. The extra precisions can tell which process consumes more memory than another.

> **The loss of display "precision" is only at the sub-megabyte level and only for programs from `1MB` to `99MB`**.. For the example you gave, it should still give `1024G` and `2047G`. Don't have programs that large.. but should look just like this:

The `1024G` and `2047G` are extreme examples I gave to tell when precisions matter. The printing routine you modified is designed to be able to show *very* large units - up to 15 zebibytes actually.

With your modifications like that, it would be impossible to tell differences between 10 GiB and 10.99 GiB. It's not only "sub-megabyte level" as you think.

> Actually, seeing that time column, that could also need a little touch-up... 🤔

That was PR #1325 for your reference.

eMPee584:
> The high precision is exactly what you'd want from a system monitor
So it matters to you whether a progress uses `94.325M` or `94.493M`? Both would be shown as `94M`. Please explain to me why you would care about those few kByte? Is it not about **magnitudes**?
I shouldn't have phrased it as a loss of precision, when it is really not relevant in this case. Please compare the two screenshots I posted and tell me whether there is a real usability loss or **ux improvement**.

> And, although extreme, but there's a real difference between 1024G and 2047G of used memory, so putting it both as 1T would be of little help.
Please carefully read my reply from earlier today again.. They will _**NOT**_ both be shown as `1T`! The change only applies to memory sizes from `1` to `99M`..

> The current display is optimized to be compact
Showing more digits is not compact, it's confusing, especially because of the difference between displaying sizes below and above `99M`.. it's incoherent, un-intuitive and confusing.

Anyways, how about a toggle then?

Explorer09:
> I shouldn't have phrased it as a loss of precision, when it is really not relevant in this case. Please compare the two screenshots I posted and tell me whether there is a real usability loss or **ux improvement**.

(emphasis in original)

I highly doubt it improves UX in any way. Apart from making the digits look more aligned, which sacrifices precision during the process.

> 
> Please carefully read my reply from earlier today again.. They will _**NOT**_ both be shown as `1T`! The change only applies to memory sizes from `1` to `99M`..

I stand corrected. But the other example I gave about `10G` and `10.9G` still holds... 

> 
> Showing more digits is not compact, it's confusing, especially because of the difference between displaying sizes below and above `99M`.. it's incoherent, un-intuitive and confusing. Look at the screenshot please. In two decades of daily use as a sys admin (on gentoo & debian), I did not figure this out.

We had trouble with base-10 megabytes and base-2 mebibytes confusion before. But I don't think your solution addresses that. (I don't even know whether you meant to address that base-2 vs base-10 issue after all.)

> Anyways, how about a toggle then?

Please give us more compelling reasons on why your format is better and why the loss of precision can be justified.

By the way, in my macOS system, I can frequently see processes that have virtual memory up to 1 terabyte large.

BenBE:
> By the way, in my macOS system, I can frequently see processes that have virtual memory up to 1 terabyte large.

You just need to start chrome for this …
image

PR #1584, opened 2025-01-09
Reflect the "show cached memory" setting in help screen

Explorer09:
Hide the "buffers/cache" text on the memory bar description when the "show cached memory" setting is off.


BenBE:
I wonder if we should completely hide these – or similar to how threads/kernel threads are handled on the main screen – simply use the shadow color to remove emphasis. Still showing them keeps these values "discoverable" to the user.

Apart from that LGTM.

fasterit:
may be add a hint why cache is not shown?

```
diff --git a/Action.c b/Action.c
index ed61881a..27daf976 100644
--- a/Action.c
+++ b/Action.c
@@ -749,7 +749,7 @@ static Htop_Reaction actionHelp(State* st) {
       addbartext(CRT_colors[MEMORY_CACHE], "/", "cache");
       addbartext(CRT_colors[BAR_SHADOW], "          ", "used");
    } else {
-      addbartext(CRT_colors[BAR_SHADOW], "                        ", "used");
+      addbartext(CRT_colors[BAR_SHADOW], "/cache hidden in setup  ", "used");
    }
    addbartext(CRT_colors[BAR_SHADOW], "/", "total");
    addattrstr(CRT_colors[BAR_BORDER], "]");
```

BenBE:
> > I wonder if we should completely hide these – or similar to how threads/kernel threads are handled on the main screen – simply use the shadow color to remove emphasis. Still showing them keeps these values "discoverable" to the user.
> 
> I'm okay with using a shadow color on the _text_ meter mode of the memory meter to hint the fact they are hidden. But on the help screen, the colors of the "buffers" and "cache" words would (in my humble opinion) confuse users when they are never shown at all.

I wouldn't change anything about the normal meter display for any of the meters. Just with the help screen.

> My idea on this change is similar to the "detail CPU time" setting - when an item is not shown in the CPU meter bar that item's description would also be hidden in the help screen.

For the detailed CPU time we could do the same on the help screen. Wouldn't mind.

> > may be add a hint why cache is not shown?
> > `addbartext(CRT_colors[BAR_SHADOW], "/cache hidden in setup  ", "used");`
> 
> If we need that hint, then we should probably hint about detailed CPU time, too.

ACK.

> Remember the reason we add an option to hide "buffers" and "cache" from the memory bars is that they don't add to the total "used" memory size. The "buffers" and "cache" can be freed up on demand when additional memory is needed for processes or kernel. If a user doesn't think they should be shown on the memory bar, we gave them this choice.

While the details are opposites for these two settings, the overall handling of these should probably be consistent. Which means if we remark the existence of one when hidden, we should also hint about the other.

BenBE (approved):

BenBE:
Okay, had a look at this in detail and I think there are two open points with this:

1. Do we want to show a hint on the help screen about cache/buffer memory when hidden
1a. Do we shadow cache/buffer memory when hidden? -> Issue with color scheme, as compressed is dark grey too
1b. Do we just show these like normal?
2. Do we hint about detailled CPU times in a similar manner?

cc @fasterit

Explorer09:
Wait. So the decision was to accept all of these changes? I thought there would be a discussion somehow.

PR #1587, opened 2025-01-14
Replaced `xMalloc+xSnprintf` with `xAsprintf` in PCP platform code

kingmidas-hack:
Hi! I submitted a PR in hopes this would suffice. After refactoring `xMalloc+xSnprintf` to `xAsprintf` within

`~/pcp/PCPDynamicColumn_addMetric.c`
`~/pcp/PCPDynamicMeter_lookupMetric.c` 
`~/pcp/PCPDynamicScreen_lookupMetric.c`

I was able to compile and test the changes (`make clean && make`) with no hassle within macOS. I'm open to any constructive criticism for writing C and overall best practices for maintaining large codebases. Thanks all!

### DIFF
--- a/pcp/PCPDynamicScreen.c
+++ b/pcp/PCPDynamicScreen.c
@@ -71,13 +71,11 @@ static void PCPDynamicScreens_appendDynamicColumns(PCPDynamicScreens* screens, P
 
 static PCPDynamicColumn* PCPDynamicScreen_lookupMetric(PCPDynamicScreen* screen, const char* name) {
    PCPDynamicColumn* column = NULL;
-   size_t bytes = strlen(name) + strlen(screen->super.name) + 1; /* colon */
-   if (bytes >= sizeof(column->super.name))
+   if ((strlen(name) + strlen(screen->super.name) + 1) >= sizeof(column->super.name)) /* colon */
       return NULL;
-
-   bytes += 16; /* prefix, dots and terminator */
-   char* metricName = xMalloc(bytes);
-   xSnprintf(metricName, bytes, "htop.screen.%s.%s", screen->super.name, name);
+    
+    char* metricName = NULL;
+    xAsprintf(&metricName, "htop.screen.%s.%s", screen->super.name, name);
 
    for (size_t i = 0; i < screen->totalColumns; i++) {
       column = screen->columns[i];
--- a/pcp/PCPDynamicScreen.c
+++ b/pcp/PCPDynamicScreen.c
@@ -73,9 +73,9 @@ static PCPDynamicColumn* PCPDynamicScreen_lookupMetric(PCPDynamicScreen* screen,
    PCPDynamicColumn* column = NULL;
    if ((strlen(name) + strlen(screen->super.name) + 1) >= sizeof(column->super.name)) /* colon */
       return NULL;
-    
-    char* metricName = NULL;
-    xAsprintf(&metricName, "htop.screen.%s.%s", screen->super.name, name);
+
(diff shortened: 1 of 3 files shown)


Explorer09:
Please remove `docs/km-notes.md` from the commit. You likely added that by mistake.

kingmidas-hack:
@Explorer09 Oof good catch, my apologies. This is done!

pcp/PCPDynamicScreen.c
```diff
@@ -71,13 +71,11 @@ static void PCPDynamicScreens_appendDynamicColumns(PCPDynamicScreens* screens, P
...
-   if (bytes >= sizeof(column->super.name))
+   if ((strlen(name) + strlen(screen->super.name) + 1) >= sizeof(column->super.name)) /* colon */
       return NULL;
-
-   bytes += 16; /* prefix, dots and terminator */
-   char* metricName = xMalloc(bytes);
-   xSnprintf(metricName, bytes, "htop.screen.%s.%s", screen->super.name, name);
+    
+    char* metricName = NULL;
+    xAsprintf(&metricName, "htop.screen.%s.%s", screen->super.name, name);
```
BenBE, line 78: Indentation should just be 3 spaces.
```suggestion
   char* metricName = NULL;
   xAsprintf(&metricName, "htop.screen.%s.%s", screen->super.name, name);
```

pcp/PCPDynamicScreen.c
```diff
@@ -71,13 +71,11 @@ static void PCPDynamicScreens_appendDynamicColumns(PCPDynamicScreens* screens, P
 
 static PCPDynamicColumn* PCPDynamicScreen_lookupMetric(PCPDynamicScreen* screen, const char* name) {
    PCPDynamicColumn* column = NULL;
-   size_t bytes = strlen(name) + strlen(screen->super.name) + 1; /* colon */
-   if (bytes >= sizeof(column->super.name))
+   if ((strlen(name) + strlen(screen->super.name) + 1) >= sizeof(column->super.name)) /* colon */
```
BenBE, line 74: `xAsprintf` already returns the actual string length. Thus you could use the return value below for this check (although this means we will allocate memory here in case of overlong column names). But given this would be on the error path this should happen rarely enough to not matter.

BenBE (commented):
Did you compile a PCP version of htop to perform your checks? Because the default on darwin won't cover these platform files.

pcp/PCPDynamicScreen.c
Explorer09: @BenBE This looks bad. While `PCPDynamicScreen_lookupMetric()` could return NULL,  there are no NULL checks on the only caller of the function, `PCPDynamicScreen_parseColumn()`. That is, you get undefined behavior with an overlong string.

BenBE: @Explorer09 This would be a bug with the current code base too. A fix would be independent of this change.

natoscott: Nice tidyup, ty @kingmidas-hack
CI is showing a clean PCP build with these changes, but looks like there's some accidentally added trailing whitespace (pcp/PCPDynamicScreen.c:76) - could you tidy that up?  Thanks.

kingmidas-hack: @natoscott No problem and most certainly! Please let me know if there’s anything else that needs to be adjusted. Thanks for catching this!

natoscott: Looks good, thanks!

PR #1588, opened 2025-01-15
Use size_t for Process offset values

Explorer09:
Convert `cmdlineBasenameStart`, `cmdlineBasenameEnd` and `procExeBasenameOffset` values in `Process` to `size_t` type. Improve functions that search for "basenames", "comm" and "procExe" to use `size_t` for offsets as well.

I made these set of code changes before the length properties of `RichString` object can be upgraded to `size_t`.

Note: This PR contains many platform specific code changes. I didn't test all of them. It is possible that I broke something.

### DIFF
--- a/Process.c
+++ b/Process.c
@@ -121,7 +121,7 @@ static int matchCmdlinePrefixWithExeSuffix(const char* cmdline, int cmdlineBaseO
     * that make htop's identification of the basename in cmdline unreliable.
     * For e.g. /usr/libexec/gdm-session-worker modifies its cmdline to
     * "gdm-session-worker [pam/gdm-autologin]" and htop ends up with
-    * proccmdlineBasenameEnd at "gdm-autologin]". This issue could arise with
+    * cmdlineBasenameStart at "gdm-autologin]". This issue could arise with
     * chrome as well as it stores in cmdline its concatenated argument vector,
     * without NUL delimiter between the arguments (which may contain a '/')
     *
--- a/Process.c
+++ b/Process.c
@@ -99,7 +99,7 @@ static bool findCommInCmdline(const char* comm, const char* cmdline, int cmdline
    return false;
 }
 
-static int matchCmdlinePrefixWithExeSuffix(const char* cmdline, int cmdlineBaseOffset, const char* exe, int exeBaseOffset, int exeBaseLen) {
+static int matchCmdlinePrefixWithExeSuffix(const char* cmdline, int* cmdlineBasenameStart, const char* exe, int exeBaseOffset, int exeBaseLen) {
    /* cmdline prefix is an absolute path: it must match whole exe. */
    if (cmdline[0] == '/') {
       int matchLen = exeBaseLen + exeBaseOffset;
@@ -124,6 +124,7 @@ static int matchCmdlinePrefixWithExeSuffix(const char* cmdline, int cmdlineBaseO
     *
     * So if needed, we adjust cmdlineBaseOffset to the previous (if any)
     * component of the cmdline relative path, and retry the procedure. */
+   int cmdlineBaseOffset = *cmdlineBasenameStart;
    bool delimFound = true; /* if valid basename delimiter found */
    do {
       /* match basename */
@@ -140,8 +141,10 @@ static int matchCmdlinePrefixWithExeSuffix(const char* cmdline, int cmdlineBaseO
             }
 
             /* full match, with exe suffix being a valid relative path */
-            if (i < 1 && j >= 1 && exe[j - 1] == '/')
+            if (i < 1 && j >= 1 && exe[j - 1] == '/') {
+               *cmdlineBasenameStart = cmdlineBaseOffset;
                return matchLen;
+            }
          }
       }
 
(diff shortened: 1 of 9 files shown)


Process.c
```diff
@@ -1057,7 +1057,7 @@ void Process_updateExe(Process* this, const char* exe) {
    if (exe) {
       this->procExe = xStrdup(exe);
       const char* lastSlash = strrchr(exe, '/');
-      this->procExeBasenameOffset = (lastSlash && *(lastSlash + 1) != '\0' && lastSlash != exe) ? (size_t)(lastSlash - exe + 1) : 0;
+      this->procExeBasenameOffset = (lastSlash > exe && *(lastSlash + 1) != '\0') ? (size_t)(lastSlash - exe + 1) : 0;
```
Explorer09, line 1060: Side note: I have no idea why this conditional needs a `lastSlash != exe`. I thought it should be legitimate when `lastSlash == exe`, that is, when the `exe` string begins with a slash such as "`/vmlinuz`".

BenBE:
FYI; the last time I needed to work on this logic, it took me the better of 3 months to track down some issue with the command line handling in this code. So be careful with your changes; the details are important with this part of the code …

Explorer09:
@BenBE 
The changes of upgrading the length values of `RichString` to `size_t` is ready here:
https://github.com/Explorer09/htop-1/commits/rich-string-chlen/

But I'm not sure I should file a new PR for it or just merge with this one. The new changes depend on this PR and I've seen you added tags and categories to this PR already.

BenBE:
@Explorer09 Make a second PR as this makes discussing the changes easier. From a quick view at your commits on that branch I already noticed some things I'd like to change with that patchset …

Process.c
```diff
@@ -64,29 +64,26 @@ void Process_fillStarttimeBuffer(Process* this) {
...
+static bool findCommInCmdline(const char* comm, const char* cmdline, size_t cmdlineBasenameStart, size_t* pCommStart, size_t* pCommLen) {
    /* Try to find procComm in tokenized cmdline - this might in rare cases
     * mis-identify a string or fail, if comm or cmdline had been unsuitably
     * modified by the process */
    const char* tokenBase;
    size_t tokenLen;
    const size_t commLen = strlen(comm);
 
-   if (cmdlineBasenameStart < 0)
-      return false;
```
BenBE, line 76: This removes a check when no basename was found in the command line …

Process.c
```diff
@@ -121,35 +117,42 @@ static int matchCmdlinePrefixWithExeSuffix(const char* cmdline, int cmdlineBaseO
...
       matchLen = exeBaseLen + cmdlineBaseOffset;
       if (cmdlineBaseOffset < exeBaseOffset &&
           strncmp(cmdline + cmdlineBaseOffset, exe + exeBaseOffset, exeBaseLen) == 0) {
-         delim = cmdline[matchLen];
+         char delim = cmdline[matchLen];
          if (delim == 0 || delim == '\n' || delim == ' ') {
-            int i, j;
+            size_t i, j;
             /* reverse match the cmdline prefix and exe suffix */
             for (i = cmdlineBaseOffset - 1, j = exeBaseOffset - 1;
```
BenBE, line 137: This could underflow for both `i` and `j` …

BenBE (commented):
Will take me some time to check consistency with all those edits. Especially with all the changes regarding edge cases re signed-nes changing due to this PR.

What might be worth doing is having the markers for "not used" become a special constant `#define STRPOS_UNUSED ((size_t)-1)`.

Process.c
Explorer09: "No basename found" is no different from matching "comm" from the start of the "cmdline" string. You notice this function does not take `cmdlineBasenameLen` as input. "No basename found" may be indicated by `cmdlineBasenameLen == 0`, but it's irrelevant to this function.

Process.c
Explorer09: Intentional. `i` and `j` may wrap over to `(size_t)-1` and that's what the conditionals below are for.

Process.c
BenBE: Just checking the change in the code has been accounted for in a different way.

Process.c
BenBE: Don't quite like doing it that way …

Especially as `x >= 0` is a no-op (always `true`) for any unsigned `x` …

Explorer09: Do you have any better way here? Note that the `i >= 0` had been rewritten to `i != (size_t)-1` by me when I changed the types to unsigned.

Process.c
```diff
@@ -99,15 +96,14 @@ static bool findCommInCmdline(const char* comm, const char* cmdline, int cmdline
    return false;
 }
 
-static int matchCmdlinePrefixWithExeSuffix(const char* cmdline, int cmdlineBaseOffset, const char* exe, int exeBaseOffset, int exeBaseLen) {
-   int matchLen; /* matching length to be returned */
-   char delim;   /* delimiter following basename */
+static size_t matchCmdlinePrefixWithExeSuffix(const char* cmdline, size_t* cmdlineBasenameStart, const char* exe, size_t exeBaseOffset, size_t exeBaseLen) {
+   size_t matchLen; /* matching length to be returned */
```
BenBE, line 100: Scope could be reduced AFAICS.

BenBE:
The AI complaint is about the `cmdlineBaseOffset - 1` and `exeBaseOffset - 1`, but the first one in particular. When you enter with `cmdlineBaseOffset` set to `0`, this underflows and is what the AI remarked about. Yes, that's defined in the standard and the code handles this situation, but a somewhat cleaner code path here would be nice.

Ideally, your last commit would only have to update the type, with all preceding ones ensuring no negative numbers appear along the modified code paths.

Process.c
Explorer09: Update: I think I can change this one a little bit. Just realized the checks against `(size_t)-1` here are too ugly, and can be avoided if we have the `i` and `j` counters incremented by 1. A decent compiler should optimise it to the same code (and if it doesn't it's a bug).

```c
            for (size_t i = cmdlineBaseOffset, j = exeBaseOffset;
               i != 0 && j != 0 && cmdline[i - 1] == exe[j - 1];
               --i, --j)
               /* empty loop body */;
            if (i == 0 && j != 0 && exe[j - 1] == '/') {
               *cmdlineBasenameStart = cmdlineBaseOffset;
               return matchLen;
            }

**Comment by Explorer09** [comment]
> The AI complaint is about the `cmdlineBaseOffset - 1` and `exeBaseOffset - 1`, but the first one in particular. When you enter with `cmdlineBaseOffset` set to `0`, this underflows and is what the AI remarked about. Yes, that's defined in the standard and the code handles this situation, but a somewhat cleaner code path here would be nice.
> 
> Ideally, your last commit would only have to update the type, with all preceding ones ensuring no negative numbers appear along the modified code paths.

[I replied here](https://github.com/htop-dev/htop/pull/1588#discussion_r2046514928). In short, maybe it helps by incrementing the `i` and `j` counters by 1.

```c
            for (size_t i = cmdlineBaseOffset, j = exeBaseOffset;
               i != 0 && j != 0 && cmdline[i - 1] == exe[j - 1];
               --i, --j)
               /* empty loop body */;
            if (i == 0 && j != 0 && exe[j - 1] == '/') {
               *cmdlineBasenameStart = cmdlineBaseOffset;
               return matchLen;
            }
```

PR #1591, opened 2025-01-18
Fix possible NULL deref when looking for PCP dynamic columns

BenBE:
Fixes a NULL deref found while working on #1587 …

This also does some code style cleanup of the caller.


natoscott (approved):
@BenBE LGTM.  I thought we were using no-braces for single-line conditionals in general these days, but I'm not particularly fussed if you really want to change that in all these places.  The new null-checking code updates there certainly all seem fine anyway.  Thanks!

BenBE:
@natoscott The brace-less variant is okay too, but within nested `if`'s the em-brace-d (scnr) variant is preferred. The top two NULL checks could have been brace-less (single-line, freestanding); no particular preference there.

PR #1585, opened 2025-01-11
Panel.scrollH clamp logic improvement

Explorer09:
Fix a logic error in commit 0128d222b94adab6b258f417722f298013a33327 (This error is not triggerable by the user.)

PR #1434, opened 2024-04-05
Properly handle RichString_extendLen when an external buffer already exists

BenBE:
The leak was found using GCC 14 and `-fanalyzer -flto`:

```
RichString.c: In function ‘RichString_extendLen’:
RichString.c:25:22: warning: leak of ‘this_25(D)->chptr’ [CWE-401] [-Wanalyzer-malloc-leak]
   25 |          this->chptr = xMalloc(charBytes(len + 1));
      |                      ^
  ‘ZramMeter_display’: events 1-2
    |
    |linux/ZramMeter.c:51:13:
    |   51 | static void ZramMeter_display(const Object* cast, RichString* out) {
    |      |             ^
    |      |             |
    |      |             (1) entry to ‘ZramMeter_display’
    |......
    |   61 |    RichString_appendAscii(out, CRT_colors[METER_TEXT], " used:");
    |      |    ~         
    |      |    |
    |      |    (2) calling ‘RichString_appendAscii’ from ‘ZramMeter_display’
    |
    +--> ‘RichString_appendAscii’: events 3-4
           |
           |RichString.c:246:5:
           |  246 | int RichString_appendAscii(RichString* this, int attrs, const char* data) {
           |      |     ^
           |      |     |
           |      |     (3) entry to ‘RichString_appendAscii’
           |  247 |    return RichString_writeFromAscii(this, attrs, data, this->chlen, strlen(data));
           |      |           ~
           |      |           |
           |      |           (4) calling ‘RichString_writeFromAscii’ from ‘RichString_appendAscii’
           |
           +--> ‘RichString_writeFromAscii’: events 5-6
                  |
                  |  135 | static inline int RichString_writeFromAscii(RichString* this, int attrs, const char* data, int from, int len) {
                  |      |                   ^
                  |      |                   |
                  |      |                   (5) entry to ‘RichString_writeFromAscii’
                  |  136 |    int newLen = from + len;
                  |  137 |    RichString_setLen(this, newLen);
                  |      |    ~               
                  |      |    |
                  |      |    (6) calling ‘RichString_setLen’ from ‘RichString_writeFromAscii’
                  |
                  +--> ‘RichString_setLen’: events 7-8
                         |
                         |   42 | static void RichString_setLen(RichString* this, int len) {
                         |      |             ^
                         |      |             |
                         |      |             (7) entry to ‘RichString_setLen’
                         |......
                         |   47 |       RichString_extendLen(this, len);
                         |      |       ~      
                         |      |       |
[...]

PR #1592, opened 2025-01-19
Upgrade lengths of RichString to `size_t` & many related changes.

Explorer09:
The main goal of this PR is to upgrade the "length" parameters and properties of `RichString` class from `int` to `size_t`. While doing the change, there are other code quality improvements that are related or required to the data type migrations.

Important changes of these include:

1. The `/proc/*/cmdline` basename matching algorithm is slightly changed to utilize the `/proc/*/exe` matching when possible. (Commit d384376d90b321c55795ed3d1a39d670a24e4c9f)
2. Add a special case to `/proc/*/comm` string highlighting. (This is required to prevent an internal variable `commStart` to go negative.)
3. Convert to `size_t` for `cmdlineBasenameStart`, `cmdlineBasenameEnd` and `procExeBasenameOffset` values in `Process` class. (#1588)
4. Add additional limit clamping to `Panel.cursorX` value.
5. Convert to `size_t` for `selectedLen` and `scrollH` values in `Panel` class. (This incorporates #1585)
6. Incorporate #1586 and fix `mbstowcs_nonfatal()`. (It's better for that PR to apply after the `RichString` lengths migrated to `size_t` type.)


DiskIOMeter.c
```diff
@@ -137,7 +137,7 @@ static void DiskIOMeter_display(ATTR_UNUSED const Object* cast, RichString* out)
 
    int color = cached_utilisation_diff > 40.0 ? METER_VALUE_NOTICE : METER_VALUE;
    int len = xSnprintf(buffer, sizeof(buffer), "%.1f%%", cached_utilisation_diff);
```
BenBE, line 139: Similar consideration re `size_t`. Maybe update signature of `xSnprintf` overall?

MainPanel.c
```diff
@@ -81,7 +82,7 @@ static HandlerResult MainPanel_eventHandler(Panel* super, int ch) {
 
    if (EVENT_IS_HEADER_CLICK(ch)) {
       int x = EVENT_HEADER_CLICK_GET_X(ch);
-      int hx = super->scrollH + x + 1;
+      size_t hx = super->scrollH + (unsigned int)x + 1;
```
BenBE, line 85: Shouldn't the second cast also be `size_t`?

Meter.c
```diff
@@ -96,46 +96,44 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
    // The text in the bar is right aligned;
    // Pad with maximal spaces and then calculate needed starting position offset
    RichString_begin(bar);
-   RichString_appendChr(&bar, 0, ' ', w);
+   RichString_appendChr(&bar, 0, ' ', (unsigned int)w);
```
BenBE, line 99: This looks overall kinda noisy. The source code shouldn't be a casting show … ;-)

Panel.h
```diff
@@ -67,10 +68,10 @@ struct Panel_ {
    Vector* items;
    int selected;
    int oldSelected;
-   int selectedLen;
+   size_t selectedLen;
    void* eventHandlerState;
    int scrollV;
```
BenBE, line 73: For consistency, this would need to change to `size_t` too.

RichString.c
```diff
@@ -21,7 +22,11 @@ in the source distribution for its full text.
 
 #define charBytes(n) (sizeof(CharType) * (n))
 
-static void RichString_extendLen(RichString* this, int len) {
+static void RichString_extendLen(RichString* this, size_t len) {
+   if (len > (SIZE_MAX - 1) / sizeof(CharType)) {
```
BenBE, line 26:
```suggestion
   if (len > SIZE_MAX / sizeof(CharType) - 1) {
```

Also, why the special handling here? Other places check only for max number of characters, not max number of bytes for storing these characters.

BenBE (changes requested):
There'S a lot of noise introduced due to the `(size_t)len` casts on `RichString_appendnAscii` calls, most of which could be avoided if we made `xSnprintf` return `size_t`. Can you take a look in that direction?

Explorer09:
> There'S a lot of noise introduced due to the `(size_t)len` casts on `RichString_appendnAscii` calls, most of which could be avoided if we made `xSnprintf` return `size_t`. Can you take a look in that direction?

Because `snprintf(3)` only returns `int`, the return type that makes more sense for `xSnprintf` is `unsigned int`, not `size_t`.

Even though I can review all calls of `xSnprintf` and update the return types, I fear that would create more noise for the changes, not less. More importantly, this would take more time of me.

Meter.c
Explorer09: That's the problem of policy and whether the compiler would complain about implicit casts to unsigned. Given there is a conditional `if (w < 1) return;` a few lines earlier, compiler should detect that the `w` value cannot go negative in this case, but I can't say for sure.

If htop is compiled with gcc or clang's `-Wsign-conversion` flag, will that cause a warning?

CPUMeter.c
```diff
@@ -137,44 +137,44 @@ static void CPUMeter_display(const Object* cast, RichString* out) {
 
    len = xSnprintf(buffer, sizeof(buffer), "%5.1f%% ", this->values[CPU_METER_NORMAL]);
```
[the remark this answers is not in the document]
Explorer09: This approach just moves the casts from the `RichString` append functions to `xSnprintf`. Unless we alter the API of `xSnprintf`, I don't think it reduces any noise overall.

How about this alternate solution: Create a `RichString_appendAsciiFormat()` function that internally combines `xSnprintf` and `RichString_appendnAscii()`?

The prototype could look something like this:
```c
/* The "buf" argument is optional. If "buf" is NULL, then this function uses an internally allocated buffer (using xAsprintf).
It's still recommended to supply "buf" for performance reasons. */
ATTR_NONNULL_N(1, 5)
size_t RichString_appendAsciiFormat(RichString* this, int attrs, char* buf, size_t bufsize, const char* format, ...);

ATTR_NONNULL_N(1, 5)
size_t RichString_appendAsciiVFormat(RichString* this, int attrs, char* buf, size_t bufsize, const char* format, va_list vl);
```

RichString.c
Explorer09: > Also, why the special handling here? Other places check only for max number of characters, not max number of bytes for storing these characters.

You don't want the total size in bytes overflow a `size_t` type, do you?
Because of the `+1` on allocated length, I cannot use `xAllocArray` to simplify things.

MainPanel.c
Explorer09: @BenBE Let me explain a bit:

* `signed int` cast to `size_t`: Perform a sign extension.
* `unsigned int` cast to `size_t`: Perform a zero extension.

When I made this patch, I made a personal rule that, if a zero extension would suffice, I would avoid sign extension. This also applies to the noisy-as-you-called-it changes about me casting the `xSnprintf` results to `size_t`.

Panel.h
Explorer09: I was considering that changing the `scrollH` type to `size_t` would be a bad move. It should have been `unsigned int` instead. It doesn't make sense to allow the UI to display 2 GiB of characters nor allow user to scroll past 2 GiB characters. In other words, it might be better to cap the `scrollH` value to `INT_MAX` whenever we have such a large data.

fasterit:
I don't like these type of "my vscode refactoring plugin had a go at the codebase" changes. There's lots of noise and it fully breaks backportability of changes to stable releases (see my Debian hat). Granted, you have done nice, small commits and for 3-4 of them there is a rationality, like "Fix potential out of bound access" or the N-in-Unicode != N-in-Ascii. I'd cherry pick these to a 3.4.1 release.
/DLange

Panel.h
BenBE: Capping the internal value is independent from the data type used (to some extend).

Also, the marked line is about `scrollV`, not `scrollH`.

RichString.c
BenBE: The expression I gave forces `len` to be limited to one byte less as its maximum. With your check, the `-1` is more or less truncated by the `/ sizeof(CharType)` …

Explorer09:
> Granted, you have done nice, small commits and for 3-4 of them there is a rationale, like "Fix potential out of bound access" or the N-in-Unicode != N-in-Ascii. I'd cherry pick these to a 3.4.1 release. /DLange

Thank you. I did intend you maintainers to cherry-pick some small commits before merging the whole thing into the mainline. (There are still style issues that need to be discussed, as you can see.)

> I don't like these type of "my vscode refactoring plugin had a go at the codebase" changes. There's lots of noise and it fully breaks backportability of changes to stable releases (see my Debian hat).

It's not the "vscode refactoring plugin" or anything like that. In fact I didn't use VSCode when refactoring this. It is my defensive coding style because of my previous experiences working in a company (that does use MSVC for compiling projects).

I'm not expecting this one gets merged quickly. What I was suggesting are things that should have been done to htop for long term maintainability. Perhaps for the next major release (when backportability is less of an issue).

RichString.c
Explorer09: Acknowledged. My logic error.

Panel.h
Explorer09: What I was considering is setting the data type of `scrollH` to `unsigned int`. And for consistency change `scrollV` to `unsigned int` as well. Even though I consider the `scrollV` changes be separate commits (because it's not directly related to or depend on `RichString`).

CPUMeter.c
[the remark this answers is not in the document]
Explorer09: Update: I've add the commit aa74b30f8f999c7c7b88e73c7bd4e65992baf1e6.

RichString.c
```diff
@@ -59,7 +59,7 @@ static void RichString_setLen(RichString* this, int len) {
 }
 
 void RichString_rewind(RichString* this, int count) {
-   RichString_setLen(this, this->chlen - count);
+   RichString_setLen(this, this->chlen > count ? this->chlen - count : 0);
```
Explorer09, line 68: Note to self: Once the types of `RichString.chlen` and `count` are changed to unsigned, we can change this expression to a `saturatingSub()` call to make the expression more concise to read. However, I found missed optimizations with this: In GCC (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=118679) and in Clang
(https://github.com/llvm/llvm-project/issues/124714)

Panel.c
```diff
@@ -386,7 +386,11 @@ bool Panel_onKey(Panel* this, int key) {
       case KEY_LEFT:
       case KEY_CTRL('B'):
          if (this->scrollH > 0) {
-            this->scrollH = MAXIMUM(this->scrollH - CRT_scrollHAmount, 0);
+            if (this->scrollH > CRT_scrollHAmount) {
+               this->scrollH = this->scrollH - CRT_scrollHAmount;
+            } else {
+               this->scrollH = 0;
+            }
```
Explorer09, line 393: This may be changed to a `saturatingSub` call if there is no [missed optimization](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=118679).

Explorer09:
@BenBE I have a question to ask when I try to improve the code of `RichString`...

I can't get the idea of what the commit a6306f090abcb1316a81444dc5514d236409349b (#1434) was trying to fix. While the PR mentioned a compiler warning about a memory leak, was that leak actually possible, or was that a false positive and the change was meant to silence it?

I could't think of a case where the buffer can actually leak, and thus I thought it was a false positive. Yet I'm not sure about that.

BenBE:
Yes, the leak was possible. Have a really close look at the trace in the compiler warning from the referenced issue.

I'd have to look at the previous state of that code before that fix, but IIRC the issue occurred, when the buffer was allocated but resized to be below the allocation limit in a later code path. In that case, the buffer no longer gets `free`'d since it's too small for the logic that `free`s the memory in the end.

Explorer09:
@BenBE I still didn't get how the leak can happen. For what I've known, the `chlen` property is only controlled by the `RichString_setLen()` function. Unless `chlen` is modified from outside the function, a `RichString` instance should have a consistent state: `this->chlen <= RICHSTRING_MAXLEN` if and only if `this->chptr == this->chstr`. What could possibly go wrong?

PR #1598, opened 2025-01-31
CRT_degreeSign code shrink

Explorer09:
Make `CRT_degreeSign` a writable buffer that will be updated after the `initDegreeSign()` function. This avoids needing a static buffer in the function.

Also use `MB_LEN_MAX` for the buffer size, which is more appropriate than the magic number 4.

One side effect of this change is that `CRT_degreeSign` will lose the const qualifier in the header (`CRT.h`).


CRT.c
```diff
@@ -94,21 +95,25 @@ const char* const* CRT_treeStr = CRT_treeStrAscii;
 
 static const Settings* CRT_settings;
 
-const char* CRT_degreeSign;
+#ifdef HAVE_LIBNCURSESW
+char CRT_degreeSign[MB_LEN_MAX] = "\xc2\xb0";
```
Explorer09, line 99: Oops. Off-by-one error. The array size should be `MB_LEN_MAX  + 1`. Will fix this in the next rebase.

CRT.c
```diff
@@ -94,21 +95,31 @@ const char* const* CRT_treeStr = CRT_treeStrAscii;
...
+      return;
+# endif
 
-   static char buffer[4];
    // this might fail if the current locale does not support wide characters
-   int r = snprintf(buffer, sizeof(buffer), "%lc", 176);
-   if (r > 0)
-      return buffer;
+   int r = snprintf(CRT_degreeSign, sizeof(CRT_degreeSign), "%lc", 176);
+   if (r <= 0)
```
BenBE:
```suggestion
   if (r <= 0 || r >= sizeof(CRT_degreeSign))
```

CRT.c
```diff
@@ -94,21 +95,31 @@ const char* const* CRT_treeStr = CRT_treeStrAscii;
 
 static const Settings* CRT_settings;
 
-const char* CRT_degreeSign;
+#ifdef HAVE_LIBNCURSESW
+# if MB_LEN_MAX >= 3 // Minimum required to support UTF-8 BMP subset
+char CRT_degreeSign[MB_LEN_MAX * 2] = "\xc2\xb0";
+# else
+char CRT_degreeSign[MB_LEN_MAX * 2] = "";
```
BenBE, line 102:
```suggestion
char CRT_degreeSign[MB_LEN_MAX + 1] = "\xc2\xb0";
# else
char CRT_degreeSign[MB_LEN_MAX + 1] = "";
```

Explorer09: Technically `+ 1` would be not enough to store the "shift sequence" to restore the shift state to initial. This only matters when the multibyte encoding is not UTF-8, though. (UTF-8 doesn't use shift states so `+ 1` is sufficient.)

CRT.c
Explorer09: I consider `r >= sizeof(CRT_degreeSign)` is impossible and that should be an assert instead.

BenBE: Completely in the realm of reality, for `MB_LEN_MAX == 1`, when `snprintf` doesn't know about `%lc` and write this to the buffer literally. And believe me, I have seen such broken implementations of `snprintf`.

Also, this check is present in the `xSnprintf` implementation you are mimicking here.

CRT.c
BenBE: This one was more of a suggestion. Not that it would make any difference based on the enforced alignment when linking anyways …

CRT.c
Explorer09: Okay. So to address a broken `snprinf` implementation. Got it.

CRT.c
Explorer09: Actually I don't like this waste of memory space either. I use `MB_LEN_MAX * 2` as a precaution. If the C language were to support Unicode only I won't need this `MB_LEN_MAX * 2` thing.

BenBE: As far as I understand this, the space reserved here is less for the UTF-8 version, but rather for non-UTF-8 encodings. So if you have an example, where things can break, I'd appreciate to knowing of that example.

That said: I think this change is more or less optional, but given we are talking 32 vs. 17 bytes, this might be worth looking into.

Explorer09: Yes, this space is reserved for non-UTF-8, multibyte encoding. [Glibc set this `MB_LEN_MAX` to 16 just to be future-proof](https://lists.gnu.org/r/bug-gnulib/2015-05/msg00001.html), but comparing to the cost of `malloc`'ing the buffer. It's still very small and is negligible.

BenBE: Ah, k. Given that musl libc has `MB_LEN_MAX` at 4 somewhat dampens the fun. So yeah, will thus remain at `2*MB_LEN_MAX`.

Explorer09: Because of the dominance of Unicode, it no longer makes sense to define `MB_LEN_MAX` to any value larger than 4 (or 6 if you want to support the full ISO 10646 range [0x0, 0x7FFFFFFF] of UTF-8). But Glibc is a historical exception (16 - I admit it's an overkill), and there is nothing we could do.

Issue #1600, opened 2025-02-04
Segmentation fault on full memory and cpu usage


APN-Pucky:
Compiling some stuff, my laptop struggles with its memory and cpu resources. It is difficult to reproduce and I only encountered it twice sofar. Nonetheless, I would not expect htop to SegFault because of that:

```
FATAL PROGRAM ERROR DETECTED
============================
Please check at https://htop.dev/issues whether this issue has already been reported.
If no similar issue has been reported before, please create a new issue with the following information:
  - Your htop version: '3.3.0'
  - Your OS and kernel version (uname -a)
  - Your distribution and release (lsb_release -a)
  - Likely steps to reproduce (How did it happen?)
  - Backtrace of the issue (see below)

Error information:
------------------
A signal 11 (Segmentation fault) was received.

Setting information:
--------------------
[...]

PR #1604, opened 2025-02-08
Fix possible buffer OOB access when parsing the stat file for a process

BenBE:
Fixes: #1600


fasterit (approved):
Looks good. Defensive programming at its best.
The alternative would be to secure the fast_* functions return a valid empty string.
/DLange

BenBE:
> The alternative would be to secure the `fast_*` functions return a valid empty string.

They already do. ;-) If you reach any non-digit, the function will stop. If the reached character is a NUL byte, the function will not advance past it. Cf. `fast_strtoull_dec` for the only function that actually advances through the provided string buffer (all others just forward to there).

Issue #1597, opened 2025-01-30
GPU monitoring (percentage and time) in Darwin/macOS


Explorer09:
This is a feature ticket, for people who are more familiar with Darwin/macOS to try and implement.

The Activity Monitor app in macOS has GPU Percentage and GPU Time columns for process monitoring. There is also a "GPU History" feature that shows the GPU usage as a graph.

This webpage (from a third party, not Apple) has brief descriptions and screenshots about the GPU monitoring in macOS: https://www.alphr.com/view-gpu-usage-mac/

It might be possible for such feature to be implemented in htop, too.

After #1288, we have GPU monitoring in Linux now. The GPU meter and columns shouldn't be limited to Linux.

aestriplex:
@BenBE cc @Explorer09 
It would be great to have that feature in htop. If it's ok for you, I'd like to start thinking about a possible implementation.
I would start by separating two aspects of the problem.
1. GPU model and its overall usage;
2. GPU utilization by process.

I attach below a sample program that prints the overall GPU usage, which could be a first step to solve the first point, once properly adapted to the htop context.
Unfortunately, as far as I know, there is no direct way to solve the second problem, as there is no API to trace this information. I would therefore start with the first point, adding a section (similar to the CPU or RAM usage section) that contains: 
* The model and version of the GPU;
* The number of cores;
* Their total GPU core utilization.

```c
#include <stdio.h>
#include <IOKit/IOKitLib.h>
#include <CoreFoundation/CoreFoundation.h>

int main() {
    io_service_t service = IOServiceGetMatchingService(kIOMainPortDefault, IOServiceMatching("IOGPU"));

    if (!service) {
        printf("GPU not found\n");
        return 1;
    }

    CFMutableDictionaryRef properties = NULL;
    if (IORegistryEntryCreateCFProperties(service, &properties, kCFAllocatorDefault, kNilOptions) != KERN_SUCCESS) {
        fprintf(stderr, "Failed to get GPU properties\n");
        IOObjectRelease(service);
        return 1;
    }

    CFDictionaryRef perfStats = CFDictionaryGetValue(properties, CFSTR("PerformanceStatistics"));

    if (perfStats && CFGetTypeID(perfStats) == CFDictionaryGetTypeID()) {
        CFNumberRef rendererUtil, tilerUtil;
        
        rendererUtil = CFDictionaryGetValue(perfStats, CFSTR("Renderer Utilization %"));
[...]

PR #1606, opened 2025-02-10
GPU monitoring (percentage and time) in Darwin/macOS

aestriplex:
This PR follows from the discussion in #1597. As said there, I split my changes in different commit, starting from the refactoring of `linux/GPUMeter.c` and `linux/GPUMeter.h`.
@Explorer09 @BenBE I open it as a draft because of the GPUMeter refactoring thing. The code compiles and runs on linux and macOS (both x86 and Silicon), but I am going to express some perplexities (as inline comments) for what concerns the unified `GPUMeter.c` in the root directory. 

I should have pushed this one on Friday, but I've had a quite busy weekend, sorry. Anyway, this week I should be more available.


GPUMeter.c
```diff
@@ -0,0 +1,80 @@
...
+#include "config.h" // IWYU pragma: keep
+
+#include "GPUMeter.h"
+
+#include "CRT.h"
+#include "RichString.h"
+
+
+static size_t activeMeters;
+static const int GPUMeter_attributes[] = {
```
aestriplex, line 17: This one is necessary for LinuxMachine, but it's useless for DarwinMachine, since the APIs return the usage metric go the GPU as a whole, and each Silicon chip as  a single GPU engine. Unfortunately, since the `GPUMeter_class` is allocated in this file, we cannot define this array in platform specific machines. This should not be a problem for macOS machines, but it can be a bit misleading for developers

GPUMeter.c
```diff
@@ -0,0 +1,80 @@
...
+}
+
+static double totalUsage;
+
+static void GPUMeter_updateValues(Meter* this) {
+   char* buffer = this->txtBuffer;
+   size_t size = sizeof(this->txtBuffer);
+   int written;
+
+	totalUsage = Machine_getGpuUsage(this->host, this->values);
```
aestriplex, line 36: Since the metric is just one, I thought it would be nice to have an API that return just the floating point number that you actually see in the metric itself. The way this number is calculated is platform specific

GPUMeter.c
```diff
@@ -0,0 +1,80 @@
...
+      .extends = Class(Meter),
+      .delete = Meter_delete,
+      .display = GPUMeter_display,
+   },
+   .init = GPUMeter_init,
+   .done = GPUMeter_done,
+   .updateValues = GPUMeter_updateValues,
+   .defaultMode = BAR_METERMODE,
+   .supportedModes = METERMODE_DEFAULT_SUPPORTED,
+   .maxItems = ARRAYSIZE(GPUMeter_attributes),
```
aestriplex, line 74: Same thing of [the comment on line 17](https://github.com/htop-dev/htop/pull/1606/files#r1948703318)

darwin/DarwinMachine.c
```diff
@@ -125,3 +125,17 @@ bool Machine_isCPUonline(const Machine* host, unsigned int id) {
 
    return true;
 }
+
+double Machine_getGpuUsage(const Machine* super, double values[]) {
```
aestriplex, line 129: I have to add the two APIs for all the platforms to make the code compile. It seems to me a good solution, because it works even if someone adds the `&GPUMeter_class` reference in the `Platform.c` for this machine, so it's quite robust.

linux/LinuxMachine.c
```diff
@@ -820,3 +824,108 @@ bool Machine_isCPUonline(const Machine* super, unsigned int id) {
...
+
+   if (value < 96)
+      return xSnprintf(buffer, size, "%3lluh", value);
+
+   value /= 24;
+
+   return xSnprintf(buffer, size, "%3llud", value);
+}
+
+double Machine_getGpuUsage(const Machine* super, double values[]) {
```
Explorer09, line 872: I consider this function belongs to Platform.c, not here.
Also, pass the pointer to the whole Meter instance (`Meter *`), not just the `values` array, as the Meter instance would be more useful.

linux/LinuxMachine.c
```diff
@@ -820,3 +824,108 @@ bool Machine_isCPUonline(const Machine* super, unsigned int id) {
...
+
+      values[i] = 100.0 * GPUMeter_engineData[i].timeDiff / (1000 * 1000) / monotonictimeDelta;
+   }
+
+   values[ARRAYSIZE(GPUMeter_engineData)] = 100.0 * saturatingSub(curResidueTime, prevResidueTime) / (1000 * 1000) / monotonictimeDelta;
+
+   return 100.0 * totalGPUTimeDiff / (1000 * 1000) / monotonictimeDelta;
+}
+
+void Machine_GPUMeterDisplay(const Object* cast, RichString* out, double totalUsage) {
```
Explorer09, line 900: I can't see how it make sense for having this function Linux-only. (This function was formerly named `GPUMeter_display`, IIRC.) The text for displaying the GPU stat should be platform independent, no?

linux/LinuxMachine.c
```diff
@@ -820,3 +824,108 @@ bool Machine_isCPUonline(const Machine* super, unsigned int id) {
    assert(id < super->existingCPUs);
    return this->cpuData[id + 1].online;
 }
+
+static int humanTimeUnit(char* buffer, size_t size, unsigned long long int value) {
```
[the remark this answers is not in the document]
Explorer09: See [my comment about `Machine_GPUMeterDisplay`](https://github.com/htop-dev/htop/pull/1606#discussion_r1948769490)

Machine.h
```diff
@@ -35,6 +35,8 @@ in the source distribution for its full text.
 typedef unsigned long long int memory_t;
 #define MEMORY_MAX ULLONG_MAX
 
+extern const int gpuMetricMaxItems;
```
Explorer09, line 38: Unused variable?

BenBE:
The `GPUMeter_attributes` is the list of attributes that a certain meter wants to display and is linked to the colors to use for these values. This can/should be platform-independent and will contain only 1 value usually (the usage). You could have more attributes, if there are distinct usage patterns that can be distinguished, like low priority, user space load, kernel space load, soft/hard IRQs, … are for the normal CPU meter.

The display function should be independent of the platform; only retrieving the values for the meter (`*_updateValues`) should call into platform-specific code as needed.

GPUMeter.c
BenBE: Treat the GPU meter as a overall GPU usage meter for now. Reworking this similar to the CPU meter to support multiple GPUs should be done separately.

Each of these `GPUMeter_attributes` is one value displayed in the bar graph (like low/normal/kernel/softirq/hardirq/… usage for the CPU meter). This is not about multiple GPUs …

GPUMeter.c
```diff
@@ -0,0 +1,80 @@
...
+   GPU_ENGINE_3,
+   GPU_ENGINE_4,
+   GPU_RESIDUE,
+};
+
+bool GPUMeter_active(void) {
+   return activeMeters > 0;
+}
+
+static double totalUsage;
```
BenBE, line 29: Try to group (module-global) variables to the top.

GPUMeter.c
```diff
@@ -0,0 +1,80 @@
...
+static void GPUMeter_updateValues(Meter* this) {
+   char* buffer = this->txtBuffer;
+   size_t size = sizeof(this->txtBuffer);
+   int written;
+
+	totalUsage = Machine_getGpuUsage(this->host, this->values);
+   
+   if (totalUsage < 0) {
+      this->values[0] = 0;
+      written = snprintf(buffer, size, "N/A");
```
BenBE, line 40: Use `xSnprintf`.

GPUMeter.c
```diff
@@ -0,0 +1,80 @@
...
+      METER_BUFFER_CHECK(buffer, size, written);
+   } else {
+      written = snprintf(buffer, size, "%.1f", totalUsage);
```
BenBE, line 43: Use `xSnprintf`.

GPUMeter.c
```diff
@@ -0,0 +1,80 @@
...
+      METER_BUFFER_CHECK(buffer, size, written);
+
+      METER_BUFFER_APPEND_CHR(buffer, size, '%');
+   }
+}
+
+static void GPUMeter_display(const Object* cast, RichString* out) {
+   Machine_GPUMeterDisplay(cast, out, totalUsage);
```
BenBE, line 51: Displaying the values of the meter should not depend on the platform.

Machine.h
```diff
@@ -94,4 +96,8 @@ void Machine_scan(Machine* this);
 
 void Machine_scanTables(Machine* this);
 
+double Machine_getGpuUsage(const Machine* super, double values[]);
```
BenBE, line 99: Passing unbounded arrays is dangerous. Also, the `Machine*` can just be extended to hold all the information for the GPU meter as necessary.

```suggestion
double Machine_getGpuUsage(const Machine* super);
```

Machine.h
```diff
@@ -94,4 +96,8 @@ void Machine_scan(Machine* this);
...
+
+void Machine_GPUMeterDisplay(const Object* cast, RichString* out, double totalUsage);
```
BenBE, line 101: This should be independent of the platform.

linux/LinuxMachine.c
BenBE: It absolutely should be platform-independent …

BenBE (changes requested):

aestriplex:
I declared three of the (formerly) static variable in `GPUMetric.c` as extern in `GPUMetric.h`
```c
extern const MeterClass GPUMeter_class;
extern struct EngineData GPUMeter_engineData[4];
extern unsigned long long int totalGPUTimeDiff;
```
Because now they must be shared between `/GPUMetric.c` and `<platform>/Platform.c`.

aestriplex:
@BenBE Thank you for [your explanation]( https://github.com/htop-dev/htop/pull/1606#discussion_r1948825802). I was trying to understand the meaning of the attributes structure by looking at other examples of meters. In some of them it was quite clear that these attributes refer to different kind of values displayed by the graph, e.g. this one
```c
static const int FileDescriptorMeter_attributes[] = {
   FILE_DESCRIPTOR_USED,
   FILE_DESCRIPTOR_MAX
};
```
In the case of the GPUMeter, I was mislead by the word "ENGINE", though

linux/LinuxMachine.c
BenBE: Given the CPU data is also handled as part of the `*Machine` structure, putting the GPU data there feels much more natural to me, than putting it in the `*Platform` code. This also allows for putting the code in the correct structures more easily, without violating the encapsulation.

BenBE:
> I declared three of the (formerly) static variable in `GPUMetric.c` as extern in `GPUMetric.h`
> 
> ```c
> extern const MeterClass GPUMeter_class;
> extern struct EngineData GPUMeter_engineData[4];
> extern unsigned long long int totalGPUTimeDiff;
> ```
> 
> Because now they must be shared between `/GPUMetric.c` and `<platform>/Platform.c`.

Sorry, for zigzagging things here, but as noted [in my reply](https://github.com/htop-dev/htop/pull/1606#discussion_r1949974304) to @Explorer09, a more natural solution is actually to put these with the `<Platform>Machine` structure, similar to how information for CPUs are handled.

The `GPUMeter_class` should go platform-independent in `GPUMeter.c`, The `GPUMeter_engineData[4]` should go into the `<Platform>Machine` structure, and `totalGPUTimeDiff` would go to the `Machine` structure. Have a look how similar fields for CPU information are placed inside these structures for comparison. If you are unsure, feel free to ask (if you would like to avoid refactoring things over and over again).

linux/Platform.c
```diff
@@ -1104,3 +1105,33 @@ void Platform_done(void) {
...
+double Platform_getGpuUsage(Meter* meter) {
+   unsigned int i;
+   uint64_t monotonictimeDelta;
+   const Machine* super = meter->host;
+   const LinuxMachine* this = (const LinuxMachine*) super;
+   GPUEngineData *gpuEngineData;
+
+   assert(ARRAYSIZE(GPUMeter_engineData) + 1 == ARRAYSIZE(GPUMeter_attributes));
+
+	totalGPUTimeDiff = saturatingSub(this->curGpuTime, this->prevGpuTime);
```
Explorer09, line 1119: Indent messed up. (There is a tab character here; please change it to spaces.)

linux/Platform.c
```diff
@@ -146,6 +146,7 @@ static enum { BAT_PROC, BAT_SYS, BAT_ERR } Platform_Battery_method = BAT_PROC;
 static time_t Platform_Battery_cacheTime;
 static double Platform_Battery_cachePercent = NAN;
 static ACPresence Platform_Battery_cacheIsOnAC;
+static unsigned long long int prevResidueTime, curResidueTime;
```
Explorer09, line 149: When moving these two variables to platform code, I suggest renaming these to `prevGPUResidueTime` and `curGPUResidueTime` to reduce ambiguity.

By the way, `curGPUResidueTime` may be converted into a local (function scope) variable, but that's a separate commit.

linux/Platform.c
```diff
@@ -1104,3 +1105,33 @@ void Platform_done(void) {
    LibSensors_cleanup();
 #endif
 }
+
+
+double Platform_getGpuUsage(Meter* meter) {
+   unsigned int i;
+   uint64_t monotonictimeDelta;
+   const Machine* super = meter->host;
+   const LinuxMachine* this = (const LinuxMachine*) super;
```
Explorer09, line 1114:
```suggestion
   const Machine* host = meter->host;
   const LinuxMachine* lhost = (const LinuxMachine*) host;
```

Avoid the names `this` and `super` for non-methds.

Explorer09:
> > The `GPUMeter_class` should go platform-independent in `GPUMeter.c`, The `GPUMeter_engineData[4]` should go into the `<Platform>Machine` structure, and `totalGPUTimeDiff` would go to the `Machine` structure. Have a look how similar fields for CPU information are placed inside these structures for comparison. If you are unsure, feel free to ask (if you would like to avoid refactoring things over and over again).
> 
> @BenBE Ok, I am putting the `Machine_setGPUValues` (with the name modified as indicate [here](https://github.com/htop-dev/htop/pull/1606#discussion_r1950514370)) in `<Platform>Machine`. I cannot use a pointer to a Meter structure in the signature though, because I have to declare the prototype in Machine.h, and the Meter structure depends on the Machine structure. So I am thinking of this
> 
> ```c
> double Machine_setGPUValues(Machine* this, double *values, size_t values_size);
> ```
> 
> That way, although we do not have the meter structure, we still have the machine and the (bounded) values array.

This function should be dedicated to updating the `Meter` instance. Without a `Meter` instance, this function would be useless. The direct output of `double *values` makes no sense.

BenBE:
I did some refactoring for the code structure (WIP), that might better illustrate, where things should go.

Code can be found at https://github.com/BenBE/htop/tree/%231597

Basically:
* GPU meter code in `GPUMeter.c`
* GPU metric fetching in `<Platform>Machine.c`
* GPU display code in `GPUMeter.c`
* Platform-independent statistics (e.g. total GPU time) to `Machine.h`
* Platform-dependent statistics (e.g. EngineData) to `<Platform>Machine.h`

The refactoring isn't done yet, but most parts should already be visible.

aestriplex:
@BenBE Thank you for sharing your refactoring. But at this point I have to say that probably I didn't get the purpose of this function in the first place. Because I put most of my efforts in trying to update the meters' values from within the <platform>Machine module (that's the main reason why I was passing the array [here](https://github.com/htop-dev/htop/pull/1606#issuecomment-2651427929), because I couldn't pass the Meter* directly). It's not a problem for me, at all. It has been an opportunity to learn something about htop's metrics. In this regard I was thinking about: what if I start a new branch writing some dev documentation about what we have done here? It could be useful for others.

Anyway, I'll build the macOS version on top of that.

Machine.h
```diff
@@ -94,4 +97,6 @@ void Machine_scan(Machine* this);
 
 void Machine_scanTables(Machine* this);
 
+double Machine_updateGpuUsage(Machine* this);
```
aestriplex, line 100: @BenBE There is one last thing about this method. The compiler returns the following warning

```
warning: passing 'const Machine *' (aka 'const struct Machine_ *') to parameter of type 'Machine *' (aka 'struct Machine_ *') discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
```
Because, when we call it from within the GPUMeter, we have a `const Machine*`. The code works, but it seems kind of strange to just gat rid of a const modifier, especially given the fact that we have to modify some internals of the machine structure in this function. Since we are not updating the meters' values anymore couldn't we just put this logic into an internal function within the <platforma>Machine, and then call it in
```c
void Machine_scan(Machine* super) {
   LinuxMachine* this = (LinuxMachine*) super;

   LinuxMachine_scanMemoryInfo(this);
   LinuxMachine_scanHugePages(this);
   LinuxMachine_scanZfsArcstats(this);
   LinuxMachine_scanZramInfo(this);
   LinuxMachine_scanCPUTime(this);
   <here, something like LinuxMachine_scanGPUUsage(this);>
   ...
```
and then, in the GPUMeter.c, just take the value from the machine structure and print it?

BenBE: I'm aware of this issue in my refactoring and I'll do a few more tweaks to fix that issue. In your work based on that refactoring you may assume the update routine for each platform gets called properly and only once (that's another known issue if you had multiple GPU meters active).

darwin/DarwinMachine.c
```diff
@@ -132,19 +133,21 @@ bool Machine_isCPUonline(const Machine* host, unsigned int id) {
...
    if (!dhost->GPUService) {
-      return -1;
+      return;
    }
 
-   if (IORegistryEntryCreateCFProperties(dhost->GPUService, &properties, kCFAllocatorDefault, kNilOptions) != KERN_SUCCESS) {
-      return -1;
+   ret = IORegistryEntryCreateCFProperties(dhost->GPUService, &properties, kCFAllocatorDefault, kNilOptions);
+   if (ret != KERN_SUCCESS) {
+      CRT_debug("IORegistryEntryCreateCFProperties failed while retrieving GPU service properties [%d]", ret);
```
BenBE, line 149: Try to avoid printing error messages repeatedly for each refresh cycle.

aestriplex: I removed those two error messages, and I just put one in `Machine_new`, just in case GPUService can't be allocated

GPUMeter.c
```diff
@@ -83,72 +73,36 @@ static int humanTimeUnit(char* buffer, size_t size, unsigned long long int value
 }
 
 static void GPUMeter_updateValues(Meter* this) {
-   const Machine* host = this->host;
-   const LinuxMachine* lhost = (const LinuxMachine*) host;
-   const GPUEngineData* gpuEngineData;
+   const Machine* machine = this->host;
```
Explorer09, line 76: Could you bring back the `host` name for consistency with other code?

darwin/DarwinMachine.c
```diff
@@ -105,6 +108,11 @@ Machine* Machine_new(UsersTable* usersTable, uid_t userId) {
    openzfs_sysctl_init(&this->zfs);
    openzfs_sysctl_updateArcStats(&this->zfs);
 
+   this->GPUService = IOServiceGetMatchingService(kIOMainPortDefault, IOServiceMatching("IOGPU"));
+   if(!this->GPUService) {
```
Explorer09, line 112: A space
```suggestion
   if (!this->GPUService) {
```

darwin/DarwinMachine.c
```diff
@@ -125,3 +135,34 @@ bool Machine_isCPUonline(const Machine* host, unsigned int id) {
...
+   CFMutableDictionaryRef properties = NULL;
+   CFDictionaryRef perfStats;
+   CFNumberRef deviceUtil;
+   int device = 0, ret;
+
+   if (!dhost->GPUService) {
+      return;
+   }
+
+   ret = IORegistryEntryCreateCFProperties(dhost->GPUService, &properties, kCFAllocatorDefault, kNilOptions);
```
Explorer09, line 150: Declare variables just before their first use. We are targeting for C99 and so there's no need to put declarations before any statement.

```suggestion
   CFMutableDictionaryRef properties = NULL;
   int ret = IORegistryEntryCreateCFProperties(dhost->GPUService, &properties, kCFAllocatorDefault, kNilOptions);
```

By the way, if `ken_return_t` is available from the included headers, then prefer `kern_return_t` to `int`.

darwin/DarwinMachine.c
```diff
@@ -125,3 +135,34 @@ bool Machine_isCPUonline(const Machine* host, unsigned int id) {
...
+   if (ret != KERN_SUCCESS) {
+      return;
+   }
+
+   perfStats = CFDictionaryGetValue(properties, CFSTR("PerformanceStatistics"));
```
Explorer09, line 155: Ditto. Declare variables just before their first use.
```suggestion
   CFDictionaryRef perfStats = CFDictionaryGetValue(properties, CFSTR("PerformanceStatistics"));
```

darwin/DarwinMachine.c
```diff
@@ -125,3 +135,34 @@ bool Machine_isCPUonline(const Machine* host, unsigned int id) {
...
+
+   assert(perfStats && CFGetTypeID(perfStats) == CFDictionaryGetTypeID());
```
Explorer09, line 157: Split to two asserts?
```suggestion
   assert(perfStats);
   assert(CFGetTypeID(perfStats) == CFDictionaryGetTypeID());
```

darwin/DarwinMachine.c
```diff
@@ -125,3 +135,34 @@ bool Machine_isCPUonline(const Machine* host, unsigned int id) {
...
+   
+   deviceUtil = CFDictionaryGetValue(perfStats, CFSTR("Device Utilization %"));
+   if (!deviceUtil) {
+      return;
```
Explorer09, line 161: I see a `CFRelease` call after this `return` statement. Would this early return cause a resource leak? Please help us check it out.

darwin/DarwinMachine.c
```diff
@@ -125,3 +135,34 @@ bool Machine_isCPUonline(const Machine* host, unsigned int id) {
...
    return true;
 }
+
+void Machine_scanGPUUsage(Machine* super) {
+   DarwinMachine* dhost = (DarwinMachine *)super;
+   CFMutableDictionaryRef properties = NULL;
+   CFDictionaryRef perfStats;
+   CFNumberRef deviceUtil;
+   int device = 0, ret;
+
```
Explorer09, line 145: If the total GPU time is not retrievable from this machine (I mean macOS), please set it to a dummy value that we can tell it's N/A.
I suggest `(unsigned long long)-1` as a placeholder here.
```c
   totalGPUTimeDiff = (unsigned long long)-1;
```

darwin/DarwinMachine.c
aestriplex: I removed the first assert and checked for perfStats not NULL instead

darwin/DarwinMachine.c
aestriplex: I added a static function that initialize the two GPU fields in the Machine structure to -1, that way they are always set even in case of early exit.
```c
static void DarwinMachine_setDeafultGPUValues(Machine* super) {
   super->totalGPUUsage = (double)-1;
   super->totalGPUTimeDiff = (unsigned long long int)-1;
}
```
This function is called only here, but I thought it would make the code more readable by clarifying the fact that -1 is the default value of these two parameters, and not just a magic number

darwin/DarwinMachine.c
aestriplex: Yes, my bad. I just used the code of the standalone script I wrote to reply to this issue, in which I obviously did not care to deallocate any resource. I fixed it in any early exit.
I tested for memory leaks on both platforms (the old Macs with x86 and those with Apple Silicon). I report the output of the tool leaks for both below
```bash
Process:         htop [1162]
Path:            /Users/USER/Documents/*/htop
Load Address:    0x1022b4000
Identifier:      htop
Version:         0
Code Type:       ARM64
Platform:        macOS
Parent Process:  leaks [1161]

Date/Time:       2025-02-16 22:27:28.742 +0100
Launch Time:     2025-02-16 22:27:06.245 +0100
OS Version:      macOS 15.1.1 (24B91)
Report Version:  7
Analysis Tool:   /usr/bin/leaks

Physical footprint:         8818K
Physical footprint (peak):  8866K
Idle exit:                  untracked
----

leaks Report Version: 4.0, multi-line stacks
Process 1162: 1260 nodes malloced for 1209 KB
Process 1162: 0 leaks for 0 total leaked bytes.
```
```bash
Process:         htop [93318]
Path:            /Users/USER/Documents/*/htop
Load Address:    0x10ccdc000
Identifier:      htop
Version:         0
Code Type:       X86-64
Platform:        macOS
Parent Process:  leaks [93317]

Date/Time:       2025-02-16 10:58:04.530 pm +0100
Launch Time:     2025-02-16 10:57:45.224 pm +0100
OS Version:      macOS 12.7.6 (21H1320)
Report Version:  7
Analysis Tool:   /usr/bin/leaks

Physical footprint:         5168K
Physical footprint (peak):  5176K
----

leaks Report Version: 4.0, multi-line stacks
Process 93318: 742 nodes malloced for 920 KB
Process 93318: 0 leaks for 0 total leaked bytes.
```

darwin/DarwinMachine.c
```diff
@@ -79,7 +79,7 @@ static void DarwinMachine_getVMStats(DarwinMachine* this) {
 
 static void DarwinMachine_setDeafultGPUValues(Machine* super) {
    super->totalGPUUsage = (double)-1;
-   super->totalGPUTimeDiff = (unsigned long long int)-1;
+   super->totalGPUTimeDiff = (unsigned long long)-1;
```
BenBE, line 82: Why remove the `int` from `unsigned long long int`?

aestriplex:
Just to make it shorter. Should I put the int  back again?

darwin/DarwinMachine.c
aestriplex: Just to make it shorter. Should I put the int  back again?

darwin/DarwinMachine.c
Explorer09: 1. Please inline the `DarwinMachine_setDeafultGPUValues` call into `Machine_new`. Don't make a separate function for that because the function is useless and doesn't align with the code styles of other objects in htop.
2. For floating point value, we tend to use `NAN` rather than `-1` for "unknown" or "not available" values. Even though `-1` could work currently, I suggest changing it to `NAN` for consistency.

darwin/DarwinMachine.c
Explorer09: @aestriplex The point is unnecessary line changes that complicated our code review. The commit history is messy currently. You should squash commits so that one commit does one logical change, which can ease other people reviewing your code.

darwin/DarwinMachine.h
```diff
@@ -9,6 +9,7 @@ in the source distribution for its full text.
 
 #include <mach/mach_host.h>
 #include <sys/sysctl.h>
+#include <IOKit/IOKitLib.h>
```
BenBE, line 12: Alphabetical, unless required differently (annotate if that's the case)

```suggestion
#include <IOKit/IOKitLib.h>
#include <mach/mach_host.h>
#include <sys/sysctl.h>
```

darwin/DarwinMachine.c
```diff
@@ -16,6 +16,8 @@ in the source distribution for its full text.
 #include <unistd.h>
 #include <sys/mman.h>
 #include <sys/sysctl.h>
+#include <IOKit/IOKitLib.h>
+#include <CoreFoundation/CoreFoundation.h>
```
BenBE, line 20: Header order.

darwin/DarwinMachine.c
Explorer09: @aestriplex Clarification. There should be two commits for this PR:

1. Refactor linux/GPUMeter into platform independent GPUMeter.
2. Add GPUMeter code for Darwin.

darwin/DarwinMachine.c
aestriplex: @Explorer09 Ok, I'll initialize them directly in Machine_new. For the NAN part, should I change the `Machine_updateGpuUsage` function of every <platform>Machine to make it return NAN  as well?

```c
double Machine_updateGpuUsage(Machine* super ATTR_UNUSED) {
   return NAN;
}
```

aestriplex:
@BenBE @Explorer09 Now There are just two commit in this branch:
1. Refactor linux/GPUMeter into platform independent GPUMeter.
2. Add GPUMeter code for Darwin.

Sorry I didn't do this sooner, but I was afraid that by squashing commits and force pushing, the comment references would be lost. This has happened to me in the past using other systems (e.g. BitBucket).

aestriplex:
@BenBE I notice that in Machine_scan of LinuxMachine, I am passing `Machine_scanGPUUsage(this);` instead of `Machine_scanGPUUsage(super);`, I fix it and squash it in `Refactor linux/GPUMeter into platform independent GPUMeter`

aestriplex:
I squashed it in `Add GPUMeter code for Darwin`, because the call in Machine_scan was made there

unsupported/UnsupportedMachine.c
```diff
@@ -54,3 +54,7 @@ void Machine_scan(Machine* super) {
    super->usedSwap = 0;
    super->cachedSwap = 0;
 }
+
+double Machine_updateGpuUsage(Machine* super ATTR_UNUSED) {
+   return -1;
+}
```
BenBE, line 60: This still seems to differ from other platform implementations.

Explorer09:
> Now There are just two commit in this branch:
> 
> 1. Refactor linux/GPUMeter into platform independent GPUMeter.
> 2. Add GPUMeter code for Darwin.
> 
> Sorry I didn't do this sooner, but I was afraid that by squashing commits and force pushing, the comment references would be lost. This has happened to me in the past using other systems (e.g. BitBucket).

When you push commits into GitHub, GitHub's database would keep permanent copies of those commits even when they're never referenced again in later branches.

GitHub can even detect whether the commit being discussed is "outdated" and indicate as such in the UI. Therefore, no need to worry about references.

darwin/DarwinMachine.c
```diff
@@ -126,8 +140,36 @@ bool Machine_isCPUonline(const Machine* host, unsigned int id) {
...
+   }
+   
+   assert(CFGetTypeID(perfStats) == CFDictionaryGetTypeID());
+   
+   CFNumberRef deviceUtil = CFDictionaryGetValue(perfStats, CFSTR("Device Utilization %"));
+   if (!deviceUtil) {
+      goto cleanup;
+   }
+   
+   int device = NAN;
```
Explorer09, line 169: Integer cannot hold a NaN value. Does this not produce a compiler warning?

aestriplex: This was clearly a mistake, as I misinterpreted the type of device var. Surprisingly enough, the compiler (clang) on macOS does not produce any warning for that. I tried to initialise another int variable to NAN an compiling it with gcc (running on Ubuntu) and I get this warning:
```bash
GPUMeter.c:97:18: warning: overflow in conversion from 'float' to 'int' changes value from '+QNaNf' to '0' [-Woverflow]
    97 |    int written = NAN;
```
This behaviour of clang is quite curious. I also tried to activate the option `-Woverflow`, but nothing happened

Explorer09:
This PR can be closed as #1620 is merged into `main`. #1620 supersedes this PR.

PR #1607, opened 2025-02-14
Fix #1599 DarwinProcess.c compiler warning when debug is enabled

### LINKED ISSUE #1599: Compile warning in DarwinProcess.c when debug is enabled
```
darwin/DarwinProcess.c:499:39: warning: comparison of integers of different signs: 'pid_t' (aka 'int') and 'uint64_t' (aka 'unsigned long long') [-Wsign-compare]
  499 |       assert(Process_getPid(tprocess) == tid);
      |              ~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/assert.h:75:25: note: expanded from macro 'assert'
   75 |     (__builtin_expect(!(e), 0) ? __assert_rtn(__func__, __ASSERT_FILE_NAME, __LINE__, #e) : (void)0)
      |                         ^
```

OS: macOS Sequoia 15.3
Compiler:

```
Apple clang version 16.0.0 (clang-1600.0.26.6)
Target: arm64-apple-darwin24.3.0
```

Reproducible also in CI: https://github.com/Explorer09/htop-1/actions/runs/13082758381/job/36509406353

Arkanic:
Fixes #1599 for debug builds

Adds an additional assert to confirm that tid (which is a signed value and potentially negative) is above zero and therefore not an error code, before casting it to uint64_t to do an equivalent comparison

### DIFF
--- a/darwin/DarwinProcess.c
+++ b/darwin/DarwinProcess.c
@@ -496,7 +496,11 @@ void DarwinProcess_scanThreads(DarwinProcess* dp, DarwinProcessTable* dpt) {
          continue;
       }
 
-      assert(Process_getPid(tprocess) == tid);
+      pid_t tprocessPid = Process_getPid(tprocess);
+      assert(tprocessPid >= 0);
+      assert((uint64_t)tprocessPid == tid);
+      (void)tprocessPid;
+
       Process_setParent(tprocess, pid);
       Process_setThreadGroup(tprocess, pid);
       tprocess->super.show       = true;


BenBE (changes requested):
This will give a warning about `tprocessPid` being unused in non-debug builds.

Add a `(void)tprocessPid;` after the `assert`s to tell the compiler this is intentional.

darwin/DarwinProcess.c
```diff
@@ -496,7 +496,9 @@ void DarwinProcess_scanThreads(DarwinProcess* dp, DarwinProcessTable* dpt) {
          continue;
       }
 
-      assert(Process_getPid(tprocess) == tid);
+      pid_t tprocessPid = Process_getPid(tprocess);
+      assert(tprocessPid >= 0);
```
Explorer09, line 500: The fix doesn't look correct. Why allow PID 0 here?
Besides, are we stuck with requiring an unsigned type for the PID argument in `ProcessTable_getProcess`?

Arkanic: I understand that it may seem odd but for simplicity shouldn't this check only make sure that the pid is a non-error code? any weird edge cases should be handled by checking that pid and tid are actually the same

As for the unsigned return of `ProcessTable_getPid`, why should it be signed if it will never return a negative number? if anything it could only create more expectations in code that getPid will return a wider range of values than it actually does, leading to dead checks in the code

Explorer09: > any weird edge cases should be handled by checking that pid and tid are actually the same

My concern is that `tid` might be a negative number wrongly converted to unsigned in the first place.

> As for the unsigned return of `ProcessTable_getPid`, why should it be signed if it will never return a negative number?

For consistency of use of `pid_t`. And we cannot assume `sizeof(pid_t)` equals `sizeof(uint64_t)`. Using wrong size for PID values can hide potential bugs.

Arkanic: looking through the code, tid is ultimately defined by `thread_info` (mach lib function), the choice to be a uint64_t is a choice by the mach devs defining the members of the `thread_identifier_info_data_t` struct rather than a misuse on this end of the codebase - formatting this particular value into a pid_t seems possible but a wrapper function would be needed for that library call and it seems that it would be a bit "messy" (would you return a new struct that has a single member changed, or would you "translate" every `uint64_t` pid into a `pid_t` by calling a method individually when you get a `uint64_t` pid?

IMO a couple of asserts seems the best and simplest fix at the moment but if anyone was planning on overhauling that would be interesting to look at

BenBE (approved):
LGTM. Would you mind squashing these two commits into one?

Arkanic:
done

BenBE:
Did a quick reword of the commit message, as the fact it was squashed is much less interesting than the issue it fixes.

Also, no explicit reference to the "unused" tagging needs to be done, as this needs to be integral part of the commit that fixes the original issue.

Arkanic:
My fault, I didn't change the default git squash commit message

Issue #622, opened 2021-05-09
Process CPU state not shown on MacOS


BenBE:
Sorted by CPU, but it doesn't shows it correctly which process is taken CPU:
<img width="1060" alt="выява" src="https://user-images.githubusercontent.com/9673243/117583418-c3a61780-b10f-11eb-9b2f-344346c421bc.png">
I've builded rust from mac ports. 
<img width="864" alt="выява" src="https://user-images.githubusercontent.com/9673243/117583447-e1737c80-b10f-11eb-8afb-f0da80c1fb18.png">

_Originally posted by @sharewax in https://github.com/htop-dev/htop/issues/446#issuecomment-835861914_

sharewax:
Apple M1 Silicon btw, Big Sur 11.3.1.

BenBE:
@sharewax: Can you test if the CPU state has been showing previously on some release version and if so, on which one? You can get to the various release versions quickest by cloning the git repository and checking out the various tags.

If you find an old version that worked. it'd be also really helpful if you could do a `git bisect` from there. If you need some assistance with that, feel free to ask.

sharewax:
Checked back to 2.1.0 version (2018 year) - it doesn't show correct state. Does it have sence to check more? Probably not.
My main question was not about state, but about CPU percent - because it's shows 0.4% for firefox and 0.0 for rustc during compilcation.

BenBE:
Ah, k. In that case I misunderstood you on that part. Nonetheless, the missing state information is something that probably should be fixed. Given that it's been broken for that long strongly hints it never worked.

Regarding the (far too) low CPU usage display, that's probably still some issue with either the read-out or some conversion. Unfortunately I don't have any Apple hardware myself, thus someone else will have to have a look into this.

risentveber:
This requires to disable SIP `csrutil disable` (in recovery mode), and after rebooting just start `sudo htop` - see original https://github.com/hishamhm/htop/issues/701

PR #1609, opened 2025-02-19
darwin: fix process state visualization in DarwinProcessTable

aestriplex:
There is a problem with the process state column on macOS. All the processes but htop itself are marked as ProcessState.UNKNOWN ('?'). This behavior was referenced in this old issue: #622.

`DarwinProcess_scanThreads` contains the code related to the mach subsystem, that retrieves the information thread by thread. For some of them (as process state information) this is not necessary. We have this information right in the `kinfo_proc` structure, allocated in `DarwinProcessTable.c`. This allows us to avoid dealing with threads unless strictly necessary, such as in the case of process state information.

We have an analogous issue with the running time, that is not updated during the execution of the program, but I'll open a separated pr for that.


darwin/DarwinProcessTable.c
```diff
@@ -102,11 +118,12 @@ void ProcessTable_goThroughEntries(ProcessTable* super) {
 
       DarwinProcess_setFromKInfoProc(&proc->super, &ps[i], preExisting);
       DarwinProcess_setFromLibprocPidinfo(proc, dpt, time_interval_ns);
-
+
```
BenBE, line 121: Whitespace @ EOL.

darwin/DarwinProcessTable.c
```diff
@@ -27,6 +27,22 @@ in the source distribution for its full text.
 #include "generic/openzfs_sysctl.h"
 #include "zfs/ZfsArcStats.h"
 
+static ProcessState ProcessTable_mapDarwinProcessState(int kinfo_proc_state) {
```
BenBE, line 30: Two blank lines after the block of headers.

BenBE (approved):

PR #745, opened 2021-08-16
Add increment and decrement controls for numeric display options

ahmedaabouzied:
* Display controls to increment & decrement the numeric options.
* Display control to select/deselect non-numeric options with the enter key.
* This allows incrementing and decrementing values with the mouse.

Resolves: #737 

Screenshot after applying the change:
Screenshot (9)

Issue #737, opened 2021-08-13
Numeric option items should have increment & decrement controls


Explorer09:
A usability problem I recently discovered.

![htop-display-numeric-options](https://user-images.githubusercontent.com/2487263/129303748-11a9fd44-86ea-4347-80f6-0fc5c2d87901.png)

In "Setup"->"Display options", several items take numeric values instead of boolean/checkbox. However, there're no obvious hints on how to adjust the value. The users have to guess which keys would adjust the values.

Suggestion: _The function bar should include at least "increment" and "decrement" buttons for the numeric option item._ This (1) shows the user which keys to press in order to adjust the values, and (2) it allows mouse users to adjust the values too.

I might also suggest a "revert" button in the function bar, but that would be a separate discussion.

ahmedaabouzied:
Hi @Explorer09 , @BenBE,

I'd like to work on this little task. It's the first time I'll be working with the `htop` source code. I think this is a good place for me to start.

As @Explorer09  suggested, we can add `| + | Increment     | - | Decrement    |` buttons  added to the function bar that only show up when the user is on a numeric value in the display options. 

I see that the keys `+`, `-` increment and decrement the numeric values. But the `Enter` key increments the values as well. I found that the `Enter` key increments the values confusing. Should we keep it that way?

BenBE:
Hi @ahmedaabouzied,

that's correct: The <kbd>+</kbd> and <kbd>-</kbd> keys increment and decrement the value respectively. The <kbd>Enter</kbd> key is intentional, as setting all the other (non-numeric) options reacts to <kbd>Enter</kbd> just the same.

If you have any questions or need some help with the implementation feel free to ask.

rustedusted:
hey there can i work on this i saw that there is no active discussion for this for last few years so just thought if i could take up the issue 

also if it's fine with you guys can you put a hacktoberfest label on this as i'm participating :D

rustedusted:
i also have a suggestion can we even have f1 f2 to increament and decreament values when hovering on that option so that people can read on bar below how to modify values 

just asking or is there a better way

fasterit:
<kbd>F7</kbd> and <kbd>F8</kbd> would make more sense as that is <tt>Nice -</tt> and <tt>Nice +</tt> on the main panel

rustedusted:
oh yeah makes sense :D will do that then

rustedusted:
hey there quick update there are only 6 spaces for writing things on the bottom row 

[...]

PR #1610, opened 2025-02-19
DisplayOptions: "Inc"/"Dec" operations in the function bar 

hishamhm:
For numeric entries in Display Options, this gives users a hint of which keys operate on the value, and allows the value to be increasable/decreasable using the mouse as well.

### DIFF
--- a/AffinityPanel.c
+++ b/AffinityPanel.c
@@ -136,20 +136,19 @@ typedef struct AffinityPanel_ {
 
 static void AffinityPanel_delete(Object* cast) {
    AffinityPanel* this = (AffinityPanel*) cast;
-   Panel* super = (Panel*) this;
-   Panel_done(super);
    Vector_delete(this->cpuids);
    #ifdef HAVE_LIBHWLOC
    hwloc_bitmap_free(this->workCpuset);
    MaskItem_delete((Object*) this->topoRoot);
    #endif
+   Panel_done(&this->super);
    free(this);
 }
 
 #ifdef HAVE_LIBHWLOC
 
 static void AffinityPanel_updateItem(AffinityPanel* this, MaskItem* item) {
-   Panel* super = (Panel*) this;
+   Panel* super = &this->super;
 
    item->value = hwloc_bitmap_isincluded(item->cpuset, this->workCpuset) ? 2 :
                  hwloc_bitmap_intersects(item->cpuset, this->workCpuset) ? 1 : 0;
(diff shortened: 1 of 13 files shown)


Explorer09:
Related: #737 and #745.

May supersede #745, but I don't understand why another pull request here.

BenBE:
I don't quite like the extra cleanup code needed for this version. Cf. #772 for a different approach to this.

hishamhm:
> May supersede #745, but I don't understand why another pull request here.

Ah, I didn't see there was already a PR for this. Feel free to discard this one. (You still might want to take note of b2e86aa.)

It honestly took me a few seconds until I remembered that the "-" key decremented the value (at first I tried pressing space multiple times assuming the value would wraparound or something). Since the original point of the function bar at the bottom was to make the UI naturally discoverable, I thought that the operations applicable to the selected item should appear there, like it happens in the Meters screen.

If that's already being taken care of, all the better! Cheers!

BenBE:
> > May supersede #745, but I don't understand why another pull request here.
> 
> Ah, I didn't see there was already a PR for this.

Happens. I had to look it up too.

> Feel free to discard this one. (You still might want to take note of [b2e86aa](https://github.com/htop-dev/htop/commit/b2e86aa8040db75d5ecc345f0138d0bca19680df).)

I already cherry-picked that commit onto main.

> It honestly took me a few seconds until I remembered that the "-" key decremented the value (at first I tried pressing space multiple times assuming the value would wraparound or something). Since the original point of the function bar at the bottom was to make the UI naturally discoverable, I thought that the operations applicable to the selected item should appear there, like it happens in the Meters screen.

Yes. And for some other screens the keys already were noted so this is somewhat for consistency.

> If that's already being taken care of, all the better! Cheers!

All mentioned PRs are somewhat incomplete for various details. In #745 there was some memory leak which I tried to fix with #772, but that PR hit a NULL deref on the way.

What bugs me with this PR is the global state introduced by the ctor logic; I'd somewhat prefer if the function bar had a somewhat more specific lifetime than "well, when the program closes".

hishamhm:
> What bugs me with this PR is the global state introduced by the ctor logic; I'd somewhat prefer if the function bar had a somewhat more specific lifetime than "well, when the program closes".

I copied this straight from the MetersPanel.c logic, as you can imagine! :)

I don't remember _why_ that one is static (because it is shared by multiple panels, perhaps?). I guess for DisplayOptions the FunctionBar could be just owned by the Panel itself. Yeah, I like that, it's way less code.

hishamhm:
cherry on top: as per @Explorer09's suggestion in #745 I flipped the button order. Also makes it consistent with "- Nice" and "+ Nice" in the main function bar!

hishamhm:
Thanks, folks! <3

PR #1613, opened 2025-02-21
Beautify the + / - options for numeric fields as shown in the status bar

fasterit:
Use Unicode full-width characters where available


BenBE:
Tests with the following consoles yielded negative feedback:

1. QTerminal on Lubuntu 24.10 (default theme) gives a noticeable shift between the two states of the status bar. The symbols are rendered correctly though.

2. Linux text console (default font used) renders the symbols as U+FFFD and has the noticeable shift in the status bar. With the symbols not properly showing, this is a usability regression.

I'd suggest to stick with the ASCII-only version where proper display of information is key (and actually carries essential information).

fasterit:
As stated on IRC: 😭
or `:-(` for those living on ASCII still

Explorer09:
@fasterit Just for your info: Many fonts designed with Latin scripts only will not contain glyphs for the fullwidth Latin block. The fullwidth characters you are using are for the CJK scripts. That is, you generally need a CJK font installed before the fullwidth glyphs can be displayed properly.

I like the alternative of using F7 and F8 instead of "-" and "+" for display in #1615, however.

PR #1615, opened 2025-02-21
Make Dec and Inc consistent on F7 and F8 as per the main panel

fasterit:
* minus and plus keys still work
* this prevents the ugly left flush "- " and "+ " key labels
* Unicode "－" and "＋" are still too hard in 2025, see PR #1613 😭

Closes #1613

### DIFF
--- a/DisplayOptionsPanel.c
+++ b/DisplayOptionsPanel.c
@@ -22,11 +22,11 @@ in the source distribution for its full text.
 #include "ScreensPanel.h"
 
 
-static const char* const DisplayOptionsFunctions[] = {"      ", "      ", "      ", "      ", "      ", "      ", "      ", "      ", "      ", "Done  ", NULL};
+static const char* const DisplayOptionsFunctions[] =       {"      ", "      ", "      ", "      ", "      ", "      ", "      ", "      ", "      ", "Done  ", NULL};
 
-static const char* const DisplayOptionsDecIncFunctions[] = {"Dec   ", "Inc   ", "                                                      ", "Done  ", NULL};
-static const char* const DisplayOptionsDecIncKeys[] = {"- ", "+ ", "  ", "F10", NULL};
-static const int DisplayOptionsDecIncEvents[] = {'-', '+', ERR, KEY_F(10)};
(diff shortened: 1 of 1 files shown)


DisplayOptionsPanel.c
```diff
@@ -22,11 +22,11 @@ in the source distribution for its full text.
 #include "ScreensPanel.h"
 
 
-static const char* const DisplayOptionsFunctions[] = {"      ", "      ", "      ", "      ", "      ", "      ", "      ", "      ", "      ", "Done  ", NULL};
+static const char* const DisplayOptionsFunctions[] =       {"      ", "      ", "      ", "      ", "      ", "      ", "      ", "      ", "      ", "Done  ", NULL};
 
-static const char* const DisplayOptionsDecIncFunctions[] = {"Dec   ", "Inc   ", "                                                      ", "Done  ", NULL};
-static const char* const DisplayOptionsDecIncKeys[] = {"- ", "+ ", "  ", "F10", NULL};
-static const int DisplayOptionsDecIncEvents[] = {'-', '+', ERR, KEY_F(10)};
+static const char* const DisplayOptionsDecIncFunctions[] = {"      ", "      ", "      ", "      ", "      ", "      ", " Dec  ", " Inc  ", "      ", "Done  ", NULL};
```
BenBE, line 27: Items are left-aligned within the button label …
```suggestion
static const char* const DisplayOptionsDecIncFunctions[] = {"      ", "      ", "      ", "      ", "      ", "      ", "Dec   ", "Inc   ", "      ", "Done  ", NULL};
```

BenBE (changes requested):
Not a fan of the whitespace alignment changes surrounding this, but well.

Currently display is inconsistent, as all function keys are shown for the Dec/Inc bar, but not for everything else.

DisplayOptionsPanel.c
fasterit: I like the middle alignment better when we have the space...

BenBE: It's just inconsistent with everywhere else in the UI though …

fasterit: Ugly wins. You may only wear grey cat ears from now on!

BenBE (approved):

Issue #1611, opened 2025-02-20
Assertion fail in macOS (#1609 regression)


Explorer09:
Just built the release candidate of htop-3.4.0 (b1474aac9a11f0abe525f10971270be20e127056) on macOS Sequoia, and this is what I got:

```
>>>>>>>>>> stderr output >>>>>>>>>>
Assertion failed: (0), function processStateChar, file Process.c, line 533.

<<<<<<<<<< stderr output <<<<<<<<<<
```

```
Error information:
------------------
A signal 6 (Abort trap: 6) was received.

Backtrace information:
----------------------
0   htop                                0x0000000102ba7d18 CRT_handleSIGSEGV + 288
1   libsystem_platform.dylib            0x000000018d686de4 _sigtramp + 56
2   libsystem_pthread.dylib             0x000000018d64ff70 pthread_kill + 288
3   libsystem_c.dylib                   0x000000018d55c908 abort + 128
4   libsystem_c.dylib                   0x000000018d55bc1c err + 0
5   htop                                0x0000000102bc3520 Process_done.cold.1 + 0
6   htop                                0x0000000102bb35a4 Process_done + 0
7   htop                                0x0000000102bb46ac Row_display + 72
8   htop                                0x0000000102bb119c Panel_draw + 1148
9   htop                                0x0000000102bb6d10 ScreenManager_run + 824
10  htop                                0x0000000102ba574c CommandLine_run + 1804
11  dyld                                0x000000018d2d0274 start + 2840
```

It's a regression of #1609.

Explorer09:
Trying to diagnose this one. The assertion error happens only when the "Hide userland process threads" option is off (i.e. show the threads), and the error happens on htop's own thread. See the screenshot.

And there is one thing strange with #1609 in which many processes now show with the status "R" while they should be "S" (sleeping) instead. This is shown on the same screenshot.

<img width="357" alt="Image" src="https://github.com/user-attachments/assets/b5a79299-5513-418a-9d1b-d07a6caff0ba" />

Cc: @aestriplex

BenBE:
Have a look at the value passed to `processStateChar` in `state`.

Also, make sure you rebuild clean, to avoid some old object files from interfering.

Explorer09:
@BenBE I did rebuild everything. And by reverting the changes of that PR, htop works again (no assertion error), even though many processes' statuses show "?" instead.

BenBE:
[...]

PR #1617, opened 2025-02-22
darwin: not every process should be in running state (#1611)

aestriplex:
This PR follows from the discussion in #1611

Summary:
* removed the static function `ProcessTable_mapDarwinProcessState`;
* added `proc->super.state = pti.pti_numrunning > 0 ? RUNNING : SLEEPING;` in `DarwinProcess_setFromLibprocPidinfo`;
* refactored `DarwinProcess_setFromLibprocPidinfo`:
    - added early exit;
    - introduced XNU constant `PROC_PIDTASKINFO_SIZE` instead of `sizeof`.
* removed a redundant system call in a closed scope

### DIFF
--- a/darwin/DarwinProcess.c
+++ b/darwin/DarwinProcess.c
@@ -369,7 +369,8 @@ void DarwinProcess_setFromKInfoProc(Process* proc, const struct kinfo_proc* ps, 
 void DarwinProcess_setFromLibprocPidinfo(DarwinProcess* proc, DarwinProcessTable* dpt, double timeIntervalNS) {
    struct proc_taskinfo pti;
 
-   if (sizeof(pti) != proc_pidinfo(Process_getPid(&proc->super), PROC_PIDTASKINFO, 0, &pti, sizeof(pti))) {
+   if (PROC_PIDTASKINFO_SIZE != proc_pidinfo(Process_getPid(&proc->super), PROC_PIDTASKINFO, 0, &pti, PROC_PIDTASKINFO_SIZE)) {
+      proc->taskAccess = false;
       return;
    }
 
@@ -390,6 +391,7 @@ void DarwinProcess_setFromLibprocPidinfo(DarwinProcess* proc, DarwinProcessTable
    }
    Process_updateCPUFieldWidths(proc->super.percent_cpu);
 
+   proc->super.state = pti.pti_numrunning > 0 ? RUNNING : SLEEPING;
    proc->super.time = nanosecondsToCentiseconds(total_current_time_ns);
    proc->super.nlwp = pti.pti_threadnum;
    proc->super.m_virt = pti.pti_virtual_size / ONE_K;
@@ -433,18 +435,6 @@ void DarwinProcess_scanThreads(DarwinProcess* dp, DarwinProcessTable* dpt) {
          CRT_debug("task_for_pid(%d) failed: %s", pid, mach_error_string(ret));
       dp->taskAccess = false;
       return;
-   }
-
-   {
-      task_info_data_t tinfo;
-      mach_msg_type_number_t task_info_count = TASK_INFO_MAX;
-      ret = task_info(task, TASK_BASIC_INFO, (task_info_t) &tinfo, &task_info_count);
-      if (ret != KERN_SUCCESS) {
-         CRT_debug("task_info(%d) failed: %s", pid, mach_error_string(ret));
-         dp->taskAccess = false;
-         mach_port_deallocate(mach_task_self(), task);
-         return;
-      }
    }
 
    thread_array_t thread_list;
--- a/darwin/DarwinProcess.c
+++ b/darwin/DarwinProcess.c
@@ -369,40 +369,41 @@ void DarwinProcess_setFromKInfoProc(Process* proc, const struct kinfo_proc* ps, 
 void DarwinProcess_setFromLibprocPidinfo(DarwinProcess* proc, DarwinProcessTable* dpt, double timeIntervalNS) {
    struct proc_taskinfo pti;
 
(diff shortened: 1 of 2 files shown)


darwin/DarwinProcess.c
```diff
@@ -434,18 +438,6 @@ void DarwinProcess_scanThreads(DarwinProcess* dp, DarwinProcessTable* dpt) {
       return;
    }
 
-   {
```
aestriplex, line 437: This syscall uses two variables that are declared in this scope. The only thing that modifies outside of its own scope is `dp->taskAccess = false`. If this call fails because of the fact that we cannot get `TASK_BASIC_INFO`, so any of the other 3 syscall done in this module would fail and set `dp->taskAccess = false`. It seems to me that avoid a syscall for every process for every iteration would be a quite decent optimization

fasterit:
Linting complains as follows:
darwin/DarwinProcess.c:393: trailing whitespace.
darwin/DarwinProcess.c:450: trailing whitespace.

aestriplex:
@fasterit Thanks, fixed

fasterit:
This works as intended. Good work!
The native `top` also shows "stuck" tasks quite often, do we have an idea what that is and do we want to represent it?

darwin/DarwinProcess.c
BenBE: Mind to split the `dp->taskAccess` optimization into its own commit? Change itself LGTM.

Explorer09:
Just tested this PR and it seems to work great.

I would argue it shows more status than macOS's `top` does, especially on the Stopped ("T") state.

There is another "stuck" state that are shown in macOS's `top`, which display as Sleeping ("S") in htop instead with this PR. Although I didn't have the need to distinguish this "stuck" state from "sleeping" right now, it's still a room for a future enhancement.

(I tested this with a little program called [stress](https://formulae.brew.sh/formula/stress) that can generate workloads, and I can occasionally see "stuck" states with `stress`. See the screenshot below.)

htop-screenshot-darwin-process-state-4

darwin/DarwinProcess.c
Explorer09: Can you please pull in this change https://github.com/Explorer09/htop-1/commit/fe73c2ff4966e45b0b3a5502f38cc8fdd1ea5096 and apply the PR after it?

I made the indent adjustment a separate commit so that the new changes can be easier to review.

1. https://github.com/Explorer09/htop-1/commit/fe73c2ff4966e45b0b3a5502f38cc8fdd1ea5096
2. https://github.com/htop-dev/htop/commit/376afa909fcba7391933ce1e1ea590815e10b711

aestriplex:
@fasterit
> The native top also shows "stuck" tasks quite often, do we have an idea what that is and do we want to represent it?

@Explorer09 
> There is another "stuck" state that are shown in macOS's top, which display as Sleeping ("S") in htop instead with this PR. Although I didn't have the need to distinguish this "stuck" state from "sleeping" right now, it's still a room for a future enhancement.

I also think that it could be a good future enhancement

fasterit:
> @fasterit
> 
> > The native top also shows "stuck" tasks quite often, do we have an idea what that is and do we want to represent it?
> 
> @Explorer09
> 
> > There is another "stuck" state that are shown in macOS's top, which display as Sleeping ("S") in htop instead with this PR. Although I didn't have the need to distinguish this "stuck" state from "sleeping" right now, it's still a room for a future enhancement.
> 
> I also think that it could be a good future enhancement

Please get an understanding what "stuck" tasks are and how to identify them. This is so we don't have to complete rewrite the logic yet again. I'd like to be sure this **can** be an incremental change.
/DLange

aestriplex:
@fasterit in `top`, a stuck stack is a sleeping task that is waiting for resources. Usually it's waiting for a I/O resource that is currently blocked, or for a mutex tat lock some global resource. In top's man page, stuck is mentioned as `"stuck" (i.e. uninterruptible sleep)`.

Anyway, by using public APIs (like the one we are using from sysctl or libproc) it's not so straightforward how to get this information. Probably we can get it by adding another syscall that shows us if a process is waiting for resources, and then we could do something like
```c
if (ps->state == SLEEPING && isWaitingForResources(ps)) {
   ps->state = STUCK;
}
```

Although it's not clear to me which syscall shows the most reliable info in this respect, as far as I know, it *should* be  an incremental change, because the set of stuck processes is a subset of sleeping processes.

fasterit:
https://github.com/apple-oss-distributions/top/blob/47e40b78502828778d0632952e193442f4d24c74/libtop.c#L644 looks useful

aestriplex:
@fasterit Yes, it seems useful, especially the way it's used [here](https://github.com/apple-oss-distributions/top/blob/47e40b78502828778d0632952e193442f4d24c74/libtop.c#L1639). Should I open a new pr for this change?

darwin/DarwinProcess.c
BenBE: I think, from this note, only the `dp->taskAccess` split is still left.

aestriplex:
Thank you @BenBE.
For what concerns stuck processes, simply adding the condition `extended_info.pth_run_state == TH_STATE_UNINTERRUPTIBLE` in `DarwinProcess_scanThreads` does not work. The threads I see in stuck state in top do not meet this condition in htop. I am currently working to see which syscalls return *reliable* information about the threads, process by process. For this reason, if it's ok with you, I would open a separate issue (with separate pr). Instances of stuck processes are quite rare (on my mac I only see them when I use the `stress` utility), so I think you might consider a small additional feature.

aestriplex:
@BenBE Sounds good to me

PR #1622, opened 2025-02-24
Darwin running time

aestriplex:
At the moment, the running time of processes on macOS is not updated correctly. I made two commit to try to fix this problem. 

The [branch](https://github.com/aestriplex/htop/tree/darwin-running-time) is rebased on the PR https://github.com/htop-dev/htop/pull/1617.

Summary:

Removed the call to Platform_machTicksToNanoseconds. The proc_taskinfo structure contains the running time (both system and user) of the process. We should just take those value in nanoseconds and convert them in the proper unit. How the kernel handles this calculation (starting from low level cpu-dependent details) should not concern us.
I also removed the static function nanosecondsToCentiseconds because for some reason we lose precision (± 1ms) with that calculation. If we divide directly by 1e7, we get a more precise measure. Probably it's a good idea to have an inline function or (maybe better) a macro, just to give a name to that operation and avoid the magic number;
Since the GPU percentage update depends on the running time of the process, I updated the condition in the if statement. In fact, suppose we are at the nth iteration of the DarwinProcess_setFromLibprocPidinfo function. We should update the CPU usage value if the total CPU running time of the process is greater than the CPU running time of the process in the previous iteration. To do that, we do not need to use the timeIntervalNS.
TODO:

remove the static double Platform_nanosecondsPerMachTick var in Darwin/Platform.c as well. It's only used in the (removed) Platform_machTicksToNanoseconds.
remove the Platform_calculateNanosecondsPerMachTick function in Darwin/PlatformHelpers.c. It's only used to set the static double Platform_nanosecondsPerMachTick var once.


BenBE:
@aestriplex Could you complete the pending two TODOs? TIA.

aestriplex:
@BenBE Yes, I am committing those asap

aestriplex:
@BenBE I made a separate commit

aestriplex:
OpenBSD build pipeline failed with error

```
[build-openbsd-latest-clang](https://github.com/htop-dev/htop/actions/runs/13520021995/job/37777020241)
The job running on runner GitHub Actions 14 has exceeded the maximum execution time of 20 minutes.
```

May be due to temporary conditions, should I re-run it manually later?

BenBE:
Can you check if `mach_timebase_info` is used anywhere else and update the `configure.ac` if there's no further reference left?

BenBE (approved):
LGTM.

PR #1623, opened 2025-02-25
darwin: nanoseconds to centi-seconds minor improvement

Explorer09:
Avoid the use of floating point division when integer division can do it.
A follow-up of commit 9944d5507f9d57f8c03b0e183523fb28eee79005.

(This very minor change doesn't need to be made in 3.4.0 release. I'm okay when this one is pulled to 3.4.1 instead.)

### DIFF
--- a/darwin/DarwinProcess.c
+++ b/darwin/DarwinProcess.c
@@ -383,7 +383,8 @@ void DarwinProcess_setFromLibprocPidinfo(DarwinProcess* proc, DarwinProcessTable
    Process_updateCPUFieldWidths(proc->super.percent_cpu);
 
    proc->super.state = pti.pti_numrunning > 0 ? RUNNING : SLEEPING;
-   proc->super.time = total_current_time_ns / 1e7;
+   // Convert from nanoseconds to hundredths of seconds
+   proc->super.time = total_current_time_ns / 10000000ULL;
    proc->super.nlwp = pti.pti_threadnum;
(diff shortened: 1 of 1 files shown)


BenBE:
Given that the function is used only once I was quite happy, that the previous commit got rid of it.

Also, it's `ULL`, not `U`.

So, just change the literal in the appropriate line and we're fine.

fasterit:
I suggest to keep the comment though so we are less likely to guess around why this is centiseconds in a few years... :smiley:

darwin/DarwinProcess.c
```diff
@@ -383,7 +388,7 @@ void DarwinProcess_setFromLibprocPidinfo(DarwinProcess* proc, DarwinProcessTable
    Process_updateCPUFieldWidths(proc->super.percent_cpu);
 
    proc->super.state = pti.pti_numrunning > 0 ? RUNNING : SLEEPING;
-   proc->super.time = total_current_time_ns / 1e7;
+   proc->super.time = nanosecondsToCentiseconds(total_current_time_ns);
```
BenBE, line 391:
```suggestion
   // Converts nanoseconds to hundredths of a second (centiseconds) as needed by the "time" field of the Process struct.
   proc->super.time = total_current_time_ns / 10000000ULL;
```

Explorer09: Done

BenBE (approved):

Issue #1601, opened 2025-02-07
On a Mac, "delete" key doesn't delete a meter


Explorer09:
This is a usability problem that's worth discussing.

Mac keyboards have the "delete" key that actually works like backspace. The delete key commonly known in PC are known as "forward delete" on a Mac (sometimes presented as ⌦ (U+2326) icon). On MacBooks and small Mac keyboards, there is usually no dedicated "forward delete" button and to forward delete one has to press "Fn + Delete" instead. (By the way, "Fn + Delete" cannot be pressed with one hand as the two buttons are positioned far away on the keyboard.)

In htop's meter setup screen, the bottom function bar tells the user can delete a meter by pressing the "Del" key, but on a Mac, pressing "delete" there does nothing. This has confused me for a while.

And please don't tell me I can also press F9 to delete a meter. My MacBook is unfortunately the model with touch bar so that the F1 - F12 keys are hidden under the Fn key by default.

The fix of this issue can be as simple as allowing Backspace to work as Delete unless there is a need to distinguish between the two keys' functions. But I can't tell if this is worth it.

BenBE:
A possible solution may be to allow backspace to be interpreted as DEL for the setup screen AND only on Mac …

Explorer09:
> A possible solution may be to allow backspace to be interpreted as DEL for the setup screen AND only on Mac …

I don't like platform specific hotkeys and if we want to implement this, I wish this Backspace/Delete behavior work consistently across platforms, too. The reason is I do run emulators inside macOS in order to test things on other OSes, including htop.

aestriplex:
> The fix of this issue can be as simple as allowing Backspace to work as Delete unless there is a need to distinguish between the two keys' functions. But I can't tell if this is worth it.

I made a little investigation. It seems to me that adding Backspace hook does not interfere with the handling of the event related to the Del key. I added this behaviour in two different points:
* ColumnsPanel.c
* MetersPanel.c

I am opening a PR with this changes. Probably I should change the description below as well. Since the backspace key works for every platform, and the del key for every platform but macOS, IMO we can follow to paths here:
* Rename Del to Bkspc for every platform. For those who were used to use Del (on others platforms) nothing change;
* Rename Del to Bkspc just for macOS

fasterit:
No need to rename everything because the Mac stuff is special.
@aestriplex: Which of the three keys you added is the one that maps to "Del" on the Mac keyboard?

aestriplex:
[...]

PR #1624, opened 2025-02-26
darwin: Del key should remove meters (#1601)

aestriplex:
I open a PR with the changes discussed in #1601.
It seems to be working (apart from the legend issue which I am implementing, and on which I would like feedback from you).
Just an implementation note. I used to use ncurses years ago, so I don't know how the handling of key bindings has changed recently. However, in the past, in order to make the backspace handling robust, I used to listen on all three of these events:
* KEY_BACKSPACE;
* 127;
* '\b'.

What are your thoughts on this?


aestriplex:
@BenBE I just kept case 127, but I defined it as a constant (`MAC_DEL_KEY`) in `CRT.h` for readability. I saw that there were some other keys defined as constants there (`KEY_WHEELUP` and so on...)

PR #1625, opened 2025-02-26
Fix a bug in function `PCPDynamicScreens_addAvailableColumns`

mugitya03:
The function `Hashtable_get` can return a NULL value. 
In function `PCPDynamicScreens_addAvailableColumns`, there is not null value check for the return value of function `Hashtable_get`. 
However, it checks the pointer `screen` instead. I suspect there is a typo.

Here is the original code:
```c
   PCPDynamicScreen* dynamicScreen = Hashtable_get(screens, key);
   if (!screen)
      return;
```

### DIFF
--- a/pcp/PCPDynamicScreen.c
+++ b/pcp/PCPDynamicScreen.c
@@ -399,7 +399,7 @@ void PCPDynamicScreens_addAvailableColumns(Panel* availableColumns, Hashtable* s
       return;
 
    PCPDynamicScreen* dynamicScreen = Hashtable_get(screens, key);
-   if (!screen)
+   if (!dynamicScreen)
       return;
 
    for (unsigned int j = 0; j < dynamicScreen->totalColumns; j++) {


mugitya03:
Hi, I added another commit: Free pt before the function CommandLine_run returns.

The function `ProcessTable_new` will return a ProcessTable object. 
In function `CommandLine_run`, the pointer pt at line 28 is not freed before the function returns. While pt->super is passed to functions at lines 30 and 89, this doesn't transfer ownership of the entire pt structure. There is no explicit free operation for pt in the function, resulting in a memory leak.
Thus, we delete the pointer `pt` at the end of the function.

BenBE:
The patch for PCP looks fine.

Will have to take a closer look at the CommandLine patch though later.

Having this as two PRs would have been preferable from a code review PoV. But no need to split this now; both seem reasonably trivial to check.

P.S.: Please avoid opening a PR from your main branch. Having a separate branch for PRs allows to work on multiple ones in parallel and still allows you to follow upstream development at reasonable ease.

BenBE:
The dtor order in `CommandLine_run` is off when assuming all resources should be `free`'d in the opposite order they were allocated in. This actually would suggest cleanup to be in the following order: `host`, `settings`, `pt`, `ds`, `dc`, `dm`, ut`.

@fasterit @natoscott : Please take a look at this.

mugitya03:
Yes. It's better to free resources in the opposite order they were allocated. Here I just want to make sure the pointer `pt` is not referenced by other pointers.

natoscott:
@mugitya03 your second commit doesn't compile, so it's definitely not tested yet.  :)

Can you run it under valgrind and see what it reports (i.e. is this really a leak?  its not clear to me because the object system in htop has a ->delete method, which could/should be causing ProcessTable_delete to be called elsewhere).  Last time I ran htop under valgrind it was not reporting a leak here, so I'm curious as to whether this change is indeed necessary.

Your initial (PCP) fix looks good to me too though.  How about moving this second commit to a new PR as its unrelated to the first issue?

natoscott (changes requested):
As discussed elsewhere, second commit here doesn't yet compile and is unclear as to whether it is necessary (post valgrind output showing the issue if possible please).

BenBE:
Taking a closer look at this, the proper dtor order should likely be something like the following:

```diff```
$ git diff
diff --git a/CommandLine.c b/CommandLine.c
index b93a0f96..f6208d84 100644
--- a/CommandLine.c
+++ b/CommandLine.c
@@ -420,13 +420,11 @@ int CommandLine_run(int argc, char** argv) {
          fprintf(stderr, "Can not save configuration to %s: %s\n", settings->filename, strerror(-r));
    }
 
-   Header_delete(header);
-   Machine_delete(host);
-
    ScreenManager_delete(scr);
    MetersPanel_cleanup();
 
-   UsersTable_delete(ut);
+   Header_delete(header);
+   Machine_delete(host);
 
    if (flags.pidMatchList)
       Hashtable_delete(flags.pidMatchList);
@@ -435,9 +433,13 @@ int CommandLine_run(int argc, char** argv) {
 
    /* Delete these last, since they can get accessed in the crash handler */
    Settings_delete(settings);
+   Object_delete(pt);
+
+   DynamicScreens_delete(ds);
    DynamicColumns_delete(dc);
    DynamicMeters_delete(dm);
-   DynamicScreens_delete(ds);
+
+   UsersTable_delete(ut);
 
    return 0;
 }
```

This patch in itself won't work, as `Machine_done` will call `Object_delete(this->processTable);` in `Machine.c:59`. So, the second patch in this PR, will cause a double-free of `pt` (And I can reproduce this crash easily in GDB).

From a code cleanliness PoV it stands to argue, that we have a bit of a mess going on with `host`, `pt`, and `settings`, as `host` has references to both `pt` and `settings`, neither being clear to becoming owned as part of `Machine_populateTablesFromSettings(host, settings, &pt->super);`. I'd thus argue to remove the cleanup from `Machine.c:59` and place it in the code of `CommandLine_run`, where the initial pointers to both `pt` and `settings` still reside.

mugitya03:
Sorry for not thoroughly checking the second commit. I didn't notice that machine_done also frees the memory of processTable. I will separate these two commits and submit the first commit as a standalone PR.

mugitya03:
Oh It seems that you've fixed the first commit. Thanks!

BenBE:
Yes. There was nothing wrong with the first commit, thus I cherry-picked it.

As noted, there's some further cleanup regarding the dtor order that could be done. If you like you can open a new PR with these changes properly tested.

PR #1628, opened 2025-03-03
Build unsupported

BenBE:
This allows to run configure in a way such that the current platform is treated as an unsupported one.

This is mainly intended to allow for testing consistency of the code inside the `unsupported` platform directory.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1137,7 +1137,7 @@ fi
 AC_MSG_RESULT([
   ${PACKAGE_NAME} ${VERSION}
 
-  platform:                  $my_htop_platform
+  platform:                  $my_htop_platform ($host_os)
   os-release file:           $with_os_release
   (Linux) proc directory:    $with_proc
   (Linux) openvz:            $enable_openvz
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -124,6 +124,20 @@ jobs:
     - name: Build
       run: make -k
 
+  build-unsupported-latest-gcc:
+    # build as if we were on an unsupported platform
+    runs-on: ubuntu-22.04
+    steps:
+    - uses: actions/checkout@v4
+    - name: Install Dependencies
+      run: sudo apt-get install --no-install-recommends libncursesw5-dev libtinfo-dev libgpm-dev
+    - name: Bootstrap
+      run: ./autogen.sh
+    - name: Configure
+      run: ./configure --host=`uname -m`-none --enable-werror --enable-unicode || ( cat config.log; exit 1; )
+    - name: Build
+      run: make -k
+
   build-ubuntu-latest-clang-analyzer:
     runs-on: ubuntu-latest
     env:
--- a/unsupported/UnsupportedMachine.c
+++ b/unsupported/UnsupportedMachine.c
@@ -5,6 +5,8 @@ Released under the GNU GPLv2+, see the COPYING file
 in the source distribution for its full text.
 */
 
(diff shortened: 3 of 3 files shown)


configure.ac
```diff
@@ -23,6 +23,17 @@ AC_CONFIG_HEADERS([config.h])
...
 AM_INIT_AUTOMAKE([-Wall std-options subdir-objects])
 
+AC_ARG_ENABLE(
+   [build-as-unsupported],
+   [AS_HELP_STRING(
+      [--enable-build-as-unsupported],
+      [Force build as an unsupported platform])
+   ],
+   [if test "x$enable_build_as_unsupported" = "xyes"; then
+      host_os="unsupported"
```
Explorer09, line 33: I think you did it wrong. You should use the variable`my_htop_platform`. `host_os` is an Autoconf-reserved variable that might imply a cross-compiler should be used.

Explorer09: Just a personal idea: I don't like adding a configure option for something that is meant for developers (porting to new platforms).
I think an easier method is to allow an undocumented override of `my_htop_platform`, and then a builder can run configure like this:

```
export my_htop_platform=unsupported; ./configure --enable-werror --enable-unicode
```

BenBE: Currently there's no easy way to influence `my_htop_platform` without duplicating much of the platform selection logic. And I'd actually also be fine if compiling for the `unsupported` platform was treated/handled as a cross-build.

I think, if we want to keep `host_os` untouched, we could initialize `htop_host_os` as `host_os` and do the override on `htop_host_os`.

I'm open to alternate suggestions. The goal for this PR is mostly to get the code for the `unsupported` platform into the usual CI builds, and using an explicit configure switch seemed like the most straight forward solution to me.

Explorer09: @BenBE The problem with `--enable-build-as-unsupported` is that it conflicts with `--enable-pcp` as both options would override `my_htop_platform`.

Explorer09: @BenBE The problem with `--enable-build-as-unsupported` is that it conflicts with `--enable-pcp` as both options would override `my_htop_platform`.

BenBE: I also noticed, that `./configure --host=none` seems to work equally well …

Explorer09: @BenBE That's perhaps unintentional. You should specify `--host=x86_64-unknown-unsupported_os` instead. That's the "target triplet" format.

BenBE: Nope, that's intentional and correct, as `config.sub` translates single-part information into the full triplet.

Cf. [config.sub](https://git.savannah.gnu.org/cgit/autoconf.git/tree/build-aux/config.sub) for the full implementation …

FWIW:
```console
$ ./configure --host=x86_64-unknown-unsupported_os
checking build system type... x86_64-pc-linux-gnu
checking host system type... Invalid configuration `x86_64-unknown-unsupported_os': OS `unsupported_os' not recognized
configure: error: /bin/bash ./build-aux/config.sub x86_64-unknown-unsupported_os failed
```

A full, arch-independent version could be:
```console
$ ./configure --host=$(uname -m)-unknown-none
```

Explorer09: After a bit of thinking, I now believe your approach `./configure --host=$(uname -m)-none` might be the cleanest one. As long as there is no cross-compiler named `$(uname -m)-none-gcc`, `configure` would be happy to grab the native compiler to do what we've intended.

natoscott (approved):
@BenBE looks good, tested and works well here.

PR #1630, opened 2025-03-04
Handle when there's no CPUMeter to display

### LINKED ISSUE #1629: Segfault with "RightCPUs" meter when CPU count is 1
I just discovered this when trying to run an "unsupported" platform of htop, with the `htoprc` file with the host.

To do this, you need an `htoprc` file with "RightCPUs" enabled (this is usually enabled by default on multi-core machines).

```text
column_meters_1=RightCPUs
column_meter_modes_1=1
```

Then, either build htop with an "unsupported" platform (like #1628), or hack the `Machine_new()` routine in `${platform}Machine.c` so that it reports only one CPU. (See. the `Machine_new()` routine in `UnsupportedMachine.c` to get what I mean.)

Finally build and run `htop` and see it crash.

LLDB shows the crash line is `CPUMeter.c` line 267 `CPUMeterCommonUpdateMode()`

```c
   int h = meters[0]->h;
```

(`meters` is nonnull, but `meters[0]` is `NULL`. Null pointer dereference.)
I don't know how to patch this case as I have no idea what _should_ happen if a meter turns out to display nothing.

BenBE:
When there's no CPU to display for RightCPUMeter, display as blank instead of crashing …

Fixes #1629

/cc @Explorer09

### DIFF
--- a/CPUMeter.c
+++ b/CPUMeter.c
@@ -261,6 +261,10 @@ static void CPUMeterCommonUpdateMode(Meter* this, MeterModeId mode, int ncol) {
    this->mode = mode;
    int start, count;
    AllCPUsMeter_getRange(this, &start, &count);
+   if (!count) {
+      this->h = 1;
+      return;
+   }
    for (int i = 0; i < count; i++) {
       Meter_setMode(meters[i], mode);
    }
--- a/CPUMeter.c
+++ b/CPUMeter.c
@@ -222,7 +222,7 @@ static void AllCPUsMeter_getRange(const Meter* this, int* start, int* count) {
          break;
       case 'R': // Second Half
          *start = (cpus + 1) / 2;
-         *count = cpus / 2;
+         *count = cpus - *start;
          break;
    }
 }
--- a/CPUMeter.c
+++ b/CPUMeter.c
@@ -208,8 +208,7 @@ static void CPUMeter_display(const Object* cast, RichString* out) {
 }
 
 static void AllCPUsMeter_getRange(const Meter* this, int* start, int* count) {
-   const CPUMeterData* data = this->meterData;
-   unsigned int cpus = data->cpus;
+   unsigned int cpus = this->host->existingCPUs;
    switch (Meter_name(this)[0]) {
       default:
       case 'A': // All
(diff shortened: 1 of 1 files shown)


Explorer09:
My question was whether the `h` value for a meter is allowed to be zero, otherwise I have no problem with this simple fix.

I also found a small issue that could be addressed in another day:
In the `CPUMeterCommonInit()` code, the `this->meterData->meters` array is allocated with the number of `cpus` as size. But since we know that "LeftCPUs" and "RightCPUs" meters only use half of the array, the `xCalloc(cpus, sizeof(Meter*))` allocation slightly wastes memory.

That's it.

BenBE:
Proposed fix for the over-allocation:

```diff
$ git diff
diff --git a/CPUMeter.c b/CPUMeter.c
index cbfffe74..34d30dc7 100644
--- a/CPUMeter.c
+++ b/CPUMeter.c
@@ -208,8 +208,7 @@ static void CPUMeter_display(const Object* cast, RichString* out) {
 }
 
 static void AllCPUsMeter_getRange(const Meter* this, int* start, int* count) {
-   const CPUMeterData* data = this->meterData;
-   unsigned int cpus = data->cpus;
+   unsigned int cpus = this->host->existingCPUs;
    switch (Meter_name(this)[0]) {
       default:
       case 'A': // All
@@ -237,16 +236,17 @@ static void AllCPUsMeter_updateValues(Meter* this) {
 }
 
 static void CPUMeterCommonInit(Meter* this) {
-   unsigned int cpus = this->host->existingCPUs;
+   int start, count;
+   AllCPUsMeter_getRange(this, &start, &count);
+
    CPUMeterData* data = this->meterData;
    if (!data) {
       data = this->meterData = xMalloc(sizeof(CPUMeterData));
-      data->cpus = cpus;
-      data->meters = xCalloc(cpus, sizeof(Meter*));
+      data->cpus = this->host->existingCPUs;
+      data->meters = xCalloc(count, sizeof(Meter*));
    }
+
    Meter** meters = data->meters;
-   int start, count;
-   AllCPUsMeter_getRange(this, &start, &count);
    for (int i = 0; i < count; i++) {
       if (!meters[i])
          meters[i] = Meter_new(this->host, start + i + 1, (const MeterClass*) Class(CPUMeter));
```

BenBE:
> My question was whether the `h` value for a meter is allowed to be zero, otherwise I have no problem with this simple fix.

Technically yes. I'm not sure if there is any code that checks for this. But having a minimum size of one row for a meter at least gives visual feedback that there's something present. So I'd rather not make this zero height when empty.

Explorer09:
> Proposed fix for the over-allocation:
> 
> ```diff
>        data = this->meterData = xMalloc(sizeof(CPUMeterData));
> -      data->cpus = cpus;
> -      data->meters = xCalloc(cpus, sizeof(Meter*));
> +      data->cpus = this->host->existingCPUs;
> +      data->meters = xCalloc(count, sizeof(Meter*));
> ```

```c
   data->meters = count > 0 ? xCalloc(count, sizeof(Meter*)) : NULL;
```

fasterit (approved):
risky, exactly what we need to make 3.4.0 less boring :-)

PR #1635, opened 2025-03-12
CI: Update LLVM repo and use ubuntu-latest for building "unsupported" htop

Explorer09:


.github/workflows/ci.yml
```diff
@@ -36,8 +36,9 @@ jobs:
     - uses: actions/checkout@v4
     - name: install clang repo
       run: |
+        ubuntu_codename=`sed 's/^ *UBUNTU_CODENAME= *\([^ ]*\).*$/\1/p; d;' /etc/os-release`
```
BenBE, line 39: Restrict the codename regexp to allow alphanumeric characters and hyphen only. Same in the other locations.

BenBE (changes requested):

.github/workflows/ci.yml
Explorer09: I thought the same thing just after I made the PR. The "everything but whitespace" set might be too loose. I narrowed the set to only lowercase alphabets.

.github/workflows/ci.yml
```diff
@@ -36,8 +36,9 @@ jobs:
     - uses: actions/checkout@v4
     - name: install clang repo
       run: |
+        ubuntu_codename=`LC_ALL=C sed 's/^ *UBUNTU_CODENAME= *\([a-z]*\).*$/\1/p; d' /etc/os-release`
```
BenBE, line 39: Why match only whitespace before, but any character after? Also why no whitespace before the equals sign?
```suggestion
        ubuntu_codename=`LC_ALL=C sed 's/^ *UBUNTU_CODENAME *= *\([a-z]*\) *$/\1/p; d' /etc/os-release`
```

Explorer09: > Why match only whitespace before, but any character after? Also why no whitespace before the equals sign?

The "any character after" is to discard any potential comment text there, `# like this`. It's not an error.

I corrected the whitespace before the equal sign, though.

BenBE: I this case be explicit about optionally matching a comment, like ` *\(#.*\)?$` (untested) …

BenBE: Better fail to detect the codename in a strangely formatted os-release file, than to execute arbitrary code on a CI system … ;-)

Explorer09: There is no arbitrary code execution at all. :) I could have just used the `source` (or `.`) command on the `os-release` file (as many lazy shell scripts do) and let arbitrary commands to be injected in. This regex has been stricter than the original `os-release` format already. For example, it doesn't recognise quoted value (`UBUNTU_CODENAME="noble"`).

BenBE: ACK. And I'm fine with being strict here.

And my comment should be read as: I'd rather have the CI fail to detect the version here than that there is any chance something strange in this place allows to execute some code to execute unintentionally.

BenBE (approved):
LGTM.

Any objections to LLVM 20?

Explorer09:
> LGTM.
> 
> Any objections to LLVM 20?

LLVM-20 would pop out a lot of `-Wshorten64-to-32` warnings (#1513), and it is fine to upgrade to that as long as the warning is disabled (`-Wno-shorten64-to-32`) temporarily.

BenBE:
Let's do this after #1513 has been addressed …

PR #1637, opened 2025-03-13
Use pkg-config to discover libnl-3

fpletz:
Since #1427, the location of libnl-3 is not discovered anymore with pkg-config. Instead, the hardcoded path `/usr/include/libnl3` was introduced for headers without any possibility to override it except by patching.

This PR adds support for pkg-config to discover the location of the headers. Additionally, it discovers the path to libnl-3 libraries and modifies the `dlopen` calls to load the configured libraries from their fully qualified path.

Also note the [discussion on the pull request](https://github.com/htop-dev/htop/pull/1427#discussion_r1991155580).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -961,14 +961,16 @@ case "$enable_delayacct" in
          enable_delayacct=no
       else
          old_CFLAGS="$CFLAGS"
-         CFLAGS="$CFLAGS -I/usr/include/libnl3"
+         htop_config_cflags=`$PKG_CONFIG --cflags libnl-3.0 2>/dev/null` || continue
+         CFLAGS="$CFLAGS $htop_config_cflags"
          AC_CHECK_HEADERS([netlink/attr.h netlink/handlers.h netlink/msg.h], [enable_delayacct=yes], [enable_delayacct=no])
          CFLAGS="$old_CFLAGS"
       fi
       ;;
    yes)
       old_CFLAGS="$CFLAGS"
-      CFLAGS="$CFLAGS -I/usr/include/libnl3"
+      htop_config_cflags=`$PKG_CONFIG --cflags libnl-3.0 2>/dev/null` || continue
+      CFLAGS="$CFLAGS $htop_config_cflags"
       AC_CHECK_HEADERS([netlink/attr.h netlink/handlers.h netlink/msg.h], [], [AC_MSG_ERROR([can not find required header files netlink/attr.h, netlink/handlers.h, netlink/msg.h])])
       CFLAGS="$old_CFLAGS"
       ;;
@@ -978,7 +980,8 @@ case "$enable_delayacct" in
 esac
 if test "$enable_delayacct" = yes; then
   AC_DEFINE([HAVE_DELAYACCT], [1], [Define if delay accounting support should be enabled.])
-  AM_CFLAGS="$AM_CFLAGS -I/usr/include/libnl3"
+  htop_config_cflags=`$PKG_CONFIG --cflags libnl-3.0 2>/dev/null`
+  AM_CFLAGS="$AM_CFLAGS $htop_config_cflags"
 fi
 AM_CONDITIONAL([HAVE_DELAYACCT], [test "$enable_delayacct" = yes])
 
--- a/configure.ac
+++ b/configure.ac
@@ -982,6 +982,8 @@ if test "$enable_delayacct" = yes; then
   AC_DEFINE([HAVE_DELAYACCT], [1], [Define if delay accounting support should be enabled.])
   htop_config_cflags=`$PKG_CONFIG --cflags libnl-3.0 2>/dev/null`
   AM_CFLAGS="$AM_CFLAGS $htop_config_cflags"
+  libdir_libnl=`$PKG_CONFIG --variable=libdir libnl-3.0 2>/dev/null`
+  AC_DEFINE_UNQUOTED([LIBDIR_LIBNL], ["$libdir_libnl"], [Path to the libnl-3 libraries])
 fi
 AM_CONDITIONAL([HAVE_DELAYACCT], [test "$enable_delayacct" = yes])
 
--- a/linux/LibNl.c
+++ b/linux/LibNl.c
@@ -78,17 +78,17 @@ static int load_libnl(void) {
    if (libnlHandle && libnlGenlHandle)
       return 0;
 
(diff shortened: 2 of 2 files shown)


configure.ac
```diff
@@ -961,14 +961,16 @@ case "$enable_delayacct" in
          enable_delayacct=no
       else
          old_CFLAGS="$CFLAGS"
-         CFLAGS="$CFLAGS -I/usr/include/libnl3"
+         htop_config_cflags=`$PKG_CONFIG --cflags libnl-3.0 2>/dev/null` || continue
```
Explorer09, line 964: I wish some quality improvements to the check:
1. There's no error handling here. What would you do if the `pkg-config` call fails?
2. Is there any reason not to use `PKG_CHECK_MODULES` macro as in the old code?
3. You should allow the user to override the compiler flags detected. If you have used `PKG_CHECK_MODULES`, then the variable `LIBNL3_CFLAGS` would be defined automatically to allow that user override. Otherwise, maybe you can define it yourself? (`AC_ARG_VAR([LIBNL3_CFLAGS], [C compiler flags for libnl3, overriding pkg-config])`)

BenBE, line 964: AFAICS, this should be `|| true`, as there is no loop.
```suggestion
         htop_config_cflags=`$PKG_CONFIG --cflags libnl-3.0 2>/dev/null` || true
```

fpletz: > There's no error handling here. What would you do if the `pkg-config` call fails?

In this case the `AC_CHECK_HEADERS` fails like it would have done before since `htop_config_cflags` is empty and the headers aren't found. But `PKG_CHECK_MODULES` should take care of that.

> Is there any reason not to use `PKG_CHECK_MODULES` macro as in the old code?

I used the current style since that is used to discover `ncurses` in the same file. I agree that `PKG_CHECK_MODULES` is the better solution and will change the code accordingly.

> You should allow the user to override the compiler flags detected. If you have used `PKG_CHECK_MODULES`, then the variable `LIBNL3_CFLAGS` would be defined automatically to allow that user override. Otherwise, maybe you can define it yourself? (`AC_ARG_VAR([LIBNL3_CFLAGS], [C compiler flags for libnl3, overriding pkg-config])`)

Yes, makes sense. I will add this. And also introduce the variable `LIBNL3_LIBDIR`.

configure.ac
```diff
@@ -961,14 +961,16 @@ case "$enable_delayacct" in
...
+         CFLAGS="$CFLAGS $htop_config_cflags"
          AC_CHECK_HEADERS([netlink/attr.h netlink/handlers.h netlink/msg.h], [enable_delayacct=yes], [enable_delayacct=no])
          CFLAGS="$old_CFLAGS"
       fi
       ;;
    yes)
       old_CFLAGS="$CFLAGS"
-      CFLAGS="$CFLAGS -I/usr/include/libnl3"
+      htop_config_cflags=`$PKG_CONFIG --cflags libnl-3.0 2>/dev/null` || continue
```
BenBE, line 972: Same here …
```suggestion
      htop_config_cflags=`$PKG_CONFIG --cflags libnl-3.0 2>/dev/null` || true
```

configure.ac
Explorer09: The reason the `ncurses` detection code doesn't use `PKG_CHECK_MODULES` is that `ncurses` may come with many different library names (as well as `.pc` file names), so the code has to implement fallbacks.

configure.ac
```diff
@@ -978,7 +980,10 @@ case "$enable_delayacct" in
 esac
 if test "$enable_delayacct" = yes; then
   AC_DEFINE([HAVE_DELAYACCT], [1], [Define if delay accounting support should be enabled.])
-  AM_CFLAGS="$AM_CFLAGS -I/usr/include/libnl3"
+  htop_config_cflags=`$PKG_CONFIG --cflags libnl-3.0 2>/dev/null`
+  AM_CFLAGS="$AM_CFLAGS $htop_config_cflags"
+  libdir_libnl=`$PKG_CONFIG --variable=libdir libnl-3.0 2>/dev/null`
```
BenBE, line 985: Failure to read these should result in a proper error message generated, or a warning notifying of a fallback to `/usr/include` being used.

BenBE (changes requested):
Please squash/fixup your changes into the appropriate commits.

configure.ac
```diff
@@ -982,6 +982,8 @@ if test "$enable_delayacct" = yes; then
   AC_DEFINE([HAVE_DELAYACCT], [1], [Define if delay accounting support should be enabled.])
   htop_config_cflags=`$PKG_CONFIG --cflags libnl-3.0 2>/dev/null`
   AM_CFLAGS="$AM_CFLAGS $htop_config_cflags"
+  libdir_libnl=`$PKG_CONFIG --variable=libdir libnl-3.0 2>/dev/null`
+  AC_DEFINE_UNQUOTED([LIBDIR_LIBNL], ["$libdir_libnl"], [Path to the libnl-3 libraries])
```
Explorer09, line 986: Um. A small question (maybe this is bike shedding).
Which name for the shell variable / C macro is better, `LIBNL_LIBDIR` or `LIBDIR_LIBNL`?

I would vote for `LIBNL_LIBDIR`.

BenBE: Other places like for hwloc use `FOO_LIBS`, thus `LIBNL_LIBDIR` would be reasonable.

linux/LibNl.c
```diff
@@ -78,17 +78,17 @@ static int load_libnl(void) {
    if (libnlHandle && libnlGenlHandle)
       return 0;
 
-   libnlHandle = dlopen("libnl-3.so", RTLD_LAZY);
+   libnlHandle = dlopen(LIBDIR_LIBNL "/libnl-3.so", RTLD_LAZY);
```
Explorer09, line 81: I noticed a slash is used here unconditionally. Normally for libraries it's better to load them from default search paths rather than forcing a directory prefix at compile time. I know the directory is grabbed from `pkg-config` with this patch, but what if no directory prefix is specified there?

Explorer09:
@fpletz I've adapted some part of your patch to the new PR, #1698. Would you please help test it out to see whether it works? It can supersede this PR.

BenBE:
@fpletz @Explorer09 Given #1698 is merged, is there anything open with this PR?

fasterit:
Superseded by #1698, closing

PR #1641, opened 2025-03-17
Support for stuck processes on macOS

aestriplex:
This is the last part of the feature implemented by #1617. In particular, this commit adds the support for stuck process.
The information about the state at a thread level (`pth_run_state`) is available only for root, so you can see the 'D' only if you run htop with sudo, otherwise, you just see a generic sleeping process (state 'S').

### DIFF
--- a/darwin/DarwinProcess.c
+++ b/darwin/DarwinProcess.c
@@ -440,6 +440,7 @@ void DarwinProcess_scanThreads(DarwinProcess* dp, DarwinProcessTable* dpt) {
    }
 
    const bool hideUserlandThreads = dpt->super.super.host->settings->hideUserlandThreads;
+   bool isProcessStuck = false;
 
    for (mach_msg_type_number_t i = 0; i < thread_count; i++) {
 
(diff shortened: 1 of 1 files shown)


darwin/DarwinProcess.c
```diff
@@ -490,6 +491,8 @@ void DarwinProcess_scanThreads(DarwinProcess* dp, DarwinProcessTable* dpt) {
       tdproc->super.time        = (extended_info.pth_system_time + extended_info.pth_user_time) / 10000000;
       tdproc->super.priority    = extended_info.pth_curpri;
 
+      isProcessStuck |= (extended_info.pth_run_state == TH_STATE_UNINTERRUPTIBLE);
```
BenBE, line 494:
```suggestion
      isProcessStuck |= extended_info.pth_run_state == TH_STATE_UNINTERRUPTIBLE;
```

darwin/DarwinProcess.c
```diff
@@ -498,6 +501,10 @@ void DarwinProcess_scanThreads(DarwinProcess* dp, DarwinProcessTable* dpt) {
          ProcessTable_add(&dpt->super, tprocess);
    }
 
+   if (isProcessStuck) {
+      dp->super.state = UNINTERRUPTIBLE_WAIT;
```
BenBE, line 505: This should be pushed into the thread-specific state for the affected `tid` too.

As is, this is only for the overall process.

BenBE (changes requested):

fasterit:
Certified as "compiles and looks good :smile:"

PR #1642, opened 2025-03-18
Unicode character support in screen tab names

Explorer09:
I would like to share a feature that I have done as part of a code refactoring process.

These changes are still work in progress, but I think it's good to file it and gather early feedback.

This feature makes the screen tab names display with the correct terminal widths when they contain Unicode characters (in UTF-8 encoding). It replaces the naive `strlen` calculation with a new function that calculates the display terminal width of a name string.

The PR introduces the following functions:
* `String_makePrintable`, a sanitization function with an API similar to`xStrndup`. It reads the string, converts all unprintable characters to replacement code points (U+FFFD), and allocates and re-encodes the string.
* `EncodePrintableString`, an internal routine for `String_makePrintable`. This routine is designed with room to support `RichString` buffers
* `String_mbswidth`, a function that calculates the terminal width of a string, or calculate the byte length of a substring if it is clipped by a terminal width limit.

This PR  is related to #1513.

### DIFF
--- a/Action.c
+++ b/Action.c
@@ -432,7 +432,8 @@ Htop_Reaction Action_setScreenTab(State* st, int x) {
    int rem = x - SCREEN_TAB_MARGIN_LEFT;
    for (unsigned int i = 0; i < settings->nScreens; i++) {
       const char* tab = settings->screens[i]->heading;
-      int width = rem >= bracketWidth ? (int)strnlen(tab, rem - bracketWidth + 1) : 0;
+      const char* ptr = tab;
+      int width = rem >= bracketWidth ? String_mbswidth(&ptr, SIZE_MAX, rem - bracketWidth + 1) : 0;
       if (width >= rem - bracketWidth + 1) {
          settings->ssIndex = i;
          Htop_Reaction reaction = HTOP_UPDATE_PANELHDR | HTOP_REFRESH | HTOP_REDRAW_BAR;
--- a/XUtils.c
+++ b/XUtils.c
@@ -10,6 +10,7 @@ in the source distribution for its full text.
 #include "XUtils.h"
 
 #include <assert.h>
+#include <ctype.h> // IWYU pragma: keep
 #include <errno.h>
 #include <fcntl.h>
 #include <limits.h>
@@ -258,6 +259,298 @@ size_t strnlen(const char* str, size_t maxLen) {
    return maxLen;
 }
 #endif
+
+#ifdef HAVE_LIBNCURSESW
+static void String_encodeWChar(WCharEncoderState* ps, wchar_t wc) {
+   assert(!ps->buf || ps->pos < ps->size);
+
+   char tempBuf[MB_LEN_MAX];
+
+   // This function will null terminate the string only upon a call
+   // with (wc == 0). It might take more than a single NUL byte to
+   // terminate a string when using the C multibyte functions and a
+   // non-Unicode encoding, thus this function won't support truncation
+   // of a string. The caller must provide the right size in ps->size
+   // if ps->buf is not NULL.
+   size_t len = wcrtomb(tempBuf, wc, &ps->mbState);
+   assert(len != 0);
+   if (len == (size_t)-1) {
+      assert(len != (size_t)-1);
+      fail();
+   }
+   if (ps->buf) {
+      if (len > ps->size - ps->pos) {
+         fail();
+      }
+      memcpy((char*)ps->buf + ps->pos, tempBuf, len);
+   }
+   ps->pos += len;
+}
+#else
+static void String_encodeWChar(WCharEncoderState* ps, int c) {
+   assert(!ps->buf || ps->pos < ps->size);
+
+   char* buf = ps->buf;
+   if (buf)
+      buf[ps->pos] = (char)c;
+
+   ps->pos += 1;
+}
+#endif
+
+void EncodePrintableString(WCharEncoderState* ps, const char* src, size_t maxLen, EncodeWChar encodeWChar) {
+   assert(src || maxLen == 0);
+
+   size_t pos = 0;
+   bool wasReplaced = false;
+
+#ifdef HAVE_LIBNCURSESW
+   const wchar_t replacementChar = CRT_utf8 ? L'\xFFFD' : L'?';
+   wchar_t ch;
+
+   mbstate_t decState = {0};
+#else
+   const char replacementChar = '?';
+   char ch;
+#endif
+
+   do {
+      size_t len = 0;
+      bool shouldReplace = false;
+      ch = 0;
+
+      if (pos < maxLen) {
+         // Read the next character from the byte sequence
+#ifdef HAVE_LIBNCURSESW
+         mbstate_t newState;
+         memcpy(&newState, &decState, sizeof(newState));
+         len = mbrtowc(&ch, &src[pos], maxLen - pos, &newState);
+
+         assert(len != 0 || ch == 0);
+         switch (len) {
+         case (size_t)-2:
+            errno = EILSEQ;
+            shouldReplace = true;
+            len = maxLen - pos;
+            break;
+
+         case (size_t)-1:
+            shouldReplace = true;
+            len = 1;
+            break;
+
+         default:
+            memcpy(&decState, &newState, sizeof(decState));
+         }
+#else
+         len = 1;
+         ch = src[pos];
+#endif
+      }
+
+      pos += len;
+
+      // Filter unprintable characters
+      if (!shouldReplace && ch != 0) {
+#ifdef HAVE_LIBNCURSESW
+         shouldReplace = !iswprint(ch);
+#else
+         shouldReplace = !isprint((unsigned char)ch);
+#endif
+      }
+
+      if (shouldReplace) {
+         ch = replacementChar;
+         if (wasReplaced)
+            continue;
+      }
+      wasReplaced = shouldReplace;
+
+      encodeWChar(ps, ch);
+   } while (ch != 0);
+}
+
+char* String_makePrintable(const char* str, size_t maxLen) {
+   WCharEncoderState encState = {0};
+
+   EncodePrintableString(&encState, str, maxLen, String_encodeWChar);
+   size_t size = encState.pos;
+   assert(size > 0);
+
+   memset(&encState, 0, sizeof(encState));
+   char* buf = xMalloc(size);
+   encState.size = size;
+   encState.buf = buf;
+   EncodePrintableString(&encState, str, maxLen, String_encodeWChar);
+   assert(encState.pos == size);
+
+   return buf;
+}
+
+bool MBStringDecoder_nextWChar(MBStringDecoder* decoder) {
+   if (!decoder->str || decoder->maxLen == 0)
+      return false;
+
+   // If the previous call of this function encounters an invalid sequence,
+   // do not continue (because the "mbState" object for mbrtowc() is
+   // undefined). The caller is supposed to reset the state.
+#ifdef HAVE_LIBNCURSESW
+   if (decoder->ch == WEOF)
+      return false;
+#endif
+
+#ifdef HAVE_LIBNCURSESW
+   wchar_t ch;
+   size_t len = mbrtowc(&ch, decoder->str, decoder->maxLen, &decoder->mbState);
+
+   // These assertions ensure the mbrtowc() implementation is correct
+   assert(len == 0 || len >= (size_t)-2 || ch != 0);
+   assert(len != 0 || ch == 0);
+
+   switch (len) {
+   case (size_t)-1:
+      // Invalid sequence. decoder->str remains at the position where
+      // the first byte of the invalid sequence is found.
+      decoder->ch = WEOF;
+      return false;
+
+   case (size_t)-2:
+      // Incomplete sequence
+      decoder->str += decoder->maxLen;
+      decoder->maxLen = 0;
+      return false;
+
+   case 0:
+      // End of string. This assignment is an optimization hint.
+      ch = 0;
+   }
+#else
+   char ch = *decoder->str;
+   const size_t len = 1;
+#endif
+
+   if (ch == 0) {
+      // Setting "str" to NULL prevents subsequent calls from reading
+      // out of bounds.
+      decoder->str = NULL;
+      decoder->maxLen = 0;
+   } else {
+      assert(decoder->maxLen >= len);
+      decoder->str += len;
+      decoder->maxLen -= len;
+   }
+   decoder->ch = ch;
+   return true;
+}
+
+int String_lineBreakWidth(const char** str, size_t maxLen, int maxWidth, char separator) {
+   assert(*str || maxLen == 0);
+
+   // The caller should ensure (maxWidth >= 0).
+   // It's possible for a Unicode string to occupy 0 terminal columns, so this
+   // function allows (maxWidth == 0).
+   if (maxWidth < 0)
+      maxWidth = INT_MAX;
+
+#ifdef HAVE_LIBNCURSESW
+   // If the character takes zero columns, include the character in the
+   // substring if the working encoding is UTF-8, and ignore it otherwise.
+   // In Unicode, combining characters are always placed after the base
+   // character, but some legacy 8-bit encodings instead place combining
+   // characters before the base character.
+   const bool isUnicode = CRT_utf8;
+#else
+   const bool isUnicode = false;
+#endif
+
+   int totalWidth = 0;
+
+   MBStringDecoder decoder = {0};
+   decoder.str = *str;
+   decoder.maxLen = maxLen;
+
+   bool inSpaces = true;
+   const char* breakPos = NULL;
+   int breakWidth = 0;
+
+   while (totalWidth < maxWidth || isUnicode) {
+      assert(totalWidth <= maxWidth);
+
+      if (!MBStringDecoder_nextWChar(&decoder))
+         break;
+      if (decoder.ch == 0)
+         break;
+
+      if (decoder.ch == ' ' && separator == ' ' && !inSpaces) {
+         inSpaces = true;
+         breakPos = *str;
+         breakWidth = totalWidth;
+      }
+
+#ifdef HAVE_LIBNCURSESW
+      int cw = wcwidth((wchar_t)decoder.ch);
+      if (cw < 0) {
+         // This function should not be used with string containing unprintable
+         // characters. Tolerate them on release build, however.
+         assert(cw >= 0);
+         break;
+      }
+#else
+      assert(isprint(decoder.ch));
+      const int cw = 1;
+#endif
+
+      if (cw > maxWidth - totalWidth) {
+         // This character cannot fit the line with the given maxWidth.
+         if (breakPos) {
+            // Rewind the scanning state to the last found separator.
+            totalWidth = breakWidth;
+            *str = breakPos;
+         }
+         break;
+      }
+
+      if (cw <= 0 && !isUnicode)
+         continue;
+
+      totalWidth += cw;
+
+      // (*str - start) will represent the length of the substring bounded
+      // by the width limit.
+      *str = decoder.str;
+
+      if (decoder.ch != ' ')
+         inSpaces = false;
+
+#ifdef HAVE_LIBNCURSESW
+      bool isSeparator = decoder.ch == (wint_t)separator;
+#else
+      bool isSeparator = decoder.ch == (int)separator;
+#endif
+      if (isSeparator && separator != ' ') {
+         breakPos = *str;
+         breakWidth = totalWidth;
+      }
+   }
+
+   return totalWidth;
+}
+
+int String_mbswidth(const char** str, size_t maxLen, int maxWidth) {
+#ifdef HAVE_LIBNCURSESW
+   return String_lineBreakWidth(str, maxLen, maxWidth, '\0');
+#else
+   assert(*str || maxLen == 0);
+
+   if (maxWidth < 0)
+      maxWidth = INT_MAX;
+
+   maxLen = MINIMUM((size_t)maxWidth, maxLen);
+   size_t len = strnlen(*str, maxLen);
+   *str += len;
+   return (int)len;
+#endif
+}
 
 int xAsprintf(char** strp, const char* fmt, ...) {
    *strp = NULL;
--- a/XUtils.h
+++ b/XUtils.h
@@ -22,7 +22,36 @@ in the source distribution for its full text.
 #include <string.h> // IWYU pragma: keep
 
 #include "Macros.h"
+#include "ProvideCurses.h"
 
+
+typedef struct WCharEncoderState_ {
+   size_t pos;
+   size_t size;
+   void* buf;
+   mbstate_t mbState;
+} WCharEncoderState;
+
+/* Object for reading wide characters from a multibyte string.
+   "str" and "maxLen" are input but will be modified during process.
+   "str" will be set to NULL when the decoding is finished with the
+   terminating L'\0' character. */
+typedef struct MBStringDecoder_ {
+   const char* str;
+   size_t maxLen;
+#ifdef HAVE_LIBNCURSESW
+   wint_t ch;
+   mbstate_t mbState;
+#else
+   int ch;
+#endif
+} MBStringDecoder;
+
+#ifdef HAVE_LIBNCURSESW
+typedef ATTR_NONNULL void (*EncodeWChar)(WCharEncoderState* ps, wchar_t wc);
+#else
+typedef ATTR_NONNULL void (*EncodeWChar)(WCharEncoderState* ps, int c);
+#endif
 
 ATTR_NORETURN
 void fail(void);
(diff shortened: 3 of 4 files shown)


XUtils.c
```diff
@@ -224,6 +232,209 @@ size_t String_safeStrncpy(char* restrict dest, const char* restrict src, size_t
...
+
+      if (shouldReplace) {
+         wc = replacementChar;
+         if (wasReplaced) {
+            continue;
+         }
+      }
+      wasReplaced = shouldReplace;
+
+      destSize += encodeWChar((dest ? (char*)dest + destSize : NULL), wc, encState);
```
BenBE, line 318: Treat the `destSize` as opaque, similar to how it is done in the CGroup compression code. The location needs to be part of the passed state, alongside the maximum buffer size. This avoids any pointer arithmetic on `dest` here and improves re-usability with other memory layouts.

XUtils.c
```diff
@@ -224,6 +232,209 @@ size_t String_safeStrncpy(char* restrict dest, const char* restrict src, size_t
...
+#endif
+
+size_t EncodePrintableString(void* dest, const char* src, size_t srcLen, EncodeWChar encodeWChar, void* encState) {
+   assert(src || srcLen == 0);
+
+   size_t srcPos = 0;
+   size_t destSize = 0;
+   bool wasReplaced = false;
+
+#ifdef HAVE_LIBNCURSESW
```
BenBE, line 264: Any chance we could split this into two smaller blocks? Currently there is much code duplication.

XUtils.c
```diff
@@ -224,6 +232,209 @@ size_t String_safeStrncpy(char* restrict dest, const char* restrict src, size_t
...
+
+   const wchar_t replacementChar = CRT_utf8 ? L'\xFFFD' : L'?';
+
+   mbstate_t decState;
+   memset(&decState, 0, sizeof(decState));
+
+   mbstate_t newState;
+   wchar_t wc;
+   do {
+      wc = L'\0';
```
BenBE, line 274:
```suggestion
      wc = 0;
```

This makes `c` and `wc` assignments compatible and thus would not require distinction in the code paths …

XUtils.c
```diff
@@ -224,6 +232,209 @@ size_t String_safeStrncpy(char* restrict dest, const char* restrict src, size_t
...
+         }
+      }
+      wasReplaced = shouldReplace;
+
+      destSize += encodeWChar((dest ? (char*)dest + destSize : NULL), c, encState);
+   } while (c != '\0');
+
+#endif /* HAVE_LIBNCURSESW */
+
+   assert(destSize > 0);
```
BenBE, line 351: We shouldn't care on this level about `destSize`. All we should return is the overall success of the conversion. Cf. CGroup compression code.

fasterit:
Fails CI:
```
XUtils.c:426:6: error: 'HAVE_STRNLEN' is not defined, evaluates to 0 [-Werror,-Wundef]
  426 | # if HAVE_STRNLEN
      |      ^
1 error generated.``

File: XUtils.c (hunk 2 above)
Comment (Explorer09) [reply XUtils.c:264]: I wished to do that, but the problem is on the data type. One is `char` and the other is `wchar_t`. How do I resolve this? (Introduce something like a `TCHAR` type in Windows programming?)

**Comment by Explorer09** [comment]
> Fails CI:
> 
> ```
> XUtils.c:426:6: error: 'HAVE_STRNLEN' is not defined, evaluates to 0 [-Werror,-Wundef]
>   426 | # if HAVE_STRNLEN
>       |      ^
> 1 error generated.``
> ```

Noted. It should have been `#ifdef`.

File: XUtils.c (hunk 4 above)
Comment (Explorer09) [reply XUtils.c:351]: This function, `EncodePrintableString`, is intended to be called twice. One for counting the size needed and the second time for doing the actual conversion. There is no buffer expansion (a la `realloc`) in the middle of the process.

`EncodePrintableString` will return the size, and then the caller will allocate the buffer, and call the same function again to do the conversion.

File: XUtils.c [hunk 5]
```diff
@@ -22,6 +23,13 @@ in the source distribution for its full text.
 #include "CRT.h"
 #include "Macros.h"
 
+#ifdef HAVE_LIBNCURSESW
```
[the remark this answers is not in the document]
Explorer09: The htop codebase doesn't support any curses**w** implementation other than `ncursesw` for now.

You are right about that I can simply include `ProvideCurses.h` instead.

XUtils.c
Explorer09: I'm reserved about this. `destSize` is required to be returned to the caller so that the caller can `malloc` the buffer that just fits. It would not be that "opaque" as you think.

It might be a good idea for `EncodePrintableString` to not care about the `destSize` at all and have the property and the counting all done by the callee (`encodeWChar`). The caller of  `EncodePrintableString` would have to extract the  `destSize` property from `encState` nevertheless.

BenBE:
> > Fails CI:
> > ```
> > XUtils.c:426:6: error: 'HAVE_STRNLEN' is not defined, evaluates to 0 [-Werror,-Wundef]
> >   426 | # if HAVE_STRNLEN
> >       |      ^
> > 1 error generated.``
> > ```
> 
> Noted. It should have been `#ifdef`.

IDK, but nowhere else do we check for `strnlen`, so why here all of a sudden?

XUtils.c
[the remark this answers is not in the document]
BenBE: The idea of `ProvideCurses.h` basically is to provide you with curses¹. So anything related there should follow along to minimize churn from messing with all the flags involved with it.

I even thought about having the `character type` like `wchar_t` be a define, where differences between enabled and disabled Unicode support arise. Though I dropped that idea …

¹pun intended …

BenBE:
Any objections to pull in 4ccfc38335aaaaad3126844b7046e310a7cf6d59 to `main` now and do the Unicode support later? That commit is mostly cleanup and doesn't affect much of the later work in this PR.

BenBE:
> > Any objections to pull in [4ccfc38](https://github.com/htop-dev/htop/commit/4ccfc38335aaaaad3126844b7046e310a7cf6d59) to `main` now and do the Unicode support later? That commit is mostly cleanup and doesn't affect much of the later work in this PR.
> 
> I have just one question with the commit: Is it okay for the constants to be defined in `CRT.h` or is there a better header to put those two constant names?

`CRT.h` seems to be fine. `ScreenManager.h` may be another place to put them, but actually I'm not too sure, where's best …

> BTW, I borrowed the "margin left" and "column gap" naming from CSS. 🙂

Fine.

XUtils.c
```diff
@@ -224,6 +226,185 @@ size_t String_safeStrncpy(char* restrict dest, const char* restrict src, size_t
...
+   size_t pos = 0;
+   bool wasReplaced = false;
+
+#ifdef HAVE_LIBNCURSESW
+   const wchar_t replacementChar = CRT_utf8 ? L'\xFFFD' : L'?';
+   wchar_t ch;
+
+   mbstate_t decState;
+   memset(&decState, 0, sizeof(decState));
+   mbstate_t newState;
```
BenBE, line 261: This can be put into the loop; no need to track it beyond loop iterations.

XUtils.c
```diff
@@ -224,6 +226,185 @@ size_t String_safeStrncpy(char* restrict dest, const char* restrict src, size_t
...
+      ch = 0;
+
+      if (pos < maxLen) {
+         // Read the next character from the byte sequence
+#ifdef HAVE_LIBNCURSESW
+         memcpy(&newState, &decState, sizeof(newState));
+         len = mbrtowc(&ch, &src[pos], maxLen - pos, &newState);
+
+         switch (len) {
+
```
BenBE, line 279: No blank line needed.
```suggestion
```

XUtils.h
```diff
@@ -23,8 +23,21 @@ in the source distribution for its full text.
...
 #include "Compat.h"
 #include "Macros.h"
+#include "ProvideCurses.h"
 
 
+typedef struct WCharEncoderState_ {
+   size_t pos;
+   void* buf;
+   mbstate_t mbState;
+} WCharEncoderState;
```
BenBE, line 33: This should also carry the buffer size, so the internal `EncodeWChar` routines can perform bound checking.

Also note, that the CGroup code deliberately uses two distinct functions (one for counting the size, another for actually writing) and that these functions return success (character properly counted/written to buffer) or failure (OOB write). Similar could be done here too.

configure.ac
```diff
@@ -372,6 +372,7 @@ AC_CHECK_FUNCS([ \
     sched_getscheduler \
     sched_setscheduler \
     strchrnul \
+    strnlen \
```
BenBE, line 375: I'm not sure I like this, when we already have plenty of code, that requires this function without prior check …

XUtils.c
```diff
@@ -224,6 +226,185 @@ size_t String_safeStrncpy(char* restrict dest, const char* restrict src, size_t
...
+   assert(len != (size_t)-1);
+   assert(len > 0);
+
+   ps->pos += len;
+}
+#else
+static void String_encodeWChar(WCharEncoderState* ps, int c) {
+   if (ps->buf) {
+      ((char*)ps->buf)[ps->pos] = (char)c;
+   }
```
BenBE, line 244:
```suggestion
   char* buf = ps->buf;
   if (buf) {
      buf[ps->pos] = (char)c;
   }
```

XUtils.c
```diff
@@ -224,6 +226,185 @@ size_t String_safeStrncpy(char* restrict dest, const char* restrict src, size_t
...
+   EncodePrintableString(&encState, str, maxLen, String_encodeWChar);
+   size_t bufSize = encState.pos;
+   assert(bufSize > 0);
+
+   memset(&encState, 0, sizeof(encState));
+   char* buf = xMalloc(bufSize);
+   encState.buf = buf;
+   EncodePrintableString(&encState, str, maxLen, String_encodeWChar);
+   assert(encState.pos == bufSize);
+   encState.buf = NULL;
```
BenBE, line 341: Why the explicit `NULL` assignment?

XUtils.c
```diff
@@ -224,6 +226,185 @@ size_t String_safeStrncpy(char* restrict dest, const char* restrict src, size_t
...
+   memset(&encState, 0, sizeof(encState));
+   EncodePrintableString(&encState, str, maxLen, String_encodeWChar);
+   size_t bufSize = encState.pos;
+   assert(bufSize > 0);
+
+   memset(&encState, 0, sizeof(encState));
+   char* buf = xMalloc(bufSize);
+   encState.buf = buf;
+   EncodePrintableString(&encState, str, maxLen, String_encodeWChar);
+   assert(encState.pos == bufSize);
```
BenBE, line 340: Technically the assertion should be:
```suggestion
   assert(encState.pos <= bufSize);
```

Practically, the encoding of the string itself should match what we counted earlier.

XUtils.c
```diff
@@ -224,6 +226,185 @@ size_t String_safeStrncpy(char* restrict dest, const char* restrict src, size_t
...
+   encState.buf = NULL;
+
+   return buf;
+}
+
+#ifndef HAVE_STRNLEN
+static size_t strnlen(const char* str, size_t maxLen) {
+   size_t len;
+   for (len = 0; len < maxLen && str[len] != '\0'; len++) {}
+   return len;
```
BenBE, line 350:
```suggestion
   for (size_t len = 0; len < maxLen; len++)
      if (!str[len])
         return len;
   return maxLen;
```

XUtils.c
```diff
@@ -224,6 +226,187 @@ size_t String_safeStrncpy(char* restrict dest, const char* restrict src, size_t
...
+   char* dest = ps->buf ? (char*)ps->buf + ps->pos : tempBuf;
+   size_t len = wcrtomb(dest, wc, &ps->mbState);
+   assert(len != (size_t)-1);
+   assert(len > 0);
+
+   ps->pos += len;
+}
+#else
+static void String_encodeWChar(WCharEncoderState* ps, int c) {
+   char* buf = (char*)ps->buf;
```
BenBE, line 242: No need to cast `void*` to `char*` …

configure.ac
Explorer09: A quick `grep -r strnlen` shows no code in htop has been using `strnlen` before this PR.

XUtils.h
Explorer09: I choose to make the size counting and the actual writing the same routine for a reason: I expect the "rules" for filtering characters be tweakable in the future. If there are additional characters besides `!iswprint` that need to be filtered out (replaced with U+FFFD), people only need to modify one routine for that.

I'm no sure about bound checking yet. I can add it into the struct, but it would be for assertions only, as I always expect the buffer allocated would just fit. (That is, no truncation.)

XUtils.c
Explorer09: Need to test if this move has an impact on code size...

XUtils.c
Explorer09: I don't expect oversized allocation here. I intended the size reported be equal. Not more or less.

XUtils.c
Explorer09: I thought Clang's analyzer would be confused, thinking `encState.buf` being the owner of the allocated buffer. It turns out it wasn't confused here and so I removed the line.

XUtils.h
BenBE: That's what "Failure to write to the buffer" could be used for. If the `EncodeWChar` function sees that the buffer is full and not NUL-terminated, it should return `false` (AKA failure), which all callees could propagate accordingly. Similar to what the CGROUP compression code¹ does …

¹Actually that code over-allocates by one byte and forces the NUL-termination in that over-allocated byte. Cf. `linux/CGroupUtils.c` in functions `CGroup_filterName` and `CGroup_filterContainer`.

XUtils.c
Explorer09:
```diff
diff --git a/XUtils.c b/XUtils.c
index f841a104..767411ba 100644
--- a/XUtils.c
+++ b/XUtils.c
@@ -269,7 +269,6 @@ void EncodePrintableString(WCharEncoderState* ps, const char* src, size_t maxLen
 
    mbstate_t decState;
    memset(&decState, 0, sizeof(decState));
-   mbstate_t newState;
 #else
    const char replacementChar = '?';
    char ch;
@@ -283,6 +282,7 @@ void EncodePrintableString(WCharEncoderState* ps, const char* src, size_t maxLen
       if (pos < maxLen) {
          // Read the next character from the byte sequence
 #ifdef HAVE_LIBNCURSESW
+         mbstate_t newState;
          memcpy(&newState, &decState, sizeof(newState));
          len = mbrtowc(&ch, &src[pos], maxLen - pos, &newState);
 
```

Result. With the small change like this,
1. GCC 13.2 (Ubuntu 24.04 "Noble") shows no code size changes.
2. Clang 18 (Ubuntu 22.04 "Jammy"): code size becomes 7 bytes bigger with `-Os` option.
3. Clang 20 (Ubuntu 24.04 "Noble"): code size becomes 27 bytes smaller with `-Os` option.

So it's interesting to see Clang missed this. The code should be equivalent and yet Clang generated the binaries with different sizes.

XUtils.h
Explorer09: @BenBE Nope. After a bit of consideration, this length check cannot be done in `EncodeWChar` like in the CGROUP code. I wrote a comment on why instead.

It's a limitation of the `wcrtomb(3)` interface. When we can detect `wcrtomb` writing out of bounds we might be too late (access violation already happens). `wcrtomb` doesn't give us the "length check before writing" we need. Trying to workaround that would make code unnecessarily complicated (`wcrtomb` called one more time - three times in total).

Also, we cannot put the `'\0'` byte in arbitrary position. Unlike in the CGROUP code, we can only put `'\0'` in character boundaries here. So overall there's nothing I can do.

Explorer09: @BenBE Update: This is what I mean by being "unnecessarily expensive".

```c
static bool String_encodeWChar(WCharEncoderState* ps, wchar_t wc) {
   assert(!ps->buf || ps->pos < ps->size);
   
   char tempBuf[MB_LEN_MAX];
   size_t len = wcrtomb(tempBuf, wc, &ps->mbState);
   if (len == (size_t)-1) {
      return false;
   }
   assert(len > 0);
   if (ps->buf) {
      if (len > ps->size - ps->pos) {
         // When this happens, ps->buf will not be null-terminated and
         // ps->mbState becomes undefined state (cannot be reused).
         // Thus we have no way to recover from here.
         return false;
      }
      char* buf = ps->buf;
      memcpy(&buf[ps->pos], tempBuf, len);
   }

   ps->pos += len;
   return true;
}
```

BenBE: Yes. And with the implementation used in the CGroup code, a `false` returned from the buffer write callback would immediately terminate processing. That's basically the same semantics used there too.

Explorer09: @BenBE That means an additional `memcpy` call, slowing things down slightly. For checking an error that should never happen unless there's a bug in code. I don't really like this.

BenBE: That `memcpy` is from L1 cache to a buffer that's likely in L1 too. As mentioned elsewhere: If in doubt, I prefer correctness over performance.

And if you want to avoid this, over-allocate by `MB_LEN_MAX + 1` and you can always write to `ps->buf`, if you force the NUL terminator on error.

Explorer09: > And if you want to avoid this, over-allocate by `MB_LEN_MAX + 1` and you can always write to `ps->buf`, if you force the NUL terminator on error.

_Over-allocation_ to solve the (actually non-existent) problem of _under-allocation_. Sorry, but I'm irritated by this logic.

The whole point of adding the bound check by your suggestion was to correct the case of _under-allocation_ (allocated buffer size is less than required). While I argued first that the problem literally won't exist, I agree on the point that we can _assert_ and _detect_ the error, at least as a precaution. The buffer should be allocated with the right size already, thus any code that attempts to _correct_ or _recover_ from any potential error is useless. I might just make the error fatal instead and call it a day.

```c
if (ps->buf && len <= ps->size - ps->pos) {
   fail();
}
```

As for why there is no NUL termination for the string when odds happen here: We are using a multibyte encoder `wcrtomb` that is capable of handling any encoding, not just UTF-8. When interpreting the standard behavior of `wcrtomb` in the strict sense, _it is not necessary for a null wide character to be encoded as 1 byte_. The standard `wcrtomb` will output shift sequences (if any) that restore the shift state to the initial state before outputting the null byte itself. Since that would require extra buffer space (up to `MB_CUR_MAX` bytes), any truncation of a multibyte string would have to be decided beforehand, and it won't be an escape hatch for any buffer size problem.

configure.ac
Explorer09: Update. `strnlen` was recently added in #1707.

Action.c
```diff
@@ -432,7 +432,8 @@ Htop_Reaction Action_setScreenTab(State* st, int x) {
    int rem = x - SCREEN_TAB_MARGIN_LEFT;
    for (unsigned int i = 0; i < settings->nScreens; i++) {
       const char* tab = settings->screens[i]->heading;
-      int width = rem >= bracketWidth ? (int)strnlen(tab, rem - bracketWidth + 1) : 0;
+      const char* ptr = tab;
```
BenBE, line 435: Why introduce this extra variable? Was this patch written by AI? ;-)

XUtils.c
```diff
@@ -259,6 +260,299 @@ size_t strnlen(const char* str, size_t maxLen) {
...
+static void String_encodeWChar(WCharEncoderState* ps, wchar_t wc) {
+   assert(!ps->buf || ps->pos < ps->size);
+
+   char tempBuf[MB_LEN_MAX];
+   char* dest = ps->buf ? (char*)ps->buf + ps->pos : tempBuf;
+
+   // It is unnecessarily expensive to fix the output string if the caller
+   // gives an incorrect buffer size. This function would not support any
+   // truncation of the output string.
+   size_t len = wcrtomb(dest, wc, &ps->mbState);
```
BenBE: Potential for buffer overruns, if a position fewer than `MB_LEN_MAX` from the end of the buffer is passed. A safer option (which should hit L1) would be to always pass `tempBuf` and copy `len` bytes on success, unless that would write to `ps->buf[ps->size - 1]` (which should be forced `0`).

XUtils.c
```diff
@@ -259,6 +260,299 @@ size_t strnlen(const char* str, size_t maxLen) {
...
+   assert(len > 0);
+   if (len == (size_t)-1) {
+      assert(len != (size_t)-1);
+      fail();
+   }
+   if (ps->buf && len > ps->size - ps->pos) {
+      assert(!ps->buf || len <= ps->size - ps->pos);
```
BenBE: This check is after an OOB write by `wcrtomb` already corrupted memory. This just cures the system if it survives til here. ASAN should complain inside the `wcrtomb` call above.

XUtils.c
```diff
@@ -259,6 +260,299 @@ size_t strnlen(const char* str, size_t maxLen) {
...
+   char* buf = xMalloc(size);
+   encState.size = size;
+   encState.buf = buf;
+   EncodePrintableString(&encState, str, maxLen, String_encodeWChar);
+   assert(encState.pos == size);
+
+   return buf;
+}
+
+bool String_decodeNextWChar(MBStringDecoderState* ps) {
```
BenBE: Naming consistency `String_encodeWChar` vs. `String_decodeNextWChar`.

Also, any reason to have this function exported from this TU?

XUtils.c
```diff
@@ -259,6 +260,299 @@ size_t strnlen(const char* str, size_t maxLen) {
...
+      if (isSeparator && separator != ' ') {
+         breakPos = *str;
+         breakWidth = totalWidth;
+      }
+   }
+
+   return totalWidth;
+}
+
+int String_mbswidth(const char** str, size_t maxLen, int maxWidth) {
```
BenBE, line 539: Why have the public API of this function modify an input argument to move to the end of the function?

XUtils.h
BenBE: ACK on making the error fatal.

Anyway, a good API makes it impossible to cause issues by abusing it. Thus if `wcrtomb` may write up to `MB_LEN_MAX` bytes you better should pass a buffer that can hold `MB_LEN_MAX` bytes, i.e. ensure the remaining space is large enough. Even if the actual call is known to write less.

And adding `MB_LEN_MAX + 1` to the allocation won't cause any noticeable overhead (malloc will just allocate the same and in rare occasions or under memory pressure may delay due to the OOM killer sweeping some swap …

BenBE (changes requested):

XUtils.c
Explorer09: This function has two outputs: One is the terminal width (i.e. num of terminal columns) and the other is the end character position after calculating the width.

The second output is useful to tell, for example, how many Unicode characters would it take to fit the number of terminal columns.

This functions has two uses combined in a single API: (1) to count the number of terminal columns for a given multibyte string; (2) to count the number characters it takes to fit the given terminal width.

Action.c
Explorer09: Technically no AI is used :)

The `ptr` is a temporary variable holding a pointer because of the API design I explained in https://github.com/htop-dev/htop/pull/1642/changes/BASE..6b82bcef3b8f7aecd9be6c83d66e84af152a1d58#r3211053429

XUtils.c
Explorer09: When I look at this code again I think your suggestion of always passing `tempBuf` to `wcrtomb` is the right choice. Not only it mitigates the buffer overrun problem, but it also prevents an undefined behavior when the input `wc` is an invalid code point.

XUtils.c
Explorer09: There's a reason for asymmetric naming here. (You may suggest a better naming.)

`encodeWChar()` and `decodeNextWChar()` are for low-level encoder and decoder with states.

For the decoder:
```c
bool String_decodeNextWChar(MBStringDecoderState* ps);
```
The input string is specified as `ps->str` and the next character decoded will be outputted in `ps->ch`.

For the encoder:
```c
/* The typedef is to allow different encoders to coexist. I have thought of an encoder that would present all non-printable characters as "\xXXXX" escape sequence. */
typedef ATTR_NONNULL void (*EncodeWChar)(WCharEncoderState* ps, /*wchar_t or int*/ wc);
void String_encodeWChar(WCharEncoderState* ps, /*wchar_t or int*/ wc);
```
The next character that would be encoded is specified as `wc` argument here and will be appended to the end of `ps->buf`. The `WCharEncoderState` object does not otherwise know the next `wchar_t` to encode.

Perhaps the name `String_encodeWChar` is obscure on what it does. Maybe the name `WCharEncoder_appendWChar` will be better? I'm not sure yet.

XUtils.c
BenBE: It might seem redundant, but I'd prefer if this output parameter was a separate (NULL-able) argument to the function. This separates concerns (input argument vs. optional additional output) and makes the API clearer. This also allows to just skip the extra local variable in case you don't care AND don't have it unexpectedly be changed.

Explorer09: > It might seem redundant, but I'd prefer if this output parameter was a separate (NULL-able) argument to the function. This separates concerns (input argument vs. optional additional output) and makes the API clearer. This also allows to just skip the extra local variable in case you don't care AND don't have it unexpectedly be changed.

A `static inline` function that wraps around this `String_mbswidth()` for a more convenient output interface? That wouldn't hurt, I guess.

By the way, the "input pointer that would be modified on output" is not my original idea. There are precedents in [mbsrtowcs(3)](https://man7.org/linux/man-pages/man3/mbsrtowcs.3.html) and [mbsnrtowcs(3)](https://man7.org/linux/man-pages/man3/mbsnrtowcs.3.html) (compare mbsrtowcs(3) with [mbstowcs(3)](https://man7.org/linux/man-pages/man3/mbstowcs.3.html) and I would argue that the `const char**` argument in the former makes the API more flexible).

XUtils.c
Explorer09: Applied the suggestion in [this comment](https://github.com/htop-dev/htop/pull/1642#discussion_r3210999167) and now the function no longer writes out of bounds.

XUtils.c
BenBE: I gave both variants of the API to Copilot to compare (without initially disclosing what the function would do) and except for "if it runs in a loop for parsing" the additional out argument was the preferred variant. This only solidified further once I gave Copilot also some more semantics of the function it just had rated. One interesting aspect was that modern compiler tend to optimize the additional out parameter variant better. So there's several aspects that lean towards that variant of the API.

Using the C standard library's string function API as a baseline for API quality sets a very low bar for quality … ;-)

Explorer09: > I gave both variants of the API to Copilot to compare (without initially disclosing what the function would do) and except for "if it runs in a loop for parsing" the additional out argument was the preferred variant.

What variants of the API did you give to Copilot in particular?

> One interesting aspect was that modern compiler tend to optimize the additional out parameter variant better. So there's several aspects that lean towards that variant of the API.

Again I'm not sure what variant you are referring to.

> Using the C standard library's string function API as a baseline for API quality sets a very low bar for quality … ;-)

I personally value utility of the API over the "quality" which I'm not sure what it is. Convenience of passing the arguments and calling the function? Register and stack usage?

It's not I disagree with changing the API, but I didn't see an API suggestion to compare with, thus I cannot judge anything regarding what Copilot said.

XUtils.c
Explorer09: Update: I've modified `String_encodeWChar` that its internal `wcrtomb` call now always writes to `tempBuf`. The function now never attempts to write out of bounds (regarding `ps->buf` and `ps->size`), however when the size is too small the error is still fatal. I don't think truncation of a string is an intended use case of this function, or forcing the output string to be null terminated (the string will be null terminated only upon the `String_encodeWChar(ps, '\0')` call).

Explorer09:
@coderabbitai The suggestion of `memset(&ps->mbState, 0, sizeof(ps->mbState));` on encountering an invalid sequence is incorrect. On a non-UTF-8 encoding, this would cause the decoder to lose the shift state if any shift state is required to decode the following string.

The safer approach is to restore the `mbstate_t` object to prior to decoding the invalid sequence.

Explorer09:
> Preserve pointer progress when decoding fails.

> At Line [478], breaking immediately on MBStringDecoder_nextWChar() failure drops decoder progress (notably the incomplete-sequence path that already advanced decoder.str), so callers can reprocess the same bytes.

The stopping at the first byte of the invalid sequence is intentional. How the invalid byte sequence would be handled is left to the caller.

Issue #1638, opened 2025-03-13
Htop 3.4.0-dev not showing CPU percentage on apple silicon


Korrd:
Since upgrading to `3.4.0-dev`, htop won't show CPU percentages on the processes' list.

**sudo htop**
![Image](https://github.com/user-attachments/assets/702d3c26-f009-4092-9687-841ef7cc4220)

I'm running macOS Sequoia 15.3.2 (24D81) on an M4 Pro Macbook

BenBE:
Can you take a look with `git bisect` if you can locate where this issue got introduced? TIA.

Explorer09:
Didn't see precisely where the error is through the screenshot. I have a M2 MacBook Pro with macOS Sequoia and didn't observe error on my side.

Korrd:
If you compare the CPU Usage bars to the CPU usage on the process list, you'll notice there are no processes reflecting the amount of CPU usage shown on the bars themselves.

aestriplex:
I reverted my previous [commit](https://github.com/htop-dev/htop/commit/802494f65f321ba78e49685760850da41701b89a). Now it shows a more reliable cpu percentage at a process level (or at least consistent with 3.3.0). 
@BenBE @fasterit I wait to open a pr, because I wish to further investigate the `Platform_calculateNanosecondsPerMachTick` function. In the case of a program compiled directly for silicon, it should not run in compatibility mode (i.e. without rosetta); here's the compiler binary from my M2:
```bash
% file htop
htop: Mach-O 64-bit executable arm64
```
so perhaps it should be enough
```c
...
else {
      // No workarounds needed, use the OS-provided value.
      mach_timebase_info(&info);
}
...
```

aestriplex:
The `mach_timebase_info` on Apple Silicon sets the structure correctly (125/3). Everything seems ok, both the cpu percentage and the running time (compared to top). Even if `isRunningUnderRosetta2` should always be false, because the program is compiled for the right architecture each time, probably we can leave those controls there, just in case.
I'll open a pr with the commit reverted

corneliusroemer:
The title of this issue wasn't quite intuitive to me so I didn't spot that what I reported in #1664 is likely a duplicate of this issue here.

In my case I _do_ see percentages for CPU, it's just that they are way too small since version 3.4.0 and on Apple silicon. Is that what this issue here is about? Why does it say "not showing percentage" when it _is_ showing?

Korrd:
Is there a release date for this fix?

[...]

Issue #1619, opened 2025-02-23
Running time of macOS process


aestriplex:
At the moment, the running time of processes on macOS is not updated correctly. I made two commit to try to fix this problem. The [branch](https://github.com/aestriplex/htop/tree/darwin-running-time) is  rebased on the PR #1617.

Summary:
* Removed the call to `Platform_machTicksToNanoseconds`. The `proc_taskinfo` structure contains the running time (both system and user) of the process. We should just take those value in nanoseconds and convert them in the proper unit. How the kernel handles this calculation (starting from low level cpu-dependent details) should not concern us.
* I also removed the static function `nanosecondsToCentiseconds` because for some reason we lose precision (± 1ms) with that calculation. If we divide directly by 1e7, we get a more precise measure. Probably it's a good idea to have an inline function or (maybe better) a macro, just to give a name to that operation and avoid the magic number;
* Since the GPU percentage update depends on the running time of the process, I updated the condition in the if statement. In fact, suppose we are at the *n*th iteration of the `DarwinProcess_setFromLibprocPidinfo` function. We should update the CPU usage value if the total CPU running time of the process is greater than the CPU running time of the process in the previous iteration. To do that, we do not need to use the `timeIntervalNS`.

TODO:
* remove the `static double Platform_nanosecondsPerMachTick` var in `Darwin/Platform.c` as well. It's only used in the (removed) `Platform_machTicksToNanoseconds`.
* remove the `Platform_calculateNanosecondsPerMachTick` function in `Darwin/PlatformHelpers.c`. It's only used to set the `static double Platform_nanosecondsPerMachTick` var once.

I attach here 3 sshot. The first one is the current version (3.3.0), the second one is the `darwin-running-time` brach, and the last one is top.

* 3.3.0:

<img width="1085" alt="Image" src="https://github.com/user-attachments/assets/07b5dfde-9fc9-4e7f-8f08-f762b1710fb9" />

* `darwin-running-time`:

<img width="1084" alt="Image" src="https://github.com/user-attachments/assets/719eb25f-aba9-4eae-873b-f33707babbd9" />

* top:

<img width="664" alt="Image" src="https://github.com/user-attachments/assets/8faad2aa-1a9b-4cb4-a6eb-d63b167c0882" />

BenBE:
Please avoid function-like macros and prefer `static inline`  functions with proper arguments. The compiler will inline as appropriate.

[...]

PR #1643, opened 2025-03-18
Revert "darwin: process' running time should not depend on mach ticks"

aestriplex:
This reverts commit 802494f65f321ba78e49685760850da41701b89a, which caused the issue #1638.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -375,6 +375,8 @@ AC_CHECK_FUNCS([ \
    ])
 
 if test "$my_htop_platform" = darwin; then
+   AC_CHECK_FUNCS([mach_timebase_info])
+
    AC_CHECK_FUNCS([host_statistics64], [
       AC_CHECK_TYPES([struct vm_statistics64], [], [], [[#include <mach/vm_statistics.h>]])
    ], [])
--- a/darwin/DarwinProcess.c
+++ b/darwin/DarwinProcess.c
@@ -370,8 +370,8 @@ void DarwinProcess_setFromLibprocPidinfo(DarwinProcess* proc, DarwinProcessTable
    const DarwinMachine* dhost = (const DarwinMachine*) proc->super.super.host;
 
    uint64_t total_existing_time_ns = proc->stime + proc->utime;
-   uint64_t user_time_ns = pti.pti_total_user;
-   uint64_t system_time_ns = pti.pti_total_system;
+   uint64_t user_time_ns = Platform_machTicksToNanoseconds(pti.pti_total_user);
+   uint64_t system_time_ns = Platform_machTicksToNanoseconds(pti.pti_total_system);
    uint64_t total_current_time_ns = user_time_ns + system_time_ns;
 
    if (total_existing_time_ns < total_current_time_ns) {
--- a/darwin/Platform.c
+++ b/darwin/Platform.c
@@ -148,11 +148,14 @@ const MeterClass* const Platform_meterTypes[] = {
    NULL
 };
 
+static double Platform_nanosecondsPerMachTick = 1.0;
+
 static double Platform_nanosecondsPerSchedulerTick = -1;
 
 static mach_port_t iokit_port; // the mach port used to initiate communication with IOKit
 
 bool Platform_init(void) {
+   Platform_nanosecondsPerMachTick = Platform_calculateNanosecondsPerMachTick();
 
    // Determine the number of scheduler clock ticks per second
    errno = 0;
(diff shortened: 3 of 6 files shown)


corneliusroemer:
Sorry for the ping here but I think it would be really good to get this merged and released asap, possibly as a hotfix sidestepping normal release schedule: The latest htop is very and subtly broken (CPU % too small by factor 40) for a large proportion of users (all Apple silicon users who've installed/updated to the latest htop released 4 weeks ago).

corneliusroemer (approved):
I've tested this PR locally on M1 Apple silicon 15.3.1 and it does fix the issue I observed for v3.4.0 (but not for v3.3.0).

image

fasterit:
@aestriplex, @corneliusroemer  Can we have a better fix than reverting everything to the previous state of brokenness?

aestriplex:
@fasterit I quote here my last comment to his #1638
> The `mach_timebase_info` on Apple Silicon sets the structure correctly (125/3). Everything seems ok, both the cpu percentage and the running time (compared to top). Even if `isRunningUnderRosetta2` should always be false, because the program is compiled for the right architecture each time, probably we can leave those controls there, just in case. I'll open a pr with the commit reverted

As I said there, in order to improve this bugfix a bit, the code that was removed *without* the rosetta compatibility check could be reintroduced (see my last commit in this PR). This would simplify the code and allow us to remove a function while preserving the old behavior

corneliusroemer:
What was broken before? It seems the current state is worse than before and unintended. So it should be reverted no matter what?

corneliusroemer:
How bad was issue https://github.com/htop-dev/htop/issues/1619?

CPU % being ~40x too small is definitely a bug issue. I can't quite tell how much off the process runtime was in the issue whose fix introduced the CPU usage bug.

You shouldn't block a revert by saying the fix brings back the original problem if the new problem is worse.

BenBE:
There are several comments from my side that point out issues that the revert would introduce. Like some division by zero issue and a request for clarification for some comment. AFAICS these have not been addressed yet.

The reason we would like to avoid plainly reverting the fix is that both behaviours are broken and instead of switching back to the old brokenness it would be better to look at the root cause and fix that instead. If reverting part or all of the previous fix is part of that solution there's nothing stopping that revert from being accepted. But as it stands ATM we are far from even knowing the root cause that needs fixing.

Development should consist of steps forward, instead of going back and forth …

aestriplex:
@BenBE Sorry, I was afk for the easter holidays. I totally agree on that. The fact is that for certain architectures (e.g. Apple Silicon) the old implementation was closer to top command. I am not sure why, since the mach apis should return the time in nanoseconds, not in cpu cycles, so no conversion should be needed. Anyway, it may took time to figure out exactly where the issue is (i.e. why this workaround - that was made for rosetta compatibility - is needed)

BenBE:
@aestriplex No problem. I was mostly explaining to @corneliusroemer the reasoning behind the current procedure and why the revert is unlikely to be merged as-is. If figuring out the root cause takes time, so be it.

Korrd:
While I understand that this revert re-introduces a bug that was fixed by it, I wonder how many are affected by it vs the issue it's causing? 

RIght now, the current htop release could be considered broken for apple silicon, so I guess fixing this makes sense, then re-tackling the original issue that may not be as high-impact as the current one.

BenBE:
The objection to the revert is not about re-introducing the bug it had previously, but rather that we'd like to avoid a back-and-forth and instead focus on actually resolving the issue.

Thus while reverting the specific commit may alleviate the pain short-term, it provides no way forward; which is what we'd like to see. You're still open to revert that specific commit in your local builds though.

erikolofsson:
`proc_taskinfo` is not in nano-seconds. See this:
https://github.com/apple-oss-distributions/shell_cmds/blob/14a2692995de90c2fde0c3c1e639a8c7d9e2312d/systime/systime.c#L313

On Intel based macOS:
```
info.numer: 1 info.denom 1
Platform_nanosecondsPerMachTick: 1.000000
```
On arm64 based macOS:
```
info.numer: 125 info.denom 3
Platform_nanosecondsPerMachTick: 41.666667
```

The reason the rosetta fix is needed is because rosetta translates `mach_timebase_info` to match that of Intel based macs, and at the same time the `proc_taskinfo` is not translated.

I have submitted a new pull request that should get rid of any precision issues that would show up because of the the 52 bit mantissa in the double conversions (should be at about 52 days of CPU time on Intel based macs, 6 years on arm64 based macs): https://github.com/htop-dev/htop/pull/1691

BenBE:
Closing this as superseeded by #1691 …

PR #1649, opened 2025-03-22
Meter: Prevent integer overflow on x positions

Explorer09:
* Clamp the `x` and `w` parameters in the meter's `draw()` functions to prevent `(x + w)` overflowing a signed int type. Also add assertions to `x` and `w` in the debug build. `(x >= 0 && w <= INT_MAX - x)`
* In `BarMeterMode_draw()`, the `offset + blockSizes[i]` calculation might also overflow. Cap the maximum of `blockSizes[i]` rather than cap the result of the addition.
* In `LEDMeterMode_draw()`, adjust the loop break conditions (on `xx` variable) so that the loop can terminate properly with `xx` values very close to the `INT_MAX` limit.

### DIFF
--- a/Meter.c
+++ b/Meter.c
@@ -10,6 +10,7 @@ in the source distribution for its full text.
 #include "Meter.h"
 
 #include <assert.h>
+#include <limits.h> // IWYU pragma: keep
 #include <math.h>
 #include <stdlib.h>
 #include <string.h>
@@ -49,6 +50,9 @@ static inline void Meter_displayBuffer(const Meter* this, RichString* out) {
 /* ---------- TextMeterMode ---------- */
 
 static void TextMeterMode_draw(Meter* this, int x, int y, int w) {
+   assert(x >= 0);
+   assert(w <= INT_MAX - x);
+
    const char* caption = Meter_getCaption(this);
    attrset(CRT_colors[METER_TEXT]);
    mvaddnstr(y, x, caption, w);
@@ -71,6 +75,9 @@ static void TextMeterMode_draw(Meter* this, int x, int y, int w) {
 static const char BarMeterMode_characters[] = "|#*@$%&.";
 
 static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
+   assert(x >= 0);
+   assert(w <= INT_MAX - x);
+
    // Draw the caption
    const char* caption = Meter_getCaption(this);
    attrset(CRT_colors[METER_TEXT]);
@@ -129,12 +136,11 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
       if (isPositive(value) && this->total > 0.0) {
          value = MINIMUM(value, this->total);
          blockSizes[i] = ceil((value / this->total) * w);
+         blockSizes[i] = MINIMUM(blockSizes[i], w - offset);
       } else {
          blockSizes[i] = 0;
       }
       int nextOffset = offset + blockSizes[i];
-      // (Control against invalid values)
-      nextOffset = CLAMP(nextOffset, 0, w);
       for (int j = offset; j < nextOffset; j++)
          if (RichString_getCharVal(bar, startPos + j) == ' ') {
             if (CRT_colorScheme == COLORSCHEME_MONOCHROME) {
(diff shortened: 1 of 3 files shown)


Meter.c
```diff
@@ -71,6 +72,12 @@ static void TextMeterMode_draw(Meter* this, int x, int y, int w) {
 static const char BarMeterMode_characters[] = "|#*@$%&.";
 
 static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
+   // Prevent (x + w) overflow
+   assert(x >= 0);
+   x = MAXIMUM(0, x);
+   assert(w <= INT_MAX - x);
+   w = MINIMUM(INT_MAX - x, w);
```
Explorer09, line 79: I was considering whether this clamping of `x` and `w` is really necessary. While the two `assert` lines are useful, it might be easier to guard against overflow on the callers of the `Meter.draw()` functions, rather than guard on the callees.

There are only four functions that would compute `x`, `y`, and `w` values for each meter and call `Meter.draw()`:
* `Header_draw`
* `CPUMeterCommonDraw`
* `SingleColCPUsMeter_draw`
* `MemorySwapMeter_draw`

Meter.c
```diff
@@ -49,16 +50,22 @@ static inline void Meter_displayBuffer(const Meter* this, RichString* out) {
 /* ---------- TextMeterMode ---------- */
 
 static void TextMeterMode_draw(Meter* this, int x, int y, int w) {
+   assert(x >= 0);
+   assert(w <= INT_MAX - MAXIMUM(0, x));
```
BenBE, line 54:
```suggestion
   assert(w <= INT_MAX - x);
```

Suffices. `x>=0` has just been checked above.

Meter.c
```diff
@@ -190,12 +202,16 @@ static const char* const GraphMeterMode_dotsAscii[] = {
 };
 
 static void GraphMeterMode_draw(Meter* this, int x, int y, int w) {
+   assert(x >= 0);
+   assert(w <= INT_MAX - MAXIMUM(0, x));
```
BenBE, line 206:
```suggestion
   assert(w <= INT_MAX - x);
```

Suffices. `x>=0` has just been checked above.

Meter.c
```diff
@@ -301,6 +321,27 @@ static void LEDMeterMode_drawDigit(int x, int y, int n) {
 }
 
 static void LEDMeterMode_draw(Meter* this, int x, int y, int w) {
+   assert(x >= 0);
+   assert(w <= INT_MAX - MAXIMUM(0, x));
```
BenBE, line 325:
```suggestion
   assert(w <= INT_MAX - x);
```

Suffices. `x>=0` has just been checked above.

BenBE (changes requested):
Some simplifications possible.

Also, please don't mix code hardening fixes with unrelated behaviour changes.

Explorer09:
> Also, please don't mix code hardening fixes with unrelated behaviour changes.

What do you mean? Which parts of the code changes do I need splitting?

BenBE:
> > Also, please don't mix code hardening fixes with unrelated behaviour changes.
> 
> What do you mean? Which parts of the code changes do I need splitting?

AFAICS, "[Meter: Prevent integer overflow on x positions](https://github.com/htop-dev/htop/pull/1649/commits/a97d5c00f4a9f00cd2bb9e60ca537af1674fc51c)" is the actual overflow fix, while the other two commits change behaviour. Or did I miss something?

Explorer09:
> > > Also, please don't mix code hardening fixes with unrelated behaviour changes.
> > 
> > 
> > What do you mean? Which parts of the code changes do I need splitting?
> 
> AFAICS, "[Meter: Prevent integer overflow on x positions](https://github.com/htop-dev/htop/pull/1649/commits/a97d5c00f4a9f00cd2bb9e60ca537af1674fc51c)" is the actual overflow fix, while the other two commits change behaviour. Or did I miss something?

"[CPUMeter: Fix negative "x" positions of sub-meters](https://github.com/htop-dev/htop/pull/1649/commits/8cc92b3f748fb98e218e6e28f2612223336f9568)" is related and necessary after the first fix because this fixes one of the conditions that x can wrap to a negative number. Since the first patch adds assertions about `x >= 0`, without this second patch you can get an assertion error.

Perhaps I can reorder the patch so that this "CPUMeter" patch is applied first, before the main one ("Meter: Prevent integer overflow on x positions"), to reduce confusion.

The third patch, "[Meter: Don't draw caption or borders if width is not enough](https://github.com/htop-dev/htop/pull/1649/commits/7ff6631ce71ce11e0786dab0374b43a85885a2d0)", is to further remove potential undefined behaviours by strengthen the logics. It does change behaviours as you said, but I believe I cannot split this one out as the original logic could involve in UB in certain conditions. In other words, this is related anyway.

PR #1655, opened 2025-03-28
CPUMeter: Improve spacing between sub-meter algorithm

### LINKED ISSUE #1654: CPU meter bar misalignment with groups with different number of sub-columns
When CPU meter groups with different number of "shorter columns" are place together in one column, and the width for the (larger) meter column is not divisible by 8, then a visual misalignment may be seen, such as in this screenshot:

htop screenshot: meter misalignment

I personally have a solution to this problem, and yet I don't know if such fix is worth it. (Thus I raise this issue for discussion first.)

Explorer09:
Fixes #1654.
The formula is derived by myself.

Interestingly, I asked both ChatGPT and Google Gemini (AI) on how to make a formula based on a table of given values, they either gave me a wrong result or simply fail to make one. (I don't trust these AIs in general, by the way.)
(https://chatgpt.com/share/67e6fb4a-da80-8004-a993-11cc19666fad)
(https://g.co/gemini/share/356fa79d2f5a)

### DIFF
--- a/CPUMeter.c
+++ b/CPUMeter.c
@@ -305,8 +305,8 @@ static void CPUMeterCommonDraw(Meter* this, int x, int y, int w, int ncol) {
    Meter** meters = data->meters;
    int start, count;
    AllCPUsMeter_getRange(this, &start, &count);
-   int colwidth = (w - ncol) / ncol + 1;
-   int diff = (w - (colwidth * ncol));
+   int colwidth = w / ncol;
+   int diff = w % ncol;
    int nrows = (count + ncol - 1) / ncol;
    for (int i = 0; i < count; i++) {
       int d = (i / nrows) > diff ? diff : (i / nrows); // dynamic spacer
--- a/CPUMeter.c
+++ b/CPUMeter.c
@@ -12,6 +12,7 @@ in the source distribution for its full text.
 #include <assert.h>
 #include <stdbool.h>
 #include <stddef.h>
+#include <stdint.h>
 #include <stdlib.h>
 #include <string.h>
 
@@ -306,11 +307,14 @@ static void CPUMeterCommonDraw(Meter* this, int x, int y, int w, int ncol) {
    int start, count;
    AllCPUsMeter_getRange(this, &start, &count);
    int colwidth = w / ncol;
-   int diff = w % ncol;
    int nrows = (count + ncol - 1) / ncol;
    for (int i = 0; i < count; i++) {
-      int d = (i / nrows) > diff ? diff : (i / nrows); // dynamic spacer
-      int xpos = x + ((i / nrows) * colwidth) + d;
+      // Spacing between meters based on the remainder of (w % ncol)
+      uint32_t v = (unsigned int)((w % ncol) * 32 / ncol);
+      v = ((v * 0x00210842U) & 0x02082082U) * (unsigned int)(i / nrows);
+      v = (uint32_t)(((v + 0x02108420U) & 0x7DE71840U) * 0x00210842ULL) >> 27;
+      int spacing = (int)v;
+      int xpos = x + ((i / nrows) * colwidth) + spacing;
       int ypos = y + ((i % nrows) * meters[0]->h);
       meters[i]->draw(meters[i], xpos, ypos, colwidth);
    }
--- a/MemorySwapMeter.c
+++ b/MemorySwapMeter.c
@@ -38,7 +38,7 @@ static void MemorySwapMeter_draw(Meter* this, int x, int y, int w) {
 
    /* Use the same width for each sub meter to align with CPU meter */
    const int colwidth = w / 2;
-   const int diff = w - colwidth * 2;
+   const int diff = w % 2;
 
    assert(data->memoryMeter->draw);
    data->memoryMeter->draw(data->memoryMeter, x, y, colwidth);


CPUMeter.c
```diff
@@ -305,12 +306,15 @@ static void CPUMeterCommonDraw(Meter* this, int x, int y, int w, int ncol) {
...
-   int diff = (w - (colwidth * ncol));
+   int colwidth = w / ncol;
    int nrows = (count + ncol - 1) / ncol;
    for (int i = 0; i < count; i++) {
-      int d = (i / nrows) > diff ? diff : (i / nrows); // dynamic spacer
-      int xpos = x + ((i / nrows) * colwidth) + d;
+      // Spacing between meters based on the remainder of (w % ncol)
+      uint32_t v = (unsigned int)((w % ncol) * 32 / ncol);
+      v = ((v * 0x00210842U) & 0x02082082U) * (unsigned int)(i / nrows);
+      v = (uint32_t)(((v + 0x02108420U) & 0x7DE71840U) * 0x00210842ULL) >> 27;
```
BenBE, line 315: Please explain those magic numbers …

Explorer09:
> Please explain those magic numbers …

It's a bit difficult to explain in code. I wonder if you or other people have ideas to make the code more readable (maybe wrap it into a function or something - I don't know how to name such function, though).

Anyway. Here're my explanations:

1. The goal is to make the sub-meters of each CPU core/thread aligned despite the number of sub-columns of a CPU meter (group).
2. Currently, there can only be 2, 4 or 8 sub-columns for a meter.
3. The idea is to adjust the distribution of "spaces" between sub-meters based on the remainder of (`w` mod `ncol`). Since the width of each sub-meter is floor(`w` / `ncol`), the total number of spaces to distribute would be equal to (`w` mod `ncol`).
4. When the remainder is a multiple of 2, there should not be a space inserted in the (`ncol` / 2) position or otherwise this meter (group) would appear misaligned with other meter groups with 2 sub-columns.
5. Apply the similar rule: When the remainder is a multiple of N, there should not be a space inserted in the (`ncol` * M / N) position (where M is any integer in the range [1, N-1]) or otherwise this meter (group) would appear misaligned with other meter groups with N sub-columns.
6. The problem may be more complicated if N can be any positive integer (it would require integer factorisation and a loop), so I narrow the problem scope to cases where N is a power of two, like 2, 4, 8, 16 and 32.
7. Pseudocode:

```text
spacing_function(ncol, i, w) {
   // 'i' is column index of the sub-meter.
   // assert(i >=0 && i < ncol);
   w %= ncol;
   res = 0;
   foreach (N in {2, 4, 8, 16, 32}) {
      if ((w + 1) % N == 0)
         res += (int)(i * N / ncol);
   }
   return res;
}
```

It technically requires a loop. So I use bit manipulation tricks to simplify this, utilising N is a power of two within the narrowed problem scope, and came up with the magic numbers for multiplication and mask values.

`((w % ncol) * 32 / ncol)` // Maximum number of columns is 32 for this formula. 
`(v * 0x00210842U) & 0x02082082U)` // Break the remainder of (`w` % `ncol`) into 5 separate bits, and reverse the bit order.
`... * (unsigned int)(i / nrows)` // Make up to 5 copies of this column index value, but if `((w + 1) % N != 0)` set that copy to zero.
`(v + 0x02108420U) & 0x7DE71840U)` // Compute the addends that's equivalent to the `(int)(i * N / ncol)` in the pseudocode, for each copy.
`(... * 0x00210842ULL) >> 27` // Sum up all 5 numbers for the final result.

```text
Before
w=56 | [99.5%][99.5%][99.5%][99.5%][99.5%][99.5%][99.5%][99.5%]
w=57 | [99.5%] [99.5%][99.5%][99.5%][99.5%][99.5%][99.5%][99.5%]
w=58 | [99.5%] [99.5%] [99.5%][99.5%][99.5%][99.5%][99.5%][99.5%]
w=59 | [99.5%] [99.5%] [99.5%] [99.5%][99.5%][99.5%][99.5%][99.5%]
w=60 | [99.5%] [99.5%] [99.5%] [99.5%] [99.5%][99.5%][99.5%][99.5%]
w=61 | [99.5%] [99.5%] [99.5%] [99.5%] [99.5%] [99.5%][99.5%][99.5%]
w=62 | [99.5%] [99.5%] [99.5%] [99.5%] [99.5%] [99.5%] [99.5%][99.5%]
w=63 | [99.5%] [99.5%] [99.5%] [99.5%] [99.5%] [99.5%] [99.5%] [99.5%]
w=64 | [|99.5%][|99.5%][|99.5%][|99.5%][|99.5%][|99.5%][|99.5%][|99.5%]
w=65 | [|99.5%] [|99.5%][|99.5%][|99.5%][|99.5%][|99.5%][|99.5%][|99.5%]
w=66 | [|99.5%] [|99.5%] [|99.5%][|99.5%][|99.5%][|99.5%][|99.5%][|99.5%]
w=67 | [|99.5%] [|99.5%] [|99.5%] [|99.5%][|99.5%][|99.5%][|99.5%][|99.5%]
w=68 | [|99.5%] [|99.5%] [|99.5%] [|99.5%] [|99.5%][|99.5%][|99.5%][|99.5%]
w=69 | [|99.5%] [|99.5%] [|99.5%] [|99.5%] [|99.5%] [|99.5%][|99.5%][|99.5%]
w=70 | [|99.5%] [|99.5%] [|99.5%] [|99.5%] [|99.5%] [|99.5%] [|99.5%][|99.5%]
w=71 | [|99.5%] [|99.5%] [|99.5%] [|99.5%] [|99.5%] [|99.5%] [|99.5%] [|99.5%]
w=72 | [||99.5%][||99.5%][||99.5%][||99.5%][||99.5%][||99.5%][||99.5%][||99.5%]

After
w=56 | [99.5%][99.5%][99.5%][99.5%][99.5%][99.5%][99.5%][99.5%]
w=57 | [99.5%][99.5%][99.5%][99.5%] [99.5%][99.5%][99.5%][99.5%]
w=58 | [99.5%][99.5%] [99.5%][99.5%][99.5%][99.5%] [99.5%][99.5%]
w=59 | [99.5%][99.5%] [99.5%][99.5%] [99.5%][99.5%] [99.5%][99.5%]
w=60 | [99.5%] [99.5%][99.5%] [99.5%][99.5%] [99.5%][99.5%] [99.5%]
w=61 | [99.5%] [99.5%][99.5%] [99.5%] [99.5%] [99.5%][99.5%] [99.5%]
w=62 | [99.5%] [99.5%] [99.5%] [99.5%][99.5%] [99.5%] [99.5%] [99.5%]
w=63 | [99.5%] [99.5%] [99.5%] [99.5%] [99.5%] [99.5%] [99.5%] [99.5%]
w=64 | [|99.5%][|99.5%][|99.5%][|99.5%][|99.5%][|99.5%][|99.5%][|99.5%]
w=65 | [|99.5%][|99.5%][|99.5%][|99.5%] [|99.5%][|99.5%][|99.5%][|99.5%]
w=66 | [|99.5%][|99.5%] [|99.5%][|99.5%][|99.5%][|99.5%] [|99.5%][|99.5%]
w=67 | [|99.5%][|99.5%] [|99.5%][|99.5%] [|99.5%][|99.5%] [|99.5%][|99.5%]
w=68 | [|99.5%] [|99.5%][|99.5%] [|99.5%][|99.5%] [|99.5%][|99.5%] [|99.5%]
w=69 | [|99.5%] [|99.5%][|99.5%] [|99.5%] [|99.5%] [|99.5%][|99.5%] [|99.5%]
w=70 | [|99.5%] [|99.5%] [|99.5%] [|99.5%][|99.5%] [|99.5%] [|99.5%] [|99.5%]
w=71 | [|99.5%] [|99.5%] [|99.5%] [|99.5%] [|99.5%] [|99.5%] [|99.5%] [|99.5%]
w=72 | [||99.5%][||99.5%][||99.5%][||99.5%][||99.5%][||99.5%][||99.5%][||99.5%]
```

PR #1663, opened 2025-04-08
PCP: Remove semicolons from empty statements

Explorer09:
These are warnings generated when compiling PCP code in Clang 

```text
pcp/PCPDynamicMeter.c:200:10: error: empty expression statement has no effect; remove unnecessary ';' to silence this warning [-Werror,-Wextra-semi-stmt]
  200 |          ;  /* skip this one, we're looking for a new header */
      |          ^
pcp/PCPDynamicScreen.c:255:10: error: empty expression statement has no effect; remove unnecessary ';' to silence this warning [-Werror,-Wextra-semi-stmt]
  255 |          ;  /* skip this one, we're looking for a new header */
      |          ^
pcp/PCPDynamicScreen.c:257:10: error: empty expression statement has no effect; remove unnecessary ';' to silence this warning [-Werror,-Wextra-semi-stmt]
  257 |          ;  /* skip this one as we always need value strings */
      |          ^
```

No changes to compiled code.

### DIFF
--- a/pcp/PCPDynamicMeter.c
+++ b/pcp/PCPDynamicMeter.c
@@ -198,16 +198,18 @@ static void PCPDynamicMeter_parseFile(PCPDynamicMeters* meters, const char* path
             meter = PCPDynamicMeter_new(meters, key + 1);
       } else if (!ok) {
          /* skip this one, we're looking for a new header */
-      } else if (value && meter && String_eq(key, "caption")) {
+      } else if (!value || !meter) {
+         /* skip this one as we always need value strings */
+      } else if (String_eq(key, "caption")) {
          char* caption = String_cat(value, ": ");
          if (caption) {
             free_and_xStrdup(&meter->super.caption, caption);
             free(caption);
             caption = NULL;
          }
-      } else if (value && meter && String_eq(key, "description")) {
+      } else if (String_eq(key, "description")) {
          free_and_xStrdup(&meter->super.description, value);
-      } else if (value && meter && String_eq(key, "type")) {
+      } else if (String_eq(key, "type")) {
          if (String_eq(config[1], "bar"))
             meter->super.type = BAR_METERMODE;
          else if (String_eq(config[1], "text"))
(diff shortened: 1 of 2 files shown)


pcp/PCPDynamicMeter.c
```diff
@@ -197,7 +197,7 @@ static void PCPDynamicMeter_parseFile(PCPDynamicMeters* meters, const char* path
          if (ok)
             meter = PCPDynamicMeter_new(meters, key + 1);
       } else if (!ok) {
-         ;  /* skip this one, we're looking for a new header */
+         (void)0;  /* skip this one, we're looking for a new header */
       } else if (value && meter && String_eq(key, "caption")) {
```
BenBE, line 201: Code style note:

Adding a check `else if (!value || !meter) { (void)0; }` can simplify the conditions below.

Change similar to what is already present in the other location affected by this change.

Explorer09:
> I'm not a fan of the `(void)0;` marker, when the solution suggested by the compiler (just skip the `;`) works equally well and is unambiguous due to the comments already present.

I don't know if any compiler would complain about an empty conditional block. Because I don't like fixing this one either.

I wonder if the person who maintains the htop PCP code can take a look at this.

natoscott:
> > I'm not a fan of the `(void)0;` marker, when the solution suggested by the compiler (just skip the `;`) works equally well and is unambiguous due to the comments already present.
> 
> I don't know if any compiler would complain about an empty conditional block. Because I don't like fixing this one either.
> 
> I wonder if the person who maintains the htop PCP code can take a look at this.

I think the empty braces (no semi-colon) with a comment is my preference here, and I'm pretty sure compilers (gcc at least) don't complain on this pattern.

Explorer09:
> I think the empty braces (no semi-colon) with a comment is my preference here, and I'm pretty sure compilers (gcc at least) don't complain on this pattern.

Just tested and there are no warnings in Clang either.

BenBE (approved):
LGTM.

PR #1667, opened 2025-04-09
pcp/Platform.c: defensive code for forthcoming PCP 7.0 change

kmcdonell:
Add conditional code (no-op for the current PMAPI_VERSION_2 aka 2) that will allow the code to continue to work with PMAPI_VERSION_3 aka 3 in PCP 7.0 that is under development.

### DIFF
--- a/pcp/Platform.c
+++ b/pcp/Platform.c
@@ -349,7 +349,14 @@ bool Platform_init(void) {
 
    if (opts.context == PM_CONTEXT_ARCHIVE) {
       gettimeofday(&pcp->offset, NULL);
+#if PMAPI_VERSION >= PMAPI_VERSION_3
+    {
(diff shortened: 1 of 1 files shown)


pcp/Platform.c
```diff
@@ -349,7 +349,14 @@ bool Platform_init(void) {
 
    if (opts.context == PM_CONTEXT_ARCHIVE) {
       gettimeofday(&pcp->offset, NULL);
+#if PMAPI_VERSION >= 3
+    {
```
natoscott, line 353: We wont need these extra braces @kmcdonell - htop code base uses c99 by default and declaring variables in-place like this (without restricting scope) is the normal htop style.

natoscott:
I cleaned up those braces, rebased, squashed commits and merged.

PR #1668, opened 2025-04-10
build: Automatically detect backtrace(3) return type

Explorer09:
The return type of the backtrace() function is actually different among operating systems. Rather than fix it to 'int' or 'size_t', detect it automatically in the configure script.

The automatic type of backtrace() would prevent the warning "-Wshorten-64-to-32" in Clang for implicitly converting 'size_t' to 'int', either during the return of backtrace() or the call of backtrace_symbols_fd() (when passing the length argument as input).

Regression from 3cfc28968feee6ebbbf2058f2046ba0bab87c89f


Explorer09:
Note: There is currently no configure message for this new check. I wonder if people want to see the `checking the return type of backtrace...` message printed.

The messages may be...

```text
checking the return type of backtrace... automatic (using typeof)
checking the return type of backtrace... automatic (using __typeof__)
checking the return type of backtrace... automatic (using __typeof)
checking the return type of backtrace... size_t
checking the return type of backtrace... int
checking the return type of backtrace... not available
```

Explorer09:
> How about a classic
> 
> ```
> #if defined(HTOP_FREEBSD) || defined(HTOP_OPENBSD) || defined(HTOP_NETBSD) || defined(HTOP_DRAGONFLYBSD) ..
> #else 
> ...
> ```
> 
> too easy? Too boring?

@fasterit That's against what Autoconf was designed for. Besides, what if a newer version of the operating system or libc suddenly changes the interface?

Explorer09:
> Operating systems like *BSD and Linux don't change their interfaces suddenly. The advantage of explicit `#if defined` is that the assumptions are documented. Just what you needed to do in the comments as nobody would easily understand your code just reading what you do there. Sometimes simpler is better.

I've already commented on why there needs to be an automatic detection. The operating systems listed are only there as examples. In the long run, the thing would just work and you won't need to add any new OS to the list.

There's one more thing worth mentioning: musl libc doesn't come with the `backtrace()` function and Linux distributions using musl libc would build htop with a separate `libexecinfo` to get the backtrace function. ([See this page as an example.](https://github.com/openwrt/packages/issues/11863)) `backtrace(3)` is not a system call, so it's not appropriate to use `HTOP_LINUX` or similar macros as conditionals for the feature.

BenBE:
But we already have a check for `libexecinfo`, so you could just check for this in the code and have this detail documented as well. AFAICS, `typeof` is a (although common) compiler extension standardized only in C23 …

Thus in a strict sense we would never compile the `typeof` variants from configure and thus would rely on proper compiler behaviour (vulg. raise a warning) to distinguish the remaining cases.

NB: `__auto_type` …

Explorer09:
> But we already have a check for `libexecinfo`, so you could just check for this in the code and have this detail documented as well.

The point is which implementation of libexecinfo. 
[FreeBSD man page](https://man.freebsd.org/cgi/man.cgi?query=backtrace&sektion=3) says its backtrace function needs to link with libexecinfo. Just like the Linux-with-musl case mentioned in my previous post.

And so we had at least two implementations of libexecinfo with different prototypes for backtrace(3)...


> AFAICS, `typeof` is a (although common) compiler extension standardized only in C23 …
> Thus in a strict sense we would never compile the `typeof` variants from configure and thus would rely on proper compiler behaviour (vulg. raise a warning) to distinguish the remaining cases.

Yes. And that's why I checked `__typeof__` and `__typeof` keywords as well. Although GCC won't allow `typeof` in `-std=c99` mode, it does allow `__typeof__` and `__typeof` to be used with otherwise C99-compliant code.

If a compiler doesn't support `typeof` at all (which would be rarer as it's now part of the C23 standard), I check `size_t` and `int` specifically. I think my way of checking (by ensuring a function pointer type is compatible) is the most reliable among all other methods.

PR #1669, opened 2025-04-10
Some '-Wshorten-64-to-32' warning fixes

Explorer09:
This patch contains a small set of changes that are intended to fix the `-Wshorten-64-to-32` warnings in Clang (enabled by default in Clang 19 and later). This only contains changes that I considered "trivial", that is, not needing any major change to the data structures or code.

### DIFF
--- a/AffinityPanel.c
+++ b/AffinityPanel.c
@@ -400,7 +400,7 @@ Panel* AffinityPanel_new(Machine* host, const Affinity* affinity, int* width) {
 
       char number[16];
       xSnprintf(number, 9, "CPU %d", Settings_cpuId(host->settings, i));
-      unsigned cpu_width = 4 + strlen(number);
+      unsigned int cpu_width = 4 + (unsigned int)strlen(number);
       if (cpu_width > this->width) {
          this->width = cpu_width;
       }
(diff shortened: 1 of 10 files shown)


BenBE:
Digging into the source for a bit, I found the ACPI battery driver implementation.
https://github.com/openbsd/src/blob/5990a1963d9d3a8702389a957caa067859ba104b/sys/dev/acpi/acpibat.c#L153-L216
Things don't line up exactly though, as I'd expect `hw.sensors.acpibat0.7.6` to be read instead of `hw.sensors.acpibat0.7.3`. The 3 would only match if you ignored the non-WATTHOUR readings in the indexing …

AffinityPanel.c
```diff
@@ -400,7 +400,7 @@ Panel* AffinityPanel_new(Machine* host, const Affinity* affinity, int* width) {
 
       char number[16];
       xSnprintf(number, 9, "CPU %d", Settings_cpuId(host->settings, i));
-      unsigned cpu_width = 4 + strlen(number);
+      unsigned cpu_width = 4 + (unsigned)strlen(number);
```
BenBE, line 403: While `unsigned` is a valid alias for `unsigned int` in C, it would be preferable to use `unsigned int` to be more explicit about the referenced type. Cf. my recent cleanup PR #1675 …

openbsd/Platform.c
```diff
@@ -397,6 +397,6 @@ void Platform_getBattery(double* percent, ACPresence* isOnAC) {
       mib[3] = SENSOR_INDICATOR;
       mib[4] = 0; /* "power supply" (status indicator) */
       if (sysctl(mib, 5, &s, &slen, NULL, 0) != -1)
-         *isOnAC = s.value;
+         *isOnAC = s.value != 0 ? AC_PRESENT : AC_ABSENT;
```
BenBE, line 400: This should be its own commit.

BenBE (changes requested):

PR #1672, opened 2025-04-11
openbsd: Document sysctl indices for ACPI battery & AC code

Explorer09:
Avoid magic numbers. No changes to the compiled code.

(Split from #1669)

### DIFF
--- a/openbsd/Platform.c
+++ b/openbsd/Platform.c
@@ -368,16 +368,16 @@ void Platform_getBattery(double* percent, ACPresence* isOnAC) {
 
    *percent = NAN;
    if (found) {
-      /* last full capacity */
-      mib[3] = 7;
-      mib[4] = 0;
+      /* See "sys/dev/acpi/acpibat.c" of OpenBSD source code for the indices
+         of the last field. */
+      mib[3] = SENSOR_WATTHOUR;
+      mib[4] = 0; /* "last full capacity" */
       double last_full_capacity = 0;
       if (sysctl(mib, 5, &s, &slen, NULL, 0) != -1)
          last_full_capacity = s.value;
       if (last_full_capacity > 0) {
-         /*  remaining capacity */
-         mib[3] = 7;
-         mib[4] = 3;
+         mib[3] = SENSOR_WATTHOUR;
+         mib[4] = 3; /* "remaining capacity" */
          if (sysctl(mib, 5, &s, &slen, NULL, 0) != -1) {
             double charge = s.value;
             *percent = 100 * (charge / last_full_capacity);
@@ -392,8 +392,10 @@ void Platform_getBattery(double* percent, ACPresence* isOnAC) {
 
    *isOnAC = AC_ERROR;
    if (found) {
-      mib[3] = 9;
-      mib[4] = 0;
+      /* See "sys/dev/acpi/acpiac.c" of OpenBSD source code.
+         There is only one "sensor" for this device. */
+      mib[3] = SENSOR_INDICATOR;
+      mib[4] = 0; /* "power supply" (status indicator) */
       if (sysctl(mib, 5, &s, &slen, NULL, 0) != -1)
          *isOnAC = s.value;
    }


openbsd/Platform.c
```diff
@@ -369,14 +369,14 @@ void Platform_getBattery(double* percent, ACPresence* isOnAC) {
    *percent = NAN;
    if (found) {
       /* last full capacity */
-      mib[3] = 7;
+      mib[3] = SENSOR_WATTHOUR;
       mib[4] = 0;
```
BenBE, line 373: Note: `Last full capacity reading`

openbsd/Platform.c
```diff
@@ -369,14 +369,14 @@ void Platform_getBattery(double* percent, ACPresence* isOnAC) {
...
       double last_full_capacity = 0;
       if (sysctl(mib, 5, &s, &slen, NULL, 0) != -1)
          last_full_capacity = s.value;
       if (last_full_capacity > 0) {
          /*  remaining capacity */
-         mib[3] = 7;
+         mib[3] = SENSOR_WATTHOUR;
          mib[4] = 3;
```
BenBE, line 380: Note: `Remaining capacity reading`

openbsd/Platform.c
```diff
@@ -392,7 +392,7 @@ void Platform_getBattery(double* percent, ACPresence* isOnAC) {
 
    *isOnAC = AC_ERROR;
    if (found) {
-      mib[3] = 9;
+      mib[3] = SENSOR_INDICATOR;
       mib[4] = 0;
```
BenBE, line 396: Note: `AC Power Status Indicator`

BenBE (commented):
When removing magic number, at least commet the ones remaining …

openbsd/Platform.c
Explorer09: Do you suggest me to change this number to 6, or should I just keep the current number (3) for now?

Explorer09: Edit: Ignore my previous message. I can tell why the number is 3 now instead of 6.

Explorer09:
Copying @BenBE 's comment from issue #1669 to here for reference:

> Digging into the source for a bit, I found the ACPI battery driver implementation. https://github.com/openbsd/src/blob/5990a1963d9d3a8702389a957caa067859ba104b/sys/dev/acpi/acpibat.c#L153-L216 Things don't line up exactly though, as I'd expect `hw.sensors.acpibat0.7.6` to be read instead of `hw.sensors.acpibat0.7.3`. The 3 would only match if you ignored the non-WATTHOUR readings in the indexing …

I guess the ordinal numbers (as mentioned in the [sysctl(2) man page](https://man.openbsd.org/sysctl.2#HW_SENSORS~2)) are intended to be sorted by `sensor_type` when retrieved from sysctl(2). That's why the number is 3 and not 6. Otherwise, you can get lots of useless values for `hw.sensors.acpibat0.7.{3,4,5}` as they would have been in the wrong types.

openbsd/Platform.c
Explorer09: I think I can document this in a better way…

openbsd/Platform.c
```diff
@@ -368,16 +368,16 @@ void Platform_getBattery(double* percent, ACPresence* isOnAC) {
 
    *percent = NAN;
    if (found) {
-      /* last full capacity */
-      mib[3] = 7;
-      mib[4] = 0;
+      /* See "sys/dev/acpi/acpibat.c" of OpenBSD source code for the indexes
+         of the last field. */
```
BenBE, line 372:
```suggestion
      /* See "sys/dev/acpi/acpibat.c" of OpenBSD source code for the indices of the last field. */
```

BenBE (approved):
LGTM. One typo left.

openbsd/Platform.c
Explorer09: Not a typo, just a different spelling of the word. You still want me to change it?

BenBE: AFAIU, indexes is mostly used for written documents, while indices is more often used in the mathematical sense. So yes, in this context "indices" is IMHO the preferred spelling¹.

¹ IANAL²
² Linguist

BenBE (approved):

openbsd/Platform.c
Explorer09: @BenBE I decided to go with "indices" not for your reasoning, but because I didn't find any use of "indexes" in htop codebase but I got one result of "indices" when I ran `grep -r indices`.

Issue #1686, opened 2025-04-27
3.4.1 build error on macOS: `DarwinProcess.c: error: ‘thread_extended_info_data_t’ undeclared`


barracuda156:
```
    /usr/bin/gcc-4.2 -std=gnu99 -DHAVE_CONFIG_H -I.  -DNDEBUG -isystem/opt/local/include/LegacySupport -I/opt/local/include -std=c99 -pedantic -D_DARWIN_C_SOURCE -DNCURSES_WIDECHAR -Wall -Wcast-align -Wcast-qual -Wextra -Wfloat-equal -Wformat=2 -Winit-self -Wmissing-format-attribute -Wmissing-noreturn -Wmissing-prototypes -Wpointer-arith -Wshadow -Wstrict-prototypes -Wundef -Wunused -Wwrite-strings -DSYSCONFDIR="\"/opt/local/etc\"" -I"./darwin" -pipe -Os -arch ppc -MT zfs/ZfsArcMeter.o -MD -MP -MF $depbase.Tpo -c -o zfs/ZfsArcMeter.o zfs/ZfsArcMeter.c &&\
    mv -f $depbase.Tpo $depbase.Po
depbase=`echo zfs/ZfsCompressedArcMeter.o | sed 's|[^/]*$|.deps/&|;s|\.o$||'`;\
    /usr/bin/gcc-4.2 -std=gnu99 -DHAVE_CONFIG_H -I.  -DNDEBUG -isystem/opt/local/include/LegacySupport -I/opt/local/include -std=c99 -pedantic -D_DARWIN_C_SOURCE -DNCURSES_WIDECHAR -Wall -Wcast-align -Wcast-qual -Wextra -Wfloat-equal -Wformat=2 -Winit-self -Wmissing-format-attribute -Wmissing-noreturn -Wmissing-prototypes -Wpointer-arith -Wshadow -Wstrict-prototypes -Wundef -Wunused -Wwrite-strings -DSYSCONFDIR="\"/opt/local/etc\"" -I"./darwin" -pipe -Os -arch ppc -MT zfs/ZfsCompressedArcMeter.o -MD -MP -MF $depbase.Tpo -c -o zfs/ZfsCompressedArcMeter.o zfs/ZfsCompressedArcMeter.c &&\
    mv -f $depbase.Tpo $depbase.Po
darwin/DarwinProcess.c: In function ‘DarwinProcess_scanThreads’:
[...]

PR #1687, opened 2025-04-27
macOS: OSX 10.4 Tiger compatibility

glebm:
2 changes to make htop compatible with OSX 10.4 Tiger

1. `mach_host_self()` and `vm_page_size` are defined in `mach_init.h`.
2. `<libproc.h>` API is not available, related functionality is disabled.

htop builds and works but cannot display per-process CPU% and task%. This can hopefully be improved in the future but at least it builds for now.

image

/cc @barracuda156

### DIFF
--- a/darwin/DarwinProcess.c
+++ b/darwin/DarwinProcess.c
@@ -7,10 +7,15 @@ in the source distribution for its full text.
 
 #include "darwin/DarwinProcess.h"
 
+#include <AvailabilityMacros.h>
+#if MAC_OS_X_VERSION_MIN_REQUIRED >= 1050
 #include <libproc.h>
+#endif
+
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <mach/mach_init.h>
 #include <mach/mach.h>
 #include <sys/dirent.h>
 
@@ -105,6 +110,7 @@ static int DarwinProcess_compareByKey(const Process* v1, const Process* v2, Proc
 }
 
 static void DarwinProcess_updateExe(pid_t pid, Process* proc) {
+#if MAC_OS_X_VERSION_MIN_REQUIRED >= 1050
    char path[PROC_PIDPATHINFO_MAXSIZE];
 
    int r = proc_pidpath(pid, path, sizeof(path));
@@ -112,9 +118,11 @@ static void DarwinProcess_updateExe(pid_t pid, Process* proc) {
       return;
 
    Process_updateExe(proc, path);
+#endif
 }
 
 static void DarwinProcess_updateCwd(pid_t pid, Process* proc) {
+#if MAC_OS_X_VERSION_MIN_REQUIRED >= 1050
    struct proc_vnodepathinfo vpi;
 
    int r = proc_pidinfo(pid, PROC_PIDVNODEPATHINFO, 0, &vpi, sizeof(vpi));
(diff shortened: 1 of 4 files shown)


barracuda156:
@glebm We have a breakage outside of Tiger, at least up to 10.6: https://github.com/htop-dev/htop/issues/1686

BenBE:
Disabling huge chunks of the existing code may have very adverse effects. Please refine this patch in a way that it works more localized.

darwin/DarwinProcess.c
```diff
@@ -398,6 +408,9 @@ void DarwinProcess_setFromLibprocPidinfo(DarwinProcess* proc, DarwinProcessTable
    dpt->super.userlandThreads += pti.pti_threadnum; /*pti.pti_threads_user;*/
    dpt->super.totalTasks += pti.pti_threadnum;
    dpt->super.runningTasks += pti.pti_numrunning;
+#else
+   proc->taskAccess = false;
+#endif
```
BenBE, line 413: Try to fill all the other fields here too; and be it with sensible "defaults" like "no time passed" for "one task"/"none running". Mark a todo to implement these with older APIs as necessary.

darwin/DarwinProcess.c
```diff
@@ -510,6 +524,7 @@ void DarwinProcess_scanThreads(DarwinProcess* dp, DarwinProcessTable* dpt) {
 
    vm_deallocate(mach_task_self(), (vm_address_t) thread_list, sizeof(thread_port_array_t) * thread_count);
    mach_port_deallocate(mach_task_self(), task);
+#endif
```
BenBE, line 527: Same here. If there's not even essential information filled in we can also just declare things fully incompatible with such old versions of MacOS …

darwin/Platform.c
```diff
@@ -18,6 +18,7 @@ in the source distribution for its full text.
 #include <net/if_types.h>
 #include <net/route.h>
 #include <sys/socket.h>
+#include <mach/mach_init.h>
```
BenBE, line 21: Why is this here, when it's not used through new code? Or is was this missing before; and by which code?

BenBE (changes requested):
That's what I mean by "too broad" …

darwin/DarwinProcess.c
```diff
@@ -7,10 +7,15 @@ in the source distribution for its full text.
 
 #include "darwin/DarwinProcess.h"
 
+#include <AvailabilityMacros.h>
+#if MAC_OS_X_VERSION_MIN_REQUIRED >= 1050
 #include <libproc.h>
+#endif
```
Explorer09, line 13: See docs/styleguide.md. Conditional includes should be below headers that are always included.

glebm:
Thanks for the review, I'll get back to it next weekend. In the meantime, I figured out how to get some of the info on Roger (though it requires root for privileged processes).

BenBE:
> Thank you for the review, I'll get back to it next weekend.
Take the time you need. With #1688 at least the immediate breakage from 10.6 onward should be taken care of.

> In the meantime, I figured out how to get some of the info on Tiger (though it requires root for privileged processes).
This sound good to hear. If information are unavailable for some processes, this usually is not a problem. Some columns include some marker for "N/A" data that should be used in these cases.

PR #1691, opened 2025-05-01
Bring back conversion of process CPU time on macOS (#1638)

### LINKED ISSUE #1638: Htop 3.4.0-dev not showing CPU percentage on apple silicon

erikolofsson:
Fixes #1638


darwin/Platform.c
```diff
@@ -148,11 +148,15 @@ const MeterClass* const Platform_meterTypes[] = {
    NULL
 };
 
+static uint64_t Platform_nanosecondsPerMachTickNumer = 1;
+static uint64_t Platform_nanosecondsPerMachTickDenum = 1;
```
Explorer09, line 152:
```suggestion
static uint64_t Platform_nanosecondsPerMachTickDenom = 1;
```

darwin/PlatformHelpers.c
```diff
@@ -86,3 +86,36 @@ bool Platform_isRunningTranslated(void) {
...
+   // Kernel version 20.0.0 is macOS 11.0 (Big Sur)
+   bool isBuggedVersion = Platform_KernelVersionIsBetween((KernelVersion) {20, 0, 0}, (KernelVersion) {999, 999, 999});
+
+   if (isRunningUnderRosetta2 && isBuggedVersion) {
+      // In this case `mach_timebase_info` provides the wrong value, so we hard-code the correct factor,
+      // as determined from `mach_timebase_info` when the process running natively.
+      info = (mach_timebase_info_data_t) { .numer = 125, .denom = 3 };
+   } else {
+      // No workarounds needed, use the OS-provided value.
+      mach_timebase_info(&info);
```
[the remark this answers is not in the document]
erikolofsson: Fixed

darwin/Platform.c
erikolofsson: Fixed

darwin/Platform.c
```diff
@@ -168,6 +172,12 @@ bool Platform_init(void) {
    return true;
 }
 
+// Converts ticks in the Mach "timebase" to nanoseconds.
+// See `mach_timebase_info`, as used to define the `Platform_nanosecondsPerMachTick` constant.
+uint64_t Platform_machTicksToNanoseconds(uint64_t mach_ticks) {
+   return (mach_ticks * Platform_nanosecondsPerMachTickNumer) / Platform_nanosecondsPerMachTickDenom;
```
Explorer09, line 178: I wonder what we can do if the multiplication overflows here.
Assertion? Crash?

erikolofsson: We could fix the overflow, which occurs after 194 cpu-years with 125 / 3 as we have on arm64.

Option 1 - Use 128 bit math:
```
uint64_t Platform_machTicksToNanoseconds(uint64_t mach_ticks) {
   return (uint64_t)(((unsigned __int128)mach_ticks * Platform_nanosecondsPerMachTickNumer) / Platform_nanosecondsPerMachTickDenom);
}
```
Micro benchmark:
```
Apple M1:     3.75 ns per call
Xeon W-3275M: 9.38 ns per call 
```

Option 2 - Use floating point math when overflow occurs:
```
uint64_t Platform_machTicksToNanoseconds(uint64_t mach_ticks) {
   if (mach_ticks >= Platform_nanosecondsPerMachTickOverflow)
      return (double)mach_ticks * Platform_nanosecondsPerMachTick;
   return (mach_ticks * Platform_nanosecondsPerMachTickNumer) / Platform_nanosecondsPerMachTickDenom;
}
```
Micro benchmark:
```
Apple M1:     2.05 ns per call
Xeon W-3275M: 6.66 ns per call 
```

I'm leaning toward 128 bit math to keep the code simpler and the precision high for the full range. The performance difference should be a non-issue. The only question is if we need care about old compilers and 32 bit builds which might not support the 128 bit type?

erikolofsson: Never mind, this is faster and avoids overflow:
```
uint64_t Platform_machTicksToNanoseconds(uint64_t mach_ticks) {
    uint64_t ticks_quot = mach_ticks / Platform_nanosecondsPerMachTickDenom;
    uint64_t ticks_rem  = mach_ticks - ticks_quot * Platform_nanosecondsPerMachTickDenom;

    uint64_t part1 = ticks_quot * Platform_nanosecondsPerMachTickNumer;
    uint64_t part2 = (ticks_rem * Platform_nanosecondsPerMachTickNumer) / Platform_nanosecondsPerMachTickDenom;

    return part1 + part2;
}
```
Micro benchmark:
```
Apple M1:     1.36 ns per call
Xeon W-3275M: 9.18 ns per call 
```

darwin/Platform.c
```diff
@@ -168,6 +172,18 @@ bool Platform_init(void) {
    return true;
 }
 
+// Converts ticks in the Mach "timebase" to nanoseconds.
+// See `mach_timebase_info`, as used to define the `Platform_nanosecondsPerMachTick` constant.
+uint64_t Platform_machTicksToNanoseconds(uint64_t mach_ticks) {
+   uint64_t ticks_quot = mach_ticks / Platform_nanosecondsPerMachTickDenom;
+   uint64_t ticks_rem  = mach_ticks - ticks_quot * Platform_nanosecondsPerMachTickDenom;
```
Explorer09, line 179: Maybe keep it simple? (A compiler should recognize both as equivalent, but if it doesn't, my testing long ago showed that the `%` operator produces smaller code in `-Os` optimization.)

```suggestion
   uint64_t ticks_rem  = mach_ticks % Platform_nanosecondsPerMachTickDenom;
```

darwin/PlatformHelpers.c
```diff
@@ -86,3 +86,36 @@ bool Platform_isRunningTranslated(void) {
    }
    return ret;
 }
+
+void Platform_calculateNanosecondsPerMachTick(uint64_t* numer, uint64_t* denom) {
+   // Check if we can determine the timebase used on this system.
+   // If the API is unavailable assume we get our timebase in nanoseconds.
+#ifdef HAVE_MACH_TIMEBASE_INFO
+   /* WORKAROUND for `mach_timebase_info` giving incorrect values on M1 under Rosetta 2.
+    *    rdar://FB9546856 https://openradar.appspot.com/radar?id=5055988478509056
```
Explorer09, line 95: The URL was broken. Perhaps we can update the reference to http://www.openradar.appspot.com/FB9546856

darwin/Platform.c
BenBE: Actually there usually is a divmod intrinsic in the compiler that should kick in if both a division and a modulo are done on the same arguments near each other. That said, using `__uint128` is non-standard and except if there's a **really** good reason for it, we should avoid it.

darwin/PlatformHelpers.c
```diff
@@ -86,3 +86,36 @@ bool Platform_isRunningTranslated(void) {
...
+    *
+    *    We don't know exactly what feature/attribute of the M1 chip causes this mistake under Rosetta 2.
+    *    Until we have more Apple ARM chips to compare against, the best we can do is special-case
+    *    the "Apple M1" chip specifically when running under Rosetta 2.
+    */
+
+   bool isRunningUnderRosetta2 = Platform_isRunningTranslated();
+
+   // Kernel version 20.0.0 is macOS 11.0 (Big Sur)
+   bool isBuggedVersion = Platform_KernelVersionIsBetween((KernelVersion) {20, 0, 0}, (KernelVersion) {999, 999, 999});
```
Explorer09, line 105:
```suggestion
   bool isBuggedVersion = 0 <= Platform_CompareKernelVersion((KernelVersion) {20, 0, 0});

File: darwin/PlatformHelpers.c (hunk 5 above)
Comment (erikolofsson) [reply darwin/PlatformHelpers.c]: Fixed

**Comment by Explorer09** [comment]
I don't have any more comment with this code now. But perhaps is good to explain what's the difference between this PR and #1643. Does this supersede #1643?

Cc @aestriplex

File: darwin/PlatformHelpers.c [hunk 7]
```diff
@@ -86,3 +86,36 @@ bool Platform_isRunningTranslated(void) {
...
+    */
+
+   bool isRunningUnderRosetta2 = Platform_isRunningTranslated();
+
+   // Kernel version 20.0.0 is macOS 11.0 (Big Sur)
+   bool isBuggedVersion = 0 <= Platform_CompareKernelVersion((KernelVersion) {20, 0, 0});
+
+   if (isRunningUnderRosetta2 && isBuggedVersion) {
+      // In this case `mach_timebase_info` provides the wrong value, so we hard-code the correct factor,
+      // as determined from `mach_timebase_info` when the process running natively.
```
BenBE, line 109: Maybe better wording?
```suggestion
      // as determined from `mach_timebase_info` as if the process was running natively.
```

darwin/PlatformHelpers.c
```diff
@@ -86,3 +86,36 @@ bool Platform_isRunningTranslated(void) {
    }
    return ret;
 }
+
+void Platform_calculateNanosecondsPerMachTick(uint64_t* numer, uint64_t* denom) {
+   // Check if we can determine the timebase used on this system.
+   // If the API is unavailable assume we get our timebase in nanoseconds.
+#ifdef HAVE_MACH_TIMEBASE_INFO
```
BenBE, line 93: What about collecting the values from the OS for the timebase here (as the default path) with corrections following later in the code?

I.e.:
- Query factor when function available, otherwise hard-code 1/1
- Check for bugged version
- If affected, override the read value.

Currently we have tons of code depending on whether a certain function is available or not, when usually these checks should be kept as minimal in covered code as possible IMHO.

erikolofsson: Something like this?
(Also limited the workaround to only run on x86 executable)
```
void Platform_calculateNanosecondsPerMachTick(uint64_t* numer, uint64_t* denom) {
   // Check if we can determine the timebase used on this system.
   // If the API is unavailable assume we get our timebase in nanoseconds.
#ifdef HAVE_MACH_TIMEBASE_INFO
   mach_timebase_info_data_t info;
   mach_timebase_info(&info);

   *numer = info.numer;
   *denom = info.denom;
#else
   *numer = 1;
   *denom = 1;
#endif

#ifdef __x86_64__
   /* WORKAROUND for `mach_timebase_info` giving incorrect values on M1 under Rosetta 2.
    *    rdar://FB9546856 http://www.openradar.appspot.com/FB9546856
    *
    *    We don't know exactly what feature/attribute of the M1 chip causes this mistake under Rosetta 2.
    *    Until we have more Apple ARM chips to compare against, the best we can do is special-case
    *    the "Apple M1" chip specifically when running under Rosetta 2.
    */

   bool isRunningUnderRosetta2 = Platform_isRunningTranslated();

   // Kernel version 20.0.0 is macOS 11.0 (Big Sur)
   bool isBuggedVersion = 0 <= Platform_CompareKernelVersion((KernelVersion) {20, 0, 0});

   if (isRunningUnderRosetta2 && isBuggedVersion) {
      // In this case `mach_timebase_info` provides the wrong value, so we hard-code the correct factor,
      // as determined from `mach_timebase_info` as if the process was running natively.
      *numer = 125;
      *denom = 3;
   }
#endif
}
```

BenBE: Maybe a short note regarding the `#ifdef __x86_64__` implying to run "as x64, possibly emulated" to clarify when this correction is applied … Else this LGTM.

darwin/PlatformHelpers.c
erikolofsson: Fixed

darwin/PlatformHelpers.c
erikolofsson: Fixed

BenBE (approved):

darwin/Platform.c
```diff
@@ -168,6 +172,18 @@ bool Platform_init(void) {
    return true;
 }
 
+// Converts ticks in the Mach "timebase" to nanoseconds.
+// See `mach_timebase_info`, as used to define the `Platform_nanosecondsPerMachTick` constant.
+uint64_t Platform_machTicksToNanoseconds(uint64_t mach_ticks) {
+   uint64_t ticks_quot = mach_ticks / Platform_nanosecondsPerMachTickDenom;
+   uint64_t ticks_rem  = mach_ticks % Platform_nanosecondsPerMachTickDenom;
+
```
Explorer09, line 180:
```c
   assert(ticks_quot <= UINT64_MAX / Platform_nanosecondsPerMachTickNumer);
   assert(Platform_nanosecondsPerMachTickDenom <= UINT64_MAX / Platform_nanosecondsPerMachTickNumer);
```

These two constraints are what prevent the overflow from happening in this function. Not sure if these two assertions are worth adding into the function, but I mention these just for completeness sake.

The first assertion is related to a [time overflow](https://en.wikipedia.org/wiki/Time_formatting_and_storage_bugs) (like Y2K) that might be out of scope of htop.
The second assertion can always hold true when `mach_timebase_info_data_t` uses 32-bit integers for both numerator and denominator fields.

BenBE: They don't exactly hurt, but given this is what ultimately gives the time values used in htop, they should not straight up cause htop to terminate seemingly at random. Instead normal handling that causes a runtime warning (once) could be done instead …

Can you elaborate on the intuition of the second assertion?

Explorer09: @BenBE 
```text
ticks_rem < Platform_nanosecondsPerMachTickDenom ;
ticks_rem * Platform_nanosecondsPerMachTickNumer < Platform_nanosecondsPerMachTickDenom * Platform_nanosecondsPerMachTickNumer

When `Platform_nanosecondsPerMachTickDenom * Platform_nanosecondsPerMachTickNumer` is less than 2^64, `ticks_rem * Platform_nanosecondsPerMachTickNumer` will be less than 2^64 as well, i.e. never overflows.

Comment (erikolofsson) [reply darwin/Platform.c:180]: Do you want me to do any changes?

The overflow should only happen when the result in nano-seconds overflows, which would already be a problem in other places in the application. This is after 584 CPU-years.

Comment (BenBE) [reply darwin/Platform.c:180]: Optional. I mostly asked for some clarifications here.

If you decided to update the PR, be sure to avoid direct `assert`s (for the reasons mentioned above), and instead choose a way that is as obvious as possible, that this is some intentional error value.

Comment (Explorer09) [reply darwin/Platform.c:180]: @erikolofsson My view is that the time overflow bug that would happen 584 years later (the overflow on `part1`) is out of scope, but without an in-code comment, the `(ticks_rem * Platform_nanosecondsPerMachTickNumer)` in `part2` can be mistaken as prone to overflow. (It actually doesn't overflow for the reason I stated above.)

**Comment by Korrd** [comment]
Is this fix expected to be merged for next release? I just had to copy old 3.3.0 htop from another mac to mine in order to make it show the right CPU usage. Something like this that makes the software fail to a point it's not showing the right data should be fixed much quicker than several months, as meanwhile it makes it unreliable and unusable for a quite common use-case, which might make users look for alternatives instead, as having the code done right is useless if the product does not work as intended.

**Comment by erikolofsson** [comment]
Rebased on main (comment added to prevent GitHub from merging multiple force pushes)

File: darwin/Platform.c (hunk 9 above)
Comment (erikolofsson) [reply darwin/Platform.c]: I added your comment

**Comment by Explorer09** [comment]
@Korrd Please keep in mind I'm just a volunteer. While I can help evaluting the quality of the patches, I don't have all day to do it and I have no guarantee on which patch would be reviewed sooner or later. (And I'm not the maintainer of this project - this is just my personal position and not the maintainers'.)

**Comment by aestriplex** [comment]
> I don't have any more comment with this code now. But perhaps is good to explain what's the difference between this PR and #1643. Does this supersede #1643?
> 
> Cc @aestriplex

@Explorer09 yes, this PR supersede #1643. If it's one with anyone, I can close it

File: darwin/PlatformHelpers.h [hunk 10]
```diff
@@ -29,6 +30,8 @@ int Platform_CompareKernelVersion(KernelVersion v);
 // lowerBound <= currentVersion < upperBound
 bool Platform_KernelVersionIsBetween(KernelVersion lowerBound, KernelVersion upperBound);
 
+void Platform_calculateNanosecondsPerMachTick(uint64_t* numer, uint64_t* denom);
```
Explorer09: Since you are only calling this `Platform_calculateNanosecondsPerMachTick()` from `Platform_init()`, I wonder if it's possible to make the former function `static` scope and not expose this as a public interface in `PlatformHelpers.h`?

darwin/PlatformHelpers.c
```diff
@@ -86,3 +86,42 @@ bool Platform_isRunningTranslated(void) {
...
+   mach_timebase_info_data_t info;
+   mach_timebase_info(&info);
```
BenBE: This probably should check for errors, AFAICS [here](https://developer.apple.com/documentation/driverkit/3433733-mach_timebase_info).. Falling back to some "sane" timebase on error, won't hurt. Otherwise the timebase is undefined.

darwin/PlatformHelpers.c
```diff
@@ -86,3 +86,42 @@ bool Platform_isRunningTranslated(void) {
...
+    *    rdar://FB9546856 http://www.openradar.appspot.com/FB9546856
+    *
+    *    We don't know exactly what feature/attribute of the M1 chip causes this mistake under Rosetta 2.
+    *    Until we have more Apple ARM chips to compare against, the best we can do is special-case
+    *    the "Apple M1" chip specifically when running under Rosetta 2.
+    */
+
+   bool isRunningUnderRosetta2 = Platform_isRunningTranslated();
+
+   // Kernel version 20.0.0 is macOS 11.0 (Big Sur)
```
BenBE: Maybe clarify the comparison:
```suggestion
   // Kernel versions >= 20.0.0 (macOS 11.0 AKA Big Sur) affected
```

Is this really intended to mark all "future" versions as buggy?

BenBE (commented):
Two minor. Rest seems fine.

Explorer09:
I slightly reordered the logic of the `Platform_calculateNanosecondsPerMachTick` function:

```c
static void Platform_calculateNanosecondsPerMachTick(uint64_t* numer, uint64_t* denom) {
   // Check if we can determine the timebase used on this system.
   // If the API is unavailable assume we get our timebase in nanoseconds.
   *numer = 1;
   *denom = 1;

#ifdef __x86_64__
   /* WORKAROUND for `mach_timebase_info` giving incorrect values on M1 under Rosetta 2.
    *    rdar://FB9546856 http://www.openradar.appspot.com/FB9546856
    *
    *    We don't know exactly what feature/attribute of the M1 chip causes this mistake under Rosetta 2.
    *    Until we have more Apple ARM chips to compare against, the best we can do is special-case
    *    the "Apple M1" chip specifically when running under Rosetta 2.
    *
    *    Rosetta 2 only supports x86-64, so skip this workaround when building for other architectures.
    */

   bool isRunningUnderRosetta2 = Platform_isRunningTranslated();

   // Kernel version 20.0.0 is macOS 11.0 (Big Sur)
   bool isBuggedVersion = 0 <= Platform_CompareKernelVersion((KernelVersion) {20, 0, 0});

   if (isRunningUnderRosetta2 && isBuggedVersion) {
      // In this case `mach_timebase_info` provides the wrong value, so we hard-code the correct factor,
      // as determined from `mach_timebase_info` as if the process was running natively.
      *numer = 125;
      *denom = 3;
      return;
   }
#endif

#ifdef HAVE_MACH_TIMEBASE_INFO
   mach_timebase_info_data_t info;
   if (mach_timebase_info(&info) == KERN_SUCCESS) {
      *numer = info.numer;
      *denom = info.denom;
      return;
   }
#endif
}
```

I feel the code path like this would be easier to follow. LGTM for other parts of the code.

BenBE:
> I slightly reordered the logic of the `Platform_calculateNanosecondsPerMachTick` function:
> 
> I feel the code path like this would be easier to follow. LGTM for other parts of the code.

Minor caveat, as with the current way the `return;`s are written, it looks like there's a path without return value, which looks odd. Instead, a slightly clearer variant to order would be:

```c
static void Platform_calculateNanosecondsPerMachTick(uint64_t* numer, uint64_t* denom) {
   // Check if we can determine the timebase used on this system.

#ifdef __x86_64__
   /* WORKAROUND for `mach_timebase_info` giving incorrect values on M1 under Rosetta 2.
    *    rdar://FB9546856 http://www.openradar.appspot.com/FB9546856
    *
    *    We don't know exactly what feature/attribute of the M1 chip causes this mistake under Rosetta 2.
    *    Until we have more Apple ARM chips to compare against, the best we can do is special-case
    *    the "Apple M1" chip specifically when running under Rosetta 2.
    *
    *    Rosetta 2 only supports x86-64, so skip this workaround when building for other architectures.
    */

   bool isRunningUnderRosetta2 = Platform_isRunningTranslated();

   // Kernel version 20.0.0 is macOS 11.0 (Big Sur)
   bool isBuggedVersion = 0 <= Platform_CompareKernelVersion((KernelVersion) {20, 0, 0});

   if (isRunningUnderRosetta2 && isBuggedVersion) {
      // In this case `mach_timebase_info` provides the wrong value, so we hard-code the correct factor,
      // as determined from `mach_timebase_info` as if the process was running natively.
      *numer = 125;
      *denom = 3;
      return;
   }
#endif

#ifdef HAVE_MACH_TIMEBASE_INFO
   mach_timebase_info_data_t info = { 0 };
   if (mach_timebase_info(&info) == KERN_SUCCESS) {
      *numer = info.numer;
      *denom = info.denom;
      return;
   }
#endif

   // No info on actual timebase found; assume timebase in nanoseconds.
   *numer = 1;
   *denom = 1;
}
```

That way, the fallback is clearly visible. This also highlights, that both `numer` and `denom` need to be assigned for each return path; although with this version it's no longer enforced as well as with the other variant.

darwin/PlatformHelpers.h
erikolofsson: In that case I would have to move it to Platform.c as it's in another compilation unit otherwise. That file is already rather long. Should I?

darwin/PlatformHelpers.c
erikolofsson: Yes, at this point I don't think this will change, and rosetta is going away soon.

erikolofsson:
> That way, the fallback is clearly visible. This also highlights, that both `numer` and `denom` need to be assigned for each return path; although with this version it's no longer enforced as well as with the other variant.

I applied the suggested fixes

darwin/PlatformHelpers.h
BenBE: I see no problem with that. Especially as this allows to reduce visibility/exposed interface for a function that is unlikely to get re-used. And regarding large files: There are far more complex ones. And just for comparison: The platform implementation (main part) for Linux is twice the size … What matters more is keeping code that belongs together and is used as an unit near each other; thus with this just being called on platform initialization and all other code related to timebase conversion in `Platform.c` it's more logical to also have this function over there.

erikolofsson: Ok, pushed change.

darwin/Platform.c
```diff
@@ -171,6 +217,21 @@ bool Platform_init(void) {
...
+   uint64_t part1 = ticks_quot * Platform_nanosecondsPerMachTickNumer;
+
+   // When Platform_nanosecondsPerMachTickDenom * Platform_nanosecondsPerMachTickNumer is less than 2^64, ticks_rem *
+   // Platform_nanosecondsPerMachTickNumer will be less than 2^64 as well, i.e. never overflows.
+   uint64_t part2 = (ticks_rem * Platform_nanosecondsPerMachTickNumer) / Platform_nanosecondsPerMachTickDenom;
+
+   return part1 + part2;
+}
+
```
BenBE: Group this above `Platform_init` to have related code closer together.

erikolofsson: Ok, I moved it up

Korrd:
Thanks a lot guys! <3

My love to you all!

PR #1620, opened 2025-02-23
GPUMeter code refactor & macOS support

Explorer09:
This is my attempt of refactoring the GPU meter code in order to support multiple platforms.
This branch includes #1618 which fixes a bug I discovered during the refactor process.
I adapted @aestriplex 's code in #1606 in the last commit of this branch. Yes I made the meter working in both bar and graph modes. (I don't want to take full credit for the last commit. If permitted, I wish to change the last commit so that @aestriplex is listed as the main author and I as a co-author.)

PR #1692, opened 2025-05-03
Preliminary support for NVIDIA Jetson boards

dmitriy-philimonov:
NVIDIA Jetson device is an insdustrial Linux based embedded aarch64 platfrom with powerful builtin GPU, which is used for AI tasks, mostly for CV purposes.

The support is provided via --enable-nvidia-jetson switch in the configure script.

All the source code related to the NVIDIA Jetson is placed in the linux/NvidiaJetson.{h,c} source files and hidden by 'NVIDIA_JETSON' C preprocessor define. So, for x86_64 platforms the source code stays unchanged.

Additional functionality added by this commit:
1. Fix for the CPU temperature reading. The Jetson device is not supported by libsensors. The CPU has 8 cores with only one CPU temperature sensor for all of them located in the thermal zone file. libsensor might be compiled in or turned off. The additional care was taken to provide successfull build with/without libsensors.
2. The Jetson GPU Meter was added: current load, frequency and temperature.

== Technical details ==

The code tries to find out the correct sensors during the application startup. As an example, the sensors location for NVIDIA Jetson Orin are the following:
- CPU temperature: /sys/devices/virtual/thermal/thermal_zone0/type
- GPU temperature: /sys/devices/virtual/thermal/thermal_zone1/type
- GPU frequency: /sys/class/devfreq/17000000.gpu/cur_freq
- GPU curr load: /sys/class/devfreq/17000000.gpu/device/load

Measure:
- The GPU frequency is provided in Hz, shown in MHz.
- The CPU/GPU temperatures are provided in Celsius multipled by 1000 (milli Celsius), shown in Cesius
- The Farenheit support is not provided

P.S. The GUI shows all temperatures for NVIDIA Jetson with additional precision comparing to the default x86_64 platform.

== NVIDIA Jetson models  ==

Tested for NVIDIA Jetson Orin and Xavier boards.


Explorer09:
I fear the option of `--enable-nvidia-jetson` will make future board-specific customizations add similar configure options. That would make things unmaintainable.

Explorer09:
Another problem is the conflict with #1620, which is an attempt to unify the GPU meter structure to one interface.

linux/NvidiaJetson.c
```diff
@@ -0,0 +1,188 @@
...
+   RichString_writeAscii(out, CRT_colors[METER_TEXT], ":");
+   xSnprintf(buffer, sizeof(buffer), "%.1f", this->values[JETSON_GPU_LOAD]);
+   RichString_appendAscii(out, CRT_colors[METER_VALUE], buffer);
+
+   RichString_appendAscii(out, CRT_colors[METER_TEXT], " freq:");
+   xSnprintf(buffer, sizeof(buffer), "%3uMHz", (unsigned)this->values[JETSON_GPU_FREQ]);
+   RichString_appendAscii(out, CRT_colors[METER_VALUE], buffer);
+
+   RichString_appendAscii(out, CRT_colors[METER_TEXT], " temp:");
+   xSnprintf(buffer, sizeof(buffer), "%.1f°C", this->values[JETSON_GPU_TEMP]);
```
Explorer09, line 164: Use `CRT_degreeSign` rather than hard-code a degree sign here.

linux/NvidiaJetson.c
```diff
@@ -0,0 +1,188 @@
...
+
+      xSnprintf(path, sizeof(path), "/sys/devices/virtual/thermal/%s/type", name);
+      ret = xReadfile(path, content, sizeof(content));
+      if (ret <= 0)
+         continue;
+
+      content[0] = tolower(content[0]);
+      content[1] = tolower(content[1]);
+      content[2] = tolower(content[2]);
+      content[3] = tolower(content[3]);
```
Explorer09, line 102: Why case conversion? Is there any reason for the letter case to vary?

dmitriy-philimonov: Yeah, even Jetson Xavier and Jetson Orin has different sensor names. NVIDIA breaks backward compatibility here.

linux/NvidiaJetson.c
dmitriy-philimonov: Sure, going to fix it.

dmitriy-philimonov:
> Another problem is the conflict with #1620, which is an attempt to unify the GPU meter structure to one interface.

I looked through 'main' branch implementation of the GpuMeter. If I understand correctly, it collects information about the GPU usage from each running process.

NVIDIA Jetson has a different approach - it provides a separate GPU statistics via sysfs / custom nvgpu driver.

You could merge the final version of a PR 1620 first, then I'll figure out how to reuse it correctly, on the next big holidays :)

dmitriy-philimonov:
> I fear the option of `--enable-nvidia-jetson` will make future board-specific customizations add similar configure options. That would make things unmaintainable.

The main purpose of the commit was to minimize the interference with the major code base for the default x86_64 platform. Honestly, I do not want to compile in the nvidia jetson board specific code anywhere else.

What approach would you recommend here?

Explorer09:
> > I fear the option of `--enable-nvidia-jetson` will make future board-specific customizations add similar configure options. That would make things unmaintainable.
> 
> The main purpose of the commit was to minimize the interference with the major code base for the default x86_64 platform. Honestly, I do not want to compile in the nvidia jetson board specific code anywhere else.
> 
> What approach would you recommend here?

There are two ideas that came in my mind.

1. The more ideal one: Make the board identifier part of the machine type, so we can have `--host=aarch64-nvidiajetson-linux-gnu`. But that requires your toolchain to be configured with the same machine type identifier, which is sometimes not feasible.
2. The less ideal, but easier approach: name the the configure option as `--with-board=nvidia_jetson`. This assumes that htop would accept patches for additional board customizations, and I don't know the maintainers' attitude on this.

dmitriy-philimonov:
> > > I fear the option of `--enable-nvidia-jetson` will make future board-specific customizations add similar configure options. That would make things unmaintainable.
> > 
> > 
> > The main purpose of the commit was to minimize the interference with the major code base for the default x86_64 platform. Honestly, I do not want to compile in the nvidia jetson board specific code anywhere else.
> > What approach would you recommend here?
> 
> There are two ideas that came in my mind.
> 
> 1. The more ideal one: Make the board identifier part of the machine type, so we can have `--host=aarch64-nvidiajetson-linux-gnu`. But that requires your toolchain to be configured with the same machine type identifier, which is sometimes not feasible.
> 2. The less ideal, but easier approach: name the the configure option as `--with-board=nvidia_jetson`. This assumes that htop would accept patches for additional board customizations, and I don't know the maintainers' attitude on this.
> 
> Update: Oh no. Nvidia didn't use a unique machine type for their GCC cross-toolchain. [Reference](https://docs.nvidia.com/jetson/archives/r36.4.3/DeveloperGuide/AT/JetsonLinuxToolchain.html)

@BenBE , as a maintainer, are you agree? If so, I will fix according to the idea №2.

BenBE:
There's some internal discussion still going on. We're still discussing which direction we'd like to move forward in.

dmitriy-philimonov:
Pushed new feature, need to rebase over the current trunk...

Feature details follows.

If htop starts with root privileges (effective user id is 0), the experimental code activates. It reads the fixed sysfs file /sys/kernel/debug/nvmap/iovmm/clients with the following content, e.g.:
```
CLIENT                        PROCESS      PID        SIZE
user                         gpu_burn     7979   23525644K
user                      gnome_shell     8119       5800K
user                             Xorg     2651      17876K
total                                            23549320K
```
Unfortunately, the /sys/kernel/debug/* files are allowed to read only for the root user, that's why the restriction applies.

The patch also adds a separate field 'GPU_MEM', which reads data from the added LinuxProcess::gpu_mem field. The field stores memory allocated for GPU in kilobytes. It is populated by the function NvidiaJetson_LoadGpuProcessTable (the implementation is located in NvidiaJetson.c), which is called at the end of the function Machine_scanTables.

Additionally, the new Action is added: actionToggleGpuFilter, which is activated by 'g' hot key (the help is updated appropriately). The GpuFilter shows only the processes which currently utilize GPU (i.e. highly extended nvmap/iovmm/clients table). It is achieved by the filtering machinery associated with ProcessTable::pidMatchList. The code below constructs 
 PU_PID_MATCH_LIST hash table, then actionToggleGpuFilter either stores it to the ProcessTable::pidMatchList or restores old value of ProcessTable::pidMatchList.

The separate LinuxProcess's PROCESS_FLAG_LINUX_GPU_JETSON (or something ...) flag isn't added for GPU_MEM, because currently the functionality of population LinuxProcess::gpu_mem is shared with the GPU consumers filter construction.
So, even if GPU_MEM field is not activated, the filter showing GPU consumers should work. This kind of architecture is chosen intentially since it saves memory for the hash table GPU_PID_MATCH_LIST (which is now actually a set), and therefore increases performance. All other approaches convert GPU_PID_MATCH_LIST to a true key/value storage (key = pid,
value = gpu memory allocated) with further merge code.

dmitriy-philimonov:
Changes:
1. Rebased. Honestly, I've left all GPU-related code unchanged. It utilized the different kernel API, might work with nvidia jetson one day, who knows?
2.  Additionally pushed the per process GPU memory allocation functionality right into the LinuxProcess class / GPU_MEM field in main screen. Marked it as experimental, because it works with root privileges only. In short, it reads the special sysfs file inside kernet/debug directory which is published by nvgpu nvidia driver, where the dictionary {pid -> gpu_memory} is published.

Added the Action for this functionality. Pressing 'g' hot key the main screen shows only the processes which uses GPU right now. Having the GPU_MEM field, you see the current GPU load per process. Useful, I guess. Hope, you'll utilize the same approach in your future development.

dmitriy-philimonov:
I've left all the deep details in both: the commit message and the NvidiaJetson.c file. Have a look, please. @BenBE 

Finally, with "Jetson GPU" Meter and "g" hot key applied, with GPU_MEM field, the "htop" looks like this:
image

CPUMeter.c
```diff
@@ -98,17 +98,30 @@ static void CPUMeter_updateValues(Meter* this) {
...
+      double cpuTemperature = this->values[CPU_METER_TEMPERATURE];
+      if (settings->degreeFahrenheit) {
+         c = 'F';
+         cpuTemperature = ConvCelsiusToFahrenheit(cpuTemperature);
+      }
+      /* snprintf correctly represents double NAN numbers as 'nan' */
+      xSnprintf(cpuTemperatureBuffer, sizeof(cpuTemperatureBuffer), "%.1f%s%c", cpuTemperature, CRT_degreeSign, c);
+   }
+   #else
    #ifdef BUILD_WITH_CPU_TEMP
```
BenBE: Maybe flatten into one `#if` block as:

```c
#if defined(BUILD_WITH_CPU_TEMP) && !defined(NVIDIA_JETSON)
   …
#elif defined(NVIDIA_JETSON)
   …
#endif
```

Machine.c
```diff
@@ -6,6 +6,7 @@ Released under the GNU GPLv2+, see the COPYING file
 in the source distribution for its full text.
 */
 
+#include "NvidiaJetson.h"
```
BenBE: Header order

MainPanel.c
```diff
@@ -195,6 +195,11 @@ static void MainPanel_drawFunctionBar(Panel* super, bool hideFunctionBar) {
    if (this->state->pauseUpdate) {
       FunctionBar_append("PAUSED", CRT_colors[PAUSED]);
    }
+#ifdef NVIDIA_JETSON
```
BenBE, line 201: Add a blank line before this block.

XUtils.h
```diff
@@ -123,6 +123,13 @@ char* xStrndup(const char* str, size_t len);
 
 ATTR_NONNULL ATTR_ACCESS3_W(2, 3)
 ssize_t xReadfile(const char* pathname, void* buffer, size_t count);
+#ifdef NVIDIA_JETSON /* uncomment if you need this functionality somewhere else */
```
BenBE: I think without the conditional define here is fine.

Maybe rename to `xReadNumberFile`?

linux/LinuxMachine.c
```diff
@@ -741,15 +742,20 @@ void Machine_scan(Machine* super) {
    const Settings* settings = super->settings;
    if (settings->showCPUFrequency
 #ifdef HAVE_SENSORS_SENSORS_H
-       || settings->showCPUTemperature
+       || settings->showCPUTemperature /* TODO: looks like this line in the condition might be removed */
```
BenBE: Can you explain your comment on this line?

linux/LinuxMachine.c
```diff
@@ -741,15 +742,20 @@ void Machine_scan(Machine* super) {
...
 #endif
    )
       LinuxMachine_scanCPUFrequency(this);
 
+   #ifdef NVIDIA_JETSON
+   if (settings->showCPUTemperature)
+      NvidiaJetson_getCPUTemperatures(this->cpuData, super->existingCPUs);
+   #else
    #ifdef HAVE_SENSORS_SENSORS_H
```
BenBE: See flattening comment regarding these defines above.

linux/LinuxProcess.c
```diff
@@ -6,6 +6,7 @@ Released under the GNU GPLv2+, see the COPYING file
 in the source distribution for its full text.
 */
 
+#include "NvidiaJetson.h"
```
BenBE: Header order.

linux/LinuxProcess.c
```diff
@@ -112,6 +113,9 @@ const ProcessFieldData Process_fields[LAST_PROCESSFIELD] = {
 #endif
    [GPU_TIME] = { .name = "GPU_TIME", .title = "GPU_TIME ", .description = "Total GPU time", .flags = PROCESS_FLAG_LINUX_GPU, .defaultSortDesc = true, },
    [GPU_PERCENT] = { .name = "GPU_PERCENT", .title = " GPU% ", .description = "Percentage of the GPU time the process used in the last sampling", .flags = PROCESS_FLAG_LINUX_GPU, .defaultSortDesc = true, },
+#ifdef NVIDIA_JETSON
+   [GPU_MEM]= { .name = "GPU_MEM", .title = "GPU_M ", .description = "GPU memory allocated for the process", .flags = 0, .defaultSortDesc = true, },
```
BenBE:
```suggestion
   [GPU_MEM] = { .name = "GPU_MEM", .title = "GPU_M ", .description = "GPU memory allocated for the process", .flags = 0, .defaultSortDesc = true, },
```

linux/NvidiaJetson.c
```diff
@@ -0,0 +1,322 @@
...
+#include "Meter.h"
+#include "XUtils.h"
+#include "linux/NvidiaJetson.h"
+#include "linux/LinuxProcess.h"
+
+
+#include <ctype.h>
+#include <dirent.h>
+#include <stdio.h>
+#include <string.h>
```
BenBE: Header order.
```suggestion
#include <ctype.h>
#include <dirent.h>
#include <stdio.h>
#include <string.h>

#include "CRT.h"
#include "Hashtable.h"
#include "Macros.h"
#include "Meter.h"
#include "XUtils.h"

#include "linux/NvidiaJetson.h"
#include "linux/LinuxProcess.h"
```

linux/NvidiaJetson.h
```diff
@@ -0,0 +1,19 @@
+#ifndef HEADER_NVIDIA_JETSON
+#define HEADER_NVIDIA_JETSON
+
+#include "config.h"
+#ifdef NVIDIA_JETSON
```
BenBE: No need to define the interface conditional. Just expose the interface, as linking will fail if otherwise unavailable.

Also, don't reference `config.h` inside header files. It should be included by the file including the header, as `config.h` needs to be the overall first header to be included.

CPUMeter.c
Explorer09: What's the relationship between the macro `NVIDIA_JETSON` and `BUILD_WITH_CPU_TEMP`? At a glance they don't seem to be mutually exclusive.

XUtils.h
Explorer09: I don't like how the function is currently designed. What if the temporary string buffer is not large enough to hold the number to be scanned?

CPUMeter.c
BenBE: AFAICS, `BUILD_WITH_CPU_TEMP` is set, whenever you are building with temperature sensor support (i.e. `libsensors` or similar source), It's also set, when building with NVIDIA Jetson support.

You also could check for `libsensors` in the first block AFAICS, but that should be properly evaluated it that hits all cases.

XUtils.h
BenBE: You could even work with a fixed stack buffer inside the function IFF you properly check for the file read to be untruncated (which `xReadFile` does).

CPUMeter.c
dmitriy-philimonov: * The CPU temperature showing code is rewritten.
* The duplication is removed.
* Additional settings showCPUTemperatureFractional is added.
* All decisions are explained in comments.

Machine.c
dmitriy-philimonov: Fixed.

MainPanel.c
dmitriy-philimonov: Fixed.

XUtils.h
dmitriy-philimonov: Fixed. Renamed to `xReadNumberFile`.

linux/LinuxMachine.c
dmitriy-philimonov: 1. Removed `/* TODO: ... */` comment. Let's keep backward compatibility.
2. Explanation.

The semantics is broken. If `showCPUTemperature` is activated, the application starts to read CPU frequencies from the system via `LinuxMachine_scanCPUFrequency` function. There's no connection with the CPU temperature. CPU frequency is not shown in case if `showCPUFrequency` is turned off. No commentary why it's done. Looks like some old forgotten hack (maybe there was a connection cpu_frequency with cpu_temperature some time ago, or other platforms have/had these connection).

linux/LinuxMachine.c
dmitriy-philimonov: Fixed

linux/LinuxProcess.c
dmitriy-philimonov: Fixed

linux/LinuxProcess.c
dmitriy-philimonov: Fxied

linux/NvidiaJetson.c
dmitriy-philimonov: Fixed

linux/NvidiaJetson.h
dmitriy-philimonov: Fixed

dmitriy-philimonov:
Fixed issues, @BenBE, have a look, please.

linux/LinuxMachine.c
BenBE: The code inside the `linux` directory only affects Linux itself, thus there should be no connection with other platforms here. And if the code guarded by this condition no longer has any references to CPU temperatures, then this condition can be removed.

IIRC libsensors is used for CPU temperature stuff, but this was refactored some time ago. I thus think it's more likely a remnant from that refactoring.

Action.c
```diff
@@ -646,6 +646,29 @@ static Htop_Reaction actionTogglePauseUpdate(State* st) {
    return HTOP_REFRESH | HTOP_REDRAW_BAR | HTOP_KEEP_FOLLOWING;
 }
 
+#ifdef NVIDIA_JETSON
+#include "NvidiaJetson.h"
+#include "ProcessTable.h"
```
BenBE: Includes should be kept near the top of the file.

CRT.h
```diff
@@ -158,6 +158,9 @@ typedef enum ColorElements_ {
    DYNAMIC_MAGENTA,
    DYNAMIC_YELLOW,
    DYNAMIC_WHITE,
+#ifdef NVIDIA_JETSON
+   GPU_FILTER,
+#endif
```
BenBE: No need to define this conditional. Also, probably group near the `GPU_ENGINE_x` fields.

CRT.c
```diff
@@ -231,6 +231,9 @@ static int CRT_colorSchemes[LAST_COLORSCHEME][LAST_COLORELEMENT] = {
       [DYNAMIC_MAGENTA] = ColorPair(Magenta, Black),
       [DYNAMIC_YELLOW] = ColorPair(Yellow, Black),
       [DYNAMIC_WHITE] = ColorPair(White, Black),
+#ifdef NVIDIA_JETSON
+      [GPU_FILTER] = A_BOLD | ColorPair(Red, Cyan),
+#endif
```
BenBE: Needs to be inserted for every color scheme. Also see the note regarding conditional defines/ordering.

XUtils.h
```diff
@@ -178,4 +185,9 @@ static inline int xDirfd(DIR* dirp) {
    return r;
 }
 
+
```
BenBE: Just one blank line between functions.

linux/NvidiaJetson.c
```diff
@@ -0,0 +1,321 @@
...
+
+#include "CRT.h"
+#include "Hashtable.h"
+#include "Macros.h"
+#include "Meter.h"
+#include "XUtils.h"
+
+#include "linux/NvidiaJetson.h"
+#include "linux/LinuxProcess.h"
```
BenBE: Missed that ordering issue …

```suggestion
#include "linux/LinuxProcess.h"
#include "linux/NvidiaJetson.h"
```

linux/NvidiaJetson.c
```diff
@@ -0,0 +1,321 @@
...
+      IsRootUser = geteuid() == 0;
+#endif
+      if (!IsRootUser) return;
+
+      GPU_PID_MATCH_LIST = Hashtable_new(MAX_GPU_PROCESS_COUNT, false);
+   }
+
+#ifdef DEBUG_ME
+   FILE *input = fopen("/home/philimonov/tmp/test.file", "r");
+#else
```
BenBE: Debugging code …

linux/NvidiaJetson.c
```diff
@@ -0,0 +1,321 @@
...
+   /* needs root privileges */
+   if (!IsRootUser) return;
+
+   /* first time function is called */
+   if (IsRootUser == -1) {
+#ifdef DEBUG_ME
+      IsRootUser = 1;
+#else
+      IsRootUser = geteuid() == 0;
+#endif
```
BenBE: Debug code …

linux/NvidiaJetson.c
```diff
@@ -0,0 +1,321 @@
...
+*/
+
+
+/* global paths per each sensor */
+char CPU_TEMP_SENSOR_PATH[64];
+char GPU_TEMP_SENSOR_PATH[64];
+char GPU_FREQ_SENSOR_PATH[64];
+char GPU_LOAD_SENSOR_PATH[64];
+
+
```
BenBE: Highlevel blocks only have one blank between them. Cf. style guide. Only exception is directly after the headers, where it's two blanks.

linux/NvidiaJetson.c
```diff
@@ -0,0 +1,321 @@
...
+/* #define DEBUG_ME 1 */
```
BenBE: Debugging code …

linux/NvidiaJetson.c
```diff
@@ -0,0 +1,321 @@
...
+   /* construct new knowledge regarding GPU allocations */
+   static const char sentinel = -128;
+   static const size_t line_sz = 256;
+
+   char line[line_sz], *last = &line[sizeof(line)-1]; *last = sentinel;
+   char *saveptr, *endptr;
+   while(fgets(line, sizeof(line), input)) {
+      if(String_startsWith(line, "CLIENT")) continue; // skip header
+      if(String_startsWith(line, "total")) break; //final line, skip for now
+      if(*last == '\0') {
```
BenBE: Space after keywords …

```suggestion
   while (fgets(line, sizeof(line), input)) {
      if (String_startsWith(line, "CLIENT")) // skip header
         continue;
      if (String_startsWith(line, "total")) //final line, skip for now
         break;

      if (*last == '\0') {
```

linux/NvidiaJetson.c
```diff
@@ -0,0 +1,321 @@
...
+      char *memPtr = strtok_r(NULL, " \n", &saveptr);
+      uint64_t gpumem = strtoull(memPtr, &endptr, 10);
+      /* memory allocation showed in kylobytes*/
+      if (endptr && *endptr != 'K') {
+         switch (*endptr) {
+            case 'M': gpumem *= 1024; break;
+            case 'G': gpumem *= 1024*1024; break;
+            default: gpumem = 0; break;
+         }
+      }
```
BenBE: Please provide a reference line of what the parsed content is expected to look like.

linux/NvidiaJetson.c
```diff
@@ -0,0 +1,321 @@
...
+   /* reset all knowledge about GPU allocations */
+   Hashtable_foreach(pidHash, NVidiaJetsonHashtableFunctor_ResetGpuMem, NULL);
+   Hashtable_clear(GPU_PID_MATCH_LIST);
+
+   /* construct new knowledge regarding GPU allocations */
+   static const char sentinel = -128;
+   static const size_t line_sz = 256;
+
+   char line[line_sz], *last = &line[sizeof(line)-1]; *last = sentinel;
+   char *saveptr, *endptr;
```
BenBE: Scope can be reduced to inside the loop near where it's first used.

linux/NvidiaJetson.c
```diff
@@ -0,0 +1,321 @@
...
+
+   /* reset all knowledge about GPU allocations */
+   Hashtable_foreach(pidHash, NVidiaJetsonHashtableFunctor_ResetGpuMem, NULL);
+   Hashtable_clear(GPU_PID_MATCH_LIST);
+
+   /* construct new knowledge regarding GPU allocations */
+   static const char sentinel = -128;
+   static const size_t line_sz = 256;
+
+   char line[line_sz], *last = &line[sizeof(line)-1]; *last = sentinel;
```
BenBE: No VLA. One variable declaration per line. No extra commands on declaration lines.

linux/NvidiaJetson.c
```diff
@@ -0,0 +1,321 @@
...
+
+      GPU_PID_MATCH_LIST = Hashtable_new(MAX_GPU_PROCESS_COUNT, false);
+   }
+
+#ifdef DEBUG_ME
+   FILE *input = fopen("/home/philimonov/tmp/test.file", "r");
+#else
+   FILE *input = fopen("/sys/kernel/debug/nvmap/iovmm/clients", "r");
+#endif
+   if (input == NULL) return;
```
BenBE:
```suggestion
   if (input == NULL)
      return;
```

linux/NvidiaJetson.c
```diff
@@ -0,0 +1,321 @@
...
+      if(String_startsWith(line, "total")) break; //final line, skip for now
+      if(*last == '\0') {
+         /* overflow */
+         *last = sentinel;
+         continue;
+      }
+      /* char *user = */ strtok_r(line, " \n", &saveptr);
+      /* char *proc = */ strtok_r(NULL, " \n", &saveptr);
+      char *pidPtr = strtok_r(NULL, " \n", &saveptr);
+      unsigned int pid = strtoul(pidPtr, NULL, 10);
```
BenBE: Probably use the `fast_strtoull` implementation, which is used elsewhere for parsing long lists of numbers.

linux/NvidiaJetson.c
```diff
@@ -0,0 +1,321 @@
...
+
+      char *memPtr = strtok_r(NULL, " \n", &saveptr);
+      uint64_t gpumem = strtoull(memPtr, &endptr, 10);
```
BenBE: Same note re `fast_strtoull` here …

linux/NvidiaJetson.c
```diff
@@ -0,0 +1,321 @@
...
+value = gpu memory allocated) with further merge code.
+
+*/
+
+
+/* global paths per each sensor */
+char CPU_TEMP_SENSOR_PATH[64];
+char GPU_TEMP_SENSOR_PATH[64];
+char GPU_FREQ_SENSOR_PATH[64];
+char GPU_LOAD_SENSOR_PATH[64];
```
BenBE: Only pre-processor macros should have uppercase names.

Also not quite happy with fixed-size buffers for filenames here.

linux/NvidiaJetson.c
```diff
@@ -0,0 +1,321 @@
...
+   }
+   closedir(dir);
+}
+
+
+static void NvidiaJetson_GoThroughThermalZones(void) {
+   char path[64], content[4];
+   const struct dirent* entry;
+
+   DIR* dir = opendir("/sys/devices/virtual/thermal");
```
BenBE: Probably use a define for the base path and reuse later when constructing paths inside this directory to avoid typos.

linux/NvidiaJetson.c
```diff
@@ -0,0 +1,321 @@
...
+void NvidiaJetson_FindSensors(void) {
+   NvidiaJetson_GoThroughThermalZones();
+   NvidiaJetson_FindGPUDevice();
+}
+
+
+void NvidiaJetson_getCPUTemperatures(CPUData* cpus, unsigned int existingCPUs) {
+   char buffer[22];
+   double temp = xReadNumberFile(CPU_TEMP_SENSOR_PATH, buffer, sizeof(buffer)) / 1000.0;
+   for (unsigned int i=0; i<=existingCPUs; ++i)
```
BenBE:
```suggestion
   for (size_t i = 0; i <= existingCPUs; i++)
```

linux/NvidiaJetson.c
```diff
@@ -0,0 +1,321 @@
...
+
+
+static void JetsonGPUMeter_updateValues(Meter* this) {
+   char buffer[22];
+   this->values[JETSON_GPU_LOAD] = xReadNumberFile(GPU_LOAD_SENSOR_PATH, buffer, sizeof(buffer));
+   this->curItems = 1; /* only show bar for JETSON_GPU_LOAD */
+
+   this->values[JETSON_GPU_TEMP] = xReadNumberFile(GPU_TEMP_SENSOR_PATH, buffer, sizeof(buffer)) / 1000.0;
+   this->values[JETSON_GPU_FREQ] = xReadNumberFile(GPU_FREQ_SENSOR_PATH, buffer, sizeof(buffer)) / 1000000.0;
+   double percent = this->values[0] / 10.0;
```
BenBE: Inconsistent use of indices into `this->values`; probably meant `JETSON_GPU_LOAD`?

BenBE (changes requested):

Action.c
dmitriy-philimonov: Fixed

CRT.h
dmitriy-philimonov: Fixed

CRT.c
dmitriy-philimonov: Fixed

XUtils.h
dmitriy-philimonov: Fixed

linux/NvidiaJetson.c
dmitriy-philimonov: Fixed

linux/NvidiaJetson.c
dmitriy-philimonov: Removed

linux/NvidiaJetson.c
dmitriy-philimonov: Removed

linux/NvidiaJetson.c
dmitriy-philimonov: Removed

linux/NvidiaJetson.c
dmitriy-philimonov: Removed

linux/NvidiaJetson.c
dmitriy-philimonov: Fixed

linux/NvidiaJetson.c
dmitriy-philimonov: Provided

linux/NvidiaJetson.c
dmitriy-philimonov: Done

linux/NvidiaJetson.c
dmitriy-philimonov: Fixed

linux/NvidiaJetson.c
dmitriy-philimonov: Fixed

linux/NvidiaJetson.c
dmitriy-philimonov: Introduced

linux/NvidiaJetson.c
dmitriy-philimonov: Fixed

linux/NvidiaJetson.c
dmitriy-philimonov: Renamed, changed to the xMalloc'ed strings

linux/NvidiaJetson.c
dmitriy-philimonov: Done

linux/NvidiaJetson.c
dmitriy-philimonov: Formatted

linux/NvidiaJetson.c
dmitriy-philimonov: Fixed

XUtils.c
```diff
@@ -263,6 +263,30 @@ int xSnprintf(char* buf, size_t len, const char* fmt, ...) {
...
+   size_t slen = n + 1;
+   void *data = malloc(slen);
+   if (!data) {
+      fail();
+   }
+
+   memcpy(data, buf, slen);
+   return data;
+}
+
```
BenBE: Isn't this a case for `xAsprintf`?

Machine.c
```diff
@@ -129,4 +130,8 @@ void Machine_scanTables(Machine* this) {
 
    Row_setUidColumnWidth(this->maxUserId);
    Row_setPidColumnWidth(this->maxProcessId);
+
+   #ifdef NVIDIA_JETSON
+   NvidiaJetson_LoadGpuProcessTable(this->activeTable->table);
+   #endif
```
BenBE: Given the Jetson code is placed in the Linux platform directory, I'd've expected this call somewhere inthe LinuxMachine code …

linux/NvidiaJetson.c
```diff
@@ -0,0 +1,321 @@
...
+   while ((entry = readdir(dir)) != NULL) {
+      const char* name = entry->d_name;
+
+      if (name[0] == '.' && (name[1] == '\0' || (name[1] == '.' && name[2] == '\0')))
+         continue;
+
+      if (IsJetsonOrinGPU(name) || IsJetsonXavierGPU(name)) {
+         char path[64];
+         GpuFreqSensorPath = xSnprintfDup(path, sizeof(path), SYS_CLASS_DEVFREQ "/%s/cur_freq", name);
+         GpuLoadSensorPath = xSnprintfDup(path, sizeof(path), SYS_CLASS_DEVFREQ "/%s/device/load", name);
```
BenBE: Yep, definitively `xAsprintf` …

With a `free` before to avoid memory leaks …

XUtils.c
dmitriy-philimonov: Thanks! Really, haven't seen the asprintf() before :)
Non POSIX, but pretty useful.

Machine.c
dmitriy-philimonov: Moved to the LinuxProcessTable::ProcessTable_goThroughEntries.

linux/NvidiaJetson.c
dmitriy-philimonov: Sure.

XUtils.c
BenBE: You rarely ever see it in code, but when you see it, it's quite helpful.

There's often the question regarding whether or not to force that one additional allocation, but when not inside a tight loop, that's usually fine.

XUtils.h
```diff
@@ -178,4 +186,23 @@ static inline int xDirfd(DIR* dirp) {
    return r;
 }
 
+static inline double ConvCelsiusToFahrenheit(const double celsius) {
```
BenBE:
```suggestion
static inline double convertCelsiusToFahrenheit(const double celsius) {
```

dmitriy-philimonov:
Polite ping, @BenBE

BenBE:
> Polite ping, @BenBE

NP. Haven't forgotten you, just busy right now (day job). Will take a look as things calm down again; this will take a few weeks though.

Issue #1678, opened 2025-04-16
Slow startup when many threads exist


abbeyj:
This may be related to #1535.

When you have a large number of threads running (e.g. 10000), htop takes a noticeable amount of time to start up and display the first output.  

I have the settings "Hide kernel threads" and "Hide userland process threads" turned on.  I have the setting "Highlight out-dated/removed programs (red) / libraries (yellow)" turned off.  I am not displaying the M_LRS column.  I'm measuring startup time with `time ./htop < /dev/null`.  To create a lot of threads for testing, I'm using `perl -Mthreads -e 'sub thd { sleep; } threads->create(\&thd) for (1 .. 8000); print "All threads created...\n"; sleep;'`

When there are only a handful of threads running, htop will start up quickly, in 0.1s or less.  When many threads are running, this gets slower, taking a second or several seconds.  I think there are few related factors that are contributing to this.

First, `LinuxProcessTable_recurseProcTree` will try to recurse into the `task` directory.  At the top level this makes sense because `/proc/N/task` will exist and will contain a list of the threads for PID N.  At the next level, this will try to open `/proc/N/task/M/task`.  At least on my system this never exists and always fails with ENOENT.  All those extra syscalls trying to check for a directory that does not exist add up  Changing https://github.com/htop-dev/htop/blob/987a47f7677fee51bede186efd04e2763d60e3fa/linux/LinuxProcessTable.c#L1603 to

```C
      if (!mainTask)
        LinuxProcessTable_recurseProcTree(this, procFd, lhost, "task", lp);
```

short-circuits this and avoids the extra open attempts.  This shows an improvement of 15% or so for me.  It may show a bigger effect if there is a virus scanner hooking into all `open` calls and adding extra overhead to them.  If there is some situation in which the task directory might show up at the second level then this change wouldn't work.

Second, `Machine_scanTables` is called 3 times on startup.  Two calls are from `CommandLine_run` and one is from `ScreenManager_run` => `checkRecalculation`.  The two calls in `CommandLine_run` are intentional according to eb196f8ba490536dc0bd26a3f316014af1823be3.  The fact that this is being run 3 times instead of 2 is seemingly not intended though.  Avoiding the call in `checkRecalculation` looked difficult so I removed the second call in `CommandLine_run` with no apparent ill effects.  That resulted in another ~15% improvement on top of the first change.

[...]

PR #1696, opened 2025-05-09
Startup boost

BenBE:
Improve startup performance by optimizing how the process list is gathered and how often.

This PR is based on the suggestions from #1678 by @abbeyj.

While this PR is tagged for Linux, it also affects all other platforms due to the second commit. The first commit (improving the process list scanning) only affects Linux, but should be noticeable throughout htop's lifetime.

I'm open for further patches related to performance optimization, in particular ones that affect/improve the startup time.

### DIFF
--- a/CommandLine.c
+++ b/CommandLine.c
@@ -297,16 +297,6 @@ static CommandLineStatus parseArguments(int argc, char** argv, CommandLineSettin
    return STATUS_OK;
 }
 
-static void CommandLine_delay(Machine* host, unsigned long millisec) {
-   struct timespec req = {
-      .tv_sec = 0,
-      .tv_nsec = millisec * 1000000L
-   };
-   while (nanosleep(&req, &req) == -1)
-      continue;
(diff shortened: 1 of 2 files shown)


CommandLine.c
```diff
@@ -398,8 +398,6 @@ int CommandLine_run(int argc, char** argv) {
    Machine_scan(host);
    Machine_scanTables(host);
    CommandLine_delay(host, 75);
-   Machine_scan(host);
-   Machine_scanTables(host);
```
natoscott, line 401: Pretty sure you can drop the delay on line 400 if we go down this path...?  I've always wondered why this second scan/scanTable was there in the original code - perhaps to guarantee values are fully setup (two samples mean all counters are rate converted and ready to display when the UI is presented).  But, I'm all for this change if it works.

natoscott: If really needed, could the delay perhaps be reduced from 75 to 1 (ms)?

BenBE: A reduction should be fine.

PR #1697, opened 2025-05-09
Warp open source software support!

ericdachen:


Explorer09:
Is this a legitimate sponsorship offer or is this a spam? Personally I don't like a banner of the sponsor to be placed on the top of the README file. It looks suspicious.

BenBE:
> Is this a legitimate sponsorship offer or is this a spam?

This is legitimate. We've been in contact with them for a bit already.

Have a look at [our HCB account](https://hcb.hackclub.com/htop) for reference.

natoscott:
@ericdachen thanks for your generous donation!  We're looking into ways we can more publicly show our appreciation.  I think something like the way the shields.io project does it may suit better [ https://shields.io/community ] but we'll continue discussing.

natoscott (changes requested):
I think we should approach this via a separate htop.dev page showing appreciation for all contributors (starting with our sponsors), as mentioned here and in separate mail.

PR #1698, opened 2025-05-10
Use pkg-config to detect libnl header and library paths

### LINKED ISSUE #1639: Regression in libnl3 detection (no longer uses pkg-config)
Hi!

We had a report in Gentoo of htop-3.4.0 failing to build on Gentoo Prefix (https://bugs.gentoo.org/951290), which is like a Homebrew/nix/whatever setup where Gentoo resides in `/home/sam/prefix` as an unprivileged prefix.

Since 24b1513296fd61722166625ad46be1c56a5efc44, configure fails with:
```
[...]
checking for netlink/attr.h... no
configure: error: can not find required header files netlink/attr.h, netlink/handlers.h, netlink/msg.h
```

... because it now hardcodes `CFLAGS="$CFLAGS -I/usr/include/libnl3"` rather than using pkg-config via `PKG_CHECK_MODULES`.

Explorer09:
Supersedes #1637. Fixes #1639.


configure.ac
```diff
@@ -1033,39 +1033,100 @@ case "$enable_capabilities" in
...
+         fi
+
+         if test "x$LIBNL3_LIBDIR" = x; then
+            AC_MSG_RESULT([(default)])
+         else
+            # The path must end with a slash
+            LIBNL3_LIBDIR=`echo "x$LIBNL3_LIBDIR" | sed 's/^x//; s|/*$|/|'`
+            AC_MSG_RESULT([$LIBNL3_LIBDIR])
+         fi
+         AC_DEFINE_UNQUOTED([LIBNL3_LIBDIR], ["$LIBNL3_LIBDIR"], [libnl-3 search path; use the default search paths if empty])
```
BenBE, line 1118: Is this the right way to define this macro? What about path's like `/usr/lib/malicious.so"); (void)StuffToFixSyntax("`?

Explorer09: @BenBE Perhaps forbidding the quote characters (`"`) can help. Technically this is not our bug - It would be the builder that injects the vulnerability and not we that do.

configure.ac
```diff
@@ -1033,39 +1033,100 @@ case "$enable_capabilities" in
...
+      htop_save_CFLAGS=$CFLAGS
+      # New include path searched after what user has specified
+      CFLAGS="$CFLAGS $LIBNL3_CFLAGS"
+      AC_CHECK_HEADERS([netlink/attr.h netlink/handlers.h netlink/msg.h],
+         [],
+         [if test "$enable_delayacct" = yes; then
+            AC_MSG_ERROR([can not find required header files netlink/attr.h, netlink/handlers.h, netlink/msg.h])
+         fi
+         enable_delayacct=no
+         break])
```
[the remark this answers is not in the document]
Explorer09: Can we stop Copilot from making false claims about the code errors? The `break` statement inside `AC_CHECK_HEADERS` was _intentional_ and its allowed in the documentation. ([Autoconf documentation](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.72/html_node/Generic-Headers.html#index-AC_005fCHECK_005fHEADERS-1) "You can give it a value of ‘`break`’ to break out of the loop on the first match.")

linux/LibNl.c
```diff
@@ -78,17 +78,17 @@ static int load_libnl(void) {
...
       if (!libnlHandle) {
          goto dlfailure;
       }
    }
 
-   libnlGenlHandle = dlopen("libnl-genl-3.so", RTLD_LAZY);
+   libnlGenlHandle = dlopen(LIBNL3_LIBDIR "libnl-genl-3.so", RTLD_LAZY);
    if (!libnlGenlHandle) {
-      libnlGenlHandle = dlopen("libnl-genl-3.so.200", RTLD_LAZY);
+      libnlGenlHandle = dlopen(LIBNL3_LIBDIR "libnl-genl-3.so.200", RTLD_LAZY);
```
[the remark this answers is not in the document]
Explorer09: Yes it is. The configure script ensured it. Copilot is making a false error claim.

configure.ac
BenBE: It is mostly a matter of proper escaping of variables. It should not fail to build, just because I have my libnl3 in `/usr/local/"special builds"/libs` … ;-)

Explorer09: @BenBE That is an undefined behavior AFAICT. At least, it's impossible to specify a path with spaces in an `-L` option to a `LDFLAGS` variable. You _may_ pass a path with spaces in an `-L` option when you run a linker directly, but the makefile format was never designed with that use case. 🙁

Paths with embedded quotation marks would be worse - variable expansions in makefiles would never work with those. The makefile format didn't come with a secure variable expansion by design, unlike variable expansion in shells. Thus if you are working with those paths, you would need the hell a lot of code hacking to make it work (at least you would modify `config.h` manually).

PR #1706, opened 2025-05-21
check totalCapacity before use it as a divisor

mugitya03:
In `Platform_getBattery`, `totalCapacity` is initialized to 0:

```c
intmax_t totalCapacity = 0;
```

It only increases inside the loop when both `isBattery` and `isPresent` are true:

```c
if (isBattery && isPresent) {
    totalCharge += curCharge;
    totalCapacity += maxCharge;
}
```

However, there are valid execution paths (no battery entries, an empty iterator, or batteries with zero capacity) where `totalCapacity` remains 0. The code then performs

```c
((double)totalCharge / (double)totalCapacity) * 100.0
```

which can divide by zero. To prevent this, I add a guard before the division:

```c
   if (totalCapacity != 0){
      *percent = ((double)totalCharge / (double)totalCapacity) * 100.0;
   }
```

### DIFF
--- a/netbsd/Platform.c
+++ b/netbsd/Platform.c
@@ -512,8 +512,9 @@ void Platform_getBattery(double* percent, ACPresence* isOnAC) {
          *isOnAC = isConnected ? AC_PRESENT : AC_ABSENT;
       }
    }
-
-   *percent = ((double)totalCharge / (double)totalCapacity) * 100.0;
+   if (totalCapacity != 0){
+      *percent = ((double)totalCharge / (double)totalCapacity) * 100.0;
+   }
 
 error:
    if (fd != -1)


Explorer09:
Floating point devision by zero is a defined behavior within IEEE 754. It results in NaN. So there isn't a bug here, although I can see it doesn't hurt to add an extra protection.

netbsd/Platform.c
```diff
@@ -512,8 +512,9 @@ void Platform_getBattery(double* percent, ACPresence* isOnAC) {
          *isOnAC = isConnected ? AC_PRESENT : AC_ABSENT;
       }
    }
-
-   *percent = ((double)totalCharge / (double)totalCapacity) * 100.0;
+   if (totalCapacity != 0){
```
Explorer09, line 515: Add a space here
```suggestion
   if (totalCapacity != 0) {
```

BenBE: Most other implementations just use a ternary `totalCapacity = 0 ? … : NAN`.

mugitya03: Thanks for the clarification. Our static analyzer flags any divide-by-zero as a potential bug, so we treat dividing by zero as an issue. I believe that adding this check will make the code more robust.

netbsd/Platform.c
```diff
@@ -512,8 +512,9 @@ void Platform_getBattery(double* percent, ACPresence* isOnAC) {
          *isOnAC = isConnected ? AC_PRESENT : AC_ABSENT;
       }
    }
-
-   *percent = ((double)totalCharge / (double)totalCapacity) * 100.0;
+   *percent = (totalCapacity > 0)
+    ? ((double)totalCharge / (double)totalCapacity) * 100.0
+    : NAN;
```
BenBE, line 517: Reasonably long lines are preferred.
```suggestion
   *percent = totalCapacity > 0 ? ((double)totalCharge / (double)totalCapacity) * 100.0 : NAN;
```

Technically you could even do:
```suggestion
   *percent = totalCapacity > 0 ? (100.0 * totalCharge) / totalCapacity : NAN;
```

Please squash commits. The commit history should tell what needed to change, not how you got there.

Explorer09: @BenBE Your latter suggestion changes semantics. I would rather not write that way. For floating points, division before multiplication is safer against overflow, and the rounding differences in the fractions part is negligible.

BenBE: The latter suggestion is the order used on Darwin and Linux. For OpenBSD the division is performed first.

Explorer09:
I thought the commits would be squashed before they are merged info `main`. Um, nevermind.

fasterit:
Yes, I thought so, too. That's what you get for reviewing a one line diff and assuming it would take one commit to get there. Sorry!

Issue #1673, opened 2025-04-14
Clang [-Wshorten-64-to-32] warnings in htop code


Explorer09:
This spreadsheet contains all `[-Wshorten-64-to-32]` warnings that I can find within htop codebase:

https://docs.google.com/spreadsheets/d/12P6VnjGBcqXA-vDI-lsBHiAsmJoruPdLBPfZ3TKNCbg/edit?usp=sharing

I didn't test all of the platforms supported in htop, but I have tried my best. 

Platforms tested: Linux, FreeBSD, OpenBSD, macOS (Darwin), PCP and Unsupported
Platforms not tested: NetBSD, DragonFly BSD and Solaris

There are <s>151</s> <s>134</s> 107 `[-Wshorten-64-to-32]` warnings in total in the attached spreadsheet.

[This page marks all lines with the warnings.](https://github.com/htop-dev/htop/commit/8400d8014ac6ef0976e8aa422f05cef4180e1f1f)

I have also tried categorizing the warnings on the real problems behind or how to fix them. The pull requests <s>#1588</s> (merged), <s>#1668</s> (merged) and <s>#1669</s> (merged) are parts of the solutions to these warnings.

#### "Cmdline"

(The PR 1588 has been merged.)

<s>Related to byte offsets and lengths of processes' "cmdline", "comm" and "exe" path strings. The ideal solution is to promote the type of these from `int` to `size_t`. I have proposed #1588 for this.</s>

#### "Terminal-width"

These variables or data have confusing uses of string length (`strlen` and friends) and the display width of the string (number of terminal columns the string can occupy, a la `wcswidth`). I cannot totally fix the problems in this category for now, as the reworking of certain functions to refer to widths instead can be difficult. I'm not sure if a quick workaround to silence the warning could be accepted. (The workaround I had in mind is using `(int)strnlen(str, INT_MAX)`.)

#### "strto*"

These are about conversion of `atol`, `strtol`, `strtoul`, `strtoull` etc. results to smaller integer types. There should be bound checks (or, for `fast-strto*` functions, assertions on the limits).

#### "Hashtable-key"

Related to bit widths allowed for `Hashtable` keys.

#### "PCP-Metric"

Related to the `Metric` data type in the PCP code. I won't fix this one as I have little knowledge with the PCP code.

campillumi:
I believe I might a solution to your problem. Check it out here [click](https://forreststonesolutions.com/robots/)

BenBE:
[...]

PR #1707, opened 2025-05-22
Introduce strnlen() and use it for some string width calculations

Explorer09:
The `strnlen` function does not calculate true display widths of the Unicode strings, but at least it works with ASCII strings. The function that can calculate display widths of Unicode strings is not implemented in htop yet (I have #1642, but the implementation of that is incomplete).

The goal of this commit is to prevent potential arithmetic overflows when calculating string widths and to allow a safe downcast from `size_t` to `int` type. This fixes some of the`-Wshorten-64-to-32` warnings produced by Clang (see #1673) as a result.


Meter.c
```diff
@@ -60,12 +60,12 @@ static void TextMeterMode_draw(Meter* this, int x, int y, int w) {
    }
    attrset(CRT_colors[RESET_COLOR]);
 
-   int captionLen = strlen(caption);
-   w -= captionLen;
-   if (w < 1) {
+   int captionWidth = (int)strnlen(caption, (unsigned int)w);
```
BenBE, line 63: Add an assert for `w >= 0` and cast the second argument to `size_t`.

Meter.c
```diff
@@ -340,11 +340,11 @@ static void LEDMeterMode_draw(Meter* this, int x, int y, int w) {
       mvaddnstr(yText, x, caption, w);
    }
 
-   int captionLen = strlen(caption);
-   if (w <= captionLen) {
+   int captionWidth = (int)strnlen(caption, (unsigned int)w);
```
BenBE, line 343:
```suggestion
   int captionWidth = (int)strnlen(caption, (size_t)w);
```

(Note the remark re an additional `assert` above)

Row.c
```diff
@@ -179,13 +179,13 @@ const char* RowField_alignedTitle(const Settings* settings, RowField field) {
 RowField RowField_keyAt(const Settings* settings, int at) {
    const RowField* fields = (const RowField*) settings->ss->fields;
    RowField field;
-   int x = 0;
+   int rem = at;
    for (int i = 0; (field = fields[i]); i++) {
-      int len = strlen(RowField_alignedTitle(settings, field));
-      if (at >= x && at <= x + len) {
+      int len = rem > 0 ? (int)strnlen(RowField_alignedTitle(settings, field), (unsigned int)rem) : 0;
```
BenBE, line 184:
```suggestion
      int len = rem > 0 ? (int)strnlen(RowField_alignedTitle(settings, field), (size_t)rem) : 0;
```

ScreenManager.c
```diff
@@ -166,11 +166,10 @@ static inline bool drawTab(const int* y, int* x, int l, const char* name, bool c
    (*x)++;
    if (*x >= l)
       return false;
-   int nameLen = strlen(name);
-   int n = MINIMUM(l - *x, nameLen);
+   int nameWidth = (int)strnlen(name, (unsigned int)(l - *x));
```
BenBE, line 169: Probably should `assert(l > 0);`. Also the parameter for `l` does not include the space used for the `[` and `]` from the tab drawing.

```suggestion
   int nameWidth = (int)strnlen(name, (size_t)(l - *x));
```

Explorer09: This assertion is probably not needed because of the `(*x >= l)` right above. It might not be obvious but `(*x < l)` → `(0 < l - *x)`.

Meter.c
Explorer09: 1. Nice catch on missing the `w >= 0` check. I add it now as a conditional (rather than an assertion).
2. The `(unsigned int)` cast was meant to be an optimization hint. I intended the cast to be `(size_t)(unsigned int)w`, but I except a decent compiler should be able to figure that out. I now make the cast implicit instead. (If a compiler can throw a warning on that, it's a compiler bug.)

ScreenManager.c
BenBE: Well, `0 < l - *x` does not imply `l > 0`, as `*x >= 0` isn't checked either …

Explorer09: Hmmm. Good catch. Although I believe the assertion should be a separate commit.

One more thing is I'm not sure if the compiler can catch that `l - *x > 0` and can optimise accordingly (on the bit extension). I am about to test that out and see whether GCC and Clang can optimise as expected or any of them has a bug.

BenBE: I think, that if compilers don't catch this micro-optimization, this doesn't hurt for this case much, as this routine isn't gonna be called thousands of times per second. Doesn't hurt to report those missed optimizations to compiler vendors though, but there's no need to explicitly make the code more complicated if they are missed.

Meter.c
Explorer09: And [here is a GCC bug report about sign extension missed optimization](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=120434).

ScreenManager.c
Explorer09: [Clang did miss this optimization](https://github.com/llvm/llvm-project/issues/142283).

Action.c
```diff
@@ -402,19 +402,27 @@ static Htop_Reaction actionPrevScreen(State* st) {
...
+   int rem = x - SCREEN_TAB_MARGIN_LEFT;
    for (unsigned int i = 0; i < settings->nScreens; i++) {
-      if (x < s) {
-         return 0;
-      }
       const char* tab = settings->screens[i]->heading;
-      int len = strlen(tab);
-      if (x < s + len + 2) {
+      int width = rem >= 2 ? (int)strnlen(tab, (unsigned int)(rem - 2 + 1)) : 0;
+      if (rem - 2 < width) {
```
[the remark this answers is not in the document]
Explorer09: I'm changing the conditional to `(width >= rem - /*bracketWidth*/2 + 1)` instead.

I really don't like the assumption of `SCREEN_TAB_MARGIN_LEFT > 0`. The `SCREEN_TAB_MARGIN_LEFT` constant is supposed to be adjustable and the real constraint for it should be `SCREEN_TAB_MARGIN_LEFT >= 0`.

BenBE:
When clicking the screen tabs there's a small off-by-one present:
- Clicking the cell right before the tab selects that tab, while clicking the  last cell of a tab does nothing (as expected of the gap between tabs).
- Clicking the gap selects the tab following the gap.

A similar issue exists when trying to select a column header:
- Clicking the last cell of a column header inverts and selects the column following it.
- Except when you are too far to the left, where it actually selects the right column once. Subsequent clicks select the column right of it.

Explorer09:
> When clicking the screen tabs there's a small off-by-one present:
> 
> * Clicking the cell right before the tab selects that tab, while clicking the  last cell of a tab does nothing (as expected of the gap between tabs).
> * Clicking the gap selects the tab following the gap.
> 
> A similar issue exists when trying to select a column header:
> 
> * Clicking the last cell of a column header inverts and selects the column following it.
> * Except when you are too far to the left, where it actually selects the right column once. Subsequent clicks select the column right of it.

Can't reproduce in my terminal emulator. They work perfectly on my side. Both on macOS Terminal and xfce4-terminal (on Devuan GNU/Linux).

BenBE:
Merging for now, as this happens for me also with the previous version without this PR applied.

PR #1713, opened 2025-06-01
Make Ctrl-U clear the current Filter or Search string while editing it

### LINKED ISSUE #1694: Filter Feature Request
Would it be possible to add a user preference to always clear the filter on F4. Currently anything you type there is never cleared and if you search for several things one after another you have to constantly backspace/delete the previous filter term over and over.

fasterit:
Closes #1694

### DIFF
--- a/IncSet.c
+++ b/IncSet.c
@@ -181,6 +181,13 @@ bool IncSet_handleKey(IncSet* this, int ch, Panel* panel, IncMode_GetPanelValue 
             }
          }
       }
+   } else if (ch == KEY_CTRL('U')) {
+      mode->index = 0;
+      mode->buffer[mode->index] = 0;
+      if (mode->isFilter) {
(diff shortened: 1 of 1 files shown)


BenBE (approved):
LGTM.

IncSet.c
```diff
@@ -181,6 +181,13 @@ bool IncSet_handleKey(IncSet* this, int ch, Panel* panel, IncMode_GetPanelValue
             }
          }
       }
+   } else if (ch == KEY_CTRL('U')) {
+      mode->index = 0;
+      mode->buffer[mode->index] = 0;
+      if (mode->isFilter) {
+            filterChanged = true;
+            this->filtering = false;
```
Explorer09: The two layers of indent here is probably unintentional.
```suggestion
         filterChanged = true;
         this->filtering = false;
```

PR #1714, opened 2025-06-01
Make Infoscreen also show uppercase FILTER when filtering (F4) is active

fasterit:
Tracescreen inherits this as well


BenBE (approved):
LGTM.

PR #1715, opened 2025-06-01
Make empty --filter= command line arguments invalid

fasterit:


BenBE (approved):
LGTM.

Took me a second though to see why you are checking for `'|'`.

Explorer09:
May I ask, why the checking for the `'|'` character?

fasterit:
because that's how you "OR" terms in the filter. Does not make sense as first char.

Explorer09:
> because that's how you "OR" terms in the filter. Does not make sense as first char.

If you want sanitization like that, then it would make no sense either with two consecutive ORs `|` in the middle or an OR in the end `|`.

`foo||bar`
`foo|bar|`

fasterit:
yes, and you would be the one to develop that into a regex engine over time. I'm not :rofl:

PR #1723, opened 2025-06-18
FreeBSD: Update the internal priority reference point

OlCe2:
This reference value has recently changed, in order for 0 and higher values to indicate timesharing/idletime priorities, and negative values are kernel or realtime priorities. A reference point is still needed as currently only the raw internal priorities, which are unsigned (uint8_t), can be communicated to userland.

### DIFF
--- a/freebsd/FreeBSDProcessTable.c
+++ b/freebsd/FreeBSDProcessTable.c
@@ -14,6 +14,7 @@ in the source distribution for its full text.
 #include <math.h>
 #include <stdlib.h>
 #include <string.h>
+#include <unistd.h>
 #include <sys/_iovec.h>
 #include <sys/errno.h>
 #include <sys/param.h> // needs to be included before <sys/jail.h> for MAXPATHLEN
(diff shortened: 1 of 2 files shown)


markjdb:
Isn't this too simple? You're checking the version of FreeBSD that htop was compiled against, not the version of the kernel, but the latter is what determines the scheduler behaviour.

freebsd/FreeBSDProcessTable.c
```diff
@@ -246,7 +246,14 @@ void ProcessTable_goThroughEntries(ProcessTable* super) {
...
 
-      proc->priority = kproc->ki_pri.pri_level - PZERO;
+      proc->priority = kproc->ki_pri.pri_level -
+          /* Reference point, as used by system's top(1) and ps(1). */
+#if __FreeBSD_version >= 1500048
+          PUSER
+#else
+          PZERO
+#endif
+          ;
```
BenBE: Please use the full statement inside the two `#ifdef` branches. I know that's repeated code, but it reads much easier that way.

Even the following would be possible.
```suggestion
      /* Reference point, as used by system's top(1) and ps(1). */
      proc->priority = kproc->ki_pri.pri_level - (__FreeBSD_version >= 1500048 ? PUSER : PZERO);
```

But also note the other comment about the running kernel's version …

OlCe2: Yes, I'm aware about the running's kernel version and see how to get the running kernel's version.

emaste: kern.osreldate sysctl description updated as a result of this PR -- https://github.com/freebsd/freebsd-src/commit/5110a74afe5680882e3274d57da881511ea7c440

OlCe2: Please see the newly pushed commit.

BenBE (approved):

OlCe2:
Thanks!

Issue #1603, opened 2025-02-08
Unicode "Grapheme Break" support (user-perceived character breaks, with libunistring)


Explorer09:
This is a discussion of a feature that would use a library.

htop often needs breaks of file paths of command lines, as well as other strings that may contain arbitrary Unicode text.

In discussions of #462 and #854, about shortening a path of a working directory for display, I think it becomes apparent that we need to support the Unicode algorithm of _user-perceived character_ boundaries.

In Unicode, this is defined in UAX 29 section 3:
https://www.unicode.org/reports/tr29/#Grapheme_Cluster_Boundaries
with a normative database:
https://www.unicode.org/Public/UCD/latest/ucd/auxiliary/GraphemeBreakProperty.txt

Not to be confused with Unicode line breaking algorithm ([UAX 14](https://www.unicode.org/reports/tr14/)).

_What are the "Grapheme Cluster Boundaries" in Unicode and why are they important for htop?_

When htop needs to truncate a file path or a command line argument string, it is supposed to  remove "characters" from the path string until the string fits the given terminal column width. However, Unicode characters (code points) do not always have one-to-one relationships to the "characters" that user could perceive as one unit.

Special cases can include:
* [Combining character sequences](https://en.wikipedia.org/wiki/Combining_character)
* Character sequences with ZWJ (Zero Width Joiner, U+200D), commonly found in emoji sequences.
* [Regional indicator symbols](https://en.wikipedia.org/wiki/Regional_indicator_symbol) (a.k.a. flag emojis). These regional indicator character should be in pairs and break cannot happen within each pair.
* Languages that use a syllable as the minimum grapheme unit. This is the least concern for htop but I mention for completeness.

I don't think people would like a series of United Kingdom flag emojis (🇬🇧🇬🇧🇬🇧🇬🇧) suddenly transform into Bulgaria flag emojis plus a letter B (🇧🇬🇧🇬🇧🇬🇧) when characters are cut at a wrong location. This is just an example.

_The grapheme cluster boundary algorithm requires a database,_ and I'm not wishing to embed the whole database into htop. What I am suggesting is to incorporate **[libunistring](https://www.gnu.org/software/libunistring/)**, a library from GNU that had the database for our needs. Specifically the [`unigbrk.h`](https://www.gnu.org/software/libunistring/manual/html_node/unigbrk_002eh.html) APIs.

[...]

PR #1726, opened 2025-06-22
CommandScreen: handle multi-byte UTF-8 code sequences while wrapping

ryenus:
In earlier implementation of the command screen, the process command was treated as a sequence of single-byte characters. When there are wide chars, especially UTF-8 byte sequences in the command, it's possible to have part of the multi-byte sequence wrapped to next line.

To fix that we now leverage [`mbrtowc`][1] and [`wcwidth`][2] to know the byte count and character width for non-ASCII characters, so that we can now wrap near window edge and more importantly at character boundaries.

@BenBE Would you mind having a look at this when you get a chance?

[1]: https://man7.org/linux/man-pages/man3/mbrtowc.3.html
[2]: https://man7.org/linux/man-pages/man3/wcwidth.3.html

### DIFF
--- a/CommandScreen.c
+++ b/CommandScreen.c
@@ -11,58 +11,125 @@ in the source distribution for its full text.
 #include "CommandScreen.h"
 
 #include <assert.h>
+#include <locale.h>
+#include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <wchar.h>
+#include <wctype.h>
 
+#include "CRT.h"
 #include "Macros.h"
 #include "Panel.h"
 #include "ProvideCurses.h"
 
 
-static void CommandScreen_scan(InfoScreen* this) {
-   Panel* panel = this->display;
-   int idx = MAXIMUM(Panel_getSelectedIndex(panel), 0);
-   Panel_prune(panel);
-
-   const char* p = Process_getCommand(this->process);
-
-   size_t line_maxlen = COLS < 40 ? 40 : COLS;
-   size_t line_offset = 0;
-   size_t last_space = 0;
-   char* line = xCalloc(line_maxlen + 1, sizeof(char));
-
-   for (; *p != '\0'; p++) {
-      if (line_offset >= line_maxlen) {
-         assert(line_offset <= line_maxlen);
-         assert(last_space <= line_maxlen);
-
-         size_t line_len = last_space <= 0 ? line_offset : last_space;
-         char tmp = line[line_len];
-         line[line_len] = '\0';
-         InfoScreen_addLine(this, line);
-         line[line_len] = tmp;
-
-         assert(line_len <= line_offset);
-         line_offset -= line_len;
-         memmove(line, line + line_len, line_offset);
-
-         last_space = 0;
+static int CommandScreen_scanAscii(InfoScreen* this, const char* p, size_t total, char* line) {
+   int line_offset = 0, line_size = 0, last_spc_offset = -1;
+   for (size_t i = 0; i < total; i++, line_offset++) {
+      assert(line_offset >= 0 && (size_t)line_offset <= total);
+      char c = line[line_offset] = p[i];
+      if (c == ' ') {
+         last_spc_offset = line_offset;
       }
 
-      line[line_offset++] = *p;
-      if (*p == ' ') {
-         last_space = line_offset;
+      if (line_offset == COLS) {
+         line_size = (last_spc_offset == -1) ? line_offset : last_spc_offset;
+         line[line_size] = '\0';
+         InfoScreen_addLine(this, line);
+
+         line_offset -= line_size;
+         last_spc_offset = -1;
+         memcpy(line, p + i - line_offset, line_offset + 1);
       }
    }
 
+   return line_offset;
+}
+
+#ifdef HAVE_LIBNCURSESW
+
+static int CommandScreen_scanWide(InfoScreen* this, const char* p, size_t total, char* line) {
+   mbstate_t state;
+   memset(&state, 0, sizeof(state));
+   int line_cols = 0, line_offset = 0, line_size = 0, width = 1;
+   int last_spc_cols = -1, last_spc_offset = -1;
+   for (size_t i = 0, bytes = 1; i < total; bytes = 1, width = 1) {
+      assert(line_offset >= 0 && (size_t)line_offset <= total);
+      unsigned char c = (unsigned char)p[i];
+      if (c < 0x80) { // skip mbrtowc for ASCII characters
+         line[line_offset] = c;
+         if (c == ' ') {
+            last_spc_offset = line_offset;
+            last_spc_cols = line_cols;
+         }
+      } else {
+         wchar_t wc;
+         bytes = mbrtowc(&wc, p + i, total - i, &state);
+         if (bytes != (size_t)-1 && bytes != (size_t)-2) {
+            width = wcwidth(wc);
+            width = MAXIMUM(width, 1);
+         } else {
+            bytes = 1;
+         }
+         memcpy(line + line_offset, p + i, bytes);
+      }
+
+      i += bytes;
+      line_offset += bytes;
+      line_cols += width;
+      if (line_cols < COLS) {
+         continue;
+      }
+
+      if (last_spc_offset >= 0) {  // wrap by last space
+         line_size = last_spc_offset;
+         line_cols -= last_spc_cols;
+         last_spc_offset = last_spc_cols = -1;
+      } else if (line_cols > COLS) {  // wrap current wide char to next line
+         line_size = line_offset - (int) bytes;
+         line_cols = width;
+      } else { // wrap by current character
+         line_size = line_offset;
+         line_cols = 0;
+      }
+
+      line[line_size] = '\0';
+      InfoScreen_addLine(this, line);
+      line_offset -= line_size;
+      if (line_offset > 0) {
+         memcpy(line, p + i - line_offset, line_offset + 1);
+      }
+   }
+
+   return line_offset;
+}
+
+#endif // HAVE_LIBNCURSESW
+
+static void CommandScreen_scan(InfoScreen* this) {
+   Panel* panel = this->display;
+   int idx = Panel_getSelectedIndex(panel);
+   Panel_prune(panel);
+
+   const char* p = Process_getCommand(this->process);
+   assert(p != NULL);
+   size_t total = strlen(p);
+   char line[total + 1];
+
+   int line_offset =
+#ifdef HAVE_LIBNCURSESW
+      CRT_utf8 ? CommandScreen_scanWide(this, p, total, line) :
+#endif
+      CommandScreen_scanAscii(this, p, total, line);
+
+   assert(line_offset >= 0 && (size_t)line_offset <= total);
    if (line_offset > 0) {
       line[line_offset] = '\0';
       InfoScreen_addLine(this, line);
    }
 
-   free(line);
-
-   Panel_setSelected(panel, idx);
+   Panel_setSelected(panel, MAXIMUM(idx, 0));
 }
 
 static void CommandScreen_draw(InfoScreen* this) {


Explorer09:
I have been trying to make a few PRs (pull requests) that could bring Unicode character width support in htop. So you are not the first one that proposed the idea.

There are many places in htop codebase that need to be upgraded for Unicode character width support. So your width calculation function would be better not limited to `CommandScreen.c` use.

To avoid duplicate effort, maybe you could look at my PR #1642 to see what you can do with the width calculation.

CommandScreen.c
```diff
@@ -49,21 +49,27 @@ static int CommandScreen_scanAscii(InfoScreen* this, const char* p, size_t total
...
+   int line_cols = 0, line_offset = 0, line_size = 0, width = 0;
+   int last_spc_cols = -1, last_spc_offset = -1;
    for (size_t i = 0; i < total; ) {
       assert(line_offset >= 0 && (size_t)line_offset <= total);
-      wchar_t wc;
-      size_t bytes = mbrtowc(&wc, p + i, total - i, &state);
-      int width = wcwidth(wc);
-      if (bytes == (size_t)-1 || bytes == (size_t)-2 || bytes == 0 || width < 0) {
+      unsigned char c = (unsigned char)p[i];
+      if (c < 0x80) { // skip mbrtowc for ASCII characters
```
Explorer09, line 60: This lacks checking or assertion on control characters.

CommandScreen.c
```diff
@@ -49,21 +49,27 @@ static int CommandScreen_scanAscii(InfoScreen* this, const char* p, size_t total
...
-
-      memcpy(line + line_offset, p + i, bytes);
-      if (wc == ' ') {
-         last_spc_offset = line_offset;
-         last_spc_cols = line_cols;
+         line[line_offset] = c;
+      } else {
+         wchar_t wc;
+         bytes = mbrtowc(&wc, p + i, total - i, &state);
+         width = wcwidth(wc);
```
Explorer09: Check `bytes` before calling `wcwidth`. If `mbrtowc` returns an error, `wc` can be left undefined.

ryenus: Good catch, thank you!

CommandScreen.c
ryenus: Could you please clarify what the expected behavior is? I’m asking because, as far as I know, it has always worked this way, even in the main screen. For now, I’d prefer to focus on resolving the line wrapping issue first—one step at a time. 😊

Explorer09: Code point range [0x00, 0x1F] and 0x7F are unprintable. You shouldn't assume the width would be 1 here.

CommandScreen.c
```diff
@@ -12,45 +12,129 @@ in the source distribution for its full text.
...
+   int last_spc_cols = -1, last_spc_offset = -1;
+   for (size_t i = 0; i < total; ) {
+      assert(line_offset >= 0 && (size_t)line_offset <= total);
+      unsigned char c = (unsigned char)p[i];
+      if (c < 0x80) { // skip mbrtowc for ASCII characters
+         if (c == ' ') {
+            last_spc_offset = line_offset;
+            last_spc_cols = line_cols;
+         }
+         bytes = width = 1;
```
Explorer09: Code style issue. Don't do multiple assignments like this, especially when the types of two variables are different (`size_t` and `int`).

Write the two assignments as separate lines.

CommandScreen.c
ryenus: Got you, appreciate your feedback, but again, for now I’d prefer to focus on resolving the line wrapping issue first.

Explorer09: I think your effort on trying to wrap the commnd line on space boundaries is unjustified. Unicode line breaking is more complicated than that, and in case of file paths, it might be more appropriate to wrap at path separators (that is, slashes) rather than spaces.

And since you are working with Unicode characters with variable display width, the issue #1603 would be relevant. I think you can read that first.

CommandScreen.c
ryenus: Updated, thank you

CommandScreen.c
ryenus: Thank you, but be it unjustified or justified, it varies depending on who's looking at it. While I understand you want to tackle all these related issues in broader scopes, however, I'd prefer to make incremental progress one after another, rather than trying to fix everything in one or a few relatively large patch(es), just one step at a time, like what I already said earlier.

ryenus:
@Explorer09, thanks a lot for sharing your thoughts and pointing me to your work on related topics. I really appreciate your insights and all the effort you’ve put into this area.

For this PR, I’m mainly focused on implementing proper line wrapping at or near the window border. I agree that a broader, unified solution for wide character support, especially an improved control character handling worth looking. However, I think those are best addressed in their own dedicated PRs and discussions, such as what you’re working on in #1642. For now, I’d prefer to keep the scope here limited so we can resolve this specific issue first.

Thanks again for your feedback and for helping move things forward!

CommandScreen.c
```diff
@@ -12,45 +12,124 @@ in the source distribution for its full text.
 
 #include <assert.h>
 #include <stdlib.h>
+#include <stdio.h>
 #include <string.h>
+#include <wchar.h>
+#include <wctype.h>
+#include <locale.h>
```
BenBE: Header order:

```suggestion
#include <assert.h>
#include <locale.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <wchar.h>
#include <wctype.h>
```

CommandScreen.c
```diff
@@ -12,45 +12,124 @@ in the source distribution for its full text.
...
+
+static void CommandScreen_scan(InfoScreen* this) {
+   Panel* panel = this->display;
+   int idx = Panel_getSelectedIndex(panel);
+   Panel_prune(panel);
+
+   const char* p = Process_getCommand(this->process);
+   assert(p != NULL);
+   size_t total = strlen(p);
+   char line[total + 1];
```
BenBE, line 118: Avoid variable length arrays.

CommandScreen.c
```diff
@@ -12,45 +12,124 @@ in the source distribution for its full text.
...
+         } else {
+            bytes = 1;
+         }
+         memcpy(line + line_offset, p + i, bytes);
+      }
+
+      i += bytes;
+      line_offset += bytes;
+      line_cols += width;
+      if (line_cols < COLS) {
```
BenBE, line 81: `COLS` should ideally be an argument to the string wrap function.

BenBE (commented):
I think, having this in one function is still reasonably readable:

```c
static int CommandScreen_scanString(InfoScreen* this, const char* p, size_t total, char* line, bool wide) {
   int line_offset = 0;
   int line_cols = 0;
   int last_spc_offset = -1;
   int last_spc_cols = -1;

#ifdef HAVE_LIBNCURSESW
   mbstate_t state;
   memset(&state, 0, sizeof(state));
#endif

   for (size_t i = 0; i < total;) {
      assert(line_offset >= 0 && (size_t)line_offset <= total);

      unsigned char c = (unsigned char)p[i];
      int bytes = 1;
      int width = 1;

      // Handle character encoding based on mode
      if (!wide || c < 0x80) {
         line[line_offset] = c;
         if (c == ' ') {
            last_spc_offset = line_offset;
            last_spc_cols = line_cols;
         }
      }
#ifdef HAVE_LIBNCURSESW
      else {
         wchar_t wc = 0;
         bytes = mbrtowc(&wc, p + i, total - i, &state);
         if (bytes != (size_t)-1 && bytes != (size_t)-2) {
            width = wcwidth(wc);
            width = MAXIMUM(width, 1);
         } else {
            bytes = 1;
         }

         assert(line_offset + bytes <= total + 1);
         assert(i + bytes <= total);
         memcpy(line + line_offset, p + i, bytes);
      }
#endif

      i += bytes;
      line_offset += bytes;
      line_cols += width;

      // Check if we need to wrap the line
      if (line_cols >= COLS) {
         int line_size;

         if (last_spc_offset >= 0) {
            // Wrap at last space
            line_size = last_spc_offset;
            line_cols -= last_spc_cols;
            last_spc_offset = last_spc_cols = -1;
         } else if (line_cols > COLS && wide) {
            // Current character would exceed COLS, wrap before it
            line_size = line_offset - bytes;
            line_cols = width;
         } else {
            // Wrap at current position
            line_size = line_offset;
            line_cols = 0;
         }

         line[line_size] = '\0';
         InfoScreen_addLine(this, line);

         // Move any remaining content to the beginning
         line_offset -= line_size;
         if (line_offset > 0) {
            memcpy(line, p + i - line_offset, line_offset);
         }
      }
   }

   return line_offset;
}

static void CommandScreen_scan(InfoScreen* this) {
   Panel* panel = this->display;
   int idx = Panel_getSelectedIndex(panel);
   Panel_prune(panel);

   const char* p = Process_getCommand(this->process);
   assert(p != NULL);
   size_t total = strlen(p);
   char line[total + 1];

   int line_offset = CommandScreen_scanString(this, p, total, line, CRT_utf8);

   assert(line_offset >= 0 && (size_t)line_offset <= total);
   if (line_offset > 0) {
      line[line_offset] = '\0';
      InfoScreen_addLine(this, line);
   }

   Panel_setSelected(panel, MAXIMUM(idx, 0));
}
```

Note this is mostly a brute-force mashup on your two functions with some restructuring, but unless I missed something this should work as expected.

The `else {` would for completeness actually be a `else if (wide) {` but the reasoning for why it's not needed to cover everything, should be easy to see.

A further refinement one could do is making adding a line governed by a callback `typedef void (* line_cb_t)(void* ctx, const char* line)`, with also the final line being handled as part of the string wrapping code.

ryenus:
@BenBE actually the code had been like that, in the sense that everything was in just one function, it was later split up :-) I just pushed the earlier commits for your reference.

CommandScreen.c
ryenus: This used to be `line[COL + 1]`, however due to it's possible for a unicode character to have arbitrary combining marks, like diacritics, therefore `COL + 1` is not sufficient and would cause segfault. therefore `strlen(p) + 1` is probably the only safe option out there.

With that said, I wonder if you mean to use malloc instead?

Explorer09: The `strlen(p)` is a bad idea as it scans the entire string rather than just stops at the potential line breaking position.

You can reference my code `String_mbswidth` in #1642 if you need some inspiration.

CommandScreen.c
Explorer09: We should `#include "ProvideCurses.h"` for wide character headers instead of including `wchar.h` and `wctype.h` directly.

CommandScreen.c
BenBE: The comment is about the VLA itself. And yes, a `malloc` is the preferred way.

That way you avoid having any dynamically sized buffers anywhere near the stack.

Explorer09: Note: I just don't like duplicated efforts. I remembered I have done this "calculate string width and length with respect to whitespace line break" before. It was in #1648. I just need to split that out into a dedicated function so that it can be reused in this PR.

BenBE: Preferably similar to the `String_split` function or with a callback as suggested.

ryenus: Haha, maybe we could just wait for String_split to magically appear :P

Jokes aside, I honestly don't have enough spare time right now to implement a String_split that would cover all use cases throughout the entire htop codebase. It seems like #1105 ran into similar challenges and ended up stalled, which is unfortunate.

Explorer09: There's an infrastructural change needed for htop to fully support Unicode.

BenBE: > Haha, maybe we could just wait for String_split to magically appear :P

There is an existing function `String_split`. I meant more like creating a `String_wrap` function with similar API.

> Jokes aside, I honestly don't have enough spare time right now to implement a String_split that would cover all use cases throughout the entire htop codebase.

If a first version only covers the cases used on `CommandScreen.c` that's totally fine. @Explorer09 can take up the work from there if you like.

> It seems like #1105 ran into similar challenges and ended up stalled, which is unfortunate.

Multibyte support is quite tricky to get right, in particular when dealing with input handling.

ryenus: > If a first version only covers the cases used on CommandScreen.c that's totally fine. @Explorer09 can take up the work from there if you like.

Yes, that would be great :-)

Explorer09: I made `String_lineBreakWidth()` in https://github.com/htop-dev/htop/pull/1648/commits/2ddc31180703939578cd45f6402c9628420d1756 that should be able to work with your use case.

BenBE: The stack buffer is still there …

PR #1728, opened 2025-06-24
pcp/Metric.c and pcp/Plaform.c: changes for pending PCP 7.0.0

kmcdonell:
timestamps in a pmResult will become struct timespec, not struct timeval, so small change to support both PMAPI_VERSION_2 (timeval) and PMAPI_VERSION_4 (timespec).


natoscott (approved):
LGTM, thanks @kmcdonell

PR #1731, opened 2025-06-27
Disable increasing numeric values on select clicks

fasterit:
This is a bit a preference issue. I find it odd that clicking "Update interval (in seconds)" increases this field (like any other numeric field) when you just select it with the mouse.

This PR disables that unexpected functionality.

On the other hand ... we toggle checkboxes on select click, too...

Discuss! :smiley:

### DIFF
--- a/DisplayOptionsPanel.c
+++ b/DisplayOptionsPanel.c
@@ -54,13 +54,10 @@ static HandlerResult DisplayOptionsPanel_eventHandler(Panel* super, int ch) {
       case ' ':
          switch (OptionItem_kind(selected)) {
             case OPTION_ITEM_TEXT:
+            case OPTION_ITEM_NUMBER:
                break;
             case OPTION_ITEM_CHECK:
                CheckItem_toggle((CheckItem*)selected);
-               result = HANDLED;
-               break;
-            case OPTION_ITEM_NUMBER:
-               NumberItem_toggle((NumberItem*)selected);
                result = HANDLED;
                break;
          }


BenBE:
I think given that you need some way to operate things "purely with your mouse", having some interaction when clicking is necessary. We might want to opt to disable the first click (which selects the item) from changing the value, but keep the update, when the click doesn't change the focus/selected item.

fasterit:
yeah, and may be make right click lower the number where left click increases it

fasterit:
"purely with the mouse" continues to work as you can click "F7Dec" and "F8Inc" fields with the mouse and they do the right thing™. Again. After you merged #1730 :smiley:.

Explorer09:
I think there are already UX researches related to numeric fields.
A summary opinion:
* Boolean fields (checkboxes) are okay when mouse click on the item toggles the checked status.
* Numeric field values should have explicit increment and decrement buttons (we have these already). A click on a numeric field should allow user to edit the value by typing on the keyboard. No value change if the user merely clicks on the field.
* Mouse right click is probably reserved for the function of opening a context menu in GUI. A context menu on a text field allows user to cut, copy or paste the value or do other things.

Explorer09:
I probably missed two more things in my last comment.

* The reason for right click not to do anything but opening the context menu is that not everyone operates their computer with a mouse. They may use touchpads or touch screens (in case of phones and tablets) and so user might not always have access to the "right click" as input. (Video games might be an exception when right mouse button can be assigned to do something, but this is not our case.)
* Mouse wheels can be use for increment and decrement of a numeric field. We could implement that, perhaps in a separate PR.

fasterit:
This has been superseded by #1732

PR #1732, opened 2025-06-27
Make first click select option line but not toggle it, toggle with next click, handle right click

fasterit:
... and assorted consistency improvements

Improved version of #1731 as discussed there

### DIFF
--- a/DisplayOptionsPanel.c
+++ b/DisplayOptionsPanel.c
@@ -49,8 +49,6 @@ static HandlerResult DisplayOptionsPanel_eventHandler(Panel* super, int ch) {
       case '\n':
       case '\r':
       case KEY_ENTER:
-      case KEY_MOUSE:
-      case KEY_RECLICK:
       case ' ':
          switch (OptionItem_kind(selected)) {
             case OPTION_ITEM_TEXT:
@@ -67,8 +65,12 @@ static HandlerResult DisplayOptionsPanel_eventHandler(Panel* super, int ch) {
          break;
       case '-':
       case KEY_F(7):
+      case KEY_RIGHTCLICK:
          if (OptionItem_kind(selected) == OPTION_ITEM_NUMBER) {
             NumberItem_decrease((NumberItem*)selected);
+            result = HANDLED;
+         } else if (OptionItem_kind(selected) == OPTION_ITEM_CHECK) {
+            CheckItem_set((CheckItem*)selected, false);
             result = HANDLED;
          }
          break;
@@ -76,6 +78,18 @@ static HandlerResult DisplayOptionsPanel_eventHandler(Panel* super, int ch) {
       case KEY_F(8):
          if (OptionItem_kind(selected) == OPTION_ITEM_NUMBER) {
             NumberItem_increase((NumberItem*)selected);
+            result = HANDLED;
+         } else if (OptionItem_kind(selected) == OPTION_ITEM_CHECK) {
+            CheckItem_set((CheckItem*)selected, true);
+            result = HANDLED;
+         }
+         break;
+      case KEY_RECLICK:
+         if (OptionItem_kind(selected) == OPTION_ITEM_NUMBER) {
+            NumberItem_increase((NumberItem*)selected);
+            result = HANDLED;
+         } else if (OptionItem_kind(selected) == OPTION_ITEM_CHECK) {
+            CheckItem_toggle((CheckItem*)selected);
             result = HANDLED;
          }
          break;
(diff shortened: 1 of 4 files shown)


DisplayOptionsPanel.c
```diff
@@ -67,16 +65,32 @@ static HandlerResult DisplayOptionsPanel_eventHandler(Panel* super, int ch) {
...
             NumberItem_increase((NumberItem*)selected);
             result = HANDLED;
+         } else if (OptionItem_kind(selected) == OPTION_ITEM_CHECK) {
+            CheckItem_set((CheckItem*)selected, true);
+            result = HANDLED;
+         }
+         break;
+      case KEY_RECLICK:
+         if (OptionItem_kind(selected) == OPTION_ITEM_NUMBER) {
+            NumberItem_increase((NumberItem*)selected);
```
BenBE, line 89: What about using:
Left click inc, right click dec?

fasterit: It's doing that, how about testing before giving feedback? :kissing_heart:

BenBE: Misread the `KEY_RECLICK` as `KEY_RCLICK` … ;-)

fasterit: That mean labeling was Hisham on 78be8201dc86c4d57226d845f04ac496cd64d9b6 in 2015. Got me, too, at first.

ScreenManager.c
```diff
@@ -304,6 +304,8 @@ void ScreenManager_run(ScreenManager* this, Panel** lastFocus, int* lastKey, con
                      }
                   }
                }
+            } else if (mevent.bstate & BUTTON3_RELEASED) {
+               ch = KEY_RIGHTCLICK;
```
BenBE, line 308: Shouldn't this replicate the `KEY_RECLICK` logic from above?

fasterit: No, we only want to be sure to filter out a right click. Right clicks are not used to select _and_ change as left clicks are (in htop, which is UI-concept wise bit wrong anyways but historic).

DisplayOptionsPanel.c
```diff
@@ -67,16 +65,32 @@ static HandlerResult DisplayOptionsPanel_eventHandler(Panel* super, int ch) {
          break;
       case '-':
       case KEY_F(7):
+      case KEY_RIGHTCLICK:
```
Explorer09, line 68: I object using right click as mentioned in [this comment](https://github.com/htop-dev/htop/pull/1731#issuecomment-3014042910). The right button to use should be `KEY_WHEELDOWN`.

DisplayOptionsPanel.c
```diff
@@ -67,16 +65,32 @@ static HandlerResult DisplayOptionsPanel_eventHandler(Panel* super, int ch) {
...
          if (OptionItem_kind(selected) == OPTION_ITEM_NUMBER) {
             NumberItem_decrease((NumberItem*)selected);
             result = HANDLED;
+         } else if (OptionItem_kind(selected) == OPTION_ITEM_CHECK) {
+            CheckItem_set((CheckItem*)selected, false);
+            result = HANDLED;
          }
          break;
       case '+':
       case KEY_F(8):
```
Explorer09, line 78: Referencing [this comment](https://github.com/htop-dev/htop/pull/1731#issuecomment-3016549163), a `case KEY_WHEELUP:` can be added here.

ScreenManager.c
BenBE: Ah, k. LGTM.

BenBE:
@Explorer09 Handling mouse wheel / touchpad scroll events is a different set of changes (even if most of the code for it may overlap.

DisplayOptionsPanel.c
fasterit: No, because then you'd scroll through the list of options with the mouse wheel and when getting to a numeric one you'd get "stuck" and it's value would start changing (rapidly) instead. That's awful UX.

Explorer09: @fasterit I saw GIMP and Inkscape have their way of solving this:
Detect the mouse cursor position when rolling mouse wheel. If the mouse cursor is on a numeric field, increment or decrement the value of that. If the mouse cursor is outside the numeric field but still in the panel, scroll thr panel instead.

So the mouse wheel can still be used for value increment and decrement.

Maybe my proposal of `case KEY_WHEELUP:` here might not make things work as intended, but it might be worth fixing this in another PR.

PR #1735, opened 2025-07-05
Pause mode fixes

eMPee584:
- **Revert "Fix pause mode ("Z") in tree view"**
- **Improve contrast & readability of PAUSED indicator in function bar**


fasterit:
Thank you for trying to improve htop!

7b739b6 was intentional as the data is not updated so the sort may be showing wrong data / wrong child nodes.

Personally I also think the "PAUSED" indicator is sufficiently visible and fits better into the design than your proposal. But I am happy to have others chime in on this. (Design) taste is always subjective.
/DLange

Explorer09:
Can the OP (@eMPee584) clarify what problem this PR is trying to fix? Without a report of an issue, I can't understand why this PR is needed.

PR #1736, opened 2025-07-06
build: 'htop_search_header_dir' function & libunwind detection logic

Explorer09:
Introduce a configure script function that allows searching a header in a list of include directories. And make the libunwind detection logic utilize this new function.

The new function can be useful in other PRs that need detection of headers outside the system include directories. For example #1270.


configure.ac
```diff
@@ -833,29 +833,29 @@ AC_ARG_ENABLE([unwind],
...
-   no)
-      ;;
-   yes)
-      AC_CHECK_LIB([unwind], [backtrace], [], [AC_MSG_ERROR([can not find required library libunwind])])
-      AC_CHECK_HEADERS([libunwind.h], [], [
-         AM_CFLAGS="$AM_CFLAGS -I/usr/include/libunwind"
-         AC_CHECK_HEADERS([libunwind/libunwind.h], [], [AC_MSG_ERROR([can not find required header file libunwind.h])])
-      ])
+
+      AC_CHECK_LIB([unwind], [backtrace], [],
```
Explorer09, line 843: I'm leaving a note here for people to test this later:

Strictly speaking, the libunwind API doesn't contain this backtrace() function. The libunwind code in htop, when enabled, doesn't call backtrace() either. Why are we checking the presence of backtrace() in the library?

While Linux and BSDs would likely use libunwind from https://www.nongnu.org/libunwind/ (David Mosberger et al.), I discover macOS has another implementation of libunwind (shipped as part of macOS LLVM toolchain). I didn't test the latter yet. But it suggests me that checking for backtrace() here looks inaccurate.

PR #1738, opened 2025-07-13
add CR trimming and early return on empty string

highpn:
Trim carriage return (\r) characters along with space, tab, and newline.
Add a check after left-trimming to return an empty string early if the input is now empty.
This avoids unnecessary right-trimming and improves clarity.

### DIFF
--- a/XUtils.c
+++ b/XUtils.c
@@ -135,12 +135,12 @@ char* String_cat(const char* s1, const char* s2) {
 }
 
 char* String_trim(const char* in) {
-   while (in[0] == ' ' || in[0] == '\t' || in[0] == '\n') {
+   while (in[0] == ' ' || in[0] == '\t' || in[0] == '\n' || in[0] == '\r') {
       in++;
    }
 
    size_t len = strlen(in);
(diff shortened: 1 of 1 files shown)


BenBE:
The second while loop already had this early return (using short circuit evaluation of the loop condition).

Thus the only change remaining is adding `'\r'` to the set of white space characters to be trimmed.

marcibarat:
I've now added `'\r'` to the set of whitespace characters to trim, so that should cover the remaining change.

Explorer09:
The question is, why? What bug does this PR fix specifically? The `\r` had some special uses in htop.

highpn:
This change is based on the assumption that leftover \r characters (from files created on Windows or old Mac systems) might cause subtle issues on Unix-like systems. I haven't seen a specific bug, but thought this might make parsing more robust.

Explorer09:
@highpn
https://github.com/htop-dev/htop/blob/eb5fcd3d38578c92059dfaa350e3c1141b68ec50/linux/LinuxProcessTable.c#L1294
Evidence of `'\r'` being used for specil purposes.

Unless there's an actual bug that this PR particularly solves, touching a utility function will likely break something you aren't aware of. That's why I am against this PR.

PR #1740, opened 2025-07-16
Keep track of the biggest PID and scale the column accordingly

### LINKED ISSUE #1645: Large PID causes misalignment of columns
Large PID numbers cause misaligned columns. Htop should either truncate PIDs to say 5 digits, or dynamically adjust the PID column's width.

Example of misalignment when PIDs are longer than 5 (in my case up to 8).

Image

a17sol:
This PR closes #1645. The bug was reported on macOS, but may occur on all the platforms where the maximum PID can be changed at runtime.

This change lets the PID column grow dynamically if a longer PID is encountered, starting from the platform's initial maximum PID length. It will not shrink once expanded.

The maximum PID is now treated similarly to the maximum UID:
    A new field Machine.maxProcessId has been introduced.
    During each scan, the longest PID encountered is stored in this field and passed to Row_setPidColumnWidth.
    The checkRecalculation function has been modified to redraw the header if the PID length has changed.


ProcessTable.c
```diff
@@ -70,10 +70,13 @@ static void ProcessTable_cleanupEntries(Table* super) {
       // tidy up Process state after refreshing the ProcessTable table
       Process_makeCommandStr(p, settings);
 
-      // keep track of the highest UID for column scaling
+      // keep track of the highest UID and PID for column scaling
       if (p->st_uid > host->maxUserId)
          host->maxUserId = p->st_uid;
 
+      if (Process_getPid(p) > host->maxProcessId)
+         host->maxProcessId = Process_getPid(p);
```
Explorer09: Please store the temporary result of `Process_getPid(p)` and avoid calling the function twice.

Machine.h
```diff
@@ -69,6 +69,8 @@ typedef struct Machine_ {
    uid_t maxUserId;  /* recently observed */
    uid_t userId;  /* selected row user ID */
 
+   pid_t maxProcessId; /* encountered */
```
Explorer09: What does "encountered" in this line comment mean?

a17sol: The comment "encountered" is meant to clarify that maxProcessId does not represent the maximum possible PID, but rather the highest PID encountered. As far as I understand, this is similar in meaning to the "recently observed" comment for maxUserId.

BenBE: I think, wording like this is a bit clearer:

```suggestion
   pid_t maxProcessId; /* largest PID seens at runtime */
```

ProcessTable.c
a17sol: Fixed this, thanks

ScreenManager.c
```diff
@@ -147,7 +148,7 @@ static void checkRecalculation(ScreenManager* this, double* oldTime, int* sortTi
       // always update header, especially to avoid gaps in graph meters
       Header_updateData(this->header);
       // force redraw if the number of UID digits was changed
-      if (Process_uidDigits != oldUidDigits) {
+      if (Process_uidDigits != oldUidDigits || Process_pidDigits != oldPidDigits) {
          *force_redraw = true;
       }
```
BenBE: Just make this two blocks.
```suggestion
      if (Process_uidDigits != oldUidDigits) {
         *force_redraw = true;
      }
      if (Process_pidDigits != oldPidDigits) {
         *force_redraw = true;
      }
```

Also, that comment above needs updating.

BenBE:
I did some code style cleanup in one of the routines and squashed the commits into one (cf. style guide).

ScreenManager.c
BenBE: I fixed the comment above.

I reverted this into one conditional to take advantage of the short-circuit logic in the evaluation. Likely doesn't make a huge difference in performance though.

Machine.h
a17sol: Isn't "seens" a typo?

BenBE: Thx. Fixed on main.

PR #1745, opened 2025-07-24
Automatically reconnect PCP metrics contexts on disconnect

natoscott:
If the PCP pmcd metrics server is restarted, pcp-htop can receive SIGPIPE and failed calls to pmFetch which currently often results in process exit.  This commit adds automatic pmcd reconnect after failure, increasing resilience and avoiding manual restarts we'll otherwise need.

### DIFF
--- a/Action.h
+++ b/Action.h
@@ -37,6 +37,7 @@ typedef struct State_ {
    struct MainPanel_* mainPanel;
    Header* header;
    bool pauseUpdate;
+   bool validUpdate;
    bool hideSelection;
    bool hideMeters;
 } State;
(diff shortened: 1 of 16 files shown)


BenBE:
You might want to delay subsequent reconnect attempts to avoid hammering the `pmcd`.

A nice to have would be a small message in the status bar, when the lost connection is noticed.

natoscott:
Thanks @BenBE - agreed on both points.

We get a natural delay between reconnection attempts due to the sampling interval (we only fetch once per interval), so I think that aspect is OK.

It'd be great to show disconnect.  The place where we observe the connection failure is distant from any screen update activity, so I think it'd have to be at some other point later on.  It needs to be a platform-specific place too of course - leave it with me, I'll see if I can find somewhere appropriate.

BenBE (approved):
LGTM.

PR #1747, opened 2025-07-27
add --no-meters option to hide meters directly from CLI

### LINKED ISSUE #1744: Add CLI parameter to hide meters
When running htop on systems with a high number of CPU cores (96+), the CPU meters occupy a significant portion of the screen. This leaves limited space for viewing the actual process information, which is often the primary reason for me using htop. Currently, I find myself immediately pressing `Shift+3` after launching htop to hide these meters and gain more usable space for process table.

It will be nice to have a command-line option to start htop with the CPU meters already hidden, similar to other existing visibility options. A new flag like `--no-meters` would be consistent with other existing htop options such as `--no-color` and `--no-unicode`.

Would this be something the maintainers would consider implementing? I'd be happy to provide any additional information if needed or submit a PR.

serg00ns:
Closes #1744

### DIFF
--- a/CommandLine.c
+++ b/CommandLine.c
@@ -59,7 +59,8 @@ static void printHelpFlag(const char* name) {
 #ifdef HAVE_GETMOUSE
    printf("-M --no-mouse                   Disable the mouse\n");
 #endif
-   printf("-n --max-iterations=NUMBER      Exit htop after NUMBER iterations/frame updates\n"
+   printf("-m --no-meters                  Hide meters\n"
+          "-n --max-iterations=NUMBER      Exit htop after NUMBER iterations/frame updates\n"
           "-p --pid=PID[,PID,PID...]       Show only the given PIDs\n"
           "   --readonly                   Disable all system and process changing features\n"
           "-s --sort-key=COLUMN            Sort by COLUMN in list view (try --sort-key=help for a list)\n"
@@ -91,6 +92,7 @@ typedef struct CommandLineSettings_ {
    bool highlightChanges;
    int highlightDelaySecs;
    bool readonly;
+   bool hideMeters;
 } CommandLineSettings;
 
 static CommandLineStatus parseArguments(int argc, char** argv, CommandLineSettings* flags) {
@@ -125,6 +127,7 @@ static CommandLineStatus parseArguments(int argc, char** argv, CommandLineSettin
       {"no-colour",  no_argument,         0, 'C'},
       {"no-mouse",   no_argument,         0, 'M'},
       {"no-unicode", no_argument,         0, 'U'},
+      {"no-meters", no_argument,          0, 'm'},
       {"tree",       no_argument,         0, 't'},
       {"pid",        required_argument,   0, 'p'},
       {"filter",     required_argument,   0, 'F'},
@@ -136,7 +139,7 @@ static CommandLineStatus parseArguments(int argc, char** argv, CommandLineSettin
 
    int opt, opti = 0;
    /* Parse arguments */
-   while ((opt = getopt_long(argc, argv, "hVMCs:td:n:u::Up:F:H::", long_opts, &opti))) {
+   while ((opt = getopt_long(argc, argv, "hVMmCs:td:n:u::Up:F:H::", long_opts, &opti))) {
       if (opt == EOF)
          break;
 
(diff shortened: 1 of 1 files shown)


BenBE:
Given the "disable X" options all have uppercase short options so far, I'd suggest `-H`/`--no-header`. Else it might be a bit too close to `-M`, which disables the mouse support.

Explorer09:
Keep in mind that the `hideMeters` option in the `State` object also hides the tab names.

serg00ns:
> Given the "disable X" options all have uppercase short options so far, I'd suggest `-H`/`--no-header`. Else it might be a bit too close to `-M`, which disables the mouse support.

Noticed that `-H` is already used, so maybe `-B` / `--no-bars` could work instead? 
Happy to update if that’s okay.

BenBE:
Given some internal feedback (and related to the vacation season) there'll be some delay for the final decision of what the option will become. Until then, suggesting and discussing the possible option names is one task to prepare.

Another one (based on internal comments) is a possible issue with existing configuration options, which might be accidentally modified when the configuration is stored while a session with this config option is running.

natoscott:
IMO, either --no-meters or --no-header are the best choices (no strong preference here) as --no-bars is ambiguous with the FunctionBar, and it doesn't really capture the Meters anyway, which have many different non-Bar visuals too.

If we do this, I wonder if the 'hideFunctionBar' functionality should be exposed via a command line option as well BTW?

And the point @Explorer09 makes re the Tabs sounds like a (pre-existing) bug, no?

Finally, the man page will also need an update with this change.

Explorer09:
@natoscott 

> And the point @Explorer09 makes re the Tabs sounds like a (pre-existing) bug, no?

Perhaps it is because too few people tested the `#` hotkey in htop. I only noticed that when testing the new command line option.

Explorer09:
It's better to rebase the branch onto `main` rather than cross-merge it (to keep the commit history clean). Also squash the two commits into one.

fasterit:
I like @natoscott's comments, so I suggest to implement both `--no-meters` and `--no-functionbar`. I don't think these need short options at all as they are very "special interest". These need to be reflected in the man page as well.

BTW `{ echo -n "#" ; cat ; } | htop` works.

CommandLine.c
```diff
@@ -370,6 +376,8 @@ int CommandLine_run(int argc, char** argv) {
       }
       ScreenSettings_setSortKey(settings->ss, flags.sortKey);
    }
+   if (flags.hideFunctionBar)
+      settings->hideFunctionBar = 2;
```
BenBE, line 383: Why this tri-state logic?

CommandLine.c
```diff
@@ -54,6 +54,7 @@ static void printHelpFlag(const char* name) {
           "-C --no-color                   Use a monochrome color scheme\n"
           "-d --delay=DELAY                Set the delay between updates, in tenths of seconds\n"
           "-F --filter=FILTER              Show only the commands matching the given filter\n"
+          "   --no-functionbar             Hide the function bar\n"
```
Explorer09: It could be just me, but I think the name `--no-function-bar` would be better. (Separate the two words.)

CommandLine.c
serg00ns: https://github.com/htop-dev/htop/blob/39a7eac3f74fdf349593010c5fc1250644801ebd/Settings.h#L108
The setting is already defined as tri-state. So I thought that it would be better to pick the permanent state which is 2

BenBE: I got somewhat confused by the `bool` dfinition in `CommandLine.c` for the `CommandLineSettings` type.

CommandLine.c
BenBE: We might even accept both, but keep the documentation to only mention the one with the additional dash.

CommandLine.c
BenBE, line 120: This initializer probably also should include the defaults (`false`) for the two new flags.

serg00ns: Yes. I had missed initialize. Adding now.

BenBE: Technically not bad, as the default would be `false` anyways, but its much cleaner to be explicit here. :)

BenBE:
Could you squash the last three commits into one? No need to document the journey there. ;-)

Otherwise seems good.

CommandLine.c
serg00ns: :))

jamel:
Thank you!

PR #1748, opened 2025-07-28
Display a platform-specific failed state message

natoscott:
Further discussion around earlier PR #1745 (relating to improved handling of PCP pmcd connection loss) centered on more descriptive function names and diagnostics.  In this commit the boolean logic in Platform_getValidState has been reversed to *FailedState, along with a pointer to a less generic error message allowing for more clear failure diagnostics.

### DIFF
--- a/Action.h
+++ b/Action.h
@@ -36,8 +36,8 @@ typedef struct State_ {
    Machine* host;
    struct MainPanel_* mainPanel;
    Header* header;
+   const char* failedUpdate; /* function bar diagnostic or NULL if no error */
    bool pauseUpdate;
-   bool validUpdate;
    bool hideSelection;
    bool hideMeters;
 } State;
--- a/ScreenManager.c
+++ b/ScreenManager.c
@@ -147,7 +147,7 @@ static void checkRecalculation(ScreenManager* this, double* oldTime, int* sortTi
       Machine_scan(host);
       if (!this->state->pauseUpdate)
          Machine_scanTables(host);
-      this->state->validUpdate = Platform_getValidState();
+      this->state->failedUpdate = Platform_getFailedState();
 
       // always update header, especially to avoid gaps in graph meters
       Header_updateData(this->header);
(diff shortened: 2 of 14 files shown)


BenBE:
Boolean+string kinda asks for being written as `NULL` on `false` and the error message on `true`.

natoscott:
Sure, can do - I'll wait for any contradictory review suggestions first this time though :-P

Action.h
```diff
@@ -36,8 +36,8 @@ typedef struct State_ {
    Machine* host;
    struct MainPanel_* mainPanel;
    Header* header;
+   const char* failedUpdate; /* function bar diagnostic or NULL if no error */
```
BenBE, line 39: Rename this to `collectionStateDiagnostic`, which somewhat better aligns with its purpose. Also better hints at the possible use case of not just representing an error state …

ScreenManager.c
```diff
@@ -147,7 +147,7 @@ static void checkRecalculation(ScreenManager* this, double* oldTime, int* sortTi
       Machine_scan(host);
       if (!this->state->pauseUpdate)
          Machine_scanTables(host);
-      this->state->validUpdate = Platform_getValidState();
+      Platform_getFailedState(&this->state->failedUpdate);
```
BenBE: Is there a particular reason to use out parameters instead of a function return value?

Action.h
natoscott: Yow, that's quite a mouthful! :)  I think I prefer the simpler, shorter name TBH, and the way it aligns with pausedUpdate on the line below.  The intention is to represent errors, not arbitrary messages (recall we're using FAILED_READ as the color here too), so I prefer if we leave this part as-is.

ScreenManager.c
natoscott: No good reason, and yep, I agree it would be nicer to use the return code instead.

Action.h
BenBE: Good point.

BenBE (approved):
LGTM. Feel free to merge, if nobody else complains … ;-)

PR #1756, opened 2025-08-13
xAsprintf fixes

BenBE:
Some minor consistency changes for calls to `xAsprintf`.

### DIFF
--- a/XUtils.c
+++ b/XUtils.c
@@ -236,6 +236,8 @@ size_t strnlen(const char* str, size_t maxLen) {
 #endif
 
 int xAsprintf(char** strp, const char* fmt, ...) {
+   *strp = NULL;
+
    va_list vl;
    va_start(vl, fmt);
    int r = vasprintf(strp, fmt, vl);


XUtils.c
```diff
@@ -236,6 +236,8 @@ size_t strnlen(const char* str, size_t maxLen) {
 #endif
 
 int xAsprintf(char** strp, const char* fmt, ...) {
+   *strp = NULL;
```
Explorer09, line 239: I think there is a redundancy here. Since in one previous commit you are trying to ensure the caller has set `*strp` to NULL, and then here you set `*strp` to NULL once again. While this might seem harmless, I wonder if this can hide a potential memory leak. Like this hypothetical scenario:

```c
char* buf = xMalloc(len);
// ... without freeing buf ...
xAsprintf(&buf, "foo %d %d", x, y);
```

I would prefer that `xAsprintf` asserts that caller sets `*strp` to NULL, and not set it by itself.

BenBE: I was thinking of putting an assert in there, something like:

```
ASSERT(!*strp);
```

But dong that would change semantics for the caller AND change the behaviour on FreeBSD, which doesn't set `strp` on error (yes, we fail in that case, but nonetheless it's a different behaviour).

And regarding memory leaks: `valgrind` is your friend.

Explorer09: For this particular `xAsprintf` case, we fail on error. Whether or not `strp` would be set is irrelevant.

If we want a more predictable behavior then `if (r < 0) *strp = NULL;` after the `vasprintf` would make more sense. For me, the issue is whether `asprintf` family of functions would _read_ `*strp` when we feed the pointer to it. If it totally ignores the value of `*strp`, then we would be doing a zero-initialization / assignment that's redundant just to "feel safe".

PR #1763, opened 2025-08-31
Split DiskIOMeter into disk IO time and rate sub-meters

Explorer09:
Split the `DiskIOMeter` into two, `DiskIOTimeMeter` and `DiskIORateMeter`, showing the time and the rate separately.

The name `DiskIOMeter` is retained but now shows the combined view of time and rate, similar to `MemorySwapMeter`.


DiskIOMeter.c
```diff
@@ -138,8 +154,46 @@ static void DiskIOMeter_display(ATTR_UNUSED const Object* cast, RichString* out)
...
+      return;
+   }
+   if (status == RATESTATUS_INIT) {
+      xSnprintf(this->txtBuffer, sizeof(this->txtBuffer), "init");
+      return;
+   }
+   if (status == RATESTATUS_STALE) {
+      xSnprintf(this->txtBuffer, sizeof(this->txtBuffer), "stale");
+      return;
+   }
```
BenBE: Any reason for not using a `switch`/`case` like in the function below?

DiskIOMeter.c
```diff
@@ -148,20 +202,117 @@ static void DiskIOMeter_display(ATTR_UNUSED const Object* cast, RichString* out)
...
    .name = "DiskIO",
    .uiName = "Disk IO",
-   .caption = "Disk IO: "
+   .description = "Disk IO time & rate combined display",
+   .caption = "Dsk: ",
+   .draw = DiskIOMeter_draw,
+   .init = DiskIOMeter_init,
+   .updateMode = DiskIOMeter_updateMode,
+   .done = DiskIOMeter_done
 };
```
BenBE, line 361: Having this moved after the rate-based meter from where it was before introduces quite a bit of noise in this commit. Try to keep the commits minimal without unnecessarily re-ordering code. I had to resort to `git lol -p --color-moved --color-moved-ws=allow-indentation-change main...` to get at least partially sensible diffs; and those were still PITA to review. ;-)

Similar goes for the `MeterClass` constants, where putting things in the order `DiskIOMeter`, `DiskIORateMeter`, `DiskIOTimeMeter` would have produced a much shorter/cleaner diff.

BenBE:
What's currently still open for this PR to become ready?

DiskIOMeter.c
Explorer09: I thought `DiskIOMeter` goes after `DiskIORateMeter` and `DiskIOTimeMeter` because the former depends on the functionality of the latter. If that makes the diff messy, I can split the reordering into a separate commit. Will that be acceptable?

DiskIOMeter.c
Explorer09: It's to prevent too much style changes to the original code. Not a big deal to switch to `switch`-`case` though. (Pun not intended)

DiskIOMeter.c
BenBE: I think just having the meter classes in the same order s previously with the new meters added below is fine. Moving code always teds to get messy no matter if you split it or not. So if you can avoid moving code (and here keeping things ordered "Ye Old Way"™ is fine), just keep it that way.

NB: The first 3 commits are fine. Only the last commit may benefit from avoiding to move code.

Explorer09:
> What's currently still open for this PR to become ready?

I originally made this PR as a "proof of concept" and to grab early comments about the meters' look and functions.

* <s>I also intended that `DiskIOTimeMeter` will include the number of disks in the display, so that the greater-than-100% disk time can be understood without reading more context or manual, however I don't have an idea on how to implement that well. I think I can postpone that part into a separate PR. I just leave a reminder here.</s> (Update: This one is done in the new revision.)

BenBE:
Just did a quick check with the design and I think the combined meter should show IO rates left, IO Time used on the right. Else you get quite a big gap with just the percentage shown in text mode.

DiskIOMeter.c
Explorer09: When I try to rework the last commit (just now), I realised the problem wasn't on the code reordering, but that you failed to recognize the old `DiskIOMeter_` functions was renamed to `DiskIOTimeMeter_` functions intentionally.

The "DiskIOMeter" name was kept for compatibility, but the new "DiskIOMeter" in this PR is, actually, entirely new. Users with older htop config files that use the "DiskIOMeter" will new see two bars with the new code.

I'm not sure what's the best way to fix the confusion though. Maybe I could introduce both `DiskIORate` and `DiskIOTime` first (`DiskIOTimeMeter` be mostly duplicate of the old `DiskIOMeter` code). Then rework `DiskIOMeter` to the now-seen, final form.

Explorer09:
> Just did a quick check with the design and I think the combined meter should show IO rates left, IO Time used on the right. Else you get quite a big gap with just the percentage shown in text mode.

Makes sense. Perhaps another good reason is that people might look at the I/O rate graph more often than the time/utilisation percentage. (Besides, the I/O rate graph can support two channels while the percentage graph cannot.)

DiskIOMeter.c
BenBE: This updated patchset looks much better. And even though the last commit is a bit more dense in changes, that's quite easy to review.

DiskIOMeter.c
```diff
@@ -21,104 +21,168 @@ in the source distribution for its full text.
...
+   Meter* diskIOTimeMeter;
+} DiskIOMeterData;
+
+static const int DiskIORateMeter_attributes[] = {
    METER_VALUE_IOREAD,
    METER_VALUE_IOWRITE,
 };
 
+static const int DiskIOTimeMeter_attributes[] = {
+   METER_VALUE_NOTICE
```
BenBE:
```suggestion
   METER_VALUE_NOTICE,
```

DiskIOMeter.c
```diff
@@ -138,30 +202,121 @@ static void DiskIOMeter_display(ATTR_UNUSED const Object* cast, RichString* out)
...
+   DiskIOMeterData* data = this->meterData;
+
+   /* Use the same width for each sub meter to align with CPU meter */
+   const int colwidth = w / 2;
+   const int diff = w % 2;
+
+   assert(data->diskIORateMeter->draw);
+   data->diskIORateMeter->draw(data->diskIORateMeter, x, y, colwidth);
+   assert(data->diskIOTimeMeter->draw);
+   data->diskIOTimeMeter->draw(data->diskIOTimeMeter, x + colwidth + diff, y, colwidth);
```
BenBE: While initializing or stale, this looks kinda strange (in text mode).

The IO Time meter itself could benefit from actually labeling the value it shows in text mode.

DiskIOMeter.c
```diff
@@ -138,30 +202,121 @@ static void DiskIOMeter_display(ATTR_UNUSED const Object* cast, RichString* out)
...
+}
+
+static void DiskIOMeter_init(Meter* this) {
+   DiskIOMeterData* data = this->meterData;
+
+   if (!data) {
+      data = this->meterData = xCalloc(1, sizeof(DiskIOMeterData));
+   }
```
BenBE: Maybe avoid this double assignment?
```suggestion
   if (!this->meterData) {
      this->meterData = xCalloc(1, sizeof(DiskIOMeterData));
   }

   DiskIOMeterData* data = this->meterData;
```

DiskIOMeter.c
BenBE: There's `Meter_drawFn` that could be used here.

Explorer09: What do you mean when you say it looks strange?

BenBE: Basically you have like:

```
Dsk: Stale ...                               Dsk: Stale ...
```

where you somewhat break the illusion of this being one meter …

I think:

```
Dsk: Stale ...
```

should suffice.

Similar, in the text mode, having values from both meters right next to each other

```
Dsk: read: 42.23 KiB/s, write: 666.69 MiB/s, usage: 1337%
```

compare to current:

```
Dsk: read: 42.23 KiB/s, write: 666.69 MiB/s                       Dsk: 1337%
```

would be somewhat more intuitive.

Explorer09: That means merging the two strings in the Text meter display. It's a bit out of scope of my PR, but I know it's a good feature. I'm not sure how to make it, though.

Explorer09: `Meter_drawFn` refers to the (custom) draw function associated with the meter class, not the meter mode. So no, they are not the same.

Explorer09:
@BenBE On the topic of avoiding double assignment in code style, I found there are three other files that had this "double assignment" syntax when initializing data. Not sure if you want these to be changed as well, but I'm reluctant to file a new PR for this minor style change: 

```diff
--- a/CPUMeter.c
+++ b/CPUMeter.c
@@ -241,9 +241,10 @@ static void CPUMeterCommonInit(Meter* this) {
 
    CPUMeterData* data = this->meterData;
    if (!data) {
-      data = this->meterData = xMalloc(sizeof(CPUMeterData));
+      data = xMalloc(sizeof(CPUMeterData));
       data->cpus = this->host->existingCPUs;
       data->meters = count ? xCalloc(count, sizeof(Meter*)) : NULL;
+      this->meterData = data;
    }
 
    Meter** meters = data->meters;
--- a/Machine.c
+++ b/Machine.c
@@ -79,9 +79,11 @@ void Machine_populateTablesFromSettings(Machine* this, Settings* settings, Table
 
    for (size_t i = 0; i < settings->nScreens; i++) {
       ScreenSettings* ss = settings->screens[i];
+
+      if (!ss->table)
+         ss->table = processTable;
+
       Table* table = ss->table;
-      if (!table)
-         table = ss->table = processTable;
       if (i == 0)
          this->activeTable = table;
 
--- a/MemorySwapMeter.c
+++ b/MemorySwapMeter.c
@@ -47,11 +47,10 @@ static void MemorySwapMeter_draw(Meter* this, int x, int y, int w) {
 }
 
 static void MemorySwapMeter_init(Meter* this) {
-   MemorySwapMeterData* data = this->meterData;
+   if (!this->meterData)
+      this->meterData = xCalloc(1, sizeof(MemorySwapMeterData));
 
-   if (!data) {
-      data = this->meterData = xCalloc(1, sizeof(MemorySwapMeterData));
-   }
+   MemorySwapMeterData* data = this->meterData;
 
    if (!data->memoryMeter)
       data->memoryMeter = Meter_new(this->host, 0, (const MeterClass*) Class(MemoryMeter));
```

BenBE:
@Explorer09 Patch applied on main with slight modifications.

DiskIOMeter.c
Explorer09: Update: I made it with a trick.

DiskIOMeter.c
```diff
@@ -136,32 +208,156 @@ static void DiskIOMeter_display(ATTR_UNUSED const Object* cast, RichString* out)
    char buffer[16];
 
    int color = cached_utilisation_diff > 40.0 ? METER_VALUE_NOTICE : METER_VALUE;
-   int len = xSnprintf(buffer, sizeof(buffer), "%.1f%%", cached_utilisation_diff);
+   int len = xSnprintf(buffer, sizeof(buffer), "%4.1f%%", cached_utilisation_diff);
```
BenBE: This looks inconsistent with signle-digit percentages having an additional blank. Just use minimal spacing.
```suggestion
   int len = xSnprintf(buffer, sizeof(buffer), "%.1f%%", cached_utilisation_diff);
```

DiskIOMeter.c
```diff
@@ -136,32 +208,156 @@ static void DiskIOMeter_display(ATTR_UNUSED const Object* cast, RichString* out)
    char buffer[16];
 
    int color = cached_utilisation_diff > 40.0 ? METER_VALUE_NOTICE : METER_VALUE;
```
BenBE, line 208: Where does the 40% come from?

Explorer09: e5184599814a3210497035e9942f154945f2b02f I would let @cgzones answer this.

PR #1777, opened 2025-10-03
refactor: remove redundant isRunningInContainer initialization

Pablito2020:
Remove the explicit initialization of proc->isRunningInContainer to TRI_OFF before calling LinuxProcessTable_readStatusFile.

[The function already initializes this field as part of its normal operation](https://github.com/htop-dev/htop/blob/main/linux/LinuxProcessTable.c#L532) when reading /proc/[pid]/status.


Explorer09:
I suggest you leave that code alone. `LinuxProcessTable_readStatusFile` is only called in one place, and when the compiler inlines the function, it can figure that out automatically.

Pablito2020:
You're right, after analyzing the compiled code, the `mov` instruction related with the initialization doesn't actually appear. 

Thanks for your time and feedback :)

PR #1779, opened 2025-10-05
Improve libunwind check in configure

### LINKED ISSUE #1778: libunwind check is wrong on musl (and maybe others)?
The ``--enable-libunwind`` check checks for the ``backtrace`` function which is normally part of either glibc or libexecinfo. If neither are installed this will fail even though libunwind exists. 

Changing ``configure.ac`` to use ``unw_getcontext`` or some other similar function makes it link and work properly. I'm not familiar with autoconf so I don't know if this is the right way. The actual code that uses ``backtrace`` is #ifdef'd to an alternative code path, so it seems like this check is broken on non-musl, ``--disable-unwind`` codepaths too?

Explorer09:
* Fix #1778 
* Add a new check in configure to ensure that `unw_init_local()` in libunwind works.


lacusdoloris:
I verified that htop now builds with ``--enable-unwind`` on both a cross-compile musl+LLVM and a host musl+LLVM. Compiling with frame pointers also produces a working "backtrace" (the signal handler) on a ``kill -SEGV``. This definitely fixes it!

BenBE:
Some hints for formatting:

```diff
diff --git a/configure.ac b/configure.ac
index 2cc39f4d..88dde700 100644
--- a/configure.ac
+++ b/configure.ac
@@ -860,9 +860,11 @@ elif htop_search_header_dir libunwind.h "/usr/include/libunwind"; then
    CPPFLAGS="$AM_CPPFLAGS $CPPFLAGS"
 
    AC_MSG_CHECKING([whether local unwinding works])
-   AC_LINK_IFELSE([AC_LANG_PROGRAM([[
+   AC_LINK_IFELSE(
+      [AC_LANG_PROGRAM(
+         [[
 #include <libunwind.h>
-      ]], [[
+         ]], [[
 /* If libunwind is built with remote unwinding only, unw_init_local()
    will be a dummy function that returns -UNW_EINVAL when called. */
 unw_context_t context;
@@ -872,33 +874,43 @@ if (!ret) {
    ret = unw_init_local(&cursor, &context);
 }
 return (ret <= 0 && ret > -126 ? -ret : 1);
-      ]])],
-      [libunwind_local_support=unknown
-      if test "$cross_compiling" = yes; then
-         AC_COMPILE_IFELSE([AC_LANG_SOURCE([[
+         ]]
+      )],
+      [
+         libunwind_local_support=unknown
+         if test "$cross_compiling" = yes; then
+            AC_COMPILE_IFELSE(
+               [
+                  AC_LANG_SOURCE([[
 #include <libunwind.h>
 #ifdef UNW_REMOTE_ONLY
 #error "local unwinding not supported"
 #endif
-            ]])],
-            [AC_MSG_RESULT([assume yes (cross compiling)])],
-            [libunwind_local_support=no])
-      elif ./conftest$EXEEXT >&AS_MESSAGE_LOG_FD; then
-         libunwind_local_support=yes
-         AC_MSG_RESULT([yes])
-      else
-         libunwind_local_support=no
-      fi
-      if test "$libunwind_local_support" = no; then
-         AC_MSG_RESULT([no])
-         if test "$enable_unwind" = yes; then
-            AC_MSG_WARN([local unwinding test failed; this build of libunwind might not support local unwinding.])
+                  ]])
+               ],
+               [AC_MSG_RESULT([assume yes (cross compiling)])],
+               [libunwind_local_support=no]
+            )
+         elif ./conftest$EXEEXT >&AS_MESSAGE_LOG_FD; then
+            libunwind_local_support=yes
+            AC_MSG_RESULT([yes])
          else
-            enable_unwind=no
+            libunwind_local_support=no
          fi
-      fi],
-      [AC_MSG_RESULT([no])
-      AC_MSG_FAILURE([there are problems with libunwind.h])])
+         if test "$libunwind_local_support" = no; then
+            AC_MSG_RESULT([no])
+            if test "$enable_unwind" = yes; then
+               AC_MSG_WARN([local unwinding test failed; this build of libunwind might not support local unwinding.])
+            else
+               enable_unwind=no
+            fi
+         fi
+      ],
+      [
+         AC_MSG_RESULT([no])
+         AC_MSG_FAILURE([there are problems with libunwind.h])
+      ]
+   )
 
    CPPFLAGS=$htop_save_CPPFLAGS
 elif test "$enable_unwind" = yes; then
@@ -915,7 +927,8 @@ else
    AC_SEARCH_LIBS([backtrace], [execinfo])
 
    AC_MSG_CHECKING([the return type of backtrace])
-   AC_COMPILE_IFELSE([AC_LANG_SOURCE([[
+   AC_COMPILE_IFELSE(
+      [AC_LANG_SOURCE([[
 #ifndef BACKTRACE_RETURN_TYPE
 #error "BACKTRACE_RETURN_TYPE not defined"
 #endif
@@ -939,10 +952,12 @@ else
          size_t \
          int \
       ; do
-         AC_LINK_IFELSE([AC_LANG_PROGRAM([[
+         AC_LINK_IFELSE(
+            [AC_LANG_PROGRAM(
+               [[
 #include <execinfo.h>
 #include <stddef.h> /* For size_t */
-         ]],[[
+               ]],[[
 #define BACKTRACE_RETURN_TYPE $backtrace_return_type
 /* Assuming the second argument of backtrace_symbols() would use the same type
    as the return type of backtrace(). */
@@ -950,10 +965,14 @@ static char** (*func)(void* const*, BACKTRACE_RETURN_TYPE) = backtrace_symbols;
 static void* addr;
 BACKTRACE_RETURN_TYPE nptrs = backtrace(&addr, 1);
 (void)nptrs;
-         ]])],
-         [AC_DEFINE_UNQUOTED([BACKTRACE_RETURN_TYPE], [$backtrace_return_type], [Defined to the return type of the 'backtrace' function.])
-         break],
-         [backtrace_return_type=""])
+               ]]
+            )],
+            [
+               AC_DEFINE_UNQUOTED([BACKTRACE_RETURN_TYPE], [$backtrace_return_type], [Defined to the return type of the 'backtrace' function.])
+               break
+            ],
+            [backtrace_return_type=""]
+         )
       done
       backtrace_return_type_msg=`echo "x$backtrace_return_type" | sed '
          s/^x *//
@@ -961,7 +980,9 @@ BACKTRACE_RETURN_TYPE nptrs = backtrace(&addr, 1);
          s/^\(_*typeof_*\) *( *backtrace( *( *void *\* *\* *) *1 *, *0 *) *) *$/automatic (using \1)/'`
       AC_MSG_RESULT([$backtrace_return_type_msg])
 
-      CFLAGS=$htop_save_CFLAGS])
+      CFLAGS=$htop_save_CFLAGS
+      ]
+   )
 fi
 
 
```

The overall change LGTM AFAICT, but only had a quick look yesterday night so far.

Explorer09:
I slightly tweaked the indentation and quoting style. Sigh. M4 quoting is hard, and there is not yet a consistent style in htop's configure script. Let's see if this style can work better in the long term, and I can try my best to change the rest of the configure code to this style.

BenBE:
The closing brackets should align with the indentation of the line where they are opened, instead of aligning the closing brackets at the start of the line.

Using indentation for the code blocks is fine (I thought about it, but didn't include it in my patch suggestion above); just be sure, that inside brackets, if you break those lines, another level of indentation should be used.

```m4
AC_FOO([
   AC_BAR([
      Do stuff indented here
   ])
])
```

Alternative:

```m4
AC_FOO(
   [
      AC_BAR([
         Do stuff indented here
      ])
   ], [
      AC_BAZ([
         Do stuff indented here
      ])
   ]
)
```

Also acceptable:

```m4
AC_FOO(
   [AC_BAR([
      Do stuff indented here
   ])],
   [AC_BAZ([
      Do stuff indented here
   ])]
)
```

Basic rule in all cases roughly: The opening brackets should be matched in the line closing them. So if you open `[([` the closing line should be aligned and close `])]` indented the same as the line with the opening bracket. No line in between (except for maybe C code blocks) should be less indented.

Explorer09:
Let me try to explain what my proposed indent style is about.

Example here:

```m4
AC_LINK_IFELSE(
   [AC_LANG_PROGRAM(
      [[
/* Checks if the "bool" definition in curses is ISO C compatible */
#include <stdbool.h>
#include "ProvideCurses.h"
]],
      [[
      keypad(stdscr, false);
      refresh();
#if defined(HAVE_LIBNCURSESW)
      {
         static cchar_t dummy;
         mvadd_wchnstr(0, 0, &dummy, 0);
      }
#endif
]]
   )],
   [AC_MSG_RESULT(yes)],
   [
   AC_MSG_RESULT(no)
   AC_MSG_FAILURE([there are problems with the curses header])
]
)
```

What I was trying to address are the peculiarities of M4 quoting. In M4 macro arguments, unquoted whitespace is stripped. Quoted whitespace in arguments would be preserved unless there are other macro expansions (such as `dnl`) that strip it.

Therefore, `MACRO([  text  ])` would get the leading and trailing spaces around "text" to be preserved.

When it comes to quoting code…

```text
      [[
      keypad(stdscr, false);
      refresh();
#if defined(HAVE_LIBNCURSESW)
      {
         static cchar_t dummy;
         mvadd_wchnstr(0, 0, &dummy, 0);
      }
#endif
]]
```

I wish to keep the indents of the code intact when expanded, so when users examine `conftest.c` when debugging `configure`, they can see the uniform indenting on the code snippet, even for the first line and last line of it.

Hence this style: (a) newline immediately after the M4 quotation mark (`[`), (b) no whitespace or indent before the closing quote (`]`). Think of this like a here-document in a shell script, where the ending token has to be on the beginning of the line - no indent allowed.

As an exception, quoted one-liner (e.g. `[AC_MSG_RESULT(yes)]`) can have opening and closing quotes on the same line as the one-line code. No space between `[` and the first non-space character, and no space between the last non-space character and the `]`.

I don't think I can change the maintainers' mind immediately, but at least I believe this coding style matter can be discussed further before a final desision.

BenBE:
I think the most pressing disagreement we have regarding the quoting styles is regarding the final line of multi-line quotes. The reason I prefer to have the code in `configure.ac` aligned is that that's the file most people will interact with. Thus the code in `configure.ac` should be the one optimized for readability. This doesn't conflict with the code indentation alignment for the generated C code in any way, as I wrote that C code may be less indented than the M4 code.

Remains the issue with the closing brackets. I know this causes those white space characters to become part of the generated C code, but as that's hardly what people ever look at, this is more a stylistic choice. On the contrary ending the quote directly at the start of the line loses the information about which indentation level the starting quotes belonged to, thus making backwards-scanning in the M4 file unnecessarily harder IMHO.

Or to stick with your example: I prefer
```m4
      [[
      keypad(stdscr, false);
      refresh();
#if defined(HAVE_LIBNCURSESW)
      {
         static cchar_t dummy;
         mvadd_wchnstr(0, 0, &dummy, 0);
      }
#endif
      ]]
```

over

```m4
      [[
      keypad(stdscr, false);
      refresh();
#if defined(HAVE_LIBNCURSESW)
      {
         static cchar_t dummy;
         mvadd_wchnstr(0, 0, &dummy, 0);
      }
#endif
]]
```

for exactly the reasons stated above … ;-)

BenBE:
Updated suggestion for indentation:

```diff
diff --git a/configure.ac b/configure.ac
index ce73b75c..a4bedf67 100644
--- a/configure.ac
+++ b/configure.ac
@@ -211,38 +211,42 @@ AC_TYPE_UINT64_T
 AC_MSG_CHECKING(for alloc_size)
 old_CFLAGS="$CFLAGS"
 CFLAGS="$CFLAGS -Wno-error -Werror=attributes"
-AC_COMPILE_IFELSE([
-   AC_LANG_SOURCE(
+AC_COMPILE_IFELSE(
+   [AC_LANG_SOURCE(
       [
          /* Attribute supported in GCC 4.3 or later */
          __attribute__((alloc_size(1))) char* my_alloc(int size) { return 0; }
-      ],[]
+      ],
+      []
    )],
    AC_DEFINE([HAVE_ATTR_ALLOC_SIZE], 1, [The alloc_size attribute is supported.])
    AC_MSG_RESULT(yes),
-   AC_MSG_RESULT(no))
+   AC_MSG_RESULT(no)
+)
 CFLAGS="$old_CFLAGS"
 
 AC_MSG_CHECKING(for access)
 old_CFLAGS="$CFLAGS"
 CFLAGS="$CFLAGS -Wno-error -Werror=attributes"
-AC_COMPILE_IFELSE([
-   AC_LANG_SOURCE(
+AC_COMPILE_IFELSE(
+   [AC_LANG_SOURCE(
       [
          /* Attribute supported in GCC 10 or later */
          __attribute__((access(read_only, 1, 2))) extern int foo(const char* str, unsigned len);
-      ],[]
+      ],
+      []
    )],
    AC_DEFINE([HAVE_ATTR_ACCESS], 1, [The access attribute is supported.])
    AC_MSG_RESULT(yes),
-   AC_MSG_RESULT(no))
+   AC_MSG_RESULT(no)
+)
 CFLAGS="$old_CFLAGS"
 
 AC_MSG_CHECKING(for nonnull)
 old_CFLAGS="$CFLAGS"
 CFLAGS="$CFLAGS -Wno-error -Werror=attributes"
-AC_COMPILE_IFELSE([
-   AC_LANG_SOURCE(
+AC_COMPILE_IFELSE(
+   [AC_LANG_SOURCE(
       [[
          /* Attribute supported in GCC 3.3 or later */
          __attribute__((nonnull)) int my_strcmp(const char* a, const char* b);
@@ -251,14 +255,15 @@ AC_COMPILE_IFELSE([
    )],
    AC_DEFINE([HAVE_ATTR_NONNULL], 1, [The nonnull attribute is supported.])
    AC_MSG_RESULT(yes),
-   AC_MSG_RESULT(no))
+   AC_MSG_RESULT(no)
+)
 CFLAGS="$old_CFLAGS"
 
 AC_MSG_CHECKING(for returns_nonnull)
 old_CFLAGS="$CFLAGS"
 CFLAGS="$CFLAGS -Wno-error -Werror=attributes"
-AC_COMPILE_IFELSE([
-   AC_LANG_SOURCE(
+AC_COMPILE_IFELSE(
+   [AC_LANG_SOURCE(
       [[
          /* Attribute supported in GCC 4.9 or later */
          __attribute__((returns_nonnull)) void* foo(void);
@@ -266,52 +271,60 @@ AC_COMPILE_IFELSE([
    )],
    AC_DEFINE([HAVE_ATTR_RETNONNULL], 1, [The returns_nonnull attribute is supported.])
    AC_MSG_RESULT(yes),
-   AC_MSG_RESULT(no))
+   AC_MSG_RESULT(no)
+)
 CFLAGS="$old_CFLAGS"
 
 AC_MSG_CHECKING(for NaN support)
 dnl Note: AC_RUN_IFELSE does not try compiling the program at all when
 dnl $cross_compiling is 'yes'.
 AC_LINK_IFELSE(
-   [AC_LANG_PROGRAM([[
+   [AC_LANG_PROGRAM(
+      [[
 #include <math.h>
-   ]], [[
+      ]], [[
       double x = NAN;
       /* Both should evaluate to false -> 0 (exit success) */
       return isgreater(x, x) || isgreaterequal(x, x);
-   ]])],
+      ]]
+   )],
    [
-   flag_finite_math_only=unknown
-   if test "$cross_compiling" = yes; then
-      AC_COMPILE_IFELSE(
-         [AC_LANG_SOURCE([[
+      flag_finite_math_only=unknown
+      if test "$cross_compiling" = yes; then
+         AC_COMPILE_IFELSE(
+            [AC_LANG_SOURCE([[
 /* __FINITE_MATH_ONLY__ is documented in Clang. */
 #ifdef __FINITE_MATH_ONLY__
 #error "should not enable -ffinite-math-only"
 #endif
-         ]])],
-         [AC_MSG_RESULT([assume yes (cross compiling)])],
-         [flag_finite_math_only=yes]
-      )
-   elif ./conftest$EXEEXT >&AS_MESSAGE_LOG_FD; then
-      flag_finite_math_only=no
-      AC_MSG_RESULT(yes)
-   else
-      flag_finite_math_only=yes
-   fi
-   if test "$flag_finite_math_only" = yes; then
+            ]])],
+            [AC_MSG_RESULT([assume yes (cross compiling)])],
+            [flag_finite_math_only=yes]
+         )
+      elif ./conftest$EXEEXT >&AS_MESSAGE_LOG_FD; then
+         flag_finite_math_only=no
+         AC_MSG_RESULT(yes)
+      else
+         flag_finite_math_only=yes
+      fi
+      if test "$flag_finite_math_only" = yes; then
+         AC_MSG_RESULT(no)
+         AC_MSG_WARN([runtime behavior with NaN is not compliant - some functionality might break; consider using '-fno-finite-math-only'])
+      fi
+   ],
+   [
       AC_MSG_RESULT(no)
-      AC_MSG_WARN([runtime behavior with NaN is not compliant - some functionality might break; consider using '-fno-finite-math-only'])
-   fi],
-   [AC_MSG_RESULT(no)
-   AC_MSG_ERROR([can not find required macros: NAN, isgreater() and isgreaterequal()])])
+      AC_MSG_ERROR([can not find required macros: NAN, isgreater() and isgreaterequal()])
+   ]
+)
 
 AC_MSG_CHECKING(for __builtin_ctz)
-AC_COMPILE_IFELSE([
-   AC_LANG_PROGRAM([], [[__builtin_ctz(1); /* Supported in GCC 3.4 or later */]])],
+AC_COMPILE_IFELSE(
+   [AC_LANG_PROGRAM([], [[__builtin_ctz(1); /* Supported in GCC 3.4 or later */]])],
    [AC_DEFINE([HAVE_BUILTIN_CTZ], 1, [Define to 1 if the compiler supports '__builtin_ctz' function.])
    AC_MSG_RESULT(yes)],
-   AC_MSG_RESULT(no))
+   AC_MSG_RESULT(no)
+)
 
 # ----------------------------------------------------------------------
 
@@ -372,7 +385,7 @@ AC_CHECK_FUNCS([ \
     sched_getscheduler \
     sched_setscheduler \
     strnlen \
-   ])
+])
 
 # strchrnul is available in macOS since 15.4, but the user may specify an older
 # macOS version ('-mmacos-version-min') to build for. We need to ensure it is
@@ -382,16 +395,18 @@ CFLAGS="$CFLAGS -Werror"
 
 AC_MSG_CHECKING([for strchrnul])
 AC_LINK_IFELSE(
-   [AC_LANG_PROGRAM([[
+   [AC_LANG_PROGRAM(
+      [[
 #include <string.h>
-   ]], [[
+      ]], [[
       static char ch;
       char* ptr = strchrnul(&ch, 0);
       return ptr != &ch; /* Should be 0 (exit success) */
-   ]])],
+      ]]
+   )],
    [
-   AC_MSG_RESULT(yes)
-   AC_DEFINE([HAVE_STRCHRNUL], [1], [Define to 1 if you have the 'strchrnul' function.])
+      AC_MSG_RESULT(yes)
+      AC_DEFINE([HAVE_STRCHRNUL], [1], [Define to 1 if you have the 'strchrnul' function.])
    ],
    [AC_MSG_RESULT(no)]
 )
@@ -424,37 +439,45 @@ fi
 dnl PKG_PROG_PKG_CONFIG initializes $PKG_CONFIG and related variables.
 dnl If the macro is not called, some pkg-config checks might be skipped
 dnl and $PKG_CONFIG might be unset.
-m4_ifdef([PKG_PROG_PKG_CONFIG], [
-   PKG_PROG_PKG_CONFIG()
-   pkg_m4_included=1 # Makefile might grep this keyword. Don't remove.
-], [
-   m4_warn(
-      [syntax],
-      [pkg.m4 is absent or older than version 0.16; this 'configure' would have incomplete pkg-config support])
-])
-
-AC_ARG_ENABLE([unicode],
-              [AS_HELP_STRING([--enable-unicode],
-                              [enable Unicode support @<:@default=yes@:>@])],
-              [],
-              [enable_unicode=yes])
+m4_ifdef(
+   [PKG_PROG_PKG_CONFIG],
+   [
+      PKG_PROG_PKG_CONFIG()
+      pkg_m4_included=1 # Makefile might grep this keyword. Don't remove.
+   ], [
+      m4_warn(
+         [syntax],
+         [pkg.m4 is absent or older than version 0.16; this 'configure' would have incomplete pkg-config support]
+      )
+   ]
+)
 
+AC_ARG_ENABLE(
+   [unicode],
+   [AS_HELP_STRING(
+      [--enable-unicode],
+      [enable Unicode support @<:@default=yes@:>@]
+   )],
+   [],
+   [enable_unicode=yes]
+)
 AC_ARG_VAR([CURSES_CFLAGS], [C compiler flags for curses; this overrides auto detected values])
 AC_ARG_VAR([CURSES_LIBS], [linker flags for curses; this overrides auto detected values])
 
-curses_pkg_names="ncurses6 ncurses5 ncurses ncursest6 ncursest5 \
-   ncursest curses"
-
+curses_pkg_names="ncurses6 ncurses5 ncurses ncursest6 ncursest5 ncursest curses"
 if test "x$enable_unicode" = xyes; then
-   curses_pkg_names="ncursesw6 ncursesw5 ncursesw ncursestw6 ncursestw5 \
-      ncursestw $curses_pkg_names"
+   curses_pkg_names="ncursesw6 ncursesw5 ncursesw ncursestw6 ncursestw5 ncursestw $curses_pkg_names"
 fi
 
-AC_ARG_WITH([curses],
-            [AS_HELP_STRING([--with-curses=NAME],
-                            [select curses package NAME to link with; e.g. ncursesw6])],
-            [],
-            [with_curses=check])
+AC_ARG_WITH(
+   [curses],
+   [AS_HELP_STRING(
+      [--with-curses=NAME],
+      [select curses package NAME to link with; e.g. ncursesw6]
+   )],
+   [],
+   [with_curses=check]
+)
 case $with_curses in
 check|yes)
    : # No-op. Use default list.
@@ -500,17 +523,19 @@ htop_check_curses_capability () {
    # explicit "-ltinfo" to link (for internal dependency).
    AC_MSG_CHECKING([for keypad in $htop_msg_linker_flags])
    AC_LINK_IFELSE(
-      [AC_LANG_PROGRAM([[
+      [AC_LANG_PROGRAM(
+         [[
 /* int keypad(WINDOW* win, bool enable); */
 int keypad(void* win, int enable);
-      ]], [[
+         ]], [[
          static char dummy;
          keypad((void*)&dummy, 0);
-      ]])],
+         ]]
+      )],
       [AC_MSG_RESULT(yes)],
       [
-      AC_MSG_RESULT(no)
-      htop_curses_status=1
+         AC_MSG_RESULT(no)
+         htop_curses_status=1
       ]
    )
 
@@ -520,18 +545,20 @@ int keypad(void* win, int enable);
    if test "$htop_curses_status" -eq 0; then
       AC_MSG_CHECKING([for doupdate in $htop_msg_linker_flags])
       AC_LINK_IFELSE(
-         [AC_LANG_PROGRAM([[
+         [AC_LANG_PROGRAM(
+            [[
 int doupdate(void);
-         ]], [[
+            ]], [[
             doupdate();
-         ]])],
+            ]]
+         )],
          [
-         AC_MSG_RESULT(yes)
-         htop_curses_capability=nonwide
+            AC_MSG_RESULT(yes)
+            htop_curses_capability=nonwide
          ],
          [
-         AC_MSG_RESULT(no)
-         htop_curses_status=1
+            AC_MSG_RESULT(no)
+            htop_curses_status=1
          ]
       )
    fi
@@ -541,21 +568,23 @@ int doupdate(void);
    if test "x$htop_curses_status$enable_unicode" = x0yes; then
       AC_MSG_CHECKING([for wadd_wchnstr in $htop_msg_linker_flags])
       AC_LINK_IFELSE(
-         [AC_LANG_PROGRAM([[
+         [AC_LANG_PROGRAM(
+            [[
 /* int wadd_wchnstr(WINDOW* win, const cchar_t* wchstr, int n); */
 int wadd_wchnstr(void* win, const void* wchstr, int n);
-         ]], [[
+            ]], [[
             static char dummy1;
             static char dummy2;
             wadd_wchnstr((void*)&dummy1, (void*)&dummy2, 0);
-         ]])],
+            ]]
+         )],
          [
-         AC_MSG_RESULT(yes)
-         htop_curses_capability=wide
+            AC_MSG_RESULT(yes)
+            htop_curses_capability=wide
          ],
          [
-         AC_MSG_RESULT(no)
-         htop_curses_status=1
+            AC_MSG_RESULT(no)
+            htop_curses_status=1
          ]
       )
    fi
@@ -748,12 +777,13 @@ CFLAGS="-I$srcdir $CFLAGS"
 # "stdscr" is a macro in ncursest (reentrant version of ncurses).
 AC_MSG_CHECKING([whether the curses header works])
 AC_LINK_IFELSE(
-   [AC_LANG_PROGRAM([[
+   [AC_LANG_PROGRAM(
+      [[
 /* Checks if the "bool" definition in curses is ISO C compatible */
 #include <stdbool.h>
 
 #include "ProvideCurses.h"
-   ]], [[
+      ]], [[
       keypad(stdscr, false);
 
       refresh();
@@ -764,11 +794,12 @@ AC_LINK_IFELSE(
          mvadd_wchnstr(0, 0, &dummy, 0);
       }
 #endif
-   ]])],
+      ]]
+   )],
    [AC_MSG_RESULT(yes)],
    [
-   AC_MSG_RESULT(no)
-   AC_MSG_FAILURE([there are problems with the curses header])
+      AC_MSG_RESULT(no)
+      AC_MSG_FAILURE([there are problems with the curses header])
    ]
 )
 
@@ -782,36 +813,47 @@ if test "$my_htop_platform" = "solaris"; then
    # Since ncurses macros use the ERR macro, we can not use another name.
    AC_DEFINE([ERR], [(-1)], [Predefine ncurses macro.])
 fi
-AC_CHECK_FUNCS( [set_escdelay] )
-AC_CHECK_FUNCS( [getmouse] )
+AC_CHECK_FUNCS([set_escdelay])
+AC_CHECK_FUNCS([getmouse])
 AC_DEFINE([NCURSES_ENABLE_STDBOOL_H], [1], [Define to enable stdbool.h in ncurses])
 
 
-AC_ARG_ENABLE([affinity],
-              [AS_HELP_STRING([--enable-affinity],
-                              [enable sched_setaffinity and sched_getaffinity for affinity support, conflicts with hwloc @<:@default=check@:>@])],
-              [],
-              [enable_affinity=check])
+AC_ARG_ENABLE(
+   [affinity],
+   [AS_HELP_STRING(
+      [--enable-affinity],
+      [enable sched_setaffinity and sched_getaffinity for affinity support, conflicts with hwloc @<:@default=check@:>@]
+   )],
+   [],
+   [enable_affinity=check]
+)
 if test "x$enable_affinity" = xcheck; then
    if test "x$enable_hwloc" = xyes; then
       enable_affinity=no
    else
       AC_MSG_CHECKING([for usable sched_setaffinity])
-      AC_RUN_IFELSE([
-         AC_LANG_PROGRAM([[
-            #include <sched.h>
-            #include <errno.h>
-            static cpu_set_t cpuset;
-         ]], [[
-            CPU_ZERO(&cpuset);
-            sched_setaffinity(0, sizeof(cpu_set_t), &cpuset);
-            if (errno == ENOSYS) return 1;
-         ]])],
-      [enable_affinity=yes
-         AC_MSG_RESULT([yes])],
-      [enable_affinity=no
-         AC_MSG_RESULT([no])],
-      [AC_MSG_RESULT([assume yes (cross compiling)])])
+      AC_RUN_IFELSE(
+         [AC_LANG_PROGRAM(
+            [[
+               #include <sched.h>
+               #include <errno.h>
+               static cpu_set_t cpuset;
+            ]], [[
+               CPU_ZERO(&cpuset);
+               sched_setaffinity(0, sizeof(cpu_set_t), &cpuset);
+               if (errno == ENOSYS) return 1;
+            ]]
+         )],
+         [
+            enable_affinity=yes
+            AC_MSG_RESULT([yes])
+         ],
+         [
+            enable_affinity=no
+            AC_MSG_RESULT([no])
+         ],
+         [AC_MSG_RESULT([assume yes (cross compiling)])]
+      )
    fi
 fi
 if test "x$enable_affinity" = xyes; then
@@ -847,7 +889,7 @@ $3
          AM_CPPFLAGS="$AM_CPPFLAGS$htop_header_dir"
          htop_header_search_status=0
          break
-],
+         ],
          [htop_msg_header_dir="(not found)"]
       )
    done
@@ -857,11 +899,15 @@ $3
    return "$htop_header_search_status"
 } # htop_search_header_dir
 
-AC_ARG_ENABLE([unwind],
-              [AS_HELP_STRING([--enable-unwind],
-              [enable unwind support for printing backtraces; requires libunwind @<:@default=check@:>@])],
-              [],
-              [enable_unwind=check])
+AC_ARG_ENABLE(
+   [unwind],
+   [AS_HELP_STRING(
+      [--enable-unwind],
+      [enable unwind support for printing backtraces; requires libunwind @<:@default=check@:>@]
+   )],
+   [],
+   [enable_unwind=check]
+)
 case "$enable_unwind" in
    no)
       ;;
@@ -870,11 +916,17 @@ case "$enable_unwind" in
          AC_CHECK_LIB([lzma], [lzma_index_buffer_decode])
       fi
 
-      AC_CHECK_LIB([unwind], [unw_init_local], [],
-         [if test "$enable_unwind" = yes; then
-            AC_MSG_ERROR([can not find required library libunwind])
-         fi
-         enable_unwind=no])
+      AC_CHECK_LIB(
+         [unwind],
+         [unw_init_local],
+         [],
+         [
+            if test "$enable_unwind" = yes; then
+               AC_MSG_ERROR([can not find required library libunwind])
+            fi
+            enable_unwind=no
+         ]
+      )
       ;;
    *)
       AC_MSG_ERROR([bad value '$enable_unwind' for --enable-unwind])
@@ -889,9 +941,10 @@ elif htop_search_header_dir libunwind.h "/usr/include/libunwind"; then
 
    AC_MSG_CHECKING([whether local unwinding works])
    AC_LINK_IFELSE(
-      [AC_LANG_PROGRAM([[
+      [AC_LANG_PROGRAM(
+         [[
 #include <libunwind.h>
-      ]], [[
+         ]], [[
          /* If libunwind is built with remote unwinding only,
             unw_init_local() will be a dummy function that returns
             -UNW_EINVAL when called. */
@@ -902,38 +955,39 @@ elif htop_search_header_dir libunwind.h "/usr/include/libunwind"; then
             ret = unw_init_local(&cursor, &context);
          }
          return (ret <= 0 && ret > -126 ? -ret : 1);
-      ]])],
+         ]]
+      )],
       [
-      libunwind_local_support=unknown
-      if test "$cross_compiling" = yes; then
-         AC_COMPILE_IFELSE(
-            [AC_LANG_SOURCE([[
+         libunwind_local_support=unknown
+         if test "$cross_compiling" = yes; then
+            AC_COMPILE_IFELSE(
+               [AC_LANG_SOURCE([[
 #include <libunwind.h>
 #ifdef UNW_REMOTE_ONLY
 #error "local unwinding not supported"
 #endif
-            ]])],
-            [AC_MSG_RESULT([assume yes (cross compiling)])],
-            [libunwind_local_support=no]
-         )
-      elif ./conftest$EXEEXT >&AS_MESSAGE_LOG_FD; then
-         libunwind_local_support=yes
-         AC_MSG_RESULT([yes])
-      else
-         libunwind_local_support=no
-      fi
-      if test "$libunwind_local_support" = no; then
-         AC_MSG_RESULT([no])
-         if test "$enable_unwind" = yes; then
-            AC_MSG_WARN([this build of libunwind might not support local unwinding])
+               ]])],
+               [AC_MSG_RESULT([assume yes (cross compiling)])],
+               [libunwind_local_support=no]
+            )
+         elif ./conftest$EXEEXT >&AS_MESSAGE_LOG_FD; then
+            libunwind_local_support=yes
+            AC_MSG_RESULT([yes])
          else
-            enable_unwind=no
+            libunwind_local_support=no
+         fi
+         if test "$libunwind_local_support" = no; then
+            AC_MSG_RESULT([no])
+            if test "$enable_unwind" = yes; then
+               AC_MSG_WARN([this build of libunwind might not support local unwinding])
+            else
+               enable_unwind=no
+            fi
          fi
-      fi
       ],
       [
-      AC_MSG_RESULT([no])
-      AC_MSG_FAILURE([there are problems with libunwind.h])
+         AC_MSG_RESULT([no])
+         AC_MSG_FAILURE([there are problems with libunwind.h])
       ]
    )
 
@@ -960,26 +1014,27 @@ else
       ]])],
       [AC_MSG_RESULT([user defined])],
       [
-      htop_save_CFLAGS=$CFLAGS
-      CFLAGS="$AM_CFLAGS $CFLAGS -Werror"
-
-      # Use 'typeof' if supported. Otherwise guess types from larger bit width
-      # to smaller.
-      # 'size_t' is used in FreeBSD, NetBSD and OpenBSD.
-      # 'int' is used in Linux (with glibc) and macOS.
-      # See also: Gnulib documentation of 'backtrace'
-      # (https://www.gnu.org/software/gnulib/manual/html_node/backtrace.html).
-      # GCC (< 15) had a bug that warns on a non-null pointer in 'typeof'.
-      # Workaround by passing an address of 1 instead.
-      for backtrace_return_type in \
-         "typeof(backtrace((void**)1, 0))" \
-         "__typeof__(backtrace((void**)1, 0))" \
-         "__typeof(backtrace((void**)1, 0))" \
-         size_t \
-         int \
-      ; do
-         AC_LINK_IFELSE(
-            [AC_LANG_PROGRAM([[
+         htop_save_CFLAGS=$CFLAGS
+         CFLAGS="$AM_CFLAGS $CFLAGS -Werror"
+
+         # Use 'typeof' if supported. Otherwise guess types from larger bit width
+         # to smaller.
+         # 'size_t' is used in FreeBSD, NetBSD and OpenBSD.
+         # 'int' is used in Linux (with glibc) and macOS.
+         # See also: Gnulib documentation of 'backtrace'
+         # (https://www.gnu.org/software/gnulib/manual/html_node/backtrace.html).
+         # GCC (< 15) had a bug that warns on a non-null pointer in 'typeof'.
+         # Workaround by passing an address of 1 instead.
+         for backtrace_return_type in \
+            "typeof(backtrace((void**)1, 0))" \
+            "__typeof__(backtrace((void**)1, 0))" \
+            "__typeof(backtrace((void**)1, 0))" \
+            size_t \
+            int \
+         ; do
+            AC_LINK_IFELSE(
+               [AC_LANG_PROGRAM(
+                  [[
 #include <execinfo.h>
 #include <stddef.h> /* For size_t */
 
@@ -987,41 +1042,48 @@ else
 /* Assume the second argument of backtrace_symbols() uses the same type
    as the return type of backtrace(). */
 char** (*fn)(void* const*, BACKTRACE_RETURN_TYPE) = backtrace_symbols;
-            ]], [[
+                  ]], [[
                static void* addr;
                BACKTRACE_RETURN_TYPE nptrs = backtrace(&addr, 1);
                (void)nptrs;
-            ]])],
-            [
-            AC_DEFINE_UNQUOTED([BACKTRACE_RETURN_TYPE], [$backtrace_return_type], [Defined to the return type of the 'backtrace' function.])
-            break
-            ],
-            [backtrace_return_type=""]
-         )
-      done
-      backtrace_return_type_msg=`echo "x$backtrace_return_type" | sed '
-         s/^x *//
-         s/^$/not available/
-         s/^\(_*typeof_*\) *( *backtrace( *( *void *\* *\* *) *1 *, *0 *) *) *$/automatic (using \1)/'`
-      AC_MSG_RESULT([$backtrace_return_type_msg])
-
-      CFLAGS=$htop_save_CFLAGS
+                  ]]
+               )],
+               [
+                  AC_DEFINE_UNQUOTED([BACKTRACE_RETURN_TYPE], [$backtrace_return_type], [Defined to the return type of the 'backtrace' function.])
+                  break
+               ],
+               [backtrace_return_type=""]
+            )
+         done
+         backtrace_return_type_msg=`echo "x$backtrace_return_type" | sed '
+            s/^x *//
+            s/^$/not available/
+            s/^\(_*typeof_*\) *( *backtrace( *( *void *\* *\* *) *1 *, *0 *) *) *$/automatic (using \1)/'`
+         AC_MSG_RESULT([$backtrace_return_type_msg])
+
+         CFLAGS=$htop_save_CFLAGS
       ]
    )
 fi
 
 
-AC_ARG_ENABLE([hwloc],
-              [AS_HELP_STRING([--enable-hwloc],
-              [enable hwloc support for CPU affinity; disables affinity support; requires libhwloc @<:@default=no@:>@])],
-              [],
-              [enable_hwloc=no])
+AC_ARG_ENABLE(
+   [hwloc],
+   [AS_HELP_STRING(
+      [--enable-hwloc],
+      [enable hwloc support for CPU affinity; disables affinity support; requires libhwloc @<:@default=no@:>@]
+   )],
+   [],
+   [enable_hwloc=no]
+)
 case "$enable_hwloc" in
    no)
       ;;
    yes)
-      m4_ifdef([PKG_PROG_PKG_CONFIG], [
-         PKG_CHECK_MODULES(HWLOC, hwloc, [
+      m4_ifdef(
+         [PKG_PROG_PKG_CONFIG],
+         [
+            PKG_CHECK_MODULES(HWLOC, hwloc, [
                AM_CFLAGS="$AM_CFLAGS $HWLOC_CFLAGS"
                LIBS="$LIBS $HWLOC_LIBS"
                AC_DEFINE([HAVE_LIBHWLOC], [1], [Define to 1 if you have the 'hwloc' library (-lhwloc).])
@@ -1032,7 +1094,8 @@ case "$enable_hwloc" in
          ], [
             AC_CHECK_LIB([hwloc], [hwloc_get_proc_cpubind], [], [AC_MSG_ERROR([can not find required library libhwloc])])
             AC_CHECK_HEADERS([hwloc.h], [], [AC_MSG_ERROR([can not find require header file hwloc.h])])
-         ])
+         ]
+      )
       ;;
    *)
       AC_MSG_ERROR([bad value '$enable_hwloc' for --enable-hwloc])
@@ -1040,11 +1103,15 @@ case "$enable_hwloc" in
 esac
 
 
-AC_ARG_WITH([os-release],
-            [AS_HELP_STRING([--with-os-release=FILE],
-                            [location of an os-release file @<:@default=/etc/os-release@:>@])],
-            [],
-            [with_os_release=/etc/os-release])
+AC_ARG_WITH(
+   [os-release],
+   [AS_HELP_STRING(
+      [--with-os-release=FILE],
+      [location of an os-release file @<:@default=/etc/os-release@:>@]
+   )],
+   [],
+   [with_os_release=/etc/os-release]
+)
 if test -n "$with_os_release" && test ! -f "$with_os_release"; then
    if test -f "/usr/lib/os-release"; then
       with_os_release="/usr/lib/os-release"
@@ -1052,11 +1119,15 @@ if test -n "$with_os_release" && test ! -f "$with_os_release"; then
 fi
 AC_DEFINE_UNQUOTED([OSRELEASEFILE], ["$with_os_release"], [File with OS release details.])
 
-AC_ARG_WITH([config],
-            [AS_HELP_STRING([--with-config=DIR],
-                            [configuration path @<:@default=/.config@:>@])],
-            [],
-            [with_config="/.config"])
+AC_ARG_WITH(
+   [config],
+   [AS_HELP_STRING(
+      [--with-config=DIR],
+      [configuration path @<:@default=/.config@:>@]
+   )],
+   [],
+   [with_config="/.config"]
+)
 dnl Performance Co-Pilot configuration location to prevent overwrite
 if test "$my_htop_platform" = pcp -a "$with_config" = /.config; then
     with_config="/.pcp"
@@ -1072,42 +1143,58 @@ AC_DEFINE_UNQUOTED([CONFIGDIR], ["$with_config"], [Configuration path.])
 # Checks for Linux features and flags.
 # ----------------------------------------------------------------------
 
-AC_ARG_WITH([proc],
-            [AS_HELP_STRING([--with-proc=DIR],
-                            [location of a Linux-compatible proc filesystem @<:@default=/proc@:>@])],
-            [],
-            [with_proc=/proc])
+AC_ARG_WITH(
+   [proc],
+   [AS_HELP_STRING(
+      [--with-proc=DIR],
+      [location of a Linux-compatible proc filesystem @<:@default=/proc@:>@]
+   )],
+   [],
+   [with_proc=/proc]
+)
 if test -z "$with_proc"; then
    AC_MSG_ERROR([bad empty value for --with-proc option])
 fi
 AC_DEFINE_UNQUOTED([PROCDIR], ["$with_proc"], [Path of proc filesystem.])
 
 
-AC_ARG_ENABLE([openvz],
-              [AS_HELP_STRING([--enable-openvz],
-                              [enable OpenVZ support @<:@default=no@:>@])],
-              [],
-              [enable_openvz=no])
+AC_ARG_ENABLE(
+   [openvz],
+   [AS_HELP_STRING(
+      [--enable-openvz],
+      [enable OpenVZ support @<:@default=no@:>@]
+   )],
+   [],
+   [enable_openvz=no]
+)
 if test "x$enable_openvz" = xyes; then
    AC_DEFINE([HAVE_OPENVZ], [1], [Define if openvz support enabled.])
 fi
 
 
-AC_ARG_ENABLE([vserver],
-              [AS_HELP_STRING([--enable-vserver],
-                              [enable VServer support @<:@default=no@:>@])],
-              [],
-              [enable_vserver=no])
+AC_ARG_ENABLE(
+   [vserver],
+   [AS_HELP_STRING(
+      [--enable-vserver],
+      [enable VServer support @<:@default=no@:>@]
+   )],
+   [],
+   [enable_vserver=no]
+)
 if test "x$enable_vserver" = xyes; then
    AC_DEFINE([HAVE_VSERVER], [1], [Define if VServer support enabled.])
 fi
 
 
-AC_ARG_ENABLE([ancient_vserver],
-              [AS_HELP_STRING([--enable-ancient-vserver],
-                              [enable ancient VServer support (implies --enable-vserver) @<:@default=no@:>@])],
-              [],
-              [enable_ancient_vserver=no])
+AC_ARG_ENABLE(
+   [ancient_vserver],
+   [AS_HELP_STRING(
+      [--enable-ancient-vserver],
+      [enable ancient VServer support (implies --enable-vserver) @<:@default=no@:>@]
+   )],
+   [],
+   [enable_ancient_vserver=no]
+)
 if test "x$enable_ancient_vserver" = xyes; then
    if test "x$enable_vserver" != xyes; then
       enable_vserver=implied
@@ -1117,11 +1204,15 @@ if test "x$enable_ancient_vserver" = xyes; then
 fi
 
 
-AC_ARG_ENABLE([capabilities],
-              [AS_HELP_STRING([--enable-capabilities],
-                              [enable Linux capabilities support; requires libcap @<:@default=check@:>@])],
-              [],
-              [enable_capabilities=check])
+AC_ARG_ENABLE(
+   [capabilities],
+   [AS_HELP_STRING(
+      [--enable-capabilities],
+      [enable Linux capabilities support; requires libcap @<:@default=check@:>@]
+   )],
+   [],
+   [enable_capabilities=check]
+)
 case "$enable_capabilities" in
    no)
       ;;
@@ -1151,12 +1242,14 @@ htop_try_link_libnl3 () {
       LDFLAGS="$LDFLAGS -L$1"
    fi
    AC_LINK_IFELSE(
-      [AC_LANG_PROGRAM([[
+      [AC_LANG_PROGRAM(
+         [[
 /* struct nl_sock* nl_socket_alloc(void); */
 void* nl_socket_alloc(void);
-      ]], [[
+         ]], [[
          void* sock = nl_socket_alloc();
-      ]])],
+         ]]
+      )],
       [htop_libnl3_link_succeed=yes],
       [htop_libnl3_link_succeed=no]
    )
@@ -1165,11 +1258,15 @@ void* nl_socket_alloc(void);
    LIBS=$htop_save_LIBS
 } # htop_try_link_libnl3
 
-AC_ARG_ENABLE([delayacct],
-              [AS_HELP_STRING([--enable-delayacct],
-                              [enable Linux delay accounting support; requires libnl-3 and libnl-genl-3 @<:@default=check@:>@])],
-              [],
-              [enable_delayacct=check])
+AC_ARG_ENABLE(
+   [delayacct],
+   [AS_HELP_STRING(
+      [--enable-delayacct],
+      [enable Linux delay accounting support; requires libnl-3 and libnl-genl-3 @<:@default=check@:>@]
+   )],
+   [],
+   [enable_delayacct=check]
+)
 case "$enable_delayacct" in
    no|yes)
       ;;
@@ -1188,9 +1285,10 @@ esac
 case "$enable_delayacct" in
    check|yes)
       if test "x${LIBNL3_CFLAGS+y}" = x; then
-         m4_ifdef([PKG_PROG_PKG_CONFIG],
+         m4_ifdef(
+            [PKG_PROG_PKG_CONFIG],
             [PKG_CHECK_MODULES(LIBNL3, libnl-3.0, [], [LIBNL3_CFLAGS="-I/usr/include/libnl3"])],
-           [LIBNL3_CFLAGS="-I/usr/include/libnl3"]
+                [LIBNL3_CFLAGS="-I/usr/include/libnl3"]
          )
       fi
 
@@ -1201,10 +1299,10 @@ case "$enable_delayacct" in
          [netlink/attr.h netlink/handlers.h netlink/msg.h],
          [],
          [
-         if test "$enable_delayacct" = yes; then
-            AC_MSG_ERROR([can not find required header files netlink/attr.h, netlink/handlers.h, netlink/msg.h])
-         fi
-         enable_delayacct=no
+            if test "$enable_delayacct" = yes; then
+               AC_MSG_ERROR([can not find required header files netlink/attr.h, netlink/handlers.h, netlink/msg.h])
+            fi
+            enable_delayacct=no
          ]
       )
       CFLAGS=$htop_save_CFLAGS
@@ -1238,17 +1336,21 @@ case "$enable_delayacct" in
       ;;
 esac
 if test "$enable_delayacct" = yes; then
-  AC_DEFINE([HAVE_DELAYACCT], [1], [Define if delay accounting support should be enabled.])
-  AM_CFLAGS="$AM_CFLAGS $LIBNL3_CFLAGS"
+   AC_DEFINE([HAVE_DELAYACCT], [1], [Define if delay accounting support should be enabled.])
+   AM_CFLAGS="$AM_CFLAGS $LIBNL3_CFLAGS"
 fi
 AM_CONDITIONAL([HAVE_DELAYACCT], [test "$enable_delayacct" = yes])
 
 
-AC_ARG_ENABLE([sensors],
-              [AS_HELP_STRING([--enable-sensors],
-                              [enable libsensors support for reading temperature data; requires only libsensors headers at compile time, at runtime libsensors is loaded via dlopen @<:@default=check@:>@])],
-              [],
-              [enable_sensors=check])
+AC_ARG_ENABLE(
+   [sensors],
+   [AS_HELP_STRING(
+      [--enable-sensors],
+      [enable libsensors support for reading temperature data; requires only libsensors headers at compile time, at runtime libsensors is loaded via dlopen @<:@default=check@:>@]
+   )],
+   [],
+   [enable_sensors=check]
+)
 case "$enable_sensors" in
    no)
       ;;
@@ -1299,40 +1401,53 @@ AM_CFLAGS="$AM_CFLAGS\
  -Wwrite-strings"
 
 dnl https://www.gnu.org/software/autoconf-archive/ax_check_compile_flag.html
-AC_DEFUN([AX_CHECK_COMPILE_FLAG],
-[
-AS_VAR_PUSHDEF([CACHEVAR],[ax_cv_check_[]_AC_LANG_ABBREV[]flags_$4_$1])dnl
-AC_CACHE_CHECK([whether _AC_LANG compiler accepts $1], CACHEVAR, [
-   ax_check_save_flags=$[]_AC_LANG_PREFIX[]FLAGS
-   _AC_LANG_PREFIX[]FLAGS="$[]_AC_LANG_PREFIX[]FLAGS $4 $1"
-   AC_COMPILE_IFELSE([m4_default([$5],[AC_LANG_PROGRAM()])],
-      [AS_VAR_SET(CACHEVAR,[yes])],
-      [AS_VAR_SET(CACHEVAR,[no])])
-   _AC_LANG_PREFIX[]FLAGS=$ax_check_save_flags])
-AS_VAR_IF(CACHEVAR,yes,
-   [m4_default([$2], :)],
-   [m4_default([$3], :)])
-AS_VAR_POPDEF([CACHEVAR])dnl
-])dnl AX_CHECK_COMPILE_FLAGS
+AC_DEFUN(
+   [AX_CHECK_COMPILE_FLAG],
+   [
+      AS_VAR_PUSHDEF([CACHEVAR],[ax_cv_check_[]_AC_LANG_ABBREV[]flags_$4_$1])dnl
+      AC_CACHE_CHECK([whether _AC_LANG compiler accepts $1], CACHEVAR, [
+         ax_check_save_flags=$[]_AC_LANG_PREFIX[]FLAGS
+         _AC_LANG_PREFIX[]FLAGS="$[]_AC_LANG_PREFIX[]FLAGS $4 $1"
+         AC_COMPILE_IFELSE([m4_default([$5],[AC_LANG_PROGRAM()])],
+            [AS_VAR_SET(CACHEVAR,[yes])],
+            [AS_VAR_SET(CACHEVAR,[no])])
+         _AC_LANG_PREFIX[]FLAGS=$ax_check_save_flags
+      ])
+      AS_VAR_IF(CACHEVAR,yes,
+         [m4_default([$2], :)],
+         [m4_default([$3], :)]
+      )
+      AS_VAR_POPDEF([CACHEVAR])dnl
+   ]
+)
+dnl AX_CHECK_COMPILE_FLAGS
 
 AX_CHECK_COMPILE_FLAG([-Wextra-semi-stmt],         [AM_CFLAGS="$AM_CFLAGS -Wextra-semi-stmt"],         , [-Werror=unknown-warning-option]) dnl the autoconf check itself generates -Wextra-semi-stmt
 AX_CHECK_COMPILE_FLAG([-Wimplicit-int-conversion], [AM_CFLAGS="$AM_CFLAGS -Wimplicit-int-conversion"], , [-Werror])
 AX_CHECK_COMPILE_FLAG([-Wnull-dereference],        [AM_CFLAGS="$AM_CFLAGS -Wnull-dereference"],        , [-Werror])
 
-AC_ARG_ENABLE([werror],
-              [AS_HELP_STRING([--enable-werror],
-                              [Treat warnings as errors @<:@default=no@:>@])],
-              [],
-              [enable_werror=no])
+AC_ARG_ENABLE(
+   [werror],
+   [AS_HELP_STRING(
+      [--enable-werror],
+      [Treat warnings as errors @<:@default=no@:>@]
+   )],
+   [],
+   [enable_werror=no]
+)
 if test "x$enable_werror" = xyes; then
    AM_CFLAGS="$AM_CFLAGS -Werror"
 fi
 
-AC_ARG_ENABLE([debug],
-              [AS_HELP_STRING([--enable-debug],
-                              [Enable compiling with maximum debug info, asserts and internal sanity checks @<:@default=no@:>@])],
-              [],
-              [enable_debug=no])
+AC_ARG_ENABLE(
+   [debug],
+   [AS_HELP_STRING(
+      [--enable-debug],
+      [Enable compiling with maximum debug info, asserts and internal sanity checks @<:@default=no@:>@]
+   )],
+   [],
+   [enable_debug=no]
+)
 if test "x$enable_debug" != xyes; then
    AM_CPPFLAGS="$AM_CPPFLAGS -DNDEBUG"
 
@@ -1344,13 +1459,13 @@ if test "x$enable_debug" != xyes; then
       ]])],
       [:],
       [
-      warning_msg="signaling NaN support is enabled; not recommended for htop"
-      case "$CC" in
-      *gcc*)
-         warning_msg="$warning_msg (use '-fno-signaling-nans' compiler flag to disable)"
-         ;;
-      esac
-      AC_MSG_WARN([$warning_msg])
+         warning_msg="signaling NaN support is enabled; not recommended for htop"
+         case "$CC" in
+         *gcc*)
+            warning_msg="$warning_msg (use '-fno-signaling-nans' compiler flag to disable)"
+            ;;
+         esac
+         AC_MSG_WARN([$warning_msg])
       ]
    )
 else
```

If I didn't miss anything, this should be consistent throughout.

Explorer09:
@BenBE You are suggesting me to apply the indent style to the whole `configure.ac`? (I was hesitant to do that one yet as it can affect `git blame` of other people's change lines.)

I think you grabbed the most ideas of my proposed style, but there are a few points I wish to discuss a bit:

1. When it comes to quoted multi-line code, the M4 quotation marks may be either on the same line as the macro call or its closing parenthesis, or on a new line of its own. When the quotation marks are on a new line they should get an extra level of indent. That is, either

```m4
   [AC_LANG_PROGRAM([[
#include <stdbool.h>
   ]], [[
      bool flag = false;
   ]])]
```

or

```m4
   [AC_LANG_PROGRAM(
      [[
#include <stdbool.h>
      ]],
      [[
      bool flag = false;
      ]]
   )]
```

But, I don't like when the code got one more level of indent, simply because the quotes are in their own level of indent. I.e. I wish to avoid this:

```m4
   [AC_LANG_PROGRAM(
      [[
#include <stdbool.h>
      ]],
      [[
         bool flag = false; // extra indent
      ]]
   )]
```

When the code is indented in a macro call, I expect at most one level of indent in total.

2. I think you should leave the `AX_CHECK_COMPILE_FLAG` code alone. That macro is copied from Autoconf Archive and we should leave the original style as is. Alternatively make it a module in the `m4` folder.

3. I think it's okay for some macros to have the first few arguments on the same line as the macro token and the opening parenthesis. This is useful for macros that either define or locate a particular token, and that token won't be very long to justify the waste of one line of screen space. Macros of this kind include `AC_DEFINE`, `m4_ifdef` `AC_CHECK_HEADERS`, `AC_CHECK_LIB` and others.

```m4
AC_CHECK_LIB([unwind], [unw_init_local],
   [],
   [
   if test "$enable_unwind" = yes; then
      AC_MSG_ERROR([can not find required library libunwind])
   fi
   enable_unwind=no
   ]
)

AC_ARG_ENABLE([werror],
   [AS_HELP_STRING([--enable-werror],
      [Treat warnings as errors @<:@default=no@:>@]
   )],
   [],
   [enable_werror=no]
)
```

configure.ac
```diff
@@ -159,26 +175,45 @@ esac
...
+   [
+      if test "x$ac_cv_header_sys_mkdev_h" != xyes; then
+         AC_CHECK_HEADER(
+            [sys/sysmacros.h],
+            [AC_DEFINE(
+               [MAJOR_IN_SYSMACROS],
+               [1],
+               [Define to 1 if 'major', 'minor', and 'makedev' are declared in <sys/sysmacros.h>.]
+            )]
+         )
```
Explorer09, line 213: Here are a pair of examples that I feel the extra line wrapping around the first and second arguments is not worth it. It's analogous to `if` conditionals in C where you are forced to put conditional on a line separate from the `if` keyword.

```c
   if \
      (String_eq(option[0], "config_reader_min_version")) {
      this->config_version = atoi(option[1]);
      // ... mode code ...
   }
   // The above looks ugly
```

I would vote for this:

```m4
         AC_CHECK_HEADER([sys/sysmacros.h],
            [AC_DEFINE([MAJOR_IN_SYSMACROS], [1],
               [Define to 1 if 'major', 'minor', and 'makedev' are declared in <sys/sysmacros.h>.]
            )]
         )
```

BenBE: That change was intentional for consistency. Where possible those checks have been kept on one line altogether for brevity though.

PR #1780, opened 2025-10-06
Update constant width font specifier in pcp-htop.5

natoscott:
Current versions of Fedora "man 5 pcp-htop" produces:

troff:<standard input>:37: warning: cannot select font 'CW'
troff:<standard input>:161: warning: cannot select font 'CW'

Newer versions of groff now issue warnings about this old name for constant width (CW) fonts - switch to a modern alternative that is well supported (courier regular - CR) to resolve this.


natoscott:
CI failure here is tackled by #1781

PR #1781, opened 2025-10-06
Fixes for build warnings on macOS with latest toolchain

natoscott:


Panel.c
```diff
@@ -421,7 +421,7 @@ bool Panel_onKey(Panel* this, int key) {
 
       case KEY_CTRL('E'):
       case '$':
-         this->scrollH = MAXIMUM(this->selectedLen - this->w, 0);
+         this->scrollH = MAXIMUM((int)this->selectedLen - this->w, 0);
```
BenBE: This could be improved to a `CLAMP`.

darwin/DarwinMachine.c
```diff
@@ -47,7 +47,7 @@ static void DarwinMachine_freeCPULoadInfo(processor_cpu_load_info_t* p) {
    *p = NULL;
 }
 
-static size_t DarwinMachine_allocateCPULoadInfo(processor_cpu_load_info_t* p) {
+static unsigned int DarwinMachine_allocateCPULoadInfo(processor_cpu_load_info_t* p) {
```
BenBE, line 50: Any particular reason for changing the signature here to `unsigned int`? Too many callers?

Potentially update `activeCPUs`/`existingCPUs` to `size_t`?

Vector.c
```diff
@@ -79,10 +79,9 @@ bool Vector_countEquals(const Vector* this, unsigned int expectedCount) {
    return n == expectedCount;
 }
 
-Object* Vector_get(const Vector* this, int idx) {
-   assert(idx >= 0);
-   assert(idx < this->arraySize);
-   assert(idx < this->items);
+Object* Vector_get(const Vector* this, size_t idx) {
+   assert(idx < (size_t)this->arraySize);
+   assert(idx < (size_t)this->items);
```
BenBE, line 84: There should be a follow-up making the whole `Vector` implementation internally use `size_t`.

BenBE (approved):
Nothing too critical …

Vector.c
natoscott: Yep, I tried to limit the scope of the change here so we can (continue to) tackle this in small sections.

darwin/DarwinMachine.c
natoscott: Yep, this was done to limit the scope of the changes here to fix macos CI - this balloons out pretty quickly to all the platforms otherwise (but I agree, its a change that would be good to follow up with - there's still alot than can transition to size_t still).

Panel.c
natoscott: Hmm, can you expand on that?  I'm not seeing how to express it more clearly with CLAMP - thanks!

Explorer09:
I need to take time to review this. This is obviously a subset of issue #1673. But a few of the changes there require promoting or demoting the data types of certain structures so I didn't start working on that.

Header.c
```diff
@@ -198,8 +198,8 @@ void Header_draw(const Header* this) {
    for (int y = 0; y < height; y++) {
       mvhline(y, 0, ' ', COLS);
    }
-   const int numCols = HeaderLayout_getColumns(this->headerLayout);
-   const int width = COLS - 2 * pad - (numCols - 1);
+   const size_t numCols = HeaderLayout_getColumns(this->headerLayout);
+   const int width = COLS - 2 * pad - ((int)numCols - 1);
```
Explorer09, line 202: I don't like this. If `numCols` needs to be downcast, then `numCols` should be an `unsigned int` type. Using `size_t` would be overkill.

Panel.c
Explorer09: This `int` cast suppresses the potential overflow of `this->selectedLen`. I didn't see code anywhere else guarding against `selectedLen` overflow (`> INT_MAX`) so I was hesitant to patch this one with a simple downcast.

Header.c
natoscott: numCols was previously being downcast implicitly (HeaderLayout_getColumns returns size_t), now it is explicit.  The use of an explicit cast here now is to limit the scope of changes to a small, manageable set.  I expect many more things will transition to size_t, ideally also in small manageable increments, and these explicit casts will be slowly dropped out of the code over time.

Explorer09: @natoscott I previously suggested that `HeaderLayout_getColumns` should return `int` or `unsigned int` in order to align with the type limit of terminal columns (`COLS`). `size_t` brings no advantage here. If the number of columns in HeaderLayout is greater than `COLS` then some columns will never be displayed. In other words it makes no sense to allocate so many columns for the Header UI.

Panel.c
BenBE: > Hmm, can you expand on that? I'm not seeing how to express it more clearly with CLAMP - thanks!

The maximum currently only ensures the lower bound for the width; when using `CLAMP` both the lower bound (`0`) and upper bound (`INT_MAX`) can be checked. Particular helpful to prepare for when further moving towards `size_t` in the code base.

natoscott: Oh yes ISWYM now, will push an update shortly.

Header.c
natoscott: @Explorer09 you're advocating for an unrelated change to HeaderLayout_getColumns - this PR aims for a minimal set of changes to resolve the immediate CI issue, so I'm going to leave this aspect as-is.

BenBE:
Changes LGTM for now. I think we can merge this as-is and expand the patch set later.

Panel.c
```diff
@@ -421,7 +422,7 @@ bool Panel_onKey(Panel* this, int key) {
 
       case KEY_CTRL('E'):
       case '$':
-         this->scrollH = MAXIMUM(this->selectedLen - this->w, 0);
+         this->scrollH = CLAMP((int)this->selectedLen - this->w, 0, INT_MAX);
```
Explorer09, line 425: I need to remind that this `CLAMP` at `INT_MAX` doesn't really work. Signed integer overflow is undefined behavior, and the compiler can safely assume any `(int)x > INT_MAX` is `false`.

BenBE: This is an intermediary solution, as the upper limit by ncurses is INT_MAX (based on their API), but when transitioning to `size_t` later, this will become a proper bound. A proper compiler will optimize this current check out (as unnecessary), but this nonetheless documents the proper, expected limits of the values in the calculation (which is why I suggested this change in the first place).

Explorer09: @BenBE Yeah, except that `Panel.scrollH` should probably not migrate to `size_t` in the first place.
I still find this `CLAMP` macro call weird. Until I get a big picture about the `Panel` data structure, I would leave this one as is for now.

BenBE: The annoying thing with n _curse_ s is their antiquated API, which has been left behind in the Stone Ages of computer science. So regardless of what you do, you'll end up doing some form of casts to put data to n _curse_ s or retrieve information from the API; but inside htop we should strive to be consistent with anything referencing object counts to use `size_t`. And that's where things _will_ escalate quickly; e.g. with the `Vector` API, which is references in many places where also drawing stuff is involved.

Explorer09: @BenBE No, that wasn't what I mean. I'm aware of "curses" APIs and their limitations. What I had is _another_ API limitation that comes from `wcwidth(3)` and its family. In particular, I don't think we can make the type of terminal-width-related things anything larger than `signed int`. Using `size_t` for `Vector` sizes is fine. My concern was on members representing terminal width or columns. Because of `wcwidth` limitation, we're doomed (i.e. using any type larger than `int` will be futile).

BenBE: Yes, and that's exactly, what the intention with the CLAMP is: Use proper types to present the semantics (`size_t`), while ensuring the proper range (`CLAMP`) …

I even had a quick look earlier to see if I remember correctly about the limits of the console output stuff (somehow was wondering if it was `int16_t` or `int32_t`).

Issue #1760, opened 2025-08-27
Multiple Screens not working


ki-aura:
** Updated with requested info. 
Macos htop v 3.4.1 (installed via HomeBrew)
Ubuntu v 3.4.0 (lastest available using apt)

Tried on macOS (direct) and Ubuntu (VMware running on macOS) with same results
1. When adding a new screen, you can add new columns immediately, but if you try to go back and edit that screen definition, only the columns for the main screen show. 
**Added: When you go back to the screen definition in setup and select the new screen, the list of active columns doesn't change. If you add or remove a column, it modifies the main screen rather than the new screen. this happens even if you have 3 screens defined; regardless of which screen is selected, the list of active columns doesn't change and only the main screen gets modified

2. If you try to add EXE as a column it crashes on save   Error information:
------------------
A signal 6 (Aborted) was received
3966 abort (core dumped)  htop

Backtrace from macOS 
Backtrace information:
----------------------
0   htop                                0x0000000102af3b2c CRT_handleSIGSEGV + 288
1   libsystem_platform.dylib            0x000000018bf396a4 _sigtramp + 56
2   libsystem_pthread.dylib             0x000000018beff88c pthread_kill + 296
3   libsystem_c.dylib                   0x000000018be08a3c abort + 124
4   libsystem_malloc.dylib              0x000000018bd0c174 malloc_vreport + 892
5   libsystem_malloc.dylib              0x000000018bd0fc90 malloc_report + 64
6   libsystem_malloc.dylib              0x000000018bd1421c ___BUG_IN_CLIENT_OF_LIBMALLOC_POINTER_BEING_FREED_WAS_NOT_ALLOCATED + 32
7   htop                                0x0000000102af76e0 ListItem_delete + 24
8   htop                                0x0000000102b069d8 Vector_delete + 64
9   htop                                0x0000000102afb464 Panel_done + 32
10  htop                                0x0000000102afb144 Panel_delete + 20
11  htop                                0x0000000102b069d8 Vector_delete + 64
12  htop                                0x0000000102b00b98 ScreenManager_delete + 24
13  htop                                0x0000000102aef270 actionSetup + 92
14  htop                                0x0000000102af80b8 MainPanel_eventHandler + 276
15  htop                                0x0000000102b0138c ScreenManager_run + 1204
16  htop                                0x0000000102af1f0c CommandLine_run + 1872
17  dyld                                0x000000018bb5eb98 start + 6076

Backtrace from Ubuntu
[...]

PR #1783, opened 2025-10-09
Cancel pending renaming action for a screen in dtor of ScreensPanel

BenBE:
Fixes an UAF issue when editing the screen configuration.

Fixes: #1760

### DIFF
--- a/ScreensPanel.c
+++ b/ScreensPanel.c
@@ -51,6 +51,20 @@ static const char* const DynamicFunctions[] = {"      ", "Rename", "      ", "  
 
 static void ScreensPanel_delete(Object* object) {
    Panel* super = (Panel*) object;
+   ScreensPanel* const this = (ScreensPanel*) super;
+
+   /* cancel any pending edit action */
+   if (this->renamingItem) {
+      ListItem* item = (ListItem*) Panel_getSelected(super);
+      assert(item == this->renamingItem);
+
+      if (item)
+         item->value = this->saved;
+      this->renamingItem = NULL;
+      super->cursorOn = false;
+
+      Panel_setSelectionColor(super, PANEL_SELECTION_FOCUS);
+   }
 
    /* do not delete screen settings still in use */
    int n = Panel_size(super);


Explorer09:
I tested this in macOS Tahoe just now and it seems like the bug fix works.

One issue remaining, though. When built in GCC with LTO there is this warning:

```text
ScreensPanel.c:62:19: error: potential null pointer dereference [-Werror=null-dereference]
   62 |       item->value = this->saved;
      |                   ^
```

So the code has to tune a little bit:

```diff
diff --git a/ScreensPanel.c b/ScreensPanel.c
index eaf06920..e5ab1ded 100644
--- a/ScreensPanel.c
+++ b/ScreensPanel.c
@@ -57,10 +57,10 @@ static void ScreensPanel_delete(Object* object) {
    /* cancel any pending edit action */
    if (this->renamingItem) {
       ListItem* item = (ListItem*) Panel_getSelected(super);
-      assert(item);
       assert(item == this->renamingItem);
 
-      item->value = this->saved;
+      if (item)
+         item->value = this->saved;
       this->renamingItem = NULL;
       super->cursorOn = false;
```

And that's it.

BenBE:
@Explorer09 Do you have an idea for a fix to the issue you mentioned in #1848? I think the fixes will likely be related thus I'd like to handle it in one go.

Explorer09:
> @Explorer09 Do you have an idea for a fix to the issue you mentioned in #1848? I think the fixes will likely be related thus I'd like to handle it in one go.

Nope. Not yet. I'm busy with my personal work these days and won't be able to look at the problem.

Explorer09:
> > @Explorer09 Do you have an idea for a fix to the issue you mentioned in #1848? I think the fixes will likely be related thus I'd like to handle it in one go.
> 

@BenBE 

I just managed to get some free time and diagnosed the problem with GDB, and I have a partial fix:

```diff
diff --git a/ScreensPanel.c b/ScreensPanel.c
index 2ba7b940..acda42ad 100644
--- a/ScreensPanel.c
+++ b/ScreensPanel.c
@@ -77,6 +77,13 @@ static HandlerResult ScreensPanel_eventHandlerRenaming(Panel* super, int ch) {
    }
 
    switch (ch) {
+      case EVENT_SET_SELECTED: {
+         ListItem* item = (ListItem*) Panel_getSelected(super);
+         if (item != this->renamingItem)
+            goto renameFinish;
+         break;
+      }
+
       case 127:
       case KEY_BACKSPACE:
          if (this->cursor > 0) {
@@ -93,8 +100,9 @@ static HandlerResult ScreensPanel_eventHandlerRenaming(Panel* super, int ch) {
          if (!item)
             break;
          assert(item == this->renamingItem);
+renameFinish:
          free(this->saved);
-         item->value = xStrdup(this->buffer);
+         this->renamingItem->value = xStrdup(this->buffer);
          this->renamingItem = NULL;
          super->cursorOn = false;
          Panel_setSelectionColor(super, PANEL_SELECTION_FOCUS);
```

This at the minimum, avoids the crash when combined with your commit dbdad2069f87b9dacc5d6192410cb13ea804cfb6. The issue though, is that the modified screen name might not get applied when the ScreensPanel itself loses focus.

In `ScreenManager_run`, it looks like the `Panel_eventHandler` function is fired only for `panelFocus`, so I didn't yet find a way to catch the event when the `ScreensPanel` loses focus.

Explorer09:
Wait! I thought this fix is no longer needed as #1928 is merged to `main`. Did I miss something?
@fasterit Can you please look at this?

BenBE:
Yes and no. In theory #1928 should avoid all instances where this issue could be triggered as an UAF, while this PR explicitly closes the UAF itself (basically DiD). So merging both is correct (although re-using the infrastructure from #1928 in the dtor could be seen as an optimization if it preserves invariants).

PR #1784, opened 2025-10-09
Use libc qsort() for Vector sorting

Explorer09:
I don't know if I need to bother fix this one. Since we could replace the whole "quickSort" algorithm with srandard C `qsort()`. This `quickSort` function can be trace to this very old commit:  7ca10817122d3b7b30fabb1cadb75e5ee14b364e. I have no idea why @hishamhm chose a custom "quickSort" algorithm rather than a standard C `qsort()` function.

### DIFF
--- a/Vector.c
+++ b/Vector.c
@@ -16,6 +16,8 @@ in the source distribution for its full text.
 #include "XUtils.h"
 
 
+static Object_Compare vectorCompareFn;
+
 Vector* Vector_new(const ObjectClass* type, bool owner, int size) {
    Vector* this;
 
@@ -95,6 +97,7 @@ void Vector_prune(Vector* this) {
 
 //static int comparisons = 0;
 
+/*
 static void swap(Object** array, int indexA, int indexB) {
    assert(indexA >= 0);
    assert(indexB >= 0);
@@ -102,31 +105,7 @@ static void swap(Object** array, int indexA, int indexB) {
    array[indexA] = array[indexB];
    array[indexB] = tmp;
 }
-
-static int partition(Object** array, int left, int right, int pivotIndex, Object_Compare compare) {
-   const Object* pivotValue = array[pivotIndex];
-   swap(array, pivotIndex, right);
-   int storeIndex = left;
-   for (int i = left; i < right; i++) {
-      //comparisons++;
-      if (compare(array[i], pivotValue) <= 0) {
-         swap(array, i, storeIndex);
-         storeIndex++;
-      }
-   }
-   swap(array, storeIndex, right);
-   return storeIndex;
-}
-
-static void quickSort(Object** array, int left, int right, Object_Compare compare) {
-   if (left >= right)
-      return;
-
-   int pivotIndex = (left + right) / 2;
-   int pivotNewIndex = partition(array, left, right, pivotIndex, compare);
-   quickSort(array, left, pivotNewIndex - 1, compare);
-   quickSort(array, pivotNewIndex + 1, right, compare);
-}
+*/
 
 // If I were to use only one sorting algorithm for both cases, it would probably be this one:
 /*
@@ -167,10 +146,15 @@ static void insertionSort(Object** array, int left, int right, Object_Compare co
    }
 }
 
+static int Vector_compareObjects(const void* v1, const void* v2) {
+   return vectorCompareFn(*(const Object* const*)v1, *(const Object* const*)v2);
+}
+
 void Vector_quickSortCustomCompare(Vector* this, Object_Compare compare) {
    assert(compare);
    assert(Vector_isConsistent(this));
-   quickSort(this->array, 0, this->items - 1, compare);
+   vectorCompareFn = compare;
+   qsort(this->array, this->items, sizeof(Object*), Vector_compareObjects);
    assert(Vector_isConsistent(this));
 }
 


natoscott:
+1 for using qsort.

hishamhm:
> I have no idea why @hishamhm chose a custom "quickSort" algorithm rather than a standard C qsort() function.

The answer is much simpler and mundane than you can imagine: I was still learning C. :-) In fact, besides "scratching my itch for a better top", practicing C was the other motivation that 20+-years-younger me had for starting this project.

That also explains other oddities from the early codebase, such as trying to avoid maintaining .h files by hand (which I found nonsensical), the Java-esque OOP collections, and my favorite, which I think I never told to anyone before: the utility module with functions for setting up drawing to the screen, "CRT", is named so as a nod to the one in Turbo Pascal. :-D

Most importantly, I think all of those rookie oddities show how you don't have to be an experienced programmer to make a significant contribution to FOSS!

Explorer09:
@hishamhm Well. It's not a bad thing for contributing to OSS for trying to learn C. The question that interests me is: You seems to be capable of using `Object_Compare` function type but not realising the `size_t` data type when you wrote a sort function.

It seems there are no objections for ditching this custom `quickSort` for standard C `qsort`. I can revise this pull request to make the switch. This gives me one less `int` type issue to worry about when I try to migrate the `Vector` size data types to `size_t`.

hishamhm:
@Explorer09 I don't recall the details; maybe I just wasn't _aware_ of the `qsort` function, who knows! And for sure I wasn't super consistent with my handling of integer types. Back then I was most certainly running Linux i686, so it was the era of 32-bit pointers, and using `int` everywhere "just worked".

Explorer09:
@hishamhm In 32-bit systems `int` type has the same size as `size_t` and so many people weren't aware of the differences. But as we are now in 64 bits the `int` vs. `size_t` differences becomes significant and this is why the recent set of pull requests to eventually address that. 🙂

Also note that `(left + right) / 2` was a common mistake for people who implement _binary search_ themselves (not technically necessary for _quicksort_, but I don't blame them of choosing `(left + right) / 2` as a pivot). `left + right` can overflow for large arrays and so the proper way should be `left + ((right - left) / 2)`.
The other potential problem of a custom quicksort is not addressing the worst case (O(n²)) properly:
* When sorting sub-arrays after partitioning, the sub-array of the _smaller_ size should get sorted first. That way the compiler can optimize tail calls when sorting the larger sub-array and minimize the recursion depth.
* Standard C library's `qsort` could work better by implementing Introsort rather than a naïve Quicksort. You know, Introsort switches to Heapsort after a certain recursion depth is reached, thus it keeps the worst case time complexity down.

hishamhm:
Another sign of the times, now that talking about it reminds me of the old days, was that when I started htop, pid numbers were much smaller. Back in the day, it was IIRC a signed 16-bit value. That means I originally didn't expect any Vectors to be larger than 32768, so overflow using 32-bit indices was not a practical concern! Fun trip down memory lane.

Of course, later Vectors were used for more things, and pids (and systems!) got larger. But back when I first implemented that Vector it was unthinkable to me that its 32-bit indices could overflow: I don't think my machine had enough RAM for that. :)

Vector.c
```diff
@@ -95,38 +97,15 @@ void Vector_prune(Vector* this) {
 
 //static int comparisons = 0;
 
+/*
```
BenBE, line 100: Don't keep commented-out code around. That's what a VCS is for …

Vector.c
```diff
@@ -167,10 +146,15 @@ static void insertionSort(Object** array, int left, int right, Object_Compare co
...
+static int Vector_compareObjects(const void* v1, const void* v2) {
+   return vectorCompareFn(*(const Object* const*)v1, *(const Object* const*)v2);
+}
+
 void Vector_quickSortCustomCompare(Vector* this, Object_Compare compare) {
    assert(compare);
    assert(Vector_isConsistent(this));
-   quickSort(this->array, 0, this->items - 1, compare);
+   vectorCompareFn = compare;
+   qsort(this->array, this->items, sizeof(Object*), Vector_compareObjects);
```
BenBE, line 157: If it weren't for `qsort_r` only being standardized in POSIX.2024, I'd opt for using it; avoiding the need for global variables. Speaking of which: the wrapper/adapter should check the `vectorCompareFn` for `NULL` (marked unlikely to the optimizer).

FWIW: If we wanna be fancy, we could detect a compatible `qsort_r` implementation in `configure.ac` and prefer that if available; falling back to `qsort` only if stars don't align.

Explorer09: I wish to keep the changes small and leave the `qsort_r` support to a separate commit.

POSIX standardised the GNU version of `qsort_r` but before GNU, FreeBSD had a version of `qsort_r` that was ABI-incompatible. Although checking for `qsort_r` is possible in `configure`, it's not trivial code. Given that we still need to support platforms without `qsort_r`, I consider this `qsort_r` thing a lower priority.

Vector.c
Explorer09: Then I would have to delete the commented-out `combSort` code as well...
Would @hishamhm comment on this?

hishamhm: I'm happy to provide historical commentary, but I defer to the current maintainers on all things code review!

Vector.c
BenBE: Same here. Exactly because `qsort_r` support is such a mess I'd suggested to skip it for now and revisit it later..

Which leaves the one NULL check I mentioned above as an item TBD …

Explorer09: @BenBE Yes. I plan to add an `assert(!vectorCompareFn)` before assignment in order to catch whether `Vector_*Sort` is called in a signal handler or multithreaded code by mistake.

Vector.c
BenBE: The `combSort` vs. `insertionSort` place is a different part of the code.

But as sorting has currently two implementations, this is some other cleanup task (#1785) anyway …

Vector.c
BenBE: That's not what I was referring to. And also won't actually help except as a canary.

BenBE:
I pushed the relevant (short-term) fix from this PR in ee1d1a840d035aa0e6c0abfdbc3f699e2ee0f12f.

The question whether we want to update the actual sorting implementation and replace it by something more sprakling should be discussed while known bugs in existing code are fixed. I.e. focus on fixing the original issue, improve the algorithm in a second step.

@Explorer09 Please try to base your algorithm replacement work onto d3cd557e. That way we can later compare the different sort implementations.

Explorer09:
I closed this in favor of #1798.

PR #1790, opened 2025-10-22
build: Update ax_check_compile_flag module

Explorer09:
Update `AX_CHECK_COMPILE_FLAG` in `configure.ac` to latest upstream version (serial 11).
Move the macro to a separate file so that it won't be subject to our coding style such as in 97073ba86e2627db9132217f19f8a3c4f2235f68, and it allows future update easier.

The new version automatically appends `-Werror` during the compile option check, so I removed the `-Werror` from our macro calls in order to make them brief.


.gitignore
```diff
@@ -41,7 +41,8 @@ pcp-htop.5
 install-sh
 libtool
 ltmain.sh
-m4/
+m4/*
+!m4/ax_check_compile_flag.m4
```
BenBE: No need to add this rule; once a file is added, it gets tracked by git.

Explorer09: Acknowledged.

PR #1791, opened 2025-10-23
Fix libunwind detection regression & detect libunwind.pc

Explorer09:
1. Fix libunwind detection regression from #1779 (found during the discussion [here](https://github.com/htop-dev/htop/pull/1270#issuecomment-3435416010)).
2. Detect libunwind.pc (pkg-config module) which is better at telling libunwind dependencies. The manual check of `-llzma` is kept as fallback.

Note to @BenBE: I think this is a _less_ PITA comparing to the checking code of curses & ncurses. 🙂

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -963,9 +963,18 @@ case "$enable_unwind" in
    no)
       ;;
    check|yes)
-      if test "$enable_static" = yes; then
-         AC_CHECK_LIB([lzma], [lzma_index_buffer_decode])
-      fi
+      HTOP_PKG_CHECK_MODULES(LIBUNWIND, libunwind,
+         [],
+         [
+            if test "$enable_static" = yes; then
+               AC_CHECK_LIB([lzma], [lzma_index_buffer_decode])
+            fi
+            : "${LIBUNWIND_LIBS='-lunwind'}"
+         ]
+      )
+
+      htop_save_CFLAGS=$CFLAGS
+      CFLAGS="$AM_CFLAGS $LIBUNWIND_CFLAGS $CFLAGS"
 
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -896,6 +896,27 @@ if test "x$enable_affinity" = xyes; then
 fi
 
 
+dnl HTOP_PKG_CHECK_MODULES(VARIABLE-PREFIX, MODULES [, ACTION-IF-FOUND
+dnl    [, ACTION-IF-NOT-FOUND]])
```
BenBE:
```suggestion
dnl HTOP_PKG_CHECK_MODULES(VARIABLE-PREFIX, MODULES [, ACTION-IF-FOUND [, ACTION-IF-NOT-FOUND]])
```

configure.ac
```diff
@@ -896,6 +896,27 @@ if test "x$enable_affinity" = xyes; then
...
+   [m4_ifdef(
+      [PKG_PROG_PKG_CONFIG],
+      [PKG_CHECK_MODULES([$1], [$2], [$3], [$4])],
+      [m4_default(
+         [$4],
+         [AC_MSG_ERROR([pkg.m4 required to check for pkg-config modules '$2'; please regenerate configure script])]
+      )]
+   )]
+)dnl HTOP_PKG_CHECK_MODULES
+dnl
```
BenBE:
```suggestion
)

```

I think the `dnl` comments aren't strictly required

MrCirdo:
Hello @Explorer09,

I apply your commits on my branch and it works! It solves the problem I had yesterday about finding the `libunwind`

Issue #1785, opened 2025-10-12
Consolidate functions for sorting vectors


BenBE:
Currently there's `Vector_quickSort`, `Vector_quickSortCustomCompare`, and `Vector_insertionSort` for sorting vectors.

Apply the Highlander rule and let there only be one: `Vector_sort`.

While at it, update the signatures to allow for context support in the comparison functions:
```
typedef int (*Object_SortFn)(const void *a, const void *b, void *ctx);
int Vector_sort(Vector *v, Object_SortFn cmp, void *ctx);
```

Note: The single replacement API MUST be a stable sorting algorithm internally to conserve behaviour of `Vector_insertionSort`.

Explorer09:
I think `Vector_insertionSort` algorithm has its merits and I oppose removing it for now. (1) Insertion sort is stable; (2) insertion sort works better for arrays that are 80% sorted.

BenBE:
There's [Timsort](https://en.wikipedia.org/wiki/Timsort) and it's derived version [Powersort](https://en.wikipedia.org/wiki/Powersort) that are a combination of stable [Merge sort](https://en.wikipedia.org/wiki/Merge_sort)  with basically insertion sort for small sub-arrays.

So no, insertion sort won't be gone entirely, but it will be only a detail in the actual implementation using something more advanced. Thus, I'm fully aware that insertion sort and quick sort have their reason for existence in the current implementation, but it's for a reason, why several languages (Java, Python) switched to something else in their standard library when it comes to sorting.

Explorer09:
@BenBE I was considering leaving two interfaces for sorting, like C++ `std::sort` and `std::stable_sort`, and omit the implementation details in the function names.

As for what algorithm will be for `stable_sort`, my idea was a kind of in-place merge sort that does not take O(n) extra memory (it would take at most O(log(n)) stack space for recursion, though).

BenBE:
I'm currently looking into PowerSort, which is both stable AND a variant of merge sort. Also, it should be reasonable to implement its merging step in-place, leaving us with static worst case memory usage of O(log(n)).

Explorer09:
> I'm currently looking into PowerSort, which is both stable AND a variant of merge sort. Also, it should be reasonable to implement its merging step in-place, leaving us with static worst case memory usage of O(log(n)).

[...]

PR #1792, opened 2025-10-23
perf-enhancement: Add Vector_sort function w/ context support

j-a-c-k-goes:
_[PR_DESCRIPTION.md](https://github.com/user-attachments/files/23108975/PR_DESCRIPTION.md)_

### Add Vector_sort function with context support

Consolidates vector sorting functions as requested in issue #1785.
Implements "Highlander rule" via providing single preferred
sorting function while maintaining backward compatibility.

**Features:**
- Context-aware comparison functions for advanced sorting logic
- Hybrid algorithm: insertion sort ≤16 elements, quicksort >16 elements  
- Backward compatible: existing functions remain unchanged
- NULL-safe context handling

**Performance benefits:**
- Optimal algorithm selection based on array size
- Up to 93% faster than pure insertion sort on large arrays
- 16-element threshold for algorithm switching

This addresses issue #1785 without breaking existing code.
Future commits can gradually migrate call sites to use `Vector_sort`.

Signed-off-by: [_ack] <[jack@flnpb.comy]>

### DIFF
--- a/Vector.c
+++ b/Vector.c
@@ -167,10 +167,61 @@ static void insertionSort(Object** array, int left, int right, Object_Compare co
    }
 }
 
+static void insertionSortWithContext(Object** array, int left, int right, Object_CompareWithContext compare, void* context) {
+   for (int i = left + 1; i <= right; i++) {
+      Object* t = array[i];
+      int j = i - 1;
+      while (j >= left) {
+         if (compare(array[j], t, context) <= 0)
+            break;
+         array[j + 1] = array[j];
+         j--;
+      }
+      array[j + 1] = t;
+   }
+}
+
+static int partitionWithContext(Object** array, int left, int right, int pivotIndex, Object_CompareWithContext compare, void* context) {
+   const Object* pivotValue = array[pivotIndex];
+   swap(array, pivotIndex, right);
+   int storeIndex = left;
+   for (int i = left; i < right; i++) {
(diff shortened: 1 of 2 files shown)


Explorer09:
Hi.

Good contribution, but I have a concern on using quicksort, which can break stability in sorting. The original code uses insertionSort probably because it's a stable sort, or because it deals with data that are more often partially sorted.

So when it comes to consolidating sorting function, I would consider any unstable sorting function be out.

j-a-c-k-goes:
@Explorer09 are you finding the implemented sort to be unstable? Could retest, but initial tests were not unstable.

Explorer09:
> are you finding the implemented sort to be unstable? Could retest, but initial tests were not unstable.

@j-a-c-k-goes 
It's computer science. A sorting algorithm is said to be "stable" when the items that are compared to be equal keep their relative order after sorted. This is important for data that can be sorted in multiple, different keys (think of a data table). When I sort the table under key A, then reorder the items under key B. I except items that are equal under B keep their relative order of A.

BenBE:
This PR should be based on the commit https://github.com/htop-dev/htop/commit/d3cd557e0534a45ab971397d4c407116547d3f6d that I mentioned in [#1784](https://github.com/htop-dev/htop/pull/1784#issuecomment-3397158874) to allow a proper comparison for the chosen implementations.

PR #1794, opened 2025-10-25
Implement different display types for bar meters

### LINKED ISSUE #647: Custom character for meter (instead of "|")

rustedusted:
Allow bar meters to enable "sub-pixel" rendering.

Bar meter styles can be changed in setup (F2) under Display Options.

Co-Authored-By: Benny Baumann <BenBE@geshi.org>

### DIFF
--- a/DisplayOptionsPanel.c
+++ b/DisplayOptionsPanel.c
@@ -205,7 +205,7 @@ DisplayOptionsPanel* DisplayOptionsPanel_new(Settings* settings, ScreenManager* 
    Panel_add(super, (Object*) NumberItem_newByRef("Hide main function bar (0 - off, 1 - on ESC until next input, 2 - permanently)", &(settings->hideFunctionBar), 0, 0, 2));
 
    #ifdef HAVE_LIBNCURSESW
-   if(CRT_utf8)
+   if (CRT_utf8)
       Panel_add(super, (Object*) NumberItem_newByRef("Bar Type (0 - default)", (int*)&(settings->barType), 0, 0, BAR_METER_NUM_STYLES-1));
    #endif
 
--- a/DisplayOptionsPanel.c
+++ b/DisplayOptionsPanel.c
@@ -203,6 +203,12 @@ DisplayOptionsPanel* DisplayOptionsPanel_new(Settings* settings, ScreenManager* 
    Panel_add(super, (Object*) CheckItem_newByRef("Highlight new and old processes", &(settings->highlightChanges)));
    Panel_add(super, (Object*) NumberItem_newByRef("- Highlight time (in seconds)", &(settings->highlightDelaySecs), 0, 1, 24 * 60 * 60));
    Panel_add(super, (Object*) NumberItem_newByRef("Hide main function bar (0 - off, 1 - on ESC until next input, 2 - permanently)", &(settings->hideFunctionBar), 0, 0, 2));
+
+   #ifdef HAVE_LIBNCURSESW
+   if(CRT_utf8)
+      Panel_add(super, (Object*) NumberItem_newByRef("Bar Type (0 - default)", (int*)&(settings->barType), 0, 0, BAR_METER_NUM_STYLES-1));
+   #endif
+
    #ifdef HAVE_LIBHWLOC
    Panel_add(super, (Object*) CheckItem_newByRef("Show topology when selecting affinity by default", &(settings->topologyAffinity)));
    #endif
--- a/Meter.c
+++ b/Meter.c
@@ -86,6 +86,18 @@ static void TextMeterMode_draw(Meter* this, int x, int y, int w) {
 
 static const char BarMeterMode_characters[] = "|#*@$%&.";
 
+#ifdef HAVE_LIBNCURSESW
+const wchar_t* bars[BAR_METER_NUM_STYLES] = {
+   L"|",
+   L"#",
+   L"⣿⡀⡄⡆⡇⣇⣧⣷",
+   L"█░░▒▒▓▓█",
+   L"█▏▎▍▌▋▊▉",
+   L"█▁▂▃▄▅▆▇",
+   L"█▌▌▌▌███"
+};
+#endif
+
 static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
    assert(x >= 0);
    assert(w <= INT_MAX - x);
@@ -149,6 +161,14 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
    assert(startPos + w <= RichString_sizeVal(bar));
 
    int blockSizes[10];
+#ifdef HAVE_LIBNCURSESW
+   Settings* settings = this->host->settings;
+   assert(settings->barType < ( sizeof(bars) / sizeof(wchar_t*) ));
+   assert(settings->barType >= 0);
+   const wchar_t* currBar = bars[settings->barType];
+   int barLen = (int)wcslen(currBar);
+   int extraWidth = 0;
+#endif
 
    // First draw in the bar[] buffer...
    int offset = 0;
@@ -158,19 +178,35 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
          value = MINIMUM(value, this->total);
          blockSizes[i] = ceil((value / this->total) * w);
          blockSizes[i] = MINIMUM(blockSizes[i], w - offset);
+
+#ifdef HAVE_LIBNCURSESW
+         extraWidth = (int)ceil((value / this->total) * w * barLen) % barLen;
+#endif
       } else {
          blockSizes[i] = 0;
       }
       int nextOffset = offset + blockSizes[i];
-      for (int j = offset; j < nextOffset; j++)
+      for (int j = offset; j < nextOffset; j++) {
          if (RichString_getCharVal(bar, startPos + j) == ' ') {
             if (CRT_colorScheme == COLORSCHEME_MONOCHROME) {
                assert(i < strlen(BarMeterMode_characters));
                RichString_setChar(&bar, startPos + j, BarMeterMode_characters[i]);
-            } else {
+            }
+#ifdef HAVE_LIBNCURSESW
+            else if(CRT_utf8 && settings->barType) {
+               if(j==nextOffset-1){
+                  RichString_setChar(&bar, startPos+nextOffset-1,  currBar[extraWidth]);
+               } else {
+                  RichString_setChar(&bar, startPos + j, currBar[0]);
+               }
+            }
+#endif
+            else {
                RichString_setChar(&bar, startPos + j, '|');
             }
          }
+      }
+
       offset = nextOffset;
    }
 
--- a/Meter.c
+++ b/Meter.c
@@ -194,7 +194,7 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
             }
 #ifdef HAVE_LIBNCURSESW
             else if (CRT_utf8 && settings->barType) {
-               if(j==nextOffset-1){
+               if (j == nextOffset - 1) {
                   RichString_setChar(&bar, startPos+nextOffset-1,  currBar[extraWidth]);
                } else {
                   RichString_setChar(&bar, startPos + j, currBar[0]);
--- a/Meter.c
+++ b/Meter.c
@@ -193,7 +193,7 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
                RichString_setChar(&bar, startPos + j, BarMeterMode_characters[i]);
             }
 #ifdef HAVE_LIBNCURSESW
-            else if(CRT_utf8 && settings->barType) {
+            else if (CRT_utf8 && settings->barType) {
                if(j==nextOffset-1){
                   RichString_setChar(&bar, startPos+nextOffset-1,  currBar[extraWidth]);
                } else {
--- a/Meter.c
+++ b/Meter.c
@@ -195,7 +195,7 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
 #ifdef HAVE_LIBNCURSESW
             else if (CRT_utf8 && settings->barType) {
                if (j == nextOffset - 1) {
-                  RichString_setChar(&bar, startPos+nextOffset-1,  currBar[extraWidth]);
+                  RichString_setChar(&bar, startPos + nextOffset - 1,  currBar[extraWidth]);
                } else {
                   RichString_setChar(&bar, startPos + j, currBar[0]);
                }
--- a/MeterMode.h
+++ b/MeterMode.h
@@ -26,4 +26,5 @@ typedef unsigned int MeterModeId;
    (1 << LED_METERMODE) |             \
    0) // Avoids edits when updating
 
+#define BAR_METER_NUM_STYLES 7
 #endif
--- a/Settings.c
+++ b/Settings.c
@@ -5,6 +5,7 @@ Released under the GNU GPLv2+, see the COPYING file
 in the source distribution for its full text.
 */
 
+#include "MeterMode.h"
 #include "config.h" // IWYU pragma: keep
 
 #include "Settings.h"
@@ -516,6 +517,13 @@ static bool Settings_read(Settings* this, const char* fileName, const Machine* h
          didReadMeters = true;
       } else if (String_eq(option[0], "hide_function_bar")) {
          this->hideFunctionBar = atoi(option[1]);
+      #ifdef HAVE_LIBNCURSESW
+      } else if (String_eq(option[0], "bar_type")) {
+         long value = strtol(option[1], NULL, 10);
+         if (value < 0 || value > BAR_METER_NUM_STYLES)
+            value = 0;
+         this->barType = (unsigned int)value;
+      #endif
       #ifdef HAVE_LIBHWLOC
       } else if (String_eq(option[0], "topology_affinity")) {
          this->topologyAffinity = !!atoi(option[1]);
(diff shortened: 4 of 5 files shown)


rustedusted:
referencing with issue #647 
the previous merge request was a mess

rustedusted:
One more thing
the bar type option is visible in F2 even when we do htop -U should i disable it when opened with -U 
or keep it there??

the function didn't have any if statement in it so felt awkward to do that
it's just a line of difference to add

rustedusted:
[x] Resolve the conflict marker in the latest commit(not using the previous commit)
[x] Handle storing the setting to disk
[ ] Proper guarding against invalid barType settings in Meter.c(i do not understand this, i just put asserts in the Meter.c line 168)(please let me know if there's anything else)
[x] Handle when htop is configured without Unicode support(done using header guards #ifdef HAVE_LIBNCURSESW)
[x] Update Commit message

Settings.c
```diff
@@ -821,6 +828,10 @@ Settings* Settings_new(const Machine* host, Hashtable* dynamicMeters, Hashtable*
    this->showMergedCommand = false;
    this->hideFunctionBar = 0;
    this->headerMargin = true;
+
+   #ifdef HAVE_LIBNCURSESW
+      this->barType = 0;
+   #endif
```
BenBE, line 838:
```suggestion

   #ifdef HAVE_LIBNCURSESW
   this->barType = 0;
   #endif

```

Settings.c
```diff
@@ -516,6 +516,10 @@ static bool Settings_read(Settings* this, const char* fileName, const Machine* h
          didReadMeters = true;
       } else if (String_eq(option[0], "hide_function_bar")) {
          this->hideFunctionBar = atoi(option[1]);
+      #ifdef HAVE_LIBNCURSESW
+      } else if (String_eq(option[0], "bar_type")) {
+         this->barType = atoi(option[1]);
```
BenBE: The value read here isn't protected against an invalid configuration file. Please ensure, that the value in `this->barType` is sanitized to be a valid bar type index.

Meter.c
```diff
@@ -158,19 +180,36 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
...
+               RichString_setChar(&bar, startPos + j, currBar[0]);
+            }
+#endif
+            else {
                RichString_setChar(&bar, startPos + j, '|');
             }
          }
+      }
+#ifdef HAVE_LIBNCURSESW
+      if(CRT_utf8 && offset != nextOffset && CRT_colorScheme != COLORSCHEME_MONOCHROME){
```
BenBE:
```suggestion
      if (CRT_utf8 && offset != nextOffset && CRT_colorScheme != COLORSCHEME_MONOCHROME) {
```

Meter.c
```diff
@@ -158,19 +180,36 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
...
+         RichString_setChar(&bar, startPos+nextOffset-1,  currBar[extraWidth]);
```
BenBE:
```suggestion
         RichString_setChar(&bar, startPos + nextOffset - 1,  currBar[extraWidth]);
```

Meter.c
```diff
@@ -158,19 +180,36 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
          value = MINIMUM(value, this->total);
          blockSizes[i] = ceil((value / this->total) * w);
          blockSizes[i] = MINIMUM(blockSizes[i], w - offset);
+
+#ifdef HAVE_LIBNCURSESW
+         extraWidth = (int)ceil((value / this->total) * w * barLen) % barLen;
+#endif
```
BenBE, line 184: I think this needs some minor update. The `extraWidth` calculation probably should re-use the value calculated in the `blockSizes[i]` assignments above.

Meter.c
```diff
@@ -86,6 +88,18 @@ static void TextMeterMode_draw(Meter* this, int x, int y, int w) {
...
 
+#ifdef HAVE_LIBNCURSESW
+static const wchar_t* bars[8] = {
+   L"||||||||",
+   L"########",
+   L"⣿⡀⡄⡆⡇⣇⣧⣷",
+   L"█░░▒▒▓▓█",
+   L"█▏▎▍▌▋▊▉",
+   L"█▁▂▃▄▅▆▇",
+   L"█▌▌▌▌███",
```
BenBE: Having the full block final seems more natural. Any reason to put it first and skip the per-profile blank instead?

DisplayOptionsPanel.c
```diff
@@ -203,6 +203,11 @@ DisplayOptionsPanel* DisplayOptionsPanel_new(Settings* settings, ScreenManager*
    Panel_add(super, (Object*) CheckItem_newByRef("Highlight new and old processes", &(settings->highlightChanges)));
    Panel_add(super, (Object*) NumberItem_newByRef("- Highlight time (in seconds)", &(settings->highlightDelaySecs), 0, 1, 24 * 60 * 60));
    Panel_add(super, (Object*) NumberItem_newByRef("Hide main function bar (0 - off, 1 - on ESC until next input, 2 - permanently)", &(settings->hideFunctionBar), 0, 0, 2));
+
+   #ifdef HAVE_LIBNCURSESW
+   Panel_add(super, (Object*) NumberItem_newByRef("Bar Type (0-6)", &(settings->barType), 0, 0, 6));
```
BenBE: Maybe `0 = default` to avoid having this needing to updated whenever there's a new character theme.

Settings.h
```diff
@@ -106,6 +106,9 @@ typedef struct Settings_ {
    bool enableMouse;
    #endif
    int hideFunctionBar;  // 0 - off, 1 - on ESC until next input, 2 - permanently
+   #ifdef HAVE_LIBNCURSESW
+   int barType;
```
Explorer09: Any reason not to use `unsigned int` for this field?

Meter.c
```diff
@@ -149,6 +163,14 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
    assert(startPos + w <= RichString_sizeVal(bar));
 
    int blockSizes[10];
+#ifdef HAVE_LIBNCURSESW
+   Settings* settings = this->host->settings;
+   assert(settings->barType < ( sizeof(bars) / sizeof(wchar_t*) ));
+   assert(settings->barType >= 0);
+   const wchar_t* currBar = bars[settings->barType];
+   int barLen = (int)wcslen(currBar);
```
Explorer09, line 169: I consider this function call redundant when you have declared the list of `bars` to have equal length.

Not sure if this is what you intended, but I think the bar meter characters can be defined with variable widths if you wish.

Like this:

```c
static const wchar_t* BarMeterMode_drawStyles[] = {
   L"|",
   L"#",
   L"⡀⡄⡆⡇⣇⣧⣷⣿",
   L"░▒▓█",
   L"▏▎▍▌▋▊▉█",
   L"▁▂▃▄▅▆▇█",
   L"▌█",
};
```

Meter.c
```diff
@@ -13,8 +13,10 @@ in the source distribution for its full text.
 #include <float.h>
 #include <limits.h> // IWYU pragma: keep
 #include <math.h>
+#include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <wchar.h>
```
Explorer09: Remove this include line. `wchar.h` is better included indirectly via `ProvideCurses.h`.

Meter.c
```diff
@@ -13,8 +13,10 @@ in the source distribution for its full text.
 #include <float.h>
 #include <limits.h> // IWYU pragma: keep
 #include <math.h>
+#include <stdio.h>
```
Explorer09: I didn't see any `stdio.h` function being used at all in this PR.

Settings.c
Explorer09: It would be a good idea to set `this->barType` to 0 whenever we read an invalid value from the config. This makes the field forward-compatible to any future extensions of the list. Also, we should use `strtol` for parsing this field to get a defined behavior on integer overflow.

Unfortunately we don't have a macro to easily produce code like this one:

```c
{
   long value = strtol(option[1], NULL, 10);
   if (value < 0 || value > BAR_METER_NUM_STYLES)
      value = 0;
   this->barType = (unsigned int)value;
}
```

Meter.c
rustedusted: i'm doing modulo below so %wcslen
when the modulo returns 0 that means the extra width is full block 

basically if the value is 15%8 => 7 would mean that the last bar has to be the 7/8 thick where 7 is the index
when i go to 16%8 => 0 it means that the last block is 8/8 thick so while indexing i put the output of modulo

seemed much more cleaner to write 
should i change it changing the positioning here
might lead me to use ternary or even if else statements below

DisplayOptionsPanel.c
rustedusted: copy that

Settings.h
rustedusted: no other reason i just copied the overall implementation of option from hideFunctionBar
should i change it ?

BenBE:
Please squash/fixup changes into the original commit by rebasing. Details are in the style guide.

rustedusted:
here's the latest commit that contains all the suggested things

DisplayOptionsPanel.c
```diff
@@ -203,6 +203,12 @@ DisplayOptionsPanel* DisplayOptionsPanel_new(Settings* settings, ScreenManager*
    Panel_add(super, (Object*) CheckItem_newByRef("Highlight new and old processes", &(settings->highlightChanges)));
    Panel_add(super, (Object*) NumberItem_newByRef("- Highlight time (in seconds)", &(settings->highlightDelaySecs), 0, 1, 24 * 60 * 60));
    Panel_add(super, (Object*) NumberItem_newByRef("Hide main function bar (0 - off, 1 - on ESC until next input, 2 - permanently)", &(settings->hideFunctionBar), 0, 0, 2));
+
+   #ifdef HAVE_LIBNCURSESW
+   if(CRT_utf8)
```
BenBE:
```suggestion
   if (CRT_utf8)
```

Meter.c
BenBE: The bars currently have all the same length, but the intention is to actually allow for bars to contain different number of symbols.

Meter.c
```diff
@@ -158,19 +178,35 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
...
-      for (int j = offset; j < nextOffset; j++)
+      for (int j = offset; j < nextOffset; j++) {
          if (RichString_getCharVal(bar, startPos + j) == ' ') {
             if (CRT_colorScheme == COLORSCHEME_MONOCHROME) {
                assert(i < strlen(BarMeterMode_characters));
                RichString_setChar(&bar, startPos + j, BarMeterMode_characters[i]);
-            } else {
+            }
+#ifdef HAVE_LIBNCURSESW
+            else if(CRT_utf8 && settings->barType) {
```
BenBE:
```suggestion
            else if (CRT_utf8 && settings->barType) {
```

Meter.c
```diff
@@ -158,19 +178,35 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
...
+               if(j==nextOffset-1){
```
BenBE:
```suggestion
               if (j == nextOffset - 1) {
```

Meter.c
```diff
@@ -158,19 +178,35 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
...
+                  RichString_setChar(&bar, startPos+nextOffset-1,  currBar[extraWidth]);
```
BenBE:
```suggestion
                  RichString_setChar(&bar, startPos + nextOffset - 1,  currBar[extraWidth]);
```

MeterMode.h
```diff
@@ -26,4 +26,5 @@ typedef unsigned int MeterModeId;
    (1 << LED_METERMODE) |             \
    0) // Avoids edits when updating
 
+#define BAR_METER_NUM_STYLES 7
```
BenBE, line 29:
```suggestion
extern const size_t bar_meter_num_styles;

```

And below the definition of `bars' for the meter styles in `Meter.c`:
```c
const size_t bar_meter_num_styles = sizeof(bars) / sizeof(*bars);
```

This avoids an additional point of manual updating.

Explorer09: @BenBE This has the problem that when compiling Setting.c, compiler cannot inline the constant but has to refer to the external variable for the number. In other words, the compiled code would be sub-optimal.

I think the better approach is to move the definitions of the bar styles into the header. Look at the `unitPrefixes` definition in XUtils.h for what I mean.

Meter.c
Explorer09: The whole block of `else if (CRT_utf8 && settings->barType) { ... }` can merge with the `else { ... }` block below so we can reduce duplicate code.

Settings.h
```diff
@@ -106,6 +106,9 @@ typedef struct Settings_ {
    bool enableMouse;
    #endif
    int hideFunctionBar;  // 0 - off, 1 - on ESC until next input, 2 - permanently
+   #ifdef HAVE_LIBNCURSESW
+   unsigned int barType;
```
Explorer09, line 110: Since the proposed bar drawing styles has `|` (vertical bar, the default), and `#` (number sign), both can be drawn in ASCII-only mode, I think we can remove the `#ifdef HAVE_LIBNCURSESW` conditional and allow the setting to be available in both ASCII-only build and Unicode build of htop.

Note: only the first two drawing styles can be selected in ASCII. The rest can be guarded in the `#ifdef HAVE_LIBNCURSESW` so they require Unicode to enable:

```c
#ifdef HAVE_LIBNCURSESW
static const wchar_t* barMeterDrawStyles[] = {
   L"|",
   L"#",
   L"⡀⡄⡆⡇⣇⣧⣷⣿",
   L"░▒▓█",
   L"▏▎▍▌▋▊▉█",
   L"▁▂▃▄▅▆▇█",
   L"▌█",
};
#else
static const char* barMeterDrawStyles[] = {
   "|",
   "#"
};
#endif
static const size_t numBarMeterStyles = ARRAYSIZE(barMeterDrawStyles);
```

Meter.c
```diff
@@ -158,19 +178,35 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
...
+         extraWidth = (int)ceil((value / this->total) * w * barLen) % barLen;
+#endif
       } else {
          blockSizes[i] = 0;
       }
       int nextOffset = offset + blockSizes[i];
-      for (int j = offset; j < nextOffset; j++)
+      for (int j = offset; j < nextOffset; j++) {
          if (RichString_getCharVal(bar, startPos + j) == ' ') {
             if (CRT_colorScheme == COLORSCHEME_MONOCHROME) {
```
Explorer09, line 191: There is one thing I wish to add when the feature of changing bar drawing character is added. It's this:

```suggestion
            if (CRT_colorScheme == COLORSCHEME_MONOCHROME && Meter_maxItems(this) != 1) {
```

Make the bar character changeable for meters with only one item.

```diff
--- a/Meter.h
+++ b/Meter.h
@@ -100,6 +100,7 @@ typedef struct MeterClass_ {
 #define Meter_attributes(this_)        As_Meter(this_)->attributes
 #define Meter_name(this_)              As_Meter(this_)->name
 #define Meter_uiName(this_)            As_Meter(this_)->uiName
+#define Meter_maxItems(this_)          As_Meter(this_)->maxItems
 #define Meter_isMultiColumn(this_)     As_Meter(this_)->isMultiColumn
 #define Meter_isPercentChart(this_)    As_Meter(this_)->isPercentChart
 
```

Settings.h
rustedusted: currently i do not display the barType option in htop -U
as seen in DisplayOptionsPanel.c line 208

MeterMode.h
BenBE: I'm aware of this downside, but AFAICT there are no real places where looking up this constant happens inside a tight loop …

Meter.c
```diff
@@ -86,6 +86,18 @@ static void TextMeterMode_draw(Meter* this, int x, int y, int w) {
...
 
+#ifdef HAVE_LIBNCURSESW
+const wchar_t* bars[BAR_METER_NUM_STYLES] = {
+   L"|",
+   L"#",
+   L"⣿⡀⡄⡆⡇⣇⣧⣷",
+   L"█░░▒▒▓▓█",
+   L"█▏▎▍▌▋▊▉",
+   L"█▁▂▃▄▅▆▇",
+   L"█▌▌▌▌███"
```
ygoe, line 97: Does this mean that there can be fractional bar blocks, i.e. the bar doesn't fill an entire character but only a part of it, increasing horizontal resolution? I know these bars primarily from the CPU meter and they have sections of different colours. While that could work for the vertically sliced blocks (foreground is the block in the left colour, background is the colour of the following section), it probably can't work with the dotted style. And still, if a bar section is shorter than a full character, such a character would need more than 2 colours and foreground/background isn't enough to properly render it.

Personally, I'd already be happy if I could replace that thin vertical line with something more filling that would stretch from the left to the right edge of the character. Doesn't need to be the whole height like █ (U+2588), could also be ■ (U+25A0) or even ▬ (U+25AC).

Explorer09: The technical limitations have removed the possibility of showing two colors in a single block. (In ANSI color terminal, there's no "bright" colors for background, only the foreground can be set to a "bright" color).

PR #1795, opened 2025-10-26
build: Simplify curses header checking code

Explorer09:
Utilize the fact that AC_CHECK_HEADERS allows a break statement in the middle of the loop. No need to nest multiple AC_CHECK_HEADERS calls just to make a list of fallback headers.

(I'd give credit to the Autotools Mythbuster book, which first proposed this idea.)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -780,27 +780,58 @@ esac
 htop_save_CFLAGS=$CFLAGS
 CFLAGS="$AM_CFLAGS $CFLAGS"
 
+have_curses_header=no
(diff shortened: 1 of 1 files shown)


BenBE:
Do you have a reference to the Autotools Mythbuster book for reference of this technique?

Explorer09:
> Do you have a reference to the Autotools Mythbuster book for reference of this technique?

https://autotools.info/autoconf/finding.html

4.2.1 Checking for one out of Multiple Headers

PR #1798, opened 2025-10-31
New Vector_sort function, replacing insertion and quick sort

Explorer09:
Introduce the new `Vector_sort()` function and obsolete the old `Vector_quickSortCustomCompare()` and `Vector_insertionSort()` APIs.

This new sort function is a natural, in-place merge sort. I.e. it takes advantage of partially sorted data, and it's stable.

Space complexity: O(log(n)) worst case
Time complexity: O(n) best case, O(n*log(n)*log(n)) average & worst case

@BenBE , I have copied some of the changes from your d3cd557e0534a45ab971397d4c407116547d3f6d commit but I didn't base on it. Because there are some data type naming that I didn't like, and so I name my "context" structure in my own way. Also to keep changes small I didn't upgrade the `Object_Compare` function type to take a new "context" parameter.

### DIFF
--- a/Vector.c
+++ b/Vector.c
@@ -10,11 +10,18 @@ in the source distribution for its full text.
 #include "Vector.h"
 
 #include <assert.h>
+#include <stdint.h>
 #include <stdlib.h>
 #include <string.h>
 
 #include "XUtils.h"
+#include "generic/Sort.h"
 
+
+typedef struct VectorSortContext_ {
+   Object_Compare compare;
+   void* compareContext;
+} VectorSortContext;
 
 Vector* Vector_new(const ObjectClass* type, bool owner, int size) {
    Vector* this;
@@ -93,97 +100,24 @@ void Vector_prune(Vector* this) {
    memset(this->array, '\0', this->arraySize * sizeof(Object*));
 }
 
-//static int comparisons = 0;
+ATTR_NONNULL
+static int Vector_sortCompare(const void* p1, const void* p2, void* context) {
+   VectorSortContext* vc = (VectorSortContext*) context;
 
-static void swap(Object** array, int indexA, int indexB) {
-   assert(indexA >= 0);
-   assert(indexB >= 0);
-   Object* tmp = array[indexA];
-   array[indexA] = array[indexB];
-   array[indexB] = tmp;
+   return vc->compare(*(const void* const*)p1, *(const void* const*)p2, vc->compareContext);
 }
 
-static int partition(Object** array, int left, int right, int pivotIndex, Object_Compare compare, void* context) {
-   const Object* pivotValue = array[pivotIndex];
-   swap(array, pivotIndex, right);
-   int storeIndex = left;
-   for (int i = left; i < right; i++) {
-      //comparisons++;
-      if (compare(array[i], pivotValue, context) <= 0) {
-         swap(array, i, storeIndex);
-         storeIndex++;
-      }
-   }
-   swap(array, storeIndex, right);
-   return storeIndex;
-}
-
-static void quickSort(Object** array, int left, int right, Object_Compare compare, void* context) {
-   if (left >= right)
-      return;
-
-   int pivotIndex = left + (right - left) / 2;
-   int pivotNewIndex = partition(array, left, right, pivotIndex, compare, context);
-   quickSort(array, left, pivotNewIndex - 1, compare, context);
-   quickSort(array, pivotNewIndex + 1, right, compare, context);
-}
-
-// If I were to use only one sorting algorithm for both cases, it would probably be this one:
-/*
-
-static void combSort(Object** array, int left, int right, Object_Compare compare, void* context) {
-   int gap = right - left;
-   bool swapped = true;
-   while ((gap > 1) || swapped) {
-      if (gap > 1) {
-         gap = (int)((double)gap / 1.247330950103979);
-      }
-      swapped = false;
-      for (int i = left; gap + i <= right; i++) {
-         comparisons++;
-         if (compare(array[i], array[i+gap], context) > 0) {
-            swap(array, i, i+gap);
-            swapped = true;
-         }
-      }
-   }
-}
-
-*/
-
-static void insertionSort(Object** array, int left, int right, Object_Compare compare, void* context) {
-   for (int i = left + 1; i <= right; i++) {
-      Object* t = array[i];
-      int j = i - 1;
-      while (j >= left) {
-         //comparisons++;
-         if (compare(array[j], t, context) <= 0)
-            break;
-
-         array[j + 1] = array[j];
-         j--;
-      }
-      array[j + 1] = t;
-   }
-}
-
-void Vector_quickSort(Vector* this, Object_Compare compare, void* context) {
-   if (!compare) {
-      assert(this->type->compare);
-      compare = this->type->compare;
-   }
+ATTR_NONNULL_N(1)
+void Vector_sort(Vector* this, Object_Compare compare, void* context) {
+   VectorSortContext vc = {
+      .compare = compare ? compare : this->type->compare,
+      .compareContext = context ? context : this,
+   };
+   assert(vc.compare);
    assert(Vector_isConsistent(this));
-   quickSort(this->array, 0, this->items - 1, compare, context);
-   assert(Vector_isConsistent(this));
-}
 
-void Vector_insertionSort(Vector* this, Object_Compare compare, void* context) {
-   if (!compare) {
-      assert(this->type->compare);
-      compare = this->type->compare;
-   }
-   assert(Vector_isConsistent(this));
-   insertionSort(this->array, 0, this->items - 1, compare, context);
+   Sort_sort(this->array, this->items, sizeof(*this->array), Vector_sortCompare, &vc);
+
    assert(Vector_isConsistent(this));
 }
 
--- a/Vector.c
+++ b/Vector.c
@@ -103,13 +103,13 @@ static void swap(Object** array, int indexA, int indexB) {
    array[indexB] = tmp;
 }
 
-static int partition(Object** array, int left, int right, int pivotIndex, Object_Compare compare) {
+static int partition(Object** array, int left, int right, int pivotIndex, Object_Compare compare, void* context) {
    const Object* pivotValue = array[pivotIndex];
    swap(array, pivotIndex, right);
    int storeIndex = left;
    for (int i = left; i < right; i++) {
       //comparisons++;
-      if (compare(array[i], pivotValue) <= 0) {
+      if (compare(array[i], pivotValue, context) <= 0) {
          swap(array, i, storeIndex);
          storeIndex++;
       }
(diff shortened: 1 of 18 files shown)


Vector.c
```diff
@@ -93,91 +101,168 @@ void Vector_prune(Vector* this) {
...
-   const Object* pivotValue = array[pivotIndex];
-   swap(array, pivotIndex, right);
-   int storeIndex = left;
-   for (int i = left; i < right; i++) {
-      //comparisons++;
-      if (compare(array[i], pivotValue) <= 0) {
-         swap(array, i, storeIndex);
-         storeIndex++;
+ATTR_NONNULL
+static void rotate(void* buffer, size_t leftSize, size_t rightSize) {
```
Explorer09: A note to leave to myself:
There are two possible interfaces for this `rotate` function. To minimize the chances that the callers pass  invalid values, I picked the sizes as the second and third parameters. When it comes to optimization, this would be more dependent on the compiler's capability in simplifying additions and subtraction than the version that passes `mid` and `end` pointers. I didn't check the code sizes produced yet, so this remains a TODO for me.

```c
rotate_a(void* buffer, size_t leftSize, size_t rightSize);
rotate_b(void* start, void* mid, const void* end);
```

Vector.c
```diff
@@ -93,91 +100,167 @@ void Vector_prune(Vector* this) {
...
+         p2 += size;
       }
    }
-}
 
-*/
+   rotate(p1, (size_t)(mid - p1), (size_t)(p2 - mid));
+
+   leftLen = (size_t)(p1 - (char*)array) / size;
+   rightLen = (size_t)(p2 - mid) / size;
```
Explorer09: Another TODO for myself (micro-optimization opportunity):
Test if the compiled assembly code can be smaller if I let `mergeRuns` accept pointers instead of lengths as arguments. There are multiplications and divisions by `size` when I convert from lengths to pointers and back to lengths, and the compiler surely cannot cancel `* size` and `/ size` operations together due to the type being unsigned.

natoscott:
Has any performance comparison been done that is sufficiently compelling that we'd want to keep all this extra code, instead of just using qsort?  Thanks.

Explorer09:
@natoscott The `qsort` will be out for two reasons: (1) It's not stable sort (important when the user would sort the table under multiple keys). (2) It does not take the third argument that can make the sort function reentrant. With `qsort` we would need global variables to pass in extra states that would affect sorting (thus unsafe in potential multithread use).

Note my current implementation is not _fully_ async-signal-safe though. The swap operation is not atomic, and can ruin the data when two sort function calls contest on the same data. It's expected that the sort function, when called in asynchronous contexts, operates on different data or the caller must guard the data with a mutex.

natoscott:
Sort stability is about what happens when two items compare equally (which can happen with/without multiple sort keys, so I don't follow your argument there).   In htop, we always have the PID we can use as a tie-breaker to guarantee stability.  Can you explain how this is insufficient?

Regarding threads and the extra parameter, htop is single threaded and is very likely to remain so.  So again this is not a valid technical argument against the use of qsort here.

qsort is highly optimised and widely reviewed.   Any custom home-grown sorting code we have in htop simply isn't, to anywhere near the same degree, and appears to be unnecessary code bloat.  So unless there is a valid technical reason otherwise (and none have been put forward so far, AFAICT), I think we should switch to qsort and delete all this code.

Explorer09:
@natoscott I assume you have read #1784 and #1785 for the discussion I had with @BenBE before.

Basically I would vote for using qsort() (#1784) if there isn't a stability requirement. I suggested that the old `insertionSort` API is there because there would be some data that need to be stable sorted. It's also for future usefulness (or utility?) for the API.

And no, I don't think tiebreaking by PID is enough. htop 's process table can be sorted with multiple keys, and there will be a need to sort the table multiple times by the user _without explicitly specifying the second sort key_. There can be also tables where there isn't an ID field for tiebreaking.

And there's a technical thing: The `Vector` object in htop stores an array of _pointers_ rather than an array of object data themselves. This mean we cannot pass the `Object_compare` function pointers directly to qsort(), but need to make a wrapper for compare function. That's why the third argument of the compare function is needed (and qsort() will not be the best API for the task). There's qsort_r(), but checking the availability of it in configure is also a pain.

Therefore the most future-proof way is to implement a stable sorting function by ourselves. Note that I've developed this algorithm for about half a month, and it would be sad for me if it's ditched finally.

natoscott:
I have read those, yes - there's little/no justification there for custom sort implementations.  It seemed like there was consensus that qsort would be fine based on Hishams reasons for the original implementation, but then it somehow became an exercise in writing new code with insufficient justification IMO.

| There can be also tables where there isn't an ID field for tiebreaking.

What tables are you referring to here?  I don't see them in the code today (something else I've missed?) - if they exist today, do these things require sort stability?  Thanks.

The main (only?) data we need to sort is the process section (lower part) of the htop display, of course, where we always have a unique identifier (PID) for every entry.  The PID provides stability with any number of secondary sort keys, its not clear what problem you are talking about there.  And the user never needs to be involved or know it is being used, again that's a tangent that just doesn't make any sense to me.  Just do all the current sort comparisons, including with multiple keys, and only use the PID as a final, stable, unique differentiator when everything else compares as equal.  Sure, some code refactoring may be required to ensure the PID is available within comparison routines, but its a solvable problem.

I understand the limitations of the qsort API (these are solvable for htop though).  I also understand the problems of creating custom sort implementations.  Its relatively straight-forward to sort things, its more difficult to perform well in general and particularly with very large process counts.

| Therefore the most future-proof way [...]

It sounds like you are designing for problems htop does not currently have.  Can you explain specifically what future features you foresee that cannot use qsort with a stable identifier tie-breaker?  Thanks.

Either way, it will be interesting to see performance numbers comparing performance at both very large and small numbers of processes.  If a different algorithm significantly outperforms qsort for sorting processes in htop, we should use it.

Explorer09:
> The main (only?) data we need to sort is the process section (lower part) of the htop display, of course, where we always have a unique identifier (PID) for every entry. The PID provides stability with any number of secondary sort keys, its not clear what problem you are talking about there. And the user never needs to be involved or know it is being used, again that's a tangent that just doesn't make any sense to me. Just do all the current sort comparisons, including with multiple keys, and only use the PID as a final, stable, unique differentiator when everything else compares as equal. Sure, some code refactoring may be required to ensure the PID is available within comparison routines, but its a solvable problem.

The current process compare functions always sort PID in the ascending order for tiebreaking. No way to make it into descending order. Sorry I just don't like this. (If we were sorting a table in a SQL database, a fallback key for tiebreaking is reasonable. But htop is not a DB.)

> 
> I understand the limitations of the qsort API (these are solvable for htop though). I also understand the problems of creating custom sort implementations. Its relatively straight-forward to sort things, its more difficult to perform well in general and particularly with very large process counts.

Global variables, I know. When I proposed that @BenBE doesn't seem to like it. So it's not my decision on this part.

qsort() uses introsort in glibc, and smoothsort in musl. 
I'm sure introsort requires O(log(n)) space in the worst case. I didn't understand smoothsort yet and so I can't judge it. For me, the memory space requirement means even the qsort() algorithm can crash htop if the system is very low on memory and there's a very large process list to sort. The space complexity is not better than the "natural, in place merge sort" that I propose here.

So, on the question of how the algorithm performs well, I think we're judging quite subjectively. And it would be better if we have some benchmarks.

(By the way, the old PR is there if you want to test the use of qsort() in htop.)

Explorer09:
> If a different algorithm significantly outperforms qsort for sorting processes in htop, we should use it.

@natoscott @BenBE 

Just a minor note. Sorting algorithms' performance can be compared with multiple kinds of data (use cases), and so it won't be a single metric that could decide which algorithm would win.

What I am saying is there can be uniformly random data as well as partially sorted data that need to be sorted. And for an algorithm like heapsort (as an example) it would work best with random data but relatively poor with data that's half sorted. Quicksort with a naïve pivot selection can easily make it the worst case (O(n²)) for data that's already sorted or reverse-sorted, you know. I'm not criticising either algorithm or qsort() API, just a note that the qsort() default algorithm is _not adaptive_ in most libc implementations, thus not as "universally useful" as you think.

And there is no "perfect" sorting algorithm. Each algorithm comes with a trade-off. Even my algorithm does. If you are looking for [block sort family](https://youtube.com/watch?v=InGeRuRk3f8) that looks "perfect" enough in the stats, I did realise there is a tradeoff (adaptive sort i.e. O(n) best case time complexity or O(1) space complexity, pick one). So really no algorithm is perfect.

natoscott:
@Explorer09 I did a little more digging and have found all supported htop platforms have a variant of qsort_r - IIRC you mentioned earlier that its missing somewhere?  (even modern Solaris has it, and Windows has qsort_s).  Which platforms were you thinking do not have it?

There are differences in the calling conventions between platforms, but they are trivially resolved via wrapper.

| And there is no "perfect" sorting algorithm. 

From a htop POV, when we can delete substantial code it's a good win - so qsort[_r] would be perfect in that sense IMO.  I honestly cannot fathom why anyone would be implementing custom sorting code for a system performance tool, that just blows my mind.

If we did need qsort_r for some unusual / new platform, there are plenty of existing, tested, robust implementations to pick from that are license-compatible with htop - choose one, put it below generic/ and add some configure magic.  Done.

Explorer09:
> There are differences in the calling conventions between platforms, but they are trivially resolved via wrapper.

The catch is how to detect the calling conventions.

```c
// Older BSD
void qsort_r(void*, size_t, size_t, void*, int(*)(void*, const void*, const void*));
// POSIX, GNU
void qsort_r(void*, size_t, size_t, int(*)(const void*, const void*, void*), void*);
```

Arguments 4 and 5 are both pointers, so only the const qualifier can tell which argument does what. At this point I am no longer interested in writing a `configure` code for this.

natoscott:
| The catch is how to detect the calling conventions.

Each platform will know the conventions of its own qsort_r, and a xxx/Platform.h wrapper can map that to whichever convention is chosen for use throughout htop.  The only requirement for configure is to detect whether there is a qsort_r at all ... which AIUI you said some platforms are missing?  If none are missing it, then no configure changes are needed.  If it is missing somewhere, configure would need to enable a generic/qsort.c implementation (with contents from somewhere like https://github.com/freebsd/freebsd-src/blob/main/lib/libc/stdlib/qsort.c for example).

Explorer09:
> If we did need qsort_r for some unusual / new platform, there are plenty of existing, tested, robust implementations to pick from that are license-compatible with htop - choose one, put it below generic/ and add some configure magic. Done.

I just heard [on DistroWatch](https://distrowatch.com/dwres-mobile.php?resource=showheadline&story=20087) that htop is [being ported to Redox OS](https://www.redox-os.org/news/this-month-251031/)! And no, I don't know them. It's just something I heard unintentionally.

And FYI, [Redox OS's libc didn't have qsort_r at the moment](https://github.com/redox-os/relibc/blob/master/src/header/stdlib/mod.rs#L1036), although I believe it's trivial for them to implement one.

> > The catch is how to detect the calling conventions.
> 
> Each platform will know the conventions of its own qsort_r, and a xxx/Platform.h wrapper can map that to whichever convention is chosen for use throughout htop. The only requirement for configure is to detect whether there is a qsort_r at all ... 

FreeBSD provides both versions of the `qsort_r` API, with a feature macro (it's [in the code you cited](https://github.com/freebsd/freebsd-src/blob/main/lib/libc/stdlib/qsort.c)). And to test the availability of `qsort_r`, `AC_SEARCH_LIBS` won't work (the macro is limited to checking unmangled symbol only). A custom test is needed in `configure`, really.

BenBE:
@coderabbitat Please note the context and requirements given in #1785.

BenBE:
@coderabbitai The benchmark branch up to d3cd557e is ready as a baseline (basically uses the existing code) but with the new interface. To minimize effects caused by function signatures, using Object_SortFn is highly encouraged. This makes comparing the results much easier to compare. If not using that signature, a good reason for diverting from it should be given though.

PR #1801, opened 2025-11-03
macOS pcp fixes

natoscott:
Final fixes that allow builds on macOS to have a functional pcp-htop with latest upstream PCP.

### DIFF
--- a/pcp/Metric.c
+++ b/pcp/Metric.c
@@ -75,36 +75,95 @@ int Metric_instanceOffset(Metric metric, int inst) {
    return 0;
 }
 
-static pmAtomValue* Metric_extract(Metric metric, int inst, int offset, pmValueSet* vset, pmAtomValue* atom, int type) {
+static pmAtomValue* Metric_extract(Metric metric, int inst, int offset, pmValueSet* vset, pmAtomValue* atom, const pmDesc **desc, int type) {
 
    /* extract value (using requested type) of given metric instance */
-   const pmDesc* desc = &pcp->descs[metric];
+   const pmDesc* outdesc = &pcp->descs[metric];
    const pmValue* value = &vset->vlist[offset];
-   int sts = pmExtractValue(vset->valfmt, value, desc->type, atom, type);
+   int sts = pmExtractValue(vset->valfmt, value, outdesc->type, atom, type);
    if (sts < 0) {
       if (pmDebugOptions.appl0)
          fprintf(stderr, "Error: cannot extract %s instance %d value: %s\n",
                          pcp->names[metric], inst, pmErrStr(sts));
       memset(atom, 0, sizeof(pmAtomValue));
    }
+   *desc = outdesc;
    return atom;
 }
 
-pmAtomValue* Metric_instance(Metric metric, int inst, int offset, pmAtomValue* atom, int type) {
-
+static const pmDesc* Metric_instanceDesc(Metric metric, int inst, int offset, pmAtomValue* atom, int type) {
    pmValueSet* vset = pcp->result->vset[metric];
    if (!vset || vset->numval <= 0)
       return NULL;
 
    /* fast-path using heuristic offset based on expected location */
-   if (offset >= 0 && offset < vset->numval && inst == vset->vlist[offset].inst)
-      return Metric_extract(metric, inst, offset, vset, atom, type);
+   const pmDesc* desc = NULL;
+   if (offset >= 0 && offset < vset->numval && inst == vset->vlist[offset].inst
+       && Metric_extract(metric, inst, offset, vset, atom, &desc, type))
+      return desc;
 
    /* slow-path using a linear search for the requested instance */
    for (int i = 0; i < vset->numval; i++) {
-      if (inst == vset->vlist[i].inst)
-         return Metric_extract(metric, inst, i, vset, atom, type);
+      if (inst == vset->vlist[i].inst
+          && Metric_extract(metric, inst, i, vset, atom, &desc, type))
+         break;
    }
+   return desc;
+}
+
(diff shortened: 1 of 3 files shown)


pcp/Metric.c
```diff
@@ -75,36 +75,95 @@ int Metric_instanceOffset(Metric metric, int inst) {
...
+   const pmDesc *desc = Metric_instanceDesc(metric, inst, offset, atom, PM_TYPE_U64);
+   if (desc)
+      return kibibytes(atom, desc->units.scaleSpace);
+   return NULL;
+}
+
+pmAtomValue* Metric_instance_milliseconds(Metric metric, int inst, int offset, pmAtomValue* atom) {
+   const pmDesc *desc = Metric_instanceDesc(metric, inst, offset, atom, PM_TYPE_U64);
+   if (desc)
+      return milliseconds(atom, desc->units.scaleTime);
```
BenBE, line 166: Minor nit: group the helper near the function where it is called in.

I.e. kibi helper before Metric_kibi, followed by milliseconds helper before Metric_milliseconds …

pcp/Metric.c
```diff
@@ -165,14 +224,15 @@ bool Metric_fetch(struct timeval* timestamp) {
...
       pcp->result = NULL;
    }
-   int sts, count = 0;
+   int sts, count = 0, total;
    if (pcp->reconnect) {
       if (pmReconnectContext(pcp->context) < 0)
          return false;
       pcp->reconnect = false;
    }
+   total = (int) pcp->totalMetrics;
```
BenBE:
```suggestion
   if (pcp->reconnect) {
      if (pmReconnectContext(pcp->context) < 0)
         return false;
      pcp->reconnect = false;
   }
   int sts, count = 0;
   int total = (int) pcp->totalMetrics;
```

Reduce scope of these variables

pcp/Platform.c
```diff
@@ -350,7 +350,7 @@ bool Platform_init(void) {
    if (opts.context == PM_CONTEXT_ARCHIVE) {
       gettimeofday(&pcp->offset, NULL);
 #if PMAPI_VERSION >= 3
-      struct timeval start = { opts.start.tv_sec, opts.start.tv_nsec / 1000 };
+      struct timeval start = { opts.start.tv_sec, (int) (opts.start.tv_nsec / 1000) };
```
Explorer09: Don't expect the type `suseconds_t` to be always synonym of `int`
```suggestion
      struct timeval start = { opts.start.tv_sec, (suseconds_t)(opts.start.tv_nsec / 1000) };
```

pcp/PCPDynamicColumn.c
```diff
@@ -166,7 +167,7 @@ static void PCPDynamicColumn_parseFile(PCPDynamicColumns* columns, const char* p
       } else if (value && column && String_eq(key, "description")) {
          free_and_xStrdup(&column->super.description, value);
       } else if (value && column && String_eq(key, "width")) {
-         column->super.width = strtoul(value, NULL, 10);
+         column->super.width = atoi(value);
```
Explorer09, line 170: I can't understand why the change from `strtoul` to `atoi` here, especially that `atoi` will make the behavior undefined for values that are out of representable range.

BenBE: FWIW, due to lack of checking the return value/error code, the former was somewhat UB too … ;-)

Explorer09: @BenBE 
Ideally I would like all the integer parsings in htop express what they would do when it comes to out-of-range values. Some could be "don't care", some others could throw errors, and yet other could clamp the values silently to allowed ranges or revert to a hard-coded default.

natoscott: @Explorer09 reason was 'width' is an int, and strtoul with those parameters is equivalent to atoi.  Since this is a config file parser and not interactively having to deal with potential errors, I went with the simplest solution not requiring a cast.

Explorer09:
Some type related fixes here seem to be related to #1673 that I reported before, but I didn't have time to review them all...

pcp/Metric.c
natoscott: Agreed, done!

pcp/PCPDynamicColumn.c
natoscott: So, I think this one is fine - if someone creates a config file specifying a column width over 2.1+ billion? - meh, they can keep both broken pieces.

pcp/Metric.c
natoscott: Done.

BenBE (approved):
LGTM.

Explorer09:
@natoscott While this PR is merged successfully, I wonder how we can set up pcp-htop monitoring in macOS, and in particular, whether we can test build it in the CI server. Do you have an idea of adding a build job in CI to constantly test-build in this environment?

natoscott:
@Explorer09 it's complicated by the PCP installation, which is a .dmg only and not brew at this stage.  Linux is easy cos PCP is in all the distros, for many years - not so for macOS packages unfortunately.

Explorer09:
@natoscott I can see the [pcp package in Homebrew](https://formulae.brew.sh/formula/pcp) is completely unrelated to the [Performance Co-Pilot](https://pcp.io/download.html) we are talking about.

But it does seem like there's a way to mount the .dmg file from the macOS Terminal, and eventually automate the installation of that.

The reason I made such a request is to prevent the similar compile warnings from appearing again. When I compiled the list of warnings you see in #1673, I definitely missed some platforms and build configurations e.g. I can't set up Clang in some BSD platforms and also Solaris.

PR #1802, opened 2025-11-03
Adjust GPUMeter text display

Explorer09:
* Character width for percentage numbers changed from 4 to 5 (same as CPUMeter percentage formatting)
* Percent signs "%" are now in the same color as the numbers preceding them (for a better look)
* No more "%" suffix if the number will display as "N/A"

htop-screenshot-gpu-meter-change

### DIFF
--- a/GPUMeter.c
+++ b/GPUMeter.c
@@ -103,9 +103,8 @@ static void GPUMeter_display(const Object* cast, RichString* out) {
       return;
    }
 
-   written = xSnprintf(buffer, sizeof(buffer), "%4.1f", totalUsage);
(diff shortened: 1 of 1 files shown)


BenBE:
With the CPU meter, the color of the % was grey if the bar didn't quite reach the spot the % covers. Is this consistent?

Explorer09:
> With the CPU meter, the color of the % was grey if the bar didn't quite reach the spot the % covers. Is this consistent?

Umm... I think you confused something. This fix is about the _text_ mode display of the GPUMeter. Nothing to do with bar mode.

PR #1804, opened 2025-11-03
Change NetworkIOMeter "packets per second" display

Explorer09:
htop-screenshot-networkio-text-change

I would like @eworm-de to come and discuss about this one.

The lower half of this picture is my proposed change to the NetworkIOMeter. Basically I feel the current "packets per second" display is unintuitive. There is no telling that the number on the left of "/" is RX and the number on the right of "/" is TX. So, I suggest the coloring on the numbers. <s>Also, because the "pkt/s" unit contains a slash, I thought it could be easier to read if I move the text to before the two numbers.</s> (Update: I moved the "pkt/s" unit back to the right after the discussion.)

### DIFF
--- a/NetworkIOMeter.c
+++ b/NetworkIOMeter.c
@@ -79,7 +79,7 @@ static void NetworkIOMeter_updateValues(Meter* this) {
 
          if (data.packetsReceived > cached_rxp_total) {
             diff = data.packetsReceived - cached_rxp_total;
-            diff = (1000 * diff) / passedTimeInMs; /* convert to pkts/s */
+            diff = (1000 * diff) / passedTimeInMs; /* convert to pkt/s */
             cached_rxp_diff = (uint32_t)diff;
          } else {
             cached_rxp_diff = 0;
@@ -96,7 +96,7 @@ static void NetworkIOMeter_updateValues(Meter* this) {
 
          if (data.packetsTransmitted > cached_txp_total) {
             diff = data.packetsTransmitted - cached_txp_total;
-            diff = (1000 * diff) / passedTimeInMs; /* convert to pkts/s */
+            diff = (1000 * diff) / passedTimeInMs; /* convert to pkt/s */
             cached_txp_diff = (uint32_t)diff;
          } else {
             cached_txp_diff = 0;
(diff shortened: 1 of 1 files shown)


eworm-de:
Well, RX and TX always had the same order, but other than that there was no indication. I like your suggestion, makes it a lot cleaner! 👍 Thanks!

BenBE:
I like the color change, but not putting the unit as a label.

eworm-de:
Having a second thought on this... If space is missing the output is just cut off. So possibly it's better to have the values first, units last? Makes more sense to cut off the units than the values.

Explorer09:
@eworm-de What do you mean by "cut off"? In the current bar meter drawing code, if there isn't enough space to print all text, the last "token" (separated by space character (0x20)) will be removed entirely to make space for it. If it's only one token left and still not enough space, the token will cut of character by character.

In the text and LED meter modes, the text is always removed character by character when there is not enough space to print all.

Because the item "(456/123 pkt/s)" or "(pkt/s:456/123)" should display as one token, there is no space character between the numbers and the unit in bar meter mode. When it has to cut off the token will be removed as a whole. Therefore it doesn't matter whether the unit goes first or the numbers go first.

The question will be whether the user can find it easier to read if I put the unit first or the unit last. But there's a problem (for me) that there are two slashes here with different meanings. One is a separator between a number pair and the other for "per" or "division" meaning.

@BenBE  I can move the "pkt/s" unit back to after the numbers. Just not sure if that would look better. Screenshot will be updated soon.

eworm-de:
Have a look at this network meter in text mode:
Screenshot_2025-11-04_12-12-57

It ends with "`(25/24 p`" as the column does not provide more space. Moving the unit there as a label would show that, but no values.

eworm-de:
How about switching the unit to just `pps` (for "**p***ackets* **p***er* **s***econd*")?

Explorer09:
> How about switching the unit to just `pps` (for "**p**_ackets_ **p**_er_ **s**_econd_")?

A quick Google search by me shows that "pps" does have some uses, but not yet as widespread as "bps" or "B/s".

So, maybe we can say "pps" in bar meter mode, but the full "packets/sec" in text meter mode. That could be a usability win for both.

BenBE:
There is an inconsistency in the format: One place adds `pps`, while the other place uses `pkt/s`.

PR #1805, opened 2025-11-03
CI: Add macos-latest-pcp build job

Explorer09:
This helps track whether we can build `pcp-htop` for macOS.
I come up this after the discussion in #1801.

Note: Currently this build job is broken. Don't merge yet, but I think @natoscott can take a look at this draft and suggest how to fix the problems.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -215,6 +215,31 @@ jobs:
     - name: Distcheck
       run: make distcheck DISTCHECK_CONFIGURE_FLAGS="--enable-werror"
 
+  build-macos-latest-pcp:
+    runs-on: macOS-latest
+    env:
+      CC: clang
+      # '-mmacos-version-min=15.4' is a workaround for PCP 7.0.2 that
+      # accidentally broke backward compatibility.
+      # (https://github.com/performancecopilot/pcp/pull/2400)
+      CFLAGS: -O3 -g -flto -mmacos-version-min=15.4
+      LDFLAGS: -O3 -g -flto
+    steps:
+    - uses: actions/checkout@v5
+    - name: Install Dependencies
+      run: |
+        pcp_version=7.0.2-1
+        brew install automake pkg-config
+        curl -f -L -O https://github.com/performancecopilot/pcp/releases/download/$(echo ${pcp_version} | sed 's/-.*$//')/pcp-${pcp_version}.dmg
+        hdiutil attach pcp-${pcp_version}.dmg
+        sudo installer -verbose -pkg /Volumes/pcp-${pcp_version}/pcp-${pcp_version}.pkg -target /
+    - name: Bootstrap
+      run: ./autogen.sh
+    - name: Configure
+      run: ./configure --enable-werror --enable-pcp --enable-unicode || ( cat config.log; exit 1; )
+    - name: Build
+      run: make -k
+
   build-dragonflybsd-latest-gcc:
       runs-on: ubuntu-22.04
       timeout-minutes: 20


natoscott:
@Explorer09 thanks!  Ah, the strchrnul on macOS warning.  Hmm, there's just one file in libpcp that uses this - probably the simplest way to tackle this will be to refactor this so we don't use strchrnul in PCP on macOS at all, and setup this pcp-htop macOS automation using the pcp-7.0.3 release onward (scheduled for later this month, 27th Nov).

Explorer09:
@natoscott I think official builds of PCP for macOS should disable `strchrnul`. You probably know why.
See also #1660, where I added the custom configure check for `strchrnul` for htop. It can be adapted also to PCP's configure script as well.

natoscott:
Yep, thanks, I saw that - since we don't really need it in PCP, I think best is to remove it entirely and the problem goes away.

natoscott:
Related: https://github.com/performancecopilot/pcp/pull/2400

Explorer09:
While the current build job builds `pcp-htop` successfully on macOS with a `-mmacos-version-min=15.4` workaround, at configure time there's an error that I'm not sure if it would matter at runtime. So still don't merge this to `main` yet.

```text
dyld[13296]: Library not loaded: libpcp.4.dylib
checking for NaN support... no
  Referenced from: <CC8E5D85-A629-3F97-A3D8-562CFE12D69E> /Users/runner/work/htop/htop/conftest
  Reason: tried: 'libpcp.4.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OSlibpcp.4.dylib' (no such file), 'libpcp.4.dylib' (no such file), '/Users/runner/work/htop/htop/libpcp.4.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/Users/runner/work/htop/htop/libpcp.4.dylib' (no such file), '/Users/runner/work/htop/htop/libpcp.4.dylib' (no such file)
./configure: line 7188: 13296 Abort trap: 6           ./conftest$EXEEXT 1>&5
configure: WARNING: runtime behavior with NaN is not compliant - some functionality might break; consider using '-fno-finite-math-only'
```

natoscott:
Looks like a DYLD_LIBRARY_PATH setting will be needed there so that libpcp.4.dylib can be found (or copy it into the working directory for the build).

Explorer09:
> Looks like a DYLD_LIBRARY_PATH setting will be needed there so that libpcp.4.dylib can be found (or copy it into the working directory for the build).

How to set it then? Note that this is the first time I tried to install PCP on macOS and I don't even have the idea on how PCP works.

natoscott:
I can't seem to find a magic DYLD environment variable incantation that works here - no matter what I use the configure-generated NaN binary check utility ignores it.  But this does the trick (from the repo clone top level directory):
```
ln -s /usr/local/lib/libpcp.4.dylib libpcp.4.dylib
```

.github/workflows/ci.yml
```diff
@@ -215,6 +215,31 @@ jobs:
...
+      # '-mmacos-version-min=15.4' is a workaround for PCP 7.0.2 that
+      # accidentally broke backward compatibility.
+      # (https://github.com/performancecopilot/pcp/pull/2400)
+      CFLAGS: -O3 -g -flto -mmacos-version-min=15.4
+      LDFLAGS: -O3 -g -flto
+    steps:
+    - uses: actions/checkout@v5
+    - name: Install Dependencies
+      run: |
+        pcp_version=7.0.2-1
```
natoscott: @Explorer09 this is already out-of-date :( - and PCP releases quite often, usually once every couple of months or less.  I wonder if there's any way we can find latest release from github?

PR #1808, opened 2025-11-07
Allow configuration without os-release file; make that default for macOS

Explorer09:
Although the `/etc/os-release` file is nearly universal for Linux distributions as well as BSDs and Solaris, macOS notably doesn't support it.

Improve the configure script logic. Make it a valid configuration without the os-release file (user can specify that as `--with-os-release=''` or `--without-os-release`). Also make `--without-os-release` the default for macOS build.

Also, don't check the existence of `/etc/os-release` or the fallback `/usr/lib/os-release` when user is cross-compiling.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1202,7 +1202,12 @@ AC_ARG_WITH(
       [location of an os-release file @<:@default=/etc/os-release@:>@]
    )],
    [],
-   [with_os_release=yes]
+   [
+      with_os_release=yes
+      if test "$my_htop_platform" = darwin; then
+         with_os_release=''
+      fi
+   ]
 )
 case x$with_os_release in
    x|xno)


configure.ac
```diff
@@ -1202,7 +1202,12 @@ AC_ARG_WITH(
       [location of an os-release file @<:@default=/etc/os-release@:>@]
    )],
    [],
-   [with_os_release=yes]
+   [
+      with_os_release=yes
+      if test "$my_htop_platform" = darwin; then
```
Explorer09, line 1207: Note: I don't know whether I should check `$host_os` or `$my_htop_platform` if I want to disable the os-release file parsing for macOS.

Explorer09:
According to [this question in Super User](https://superuser.com/questions/75166/how-to-find-out-mac-os-x-version-from-terminal), macOS stores the version information in this `/System/Library/CoreServices/SystemVersion.plist` file. Unfortunately this file is in an XML format, so if htop were to support this file, it needs a custom parser.

configure.ac
BenBE: Use `$my_htop_platform`, which is the abstration that ultimtely decides, which parts of the platform-specific code is linked into htop.

Explorer09: @BenBE What to do if `$my_htop_platform` is `pcp` but `$host_os` is `darwin*` (like #1805)? That's the question I'm asking.

For me, `$my_htop_platform` seems to be referring to what set of API htop should use for monitoring (it's like `--target` in some way even when htop is not a compiling tool). It's less appropriate for specifying dependencies of host libraries or host environment.

Maybe @natoscott can help answer this?

natoscott: @Explorer09 @BenBE good morning!  I think our approach here should be different.  AIUI, what you're trying to do is to make the generic/uname.c code work with macOS, because ...

```
$ grep Generic_uname darwin/*
darwin/Platform.h:   *string = Generic_uname();
```

An approach that I think would be more fitting with the design of the generic/ code is to accept that, for macOS, the implementation of generic/uname.c is incorrect.  Instead, the approach could be to implement Platform_getRelease without using Generic_uname.

Ah but its more subtle than that, I see.  We still want to use the parts of Generic_uname that call uname(3) on macOS, but not the os-release parts that are embedded inconveniently in the middle of that function.

What I suggest we do in that case is to move parseOSRelease() into a separate Generic_OS_release function/file such that for macOS we can provide a different implementation of just that part (in darwin/Platform.c).

BenBE: Correct. The `generic/` code was split out as "works on most platforms" to avoid code duplication where things are almost the same across platforms.

Without looking into too much details, we might be able to split out the "lookup OS release info" part into an optional callback, which if not provided calls a default (generic) implementation doing as it does not, but if provided replaces how that function looks up the OS release info. (Which iIUC the previous comment correctly should mostly align with @natoscott's approach)

Explorer09: @BenBE I like the callback idea, but the purpose of this PR is to disable looking for `/etc/os-release` in macOS since the file never exists in Mac. The callback can be implemented later.

By the way, I even tried [PureDarwin](https://www.puredarwin.org/) in a VM and PureDarwin doesn't provide `/etc/os-release` file either so I can say it's safe to disable by default for all Darwin-based systems.

natoscott: @BenBE yep, a callback should work well here.  @Explorer09 it's OK for configure to look for files that don't exist, but we should not put a NULL-check hack into the parseOSRelease function IMO - that routine should just not be called on macOS.

Explorer09: @natoscott If it's not the NULL-check hack, what is your suggestion?

natoscott: Replace the parseOSRelease function (which becomes the Linux/*BSD callback in Benny's approach) with a macOS-specific callback (e.g. darwin_parseOSRelease or some such name).

Explorer09:
@natoscott When I tried to find a way to refactor `Platform_getRelease` and related functions, I realized the current code structure of `Platform_getRelease` is a mess.

In particular, SysArchMeter is currently the main consumer of `Platform_getRelease`, and SysArchMeter caches the "address to the string" that `Platform_getRelease` outputs. All platform code in htop expect that `Platform_getRelease` would use a static buffer that lasts until the htop program end, except two platforms: Unsupported and PCP.

* `Platform_getRelease` in Unsupported outputs an `xStrdup`'d string without a corresponding `free` call.

* PCP's `Platform_getRelease` outputs an allocated string buffer that's owned by `pcp` object (in `pcp/Platform.c`) and that buffer can be `free`'d in `Platform_done`. That creates an inconsistency in managing the string buffer lifetime (a theoretical use-after-free possibility).

The PCP one is what I'm concerning now. I'm asking a permission to remove the `release` member from the `pcp` object and use a separate, static buffer for PCP's `Platform_getRelease` instead. That would make the objects' lifetime consistent. I also plan to change the argument type of `Platform_getRelease` from `char**` to `const char**` to indicate more clearly the data ownership (SysArchMeter shouldn't alter or `free` the buffer that `Platform_getRelease` outputs).

natoscott:
| All platform code in htop expect that Platform_getRelease would use a static buffer that lasts until the htop program end

The expectation is not "until htop ends", rather its until the (SysArch) Meter code can never access it again.  What both PCP and Unsupported platforms are doing is perfectly fine (although Unsupported should free it, I agree) - it can be argued the PCP way is superior to yet another static variable embedded in some Meter somewhere, and the other platforms should do something more like PCP (perhaps storing the pointer in Machine.h?).

What actual problem are you trying to solve here?  So far this seems like code churn to me, with little if any gain.

Explorer09:
> > All platform code in htop expect that Platform_getRelease would use a static buffer that lasts until the htop program end
> 
> The expectation is not "until htop ends", rather its until the (SysArch) Meter code can never access it again. What both PCP and Unsupported platforms are doing is perfectly fine (although Unsupported should free it, I agree) - it can be argued the PCP way is superior to yet another static variable embedded in some Meter somewhere, and the other platforms should do something more like PCP (perhaps storing the pointer in Machine.h?).
> 
> What actual problem are you trying to solve here? So far this seems like code churn to me, with little if any gain.

Split the `Platform_getRelease` into two parts, one grabs the `uname` strings and another the distro.

For now I am thinking how to make the new API until I found out the inconsistency on the object storage lifetimes.

The alternative is to remove the static (cached) string pointer (which can be _dangling_) in the SysArchMeter and force the meter to call `Platform_getRelease` every time to retrieve the `uname` and `os-release` info (while `Platform_getRelease` caches the string and return its address).

The idea is to not keep the address to the buffer that could be `free`'d by someone else. (That can confuse static analyzers or cause the UAF problem later on.) It's a memory safety problem that, if written in a language with stricter memory safety requirements (e.g. Rust), would not have been allowed.

natoscott:
So in summary, no actual issues that a user might observe are being fixed here.

I've opened #1812 to fix the two actual issues identified here so far - i.e. lack of macOS version support in the SysArchMeter and the strdup with no matching free on the 'unsupported' platform.  I used the callback approach @BenBE described.  Not 100% convinced this is better than separating out the distro logic completely, but I do agree it's a simpler change to make than my earlier suggestion of splitting the generic uname routine in two, so I'm happy enough with that.

Explorer09:
I'm closing this PR until I can propose something better. The discussion of macOS release file will continue in #1812.

PR #1809, opened 2025-11-08
NetBSD: Improve process state retrieval code

GuillaumeGomez:
Also: only iterate on possible states if `kinfo_lwp` struct is not `NULL`.

If you're interested into it (if you agree, gonna be in a follow-up PR): we currently need to check at each iteration of the loop if the state is not `UNKNOWN` because of the `switch`. Instead of doing that, I'm thinking of moving this code into a function so we can return the value to put into `proc->state` directly, allowing to remove this check. What do you think?

### DIFF
--- a/netbsd/NetBSDProcessTable.c
+++ b/netbsd/NetBSDProcessTable.c
@@ -150,6 +150,24 @@ static double getpcpu(const NetBSDMachine* nhost, const struct kinfo_proc2* kp) 
    return 100.0 * (double)kp->p_pctcpu / nhost->fscale;
 }
 
+static ProcessState get_active_status(const NetBSDMachine* nhost, const struct kinfo_proc2* kproc) {
+   int nlwps = 0;
+   const struct kinfo_lwp* klwps = kvm_getlwps(nhost->kd, kproc->p_pid, kproc->p_paddr, sizeof(struct kinfo_lwp), &nlwps);
+   if (!klwps) {
+      return UNKNOWN;
+   }
+   // We only consider the first LWP that has one of the states below.
+   for (int j = 0; j < nlwps; j++) {
+      switch (klwps[j].l_stat) {
+         case LSONPROC: return RUNNING;
+         case LSRUN:    return RUNNABLE;
+         case LSSLEEP:  return SLEEPING;
+         case LSSTOP:   return STOPPED;
+      }
+   }
+   return UNKNOWN;
+}
+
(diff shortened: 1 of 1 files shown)


BenBE:
Can you elaborate on the rationale of this change?

Please fix the errors from the compiler as well as the code style of your change.

GuillaumeGomez:
Sure:

* `klwps` is only used if `kproc->p_realstat` is `SACTIVE`. Meaning that the rest of the time, it's useless to query this information, so I moved this call where it's used.
* If `klwps` is `NULL`, there is no point in iterating `nlwps`, so I moved the `if (klwps)` outside of the `for` loop so it's only checked once.

GuillaumeGomez:
Ah I see, so for fixing the error and making my fix possible, I actually need to do what I suggested as a follow-up. Oh well, I guess a second commit won't hurt, sorry for the noise.

GuillaumeGomez:
Fixed the warning, CI is happy. :)

BenBE:
Please squash the two commits into one (cf. style guide).

Explorer09:
Your commit title is kinda long. When you squash the commit, I suggest this title:

```text
NetBSD: Improve process state retrieval code

Only call `kvm_getlwps` when `p_realstat` is `SACTIVE`, and only
iterate on possible states if `kinfo_lwp` struct is not NULL.
```

GuillaumeGomez:
Done!

netbsd/NetBSDProcessTable.c
```diff
@@ -150,6 +150,24 @@ static double getpcpu(const NetBSDMachine* nhost, const struct kinfo_proc2* kp)
    return 100.0 * (double)kp->p_pctcpu / nhost->fscale;
 }
 
+static ProcessState get_active_status(const NetBSDMachine* nhost, const struct kinfo_proc2* kproc) {
+   int nlwps = 0;
+   const struct kinfo_lwp* klwps = kvm_getlwps(nhost->kd, kproc->p_pid, kproc->p_paddr, sizeof(struct kinfo_lwp), &nlwps);
+   if (!klwps) {
+      return UNKNOWN;
+   }
+   // We only consider the first LWP with a one of the below states.
```
BenBE:
```suggestion
   // We only consider the first LWP that has one of the states below.
```

GuillaumeGomez: Done!

PR #1812, opened 2025-11-11
darwin: implement macOS version reporting in SysArchMeter

natoscott:
Implemented using the callback approach discussed in #1808.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1196,33 +1196,6 @@ esac
 
 
 AC_ARG_WITH(
-   [os-release],
-   [AS_HELP_STRING(
-      [--with-os-release=FILE],
-      [location of an OS release file @<:@default=check@:>@]
-   )],
-   [],
-   [with_os_release=check]
-)
-if test "$with_os_release" = check; then
-   case $host_os in
-   darwin*)
-      with_os_release=/System/Library/CoreServices/SystemVersion.plist
-      ;;
-   *)
-      if test -f /etc/os-release; then
-         with_os_release=/etc/os-release
-      elif test -f /usr/lib/os-release; then
-         with_os_release=/usr/lib/os-release
-      else
-         with_os_release=""
-      fi
-      ;;
-   esac
-fi
-AC_DEFINE_UNQUOTED([OSRELEASEFILE], ["$with_os_release"], [File with OS release details.])
-
-AC_ARG_WITH(
    [config],
    [AS_HELP_STRING(
       [--with-config=DIR],
@@ -1618,7 +1591,6 @@ AC_MSG_RESULT([
   ${PACKAGE_NAME} ${VERSION}
 
   platform:                  $my_htop_platform ($host_os)
-  OS release file:           $with_os_release
   (Linux) proc directory:    $with_proc
   (Linux) openvz:            $enable_openvz
   (Linux) vserver:           $enable_vserver
--- a/darwin/Platform.c
+++ b/darwin/Platform.c
@@ -726,9 +726,7 @@ void Platform_gettime_monotonic(uint64_t* msec) {
 
 }
 
-#ifndef OSRELEASEFILE
 #define OSRELEASEFILE "/System/Library/CoreServices/SystemVersion.plist"
-#endif
 
 static void Platform_getOSRelease(char* buffer, size_t bufferLen) {
    const UInt8* osfile = (const UInt8 *)OSRELEASEFILE;
--- a/generic/uname.c
+++ b/generic/uname.c
@@ -21,12 +21,13 @@ in the source distribution for its full text.
 #endif
 
 
-#ifndef OSRELEASEFILE
-#define OSRELEASEFILE "/etc/os-release"
-#endif
-
 static void parseOSRelease(char* buffer, size_t bufferLen) {
-   FILE* fp = fopen(OSRELEASEFILE, "r");
+   const char* osfiles[] = { "/etc/os-release", "/usr/lib/os-release" };
+   FILE* fp = NULL;
+   for (size_t i = 0; i < sizeof(osfiles)/sizeof(char*); i++) {
+      if ((fp = fopen(osfiles[i], "r")) != NULL)
+         break;
+   }
    if (!fp) {
       xSnprintf(buffer, bufferLen, "No OS Release");
       return;


configure.ac
```diff
@@ -1199,15 +1199,26 @@ AC_ARG_WITH(
...
+   [with_os_release=check]
 )
-if test -n "$with_os_release" && test ! -f "$with_os_release"; then
-   if test -f "/usr/lib/os-release"; then
-      with_os_release="/usr/lib/os-release"
-   fi
+if test "$with_os_release" = check; then
+   case $host_os in
+   darwin*)
+      with_os_release=/System/Library/CoreServices/SystemVersion.plist
```
Explorer09: I think for the Darwin / macOS case, it's fine to just hard-code the file path. No need to make it a configure option since it's unlikely that the file could be located in somewhere else.

darwin/Platform.c
```diff
@@ -725,3 +725,43 @@ void Platform_gettime_monotonic(uint64_t* msec) {
...
+         break;
+      }
+   }
+   fclose(fp);
+
+   snprintf(buffer, bufferLen, "macOS %s", version[0] ? version : "");
+}
+
+void Platform_getRelease(char** string) {
+   *string = Generic_unameRelease(Platform_getOSRelease);
```
Explorer09, line 782: I believe the callback can be an overkill here.

How about:

```suggestion
   static char osRelease[100] = {0};
   if (!osRelease[0])
      Platform_getOSRelease(osRelease, sizeof(osRelease));
   Generic_unameRelease(osRelease);
```

Yes, this way you don't need a callback function, and you can locally cache the `osRelease` string just parsed from the `.plist` file.

darwin/Platform.c
```diff
@@ -725,3 +725,43 @@ void Platform_gettime_monotonic(uint64_t* msec) {
...
+      if (String_startsWith(sp, "<key>ProductVersion</key>")) {
+         getKeyValue = true;
+         continue;
+      }
+      if (getKeyValue == false)
+         continue;
+      if (String_startsWith(sp, "<string>")) {
+         const char* start = sp + strlen("<string>");
+         const char* stop = strstr(sp, "</string>");
+         String_safeStrncpy(version, start, MINIMUM(sizeof(version), (size_t)(stop - start + 1)));
```
Explorer09: I think `sscanf` can do the job with a smaller code...

darwin/Platform.c
```diff
@@ -725,3 +725,43 @@ void Platform_gettime_monotonic(uint64_t* msec) {
...
+         break;
+      }
+   }
+   fclose(fp);
+
+   snprintf(buffer, bufferLen, "macOS %s", version[0] ? version : "");
```
Explorer09: Don't hard-code "macOS". That's what the `<key>ProductName</key>` is for. Besides, Darwin is also the core of iOS, iPadOS, etc. You know it.

pcp/Platform.c
```diff
@@ -717,6 +712,12 @@ void Platform_getRelease(char** string) {
    free(sysname.cp);
 }
 
+void Platform_getRelease(const char** string) {
+   if (pcp->release == NULL)
+      Platform_setRelease();
+   *string = (const char*)pcp->release;
```
Explorer09: No need to cast.

darwin/Platform.c
```diff
@@ -725,3 +725,43 @@ void Platform_gettime_monotonic(uint64_t* msec) {
...
+   }
+
+   /* extract version in: <key>ProductVersion</key>\n<string>X.Y.Z</string>\n */
+   bool getKeyValue = false;
+   char version[64] = {'\0'};
+   char lineBuffer[256];
+   char* sp;
+   while ((sp = fgets(lineBuffer, sizeof(lineBuffer), fp))) {
+      while (*sp != '\0' && isspace(*sp)) sp++;
+      if (String_startsWith(sp, "<key>ProductVersion</key>")) {
```
Explorer09: Parsing an XML is different from parsing a config file with assignment syntaxes. Here I'm sure `String_startsWith` is a wrong function to use. Use `strstr`:

`if (strstr(sp, "<key>ProductVersion</key>") != NULL) ...`

configure.ac
natoscott: I prefer to keep the logic for all platforms together this way so its clear what the situation on macOS is.

darwin/Platform.c
natoscott: I think the callback is OK and it helps keep the logic consistent across the major platforms.

darwin/Platform.c
natoscott: Yep, will extract the name too, good point.  Darwin is specifically referring to the kernel - it can and often does has a different version number to the rest of macOS, interestingly...
Screenshot 2025-11-11 at 1 11 05 pm

| You know it.

Please don't make little comments like this.  You have no idea what I know, and I have no idea what you know (neither of us are mind readers, as far as I know), and I don't have any clue what you're implying when you say this kind of thing (not the first time).

pcp/Platform.c
natoscott: Yep, will fix, thanks.

darwin/Platform.c
natoscott: Agreed, that does away with the leading whitespace handling - will fix up, thanks.

generic/uname.h
```diff
@@ -7,6 +7,12 @@ Released under the GNU GPLv2+, see the COPYING file
 in the source distribution for its full text.
 */
 
+#include <string.h>
+
+typedef void(*releaseFunction)(char* buffer, size_t length);
```
BenBE:
```suggestion
typedef void (*releaseFunction)(char* buffer, size_t length);
```

darwin/Platform.c
```diff
@@ -725,3 +725,58 @@ void Platform_gettime_monotonic(uint64_t* msec) {
 #endif
 
 }
+
+#ifndef OSRELEASEFILE
+#define OSRELEASEFILE "/System/Library/CoreServices/SystemVersion.plist"
+#endif
```
BenBE: Isn't this always defined in `config.h` through `configure`? And if not, this code should FTBTS or at least properly fallback to "No OS release" if `--without-os-release` was specified for `configure`.

Explorer09:
@natoscott @BenBE 
I have one concern here:

Generally I am not impressed that htop would need _two_ different parsers for querying the distro information. Although that reflects the reality that there lacks a uniform standard for the distro information. **That problem was exactly what `/etc/os-release` file was meant to solve** (from the systemd folks). So in a sense, we came back to square one on the distro information thing. (In the Linux world there used to be a LOT of distro-specifc release information file - see [this unofficial list](http://linuxmafia.com/faq/Admin/release-files.html), and maybe also [this](https://unix.stackexchange.com/questions/6345).) Considering that even BSDs and Solaris are moving to support `/etc/os-release` now, I really don't like the Darwin-specific parser to be kept for long. It's ugly for sure.

I mean, theoretically Darwin & macOS can also move to support `/etc/os-release` (or at least provide in parallel with `SystemVersion.plist`). The current callback code kills the possibility.

Yes, this issue is political, especially not all people like systemd or what they are doing, but on the specific case, the systemd prople on unifying to one file for distro information was the right approach here. And any other parser (or "callback") should be there only if the `/etc/os-release` doesn't work on a particular distro.

By the way, since we have two different parsers for two different files, maybe we can considering **killing the `--with-os-release` configure option entirely**. The configure option doesn't do what users would except and can be confusing. Hard code the two paths `/etc/os-release` and `/usr/lib/os-release`. Yes that would alienate distros with unusual file system layout (GoboLinux and others), but those distros would either need symlinks for FHS compatibility (already) or a local patch to htop anyway so I don't see it as a downside.

darwin/Platform.c
```diff
@@ -725,3 +725,58 @@ void Platform_gettime_monotonic(uint64_t* msec) {
...
+      }
+      else if (state == OS_NAME && ((vp = strstr(sp, "<string>")))) {
+         const char* start = vp + strlen("<string>");
+         const char* stop = strstr(vp, "</string>");
+         if (stop && stop > start)
+            String_safeStrncpy(name, start, MINIMUM(sizeof(version), (size_t)(stop - start + 1)));
+      }
+      state = OS_NONE;
+   }
+   fclose(fp);
```
BenBE: Why not use the CF APIs for plist handling when on darwin?

```c
#include <CoreFoundation/CoreFoundation.h>
#include <stdio.h>

static void printCString(CFStringRef cfstr) {
    if (!cfstr) return;
    char buf[256];
    if (CFStringGetCString(cfstr, buf, sizeof(buf), kCFStringEncodingUTF8)) {
        printf("%s\n", buf);
    }
}

int main(int argc, char *argv[]) {
    if (argc != 2) {
        fprintf(stderr, "Usage: %s plist-file\n", argv[0]);
        return 1;
    }

    CFURLRef url = CFURLCreateFromFileSystemRepresentation(
        NULL, (const UInt8 *)argv[1], strlen(argv[1]), false);
    if (!url) {
        fprintf(stderr, "Invalid URL\n");
        return 1;
    }

    CFDataRef data;
    SInt32 errorCode;
    Boolean status = CFURLCreateDataAndPropertiesFromResource(
        NULL, url, &data, NULL, NULL, &errorCode);
    CFRelease(url);

    if (!status || !data) {
        fprintf(stderr, "Failed to read plist\n");
        return 1;
    }

    CFPropertyListRef plist = CFPropertyListCreateWithData(
        NULL, data, kCFPropertyListImmutable, NULL, NULL);
    CFRelease(data);

    if (!plist || CFGetTypeID(plist) != CFDictionaryGetTypeID()) {
        fprintf(stderr, "Not a dictionary plist\n");
        if (plist) CFRelease(plist);
        return 1;
    }

    CFDictionaryRef dict = (CFDictionaryRef)plist;
    CFStringRef productName = CFDictionaryGetValue(dict, CFSTR("ProductName"));
    CFStringRef productVersion = CFDictionaryGetValue(dict, CFSTR("ProductVersion"));

    printf("ProductName: ");
    printCString(productName);
    printf("ProductVersion: ");
    printCString(productVersion);

    CFRelease(plist);
    return 0;
}
```

Only needs a `-framework CoreFoundation` added to `LDFLAGS`.

IMHO that's more robust than hand-rolling our own XML parser.

P.S.: The code itself is generated/untested.

generic/uname.c
```diff
@@ -67,7 +67,7 @@ static void parseOSRelease(char* buffer, size_t bufferLen) {
    snprintf(buffer, bufferLen, "%s%s%s", name[0] ? name : "", name[0] && version[0] ? " " : "", version);
 }
 
-char* Generic_uname(void) {
+char* Generic_unameRelease(releaseFunction platform_release) {
```
BenBE: Not quite sure about the naming of the argument (and its type).

I think `Platform_FetchReleaseFunction` and `fetchRelease` might be better choices.

pcp/Platform.c
```diff
@@ -717,6 +712,12 @@ void Platform_getRelease(char** string) {
    free(sysname.cp);
 }
 
+void Platform_getRelease(const char** string) {
+   if (pcp->release == NULL)
+      Platform_setRelease();
+   *string = pcp->release;
```
BenBE, line 719:
```suggestion
   if (pcp->release == NULL)
      Platform_setRelease();

   *string = pcp->release;
```

darwin/Platform.c
Explorer09: I think CoreFoundation is already mandatory for Darwin build for htop. So I think it wouldn't hurt by using their API rather than rolling our own XML parser.

darwin/Platform.c
Explorer09: Note my opinion in [this comment](https://github.com/htop-dev/htop/pull/1812#issuecomment-3515676005). I think with this extra parser the `--with-os-release` configure option should be killed.

natoscott: @BenBE it is indeed, I kept this macro handling in the code just like it is for the other platforms today.  For the configure script, TBH I don't think "without-os-release" makes alot of sense (much like today).

@Explorer09 I'm happy to keep the --with-os-release configure option, I find it handy to see where this info is found for each platform, and it can be useful for testing different file content handling.

darwin/Platform.c
natoscott: @BenBE good idea - I'll take a look.

generic/uname.c
natoscott: Yep, OK, will adjust that, thanks.

generic/uname.h
natoscott: 👍

darwin/Platform.c
Explorer09: Rather than reading an AI-generated example code, I cite an example code from Apple documentation, which is more helpful: https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFPropertyLists/Articles/Saving.html#//apple_ref/doc/uid/20001175-CJBEHAAG

natoscott: @BenBE that code works but gives a deprecation warning ... I'll see if I can find a replacement.
```
x.c:27:22: warning: 'CFURLCreateDataAndPropertiesFromResource' is deprecated: first deprecated in macOS 10.9 - For resource data, use the CFReadStream API. For file resource properties, use CFURLCopyResourcePropertiesForKeys. [-Wdeprecated-declarations]
   27 |     Boolean status = CFURLCreateDataAndPropertiesFromResource(
      |                      ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/CoreFoundation.framework/Headers/CFURLAccess.h:52:9: note: 'CFURLCreateDataAndPropertiesFromResource' has been explicitly marked deprecated here
   52 | Boolean CFURLCreateDataAndPropertiesFromResource(CFAllocatorRef alloc, CFURLRef url, CFDataRef *resourceData, CFDictionaryRef *properties, CFArrayRef desiredProperties, SInt32 *errorCode) API_DEPRECATED("For resource data, use the CFReadStream API. For file resource properties, use CFURLCopyResourcePropertiesForKeys.", macos(10.0,10.9), ios(2.0,7.0), watchos(2.0,2.0), tvos(9.0,9.0));
      |         ^
1 warning generated.
```

darwin/Platform.c
Explorer09: @natoscott My intent is to remove user confusion when the `--with-os-release` option is there. If you do need an override, `CFLAGS='-DOSRELEASEFILE="/foo/os-release"'` is still available (I'm not suggesting the removal of that macro token).

The configure option is badly designed - first it didn't handle what to do `--without-os-release`, second it checks the file in the build machine even when cross compiling (one of the issues that I tried to fix in #1808). Needless to say that the paths `/etc/os-release` and `/usr/lib/os-release` are meant to be hard-coded according to the spec. Distros should not use other locations for the file.

natoscott:
> I have one concern here:
> [musings about /etc/os-versions as a standard]

We need to work with the tools available and for all existing macOS releases (and TBH likely all future releases too), there is no /etc/os-versions file.  This is true of almost all metrics we gather from different platforms, there are many different interfaces for the same things across platforms.  So, I'm happy enough with the solution here - its the best we can do under the circumstances IMO.

darwin/Platform.c
natoscott: OK, I agree - I've added a follow-up commit that removes the configure option entirely.

darwin/Platform.c
BenBE: @Explorer09 That's why I added this note; to give a broad direction, fully suspecting, there will be issues.

generic/uname.h
```diff
@@ -7,6 +7,12 @@ Released under the GNU GPLv2+, see the COPYING file
 in the source distribution for its full text.
 */
 
+#include <string.h>
```
BenBE:
```suggestion
#include <stddef.h>
```

natoscott: Done, thanks.

BenBE (approved):
LGTM.

generic/uname.c
```diff
@@ -21,12 +21,13 @@ in the source distribution for its full text.
 #endif
 
 
-#ifndef OSRELEASEFILE
-#define OSRELEASEFILE "/etc/os-release"
-#endif
-
 static void parseOSRelease(char* buffer, size_t bufferLen) {
-   FILE* fp = fopen(OSRELEASEFILE, "r");
+   const char* osfiles[] = { "/etc/os-release", "/usr/lib/os-release" };
```
Explorer09, line 25: A minor deal: If you really have a need for a custom file path to the `os-release` file for testing, this is the idea I can suggest to you:

```suggestion
   const char* osfiles[] = {
#ifdef OSRELEASEFILE
      /* Custom path for testing only; undefined by default */
      OSRELEASEFILE,
#endif
      "/etc/os-release",
      "/usr/lib/os-release",
   };
```

And then builders can use `"-DOSRELEASEFILE=\"foo\""` in `CFLAGS` to use a custom path they want. This is a configuration that's so advanced that I, again, vote against having a `--with-` configure option for.

EDIT: I mean, I can propose this in a PR. So that you guys can still have the custom configuration for the testing use case. It's just whether this PR will be worth it.

PR #1813, opened 2025-11-11
SysArchMeter & Platform_getRelease code improvements

Explorer09:
<s>This is cherry-picked and reworked from #1812 so I hope it can be merged earlier than the Darwin release code thing.</s>

<s>I attribute natoscott as the main author - I hope he doesn't mind.</s>

Update: I reworked this as a follow-up of #1812.

These are improvements to the `Platform_getRelease` code and APIs. Notably I rewrote a major part of Darwin `Platform_getOSRelease` function, so that the API calls are simplified (better than the previous "generated" version).

### DIFF
--- a/darwin/Platform.c
+++ b/darwin/Platform.c
@@ -792,7 +792,7 @@ static void Platform_getOSRelease(char* buffer, size_t bufferLen) {
       return;
 
 fail:
-   xSnprintf(buffer, bufferLen, "No OS Release");
+   buffer[0] = '\0';
 }
 
 const char* Platform_getRelease(void) {
--- a/generic/uname.c
+++ b/generic/uname.c
@@ -22,14 +22,25 @@ in the source distribution for its full text.
 
 
 static void parseOSRelease(char* buffer, size_t bufferLen) {
-   const char* osfiles[] = { "/etc/os-release", "/usr/lib/os-release" };
+   static const char* const osfiles[] = {
+#ifdef OSRELEASEFILE
+      OSRELEASEFILE /* Custom path for testing; undefined by default */,
+#endif
+      "/etc/os-release",
+      "/usr/lib/os-release",
+   };
+
+   if (!bufferLen)
+      return;
+
    FILE* fp = NULL;
-   for (size_t i = 0; i < sizeof(osfiles)/sizeof(char*); i++) {
-      if ((fp = fopen(osfiles[i], "r")) != NULL)
+   for (size_t i = 0; i < ARRAYSIZE(osfiles); i++) {
+      fp = fopen(osfiles[i], "r");
+      if (fp)
          break;
    }
    if (!fp) {
-      xSnprintf(buffer, bufferLen, "No OS Release");
+      buffer[0] = '\0';
       return;
    }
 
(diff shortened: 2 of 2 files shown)


BenBE:
Any particular reason to "rush" this part of the mentioned PR?

darwin/Platform.c
```diff
@@ -728,58 +728,61 @@ void Platform_gettime_monotonic(uint64_t* msec) {
...
 #define OSRELEASEFILE "/System/Library/CoreServices/SystemVersion.plist"
+#endif
 
 static void Platform_getOSRelease(char* buffer, size_t bufferLen) {
-   const UInt8* osfile = (const UInt8 *)OSRELEASEFILE;
-   const size_t length = sizeof(OSRELEASEFILE) - 1;
-   CFURLRef url = CFURLCreateFromFileSystemRepresentation(NULL, osfile, length, false);
-   if (!url) {
-      xSnprintf(buffer, bufferLen, "No OS Release");
+   if (bufferLen == 0)
```
```suggestion
   if (!bufferLen)
```

darwin/Platform.c
```diff
@@ -728,58 +728,61 @@ void Platform_gettime_monotonic(uint64_t* msec) {
...
+   if (CFGetTypeID(plist) == CFDictionaryGetTypeID()) {
+      CFDictionaryRef dict = (CFDictionaryRef)plist;
+      CFStringRef productName = CFDictionaryGetValue(dict, CFSTR("ProductName"));
+      CFStringRef productVersion = CFDictionaryGetValue(dict, CFSTR("ProductVersion"));
+      CFStringRef separator = productName && productVersion ? CFSTR(" ") : CFSTR("");
+      if (!productName)
+         productName = CFSTR("");
+      if (!productVersion)
+         productVersion = CFSTR("");
+      str = CFStringCreateWithFormat(kCFAllocatorDefault, NULL, CFSTR("%@%@%@"), productName, separator, productVersion);
```
BenBE, line 781:
```suggestion

      str = CFStringCreateWithFormat(kCFAllocatorDefault, NULL, CFSTR("%@%@%@"), productName, separator, productVersion);
```

BenBE (commented):
Please split the changes in the second commit between Darwin and the generic ones.

NB: That whether `OSRELEASEFILE` has a default or not depends on the platform isn't quite optimal.

darwin/Platform.c
```diff
@@ -728,56 +728,71 @@ void Platform_gettime_monotonic(uint64_t* msec) {
...
+
+      CFReadStreamRef stream = CFReadStreamCreateWithFile(kCFAllocatorDefault, url);
+      CFRelease(url);
+
+      if (!stream)
+         continue;
+
+      bool canRead = CFReadStreamOpen(stream);
+      if (canRead) {
+         plist = CFPropertyListCreateWithStream(kCFAllocatorDefault, stream, 0, kCFPropertyListImmutable, NULL, NULL);
```
Explorer09, line 756: Minor note: The old code had a read file size limit of 4096 bytes. I removed the limit here because I'm not sure whether keeping the size limit would make sense. (The third argument to [CFPropertyListCreateWithStream](https://developer.apple.com/documentation/corefoundation/cfpropertylistcreatewithstream(_:_:_:_:_:_:)?language=objc) is that size limit.)

BenBE: The htop internal APIs usually have that limit to avoid resource exhaustion issues. Assuming the CF stuff is half-way sane¹, there's little incentive to keep it.

¹I doubt it, it's Darwin after all …

PR #1814, opened 2025-11-12
NetBSD: Loop through kernel times rather than calling them one after the other

GuillaumeGomez:
Instead of naming each field we want to set, we simply go through the array (just like how it's done two lines below).

Not too sure if it was done this way on purpose or not, so just in case, here is a PR.

### DIFF
--- a/netbsd/NetBSDMachine.c
+++ b/netbsd/NetBSDMachine.c
@@ -205,11 +205,9 @@ static void NetBSDMachine_scanCPUTime(NetBSDMachine* this) {
       CPUData* cpu = &this->cpuData[i + 1];
       kernelCPUTimesToHtop(kernelTimes, cpu);
 
-      avg[CP_USER] += cpu->userTime;
-      avg[CP_NICE] += cpu->niceTime;
-      avg[CP_SYS] += cpu->sysTime;
-      avg[CP_INTR] += cpu->intrTime;
-      avg[CP_IDLE] += cpu->idleTime;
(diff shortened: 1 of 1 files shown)


netbsd/NetBSDMachine.c
```diff
@@ -205,11 +205,9 @@ static void NetBSDMachine_scanCPUTime(NetBSDMachine* this) {
...
       kernelCPUTimesToHtop(kernelTimes, cpu);
 
-      avg[CP_USER] += cpu->userTime;
-      avg[CP_NICE] += cpu->niceTime;
-      avg[CP_SYS] += cpu->sysTime;
-      avg[CP_INTR] += cpu->intrTime;
-      avg[CP_IDLE] += cpu->idleTime;
+      for (int j = 0; j < CPUSTATES; j++) {
+         avg[j] += kernelTimes[j];
+      }
```
BenBE, line 210: These are not equivalent. Take a look at `kernelCPUTimesToHtop` just above this function. The values in `cpu` are partially updated compared to the values present in `kernelTimes`.

And while the major values that actually differ are skipped in the structure assignment, this is a potential pitfall for introducing errors down the line.

Also note, modern compilers will usually optimize multiple stores and coalesce them into one `memcpy` where applicable.

GuillaumeGomez: Oh indeed, missed that. Thanks for the clarification!

GuillaumeGomez:
Closing as the change is wrong.

Issue #956, opened 2022-03-15
The Mem bar's ANSI escape color code bleeds into percentage


Integralist:
<img width="166" alt="Screenshot 2022-03-15 at 13 05 01" src="https://user-images.githubusercontent.com/180050/158384177-2fea73cd-8226-476f-8bdf-eea3252830b2.png">

In the above screenshot (taken on macOS Monterey `12.2.1` with htop version `3.1.2`) we can see the yellow colour is bleeding into the printed percentages.

PR #1818, opened 2025-11-14
Fix bar meter layout to not color the right-adjusted text

yinghao-w:
Small change to the bar meter drawing so that when the bar is long, the text on the right isn't partially colored by the bar coloring. Also adjusts the bar color proportions to be more accurate.

### DIFF
--- a/Meter.c
+++ b/Meter.c
@@ -156,8 +156,8 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
       double value = this->values[i];
       if (isPositive(value) && this->total > 0.0) {
          value = MINIMUM(value, this->total);
-         blockSizes[i] = ceil((value / this->total) * w);
-         blockSizes[i] = MINIMUM(blockSizes[i], w - offset);
+         blockSizes[i] = ceil((value / this->total) * (w - startPos));
+         blockSizes[i] = MINIMUM(blockSizes[i], w - startPos - offset);
       } else {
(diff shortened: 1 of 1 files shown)


BenBE:
Can you provide a screenshot of before and after? TIA.

Explorer09:
This is in [htop's FAQ](https://htop.dev/faq.html):

> The numbers in the meters occasionally change colors when the bar is full. Is this a bug?
>
> No, that's the expected behavior. :) It's like when progress bars fill over the number on GUI apps.

So what you're trying to fix is a non-bug.

And there have been issue reports about this (#956, closed as won't fix).

By the way, even though I personally would want that as a toggleable option, I don't think your proposed fix is the right way. The text on the bar can have variable width, so visually the length of representing 100% would be unstable if you fix it that way.

BenBE:
> This is in [htop's FAQ](https://htop.dev/faq.html):
> 
> > The numbers in the meters occasionally change colors when the bar is full. Is this a bug?
> > No, that's the expected behavior. :) It's like when progress bars fill over the number on GUI apps.
> 
> So what you're trying to fix is a non-bug.

Exactly. This is intended behavior.

> And there have been issue reports about this (#956, closed as won't fix).

Okay, closing as duplicate then …

> The text on the bar can have variable width, so visually the length of representing 100% would be unstable if you fix it that way.

Exactly. And that's why it will stay as it is.

PR #1822, opened 2025-11-23
ci: Bump BSD versions in ci.

fraggerfox:
- Bump FreeBSD version to the latest supported `14.3`
- Bump DragonFlyBSD version to the latest supported `6.4.2`
- Bump OpenBSD version to the latest supported `7.8`

Release notes can be found here: 
- https://www.freebsd.org/releases/14.3R/relnotes/
- https://www.dragonflybsd.org/release64/
- https://www.openbsd.org/78.html

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -247,7 +247,7 @@ jobs:
         - name: Compile
           uses: vmactions/dragonflybsd-vm@v1
           with:
-            release: '6.4.0'
+            release: '6.4.2'
             usesh: true
             prepare: |
               pkg install -y gmake autoconf automake ncurses git
@@ -270,7 +270,7 @@ jobs:
         - name: Compile
           uses: vmactions/freebsd-vm@v1
           with:
-            release: '14.2'
+            release: '14.3'
             usesh: true
             prepare: |
               pkg install -y gmake autoconf automake pkgconf git
@@ -316,7 +316,7 @@ jobs:
         - name: Compile
           uses: vmactions/openbsd-vm@v1
           with:
-            release: '7.7'
+            release: '7.8'
             usesh: true
             prepare: |
               pkg_add gmake git


openbsd/OpenBSDProcessTable.c
```diff
@@ -200,7 +200,7 @@ static void OpenBSDProcessTable_scanProcs(OpenBSDProcessTable* this) {
       proc->nice = kproc->p_nice - 20;
       proc->time = 100 * (kproc->p_rtime_sec + ((kproc->p_rtime_usec + 500000) / 1000000));
       proc->priority = kproc->p_priority - PZERO;
-      proc->processor = kproc->p_cpuid;
+      proc->processor = (int) kproc->p_cpuid;
```
fraggerfox, line 203: Based on this https://github.com/openbsd/src/blob/master/sys/sys/sysctl.h , I think this conversion should work?

openbsd/OpenBSDMachine.h
```diff
@@ -45,8 +45,8 @@ typedef struct OpenBSDMachine_ {
 
    long fscale;
    int cpuSpeed;
-   int pageSize;
-   int pageSizeKB;
+   size_t pageSize;
```
fraggerfox, line 48: Not sure if making them `size_t` is the correct fix here.

openbsd/OpenBSDMachine.c
```diff
@@ -105,8 +106,9 @@ Machine* Machine_new(UsersTable* usersTable, uid_t userId) {
       CRT_fatalError("fscale sysctl call failed");
    }
 
-   if ((this->pageSize = sysconf(_SC_PAGESIZE)) == -1)
+   if ((pageSize = sysconf(_SC_PAGESIZE)) == -1)
       CRT_fatalError("pagesize sysconf call failed");
+   this->pageSize = (size_t) pageSize;
```
fraggerfox: Converting `pageSize` into `size_t` helped resolve this.

```
  openbsd/OpenBSDMachine.c:108:26: error: implicit conversion loses integer precision: 'long' to 'int' [-Werror,-Wshorten-64-to-32]
    108 |    if ((this->pageSize = sysconf(_SC_PAGESIZE)) == -1)
        |                        ~ ^~~~~~~~~~~~~~~~~~~~~
  1 error generated.
```

Scheduling.c
```diff
@@ -53,7 +53,7 @@ Panel* Scheduling_newPolicyPanel(int preSelectedPolicy) {
 
       Panel_add(this, (Object*) ListItem_new(policies[i].name, policies[i].id));
       if (policies[i].id == preSelectedPolicy)
-         Panel_setSelected(this, i);
+         Panel_setSelected(this, (int) i);
```
fraggerfox, line 56:
```
   Scheduling.c:56:34: error: implicit conversion loses integer precision: 'size_t' (aka 'unsigned long') to 'int' [-Werror,-Wshorten-64-to-32]
     56 |          Panel_setSelected(this, i);
        |          ~~~~~~~~~~~~~~~~~       ^
```

openbsd/OpenBSDMachine.h
BenBE: That's literally what size_t is for. ;-) Denoting the number of elements that can fit into memory.

openbsd/OpenBSDMachine.c
BenBE: Yes, the right size is probably int or long, but needs to be `size_t`:

```c
if ((this->pageSize = (size_t)sysconf(…)) == (size_t)-1) …
```

Scheduling.c
BenBE: This will have to do for now; The internal API is scheduled to get bumped up to `size_t`; that's what the various `Vector` class refactorings are ultimately aimed at.

BenBE:
Reorder the commits to first fix the issues and then bump the versions:
- Merge the int size fixes (commits 3+4)
- Do the OpenBSD fixes (may also be squashed into the first step)
- Bump the versions

Apart from this LGTM.

openbsd/OpenBSDMachine.c
Explorer09: Move the `pageSize` declaration to just before the `sysconf` call. And if we're casting it to `size_t`, filter out all negative values. (EDIT: Also filter out 0 as that value won't make sense.)

```c
   long pageSize = sysconf(_SC_PAGESIZE);
   if (pageSize <= 0)
      CRT_fatalError("pagesize sysconf call failed");
   this->pageSize = (size_t) pageSize;
```

openbsd/OpenBSDProcessTable.c
natoscott: Looks fine for now, I don't know of any machines with MAXINT CPUs today or in the near future.

openbsd/OpenBSDMachine.c
fraggerfox: I replaced it with the one-liner suggestion from @BenBE

openbsd/OpenBSDProcessTable.c
BenBE: Wondering when loongarch will reach MAXLONG CPUs … /jk

BenBE (approved):

openbsd/OpenBSDMachine.c
Explorer09: I re-proposed this in #1827.
(No, I don't think the one-liner is good for what I was trying to solve. My aim is to filter out all invalid values and make the signed `long` to unsigned `size_t` cast safe.)

PR #1827, opened 2025-11-28
Improve the page size retrieval code via sysconf(3)

Explorer09:
Machine code that retrieves the page size through
`sysconf(_SC_PAGESIZE)` now behaves consistently across platforms. The return values of 0 and negative are now treated as invalid. The `pageSize` members for these Machine objects are now all `size_t` type (the return value type of `sysconf()` is `long`).

### DIFF
--- a/linux/LinuxMachine.h
+++ b/linux/LinuxMachine.h
@@ -8,6 +8,7 @@ in the source distribution for its full text.
 */
 
 #include <stdbool.h>
+#include <stddef.h>
 
 #include "Machine.h"
 #include "linux/ZramStats.h"
@@ -67,8 +68,8 @@ typedef struct LinuxMachine_ {
    Machine super;
 
    long jiffies;
-   int pageSize;
-   int pageSizeKB;
+   size_t pageSize;
+   size_t pageSizeKB;
 
    /* see Linux kernel source for further detail, fs/proc/stat.c */
    unsigned int runningTasks;   /* procs_running from /proc/stat */
--- a/netbsd/NetBSDMachine.h
+++ b/netbsd/NetBSDMachine.h
@@ -12,7 +12,7 @@ in the source distribution for its full text.
 
 #include <kvm.h>
 #include <stdbool.h>
-#include <sys/types.h>
+#include <stddef.h>
 
 #include "Machine.h"
 #include "ProcessTable.h"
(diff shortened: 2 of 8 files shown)


linux/LinuxMachine.h
```diff
@@ -8,6 +8,7 @@ in the source distribution for its full text.
 */
 
 #include <stdbool.h>
+#include <sys/types.h>
```
BenBE: Use `stddef.h` for `size_t`.

netbsd/NetBSDMachine.h
```diff
@@ -45,8 +45,8 @@ typedef struct NetBSDMachine_ {
    kvm_t* kd;
 
    long fscale;
-   int pageSize;
-   int pageSizeKB;
+   size_t pageSize;
+   size_t pageSizeKB;
```
BenBE, line 49: Missing `stddef.h`

solaris/SolarisMachine.h
```diff
@@ -49,8 +50,8 @@ typedef struct SolarisMachine_ {
    kstat_ctl_t* kd;
    CPUData* cpus;
 
-   int pageSize;
-   int pageSizeKB;
+   size_t pageSize;
+   size_t pageSizeKB;
```
BenBE, line 54: Use `stddef.h`

BenBE (changes requested):
Having this in a separate PR from #1822 allows for properly cleaning up this code. Also prefer to use `stddef.h` for `size_t`. Cf. [cpp-reference](https://en.cppreference.com/w/c/types/size_t.html). C99 notes the default declaration to live in `stddef.h`; thus that's what we should use, even if other headers also define it (and `sys/types.h` is not among them canonically).

Explorer09:
> Having this in a separate PR from #1822 allows for properly cleaning up this code. Also prefer to use `stddef.h` for `size_t`. Cf. [cpp-reference](https://en.cppreference.com/w/c/types/size_t.html). C99 notes the default declaration to live in `stddef.h`; thus that's what we should use, even if other headers also define it (and `sys/types.h` is not among them canonically).

`size_t` is also defined in `sys/types.h` according to POSIX. So I was avoiding including an additional header when I wrote this patch. Although including `stddef.h` wouldn't hurt anyway...

BenBE (approved):

PR #1828, opened 2025-11-29
[feat] Add basic support for iOS 6.1

tslmy:
Tested working on an iPod touch (4th gen) running iOS 6.1.6:

| process information | IO information |
| ---- | ---- |
| image | image | 

# How to build

Prerequisites:
- macOS with Homebrew. Tested on macOS 15.7.2 (24G325).
- Xcode 4.6.3 DMG mounted at `/Volumes/Xcode`.
- automake (`brew install automake`)
- ncurses (`brew install ncurses`)

## Prepare the SDK

`htop` needs the IOKit framework for probing battery, disk I/O, and GPU statuses. The vanilla iPhoneOS6.1 SDK does not provide headers for IOKit, so we'll have to stitch together a Frankenstein-style SDK for getting around it.

Assuming you are at a directory outside of `htop` codebase:

```sh
set -gx FOO "/Volumes/Xcode/Xcode.app/Contents/Developer/Platforms"
cp -r $FOO/iPhoneOS.platform/Developer/SDKs/iPhoneOS6.1.sdk .
cp -r $FOO/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator6.1.sdk/System/Library/Frameworks/IOKit.framework/Versions/A/Headers iPhoneOS6.1.sdk/System/Library/Frameworks/IOKit.framework/
cp $FOO/MacOSX.platform/Developer/SDKs/MacOSX10.7.sdk/System/Library/Frameworks/IOKit.framework/Versions/A/Headers/storage/IOBlockStorageDriver.h iPhoneOS6.1.sdk/System/Library/Frameworks/IOKit.framework/Headers/storage/
```

Don't unmount the Xcode DMG just yet; we still need the simulator's SDK (for `libproc.h`).

## Build

Assuming you are now in the `htop` directory, here's the full commands needed to build:

```sh
set -gx IOS6_SDK_PATH "/path/to/the/frankenstein/sdk/above/iPhoneOS6.1.sdk"

./autogen.sh
./configure \
  --host=armv7-apple-darwin \
  --disable-unicode \
  --disable-sensors \
  --disable-hwloc \
  CFLAGS="-arch armv7 \
        -miphoneos-version-min=6.1 \
        -isysroot $IOS6_SDK_PATH \
        -I/Volumes/Xcode/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator6.1.sdk/usr/include \
        -I/opt/homebrew/opt/ncurses/include"
make
```

### DIFF
--- a/darwin/Platform.c
+++ b/darwin/Platform.c
@@ -15,11 +15,11 @@ in the source distribution for its full text.
 #include <math.h>
 #include <stdlib.h>
 #include <unistd.h>
-#include <net/if.h>
 #include <net/if_types.h>
 #include <net/route.h>
 #include <sys/socket.h>
 #include <mach/port.h>
+#include <net/if.h> // After `sys/socket.h` for struct `sockaddr` (for iOS6 SDK)
 
 #include <CoreFoundation/CFBase.h>
 #include <CoreFoundation/CFDictionary.h>


Explorer09:
While this thing might work, the fact that it's generated suggests we cannot merge it as is. Because there's no human review, and there is a copyright concern (the AI might have copied code from someone else that is unlicensed).

Putting the general issue of AI-generated code aside, i see there's a massive style violation. We're not defining or using a macro token like `TARGET_OS_IPHONE`. Besides, the missing libraries and headers in iPhone should be addressed with `configure` script macros. So a `AC_CHECK_HEADERS(libproc.h)` call and a substitute header `ProvideLibproc.h` might be a good way to start.

darwin/DarwinProcess.c
```diff
@@ -9,13 +9,63 @@ in the source distribution for its full text.
...
 #include <sys/dirent.h>
 
+/* iOS 6.1 SDK doesn't have libproc.h, define stubs */
+#if TARGET_OS_IPHONE
+#define PROC_PIDVNODEPATHINFO 9
+#define PROC_PIDTASKINFO 4
+#define PROC_PIDTASKINFO_SIZE sizeof(struct proc_taskinfo)
+#define PROC_PIDPATHINFO_MAXSIZE 4096
+
+struct proc_taskinfo {
```
Explorer09: This structure's definition is far beyond being a stub. That's my concern when it comes to AI-generated code. Especially copyright issues.

darwin/DarwinProcess.c
```diff
@@ -9,13 +9,63 @@ in the source distribution for its full text.
...
+
+static inline int proc_pidpath(int pid, void *buffer, uint32_t buffersize) {
+   (void)pid; (void)buffer; (void)buffersize;
+   return -1;
+}
+
+static inline int proc_pidinfo(int pid, int flavor, uint64_t arg, void *buffer, int buffersize) {
+   (void)pid; (void)flavor; (void)arg; (void)buffer; (void)buffersize;
+   return -1;
+}
```
BenBE: If these are implemented as stubs, you shouldn't be defining the PROC_EXE and CWD columns for the platform either. And as those two columns are kinda essential these information are kinda essential for an implementation to provide.

darwin/DarwinProcess.c
```diff
@@ -9,13 +9,63 @@ in the source distribution for its full text.
 
 #include "darwin/DarwinProcess.h"
 
+#if !TARGET_OS_IPHONE
```
BenBE: As @Explorer09 mentioned: This should be detected via `configure`.

Also with headers that need defines for their inclusion, the inclusion order changes slightly (cf. style guide).

darwin/DarwinProcess.c
BenBE: If this structure is public knowledge, it's lacking a source reference from where it's copied.

BUT: As this structure is very likely to have changed over the years, you'd have to implement this in a way that tracks those changes too.

Also, you're missing the include of `stdint.h` for the fixed-width integer types used below.

darwin/Platform.c
```diff
@@ -506,7 +536,11 @@ void Platform_getFileDescriptors(double* used, double* max) {
 }
 
 bool Platform_getDiskIO(DiskIOData* data) {
-
+#if TARGET_OS_IPHONE
```
BenBE: Should be checked using `configure`.

darwin/DarwinMachine.c
```diff
@@ -105,18 +105,22 @@ Machine* Machine_new(UsersTable* usersTable, uid_t userId) {
...
+#endif
 
    return super;
 }
 
 void Machine_delete(Machine* super) {
    DarwinMachine* this = (DarwinMachine*) super;
 
+#if !TARGET_OS_IPHONE
    IOObjectRelease(this->GPUService);
```
BenBE, line 124: The original line here is missing a NULL check, which the initialization code above includes.

tslmy:
Thank you @BenBE and @Explorer09 for the very detailed reviews! Let me address them and get back to you. Putting this to `draft` mode while I'm at it.

Is there unit tests that I can run?

darwin/DarwinMachine.c
```diff
@@ -105,18 +105,22 @@ Machine* Machine_new(UsersTable* usersTable, uid_t userId) {
    openzfs_sysctl_init(&this->zfs);
    openzfs_sysctl_updateArcStats(&this->zfs);
 
+#if !TARGET_OS_IPHONE
    this->GPUService = IOServiceGetMatchingService(kIOMainPortDefault, IOServiceMatching("IOGPU"));
    if (!this->GPUService) {
       CRT_debug("Cannot initialize IOGPU service");
    }
+#endif
```
Explorer09: When compiling out this code, please at least set `this->GPUService` to a dummy value like NULL.

darwin/Platform.c
```diff
@@ -15,23 +15,34 @@ in the source distribution for its full text.
 #include <math.h>
 #include <stdlib.h>
 #include <unistd.h>
-#include <net/if.h>
-#include <net/if_types.h>
-#include <net/route.h>
 #include <sys/socket.h>
+#include <sys/sysctl.h>
+#include <net/if.h>
```
Explorer09: Don't change header order without reason. (Please refer to the style guide.)

darwin/Platform.c
```diff
@@ -656,6 +691,7 @@ void Platform_getBattery(double* percent, ACPresence* isOnAC) {
    *percent = NAN;
    *isOnAC = AC_ERROR;
 
+#if !TARGET_OS_IPHONE
```
Explorer09: Why comment this block out?
People surely want to read batteries of their iPhones, no?
Better provide a proper implementation than leave a dummy code like that.

darwin/Platform.c
```diff
@@ -403,16 +423,26 @@ void Platform_setMemoryValues(Meter* mtr) {
...
+#if !TARGET_OS_IPHONE  /* iOS 6.1 SDK doesn't have compressor_page_count and external_page_count */
    natural_t used = vm->active_count + vm->inactive_count +
               vm->speculative_count + vm->wire_count +
               vm->compressor_page_count - vm->purgeable_count - vm->external_page_count;
    mtr->values[MEMORY_METER_USED] = (double)(used - vm->compressor_page_count) * page_K;
+#else
+   natural_t used = vm->active_count + vm->inactive_count +
+              vm->speculative_count + vm->wire_count - vm->purgeable_count;
+   mtr->values[MEMORY_METER_USED] = (double)used * page_K;
+#endif
```
Explorer09: Nonexistent fields should be better checked in `configure`.

darwin/DarwinMachine.c
BenBE: If I see this correctly, `GPUService` is not even declared for iPhone 6 with this PR.

darwin/Platform.c
BenBE: That said, the original ordering had its issues too …

darwin/Platform.c
BenBE: This overall nesting of ifdef checks is hard to red overall. A better approach might be needed.

darwin/DarwinMachine.c
tslmy: Sure, I can patch that. ~~It also removes the need of using a `#ifndef`, which I like anyway.~~ No it doesn't. I should have known my code.

darwin/DarwinProcess.c
tslmy: Hi @BenBE , please review the commit 7891bc3fe4e55dff1340804e7b43c14cff130e40. Is the design OK to you?

darwin/Platform.c
tslmy: Ditto. I'll resolve this one comment thread to reduce duplicate work.

darwin/DarwinMachine.c
tslmy: Yeah... Which would you all prefer -- leaving the field in the struct and provide a null check everywhere it's used, or don't compile it to the struct entirely (current implementation)?

configure.ac
```diff
@@ -56,6 +67,14 @@ darwin*)
...
    AC_DEFINE([HTOP_DARWIN], [], [Building for Darwin.])
    ;;
+ios6*)
+   # iOS 6 is really a special case of Darwin. We want to reuse existing logic for Darwin as much as
+   # possible, so we are not giving `my_htop_platform` a different value.
+   my_htop_platform=darwin
+   htop_ios6_platform=yes
+   AC_DEFINE([HTOP_DARWIN], [], [Building for Darwin.])
+   AC_DEFINE([HTOP_DARWIN_IOS6], [], [Building for Darwin (iOS 6).])
+   ;;
```
Explorer09: I'm not impressed with this. You surely have no idea of how Autoconf-powered `configure` works.
If it's merely about workarounds of certain missing headers or libraries, then you should check those headers/libraries individually and use the relevant macros for those headers/libraries.

Otherwise we could end up unmaintainable code where there are macros of iOS 6, iOS 7, iOS N... everywhere.

darwin/Platform.c
tslmy: Yeahhh.. I also would love to implement this. Battery information is read via:

```c
#include <IOKit/IOKitLib.h>
#include <IOKit/IOTypes.h>
#include <IOKit/ps/IOPowerSources.h>
```

but IOKit header files aren't available in the iOS 6.1 SDK:

```shell
$ ls /Volumes/Xcode/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS6.1.sdk/System/Library/Frameworks/IOKit.framework
IOKit*    Versions/
```

Apparently, it was still considered a private framework back in iOS 6.1.

Fortunately, Apple has open-sourced a version of those files. The most sensible workaround I could think of was to, in the build instructions, ask the developer to 

1. Clone this repo: https://github.com/opensource-apple/IOKitUser/blob/master/IOKitLib.h
2. point the compiler to it.

What do you all think?

configure.ac
tslmy: I see what you mean. You are saying that

1. you'd rather `my_htop_platform` stay generic at the Darwin level and not expose OS-version-specific macros, and that
2. I should figure out exactly what libraries / headers need special treatment for iOS 6 and just have flags for each of them.

Those make sense to me. Admittedly I only write C code at work and we use a very different build system than Autoconfig. Let me read around and see how that can be done.

darwin/DarwinProcess.c
tslmy: Sorry, could you elaborate? Not sure what `PROC_EXE` and `CWD` map to things in this stub here.

But yeah, I'm not trying to provide a fully working implementation just yet. This first commit is just enough to make things compile. Perhaps I should mark stubs with TODOs. What's a todo style that we use here in this codebase?

configure.ac
tslmy: Done! Please take a look. `AM_CONDITIONAL` and `AC_CHECK_HEADER` are fun.

darwin/Platform.c
tslmy: ~~Reverted. Thanks for catching that. I'll also move the #defines below #includes while I'm at it.~~

See below. Turns out we still need at least one moving-around.

darwin/DarwinMachine.c
BenBE: Given the original source in the cleanup function is wrong currently too, I'd much prefer to leave the field in and correct the existing code. That way we're not un-defining and re-defining fields based on whether we believe them to be present or not and instead deferring these checks to runtime, where they can be decided much more comprehensively. AFAICT the type used for the field should be available even when compiling for the iPhone 6.

darwin/Platform.c
BenBE: I think this can be best resolved with a configure option on Darwin like `--with-iokit-fallback=/path/to/IOKit` with a description providing the instructions to pass this option when dealing with these very old environments.

Note that this is intentionally framed as a fallback, i.e. the configure checks should try to establish the IOKit header locations from the normal Apple system headers first, and only rely on the path in this option IFF they otherwise cannot be found. That said, this _should not_ interfere with builds for plain darwin.

BenBE:
Please note the instruction regarding rebasing/amending your commits for PRs in the style guide.

darwin/Platform.c
Explorer09: @BenBE Can't the regular `-I` include path options (that user can add in `CPPFLAGS`) already cover that?

configure.ac
```diff
@@ -239,9 +220,48 @@ m4_version_prereq(
 
 AC_CHECK_HEADERS([execinfo.h])
 
+htop_have_iokit=no
+htop_have_libproc=no
 if test "$my_htop_platform" = darwin; then
    AC_CHECK_HEADERS([mach/mach_time.h])
    AC_CHECK_TYPES([thread_extended_info_data_t], [], [], [[#include <mach/thread_info.h>]])
+   AC_CHECK_HEADERS([net/if_types.h])
+   AC_CHECK_HEADERS([net/route.h])
```
Explorer09: Can be merged to one call along with `mach/mach_time.h` above.

```m4
   AC_CHECK_HEADERS([ \
      mach/mach_time.h \
      net/if_types.h \
      net/route.h \
   ])
```

configure.ac
```diff
@@ -239,9 +220,48 @@ m4_version_prereq(
...
+      if test "x$htop_have_libproc" = xyes; then
+         AC_DEFINE([HAVE_LIBPROC], [1], [Define if libproc APIs are available.])
+      fi
+   fi
+
+   htop_have_iokit=yes
+   AC_CHECK_HEADER([IOKit/IOKitLib.h], [], [htop_have_iokit=no])
+   AC_CHECK_HEADER([IOKit/IOTypes.h], [], [htop_have_iokit=no])
+   AC_CHECK_HEADER([IOKit/ps/IOPowerSources.h], [], [htop_have_iokit=no])
+   AC_CHECK_HEADER([IOKit/ps/IOPSKeys.h], [], [htop_have_iokit=no])
```
Explorer09: Just check for one header `IOKit/IOKitLib.h` unless you want to fine-tune what components of IOKit are supported.

configure.ac
```diff
@@ -239,9 +220,48 @@ m4_version_prereq(
...
+   AC_CHECK_HEADER([IOKit/storage/IOBlockStorageDriver.h], [], [htop_have_iokit=no])
+   if test "x$htop_have_iokit" = xyes; then
+      AC_MSG_CHECKING([for usable IOKit framework])
+      save_LIBS="$LIBS"
+      LIBS="-framework IOKit -framework CoreFoundation $LIBS"
+      AC_LINK_IFELSE(
+         [AC_LANG_PROGRAM([[#include <IOKit/IOKitLib.h>]], [[return IOServiceMatching("IOGPU") == NULL;]])],
+         [have_iokit_link=yes],
+         [have_iokit_link=no]
+      )
```
Explorer09: Why do we need this?

Makefile.am
```diff
@@ -382,7 +382,10 @@ darwin_platform_sources = \
 	zfs/ZfsCompressedArcMeter.c
 
 if HTOP_DARWIN
-AM_LDFLAGS += -framework IOKit -framework CoreFoundation
+AM_LDFLAGS += -framework CoreFoundation
+if HAVE_IOKIT
+AM_LDFLAGS += -framework IOKit
+endif
```
Explorer09: Don't add `AM_LDFLAGS` here. It's more ideal to add that in `configure`.

darwin/Platform.c
```diff
@@ -84,6 +72,17 @@ in the source distribution for its full text.
...
+#ifndef IFT_LOOP
+#define IFT_LOOP 24 /* Software loopback network */
+#endif
+#endif
+
+#ifndef HAVE_NET_ROUTE_H
+#ifndef RTM_IFINFO2
+#define RTM_IFINFO2 0x12
+#endif
+#endif
```
Explorer09: Need explanation on why we define the macro constants to the values presented. If a system doesn't have the macro constants defined, it is likely that the properties associated to the constants are not applicable to the system, in which case we should compile out the relevant code rather than define the constants to the value that you (@tslmy) even don't know what they do.

configure.ac
```diff
@@ -1585,7 +1605,8 @@ AM_CONDITIONAL([HTOP_DRAGONFLYBSD], [test "$my_htop_platform" = dragonflybsd])
 AM_CONDITIONAL([HTOP_NETBSD], [test "$my_htop_platform" = netbsd])
 AM_CONDITIONAL([HTOP_OPENBSD], [test "$my_htop_platform" = openbsd])
 AM_CONDITIONAL([HTOP_DARWIN], [test "$my_htop_platform" = darwin])
-AM_CONDITIONAL([HTOP_DARWIN_IOS6], [test "$htop_ios6_platform" = yes])
+AM_CONDITIONAL([HAVE_IOKIT], [test "x$htop_have_iokit" = xyes])
+AM_CONDITIONAL([HAVE_LIBPROC], [test "x$htop_have_libproc" = xyes])
```
Explorer09: Use `AM_CONDITIONAL` sparingly. That macro is more useful for adding or removing what source files to compile within the `Makefile`. It's not meant for appending `AM_CFLAGS` or `AM_CPPFLAGS` or relevant build variables.

tslmy:
Still grinding on it! 💪

I think I’m getting somewhere pointing the compiler to a libproc from another SDK. If that works, we don’t need a code change, and it will just be a matter of documenting what build parameters to use.

Thank you for the continuous discussion today!

darwin/DarwinProcess.c
tslmy: To answer your question, this struct is adapted from https://github.com/apple-opensource/xnu/blob/master/libsyscall/wrappers/libproc/libproc.h . 

If we can afford to tell developers to bring their own local copy of `xnu`, then we can actually flesh out the feature for iOS 6.

I was able to make it work:

image

but it required a lot of compatibility code. I'm still looking for a more concise way to do this. I'll keep you all posted.

darwin/Platform.c
tslmy: We need to move it to avoid this error in iOS 6 SDK:

```
In file included from darwin/Platform.c:18:
/Volumes/Xcode/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS6.1.sdk/usr/include/net/if.h:307:19: error: field has incomplete type 'struct sockaddr'
  307 |                 struct  sockaddr ifru_addr;
      |                                  ^
```

configure.ac
tslmy: Done.

tslmy: (I'll resolve trivial threads like this to keep things tidy. If you prefer to only have reviewers resolve conversations, please reopen, and I'll keep it open.)

configure.ac
tslmy: Done. 

(I'll resolve trivial threads like this to keep things tidy. If you prefer to only have reviewers resolve conversations, please reopen, and I'll keep it open.)

configure.ac
tslmy: Not really. I was trying to get fancy. I decided to just disable IOKit if we don't see one file.

Makefile.am
tslmy: I'm not sure if I can make the call. The flag (which adds 2 frameworks) was already there; I'm merely guarding one framework with a `if`. 

If I have to move it, I prefer to move the flag for both frameworks. But I'm afraid that if I move both to `configure` today, I may cause problems beyond my comprehension. So I suggest we keep this as-is.

configure.ac
tslmy: Never mind about this. I removed this branch entirely.

configure.ac
tslmy: I deleted these lines. I'll resolve this conversation.

darwin/Platform.c
tslmy: Deleted. No longer applies. I'll resolve this conversation.

configure.ac
tslmy: Yeah. I just need one now.

darwin/Platform.c
BenBE: Yes. It's mostly about the semantics to only add this path using `-I` as a fallback. But the regular `-I` without that logic may use this path even when there's an IOKit from the regular framework. That'S the part that `configure` should check.

tslmy: I gave up. I'll just disable IOKit-related features when it's not available. Read my work log [here](https://gist.github.com/tslmy/984813b7d9ec4a9919928c305aa17583).

IOKit is much trickier than `libproc`, which has a somewhat reusable counterpart officially open-sourced (https://github.com/apple-opensource/xnu/blob/master/libsyscall/wrappers/libproc/libproc.h). I just had to implement some lightweight compatibility layer (`compat/`).

IOKit is also less mission-critical than `libproc`, since `htop` is an interactive **process** viewer. I'm not as motivated to fix IOKit, to be frank. I'll probably mark this as a to-do and work on it in a next weekend. Perhaps I'll try using macOS's IOKit headers for iOS 6, per suggestion [here](https://www.malcolmhall.com/2017/01/29/where-to-find-iokit-headers-for-ios/).

darwin/DarwinProcess.c
tslmy: No longer applicable. Fixed.

darwin/DarwinProcess.c
tslmy: No longer applicable.

darwin/Platform.c
BenBE: In that case switching the header order minimally (and documenting the reason) is the way to go:

```c
#include "problematic.h" // After "dependency.h" for typename (for iOS6 SDK)
```

darwin/DarwinMachine.c
tslmy: Can I shelf this for now? I'd prefer to spend my next hours on `htop` looking into how to get IOKit working (in another PR), instead of correct the existing code. Perhaps I'm not feeling ownership towards `htop` code quality just yet.

Makefile.am
BenBE: That's quite interesting as `IOKit` has been used in the code base previously, thus adding it to the link flags should have been present already. If not, then the configure was lacking these checks previously for the two existing ones, as direct usage of frameworks, just like other shared libraries, should have been checked in `configure`.

In this case, it's mostly a matter of checking "Has framework X be mentioned explicitly" in the configure and adding the proper flags for the compiler/linker.

tslmy: Ah, intriguing... The closest clue I can find is this line from the changelog:

```
Remove IOKit / IOMainPort / IOMMasterPort logic for Darwin builds
```

darwin/Platform.c
tslmy: You bet!

darwin/DarwinProcess.c
tslmy: After trying a bunch of tricks, introducing a compatibility layer is still the most sensible choice I got. It's not like that everyday people create a top-level directory in this codebase, so please let me know if I should move it somewhere else.

darwin/os/availability.h
BenBE, line 1: As this header is for interoperability of different versions of Darwin-based OS/SDK, this should best stay in the `darwin/` folder. For cross-platform compatibility there is the `generic/` folder, but that's aimed for functionality that's mostly identical across most platforms, but needs special handling on a few platforms.

Thus this header probably should just go to `darwin/sdk-compat.h` which furthermore also better hints at its purpose.

compat/libproc_preinclude.h
BenBE, line 1: Rename/move to `darwin/ProvideLibproc.h` and structure similar to the other existing `Provide*.h` headers. Note the various IWYU markers in the existing headers.

darwin/DarwinMachine.h
BenBE:
```suggestion
#include <mach/mach_host.h>
#include <sys/sysctl.h>

#include "Machine.h"
#include "zfs/ZfsArcStats.h"

#ifdef HAVE_IOKIT
#include <IOKit/IOKitLib.h>
#endif
```

Conditional includes go last. Think of it as three groups: system headers, project headers, conditional headers.

BenBE, line 17:
```suggestion

#if !defined(HAVE_IOKIT) || !HAVE_IOKIT
// Provide missing type for IOKit
typedef void* io_service_t;
#endif

```

Now everything refering to the `GPUService` field can stay as is; and only needs an NULL check added before use (which these accesses should have anyway, because loading the GPU service may fail at runtime.

darwin/DarwinMachine.h
```diff
@@ -27,7 +31,9 @@ typedef struct DarwinMachine_ {
    processor_cpu_load_info_t prev_load;
    processor_cpu_load_info_t curr_load;
 
+#ifdef HAVE_IOKIT
    io_service_t GPUService;
+#endif
```
BenBE: See remark above.

darwin/Platform.c
```diff
@@ -656,6 +681,9 @@ void Platform_getBattery(double* percent, ACPresence* isOnAC) {
    *percent = NAN;
    *isOnAC = AC_ERROR;
 
+#ifndef HAVE_IOKIT
+   return;
+#else
```
```suggestion
#ifdef HAVE_IOKIT
```

configure.ac
```diff
@@ -219,10 +219,32 @@ m4_version_prereq(
...
+   #
+   # To build for iOS 6, provide the following `CFLAGS`:
+   # * `-I$XNU_PATH/libsyscall/wrappers/libproc` (for access to `libproc.h` itself)
+   # * `-idirafter $XNU_PATH/bsd` (for access to `sys/proc_info.h`)
+   #
+   # where `$XNU_PATH` is the path to a checked-out copy of the XNU source code.
+   #
+   # At this point, we have just one missing item: `os/availability.h`. For that, we have provided
+   # a compatibility layer for XNU's `libproc.h` in `compat/libproc_preinclude.h`:
+   CPPFLAGS="$CPPFLAGS -I$srcdir/compat -include $srcdir/compat/libproc_preinclude.h"
```
BenBE: Cf. comments on the two files in `compat/` above and the note on better extracting things to some `ProvideLibproc.h` header in `darwin/` instead.

Makefile.am
```diff
@@ -16,7 +16,9 @@ EXTRA_DIST = \
 	build-aux/compile \
 	build-aux/depcomp \
 	build-aux/install-sh \
-	build-aux/missing
+	build-aux/missing \
+	compat/os/availability.h \
+	compat/libproc_preinclude.h
```
BenBE: These should be picket up automatically when properly used with the darwin platform.

darwin/os/availability.h
```diff
@@ -0,0 +1,67 @@
...
+// compiled against the ancient iOS 6 SDK.  The original header only provides
+// annotation macros, therefore this version keeps the interface surface small
+// and header-only by mapping all of the macros to no-ops when they are
+// otherwise unavailable.
+
+#ifndef HTOP_COMPAT_OS_AVAILABILITY_H
+#define HTOP_COMPAT_OS_AVAILABILITY_H
+
+#ifdef __has_include
+# if __has_include(<Availability.h>)
```
Explorer09: Not fitting the current htop coding style. You know, you should check for `Availability.h` in `configure`. (This is why I hate AI generated commits and this whole pull request.)

Explorer09:
The entire PR has become too messy for me to review further. There are still coding style violations here and there. But it seems like there are some compatibility fixes that can be cherry-picked early (such as the check of `struct vm_statistics64.compressor_page_count`).

BenBE:
Dear @Explorer09,

Please keep discussion in here civilized.

If you don't like AI-generated code: Fine.

If someone has no experience with a certain tool: Help them to get to know that tool or point them to resources that provide an introduction or explain necessary concepts.

If code doesn't follow the coding style: Explain the issue and provide possible alternatives.

BUT: This is no place for attacks or personal vendettas against any particular technology, just because someone chooses to use them.

This is voluntary work. Nobody forces you to respond to each and every issue/PR.

darwin/DarwinMachine.c
BenBE: Well, it's fine in your current code, as that field is not used for the iOS 6 SDK, but for normal MacOS X the code may try to release `this->GPUService`, which may or may not have been properly initialized (to non-NULL) in the accompanying initialization function. That's why I wrote "the original line" with emphasis on "original". ;-)

Because AFAIR, all those IOKit functions aren't NULL-safe. In particular the `*Release` functions often cause trouble when being passed NULL pointers.

tslmy:
I'm sorry you feel this way @Explorer09 . Please take a break and you're welcome back when you feel better.

To both of you @Explorer09 and @BenBE :

Thanks for the reviews again. Let me put this PR to draft mode again while I'm at it. (I might need to pick this up only at a next weekend, though.)

In general, I'm a bit **confused about the "code style"** that you all have been referring to. I thought those changes are applied automatically, yet I don't see a pre-commit hook or a CI job kicking in.

I've been treating the code style doc as "hey don't be surprised if we modify your code this way eventually / don't be offended when a script keeps formatting your code this way", but judging from your comments this isn't the case.

Are those code style rules to be followed manually?

darwin/Platform.c
tslmy: I'll resolve this one.

darwin/DarwinMachine.c
tslmy: Yup. Hope this null check I added will save someone some headache. (I'll resolve this discussion; we can revisit whether we want to fit a total null pointer exception clean-up across the whole repo to the roadmap later.)

Explorer09:
@tslmy 

> In general, I'm a bit **confused about the "code style"** that you all have been referring to. I thought those changes are applied automatically, yet I don't see a pre-commit hook or a CI job kicking in.
> 
> I've been treating the code style doc as "hey don't be surprised if we modify your code this way eventually / don't be offended when a script keeps formatting your code this way", but judging from your comments this isn't the case.
> 
> Are those code style rules to be followed manually?

Some code style cannot be automated prior to commit, and our attitude to pull requests is that: **it is the OP's responsibility to code in our style when proposing code changes**. If someone cannot bother to adapt to our code style, then it's better to maintain htop in their fork and not propose upstream.

Significant disruption to code style and/or paradigms can complicate other developers when reviewing your code. And note that I only review htop PRs on the voluntary basis - and it would be unpleasant for me if someone seems to be deliberately wasting my time.

tslmy:
@Explorer09 :

Sure, I'll comply my PR to the code style. It is going to take me some time, but that's standard work anywhere. No problem with that.

I don't believe anyone is interested in deliberately spending their time on wasting your time. Has that happened to you? Do you want to talk about it?

compat/libproc_preinclude.h
tslmy: Makes sense! Moved. (I'll resolve this.)

darwin/DarwinMachine.h
tslmy: Done.

Makefile.am
tslmy: Got it. Removed. (I'll resolve this.)

darwin/os/availability.h
tslmy: I tried moving `compat/os/availability.h` to `darwin/sdk-compat.h`, and then XNU started complaining not seeing `os/availability.h`:

```
In file included from darwin/DarwinProcess.c:12:
.../xnu/libsyscall/wrappers/libproc/libproc.h:39:10: fatal error: 'os/availability.h' file not found
   39 | #include <os/availability.h>
      |          ^~~~~~~~~~~~~~~~~~~
```

The best I can do is to move `compat/os/availability.h` to `darwin/os/availability.h`. Would that be OK?

darwin/os/availability.h
tslmy: You're right. I don't really need this, now that I can just include this conditionally in `configure`. I've deleted this and will resolve this conversation.

darwin/Platform.c
tslmy: Done.

darwin/DarwinMachine.h
tslmy: Good idea. I'll add this.

I also thought for a moment that you might appreciate adding:

```c
#if !defined(HAVE_IOKIT) || !HAVE_IOKIT
// Provide missing type for IOKit
typedef void* mach_port_t;
#endif
```

to `Platform.c`, but then `mach_port_t iokit_port` is only used at:

```c
if (IOServiceGetMatchingServices(iokit_port, IOServiceMatching("IOBlockStorageDriver"), &drive_list))
```

so it seems much cleaner to just disable the whole code path (current proposed change):

```c
#ifndef HAVE_IOKIT
   (void)data;
   return false;
#else
```

configure.ac
tslmy: Renamed. (I'll resolve this thread.)

darwin/DarwinMachine.h
tslmy: Removed.

Actually, I also removed two `#ifdef HAVE_IOKIT` fences in `DarwinMachine.c`, thinking that it's no much work to stub those functions. Apparently that's another rabbit hole. I decided to extract a `ProvideIOKit.h` (see the latest commit). What about the fences in `Platform.c`? Those are too much. I'm in a dilemma of whether I should just revert the two `#ifdef HAVE_IOKIT` fences back to `DarwinMachine.c` and leave the stub as lean as what you originally suggested -- `typedef void* io_service_t;`, nothing else.

What do you suggest?

BenBE:
> In general, I'm a bit **confused about the "code style"** that you all have been referring to. I thought those changes are applied automatically, yet I don't see a pre-commit hook or a CI job kicking in.
> 
> I've been treating the code style doc as "hey don't be surprised if we modify your code this way eventually / don't be offended when a script keeps formatting your code this way", but judging from your comments this isn't the case.
> 
> Are those code style rules to be followed manually?

The code style is not done automatically, as the code base has several aspects that would break any automatic formatting in a way, which causes unnecessary noise. One example is indentation of `case` labels inside `switch` statements, where you have several sections of the code that indent them, with other parts not indenting them. So the general consensus on the code style document is "write your code in such a way that someone looking at the changes would not notice them being made". That's also why the code style gives some specific hints on certain aspects (these are the ones that are fixed throughout), but is more open on other aspects. There is some rough command mentioned in the style guide that does the formatting approximately, but still you'd need to select only the parts that concern your patch.

Same goes for the header includes: As you noticed, there are some situations where the usual alphabetic order just wouldn't work. Thus automatic enforcement just wouldn't work. And minor deviations are acceptable if they are missed or justified. When writing a PR the style guide actually encourages you to fix minor style issues you notice along the way, like the missing NULL check or even the header ordering issue that was present even before your PR. The only thing the style guide asks in this case is to have these changes in separate commits. More on that in a moment.

As code style also includes other aspects like how you perform certain operations. For example with the `libproc` situation you have multiple ways to go, like distributing several `#ifdef`s all over the code for each site where you need access to that library, or you could do a central approach like has been done with e.g. `ProvideCurses.h` that streamlines the various ways that `ncurses` wants to be babysitted and just makes things "work"™. This latter part is where AI tools love to stray off-road any time you're not watching carefully, which is part of why they cause some friction on things like with the build system. (Well, `autotools` is horrible to work with, even if you know what you are doing, but getting reasonable results from AI tools there is like fighting two up-hill battles at once – one against `autotools` and one against the AI tool).

And no: There is no commit hook magically fixing things. Writing well structured code is the responsibility of the developer – even if they decide to support their work with tools like code formatters (like `astyle`) or AI agents.

The style guide is as the word implies a guide and it's intentionally written with a lot of room for things that are acceptable.

Okay, not for the commits: The usual PR development easily spirals into dozens of commits with advancing things one step and rewinding others several times in a row. This makes the (version control) history terrible to read if left untreated. That's why the style guide asks you to group your commits not by when you changed things, but by the aspects that you changed in order to arrive at your goal. So any PR that's like more than 3 to 5 commits usually needs a VERY good explanation why it splits things into so many steps. The largest PR I can remember was one where I cleaned up a cross-platform feature to align on all platforms. And that one had only about 30 commits, when intermediate versions of that same PR peaked at over 90 commits. And these 30 commits were not the usual "trial-and-error" "back-and-forth" you have while developing, but the exact steps you needed to go to arrive at the new structure by transforming things from its starting point over to the desired code structure. And that's what is meant by the "rebase early, rebase often" in the style guide: Don't document your journey through the maze, but the best path to reach its goal.

tslmy:
@BenBE that totally makes sense to me. Thanks for the very detailed explanation of the rationales. (What you wrote should really be stored in `doc/`!)

Let me undo my commits, regroup them, and then force-push a cleaner "history". 

----

Spoiler alert -- With a Frankenstein-style SDK I pieced together from iOS and macOS SDKs, I was able to make the `IOKit` work:

image

I'll revert some of the changes in this PR soon and provide instructions on how to compose this SDK instead.

Explorer09:
> And no: There is no commit hook magically fixing things. Writing well structured code is the responsibility of the developer – even if they decide to support their work with tools like code formatters (like `astyle`) or AI agents.

While automated code style fixing tools can correct the line breaking, indents and token spacing, there are many things that would be off the scope of said tools: Variable or function naming, choice on whether to define a constant as a macro or enum, the overall code structure in a header or source file, etc.

Therefore, don't expect there would be a tool, AI powered or not, that magically fixes everything.

darwin/DarwinProcess.c
tslmy: This Frankenstein-style SDK worked better than I expected (🤦 ). It inspired me to use iPhone Simulator's SDK files for this `libproc` support, so I don't even need XNU now. This is resolved.

darwin/DarwinMachine.c
tslmy: No longer applicable. Resolving.

darwin/Platform.c
tslmy: This Frankenstein-style SDK worked! See PR description for how it is composed. I'll resolve this thread.

Makefile.am
tslmy: I don't need this anymore. Resolving this thread.

darwin/os/availability.h
tslmy: Ditto.

darwin/DarwinMachine.h
tslmy: Ditto.

darwin/DarwinMachine.h
tslmy: No longer applicable. Resolving.

tslmy:
Hi all,

What a journey!

This Frankenstein-style SDK worked better than I expected (🤦 ). It inspired me to use iPhone Simulator's SDK files for this `libproc` support, so I don't even need XNU now. This PR has now boiled down to just 4 small patches, which I've re-organized as atomic commits.

I guess this "simulator SDK will help a lot" is common knowledge to the jailbreak community 🤷 . But when it comes to learning something, _later_ is better than _never_! :D 

I appreciate you all guiding me through this weekend hobbyist project of mine, as I'm quite amateur when it comes to `htop` codebase itself, open-source C programs in general, and iOS/Mach/XNU development. 🫡

configure.ac
tslmy: Well I don't even need this anymore.

Compat.c
```diff
@@ -24,6 +24,15 @@ in the source distribution for its full text.
 # define PATH_MAX 4096
 #endif
 
+/* iOS 6.1 SDK does not have these constants */
+#ifndef AT_FDCWD
+# define AT_FDCWD -100
+#endif
+
+#ifndef AT_SYMLINK_NOFOLLOW
+# define AT_SYMLINK_NOFOLLOW 0x100
```
Explorer09: How can we be sure every OS define `AT_FDCWD` and `AT_SYMLINK_NOFOLLOW` to the same value? Both constant values seem to be internal to the OS.

darwin/Platform.c
```diff
@@ -404,15 +404,26 @@ void Platform_setMemoryValues(Meter* mtr) {
...
-              vm->compressor_page_count - vm->purgeable_count - vm->external_page_count;
+              vm->speculative_count + vm->wire_count - vm->purgeable_count;
+#ifdef HAVE_STRUCT_VM_STATISTICS64_EXTERNAL_PAGE_COUNT
+   used -= vm->external_page_count;
+#endif
+#ifdef HAVE_STRUCT_VM_STATISTICS64_COMPRESSOR_PAGE_COUNT
+   used += vm->compressor_page_count;
    mtr->values[MEMORY_METER_USED] = (double)(used - vm->compressor_page_count) * page_K;
+#else
+   mtr->values[MEMORY_METER_USED] = (double)used * page_K;
```
Explorer09: A separate `mtr->values[MEMORY_METER_USED]` assignment line doesn't look right.

configure.ac
natoscott: > 'You surely have no idea of how Autoconf-powered configure works.'

@Explorer09 this is no way to communicate with new contributors.  Keep it friendly and supportive, thanks.

darwin/Platform.c
tslmy: Well, there are already two branches. To avoid assigning again, I'll have to introduce a new variable:

```diff
@@ -401,6 +401,7 @@
    const struct vm_statistics* vm = &dhost->vm_stats;
 #endif
    double page_K = (double)vm_page_size / (double)1024;
+   double usedPages = 0.0;
 
    mtr->total = dhost->host_info.max_mem / 1024;
 #ifdef HAVE_STRUCT_VM_STATISTICS64
@@ -411,13 +412,14 @@
 #endif
 #ifdef HAVE_STRUCT_VM_STATISTICS64_COMPRESSOR_PAGE_COUNT
    used += vm->compressor_page_count;
-   mtr->values[MEMORY_METER_USED] = (double)(used - vm->compressor_page_count) * page_K;
+   usedPages = (double)(used - vm->compressor_page_count);
 #else
-   mtr->values[MEMORY_METER_USED] = (double)used * page_K;
+   usedPages = (double)used;
 #endif
 #else
-   mtr->values[MEMORY_METER_USED] = (double)(vm->active_count + vm->wire_count) * page_K;
+   usedPages = (double)(vm->active_count + vm->wire_count);
 #endif
+   mtr->values[MEMORY_METER_USED] = usedPages * page_K;
    // mtr->values[MEMORY_METER_SHARED] = "shared memory, like tmpfs and shm"
 #ifdef HAVE_STRUCT_VM_STATISTICS64
 #ifdef HAVE_STRUCT_VM_STATISTICS64_COMPRESSOR_PAGE_COUNT
```

which doesn't look too elegant to me. What do you say?

configure.ac
Explorer09: @natoscott 

> this is no way to communicate with new contributors. Keep it friendly and supportive, thanks.

I was irritated by the attitude that the person making a pull request didn't try to pay responsibility for the code AI generated and attempted to push the code review work onto me or other volunteers.

It's just lucky that I have more knowledge with Autoconf that I can tell the patch was wrong in many places.

Example: The `host_os` variable in Autoconf can never be `ios6`. Both macOS and iOS are detected as `darwin` in any Autoconf-based configure (specifically it's `config.guess` and `config.sub` that define the OS keywords for Autoconf). A simple build testing could tell that the patch didn't work. And yet they pushed that assuming this very simple error doesn't matter.

Compat.c
tslmy: That's actually a good point. I simply saw the code above

```c
/* GNU/Hurd does not have PATH_MAX in limits.h */
#ifndef PATH_MAX
# define PATH_MAX 4096
#endif
```

and assumed hardcoding platform-specific constants is acceptable here at `htop`. 

I've surrounded these two constants with `#ifdef __APPLE__`. That should be just safe enough.

Explorer09: @tslmy For limit constants it's okay to define it to an arbitrary value as long as it's a common denominator across all OSes.

Not the case for constants that are used for option flags. Even you can define them, chances are the flags won't work, or worse, it might enable some undocumented feature of that OS's API that could wreak havoc.

darwin/Platform.c
Explorer09: I mean, it might be good to examine why there were two branches in the first place. Ideally the formula for counting "used" memory shouldn't differ between 32-bits and 64-bits.

tslmy: It looks like Apple Silicon support (https://github.com/htop-dev/htop/commit/731812d92a7d566d82b30a7f4ae44353f16290fe), perhaps something to do with Unified Memory if I'm reading it right.

That said, it's not yet clear to me how to unify all three branches together.

Compat.c
tslmy: Yeah that's fair. For now I'll just leave the two statements guarded by `#ifdef __APPLE__`. I'll wait and see what CI says (assuming our test suite covers this).

natoscott: Worth taking a step back here - the Compat functions using these macros are only used on Linux (same is true for xReadfileat AFAICT).  A simpler fix may involve moving these things into the linux/ subdirectory so that they're not compiled at all on on darwin (or anywhere else).

tslmy: Hmmm `Compat` is also used in `FreeBSDMachine.c` and `FreeBSDProcessTable.c`...

Explorer09: @tslmy Can you examine why in iOS 6 `AT_FDCWD` and `AT_SYMLINK_NOFOLLOW` are undefined? This doesn't look normal as the `AT_FDCWD` is part of the [open(2)](https://pubs.opengroup.org/onlinepubs/9699919799.orig/functions/open.html) syscall API and `AT_SYMLINK_NOFOLLOW` is part of [fstatat(2)](https://pubs.opengroup.org/onlinepubs/9699919799/functions/fstatat.html) API. Both are specified in POSIX.

tslmy:
Turns out the deprecated name (`*master*`) is now caught as an error when building for modern macOS.

I've force-pushed again to make use of 

```c
#if __IPHONE_OS_VERSION_MIN_REQUIRED <= __IPHONE_6_1
```

as a stricter way to apply the patches. When building for iOS, this macro is needed anyway. 

I _can_ put this in `configure`, but another part of me doesn't want to expose this macro to all other platforms.

Compat.c
natoscott: > Hmmm Compat is also used in FreeBSDMachine.c and FreeBSDProcessTable.c...

The header is included there, but the filesystem related calls are not used AFAICS.  It doesn't make sense that they would be either, these calls were added to optimize the /proc accesses, which is specific to Linux.

There are unrelated definitions relating to static_assert in this header, which were added later - it may be these asserts that are required on other platforms (I see a use in MemoryMeter.c, could be the cause).  

We should be able to shift those (perhaps into Macros.h) and then move the rest of Compat.[ch] into linux/ where it kinda belongs, since those routines appear to be unused on other platforms.

tslmy: @Explorer09 :

> Can you examine why in iOS 6 AT_FDCWD and AT_SYMLINK_NOFOLLOW are undefined? This doesn't look normal 

I think there's an important distinction to make here: iOS the software itself and the iOS SDK.

I believe it _is_ defined in iOS 6's _source code_ itself (as deep down as XNU); it's just that it's not exposed to us developers in its official SDK.

iOS isn't known for its openness back in the 2010s. It doesn't really surprise me that Apple decided to not expose current working directory (`AT_FDCWD`) to developers.

That said, when it comes to `AT_SYMLINK_NOFOLLOW`, there's actually a counterpart in `iPhoneOS6.1.sdk/usr/include/sys/fcntl.h`:

```c
#define O_NOFOLLOW  0x0100      /* don't follow symlinks */
```

The bottom line: reusing the values defined in XNU is still the best idea I have so far 🤷

tslmy: @natoscott :

I hear you. Honestly I didn't know `/proc` was a Linux-specific thing (TIL a lot). I'm not very confident dealing with the clean-up just yet -- still on my warm-up reps here. Do you think we can find someone more experienced in FreeBSD & Linux development to look at it, or could you wait till I make another PR dedicated for this next weekend?

natoscott: @tslmy leave it with me, I'll see if this approach is feasible tomorrow.

Compat.c
```diff
@@ -24,6 +24,20 @@ in the source distribution for its full text.
 # define PATH_MAX 4096
 #endif
 
+#if __IPHONE_OS_VERSION_MIN_REQUIRED <= __IPHONE_6_1
```
BenBE:
```suggestion
#if defined(__IPHONE_OS_VERSION_MIN_REQUIRED) && defined (__IPHONE_6_1) && __IPHONE_OS_VERSION_MIN_REQUIRED <= __IPHONE_6_1
```

Otherwise this may cause issues with `-Wundef`.

Explorer09: I have a better idea:

Just check for `defined(__IPHONE_OS_VERSION_MIN_REQUIRED)` without the `__IPHONE_6_1` part. That way people compiling for iOS 7, 8 or similar would simply work (tm) as the `AT_FDCWD` and `AT_SYMLINK_NOFOLLOW` are guarded by the conditionals.

Keep the code comment saying that the workaround is intended for old iOS versions (iOS 6).

```c
#ifdef __IPHONE_OS_VERSION_MIN_REQUIRED
// iOS 6.1 SDK does not have these constants
// These values are taken from the XNU kernel so they should be compatible
// <https://github.com/apple-opensource/xnu/blob/4f43d4276fc6a87f2461a3ab18287e4a2e5a1cc0/bsd/sys/fcntl.h>

# ifndef AT_FDCWD
#  define AT_FDCWD -2
# endif

# ifndef AT_SYMLINK_NOFOLLOW
#  define AT_SYMLINK_NOFOLLOW 0x0020
# endif
#endif // __IPHONE_OS_VERSION_MIN_REQUIRED
```

tslmy: Sounds good. Updated.

darwin/DarwinMachine.c
```diff
@@ -24,6 +25,12 @@ in the source distribution for its full text.
 #include "generic/openzfs_sysctl.h"
 #include "zfs/ZfsArcStats.h"
 
+// Compatibility for older SDKs.
+#if __IPHONE_OS_VERSION_MIN_REQUIRED <= __IPHONE_6_1
+#ifndef kIOMainPortDefault
```
Explorer09: This compatibility definition shouldn't be limited to iOS. (The internet says some older macOS versions also need this.)

And, because `kIOMainPortDefault` is defined as a constant variable, not a macro, a better way to check it is to use `AC_CHECK_DECLS` in `configure`.

```m4
AC_CHECK_DECLS([kIOMainPortDefault], [], [], [[#include <IOKit/IOKitLib.h>]])
```

```c
#if defined(HAVE_DECL_KIOMAINPORTDEFAULT) && !HAVE_DECL_KIOMAINPORTDEFAULT
#define kIOMainPortDefault kIOMasterPortDefault
#endif
```

This way would work better.

tslmy: Wow, all the magical things `configure` can do! Will add it now.

tslmy: Done. Will resolve.

Compat.c
tslmy: Resolving this.

darwin/Platform.c
```diff
@@ -15,10 +15,11 @@ in the source distribution for its full text.
 #include <math.h>
 #include <stdlib.h>
 #include <unistd.h>
-#include <net/if.h>
+#include <sys/socket.h>
+#include <net/if.h> // After `sys/socket.h` for struct `sockaddr` (for iOS6 SDK)
```
Explorer09: Didn't quite understand the comment. Did you get a build error saying that `struct sockaddr` is undefined if `<net/if.h>` is included before `<sys/socket.h>`? (This sounds like a broken header in iOS SDK.)

tslmy: Yeah. Moving this line:
```diff
+#include <sys/socket.h>
#include <net/if.h>
#include <net/if_types.h>
-#include <sys/socket.h>
```

helped resolving:

```
In file included from darwin/Platform.c:18:
/Volumes/Xcode/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS6.1.sdk/usr/include/net/if.h:307:19: error: field has incomplete type 'struct sockaddr'
  307 |                 struct  sockaddr ifru_addr;
      |                                  ^
```

Explorer09: Does the `net/if_types.h` header also have such hidden dependency?

tslmy: Not sure. Did you meant to ask why moving `socket.h` instead of moving `net/if.h`? I just thought it would be tidier keeping `net/*` imports together.

Compat.c
natoscott: @tslmy it worked out nicely - https://github.com/htop-dev/htop/pull/1832 - you can drop those changes to Compat.h once this is merged.

tslmy: Awesome. Thanks!

darwin/Platform.c
Explorer09: @tslmy My idea is to put headers with special dependencies after other alphabetically-ordered headers, with comment noting which headers depends on which (your comment is okay).

The reason for alphabetical ordering is to make it easy (for new patches) to find where to insert new headers. The ordering can also reduce cognitive load for programmers reading the code. The "header X must be included after header Y" cases are exceptions and so I like to move exceptions to outside the list of headers that don't need exceptions.

In the particular case, I would put `#include <net/if.h>` line to after `#include <mach/port.h>` but before `#include <CoreFoundation/CFBase.h>`.

tslmy: Sure. Done.

darwin/DarwinMachine.c
```diff
@@ -24,6 +25,10 @@ in the source distribution for its full text.
 #include "generic/openzfs_sysctl.h"
 #include "zfs/ZfsArcStats.h"
 
+// Compatibility for older SDKs.
+#if defined(HAVE_DECL_KIOMAINPORTDEFAULT) && !HAVE_DECL_KIOMAINPORTDEFAULT
```
BenBE: Shouldn't this be

```suggestion
#if !defined(HAVE_DECL_KIOMAINPORTDEFAULT) || !HAVE_DECL_KIOMAINPORTDEFAULT
```

darwin/Platform.c
BenBE: Try to sort out the amount of used/external/compressed memory first, before assigning. The external and compressed memory should be restricted to cases where they are provided by the OS, thus in these cases modifying the `used` value should be trivial in the appropriate branches. After that's done, updating the `mtr->values[MEMORY_METER_USED]` should be straight forward.

It furthermore helps to limit nesting things to the bare minimum.

darwin/DarwinMachine.c
tslmy: Let me think... `HAVE_DECL_KIOMAINPORTDEFAULT` will only be defined iff `"$my_htop_platform" = darwin` (in `configure.ac`). 

If we are on macOS, then `HAVE_DECL_KIOMAINPORTDEFAULT` will be defined and `= true`, so
- the current expression evaluates to `true and not true -> true and false -> false` and
- the proposed expression to `not true or not true -> false or false -> false`.

If we are on iOS 6, then `HAVE_DECL_KIOMAINPORTDEFAULT` will be defined and `= false`, so
- the current expression evaluates to `true and not false -> true and true -> true` and
- the proposed expression to `not true or not false -> false or true -> true`.

If we are on other platforms, this `.c` file won't be compiled at all. But I don't really very well about C's build systems, so let's have the mental exercise of when it does get compiled:
- the current expression evaulates to `false && undefined`. The expression will shortcircuit and will not arrive at "undefined".
- the proposed expression evaulates to `not false || undefined -> true || undefined`. The expression will shortcircuit and will not arrive at "undefined".

So they are equivalent. I'm OK doing either.

I'd prefer `if defined(HAVE_DECL_KIOMAINPORTDEFAULT)`, because -- as long as `defined(HAVE_DECL_KIOMAINPORTDEFAULT)` is evaluated -- it will always be `true`. It's just one mental hoop to me to think "when it will not be true". But let me incoporate this change.

darwin/Platform.c
tslmy: Resolved in another thread.

darwin/Platform.c
tslmy: Gave it a pass. I introduced two local variables: `used_counts_from_statistics64` (to decouple the existing `mtr->values[MEMORY_METER_USED]` assignments from the existing `#ifdef HAVE_STRUCT_VM_STATISTICS64` branches) and `external_page_count` (to hide `#ifdef HAVE_[...]_EXTERNAL_PAGE_COUNT` behind it, so that we avoid a reassignment). There's actually one more local variable `used_count` I introduced, but it's mainly for typography.

BenBE (approved):
LGTM.

PR #1830, opened 2025-11-30
Contributor policy updates - Code of Conduct and AI guidelines

natoscott:


CONTRIBUTING.md
```diff
@@ -43,3 +49,20 @@ a first orientation to go by when contributing to this repository.
 
 The details of the coding style as well as what to take care about with your
 contributions can be found in our [style guide](docs/styleguide.md).
+
+Use of AI
+---------
```
BenBE, line 51:
```suggestion
Use of AI and other assistive technologies
------------------------------------------
```

CONTRIBUTING.md
BenBE: We may want to move this section below the style guide section. The alternative is moving the style guide and PR section above bug reports and feature requests. Currently the style guide is a bit too hidden in the text.

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,134 @@
...
+  community
+
+Examples of unacceptable behavior include:
+
+* The use of sexualized language or imagery, and sexual attention or advances of
+  any kind
+* Trolling, insulting or derogatory comments, and personal or political attacks
+* Public or private harassment
+* Publishing others' private information, such as a physical or email address,
+  without their explicit permission
```
BenBE, line 35:
```suggestion
* Publishing others' private information ("doxxing"), such as a physical or
  email address, without their explicit permission
```

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,134 @@
...
+community include:
+
+* Demonstrating empathy and kindness toward other people
+* Being respectful of differing opinions, viewpoints, and experiences
+* Giving and gracefully accepting constructive feedback
+* Accepting responsibility and apologizing to those affected by our mistakes,
+  and learning from the experience
+* Focusing on what is best not just for us as individuals, but for the overall
+  community
+
```
BenBE, line 27:
```suggestion
* Providing technical insights into issues and outlining solutions to problems

```

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,134 @@
...
```
BenBE: No need for this first blank line … ;-)

CONTRIBUTING.md
natoscott: Agreed, I think moving style guide earlier will be helpful, I'll add a followup-commit.

CODE_OF_CONDUCT.md
natoscott: Agree with the sentiment, but not sure this adds new information?  It'd be better to deviate minimally from the contributor covenant text IMO.

CODE_OF_CONDUCT.md
natoscott: :)  (nor the blank last line the linter found) Fixed, thanks.

CODE_OF_CONDUCT.md
natoscott: Definitely - I feel like that is covered by "Giving and gracefully accepting constructive feedback" tho.

CONTRIBUTING.md
natoscott: Other assistive technologies?  Such as?  Do they have the same concerns (copyright, etc) as some folk have expressed for AI?

CONTRIBUTING.md
```diff
@@ -49,3 +49,20 @@ a first orientation to go by when contributing to this repository.
...
+
+You are welcome to use whatever tools work best for you, which includes use
+of AI tools to assist with developing htop. AI tools have proven invaluable
+as learning aids for new contributors, helped to identify problematic code,
+and have been used to fill cross-platform knowledge gaps.
+
+With great power comes great responsibility. Please ensure you review and
+understand any changes made by tools you use before opening pull requests.
+Be aware that AI tools may generate content which is an exact copy of code
```
Explorer09, line 60: I prefer to remove the them "exact":

```suggestion
Be aware that AI tools may generate code or content that is copied
```

From the legal perspective it does not have to be exact copy of code to constitute copyright infringement. If the AI generates a substantial, unattributed block of code, that's enough to raise suspicions on the copyright.

My personal attitude towards AI (not suggesting as an htop policy) is: don't use it for generating anything more than trivial code. Trivial patches may be okay for AI because it would be easy to argue (e.g. the "merger doctrine" as a legal defense). But IANAL (I am not a lawyer). Don't take that as a serious legal advice.

P.S. By "trivial code" I mean bug fixes that can be done by changing just one or two lines of code. If an average human programmer looks at the bug and tries to fix it, they would fix it the same way. (The "merger doctrine" refers to ideas that have only a limited way of expressing it. And for software code this would be a good example for defending "merger doctrine" in court.)

natoscott: I chose (and prefer) the existing wording because it presents the most extreme case.  It doesn't preclude that less exact similarity matches can also happen, they certainly can.  Changing the noun ("exact copy") to a verb ("is copied") is not a correct change in this context though.  Full disclosure, I hold a Masters in Applied AI (graduated 2024) and I work on AI performance in my day job, so I believe I have a solid, current foundation for crafting some initial guidelines for the project.

Like you, I am not a lawyer.  I have been fortunate enough to have access to several over the past 30 years of working on open source projects though, and discussed copyright issues on a good number of occasions, so I feel like this is also an area I'm comfortable in crafting guidelines.

CONTRIBUTING.md
BenBE: If you use code generators and auto-completion tools you may have similar problems, although not to this extend.

Your auto-completion tool might include a template library of code snippets to choose from that might have originated from various places and thus have varying licenses.

For Verilog/VHDL you often have so-called IP cores of proprietary functionality where you need to be careful with licenses.

Furthermore the copyright/license issues aren't unique to AI alone, even if they are most pronounced with these tools. And last I don't like keeping this limited to AI as this might apply to other tools even when the AI hype has moved on …

natoscott: > I don't like keeping this limited to AI as this might apply to other tools even when the AI hype has moved on …

Heh.  If that ever happens, I'd much prefer to delete this section completely.

The goal here is to answer questions and set guidelines specifically related to AI, which is the question on many peoples minds.  AFAIK, noone ever asked us these about these other coding tools, despite them being around for many years, so at this stage I suggest we don't muddy the water.  My 2c, happy to hear other opinions too (anyone using such tools here OOC?).

CONTRIBUTING.md
```diff
@@ -1,45 +1,65 @@
...
 Bug reports should be posted in the [Github issue
 tracker](https://github.com/htop-dev/htop/issues).
 Bug reports are extremely important since it's impossible for us to test
 htop in every possible system, distribution and scenario. Your feedback
-is what keeps the tool stable and always improving!  Thank you!
+is what keeps the tool stable and always improving - thank you!
+
+Please label Github issues that are feature requests with a `feature request`
+label. If you can't do this yourself, don't worry, the friendly folk from the
+core team update labels as part of their regular reviews.
```
Explorer09, line 23: I just discovered I do not have the ability to add labels to an issue report. Perhaps this privilege is limited to the core team? Besides, there is no "feature request" label, only the "[new feature](https://github.com/htop-dev/htop/issues?q=state%3Aopen%20label%3A%22new%20feature%22)" label.

fasterit: I have renamed the label to "feature request" as that is much more an industry term than "new feature". The latter was probably meant to differentiate from "enhancement" but at 30+ labels we have now ... that won't work (any more).

PR #1832, opened 2025-12-02
Refactor and move the Compat code to assist portability

natoscott:
Related: #1828

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -40,7 +40,6 @@ myhtopsources = \
 	ColumnsPanel.c \
 	CommandLine.c \
 	CommandScreen.c \
-	Compat.c \
 	CPUMeter.c \
 	CRT.c \
 	DateMeter.c \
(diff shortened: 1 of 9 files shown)


Debug.h
```diff
@@ -0,0 +1,34 @@
+#ifndef HEADER_Debug
+#define HEADER_Debug
+/*
+htop - Debug.h
+(C) 2020 htop dev team
```
BenBE, line 5:
```suggestion
(C) 2025 htop dev team
```

Explorer09:
There is only one place that uses `static_assert`? This makes me wonder if we still need this macro anymore, or whether we need a new policy regarding the use of `static_assert`.

I have no objection to this PR; just point out something that's worth discussing in the future. (In #1829 I have removed that only `static_assert` use.)

Debug.h
natoscott: I didn't change any code here, so ended up just leaving the same copyright notice as the original file.

Issue #506, opened 2021-01-31
Create the ability to run arbitrary commands on keypress


fasterit:
Generalization of #27. From there:

The htop-dev team would like to see implemented is the functionality to start a `$XDG_CONFIG_HOME/htop/run_script` on a key press (possibly r):
`run_script` would be called with "PID" "USER" "COMMAND" on the currently highlighted line.
Bonus points for providing a sequence of "PID" "USER" "COMMAND" on stdin if multiple processes are tagged (=marked).
Bonus points for making the script location configurable in `/etc/htoprc` so that system-wide defaults could be set (having a `$XDG_CONFIG_HOME/htop/run_script` present would override the global setting still). This would (automatically) also allow to override the script location for a user in their local `htoprc` (same config reader). 

If `run_script` outputs on stdout and/or stderr this would go into the `InfoScreen` pager. Otherwise htop would just return to normal operations once the `exec*` returns. See `OpenFilesScreen.c` or `TraceScreen.c`for inspiration.

Attaching a gdb could be a htop-supplied example of how to use such a script.

PR #1843, opened 2025-12-25
Add ability to run arbitrary commands on keypress

mchlyu:
Addresses [#506](https://github.com/htop-dev/htop/issues/506)  

All input to script is given on stdin as a sequence of "PID" "USER" "COMMAND" in TSV format.  
Users can configure htoprc to provide a script path setting that will act as fallback if no file exists at `$XDG_CONFIG_HOME/htop/run_script`.

I tested functionality using this gdb script on Ubuntu.
```
#!/bin/bash

# read input from stdin, expecting TSV formatted tuple [int, string, string]
while IFS=$'\t' read -r pid usr cmd; do
    gdb -p "$pid" -batch -ex "thread apply all bt" -ex "quit"
done
```

Created new InfoScreen pager heavily inspired by `OpenFilesScreen` and `TraceScreen` to display stdout and stderr from script.  
Also updated help screen and man page to include this feature.

Valgrind did not report any new memory leaks.

I'm open to any suggestions for improving the structure, security, etc. of the code or clarity of the documentation.  
This is my first contribution to this project (and open source in general), so please let me know if I’ve missed any conventions.

### DIFF
--- a/RunScript.c
+++ b/RunScript.c
@@ -12,10 +12,12 @@ in the source distribution for its full text.
 #include <assert.h>
 #include <errno.h>
 #include <fcntl.h>
+#include <grp.h>
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/stat.h>
 #include <unistd.h>
 
 #include "Action.h"
@@ -34,33 +36,29 @@ static void write_row(Row* row, int write_fd) {
    Process* this = (Process*) row;
    assert(Object_isA((const Object*) this, (const ObjectClass*) &Process_class));
 
-   int pid_length = snprintf(NULL, 0, "%d", row->id);
-   char pid[pid_length + 1];
-   snprintf(pid, pid_length + 1, "%d", row->id);
+   int pid_len = 0;
+   int pid = row->id;
+   while (pid > 0) {
+      pid /= 10;
+      pid_len++;
+   }
 
-   char* pid_str = String_cat(pid, "\t");
-   char* user = String_cat(this->user, "\t");
-   char* cmd = String_cat(Process_getCommand(this), "\n");
-   char* user_and_cmd = String_cat(user, cmd);
-   char* line = String_cat(pid_str, user_and_cmd);
+   char* line;
+   int user_len = strlen(this->user);
+   int cmd_len = strlen(Process_getCommand(this));
+   // writes pid_len:PID,user_len:User,cmd_len:Command\n in netstring format
+   xAsprintf(&line, "%d:%d,%d:%s,%d:%s\n", pid_len, row->id, user_len, this->user, cmd_len, Process_getCommand(this));
 
-   // writes PID\tUser\tCommand\n, using TSV format
    size_t count = strlen(line);
    char* line_start = line;
    while (count > 0) {
-      ssize_t res = write(write_fd, line_start, strlen(line));
+      ssize_t res = write(write_fd, line_start, strlen(line_start));
       if ((res == -1 && errno != EINTR) || res == 0)
          break;
-
       count -= res;
       line_start += res;
    }
-
-   free(pid_str);
-   free(user);
-   free(cmd);
-   free(user_and_cmd);
-   free(line);
+  free(line);
 }
 
 void RunScript(State* st) {
@@ -95,26 +93,26 @@ void RunScript(State* st) {
 
       char* home = getenv("XDG_CONFIG_HOME");
       if (!home)
-         home = String_cat(getenv("HOME"), "./config");
+         home = String_cat(getenv("HOME"), "/.config");
 
       const char* path = String_cat(home, "/htop/run_script");
       FILE* file = fopen(path, "r");
       if (file) {
-         execl(path, path, NULL);
-         // should not reach here unless execl fails
+         // executing script in root's directory, probably not malicious
+         root_exec(path, false);
+         // should not reach here unless fexecve fails
          fprintf(stderr, "error excuting %s\n", path);
-         perror("execl");
       } else {
          // check if htoprc has something
          const char* htoprc_path = st->host->settings->scriptLocation;
-         execl(htoprc_path, htoprc_path, NULL);
+         // path can point to anything, so drop sudo for safety
+         root_exec(htoprc_path, true);
 
-         // only reach here if execl fails
-         fprintf(stderr, "error executing %s from htoprc", htoprc_path);
+         // only reach here if fexecve fails
+         fprintf(stderr, "error executing %s from htoprc. ", htoprc_path);
          fprintf(stderr, "if you expected your runscript to be executed, htop looked for it at %s", path);
-         perror("execl");
       }
-      exit(1);
+      exit(127);
    }
 
    close(child_read[0]);
@@ -150,4 +148,72 @@ void RunScript(State* st) {
       InfoScreen_run((InfoScreen*)sos);
    }
    ScriptOutputScreen_delete((Object*)sos);
+}
+
+void root_exec(const char* path, bool drop_sudo) {
+   // do not use O_CLOEXEC flag as that will cause fexecve to fail with ENOENT on a script
+   int fd = open(path, O_RDONLY);
+   if (fd < 0) {
+      perror("open");
+      return;
+   }
+   // check that path is even a file
+   struct stat st;
+   if (fstat(fd, &st) == -1) {
+      perror("fstat");
+      return;
+   }
+
+   uid_t curr_uid = getuid();
+   if (drop_sudo) {
+      // need to remove root from ourselves if we are root
+      if (curr_uid == 0) {
+         char* sudo_uid_str = getenv("SUDO_UID");
+         if (!sudo_uid_str) {
+            fprintf(stderr, "sudo uid envar does not exist\n");
+            return;
+         }
+         uid_t uid = strtoul(sudo_uid_str, NULL, 10);
+         if (uid == 0) {
+            fprintf(stderr, "sudo uid envar is root, failed to get uid of invoking user\n");
+            return;
+         }
+
+         char* sudo_gid_str = getenv("SUDO_GID");
+         if (!sudo_gid_str) {
+            fprintf(stderr, "sudo gid envar does not exist\n");
+            return;
+         }
+         gid_t gid = strtoul(sudo_gid_str, NULL, 10);
+         if (gid == 0) {
+            fprintf(stderr, "sudo gid envar is root group, failed to get gid of invoking user\n");
+            return;
+         }
+         
+         // remove supplementary groups
+         if (setgroups(0, NULL) == -1) {
+            perror("setgroups");
+            return;
+         }
+         if (setgid(gid) == -1) {
+            perror("setgid");
+            return;
+         }
+         if (setuid(uid) == -1) {
+            perror("setuid");
+            return;
+         }
+      }
+   } else if (curr_uid == 0 && (st.st_gid != 0 || st.st_uid != 0)) {
+      // we are root and script does not belongs to root, consider it unsafe
+      fprintf(stderr, "%s does not belong to root; has gid %u and uid %u\n", path, st.st_gid, st.st_uid);
+      return;
+   }
+
+   static char* argv[] = {NULL, NULL};
+   argv[0] = xStrdup(path);
+   static char* env[] = {NULL};
+   fexecve(fd, argv, env);
+
+   perror("fexecve");
 }
--- a/ScriptOutputScreen.c
+++ b/ScriptOutputScreen.c
@@ -76,43 +76,41 @@ static void ScriptOutputScreen_scan(InfoScreen* super) {
          break;
       }
 
-      if (res > 0) {
-         int start = 0;
-         int num_tabs = 0;
-         for (int i = 0; i <= res; i++) {
-            num_tabs += (buffer[i] == '\t');
-            // split line when find \n or exhaust buffer
-            if (i == res || buffer[i] == '\n') {
-               buffer[i] = '\0';
-               char* str;
-               if (num_tabs > 0) {
-                  // manually replace all \t with TABSIZE spaces
-                  str = xMalloc((num_tabs * TABSIZE + i - start + 1) * sizeof(char));
-                  int index = 0;
-                  for (int j = start; j <= i; j++) {
-                     if (buffer[j] == '\t') {
-                        for (int k = 0; k < TABSIZE; k++) {
-                           str[index++] = ' ';
-                        }
-                     } else {
-                        str[index++] = buffer[j];
+      size_t start = 0;
+      int num_tabs = 0;
+      for (size_t i = 0; i <= res; i++) {
+         num_tabs += (buffer[i] == '\t');
+         // split line when find \n or exhaust buffer
+         if (i == res || buffer[i] == '\n') {
+            buffer[i] = '\0';
+            char* str;
+            if (num_tabs > 0) {
+               // manually replace all \t with TABSIZE spaces
+               str = xMalloc((num_tabs * TABSIZE + i - start + 1) * sizeof(char));
+               size_t index = 0;
+               for (size_t j = start; j <= i; j++) {
+                  if (buffer[j] == '\t') {
+                     for (int k = 0; k < TABSIZE; k++) {
+                        str[index++] = ' ';
                      }
+                  } else {
+                     str[index++] = buffer[j];
                   }
-               } else {
-                  str = buffer + start;
                }
-               InfoScreen_addLine(super, str);
-               // store line for next redraw
-               *(sos->data_tail) = xMalloc(sizeof(Node));
-               (*sos->data_tail)->line = xStrdup(str);
-               (*sos->data_tail)->next = NULL;
-               *(&sos->data_tail) = &((*sos->data_tail)->next);
-
-               if (num_tabs > 0)
-                  free(str);
-               start = i + 1;
-               num_tabs = 0;
+            } else {
+               str = buffer + start;
             }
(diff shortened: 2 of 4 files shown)


ScriptOutputScreen.c
```diff
@@ -0,0 +1,133 @@
...
+            num_tabs += (buffer[i] == '\t');
+            // split line when find \n or exhaust buffer
+            if (i == res || buffer[i] == '\n') {
+               buffer[i] = '\0';
+               char* str;
+               if (num_tabs > 0) {
+                  // manually replace all \t with TABSIZE spaces
+                  str = xMalloc((num_tabs * TABSIZE + i - start + 1) * sizeof(char));
+                  int index = 0;
+                  for (int j = start; j <= i; j++) {
```
Explorer09: Use `size_t` for array offsets.

Explorer09:
There are a lot of feature suggestions of running arbitrary commands through htop, but almost all the pull request code suffers from a problem:

htop is often run with elevated privileges (e.g. `sudo`), and there is a need for (1) an indicator on whether the external program should be run with elevated privileges or dropped privileges, and (2) a security check that, if the script is meant to be run with elevated privileges, it's not installed by users other than `root` (which can prevent accidental running of malicious programs).

So this pull request doesn't have either security mechanism yet.

RunScript.c
```diff
@@ -0,0 +1,153 @@
...
+#include "ScriptOutputScreen.h"
+#include "Settings.h"
+#include "XUtils.h"
+
+
+static void write_row(Row* row, int write_fd) {
+   Process* this = (Process*) row;
+   assert(Object_isA((const Object*) this, (const ObjectClass*) &Process_class));
+
+   int pid_length = snprintf(NULL, 0, "%d", row->id);
```
Explorer09: Why this empty `snprintf` call?

RunScript.c
```diff
@@ -0,0 +1,153 @@
...
+      close(child_write[1]);
+
+      char* home = getenv("XDG_CONFIG_HOME");
+      if (!home)
+         home = String_cat(getenv("HOME"), "./config");
+
+      const char* path = String_cat(home, "/htop/run_script");
+      FILE* file = fopen(path, "r");
+      if (file) {
+         execl(path, path, NULL);
```
Explorer09:
``` suggestion
         execl(path, path, (void*)NULL);
```

RunScript.c
```diff
@@ -0,0 +1,153 @@
...
+         // should not reach here unless execl fails
+         fprintf(stderr, "error excuting %s\n", path);
+         perror("execl");
+      } else {
+         // check if htoprc has something
+         const char* htoprc_path = st->host->settings->scriptLocation;
+         execl(htoprc_path, htoprc_path, NULL);
```
Explorer09:
```suggestion
         execl(htoprc_path, htoprc_path, (void*)NULL);
```

RunScript.c
Explorer09: One more thing. Since we need to verify the ownership of the `run_script` to be executed, the `execl`, `execv` family of API would not fit the job.

Consider the `fexecve(3)` or `execveat(2)` API instead.

RunScript.c
```diff
@@ -0,0 +1,153 @@
...
+#include "XUtils.h"
+
+
+static void write_row(Row* row, int write_fd) {
+   Process* this = (Process*) row;
+   assert(Object_isA((const Object*) this, (const ObjectClass*) &Process_class));
+
+   int pid_length = snprintf(NULL, 0, "%d", row->id);
+   char pid[pid_length + 1];
+   snprintf(pid, pid_length + 1, "%d", row->id);
```
BenBE: No VLA. Also use internal `xSnprintf` APIs. Also, why not allocate a static buffer that's large enough for any `%d` like 32 bytes? Or use multiple writes for the target FD?

RunScript.c
```diff
@@ -0,0 +1,153 @@
...
+
+   char* pid_str = String_cat(pid, "\t");
+   char* user = String_cat(this->user, "\t");
+   char* cmd = String_cat(Process_getCommand(this), "\n");
+   char* user_and_cmd = String_cat(user, cmd);
+   char* line = String_cat(pid_str, user_and_cmd);
```
BenBE: Why not make this multiple writes into `stdin` of the new process instead? Apart from `xSnprintf` being the better tool for handling the `%d\t%s\t` part for PID+username here anyway.

RunScript.c
```diff
@@ -0,0 +1,153 @@
...
+         // check if htoprc has something
+         const char* htoprc_path = st->host->settings->scriptLocation;
+         execl(htoprc_path, htoprc_path, NULL);
+
+         // only reach here if execl fails
+         fprintf(stderr, "error executing %s from htoprc", htoprc_path);
+         fprintf(stderr, "if you expected your runscript to be executed, htop looked for it at %s", path);
+         perror("execl");
+      }
+      exit(1);
```
BenBE: IIRC this is inconsistent with other locations that exec where return on error is 127.

Settings.c
```diff
@@ -53,6 +53,8 @@ void Settings_delete(Settings* this) {
    free(this->initialFilename);
    Settings_deleteColumns(this);
    Settings_deleteScreens(this);
+   if (this->scriptLocation)
+      free(this->scriptLocation);
```
BenBE: `free` on `NULL` is a no-op. No need to check.

ScriptOutputScreen.c
```diff
@@ -0,0 +1,133 @@
...
+         break;
+      }
+
+      if (res == 0) {
+         break;
+      }
+
+      if (res > 0) {
+         int start = 0;
+         int num_tabs = 0;
```
BenBE: Should use `size_t`

ScriptOutputScreen.c
```diff
@@ -0,0 +1,133 @@
...
+            continue;
+
+         break;
+      }
+
+      if (res == 0) {
+         break;
+      }
+
+      if (res > 0) {
```
BenBE: Invert condition and `continue` to reduce level of indentation. Cf. `XReadFile` API for other places that need this.

RunScript.c
BenBE: To retrieve the size. But the usage if wrong regardless …

RunScript.c
mchlyu: Got it, thank you for the clarification and suggestions. I'll see if I'm able to implement something that fits [#1844](https://github.com/htop-dev/htop/issues/1844) to address this security issue.

RunScript.c
mchlyu: Ah I missed `xSnprintf` when I was going through `xUtils`, apologies. 
I chose to do all these string concatenations and a singular write because I thought (although I have no data to back this up) a write is much more expensive than string operations, but I see how it makes the code messier. I'm open to switching to multiple writes though.

RunScript.c
mchlyu: What I said [here](https://github.com/htop-dev/htop/pull/1843#discussion_r2647662150), although  I could be wrong or the performance difference doesn't matter that much or both.

RunScript.c
mchlyu: Oh yes you're right, thank you for pointing that out.

Settings.c
mchlyu: Good to know, thanks for the clarification.

RunScript.c
Explorer09: @mchlyu There is also `xAsprintf` if you need to allocate and print the format string in one call.

mchlyu: Ooh that might be useful, thanks for pointing that out!

ScriptOutputScreen.c
mchlyu: Got it, removed one level of indentation so far.
Regarding `xReadFile`, is that meant to be a function in `xUtils`? Because I can't find it in that file nor does searching online yield anything so I'm a bit confused.

RunScript.c
mchlyu: Yup, it was meant to get the size of the pid; I'll probably replace this with one of the printf wrappers in `xUtils`.

RunScript.c
BenBE: The difference performance-wise is one vs. multiple syscalls, but given that malloc might syscall itself (mmap), there is not much of a difference in practice. Also I don't consider this a hot path, thus there's no pressing need for optimizations there.

ScriptOutputScreen.c
BenBE: Has been moved in the latest sources to `Compat.c` as `readfd_internal` with `Compat_readfile` and `Compat_readfileat` being official callers.

RunScript.c
mchlyu: Okay, thanks for clarifying; multiple writes will probably be much cleaner here.

ScriptOutputScreen.c
mchlyu: Got it, thanks for pointing out the new location.

mchlyu:
@BenBE 
I used Gemini to help me understand the codebase, explain portions of the feature request, help debug some segfaults, and suggest approaches. No code was copied from Gemini. 
The last line of this PR and the gdb script were written by ChatGPT.
Should I update the PR to say co-authored by Gemini or GPT? 

Thanks for pointing out a more robust approach. Regarding the command line of the process, could you please elaborate on how would I get the exact length of the command line, since right now I take whatever `Process_getCommand` gives me, and wherever that's terminated is the string I'd copy. If any program overwrote that string, I wouldn't know it's been changed in the first place right? 

For merging the screens, that sounds like a good idea to me, although I'm less clear on which screen(s) this one is similar enough to be combined with. Perhaps something like `CommandScreen`?

RunScript.c
Explorer09: I wonder if we can use a separator other than tab.

The problem is that the `cmdline` can contain almost any character, including tab. I know it could be a pain to securely parse a `cmdline` string when it contains special characters, but we probably have to pass the `cmdline` string securely to the run script.

BenBE:
For merging the existing screens, the most likely candidates are the `strace` screen, which IIRC  is already based on the Command Screen implementation. Thus for running your own commands a similar base can be used.

mchlyu:
Hi, I've addressed the original feedback for the PR. Could you please take a look at it? 

For the dropping sudo permissions, I set it to only drop if reading a location from the htoprc since that could point to anything versus a run_script inside root's directory. Additionally, a program is only considered trusted if its owned by root. I'm unsure if this is a reasonable enough criteria so I'm open to suggestions to change it. 

I've also changed the passing of data to the script to use netstrings instead of TSV.

Valgrind reports that some FDs that are still open, which I believe are the FDs used by fexecve. However, because setting O_CLOEXEC on a file that's a script causes fexecve to fail, I do not know if there is anyway to close these FDs after the exec call.

Gemini helped explain [#1844](https://github.com/htop-dev/htop/issues/1844) for what dropping permissions would look like in Linux.

RunScript.c
```diff
@@ -151,3 +149,71 @@ void RunScript(State* st) {
...
+         if (setgid(gid) == -1) {
+            perror("setgid");
+            return;
+         }
+         if (setuid(uid) == -1) {
+            perror("setuid");
+            return;
+         }
+      }
+   } else if (curr_uid == 0 && (st.st_gid != 0 || st.st_uid != 0)) {
```
Explorer09, line 207: A program installed in `/bin` or `/usr/bin` may be owned by root but associated with a non-root group (e.g. "admin" or "wheel"). So I personally consider it too restrictive to require GID to be 0.

That's one issue I can think of. I would review other pieces of code if I have time.

RunScript.c
```diff
@@ -151,3 +149,71 @@ void RunScript(State* st) {
...
+      return;
+   }
+   // check that path is even a file
+   struct stat st;
+   if (fstat(fd, &st) == -1) {
+      perror("fstat");
+      return;
+   }
+
+   uid_t curr_uid = getuid();
```
Explorer09, line 167: I think this should be `geteuid`, not `getuid`. The difference is not in the sudo use case, but when htop is launched with a SUID privilege.

RunScript.c
```diff
@@ -34,33 +36,29 @@ static void write_row(Row* row, int write_fd) {
...
 
-   int pid_length = snprintf(NULL, 0, "%d", row->id);
-   char pid[pid_length + 1];
-   snprintf(pid, pid_length + 1, "%d", row->id);
+   int pid_len = 0;
+   int pid = row->id;
+   while (pid > 0) {
+      pid /= 10;
+      pid_len++;
+   }
```
Explorer09, line 44: We have `countDigits` function in XUtils.h. Consider using it.

Explorer09:
> 
> For the dropping sudo permissions, I set it to only drop if reading a location from the htoprc since that could point to anything versus a run_script inside root's directory. 

We can't assume the `run_script` will always reside in a directory owned by root, or the script is always owned by root. Check it (see #1866 on how it can be done).

RunScript.c
```diff
@@ -34,33 +36,29 @@ static void write_row(Row* row, int write_fd) {
...
-   char* pid_str = String_cat(pid, "\t");
-   char* user = String_cat(this->user, "\t");
-   char* cmd = String_cat(Process_getCommand(this), "\n");
-   char* user_and_cmd = String_cat(user, cmd);
-   char* line = String_cat(pid_str, user_and_cmd);
+   char* line;
+   int user_len = strlen(this->user);
+   int cmd_len = strlen(Process_getCommand(this));
+   // writes pid_len:PID,user_len:User,cmd_len:Command\n in netstring format
+   xAsprintf(&line, "%d:%d,%d:%s,%d:%s\n", pid_len, row->id, user_len, this->user, cmd_len, Process_getCommand(this));
```
Explorer09, line 50: Why the netstring format? I don't think this makes passing the data safer while it unnecessarily complicates things.

RunScript.c
```diff
@@ -95,26 +93,26 @@ void RunScript(State* st) {
...
-         execl(path, path, NULL);
-         // should not reach here unless execl fails
+         // executing script in root's directory, probably not malicious
+         root_exec(path, false);
+         // should not reach here unless fexecve fails
          fprintf(stderr, "error excuting %s\n", path);
-         perror("execl");
       } else {
          // check if htoprc has something
          const char* htoprc_path = st->host->settings->scriptLocation;
```
Explorer09, line 107: I need clarification on what is the use case here.

Why do we have two ways to specify which command to execute?

I do have concern about this implementation: that the presence of one file can prevent the execution of another. It can also bring in an issue when the `run_script` file can be created or inserted by anyone other than the user running htop when an htop instance is being run.

mchlyu: This was from one of the feature requests in [#506](https://github.com/htop-dev/htop/issues/506)

>Bonus points for making the script location configurable in `/etc/htoprc` so that system-wide defaults could be set (having a `$XDG_CONFIG_HOME/htop/run_script` present would override the global setting still). This would (automatically) also allow to override the script location for a user in their local `htoprc` (same config reader).

I might've misinterpreted this but I understood it as check for a `run_script` in `$XDG_CONFIG_HOME`, and if that doesn't exist, check `htoprc`. 

This is also why I made the assumption earlier that when running htop as sudo, the `run_script` would need to be in directory owned by root, which could also be an incorrect assumption (my path resolves to `/root/.config/htop/run_script` when running with sudo).

RunScript.c
mchlyu: Apologies for missing that, thanks for pointing it out!

RunScript.c
Explorer09: > This was from one of the feature requests in [#506](https://github.com/htop-dev/htop/issues/506)
> 
> > Bonus points for making the script location configurable in `/etc/htoprc` so that system-wide defaults could be set (having a `$XDG_CONFIG_HOME/htop/run_script` present would override the global setting still). This would (automatically) also allow to override the script location for a user in their local `htoprc` (same config reader).
> 
> I might've misinterpreted this but I understood it as check for a `run_script` in `$XDG_CONFIG_HOME`, and if that doesn't exist, check `htoprc`.

I would turn down on that part of the feature request because that doesn't help in the security at all.

Having a path in an htoprc file means htop has to memorize it during the settings' load procedure and write the same path when the setting file is saved.

The concrete use case of that extra path is unclear. And nothing can stop the user from making a symlink in the hard-coded location (`$XDG_CONFIG_HOME/htop/run_script`). So that custom path is redundant and should be turned down.

> 
> This is also why I made the assumption earlier that when running htop as sudo, the `run_script` would need to be in directory owned by root, which could also be an incorrect assumption (my path resolves to `/root/.config/htop/run_script` when running with sudo).

When htop is run with superuser privilege, the `run_script` should be run with superuser privilege only if it's owned by the same EUID, and has `u+rx` permission bits. Otherwise, if `SUDO_USER` is defined and the `run_script` is owned by that same user, run with that UID's privilege (`setuid(2)`). Similar principle applies when htop is run with SUID (i.e. `geteuid() != getuid()`).

RunScript.c
mchlyu: The netstring was meant to address this concern from @BenBE:
>Also, as the command line of a process may contain anything, A more robust solution would provide the exact length of the data to follow and push the data after that literal. Arguments are split by \0 by the kernel, but since programs can overwrite that buffer anything goes …

I understand that your original concern is the possibility of `cmdline` containing \t, and if netstrings don't improve safety, would it be acceptable to sanitize `cmdline` to remove any \t characters before passing it in the original TSV format?

Explorer09: @mchlyu netstring is safe only if there's an agreement between sender and receiver on the length limit. You notice that the length field can contain any number of digits, so it's possible to craft a netstring that would cause arithmetic overflow on an unguarded receiver.

The security of the format comes with caveats, and I would say you didn't know it yet.

PR #1847, opened 2025-12-31
Async refix

DeltachangeOG:
Part of #1570: Graceful Signal Handling

This PR implements a graceful shutdown path for non-fatal termination signals, ensuring the terminal state is restored before the process exits.

Key Improvements:
Graceful Lifecycle: Transitions from immediate termination to a flag-based loop break, allowing the program to run its standard cleanup routines (Platform_done, CRT_done).

Double-Tap Safety: If a second termination signal is received during the shutdown phase, the handler invokes _exit() to ensure an immediate emergency stop.

Fatal Signal Integrity: Does not modify fatal signal handling (e.g., SIGSEGV); fatal handlers still take precedence and behave as they did previously.

POSIX Compliance: Returns a standard 128 + signal exit code.

Technical Implementation:
Flagging (CRT.c): The signal handler captures the signal number and sets a volatile sig_atomic_t terminate_requested flag.

Loop Interruption (ScreenManager.c): The main TUI loop checks the flag and breaks immediately upon signal receipt.

Shutdown Path (CommandLine.c): Once the loop exits and cleanup is complete, the program checks if termination was user-requested to print a status message to stderr and exit with the appropriate status.

### DIFF
--- a/CommandLine.c
+++ b/CommandLine.c
@@ -419,6 +419,11 @@ int CommandLine_run(int argc, char** argv) {
 
    CRT_done();
 
+   if (terminate_requested) {
+      fprintf(stderr, "\nInterrupted by signal %d, terminating.\n", (int)terminate_signal);
+      exit(128 + (int)terminate_signal);
+   }
+
    if (settings->changed) {
 #ifndef NDEBUG
       if (!String_eq(settings->initialFilename, settings->filename))
--- a/CRT.c
+++ b/CRT.c
@@ -951,29 +951,22 @@ static int CRT_colorSchemes[LAST_COLORSCHEME][LAST_COLORELEMENT] = {
 
 static bool CRT_retainScreenOnExit = false;
 
+volatile sig_atomic_t terminate_requested = 0;
+volatile sig_atomic_t terminate_signal = 0;
+
 int CRT_scrollHAmount = 5;
 
 int CRT_scrollWheelVAmount = 10;
 
 ColorScheme CRT_colorScheme = COLORSCHEME_DEFAULT;
 
-ATTR_NORETURN
 static void CRT_handleSIGTERM(int sgn) {
-   CRT_done();
+   if (terminate_requested) {
+      _exit(128 + sgn);
+   }
 
-   if (!CRT_settings->changed)
-      _exit(0);
-
-   const char* signal_str = strsignal(sgn);
-   if (!signal_str)
-      signal_str = "unknown reason";
-
-   char err_buf[512];
-   snprintf(err_buf, sizeof(err_buf),
-           "A signal %d (%s) was received, exiting without persisting settings to htoprc.\n",
-           sgn, signal_str);
-   full_write_str(STDERR_FILENO, err_buf);
-   _exit(0);
+   terminate_signal = (sig_atomic_t)sgn;
+   terminate_requested = (sig_atomic_t)1;
 }
 
 #ifndef NDEBUG
--- a/CRT.h
+++ b/CRT.h
@@ -12,6 +12,7 @@ in the source distribution for its full text.
 #include "Macros.h"
 #include "ProvideCurses.h"
 #include "Settings.h"
+#include "signal.h"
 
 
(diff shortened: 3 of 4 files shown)


Explorer09:
Please disclose the use of AI in this contribution.

DeltachangeOG:
AI did not generate any code in this PR. I used AI to double check syntax and have concepts I am still learning explained, but I wrote and implemented all code myself by hand. I also had AI check to ensure that my PR explanation was concise.

Explorer09:
I saw `CRT_getSignal`, a function that doesn't exist and likely a hallucination from an automated tool. Also the description of the pull request is unnecessarily detailed comparing to the code changes (another symptom of genAI tool being involved).

DeltachangeOG:
I originally wrote crt_getsignal as a public function but decided to surface the atomic flags I set instead, but wrote the code with CRT_getsignal() before changing it, and forgot to refactor that file. the latest PR should not have any reference to CRT_getsignal(). I used AI to try to ensure I was clearly explaining the code I changed. AI was used to understand concepts and check code but was expressly prohibited from generating any code as I am using this contribution to learn C and having something spoon feed it won't let me learn anything.

DeltachangeOG:
> I saw `CRT_getSignal`, a function that doesn't exist and likely a hallucination from an automated tool. Also the description of the pull request is unnecessarily detailed comparing to the code changes (another symptom of genAI tool being involved).

I am not trying to be argumentative, just trying to clarify: I copy pasted code snippets as I wrote them into chatgpt/gemini to make sure I was using the right syntax, and to ensure understanding of the changes I made, but no tool generated the code I put in the PR. I am happy to add AI use to the PR, I wasn't sure what level of AI involvement required acknowledgement.

Explorer09:
Firstly, I apologise for the wrong accusation of using AI. 

Second, I'm not sure if this is the right way to the problem. If a user-triggered signal is going to terminate htop anyway, perhaps we don't need a `terminate_requested` flag or a return from the signal handler at all. I was expecting a `longjmp` or a `siglongjmp` at the end of `CRT_handleSIGTERM`. Why is a return needed, anyway?

DeltachangeOG:
All good. I used AI to write the PR because I am not sure how much to put in about what I did.

The return is to allow the graceful shutdown mechanism already in place to process. the signal handler has double tap protection, so if the user signals again it exits immediately but this allows the existing shutdown path that includes CRT_done() and other clean up to execute with a delay that shouldn't be noticeable to end users. Since this is a graceful shutdown handling the clean up in the handler itself isn't signal safe, using ncurses and a other stdlibc stuff so I just set the flag, let the program do the graceful shutdown then print to std err out that it was an interrupt signal not a normal exit.

DeltachangeOG:
Fixed freeBSD errors ready for new CI run.

CommandLine.c
```diff
@@ -420,8 +420,8 @@ int CommandLine_run(int argc, char** argv) {
    CRT_done();
 
    if (terminate_requested) {
-      fprintf(stderr, "\nInterrupted by signal %d, terminating.\n", terminate_signal);
-      exit(128 + terminate_signal);
+      fprintf(stderr, "\nInterrupted by signal %ld, terminating.\n", (long)terminate_signal);
```
Explorer09: Why the cast to `long`? In Unix systems signals can only have the type of `int`. Even if you use `sig_atomic_t` for temporary storage, it should cast back to `int`.

DeltachangeOG: My bad, I was focused on matching types and making compiler happy. I'll get this cast to an int and recommit.

DeltachangeOG:
These changes are done and ready for CI rerun.

BenBE:
Sorry for the late feedback, but could you please rebase your changes to squash the fixes for your initial commits into those commits? The PR should only contain the actual changes required for change instead of listing every bug fix for your bug fix (cf. style guide for details).

DeltachangeOG:
@BenBE This has been addressed, thank you for the feedback.

PR #1849, opened 2026-01-03
Typographical Fixes

THE-Spellchecker:
Summary:
- Replace "can not" with "cannot" in user-facing errors, comments, and docs
- Fix minor spelling/spacing typos in comments

AI usage:
- None.

PR #1851, opened 2026-01-06
More typographical error fixes

Explorer09:
Follow-up of #1849.


BenBE:
Please split the `configure.ac` changes, the typo fix of that cursed ;-) library and the stylistic changes into separate commits each. Otherwise LGTM.

Issue #1632, opened 2025-03-06
Metrics on macOS


aestriplex:
After developing metrics on overall GPU utilization on macOS, I encountered a limitation in accessing the list of processes using the GPU and the GPU utilization (percentage) per process. Indeed, this information is not publicly available through any API. The only way to obtain it on macOS (as a user) is through the Activity Monitor.
This issue prompted me to consider other metrics that could be implemented for macOS, utilizing reliable information and public APIs. Here are a few ideas that came to my mind:
* `Network interfaces` and related `IP` addresses (via `struct ifaddrs`).
* `Page ins`, `faults`, and number of `wakeups` (all via `struct rusage_info_v2`).
* A metric that displays the GPU utilization splitter by its two main components: `Renderer Utilization %` and `Tiler Utilization %`. I suggest structuring this metric in two colors, similar to memory usage.

PR #1855, opened 2026-01-08
Darwin: Add GPU monitoring support

shatyuka:
Add gpu percentage and time monitoring to macOS.
The data is consistent with Activity Monitor, and the table header titles have been slightly adjusted.

Close: #1597
Related: #1632

### DIFF
--- a/darwin/DarwinProcessTable.c
+++ b/darwin/DarwinProcessTable.c
@@ -124,8 +124,12 @@ void ProcessTable_goThroughEntries(ProcessTable* super) {
          DarwinProcess_scanThreads(proc, dpt);
       }
 
-      // Reset GPU time updated flag
-      proc->gpu_time_updated = false;
+      // Reset GPU time/percent
+      if (ss->flags & PROCESS_FLAG_GPU) {
+         proc->gpu_time_last = proc->gpu_time;
+         proc->gpu_time = 0;
+         proc->gpu_percent = 0.0F;
+      }
 
       super->totalTasks += 1;
 
--- a/darwin/Platform.c
+++ b/darwin/Platform.c
@@ -866,27 +866,17 @@ void Platform_setGPUProcesses(DarwinProcessTable* dpt) {
          goto cleanup;
       }
 
-      if (gpuTime > 0) {
-         uint64_t gputimeDelta = saturatingSub(gpuTime, dp->gpu_time);
+      dp->gpu_time += gpuTime;
+
+      // Only calculate GPU percent if we have a previous sample, and the GPU time has increased
+      if (dp->gpu_time > dp->gpu_time_last && dp->gpu_time_last > 0) {
+         uint64_t gputimeDelta = saturatingSub(dp->gpu_time, dp->gpu_time_last);
          uint64_t monotonicTimeDelta = host->monotonicMs - host->prevMonotonicMs;
          dp->gpu_percent = 100.0F * gputimeDelta / (1000 * 1000) / monotonicTimeDelta;
-         dp->gpu_time = gpuTime;
-         dp->gpu_time_updated = true;
       }
 
 cleanup:
       IOObjectRelease(entry);
-   }
-
-   // Clear GPU time and percent for processes not found in the GPU process list
-   const Vector* rows = dpt->super.super.rows;
-   int table_size = Vector_size(rows);
-   for (int i = 0; i < table_size; ++i) {
-      DarwinProcess* dp = (DarwinProcess*) Vector_get(rows, i);
-      if (!dp->gpu_time_updated) {
-         dp->gpu_time = 0;
-         dp->gpu_percent = 0.0F;
-      }
    }
 
    IOObjectRelease(iterator);
--- a/darwin/DarwinProcess.h
+++ b/darwin/DarwinProcess.h
@@ -26,10 +26,10 @@ typedef struct DarwinProcess_ {
 
    /* Total GPU time used in nano seconds */
    uint64_t gpu_time;
+   /* Total GPU time used in nano seconds in the last scan */
+   uint64_t gpu_time_last;
    /* GPU utilization in percent */
    float gpu_percent;
-   /* Got GPU time info from last scan */
-   bool gpu_time_updated;
 } DarwinProcess;
 
 extern const ProcessClass DarwinProcess_class;


linux/LinuxProcess.c
```diff
@@ -110,8 +110,8 @@ const ProcessFieldData Process_fields[LAST_PROCESSFIELD] = {
 #ifdef SCHEDULER_SUPPORT
    [SCHEDULERPOLICY] = { .name = "SCHEDULERPOLICY", .title = "SCHED ", .description = "Current scheduling policy of the process", .flags = PROCESS_FLAG_SCHEDPOL, },
 #endif
-   [GPU_TIME] = { .name = "GPU_TIME", .title = "GPU_TIME ", .description = "Total GPU time", .flags = PROCESS_FLAG_LINUX_GPU, .defaultSortDesc = true, },
-   [GPU_PERCENT] = { .name = "GPU_PERCENT", .title = " GPU% ", .description = "Percentage of the GPU time the process used in the last sampling", .flags = PROCESS_FLAG_LINUX_GPU, .defaultSortDesc = true, },
+   [TIME_GPU] = { .name = "TIME_GPU", .title = "GPU TIME+", .description = "Total GPU time", .flags = PROCESS_FLAG_LINUX_GPU, .defaultSortDesc = true, },
+   [PERCENT_GPU] = { .name = "PERCENT_GPU", .title = " GPU% ", .description = "Percentage of the GPU time the process used in the last sampling", .flags = PROCESS_FLAG_LINUX_GPU, .defaultSortDesc = true, },
```
Explorer09: Please don't change the `.name` fields of these. It would break backward compatibility with older htop configuration files (htoprc).

shatyuka: okay, fixed

darwin/Platform.c
```diff
@@ -815,10 +815,14 @@ Hashtable* Platform_getGPUProcesses(const Machine* host) {
          if (sscanf(buffer, "pid %d", &pid) != 1)
             goto cleanup;
 
+#if __LP64__
+         unsigned long long gpuTime = (unsigned long long) Hashtable_get(gps, (ht_key_t) pid);
```
Explorer09: Please don't do this. This makes code hard to maintain. You are trying to type-pun, aren't you, by treating the `void*` member as `uintptr_t` and storing an integer to it?

darwin/Platform.c
```diff
@@ -365,14 +365,9 @@ void Platform_setGPUValues(Meter* mtr, double* totalUsage, unsigned long long* t
       return;
    }
 
-   CFMutableDictionaryRef properties = NULL;
-   kern_return_t ret = IORegistryEntryCreateCFProperties(dhost->GPUService, &properties, kCFAllocatorDefault, kNilOptions);
-   if (ret != KERN_SUCCESS || !properties)
-      return;
-
-   CFDictionaryRef perfStats = CFDictionaryGetValue(properties, CFSTR("PerformanceStatistics"));
+   CFDictionaryRef perfStats = IORegistryEntryCreateCFProperty(dhost->GPUService, CFSTR("PerformanceStatistics"), kCFAllocatorDefault, kNilOptions);
```
Explorer09: Good work on simplifying this code. Can you please split this to a separate commit? Because this is unrelated to the process GPU info thing you did later.

shatyuka: sure

darwin/Platform.c
```diff
@@ -798,3 +798,86 @@ static void Platform_getOSRelease(char* buffer, size_t bufferLen) {
...
+Hashtable* Platform_getGPUProcesses(const Machine* host) {
+   const DarwinMachine* dhost = (const DarwinMachine*)host;
+   if (!dhost->GPUService) {
+      return NULL;
+   }
+
+   Hashtable* gps = Hashtable_new(64, true);
+
+   io_iterator_t iterator;
+   if (IORegistryEntryGetChildIterator(dhost->GPUService, kIOServicePlane, &iterator) == KERN_SUCCESS) {
```
BenBE: Invert condition to avoid indentation level.

darwin/DarwinProcessTable.c
```diff
@@ -122,12 +129,18 @@ void ProcessTable_goThroughEntries(ProcessTable* super) {
...
       super->totalTasks += 1;
 
       if (!preExisting) {
          ProcessTable_add(super, &proc->super);
       }
    }
 
+   if (gps) {
+      Hashtable_delete(gps);
+   }
```
BenBE: I don't quite like allocating/deallocating memory on each refresh cycle. The hashtable should be kept across updates until program termination.

darwin/Platform.c
```diff
@@ -798,3 +798,86 @@ static void Platform_getOSRelease(char* buffer, size_t bufferLen) {
...
+         } else if (IOObjectConformsTo(entry, "IOAccelSubmitter2")) {
+            CFNumberRef accumulatedGPUTimeRef = IORegistryEntryCreateCFProperty(entry, CFSTR("accumulatedGPUTime"), kCFAllocatorDefault, kNilOptions);
+            if (!accumulatedGPUTimeRef)
+               goto cleanup;
+
+            unsigned long long accumulatedGPUTime = 0;
+            CFNumberGetValue(accumulatedGPUTimeRef, kCFNumberLongLongType, &accumulatedGPUTime);
+            CFRelease(accumulatedGPUTimeRef);
+
+            gpuTime += accumulatedGPUTime;
```
```suggestion

            gpuTime += accumulatedGPUTime;

            CFRelease(accumulatedGPUTimeRef);
```

Re-order things to have resource management as last action in branch.

darwin/Platform.c
```diff
@@ -798,3 +798,86 @@ static void Platform_getOSRelease(char* buffer, size_t bufferLen) {
...
+               data = xCalloc(1, sizeof(gpuTime));
+               Hashtable_put(gps, (ht_key_t) pid, data);
+            }
+            *data = gpuTime;
+         }
+
+      cleanup:
+         IOObjectRelease(entry);
+      }
+      IOObjectRelease(iterator);
```
```suggestion
cleanup:
         IOObjectRelease(entry);
      }

      IOObjectRelease(iterator);
```

Labels should go first column of the line. Also let's give the code some space.

darwin/Platform.c
```diff
@@ -798,3 +798,86 @@ static void Platform_getOSRelease(char* buffer, size_t bufferLen) {
...
+            gpuTime += accumulatedGPUTime;
+         } else {
+            goto cleanup;
+         }
+
+         if (gpuTime > 0) {
+            if (!data) {
+               data = xCalloc(1, sizeof(gpuTime));
+               Hashtable_put(gps, (ht_key_t) pid, data);
+            }
```
BenBE: Why not handle this early when fetching the item from the hashtable. This will create some entries with zeros, but it makes the overall handling much simpler, as you can skip later NULL checks (knowing you have memory).

Another thing I kinda dislike here is the amount of malloc calls (memory fragmentation) this causes. But given we are talking `ull` here (why not write `uint64_t` instead?), why not drop this hole thing entirely and store the actual data into the pointers directly using `uintptr_t`? Given the unit of `gpuTime` this might be limiting on 32 bit systems, of which we do not have too many (`~0`) left that would also support this interface, I guess …

Given this only affects Darwin 10.13 and 10.14 in compat mode, let's put an `assert(sizeof(uintptr_t) > sizeof(uint32_t));` in and be done on that front …

BenBE (changes requested):
Why the `+` in `GPU Time+`? We should try to minimize active allocations here to avoid memory fragmentation where possible.

darwin/Platform.c
BenBE: Yes, and given the only thing stored is an `uint64_t` it's much more efficient (in the long run) to use this storage directly instead of fragmenting the memory with thousands of malloc calls over time.

BUT, instead of polluting the code line with `unsigned long long int` abominations, just using `uintptr_t` here is much clearer and much friendlier on the eye. Cf. note in my review regarding some precautions related to this.

BenBE:
Please rebase things properly, cf. style guide for details as to why …

darwin/Platform.c
Explorer09: @BenBE No, I can think of a better idea. It is to remove this new `Hashtable`, and instead expand `DarwinProcess` for this GPU time information.

Consider that this new `Hashtable` also uses PID as the indexing key, the two tables can be merged to one.

shatyuka:
I use GPU TIME+ to keep it consistent with TIME+, in terms of numerical formatting. However, considering that there is no "GPU TIME" without the plus sign, I can change it back.

darwin/Platform.c
shatyuka: @Explorer09 That make sense.

BenBE: Be careful that you may encounter items where no process object has been stored yet. There are some items in the process struct that are only set upon first encountering the entry. Apart from this, the idea makes sense.

BenBE:
> I use GPU TIME+ to keep it consistent with TIME+, in terms of numerical formatting. However, considering that there is no "GPU TIME" without the plus sign, I can change it back.

It's fine. I just wanted to know the origins.

darwin/Platform.c
```diff
@@ -799,85 +799,97 @@ const char* Platform_getRelease(void) {
...
+      if (IORegistryEntryGetProperty(entry, "IOUserClientCreator", buffer, &size) != KERN_SUCCESS)
+         goto cleanup;
+
+      pid_t pid;
+      if (sscanf(buffer, "pid %d", &pid) != 1)
+         goto cleanup;
+
+      uint64_t gpuTime = 0;
+      const Table* table = &dpt->super.super;
+      DarwinProcess* dp = (DarwinProcess*) Hashtable_get(table->table, pid);
```
BenBE: Use `ProcessTable_findProcess` cast to `DarwinProcess*`, which is always valid on Darwin. Will return NULL if no such process exists.

Also avoids this extra variable only used once.

darwin/Platform.c
```diff
@@ -799,85 +799,97 @@ const char* Platform_getRelease(void) {
...
+   io_registry_entry_t entry;
+   while ((entry = IOIteratorNext(iterator))) {
+      io_struct_inband_t buffer;
+      uint32_t size = sizeof(buffer);
+      if (IORegistryEntryGetProperty(entry, "IOUserClientCreator", buffer, &size) != KERN_SUCCESS)
+         goto cleanup;
+
+      pid_t pid;
+      if (sscanf(buffer, "pid %d", &pid) != 1)
+         goto cleanup;
```
BenBE, line 823: Please provide some (short) inline documentation about what the contents of buffers are expected to be. Reference to corresponding documentation is fine; throw a copy to archive.org if possible in case the link breaks.

darwin/Platform.c
```diff
@@ -799,85 +799,97 @@ const char* Platform_getRelease(void) {
...
+         for (size_t i = 0; i < count; ++i) {
+            CFDictionaryRef appUsageEntry = CFArrayGetValueAtIndex(appUsage, i);
+            assert(CFGetTypeID(appUsageEntry) == CFDictionaryGetTypeID());
 
-            assert(CFGetTypeID(appUsage) == CFArrayGetTypeID());
+            CFNumberRef accumulatedGPUTimeRef = CFDictionaryGetValue(appUsageEntry, CFSTR("accumulatedGPUTime"));
+            if (!accumulatedGPUTimeRef) {
+               CFRelease(appUsage);
+               goto cleanup;
+            }
```
BenBE: Shouldn't this just try to continue with the next entry in the `appUsage` array instead of failing hard?

darwin/Platform.c
```diff
@@ -799,85 +799,97 @@ const char* Platform_getRelease(void) {
...
-               assert(CFGetTypeID(appUsageEntry) == CFDictionaryGetTypeID());
+            uint64_t accumulatedGPUTime = 0;
+            CFNumberGetValue(accumulatedGPUTimeRef, kCFNumberLongLongType, &accumulatedGPUTime);
 
-               CFNumberRef accumulatedGPUTimeRef = CFDictionaryGetValue(appUsageEntry, CFSTR("accumulatedGPUTime"));
-               if (!accumulatedGPUTimeRef) {
-                  CFRelease(appUsage);
-                  goto cleanup;
-               }
+            gpuTime += accumulatedGPUTime;
```
BenBE, line 850: Shouldn't this only sum non-negative entries?

darwin/Platform.c
```diff
@@ -799,85 +799,97 @@ const char* Platform_getRelease(void) {
...
-            }
+         uint64_t accumulatedGPUTime = 0;
+         CFNumberGetValue(accumulatedGPUTimeRef, kCFNumberLongLongType, &accumulatedGPUTime);
 
-            CFRelease(appUsage);
-         } else if (IOObjectConformsTo(entry, "IOAccelSubmitter2")) {
-            CFNumberRef accumulatedGPUTimeRef = IORegistryEntryCreateCFProperty(entry, CFSTR("accumulatedGPUTime"), kCFAllocatorDefault, kNilOptions);
-            if (!accumulatedGPUTimeRef)
-               goto cleanup;
+         gpuTime += accumulatedGPUTime;
```
BenBE, line 862: Shouldn't this too be only adding non-negative entries?

darwin/Platform.c
```diff
@@ -799,85 +799,97 @@ const char* Platform_getRelease(void) {
...
+   const Vector* rows = dpt->super.super.rows;
+   int table_size = Vector_size(rows);
+   for (int i = 0; i < table_size; ++i) {
+      DarwinProcess* dp = (DarwinProcess*) Vector_get(rows, i);
+      if (!dp->gpu_time_updated) {
+         dp->gpu_time = 0;
+         dp->gpu_percent = 0.0F;
       }
-      IOObjectRelease(iterator);
    }
```
BenBE: Why not reset these all back in `ProcessTable_goThroughEntries`? They'll stay 0 if no update occured.

Keep the `gpu_time_updated` flag for rendering to shadow/darken the entry if no GPU time was used.

linux/LinuxProcess.c
```diff
@@ -110,7 +110,7 @@ const ProcessFieldData Process_fields[LAST_PROCESSFIELD] = {
 #ifdef SCHEDULER_SUPPORT
    [SCHEDULERPOLICY] = { .name = "SCHEDULERPOLICY", .title = "SCHED ", .description = "Current scheduling policy of the process", .flags = PROCESS_FLAG_SCHEDPOL, },
 #endif
-   [TIME_GPU] = { .name = "GPU_TIME", .title = "GPU TIME+", .description = "Total GPU time", .flags = PROCESS_FLAG_LINUX_GPU, .defaultSortDesc = true, },
+   [TIME_GPU] = { .name = "GPU_TIME", .title = "GPU TIME", .description = "Total GPU time", .flags = PROCESS_FLAG_LINUX_GPU, .defaultSortDesc = true, },
```
BenBE: Given this is meant as an analog to `TIME+` for indicating the total, including the `+` is fine.

darwin/Platform.c
shatyuka: It is an unsigned type, so not necessary.

darwin/Platform.c
shatyuka: Cannot be reset, as the percentage requires calculating the difference from the last gpu time value.

linux/LinuxProcess.c
```diff
@@ -455,14 +455,14 @@ static int LinuxProcess_compareByKey(const Process* v1, const Process* v2, Proce
       return SPACESHIP_NUMBER(p1->autogroup_id, p2->autogroup_id);
    case AUTOGROUP_NICE:
       return SPACESHIP_NUMBER(p1->autogroup_nice, p2->autogroup_nice);
-   case GPU_PERCENT: {
+   case PERCENT_GPU: {
```
Explorer09: I have a question. What motivated you to rename the field from `GPU_PERCENT` to `PERCENT_GPU`? (Similar question to `GPU_TIME` to `TIME_GPU`.)

For me, the renaming of the fields doesn't make them easier to read.

This is beside the compatibility problem I pointed out to you before, which you have fixed. I just wonder why this idea in the first place.

shatyuka: Because I noticed that the previous percentages were all named like `PERCENT_CPU` and `PERCENT_MEM`. Of course, due to compatibility issues, these modifications are now meaningless. Should I revert it back to GPU_PERCENT?

Explorer09: > Because I noticed that the previous percentages were all named like `PERCENT_CPU` and `PERCENT_MEM`.

Oh I see. Thus it was an error when the GPU percentage fields were introduced in htop Linux monitoring.

Because htop had no requirement on whether the word PERCENT should be placed at the start or at the end.

If you `grep -r TIME` in the htop codebase, you would get what I mean: We had a lot of fields with "TIME" word at the end. `STARTTIME`, `CPU_TOTAL_TIME`, `GUEST_TIME`, etc.

> Of course, due to compatibility issues, these modifications are now meaningless. Should I revert it back to GPU_PERCENT?

It's not my call. Maybe BenBE can decide this.

darwin/Platform.c
Explorer09: By the way, I don't like the assumption that `pid_t` always has the same width as `int`. It's better to `sscanf` into an `int`-type buffer then cast to `pid_t` when using.

darwin/Platform.c
Explorer09: @shatyuka If the `gpu_time` stored in `DarwinProcess` is cumulative, you shouldn't reset it here.
Also, you probably don't need the `gpu_time_updated` boolean flag.

My thought is this: First, loop over all DarwinProcess entries and set `gpu_percent` to zero. Then, loop on the accumulatedGPUTimeRef objects and update the `gpu_time` and `gpu_percent` with the respective processes.

The processes that did not have GPU time updated would keep the old `gpu_time` value but `gpu_percent` reset to zero.

shatyuka: @Explorer09 Yes, I've considered that approach as well. My original intention was to handle situations where the GPU context is released by the process.
Keeping the GPU time is also a good solution.

shatyuka: A process can have multiple IOGPUDeviceUserClient entries, so it appears we must add an additional variable to accumulate values during iteration. I overlooked this when deleting the Hashtable, and the current implementation has some issues.

Explorer09: @shatyuka What should happen if a process has multiple IOGPUDeviceUserClient entries? Are those values in the entries supposed to sum up or what?

shatyuka: > @shatyuka What should happen if a process has multiple IOGPUDeviceUserClient entries? Are those values in the entries supposed to sum up or what?

Yes, sum up.

linux/LinuxProcess.c
BenBE: Given there are other fields like `PERCENT_CPU` etc. I actually prefer this change. Otherwise I'd already mention this change. Thus can stay as is rn. ;-)

darwin/Platform.c
BenBE: How is calculating the delta between the previous run and the current one supposed to work in situations where one of the contexts got removed?

shatyuka: The current behavior is that when the total GPU time decreases, the GPU percentage calculation for this process is skipped.

Explorer09: > The current behavior is that when the total GPU time decreases, the GPU percentage calculation for this process is skipped.

I agree with this behavior. When the cumulative GPU time decreases, we "update" our cached value to a lower one but show 0% for the percentage. The next measurement would compute the right percentage then.

BenBE: While this is acceptable, it's not fully correct technically.

Example:
Two components for one process specifying 1000 and 2000 time units were needed previously. Now on the next run one resource is freed and we are left with 2500 as the single usage value for the process. So while we were at 3000 previously and are now at 2500, we had a usage of (at least) 500, but will report 0.

PR #1856, opened 2026-01-08
Darwin: Simplify GPU PerformanceStatistics property code

shatyuka:


darwin/Platform.c
```diff
@@ -365,14 +365,9 @@ void Platform_setGPUValues(Meter* mtr, double* totalUsage, unsigned long long* t
...
-   CFMutableDictionaryRef properties = NULL;
-   kern_return_t ret = IORegistryEntryCreateCFProperties(dhost->GPUService, &properties, kCFAllocatorDefault, kNilOptions);
-   if (ret != KERN_SUCCESS || !properties)
-      return;
-
-   CFDictionaryRef perfStats = CFDictionaryGetValue(properties, CFSTR("PerformanceStatistics"));
+   CFDictionaryRef perfStats = IORegistryEntryCreateCFProperty(dhost->GPUService, CFSTR("PerformanceStatistics"), kCFAllocatorDefault, kNilOptions);
    if (!perfStats)
-      goto cleanup;
+      return;
```
BenBE, line 370: This change misses the final assignment of `mtr->values[0] = *totalUsage;`

shatyuka: ~This is an issue with the original code, which directly returns and skips the assignment after IORegistryEntryCreateCFProperties fails.~
We have `mtr->values[0] = NAN;` before.

PR #1862, opened 2026-01-17
fix: add null check to avoid NetBSD compile warning

### LINKED ISSUE #1662: NetBSD compile warning ("potential null pointer dereference")
This warning is when building htop for OpenBSD, with GCC with `-flto` enabled

```text
lto-wrapper: warning: using serial compilation of 3 LTRANS jobs
DisplayOptionsPanel.c: In function ‘DisplayOptionsPanel_eventHandler’:
DisplayOptionsPanel.c:98:14: error: potential null pointer dereference [-Werror=null-dereference]
   98 |          if (OptionItem_kind(selected) == OPTION_ITEM_NUMBER) {
      |              ^
lto1: all warnings being treated as errors
lto-wrapper: fatal error: gcc returned 1 exit status
```

I actually caught this one in CI ([build log](https://github.com/Explorer09/htop-1/actions/runs/14294810113/job/40060608485)), and yet I don't know how to fix it at the moment.

justinluque:
Adds an extra null check to avoid a compiler warning on NetBSD. Resolves #1662.

### DIFF
--- a/DisplayOptionsPanel.c
+++ b/DisplayOptionsPanel.c
@@ -107,7 +107,7 @@ static HandlerResult DisplayOptionsPanel_eventHandler(Panel* super, int ch) {
          }
          /* fallthrough */
       case EVENT_SET_SELECTED:
-         if (OptionItem_kind(selected) == OPTION_ITEM_NUMBER) {
+         if (selected && OptionItem_kind(selected) == OPTION_ITEM_NUMBER) {
             super->currentBar = this->decIncBar;
          } else {
             Panel_setDefaultBar(super);


BenBE:
Please note the comment by @Explorer09 in the original issue regarding `OptionItem_kind(selected)` and the side effects … A proper fix likely will need slightly more care for the details.

Explorer09:
The point is the issue only appeared in NetBSD build and not in other OS. So the better way is to find out in NetBSD code how the `selected` item can be NULL in the first place.

BenBE:
Should be resolved by 70f873e …

PR #1865, opened 2026-01-18
Allocate COMMAND (cmdline) and comm buffers dynamically

### LINKED ISSUE #1860: F4 filtering does not find all results
htop seems to turncate long command lines, and thus misses results when using the F4 Filter functionality.

* I have a program whose `arvg` joined by ` ` is 22951 characters long, which contains the string `-M100G`.
* F4-filtering in htop for `M100G` does not show this line (but shows its parent process's line, which also happens to contain `-M100G` and is shorter).

Pressing <kbd>W</kbd> on the long process to bring up htop's detail command viewer reveals that that one seems to be cut off after 4173 chars.

`ps aux | grep M100G` finds the long-named process and correctly outputs the full length of its command line.

fasterit:
Fixes #1860

This PR was developed by DLange/Faster IT with Microsoft Copilot

### DIFF
--- a/linux/LinuxProcessTable.c
+++ b/linux/LinuxProcessTable.c
@@ -1,6 +1,7 @@
 /*
 htop - LinuxProcessTable.c
 (C) 2014 Hisham H. Muhammad
+(C) 2020-2026 htop dev team
 Released under the GNU GPLv2+, see the COPYING file
 in the source distribution for its full text.
 */
@@ -58,6 +59,9 @@ in the source distribution for its full text.
 #ifndef PF_KTHREAD
 #define PF_KTHREAD 0x00200000
 #endif
+
+/* Maximum buffer size for reading COMMAND / comm */
+#define MAX_CMDLINE_BUFFER_SIZE (2 * 1024 * 1024 + 1)
 
 /* Inode number of the PID namespace of htop */
 static ino_t rootPidNs = (ino_t)-1;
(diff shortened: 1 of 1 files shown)


fasterit:
I left the initial buffer at 4096 +1 bytes so we can compare performance. I suggest we ramp up from 512 bytes if/when we commit this to main as we can save memory this way and it comes rather "for free" with the dynamic allocation.

Explorer09:
@fasterit This thing is very similar to the dynamic buffer code I've done in #1821.

My code snippet ([generic/Demangle.c](https://github.com/Explorer09/htop-1/blob/131d2387b3366a90744c6bb832d019ee87f1291f/generic/Demangle.c)):

```c
char* Generic_Demangle(const char* mangled) {
   // The cplus_demangle() API from libdemangle is flawed. It does not
   // provide us the required size of the buffer to store the demangled
   // name, and it leaves the buffer content undefined if the specified
   // buffer size is not big enough.
   static size_t bufSize = 128;
   static char* buf;

   while (true) {
      if (!buf) {
         buf = malloc(bufSize);
         if (!buf)
            return NULL;
      }

      int res = cplus_demangle(mangled, buf, bufSize);
      if (res == 0)
         break;
      if (res != DEMANGLE_ESPACE || bufSize > SIZE_MAX / 2)
         return NULL;

      bufSize *= 2;
      free(buf);
      buf = NULL;
   }

   return strdup(buf);
}
```

Notice that I keep the buffer and not deallocate it after use. I think this can save some time when there are lots of strings needed to fetch (and all of them need a dynamically-sized temporary buffer). I don't know if it is a good idea.

linux/LinuxProcessTable.c
```diff
@@ -59,6 +60,9 @@ in the source distribution for its full text.
 #define PF_KTHREAD 0x00200000
 #endif
 
+/* Maximum buffer size for reading COMMAND / comm */
+#define MAX_CMDLINE_BUFFER_SIZE (2 * 1024 * 1024 + 1)
```
BenBE: Given the 4K+1 is multiplied a bunch of times, by the time we reach this size, it's more like 2M+512.

fasterit: Yes, see my [comment](https://github.com/htop-dev/htop/pull/1865#issuecomment-3765289691) on suggesting to start with 512 bytes in production

Explorer09:
I have a question: How does this solve the original bug (#1860) when the new proposed solution still has a length limit?

fasterit:
@Explorer09 Because the length limit is 2 MB now, not 4096 Bytes. As much as I know currently, 2MB seems to be the practical limit of cmdlines in the Linux kernel these days. At least `xargs` thinks so, see [this comment in the linked issue](https://github.com/htop-dev/htop/issues/1860#issuecomment-3765235255).

Explorer09:
> @Explorer09 Because the length limit is 2 MB now, not 4096 Bytes. As much as I know currently, 2MB seems to be the practical limit of cmdlines in the Linux kernel these days. At least `xargs` thinks so, see [this comment in the linked issue](https://github.com/htop-dev/htop/issues/1860#issuecomment-3765235255).

I am uncomfortable when the whole 2 MB long command line is kept in htop's memory. If we keep that line in memory only to let the filter function work, I think an alternative approach is to read and "grep" from the file in procfs on demand.

BenBE:
> > @Explorer09 Because the length limit is 2 MB now, not 4096 Bytes. As much as I know currently, 2MB seems to be the practical limit of cmdlines in the Linux kernel these days. At least `xargs` thinks so, see [this comment in the linked issue](https://github.com/htop-dev/htop/issues/1860#issuecomment-3765235255).
> 
> I am uncomfortable when the whole 2 MB long command line is kept in htop's memory. If we keep that line in memory only to let the filter function work, I think an alternative approach is to read and "grep" from the file in procfs on demand.

Having 2MiB in memory is much better from a performance PoV than having to grep 2000+ files just in case one of the processes matches something. It's a trade-of between memory vs. repeatedly doing several syscalls just to safe a few kB of memory. You can still do "keep the buffer for next read and strdup for internal storage".

fasterit:
Added the `Assisted-by:` line as per our [AI-Assisted Contributions Policy](https://github.com/htop-dev/htop/blob/main/docs/ai-contributions-policy.md).

linux/LinuxProcessTable.c
BenBE: Issue is, the loop in line 1280 will see `bufferSize == 2M + 512` and thus skip the 2MiB round despite having the maximum still allow for a slightly smaller buffer. Thus updating the maximum to be 2MiB + 512 makes the buffer actually reach to what it could be for its maximal size.

BenBE (approved):

PR #1868, opened 2026-01-22
Message when send signal fails

### LINKED ISSUE #1863: Attempting to terminate a process without permissions doesn't produce a warning
I tried using htop to kill a process and didn't know why it kept failing.

When attempting to terminate/kill an elevated process, and htop isn't running in sudo, it just does nothing on pressing Enter. There's no feedback. In the same situation, btop++ informs you it doesn't have permissions to end the process.

DeltachangeOG:
Closes #1863 

Changed send signal and related handlers to preserve errno. Message detailing permission denied for permission issues and generic failed to send signal for all others.

No change of successful behavior.

### DIFF
--- a/Action.c
+++ b/Action.c
@@ -14,6 +14,7 @@ in the source distribution for its full text.
 #include <stdbool.h>
 #include <stdlib.h>
 #include <string.h>
+#include <errno.h>
 
 #include "CRT.h"
 #include "CategoriesPanel.h"
@@ -104,8 +105,8 @@ static void Action_runSetup(State* st) {
 
 static bool changePriority(MainPanel* panel, int delta) {
    bool anyTagged;
-   bool ok = MainPanel_foreachRow(panel, Process_rowChangePriorityBy, (Arg) { .i = delta }, &anyTagged);
-   if (!ok)
+   int rc = MainPanel_foreachRow(panel, Process_rowChangePriorityBy, (Arg) { .i = delta }, &anyTagged);
+   if (rc != 0)
       beep();
    return anyTagged;
 }
@@ -456,8 +457,8 @@ static Htop_Reaction actionSetAffinity(State* st) {
    const void* set = Action_pickFromVector(st, affinityPanel, width, true);
    if (set) {
       Affinity* affinity2 = AffinityPanel_getAffinity(affinityPanel, host);
-      bool ok = MainPanel_foreachRow(st->mainPanel, Affinity_rowSet, (Arg) { .v = affinity2 }, NULL);
-      if (!ok)
+      int rc = MainPanel_foreachRow(st->mainPanel, Affinity_rowSet, (Arg) { .v = affinity2 }, NULL);
+      if (rc != 0)
          beep();
       Affinity_delete(affinity2);
    }
@@ -502,8 +503,8 @@ static Htop_Reaction actionSetSchedPolicy(State* st) {
 
       SchedulingArg v = { .policy = preSelectedPolicy, .priority = preSelectedPriority };
 
-      bool ok = MainPanel_foreachRow(st->mainPanel, Scheduling_rowSetPolicy, (Arg) { .v = &v }, NULL);
-      if (!ok)
+      int rc = MainPanel_foreachRow(st->mainPanel, Scheduling_rowSetPolicy, (Arg) { .v = &v }, NULL);
+      if (rc != 0)
          beep();
    }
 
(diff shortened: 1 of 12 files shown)


Explorer09:
This change is too invasive. And part of this feature overlaps with #1231. I expect there should some collaborations between two PRs.

Also, this code seems vibe coded. Please disclose any use of AI in the process.

linux/LinuxProcess.c
```diff
@@ -169,15 +170,21 @@ IOPriority LinuxProcess_updateIOPriority(Process* p) {
...
+   long rc = syscall(SYS_ioprio_set, IOPRIO_WHO_PROCESS, Process_getPid(p), ioprio.i);
+   if (rc == -1)
+      return -errno;
 #endif
-   return LinuxProcess_updateIOPriority(p) == ioprio.i;
+   // Fallback / verify
+   if (LinuxProcess_updateIOPriority(p) == ioprio.i)
+      return 0;
+   
+   return -1;
```
Explorer09, line 184: If you're using `-errno` as a return value, don't simultaneously use an unnamed constant `-1` for an unknown error. In Linux the `errno` value of 1 would be interpreted as `EPERM`, which is probably not what you want.

This logic mistake should never happen if a human programmer wrote it. Chances are this code is AI generated.

linux/LinuxProcess.c
```diff
@@ -190,29 +197,32 @@ bool LinuxProcess_isAutogroupEnabled(void) {
...
    if (!file)
-      return false;
+      return -errno;
 
    long int identity;
    int nice;
    int ok = fscanf(file, "/autogroup-%ld nice %d", &identity, &nice);
-   bool success = false;
+
+   int rc = -1;
```
Explorer09, line 213: Ditto. No `-errno` and `-1` return values in the same code.

Action.c
```diff
@@ -526,11 +527,20 @@ static Htop_Reaction actionKill(State* st) {
       Panel_setHeader((Panel*)st->mainPanel, "Sending...");
       Panel_draw((Panel*)st->mainPanel, false, true, true, State_hideFunctionBar(st));
       refresh();
-      bool ok = MainPanel_foreachRow(st->mainPanel, Process_rowSendSignal, (Arg) { .i = sgn->key }, NULL);
-      if (!ok) {
-         beep();
+      int rc = MainPanel_foreachRow(st->mainPanel, Process_rowSendSignal, (Arg) { .i = sgn->key }, NULL);
+      if (rc != 0) {
+         int err = (rc < 0) ? -rc : rc;
```
Explorer09, line 532: Why this? Logically it doesn't make sense. You either use a positive value range for `errno` or a negative range for it. Not simultaneously.

linux/LinuxProcess.c
DeltachangeOG: I am still new to C and learning. I did the split in two sections. one I just changed logic to be 0 for success and -1 for failure. Then I added errno in and didn't catch all the places I changed it. I am not vibe coding I am just new and self taught, and don't know all of the conventions / correct ways to do things yet.

Action.c
```diff
@@ -526,11 +527,20 @@ static Htop_Reaction actionKill(State* st) {
...
+         if (err != ESRCH)
+            beep();
+         if (err == EPERM) {
+            Panel_setHeader((Panel*)st->mainPanel, "Permission denied (try running as root)");
```
Explorer09, line 536: strerror(3) is your friend.

linux/LinuxProcess.c
Explorer09: > I am still new to C and learning. I did the split in two sections. one I just changed logic to be 0 for success and -1 for failure. Then I added errno in and didn't catch all the places I changed it. I am not vibe coding I am just new and self taught, and don't know all of the conventions / correct ways to do things yet.

You can make small changes as separate commits so one commit does one thing. That way you can also revert or manage changes if you make a mistake in the middle. That's how VCS (version control system) can help.

Action.c
DeltachangeOG: it is a guard to ensure rc is always negative because I don't fully know the codebase or things that could change errno.

linux/LinuxProcess.c
DeltachangeOG: I thought this was all one thing. 1231 is several years old at this point, how can I tell what is an active PR with work being done on it vs not. I am completely open to feedback and willing to implement changes.

Thank you for the time  you spent to respond.

PR #1869, opened 2026-01-22
Add total seconds to uptime meter

powerpellet1077:
This will add a total seconds count appended to the uptime counter. This would help me (and others) for viewing the live system uptime in seconds, as it is used for the kernel logs and other internal counters. Thank you for your time and consideration.

Preview below:
image

### DIFF
--- a/UptimeMeter.c
+++ b/UptimeMeter.c
@@ -25,6 +25,7 @@ static void UptimeMeter_updateValues(Meter* this) {
       xSnprintf(this->txtBuffer, sizeof(this->txtBuffer), "(unknown)");
       return;
    }
+
    int seconds = totalseconds % 60;
    int minutes = (totalseconds / 60) % 60;
    int hours = (totalseconds / 3600) % 24;
@@ -56,5 +57,31 @@ const MeterClass UptimeMeter_class = {
    .attributes = UptimeMeter_attributes,
    .name = "Uptime",
    .uiName = "Uptime",
+   .caption = "Uptime: "
+};
+
+
+static void SecondsUptimeMeter_updateValues(Meter* this) {
+   int totalseconds = Platform_getUptime();
+   if (totalseconds <= 0) {
+      xSnprintf(this->txtBuffer, sizeof(this->txtBuffer), "(unknown)");
+      return;
+   }
+   xSnprintf(this->txtBuffer, sizeof(this->txtBuffer), "%d s", totalseconds);
+}
+
+const MeterClass SecondsUptimeMeter_class = {
+   .super = {
+      .extends = Class(Meter),
+      .delete = Meter_delete
+   },
+   .updateValues = SecondsUptimeMeter_updateValues,
+   .defaultMode = TEXT_METERMODE,
+   .supportedModes = (1 << TEXT_METERMODE) | (1 << LED_METERMODE),
+   .maxItems = 0,
+   .total = 0.0,
+   .attributes = UptimeMeter_attributes,
+   .name = "SecondsUptime",
+   .uiName = "Uptime (seconds)",
    .caption = "Uptime: "
 };
--- a/UptimeMeter.h
+++ b/UptimeMeter.h
@@ -11,5 +11,7 @@ in the source distribution for its full text.
 
 
 extern const MeterClass UptimeMeter_class;
+extern const MeterClass SecondsUptimeMeter_class;
+
 
 #endif
--- a/unsupported/Platform.c
+++ b/unsupported/Platform.c
@@ -59,6 +59,7 @@ const MeterClass* const Platform_meterTypes[] = {
    &HostnameMeter_class,
    &SysArchMeter_class,
    &UptimeMeter_class,
+   &SecondsUptimeMeter_class,
    &AllCPUsMeter_class,
    &AllCPUs2Meter_class,
    &AllCPUs4Meter_class,
--- a/darwin/Platform.c
+++ b/darwin/Platform.c
@@ -129,6 +129,7 @@ const MeterClass* const Platform_meterTypes[] = {
    &HostnameMeter_class,
    &SysArchMeter_class,
    &UptimeMeter_class,
+   &SecondsUptimeMeter_class,
    &AllCPUsMeter_class,
    &AllCPUs2Meter_class,
    &AllCPUs4Meter_class,
--- a/dragonflybsd/Platform.c
+++ b/dragonflybsd/Platform.c
@@ -105,6 +105,7 @@ const MeterClass* const Platform_meterTypes[] = {
    &SwapMeter_class,
    &TasksMeter_class,
    &UptimeMeter_class,
+   &SecondsUptimeMeter_class,
    &BatteryMeter_class,
    &HostnameMeter_class,
    &SysArchMeter_class,
--- a/freebsd/Platform.c
+++ b/freebsd/Platform.c
@@ -112,6 +112,7 @@ const MeterClass* const Platform_meterTypes[] = {
    &MemorySwapMeter_class,
    &TasksMeter_class,
    &UptimeMeter_class,
+   &SecondsUptimeMeter_class,
    &BatteryMeter_class,
    &HostnameMeter_class,
    &SysArchMeter_class,
--- a/linux/Platform.c
+++ b/linux/Platform.c
@@ -222,6 +222,7 @@ const MeterClass* const Platform_meterTypes[] = {
    &HugePageMeter_class,
    &TasksMeter_class,
    &UptimeMeter_class,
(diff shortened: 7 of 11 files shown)


natoscott:
@powerpellet1077 no major objection here, though that number is often going to be alot larger than four digits in practice - I wonder if it may become a little distracting.  Another option would be to create a new Meter for this UI change, similar to the way the DateMeter.c and DateTimeMeter.c handle their differences in output.

Explorer09:
I don't like two counters on the same meter. It's as if the users are too lazy to perform the unit conversions themselves. If you want it, it's better to implement it as a separate meter. E.g. `UptimeSecondsMeter`. Both the `UptimeMeter` and the proposed `UptimeSecondsMeter` can share a lot of code internally and reside in one C file.

powerpellet1077:
@Explorer09 @natoscott 
I completely agree with both statements, this change was made without detailed enough knowledge of C to make something complicated as that. If you would like, you could change/modify my pull request or simply open a new one, as I don't have the capabilities to make something of that level atm. Thank you for your consideration anyways!

BenBE:
The basic idea LGTM, but there's a few minor details to work out:
- @natoscott Even with ~3 years of runtime you still only have about 8 digits. Typical runtimes (~100 days) will have 6-7 digits.
- Splitting this into its own meter is probably the best, both for existing users who prefer the current behaviour and users that want to chose which layout they want explicitly. The alternative would be a setting somewhere, which is generally discouraged unless there are good alternatives
- The output of the meter probably should be structured using normal text and bold for highlights. E.g. UpSec: **420**(!) days, **23:23:23** (**36372203** seconds)
- Code should be re-used where possible.
- Don't open a new PR for the same feature/enhancemet; update this existing PR by rebasing/updating its commits to accommodate changes instead.

powerpellet1077:
I managed to add the SecondsUptimeMeter to the repo. I would recommend looking over my changes to make sure it works best for the team.

EDIT: Also, I did not change any copyright comments, I don't know if this will need modification in time.

SecondsUptimeMeter.c
BenBE, line 1: Why not keep the changes as part of the normal UptimeMeter implementation; as suggested in the discussion?

powerpellet1077: > Splitting this into its own meter is probably the best, both for existing users who prefer the current behaviour and users that want to chose which layout they want explicitly. The alternative would be a setting somewhere, which is generally discouraged unless there are good alternatives

I am confused, I thought you said it would be best for a separate meter object? The original implementation was replaced by the new meter object, so I deleted the original edits. Please let me know what works best for you.

powerpellet1077:
@BenBE 
Apologies, I do not have a lot of experience with git as of now. I think I rebased it properly. I still left the first commit available so I could preview what I changed if future edits are needed. I will squash when ready for pull.

SecondsUptimeMeter.c
Explorer09: @powerpellet1077 Sometimes we put meter classes with similar functions into the same `.c` source file. For an example you can see `LoadMeter` and `LoadAverageMeter` both in `LoadAverageMeter.c` (and `LoadAverageMeter.h`).

powerpellet1077: @Explorer09 This does make sense. Is this the only change that needs to be made?

Explorer09: @powerpellet1077 Move your `SecondsUptimeMeter` code into the `UptimeMeter.{c,h}` file and the things would be mostly ready to merge.

powerpellet1077: @Explorer09 
Nice, I will get that ready. Is it also fine if I make one incorporating the two values together (as proposed originally)? It may take longer if I add that as well. Please let me know what works best.

Explorer09: @powerpellet1077 It's not my use case to have the number of seconds in the Uptime meter. I am okay with a "number of seconds only" uptime meter, but showing both units would be an overkill.

Note: These time & clock related meters may be displayed with LARGE, 7-segment-like digits in the "LED" display mode.

powerpellet1077: @Explorer09 
I understand, thank you for the help! I will be committing the final changes and squashing.

unsupported/Platform.c
```diff
@@ -27,7 +27,6 @@ in the source distribution for its full text.
 #include "TasksMeter.h"
 #include "UptimeMeter.h"
 
-
```
Explorer09, line 30: This double-newlines is intentional. See the style guide. Don't change it.

UptimeMeter.h
```diff
@@ -9,7 +9,8 @@ in the source distribution for its full text.
 
 #include "Meter.h"
 
-
```
Explorer09, line 12: Ditto. (Keep the two newlines per the style guide.)

UptimeMeter.c
```diff
@@ -58,3 +59,33 @@ const MeterClass UptimeMeter_class = {
...
+static void SecondsUptimeMeter_updateValues(Meter* this) {
+   int totalseconds = Platform_getUptime();
+   if (totalseconds <= 0) {
+      xSnprintf(this->txtBuffer, sizeof(this->txtBuffer), "(unknown)");
+      return;
+   }
+   xSnprintf(this->txtBuffer, sizeof(this->txtBuffer), "%d s", totalseconds);
+}
+
+const MeterClass SecondsUptimeMeter_class = {
```
Explorer09, line 73: Just curious, how did you come up with the name `SecondsUptimeMeter` while I initially suggested `UptimeSecondsMeter`? I'm not suggesting a name change, I'm just asking the reason of your naming.

unsupported/Platform.c
powerpellet1077: Apologies for both. I just naturally changed the newlines because it was what I was use to. Would you like me to modify it and fix it?

UptimeMeter.c
powerpellet1077: I honestly completely forgot. There might have been a better reason at the time, but I think I chose the name without thinking too much on it. I guess it works though.

unsupported/Platform.c
Explorer09: Yes. Change it back when you rebase your commits.

powerpellet1077: Fixed. Thanks for the help!

natoscott:
| EDIT: Also, I did not change any copyright comments, I don't know if this will need modification in time.

@powerpellet1077 no need to change copyright notice for this sort of (relatively small) change.

UptimeMeter.c
```diff
@@ -58,3 +59,33 @@ const MeterClass UptimeMeter_class = {
...
+      .extends = Class(Meter),
+      .delete = Meter_delete
+   },
+   .updateValues = SecondsUptimeMeter_updateValues,
+   .defaultMode = TEXT_METERMODE,
+   .supportedModes = (1 << TEXT_METERMODE) | (1 << LED_METERMODE),
+   .maxItems = 0,
+   .total = 0.0,
+   .attributes = SecondsUptimeMeter_attributes,
+   .name = "Uptime (seconds)",
```
Explorer09: This `name` field is for identification in the config file. No space should be used in this field.

```suggestion
   .name = "SecondsUptime",
```

UptimeMeter.c
```diff
@@ -58,3 +59,33 @@ const MeterClass UptimeMeter_class = {
...
+   .super = {
+      .extends = Class(Meter),
+      .delete = Meter_delete
+   },
+   .updateValues = SecondsUptimeMeter_updateValues,
+   .defaultMode = TEXT_METERMODE,
+   .supportedModes = (1 << TEXT_METERMODE) | (1 << LED_METERMODE),
+   .maxItems = 0,
+   .total = 0.0,
+   .attributes = SecondsUptimeMeter_attributes,
```
Explorer09: Save code size 😉

```suggestion
   .attributes = UptimeMeter_attributes,
```

UptimeMeter.c
powerpellet1077: Apologies, I did not know what this was for and followed the naming scheme I saw from the original UptimeMeter. I will fix and commit this change soon.

UptimeMeter.c
powerpellet1077: Changing and comitting soon. Thank you!

powerpellet1077:
@Explorer09 
Fixed!

natoscott:
Chatting to @BenBE on IRC yesterday, consensus was the remaining "probably should be structured using normal text and bold for highlights" issue issue can be a follow-up commit if there's interest.  I'll merge shortly and also include Benny's request to consolidate the date/datetime meters into a single file, like was done here.

PR #1876, opened 2026-01-24
Implement explicit NO_COLOR env support

### LINKED ISSUE #1875: htop should respect NO_COLOR env var
Hi!

Since many other programs do too. So, `NO_COLOR=1 htop` should do the same as `htop --no-color`.

BenBE:
This provides further options to disable color alongside the existing:
- `--no-color` command line argument (since 2010-02-25, 62a31673c)
- Using  "monochrome" color scheme in the Setup screen (f5f29ceeb)

Fixes #1875

### DIFF
--- a/CRT.c
+++ b/CRT.c
@@ -1327,6 +1327,10 @@ void CRT_enableDelay(void) {
 }
 
 void CRT_setColors(int colorScheme) {
+   if (colorScheme >= LAST_COLORSCHEME || colorScheme < 0) {
+      colorScheme = COLORSCHEME_DEFAULT;
+   }
+
    CRT_colorScheme = colorScheme;
 
    for (short int i = 0; i < 8; i++) {


fasterit:
Please update to not save this setting:
```bash
NO_COLOR=1 ./htop
#do something that changes config, e.g. press `t`
#quit htop
./htop
#--> Monochrome mode persists
```

BenBE:
This is the same behaviour/bug that also happens with `htop --no-color` / `htop -C`.

Fehlerbestandsschutz™

fasterit:
^ There you go @BenBE, fixing the past. No free pass for htop folklore (~historic mistakes) today :smile:

fasterit:
^ Simplified a bit

CommandLine.c
```diff
@@ -121,8 +121,9 @@ static CommandLineStatus parseArguments(int argc, char** argv, CommandLineSettin
 
    {
       const char* no_color = getenv("NO_COLOR");
```
Explorer09, line 124: Can we add a code comment to reference https://no-color.org/ for the specification?

ColorsPanel.c
```diff
@@ -68,6 +68,7 @@ static HandlerResult ColorsPanel_eventHandler(Panel* super, int ch) {
          CheckItem_set((CheckItem*)Panel_get(super, mark), true);
 
          this->settings->colorScheme = mark;
+         this->settings->colorSchemeOverridden = false;
```
Explorer09: My idea: Set both `this->settings->colorScheme` and `this->settings->colorSchemeFromConfig` to the same value here, and technically we won't need an additional `colorSchemeOverridden` boolean flag.

CommandLine.c
fasterit: sure

ColorsPanel.c
fasterit: Hm, this is a clever hack but I think it makes the functionality much less discoverable, so I'd lean towards "no".
But this is @BenBE's PR ... what do you think?

BenBE: Not quite happy with the current implementation of the second commit.

Why not init `colorSchemeChanged=false` always, and set to true, once changed in the settings. When writing the config, write `colorSchemeFromConfig` if unchanged, otherwise write the current (active) color scheme from `colorScheme`.

Note the naming of the flag AND its usage.

Explorer09: @BenBE 

The current color scheme is in `CRT_colorScheme` global variable.

I'm favoring _not_ introducing any flag variable or field in `Setting` because any extra field is technically redundant.

<s>I can simplify the color scheme override code so it works something like this:</s> (Update: see #1880)

```c
   // In CommandLine_run()
   CRT_init(settings, flags.allowUnicode, flags.iterationsRemaining != -1);
   if (!flags.useColors) {
      // Color scheme override here
      // If CRT_colorScheme != settings->colorScheme it means there's an override.
CRT_setColors(COLORSCHEME_MONOCHROME);
   }
```

fasterit: > Why not init `colorSchemeChanged=false` always, and set to true, once changed in the settings. When writing the config, write `colorSchemeFromConfig`if unchanged, otherwise write the current (active) color scheme from`colorScheme`.

I like that idea! Go for it @BenBE

CommandLine.c
```diff
@@ -393,6 +391,12 @@ int CommandLine_run(int argc, char** argv) {
    host->iterationsRemaining = flags.iterationsRemaining;
    CRT_init(settings, flags.allowUnicode, flags.iterationsRemaining != -1);
 
+   if (!flags.useColors) {
+      // Override color scheme here so that 'settings' can keep the
+      // original color scheme when desirable.
+      CRT_setColors(COLORSCHEME_MONOCHROME);
```
Explorer09: @BenBE I found another solution and this one could be cleaner than what I originally proposed in #1880.

This solution would not call `CRT_setColors` twice during init.

```c
   int colorSchemeFromConfig = settings->colorScheme;
   if (!flags.useColors)
      settings->colorScheme = COLORSCHEME_MONOCHROME;

   // ...

   CRT_init(settings, flags.allowUnicode, flags.iterationsRemaining != -1);

   // The color scheme override is not saved to the `htoprc` file.
   // 'settings' will keep the original color scheme until the user
   // changes it further in the Setup.
   settings->colorScheme = colorSchemeFromConfig;
```

fasterit: @Explorer09: Please make that into a proper patch so it can be joined here

Explorer09: @fasterit , I rebased #1880 for the new version.

PR #1877, opened 2026-01-24
Feature/kill perm warning

DeltachangeOG:
Displays context in panel for failure of sendSignal. Narrowly scoped in reference of #1863. Explicit Permission denied or strerror. Currently prioritizes EPERM failures.

LLM used to check code syntax and verify understanding of code changes.

### DIFF
--- a/Action.c
+++ b/Action.c
@@ -535,7 +535,7 @@ static Htop_Reaction actionKill(State* st) {
       (void) ok;
       if (ctx.sawEperm) {
          beep();
-         Panel_setHeader((Panel*)st->mainPanel, "Permission denied (try running as root)");
+         Panel_setHeader((Panel*)st->mainPanel, "Permission denied");
          Panel_draw((Panel*)st->mainPanel, false, true, true, State_hideFunctionBar(st));
          refresh();
          napms(1500);


Action.c
```diff
@@ -526,11 +526,25 @@ static Htop_Reaction actionKill(State* st) {
       Panel_setHeader((Panel*)st->mainPanel, "Sending...");
       Panel_draw((Panel*)st->mainPanel, false, true, true, State_hideFunctionBar(st));
       refresh();
-      bool ok = MainPanel_foreachRow(st->mainPanel, Process_rowSendSignal, (Arg) { .i = sgn->key }, NULL);
-      if (!ok) {
+      bool ok = MainPanel_foreachRow(st->mainPanel, Process_rowSendSignal, (Arg) { .v = &ctx }, NULL);
+      (void) ok;
+      if (ctx.sawEperm) {
          beep();
+         Panel_setHeader((Panel*)st->mainPanel, "Permission denied (try running as root)");
```
Explorer09: I suggest ***not*** to inform the user about running as root. Just say "Permission denied" (`strerror(EPERM)`) is enough.

The reason? The user running htop might not have the privilege to run as root (or `sudo`). It's not the business of htop developers to tell what the user should do when they lack a privilege. Similarly we won't inform the user to "tell your manager as you don't have the privilege to do X".

DeltachangeOG: Easily fixed. Some applications I see offer the suggestion and some don't. Thank you for the feedback

DeltachangeOG:
Closing as #1879 addresses this PR with framework for other errors as well.

PR #1878, opened 2026-01-24
Add Tctl temperature reading

### LINKED ISSUE #1776: Support Tctl temps in AMD CPUs without Tccd temps
Hello - this is more of a feature request.

Unfortunately, I noticed when checking htop I was not getting the temps from the CPU exposed. Poking around, I realized my laptop only shows Tctl temps. It appears that if a bit isn't set to say Tccd data is valid, hwmon ignores it ([relevant code](https://github.com/torvalds/linux/blob/22bdd6e68bbe270a916233ec5f34a13ae5e80ed9/drivers/hwmon/k10temp.c#L392)]

Is there any way to add a check for this? I noticed two places in the code that reference k10 so I'm not sure what all would need to be changed, but it looks like this may be the [place](https://github.com/htop-dev/htop/blob/7d59e0463fffac23c0b37bfff5d5909a46f8f9b9/linux/LibSensors.c#L362)

Either way, it's nothing urgent. It would be nice to see my CPU temps in htop so I don't have to use something like watch. 

Thanks!

BarbUk:
For AMD k10temp, when Tccd is not present, we only have a Tctl value:

```
k10temp-pci-00c3
Adapter: PCI adapter
Tctl:         +39.1°C
```

I populate only `data[1]`, htop's existing fallback logic will copy this temperature to all individual cores.

Fixes #1776

### DIFF
--- a/linux/LibSensors.c
+++ b/linux/LibSensors.c
@@ -368,6 +368,17 @@ void LibSensors_getCPUTemperatures(CPUData* cpus, unsigned int existingCPUs, uns
                   }
                }
                ccdID++;
+            }
+
+            /* AMD k10temp with only one general Tctl */
+            else if (String_eq(label, "Tctl")) {
+               for (size_t i = 0; i <= existingCPUs; i++) {
+                  if (isNaN(data[i])) {
+                     data[i] = temp;
+                     if (i > 0)
+                        coreTempCount++;
+                  }
+               }
             } else {
                skip = false;
             }


BenBE:
Shouldn't this check that no value had been set there before? Otherwise a Tctl could overwrite the Tccd temperature.

BarbUk:
> Shouldn't this check that no value had been set there before?

I was understanding the loop like that:
 - set temp for `data[1]` if `Tctl` exists
 - if after we have a `TccdX` value, it will overwrite the temp

I changed the logic in https://github.com/htop-dev/htop/pull/1878/changes/3ee20f9ab32e84c76c37bd5adf66e035c9fa0c3e:
- Keep the `TccdX` handling first
- Else if we have a `Tctl` label, check if `data[1]` does not have a temp defined to use this temp.

This should be better ?

Tested on a AMD Ryzen 7 PRO 6850U with:
```bash
❯ sensors k10temp-pci-00c3 -c /dev/null
k10temp-pci-00c3
Adapter: PCI adapter
Tctl:         +46.6°C
``` 

image

BarbUk:
This is not ready, I'll convert it to draft as I continue testing.

BarbUk:
I did some gdb to understand how the coreTempCount and data array are allocated.

I think https://github.com/htop-dev/htop/pull/1878/changes/7fefa103c1794f94fd7954a2f6cde4563d97ae84 will work better:
 - if we have a tctl label
 - and no package or core temp already defined
 - we use the tctl temp for all cores

linux/LibSensors.c
```diff
@@ -373,6 +368,13 @@ void LibSensors_getCPUTemperatures(CPUData* cpus, unsigned int existingCPUs, uns
                   }
                }
                ccdID++;
+            }
+
+            /* AMD k10temp with only one general Tctl */
+            else if (String_startsWith(label, "Tctl")) {
+               if (isNaN(data[1])) {
+                  data[1] = temp;
+               }
```
BenBE, line 381: Why move this code block?

Also, given that we iterate over all attributes in any order, it doesn't quite matter, which check comes first.

Only ensure, that Tccd always overwrites a value from Tctl, but not vice versa.

BenBE:
Would you mind squashing/rebasing your changes into one commit (cf. style guide)?

The fallback only happens for all cores, if there were no Tccd values in between. If you get Tctl, Tccd for cores 3 and 4 on n 8 core system, Tctl will only be written to cores 1 and 2, while Tccd4 is copied to cores 4 through 8. Thus best to fill in the Tctl in a loop for all cores directly (if not yet set otherwise).

BarbUk:
> Would you mind squashing/rebasing your changes into one commit (cf. style guide)?

Yes, it's done.

> The fallback only happens for all cores, if there were no Tccd values in between. If you get Tctl, Tccd for cores 3 and 4 on n 8 core system, Tctl will only be written to cores 1 and 2, while Tccd4 is copied to cores 4 through 8. Thus best to fill in the Tctl in a loop for all cores directly (if not yet set otherwise).

Ryzen CPUs report per-CCD temp.
The code from checking from Tccd should map a `data[i]` for each core from what I understand ?

The last version of the PR only add a temperature from Tctl in a loop for all core if not already set.

BarbUk:
I have access to a new server with a:
```
processor	: 0
vendor_id	: AuthenticAMD
cpu family	: 25
model		: 97
model name	: AMD EPYC 4244P 6-Core Processor
stepping	: 2
```

Sensors:
```
k10temp-pci-00c3
Adapter: PCI adapter
Tctl:         +78.2°C  
Tccd1:        +64.0°C
```

image

I've put a breakpoint on `linux/LibSensors.c:363`, and on `linux/LibSensors.c:374`.
I've seen:
 - `Tctl` come first, the `data[]` array is still null, the temp are filled with `Tctl`
 - `Tcdd1` come after, the `data[]` is replaced with the tccd1 temp

image


**My question is**: Should I still set the `data[0]` (package temperature) or only loop from 1 to `existingCPUs` ?

BenBE:
I'd have to look into the code again how other CPUs are handled, but if Tccd and Tctl are available, it's probably a good idea to have Tctl in data[0] too. After all, if no other core overwrites it, `data[0]` is calculated from the CPU cores as the average temperature, i.e. is filled in. So if we get an explicit package temperature we might as well use it directly without inferring it if we don't have to.

BarbUk:
> it's probably a good idea to have Tctl in data[0] too

Perfect, it's what i've done.

PR #1880, opened 2026-01-26
Don't let '--no-color' affect htoprc

Explorer09:
Delay the color scheme override by '--no-color' option to after the CRT_init call and don't make it change the 'settings' object.

Also included in this patch is a minor simplification of CRT_init() code.

(Related to #1876)

### DIFF
--- a/CRT.c
+++ b/CRT.c
@@ -1192,8 +1192,6 @@ void CRT_init(const Settings* settings, bool allowUnicode, bool retainScreenOnEx
    redirectStderr();
    noecho();
    CRT_settings = settings;
-   CRT_colors = CRT_colorSchemes[settings->colorScheme];
-   CRT_colorScheme = settings->colorScheme;
 
    for (int i = 0; i < LAST_COLORELEMENT; i++) {
       unsigned int color = CRT_colorSchemes[COLORSCHEME_DEFAULT][i];
@@ -1260,9 +1258,8 @@ IGNORE_WCASTQUAL_END
    CRT_installSignalHandlers();
 
    use_default_colors();
-   if (!has_colors())
-      CRT_colorScheme = COLORSCHEME_MONOCHROME;
-   CRT_setColors(CRT_colorScheme);
+
+   CRT_setColors(has_colors() ? settings->colorScheme : COLORSCHEME_MONOCHROME);
 
 #ifdef HAVE_LIBNCURSESW
    if (allowUnicode && String_eq(nl_langinfo(CODESET), "UTF-8")) {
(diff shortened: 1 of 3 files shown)


fasterit:
Small caveat: The settings show the non-active color scheme. Would this confuse users?

ColorsPanel.c
```diff
@@ -104,6 +104,6 @@ ColorsPanel* ColorsPanel_new(Settings* settings) {
    for (int i = 0; ColorSchemeNames[i] != NULL; i++) {
       Panel_add(super, (Object*) CheckItem_newByVal(ColorSchemeNames[i], false));
    }
-   CheckItem_set((CheckItem*)Panel_get(super, settings->colorScheme), true);
+   CheckItem_set((CheckItem*)Panel_get(super, (int)CRT_colorScheme), true);
```
fasterit, line 107: Hack, baaaad hack

CRT.c
```diff
@@ -1192,8 +1192,6 @@ void CRT_init(const Settings* settings, bool allowUnicode, bool retainScreenOnEx
    redirectStderr();
    noecho();
    CRT_settings = settings;
-   CRT_colors = CRT_colorSchemes[settings->colorScheme];
```
fasterit, line 1195: Line 1289:
int resetColor = CRT_colors ? CRT_colors[RESET_COLOR] : CRT_colorSchemes[COLORSCHEME_DEFAULT][RESET_COLOR];

Explorer09: > Line 1289: int resetColor = CRT_colors ? CRT_colors[RESET_COLOR] : CRT_colorSchemes[COLORSCHEME_DEFAULT][RESET_COLOR];


Um. What's the problem with that line? That line is in `CRT_done`.

ColorsPanel.c
Explorer09: I actually prefer this way so that the `settings` object can reflect just the "settings that we would write to htoprc", not always "settings that are currently active".

The `Settings` object design was a missed opportunity: At first sight it look like multiple instances of `Settings` are allowed but the rest of htop codebase had the one `Settings` object only assumption and would cause bugs if the assumption was broken.

CRT.c
fasterit: yes, and it tests CRT_colors which you don't initialize any more

Explorer09: @fasterit `CRT_colors` was initialized in `CRT_setColors`, which was still called inside `CRT_init` (I didn't leave that call out).

The assignment of `CRT_colors` here was technically redundant. And I would like the assignment of that global var to be consolidated to `CRT_setColors` for code consistency.

fasterit: ok, that sounds reasonable

BenBE:
Merged as part of #1876 …

PR #1886, opened 2026-02-02
Resolve remaining clang-21 warnings

natoscott:


linux/SystemdMeter.c
```diff
@@ -275,13 +276,21 @@ static void updateViaExec(bool user) {
          }
          free_and_xStrdup(&ctx->systemState, lineBuffer + strlen("SystemState="));
       } else if (String_startsWith(lineBuffer, "NFailedUnits=")) {
-         ctx->nFailedUnits = strtoul(lineBuffer + strlen("NFailedUnits="), NULL, 10);
+         unsigned long value = strtoul(lineBuffer + strlen("NFailedUnits="), &endptr, 10);
+         if (value <= UINT_MAX && *endptr != '\0')
```
Explorer09: What does this `endptr` do in this check?

linux/LinuxProcessTable.c
```diff
@@ -77,7 +77,7 @@ static FILE* fopenat(openat_arg_t openatArg, const char* pathname, const char* m
    return fp;
 }
 
-static inline uint64_t fast_strtoull_dec(char** str, int maxlen) {
+static inline uint64_t fast_strtoull_dec(char** str, ssize_t maxlen) {
```
Explorer09: The `maxlen` argument can be an unsigned `size_t`.

linux/SystemdMeter.c
BenBE: Should be `*endptr == '\0'` to check for "converted to end of string"

linux/SystemdMeter.c
```diff
@@ -267,6 +267,7 @@ static void updateViaExec(bool user) {
    }
 
    char lineBuffer[128];
+   char* endptr;
```
BenBE: Scope can be reduced to inside this loop.

linux/LinuxProcessTable.c
```diff
@@ -738,8 +747,8 @@ static void LinuxProcessTable_readMaps(LinuxProcess* process, openat_arg_t procF
       uint64_t map_start;
       uint64_t map_end;
       bool map_execute;
-      unsigned int map_devmaj;
-      unsigned int map_devmin;
+      uint64_t map_devmaj;
+      uint64_t map_devmin;
```
BenBE, line 751: Technically the kernel returns two `unsigned int` here, which are combined into a `dev_t` with the `makedev` macro. So while not wrong, this would force proper parsing/range checking if we processed these further than just simply checking for zero.

linux/LinuxProcessTable.c
BenBE:
```suggestion
static inline uint64_t fast_strtoull_dec(char** str, size_t maxlen) {
```

As noted above.

linux/LinuxProcessTable.c
```diff
@@ -92,7 +92,7 @@ static inline uint64_t fast_strtoull_dec(char** str, int maxlen) {
    return result;
 }
 
-static long long fast_strtoll_dec(char** str, int maxlen) {
+static long long fast_strtoll_dec(char** str, ssize_t maxlen) {
```
BenBE:
```suggestion
static long long fast_strtoll_dec(char** str, size_t maxlen) {
```

linux/LinuxProcessTable.c
```diff
@@ -107,20 +107,29 @@ static long long fast_strtoll_dec(char** str, int maxlen) {
    return neg ? -result : result;
 }
 
-static long fast_strtol_dec(char** str, int maxlen) {
+static int fast_strtoi_dec(char** str, ssize_t maxlen) {
```
BenBE:
```suggestion
static int fast_strtoi_dec(char** str, size_t maxlen) {
```

linux/LinuxProcessTable.c
```diff
@@ -107,20 +107,29 @@ static long long fast_strtoll_dec(char** str, int maxlen) {
...
+   if (!maxlen)
+      maxlen = 10; // length of maximum value of 2147483647
+   long long result = fast_strtoll_dec(str, maxlen);
+   assert(result <= INT_MAX);
+   assert(result >= INT_MIN);
+   return (int)result;
+}
+
+static long fast_strtol_dec(char** str, ssize_t maxlen) {
```
BenBE:
```suggestion
static long fast_strtol_dec(char** str, size_t maxlen) {
```

linux/LinuxProcessTable.c
```diff
@@ -107,20 +107,29 @@ static long long fast_strtoll_dec(char** str, int maxlen) {
...
    long long result = fast_strtoll_dec(str, maxlen);
    assert(result <= LONG_MAX);
    assert(result >= LONG_MIN);
    return (long)result;
 }
 
-static unsigned long fast_strtoul_dec(char** str, int maxlen) {
+static unsigned long fast_strtoul_dec(char** str, ssize_t maxlen) {
```
BenBE:
```suggestion
static unsigned long fast_strtoul_dec(char** str, size_t maxlen) {
```

linux/LinuxProcessTable.c
```diff
@@ -107,20 +107,29 @@ static long long fast_strtoll_dec(char** str, int maxlen) {
...
    unsigned long long result = fast_strtoull_dec(str, maxlen);
    assert(result <= ULONG_MAX);
    return (unsigned long)result;
 }
 
-static inline uint64_t fast_strtoull_hex(char** str, int maxlen) {
+static inline uint64_t fast_strtoull_hex(char** str, ssize_t maxlen) {
```
BenBE:
```suggestion
static inline uint64_t fast_strtoull_hex(char** str, size_t maxlen) {
```

linux/LinuxProcessTable.c
natoscott: Yep, 💯, thanks @BenBE - will get the other items all cleaned up shortly.

linux/LinuxProcessTable.c
natoscott: @BenBE hmm, not sure about this last one - I think this breaks the logic within the first 'if' statement...?  (TBH the current code for this initial maxlen zero handling looks a bit questionable).

Its all theoretical though because maxlen is never zero in any of the current fast_strtoull_hex calls.

I'll stick with the size_t arg change and update fast_strtoull_hex to assert that a positive maxlen is given instead.

Issue #1844, opened 2025-12-25
Security check of external programs before launching


Explorer09:
This is a feature ticket.

There are several feature requests in htop that would involve in launching an external program or script (#506, #526). The `strace` feature implemented in htop also involves in launching a program.

For now, there are volunteer pull requests (#1571, #1843, #1793) that try to implement some the features above, however, a major problem in these implementations are that they simply launch external programs without check, which is a bad idea regarding security.

htop is often run with root privileges, and if user is not careful, they can execute a malicious program accidentally, or run a program with elevated privileges that is not technically necessary.

Even for `strace` it can present a similar problem: The program might be a malicious program with the same name installed by a malicious local user, and `root` user isn't aware of that ahead of time.

This feature ticket is to encourage new contributers to implement internal APIs that allow secure checking and launching programs in htop.

These APIs:
(1) must have an option to decide whether the external program should have elevated privileges, or drop privileges whenever possible,
(2) if the program is to be launched with elevated privileges, must ensure the program is trusted by `root` (check if the program is owned by `root` and has user execute permission),

For security reasons, these system call / APIs should be avoided: `execl`, `execv` and similar, `popen(3)`, `system(3)`.

Instead try one of these: [`fexecve(3)`](https://man7.org/linux/man-pages/man3/fexecve.3.html), [`execveat(2)`](https://man7.org/linux/man-pages/man2/execveat.2.html) These APIs allow launching an external program by the opened file descriptor rather than the path. Programs launching through FD can avoid the TOCTOU vulnerability that paths cannot .

rubyFeedback:
Perhaps I do not completely understand the issue raised.

I am not against (2) nor am I against using fexecve or execveat, but isn't the whole point of being the superuser to be able to do whatever he or she wants? They can always use a regular user account if they want to and run "htop" there. 

> htop is often run with root privileges, and if user is not careful, they can execute a malicious program accidentally

Right - but it should not up to the htop developers to dictate what others do with the program at hand. It was their decision to run as superuser. Why should that be the htop developers concern?

Explorer09:
> 
> I am not against (2) nor am I against using fexecve or execveat, but isn't the whole point of being the superuser to be able to do whatever he or she wants? They can always use a regular user account if they want to and run "htop" there.
> 
> > htop is often run with root privileges, and if user is not careful, they can execute a malicious program accidentally
> 
[...]

PR #1887, opened 2026-02-03
Added support OpenRC init system and metrics

GermanAizek:
@natoscott,

htop detect SystemD or OpenRC installed

Main repo: https://github.com/OpenRC/openrc

More info about OpenRC: https://en.wikipedia.org/wiki/OpenRC

fix `strtoll` 32-bit overflow change `int` to `long long`

References:
- https://softwareengineering.stackexchange.com/questions/251508/is-there-a-stricter-strtoull-in-any-ubiquitous-c-library

Final result:

https://github.com/user-attachments/assets/36a86fdd-9b18-49f4-b2cb-ecf21e832066

### DIFF
--- a/linux/Platform.c
+++ b/linux/Platform.c
@@ -21,6 +21,7 @@ in the source distribution for its full text.
 #include <string.h>
 #include <time.h>
 #include <unistd.h>
+#include <sys/stat.h>
 #include <sys/sysmacros.h>
 
 #include "BatteryMeter.h"
@@ -228,58 +229,14 @@ void Platform_setBindings(Htop_Action* keys) {
    keys[KEY_F(20)] = Platform_actionHigherAutogroupPriority; // Shift-F8
 }
 
-const MeterClass* const Platform_meterTypes[] = {
-   &CPUMeter_class,
-   &ClockMeter_class,
-   &DateMeter_class,
-   &DateTimeMeter_class,
-   &LoadAverageMeter_class,
-   &LoadMeter_class,
-   &MemoryMeter_class,
-   &SwapMeter_class,
-   &MemorySwapMeter_class,
-   &SysArchMeter_class,
-   &HugePageMeter_class,
-   &TasksMeter_class,
-   &UptimeMeter_class,
-   &SecondsUptimeMeter_class,
-   &BatteryMeter_class,
-   &HostnameMeter_class,
-   &AllCPUsMeter_class,
-   &AllCPUs2Meter_class,
-   &AllCPUs4Meter_class,
-   &AllCPUs8Meter_class,
-   &LeftCPUsMeter_class,
-   &RightCPUsMeter_class,
-   &LeftCPUs2Meter_class,
-   &RightCPUs2Meter_class,
-   &LeftCPUs4Meter_class,
-   &RightCPUs4Meter_class,
-   &LeftCPUs8Meter_class,
-   &RightCPUs8Meter_class,
-   &BlankMeter_class,
-   &PressureStallCPUSomeMeter_class,
-   &PressureStallIOSomeMeter_class,
-   &PressureStallIOFullMeter_class,
-   &PressureStallIRQFullMeter_class,
-   &PressureStallMemorySomeMeter_class,
-   &PressureStallMemoryFullMeter_class,
-   &ZfsArcMeter_class,
-   &ZfsCompressedArcMeter_class,
-   &ZramMeter_class,
-   &DiskIORateMeter_class,
-   &DiskIOTimeMeter_class,
-   &DiskIOMeter_class,
-   &NetworkIOMeter_class,
-   &SELinuxMeter_class,
-   &SystemdMeter_class,
-   &SystemdUserMeter_class,
-   &OpenRCMeter_class,
-   &OpenRCUserMeter_class,
-   &FileDescriptorMeter_class,
-   &GPUMeter_class,
-   NULL
-};
+const MeterClass* Platform_meterTypes[64];
+static int Platform_meterTypes_size = 0;
+
+static void Platform_meterTypes_add(const MeterClass* mc) {
+   if (Platform_meterTypes_size < 63) {
+      Platform_meterTypes[Platform_meterTypes_size++] = mc;
+   }
+}
 
 int Platform_getUptime(void) {
    char uptimedata[64] = {0};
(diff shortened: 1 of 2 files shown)


linux/Platform.c
```diff
@@ -555,9 +515,10 @@ FileLocks_ProcessData* Platform_getProcessLocks(pid_t pid) {
 
       errno = 0;
       char* end = de->d_name;
-      int file = strtoull(de->d_name, &end, 10);
-      if (errno || *end)
+      long long file_ll = strtoll(de->d_name, &end, 10);
+      if (errno || *end || file_ll < 0 || file_ll > INT_MAX)
          continue;
+      int file = (int)file_ll;
```
BenBE: Please split this change into its own PR.

linux/Platform.c
```diff
@@ -207,56 +209,14 @@ void Platform_setBindings(Htop_Action* keys) {
...
-   NULL
-};
+const MeterClass* Platform_meterTypes[64];
+static int Platform_meterTypes_size = 0;
+
+static void Platform_meterTypes_add(const MeterClass* mc) {
+   if (Platform_meterTypes_size < 63) {
+      Platform_meterTypes[Platform_meterTypes_size++] = mc;
+   }
+}
```
BenBE, line 239: This makes things unnecessarily complicated by introducing dynamic lists where it's not really necessary.

If we are to go this route it's best to also have this change to make the meter list dynamic a) global for all platforms and b) a separate commit to split changes cleanly.

BenBE (changes requested):
Given that OpenRC is used on both Linux and some *BSDs it's probably best to write the meter as platform-agnostic as possible (possibly something for the `generic/` folder?) and integrate it with the various *BSD variants that may use OpenRC too (at least FreeBSD has experimental support AFAIK).

linux/Platform.c
natoscott: This issue is resolved via #1886

linux/Platform.c
natoscott: +1 - FWIW, I do think this is a neat idea and worth tackling, being able to cull Meters at runtime for optional system features (selinux, GPUs, ZFS, etc) makes for a more pleasant experience for users without those things.

BenBE:
I rebased the commit onto latest `main` to include the existing bugfix for the overflow issue.

Furthermore I split the patch into two separate commits to separate the individual concerns (OpenRC vs. dynamic meters list).

As noted, I'd suggest to merge the first commit with this PR and have some more planning for the second part.

Apart from this: Does OpenRC provide some library to query the necessary information without the need for starting sub-processes (similar to how the systemd meter is implemented)?

linux/OpenRCMeter.c
```diff
@@ -0,0 +1,204 @@
...
+   ctx->runlevel = NULL;
+   ctx->services_stopped = INVALID_VALUE;
+   ctx->services_started = INVALID_VALUE;
+
+   updateViaExec(user);
+
+   xSnprintf(this->txtBuffer, sizeof(this->txtBuffer), "%s", ctx->runlevel ? ctx->runlevel : "???");
+}
+
+static void _OpenRCMeter_display(ATTR_UNUSED const Object* cast, RichString* out, OpenRCMeterContext_t* ctx) {
```
natoscott: BTW, this pattern (leading underscore for 'internal' functions) is not used in htop - I recommend just dropping that underscore off here.

natoscott: Heh, I just noticed the one place in the code base its used is in the systemd meter :P  ... which would have been the starting point for this new code.  Its not a good pattern to continue, for consistency anyway.

BenBE: ACK, but collides with an existing function of that name. Will take a look.

linux/OpenRCMeter.c
```diff
@@ -0,0 +1,204 @@
...
+      close(fdpair[1]);
+      int fdnull = open("/dev/null", O_WRONLY);
+      if (fdnull < 0)
+         exit(1);
+      dup2(fdnull, STDERR_FILENO);
+      close(fdnull);
+      if (user) {
+         execlp("rc-status", "rc-status", "--user", "-a", (char*)NULL);
+      } else {
+         execlp("rc-status", "rc-status", "-a", (char*)NULL);
```
Explorer09, line 74: An `exec` without a security check could become an issue in the future (see #1844).

In particular:

(1) It looks for `rc-status` through `PATH`. The environment variable is susceptible to attack (perhaps allow the program name to be overridden at compile time can solve it).

(2) We aren't sure the `rc-status` program is trustworthy. My suggestion to this trustworthiness checking is the ensure the program is owned by `root` before running.

linux/Platform.c
```diff
@@ -1118,6 +1075,65 @@ bool Platform_init(void) {
...
+   Platform_meterTypes_add(&PressureStallMemorySomeMeter_class);
+   Platform_meterTypes_add(&PressureStallMemoryFullMeter_class);
+   Platform_meterTypes_add(&ZfsArcMeter_class);
+   Platform_meterTypes_add(&ZfsCompressedArcMeter_class);
+   Platform_meterTypes_add(&ZramMeter_class);
+   Platform_meterTypes_add(&DiskIORateMeter_class);
+   Platform_meterTypes_add(&DiskIOTimeMeter_class);
+   Platform_meterTypes_add(&DiskIOMeter_class);
+   Platform_meterTypes_add(&NetworkIOMeter_class);
+   Platform_meterTypes_add(&SELinuxMeter_class);
```
Explorer09, line 1141: Why do we need this? A bunch of `Platform_meterTypes_add` calls that won't do anything more than increasing the code size. What was the problem with using the static array before?

linux/Platform.c
```diff
@@ -1118,6 +1075,65 @@ bool Platform_init(void) {
...
+
+   struct stat st;
+   if (stat("/bin/systemctl", &st) == 0) {
+      Platform_meterTypes_add(&SystemdMeter_class);
```
Explorer09, line 1145: I don't like these `stat` calls. Hiding available meters at runtime to end user is a bad idea from the start. Either make all of them available, or allow builders to disable them at compile time. (Something like `--without-openrc` and `--without-systemd` in `configure`.)

linux/Platform.c
natoscott: The answer to this should be clear from code inspection, and was also discussed in earlier review notes.
Please double check - feel free to ask again if not obvious though.

linux/Platform.c
natoscott: Why is it a bad idea?  Its good from a UI perspective IMO - why present options to the user that cannot work?

natoscott: Oh, its not obvious - we do not know at compile time which system will be selected by the user, so it needs to be done at runtime.

Explorer09: > Oh, its not obvious - we do not know at compile time which system will be selected by the user, so it needs to be done at runtime.

@natoscott The meter can show `N/A` data for init systems that the OS isn't using. And if an OS distro doesn't use, for example, OpenRC at all, they should be able to compile it out so as to not waste code.

Distros that allow multiple init systems can have both systemd and OpenRC meters visible. And the UI confusion remains. You need more code to check which init system is active - just `stat` the existence of the contol program is not enough.

linux/Platform.c
Explorer09: You mean [this comment](https://github.com/htop-dev/htop/pull/1887#discussion_r2760565583), no?

I'm so sorry, but my concern is the wasted code size. My argument is not about whether there's a need of a dynamic meter list. Most of the meters in this list are static - so they should be kept as static array to avoid useless `add()` calls. Some other meters can have a boolean "visible" flag each. The real dynamic meters are then put into the end.

linux/Platform.c
natoscott: | The meter can show N/A data for init systems that the OS isn't using.

That's poor UX - better to not present useless options, we can do better than that.

| [...] the existence of the contol program is not enough.

Sure, it seems good enough for a first pass though - if both binaries are installed, its fine to present both options IMO.

linux/Platform.c
natoscott: | I'm so sorry, but my concern is the wasted code size.

It's negligible, and this is a pattern we should do more of (runtime removal of non-useful elements).  I'm sure there's substantially more code size increases from custom sort code which could be removed too.  :)

linux/Platform.c
Explorer09: > | The meter can show N/A data for init systems that the OS isn't using.
> 
> That's poor UX - better to not present useless options, we can do better than that.

I need to think of this a bit. My concern is when a meter that is available in htop suddenly becomes unavailable in one invocation, or, a meter that was hidden before suddenly pops up in another htop invocation. I don't think that a better UX if a meter can disappear or re-appear without a good reason.

And there could be another scenario: A user may have both `SystemdMeter` and `OpenRCMeter` enabled in their config file, but because only one init system can be active at time, one of the meters would definitely have no data shown.

linux/Platform.c
Explorer09: > I'm sure there's substantially more code size increases from custom sort code which could be removed too. :)

I consider your argument like this comes in bad faith. Not funny. If it weren't for the stable sort requirement I would not have proposed the custom sort code... (please keep the discussion of that in #1785)

natoscott: | Not funny.

You're right, not funny.

Noone gets to have one set of rules for their own changes while applying different criteria for everyone else.

linux/Platform.c
natoscott: | My concern is when a meter that is available in htop suddenly becomes unavailable

This issue exists already with conditionally compiled code, or when new meters arrives that the previous version didn't support, etc.  The code is written to deal with it as best it can.

|  I don't think that a better UX if a meter can disappear or re-appear without a good reason.

Yes of course - if the system doesn't support some functionality, that's a very good reason.  Do not present it in the UI when its use is guaranteed to fail.

Explorer09: > | My concern is when a meter that is available in htop suddenly becomes unavailable
> 
> This issue exists already with conditionally compiled code

No. For the conditionally compiled code, the builder/packager who builds htop is responsible for what they are doing. If the user sees something not built in, they would blame the distro, not us.

> | I don't think that a better UX if a meter can disappear or re-appear without a good reason.
> 
> Yes of course - if the system doesn't support some functionality, that's a very good reason. Do not present it in the UI when its use is guaranteed to fail.

But htop would still accept the meter when it's written in the config file. That would feel like an "easter egg". We don't have a "show all meters even when the meter won't work in your system" option.

Explorer09:
@GermanAizek Is this PR still live?

I noticed a part of the commits in this PR has been merged to `main`, but then I suddenly discovered that: There's no `--user` option in rc-status(8) utility in OpenRC.

I didn't test running myself so I can't say whether this is a real issue, but I've read the man page and the source code of `rc-status` and there's no `--user` option there.

BenBE:
I merged the part of the PR that I indicated was in scope of the actual "OpenRC support" for htop. Thus the major refactoring in the third commit is intentionally left out. That said, if this PR is rebased onto main only the third commit would remain, which should be a separate PR.

Regarding the `--user` option you mentioned: Please make a quick bugfix PR for that, as this PR is "done".

PR #1889, opened 2026-02-05
Fix regression on inappropriately truncated command name

edgar-bonet:
As mentioned in #1817, in `LinuxProcessTable_readCmdlineFile()`, `argSepNUL` is being set if **any** character is found after the first NUL, which is incorrect.

This pull request changes the behavior to only set the flag when a non-NUL character is found after the first NUL.

Tested on Debian trixie. Both problematic processes mentioned in the issue are now displayed correctly.

Fixes: #1817

### DIFF
--- a/linux/LinuxProcessTable.c
+++ b/linux/LinuxProcessTable.c
@@ -1336,11 +1336,6 @@ static bool LinuxProcessTable_readCmdlineFile(Process* process, openat_arg_t pro
    bool argSepSpace = false;
 
    for (size_t i = 0; i < (size_t)amtRead; i++) {
-      // If this is true, there's a NUL byte in the middle of command
-      if (tokenEnd != (size_t)-1) {
-         argSepNUL = true;
-      }
-
       const char argChar = command[i];
 
       /* newline used as delimiter - when forming the mergedCommand, newline is
@@ -1360,6 +1355,11 @@ static bool LinuxProcessTable_readCmdlineFile(Process* process, openat_arg_t pro
          }
 
          continue;
+      }
+
+      // If this is true, there's a NUL byte in the middle of command
+      if (tokenEnd != (size_t)-1) {
+         argSepNUL = true;
       }
 
       /* Record some information for the argument parsing heuristic below. */


edgar-bonet:
Gentle ping.

Issue #1817 is still current. Is there anything I can do to help this pull request progress? Maybe rebase on main and re-test?

BenBE:
Sorry, had little time recently and reviewing this one takes a bit of time as there are several edge cases we need to take care of.

Rebasing onto main should be trivial enough with this one so not strictly necessary.. That's more important with more complicated patch sets that affect a broader part of the code.

With this PR it's mostly looking at things like ssh, processes that have square brackets in their cmdline, … Or things like postgres (`main: walwriter`), php-fpm, dbus-daemon. Also look at the kernel thread list (<kbd>SHIFT</kbd>+<kbd>K</kbd>) if there's any effects.

edgar-bonet:
@BenBE: Thanks for your feedback!

I took a look on my Debian trixie server, comparing the output of htop from main (3066c681 "Merge pull request #1918 from ViryBe/main") and from this pull request. I enabled the display of kernel threads, as you suggested. Below are the only differences I could find:

* as shown by htop(main):
  ```text
     PID USER       PRI  NI  VIRT   RES   SHR S  CPU% MEM%▽  TIME+  Command
     798 root        20   0  294M 36812 27552 S   0.0  1.8  8:17.36 php-fpm.conf)
  551691 debian      20   0 20720  7984  5480 S   0.7  0.4  0:01.94 0
  551377 root        20   0     0     0     0 I   0.0  0.0  0:00.11 kworker/u8:3-flush-254:0
  554353 root        20   0     0     0     0 I   0.0  0.0  0:00.00 kworker/u8:2-writeback
  ```
* as shown by this PR's version of htop:
  ```text
     PID USER       PRI  NI  VIRT   RES   SHR S  CPU% MEM%▽  TIME+  Command
     798 root        20   0  294M 36812 27552 S   0.0  1.8  8:17.36 php-fpm: master process (/etc/php/8.4/fpm/php-fpm.conf)
  551691 debian      20   0 20720  7984  5480 S   0.7  0.4  0:01.91 sshd-session: debian@pts/0
  551377 root        20   0     0     0     0 I   0.0  0.0  0:00.11 kworker/u8:3-events_unbound
  554353 root        20   0     0     0     0 I   0.0  0.0  0:00.00 kworker/u8:2-events_unbound
  ```

The fist two processes (PHP and SSH) are as I expected: they were my original motivation for this. The two kernel threads that follow are a surprise to me: I do not understand why they are displayed differently. FWIW, this pull-request's version seems to get the names right:

```shell
$ sudo file /proc/551377/exe
/proc/551377/exe: unreadable symlink `/proc/551377/exe' (No such file or directory)
$ sudo cat /proc/551377/cmdline | hd
$ sudo cat /proc/551377/comm | hd
00000000  6b 77 6f 72 6b 65 72 2f  75 38 3a 33 2d 65 76 65  |kworker/u8:3-eve|
00000010  6e 74 73 5f 75 6e 62 6f  75 6e 64 0a              |nts_unbound.|
0000001c
```

The other kernel thread (554353) shows exactly the same:
* `/proc/$pid/exe` is unreadable, even by root
* `/proc/$pid/cmdline` is empty
* `/proc/$pid/comm` holds `"kworker/u8:3-events_unbound\n"`

> processes that have square brackets in their cmdline

I have two of them, and both of them seem to get correctly displayed, with or without this PR:
* `sshd-session: debian [priv]`
* `sshd: /usr/sbin/sshd -D [listener] 0 of 10-100 startups`

> Or things like postgres

I do not have Postgres here.

> dbus-daemon

I have a couple of these. Again, they both seem fine, with or without this PR:
* `dbus-daemon --system --address=systemd: --nofork --nopidfile --systemd-activation --syslog-only`
* `dbus-daemon --session --address=systemd: --nofork --nopidfile --systemd-activation --syslog-only`

PR #1891, opened 2026-02-09
cmdline: use strtoul() for safety check on overflow username arg

GermanAizek:
References:
- https://stackoverflow.com/a/3792686

### DIFF
--- a/CommandLine.c
+++ b/CommandLine.c
@@ -225,7 +225,7 @@ static CommandLineStatus parseArguments(int argc, char** argv, CommandLineSettin
             if (!username) {
                flags->userId = geteuid();
             } else if (!Action_setUserOnly(username, &(flags->userId))) {
-               char* endptr;
+               const char* endptr;
                errno = 0;
                unsigned long val = strtoul(username, &endptr, 10);
                if (errno == ERANGE || *endptr != '\0' || username == endptr || val > UINT_MAX) {


CommandLine.c
```diff
@@ -223,12 +224,14 @@ static CommandLineStatus parseArguments(int argc, char** argv, CommandLineSettin
...
-               for (const char* itr = username; *itr; ++itr)
-                  if (!isdigit((unsigned char)*itr)) {
-                     fprintf(stderr, "Error: invalid user \"%s\".\n", username);
-                     return STATUS_ERROR_EXIT;
-                  }
-               flags->userId = (uid_t)atol(username);
+               char* endptr;
+               errno = 0;
+               unsigned long val = strtoul(username, &endptr, 10);
+               if (errno == ERANGE || *endptr != '\0' || username == endptr) {
```
natoscott: uid_t is an unsigned 32 bit integer (at least on Linux), so we should also check for val >= UINT_MAX here (this may make the ERANGE check redundant too - we haven't used that errno check in any other location in htop currently, anyway).

CommandLine.c
```diff
@@ -223,12 +224,14 @@ static CommandLineStatus parseArguments(int argc, char** argv, CommandLineSettin
             if (!username) {
                flags->userId = geteuid();
             } else if (!Action_setUserOnly(username, &(flags->userId))) {
-               for (const char* itr = username; *itr; ++itr)
-                  if (!isdigit((unsigned char)*itr)) {
-                     fprintf(stderr, "Error: invalid user \"%s\".\n", username);
-                     return STATUS_ERROR_EXIT;
-                  }
-               flags->userId = (uid_t)atol(username);
+               char* endptr;
```
Explorer09:
```suggestion
               const char* endptr;
```

CommandLine.c
Explorer09: If we want an overflow check, perhaps the right constant to check for is `UID_T_MAX`.

```c
#ifndef UID_T_MAX
#define UID_T_MAX ((((uid_t)1<<(sizeof(uid_t)*CHAR_BIT-1-((uid_t)-1<0)))-1)*2+1)
#endif
```

GermanAizek: > uid_t is an unsigned 32 bit integer (at least on Linux), so we should also check for val >= UINT_MAX here (this may make the ERANGE check redundant too - we haven't used that errno check in any other location in htop currently, anyway).

I have saved errno == ERANGE check to ensure correct operation on 32-bit platforms.

GermanAizek:
@Explorer09 your commit Actions CI is failed.

Explorer09:
> @ Explorer09 your commit Actions CI is failed.

@GermanAizek 
Add an explicit cast to fix it:

```c
unsigned long val = strtoul(username, (const char**)&endptr, 10);
```

Actually this was a design defect of the `strtoul` API. C didn't support function overloading, and a function that can return a pointer whose address derived from an input string would suffer from a similar problem (`strchr` and like).

```c
/* standard */
unsigned long strtoul(const char *restrict nptr, char **restrict endptr, int base);
/* more ideal API */
unsigned long strtoul(const char *restrict nptr, const char **restrict endptr, int base);
unsigned long strtoul(char *restrict nptr, char **restrict endptr, int base);
```

CommandLine.c
natoscott: @GermanAizek this suggested change is incorrect and inconsistent with every other strtoxxx call site in htop, please ignore this and keep it non-const as you had it.

CommandLine.c
natoscott: I think we can safely remove it - its inconsistent with other htop code and unnecessary.  Check the return value description on the strtoul man page which has "strtoul()  returns ULONG_MAX  and  sets errno to ERANGE." ... so my suggested range change from before covers this case already, and ERANGE checking becomes redundant as a result.

Just the UINT_MAX check will suffice, keep it simple.

Explorer09: @natoscott On 32-bit systems ULONG_MAX == UINT_MAX so it is _not_ technically redundant.

CommandLine.c
Explorer09: @natoscott What "incorrect" are you technically saying?

While the consistency of other `strto*` calls in htop codebase could be a reason against my suggested change, I don't think it's "incorrect" in any way. It's a language defect (as I've noted in a [comment below](https://github.com/htop-dev/htop/pull/1891#issuecomment-3872246354)). The pointed address should have been `const`-qualified if we are coding for memory safety.

natoscott:
Simplified and merged, thanks for the contribution @GermanAizek

natoscott:
> @natoscott
> I forgot to mention one thing.
> strtoul allows negative numbers and would convert negative numbers to a positive number by (2^UINT_WIDTH + x) (while INT_MIN < x <= -1).
>
> So we probably need this check before calling strtoul : (username[0] >= '0' && username[0] <= '9').

@Explorer09 not needed - think it through, the situation you describe cannot happen...

```
$ ./htop -u -1
./htop: invalid option -- '1'
```

Explorer09:
> > @natoscott
> > I forgot to mention one thing.
> > strtoul allows negative numbers and would convert negative numbers to a positive number by (2^UINT_WIDTH + x) (while INT_MIN < x <= -1).
> > So we probably need this check before calling strtoul : (username[0] >= '0' && username[0] <= '9').
> 
> @Explorer09 not needed - think it through, the situation you describe cannot happen...
> 
> ```
> $ ./htop -u -1
> ./htop: invalid option -- '1'
> ```

What about `./htop -u-1` ? Or, `./htop -u ' -1'`?
I thought the interpretation of `-1` as an option was an accident in parsing.

natoscott:
Please try these things yourself before posting.  We're all busy people, the signal-to-noise ratio is terrible currently.

```
$ ./htop -u-1
Error: invalid user "-1".
$ ./htop -u ' -1'
Error: invalid user " -1".
```

PR #1892, opened 2026-02-09
panel: reuse single rich str instead two alloc rich str

GermanAizek:
Reallocation rich string inside body for loop may be impractical with freeing inside, same applies to new and old rich strings, it would be more logical to use one object and reuse it to save memory

### DIFF
--- a/Panel.c
+++ b/Panel.c
@@ -277,11 +277,13 @@ void Panel_draw(Panel* this, bool force_redraw, bool focus, bool highlightSelect
       ? CRT_colors[this->selectionColorId]
       : CRT_colors[PANEL_SELECTION_UNFOCUS];
 
+   RichString_begin(item);
    if (this->needsRedraw || force_redraw) {
       int line = 0;
       for (int i = first; line < h && i < upTo; i++) {
          const Object* itemObj = Vector_get(this->items, i);
-         RichString_begin(item);
+         RichString_rewind(&item, RichString_size(&item));
+         item.highlightAttr = 0;
          Object_display(itemObj, &item);
          int itemLen = RichString_sizeVal(item);
          int amt = MINIMUM(itemLen - scrollH, this->w);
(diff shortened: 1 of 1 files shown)


Panel.c
```diff
@@ -219,6 +219,12 @@ void Panel_splice(Panel* this, Vector* from) {
    this->needsRedraw = true;
 }
 
+static void RichString_reset(RichString* this) {
+   this->chlen = 0;
```
Explorer09: This would cause memory leak. `this->chptr` is not freed when you set the length of RichString to 0.

EDIT: You should use `RichString_setLen` for what you are doing.

GermanAizek: Many thanks. Good catch.

GermanAizek: @Explorer09, fix it here https://github.com/htop-dev/htop/pull/1892/commits/1dad08301e5e12eab5e42684f4246755797a9e50

Review again please.

GermanAizek: @Explorer09, I can't use function `RichString_setLen` because it's static.
I use hack `RichString_rewind(&item, RichString_size(&item));` that equivalent `RichString_setLen(rs, 0)`

Panel.c
```diff
@@ -278,10 +278,12 @@ void Panel_draw(Panel* this, bool force_redraw, bool focus, bool highlightSelect
       : CRT_colors[PANEL_SELECTION_UNFOCUS];
 
    if (this->needsRedraw || force_redraw) {
+      RichString_begin(item);
```
Explorer09: If you are gonna reuse the buffer, why not use it throughout the whole function? Don't initialize the string inside an if-block.

fasterit: @GermanAizek : Will you continue to improve the PR? ^ this has been requested and confirmed by @BenBE on IRC

GermanAizek: > @GermanAizek : Will you continue to improve the PR? ^ this has been requested and confirmed by @BenBE on IRC

It's been so long, I've already forgotten what I was doing, but I think I did as I was asked @Explorer09

PR #1893, opened 2026-02-10
TraceScreen: fix excess copying parent memory page table

GermanAizek:
posix_spawn() is a more lightweight and modern alternative to fork()/exec(), which are used inside popen(). The main advantage of posix_spawn is that it can create a new process without completely duplicating the address space of the parent process.

References:
- https://blog.famzah.net/2018/12/19/posix_spawn-performance-benchmarks-and-usage-examples/
- https://lobste.rs/s/smbsd5/fork_road
- https://www.reddit.com/r/C_Programming/comments/1lvdhp2/fork_vs_posix_spawn/

### DIFF
--- a/OpenFilesScreen.c
+++ b/OpenFilesScreen.c
@@ -12,6 +12,7 @@ in the source distribution for its full text.
 #include <errno.h>
 #include <fcntl.h>
 #include <inttypes.h>
+#include <spawn.h>
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -26,6 +27,9 @@ in the source distribution for its full text.
 #include "Vector.h"
 #include "XUtils.h"
 
+
+// Helper to create a mutable string for argv arrays
+#define MUTABLE_STR(s) (char[]){s}
 
 // cf. getIndexForType; must be larger than the maximum value returned.
 #define LSOF_DATACOL_COUNT 8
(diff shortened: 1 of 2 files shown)


OpenFilesScreen.c
```diff
@@ -106,32 +110,41 @@ static OpenFiles_ProcessData* OpenFilesScreen_getProcessData(pid_t pid) {
       return pdata;
    }
 
-   pid_t child = fork();
-   if (child == -1) {
+   pid_t child;
+   posix_spawnattr_t attr;
+   posix_spawnattr_init(&attr);
```
BenBE, line 115: This can fail, thus return values should be checked.

Also, there are no attributes set on this empty object, thus for `posix_spawnp` the attributes could be passed as `NULL`.

OpenFilesScreen.c
```diff
@@ -106,32 +110,41 @@ static OpenFiles_ProcessData* OpenFilesScreen_getProcessData(pid_t pid) {
...
+   posix_spawn_file_actions_t fa;
+   posix_spawn_file_actions_init(&fa);
```
BenBE, line 117: This can fail, thus return values should be checked.

OpenFilesScreen.c
```diff
@@ -106,32 +110,41 @@ static OpenFiles_ProcessData* OpenFilesScreen_getProcessData(pid_t pid) {
...
+      MUTABLE_STR("lsof"),
+      MUTABLE_STR("-P"),
+      MUTABLE_STR("-o"),
+      MUTABLE_STR("-p"),
+      buffer,
+      MUTABLE_STR("-F"),
+      NULL
+   };
+
+   int spawn_ret = posix_spawnp(&child, "lsof", &fa, &attr, args, NULL);
```
BenBE, line 136: The environment `envp` should not be passed as `NULL`. This is accepted by some implementations, but is not generally portable.

Passing an (mostly¹) empty environment instead should suffice (also makes program output more predictable).

¹override `LC_ALL=C.UTF-8` and pass on `PATH` (optional).

OpenFilesScreen.c
```diff
@@ -106,32 +110,41 @@ static OpenFiles_ProcessData* OpenFilesScreen_getProcessData(pid_t pid) {
...
+
+   posix_spawnattr_destroy(&attr);
+   posix_spawn_file_actions_destroy(&fa);
```
BenBE, line 139: Resource cleanup should be reverse of acquisition.

TraceScreen.c
```diff
@@ -78,45 +82,43 @@ bool TraceScreen_forkTracer(TraceScreen* this) {
...
-      #elif defined(HTOP_LINUX)
-         execlp("strace", "strace", "-T", "-tt", "-s", "512", "-p", buffer, (void*)NULL);
-
-         // Should never reach here, unless execlp fails ...
-         const char* message = "Could not execute 'strace'. Please make sure it is available in your $PATH.";
-         (void)! write(STDERR_FILENO, message, strlen(message));
-      #else // HTOP_DARWIN, HTOP_PCP == HTOP_UNSUPPORTED
-         const char* message = "Tracing unavailable on not supported system.";
-         (void)! write(STDERR_FILENO, message, strlen(message));
-      #endif
```
BenBE, line 112: Behaviour change on platforms not supporting `truss`/`strace` …

TraceScreen.c
```diff
@@ -78,45 +82,43 @@ bool TraceScreen_forkTracer(TraceScreen* this) {
...
+   xSnprintf(buffer, sizeof(buffer), "%d", Process_getPid(this->super.process));
+
+   char* const* args;
+#if defined(HTOP_FREEBSD) || defined(HTOP_OPENBSD) || defined(HTOP_NETBSD) || defined(HTOP_DRAGONFLYBSD) || defined(HTOP_SOLARIS)
+   args = (char* const[]){MUTABLE_STR("truss"), MUTABLE_STR("-s"), MUTABLE_STR("512"), MUTABLE_STR("-p"), buffer, NULL};
+#elif defined(HTOP_LINUX)
+   args = (char* const[]){MUTABLE_STR("strace"), MUTABLE_STR("-T"), MUTABLE_STR("-tt"), MUTABLE_STR("-s"), MUTABLE_STR("512"), MUTABLE_STR("-p"), buffer, NULL};
+#else
+   args = NULL;
+#endif
```
BenBE, line 105: That's become borderline unreadable. Much more noise compared to previously.

Explorer09: @GermanAizek I have a concern about `MUTABLE_STR` macro. It might not do what you want with the syntax, and I feel it strange when it doesn't trigger a "-Wcast-qual" warning from your compiler. (I have a better workaround for this, but it is part of a private email to BenBE that's I'm not sure I can share it publicly for now.)

@BenBE Please check your mailbox. I have sent a private email requesting code review about a security feature when launching exteral programs. It (and issue #1844) would be related to this PR.

OpenFilesScreen.c
Explorer09: Declare `extern char** environ;` at the top of the .c file and pass `environ` as `envp` instead of NULL.

Explorer09:
Overall I'm not sure if this API change is worth it. Modern OS kernels should have implemented the copy-on-write mechanisms for process pages making the performance differences between `fork()` and `posix_spawn()` too small to be noticeable.

@GermanAizek , Do you have benchmark data to back your claim about using `posix_spawn()` in htop? Such as whether it reduces CPU time on average.

The [Linux man page about posix_spawn](https://man7.org/linux/man-pages/man3/posix_spawn.3.html) doesn't say about performance being a primary reason of using `posix_spawn()`, rather it's for "machines that lack the capability to support the fork(2) system call[; t]hese machines are generally small, embedded systems lacking MMU support". Since `fork()` works in htop, that is one less reason for `posix_spawn()`.

Also: I reported issue #1844, about the security of launching external programs. And this PR code is still vulnerable (to, for example, PATH environment variable attack).

EDIT just to mention two more things: (1) For security reasons we need `posix_fspawn()` that operate on file descriptors rather than paths. (2) Please set `POSIX_SPAWN_RESETIDS` to `attr`. It gives us security of dropping setuid privileges with little code.

OpenFilesScreen.c
Explorer09: Let's set `POSIX_SPAWN_RESETIDS` to this `attr` object for security. It drops setuid privileges for us with very little code needed. 🙂

TraceScreen.c
Explorer09: Update: I've made the public version of the code review request as #1915. And I would invite @GermanAizek to take a look of the feature and code changes. It would be more secure than using `posix_spawn()` as proposed here.

PR #1898, opened 2026-02-12
Restore available memory color from previous htop versions

### LINKED ISSUE #1897: memory meter shows *available*, changes colors and visible metrics
Current git `main` started showing the *available* memory in bar meters in red:

Image

This is also the reason I opened #1895 - possibly we should hold on that:

Image

My initial analysis pointed to 1ca6213bac5e8a926c9c27974e3b753d3ab59fb9, but reverting that still shows available memory, though in a different color:

Image

Just noticed that *all* colors changed there for memory...

Image

I have to revert fcc305be4678390bbded99855dc2c9727eea67f7 as well to get expected behavior (where nothing is shown for available memory) back.

So to sum up:

- fcc305be4678390bbded99855dc2c9727eea67f7 changed to show available memory
- 1ca6213bac5e8a926c9c27974e3b753d3ab59fb9 changed colors for memory

natoscott:
Previously METER_TEXT was used as the default memory colour for the "available" class (hack upon hack upon hack).  Restore this colour instead of picking a new (inappropriate) choice in "red" by default for this memory class.

Resolves #1897


eworm-de:
This is far from resolved... It *restores* the color for available memory, but it is still shown. Also other colors are still wrong.

natoscott:
> This is far from resolved

@eworm-de your bug report states"restore available color", so it sounds like that aspect is indeed resolved.

Can you be more explicit regarding the other aspects you didn't mention previously?  Which colours are still wrong?  Thanks!

eworm-de:
Added another comment in https://github.com/htop-dev/htop/issues/1897 to compare the states of memory meters in different git commits directly. I hope that helps and makes it clear...

natoscott:
| I hope that helps and makes it clear...

Thanks, that's perfect - I'll dig deeper soon as I'm able.

eworm-de:
I will continue to give comments in the issue, as the screenshots are there. Is that ok for you?

natoscott:
> I will continue to give comments in the issue, as the screenshots are there. Is that ok for you?

Yep, no problem at all - thanks!

eworm-de:
That looks perfectly fine now. 👍 Thanks a lot!

PR #1902, opened 2026-02-16
Add option for hyperthread visibility in CPU meter labels

tjk213:
This PR adds a `show_smt_cpu_labels` setting which, when enabled, labels cpu meters with a `X{a,b,c,d}` convention where `X` is the physical core ID and `a/b/c/d` indicates which hyperthread on the core the labeled meter is watching. For example, 8 cores with SMT-2 renders as follows:

```
   0a[|              ]  4a[|              ]  0b[               ]  4b[               ]
   1a[               ]  5a[|              ]  1b[               ]  5b[|              ]
   2a[               ]  6a[               ]  2b[               ]  6b[||             ]
   3a[               ]  7a[||             ]  3b[               ]  7b[               ]
```

The setting is off by default and currently only implemented for Linux.

### DIFF
--- a/CPUMeter.c
+++ b/CPUMeter.c
@@ -52,6 +52,13 @@ static void CPUMeter_init(Meter* this) {
          int coreID = Machine_getCPUPhysicalCoreID(host, cpu - 1);
          int threadIndex = Machine_getCPUThreadIndex(host, cpu - 1);
          char threadLetter = 'a' + (char)(threadIndex % 26);
+         // if we have more than 26 threads per core, then add the capital
+         // letters into the mix. If we have more than 52 threads per core, then
+         // some letters will still be repeated, but they'll be far apart from
+         // each other.
+         if ((threadIndex % 52) > 26) {
+             threadLetter -= ('a' - 'A');
+         }
          xSnprintf(caption, sizeof(caption), "%2d%c", coreID, threadLetter);
       } else {
          xSnprintf(caption, sizeof(caption), "%3u", Settings_cpuId(host->settings, cpu - 1));
--- a/CPUMeter.c
+++ b/CPUMeter.c
@@ -48,7 +48,14 @@ static void CPUMeter_init(Meter* this) {
       Meter_setCaption(this, "Avg");
    } else if (host->activeCPUs > 1) {
       char caption[10];
-      xSnprintf(caption, sizeof(caption), "%3u", Settings_cpuId(host->settings, cpu - 1));
+      if (host->settings->showCPUSMTLabels) {
+         int coreID = Machine_getCPUPhysicalCoreID(host, cpu - 1);
+         int threadIndex = Machine_getCPUThreadIndex(host, cpu - 1);
+         char threadLetter = 'a' + (char)(threadIndex % 26);
+         xSnprintf(caption, sizeof(caption), "%2d%c", coreID, threadLetter);
+      } else {
+         xSnprintf(caption, sizeof(caption), "%3u", Settings_cpuId(host->settings, cpu - 1));
+      }
       Meter_setCaption(this, caption);
    }
 }
--- a/DisplayOptionsPanel.c
+++ b/DisplayOptionsPanel.c
@@ -179,6 +179,7 @@ DisplayOptionsPanel* DisplayOptionsPanel_new(Settings* settings, ScreenManager* 
    Panel_add(super, (Object*) CheckItem_newByRef("Leave a margin around header", &(settings->headerMargin)));
    Panel_add(super, (Object*) CheckItem_newByRef("Detailed CPU time (System/IO-Wait/Hard-IRQ/Soft-IRQ/Steal/Guest)", &(settings->detailedCPUTime)));
    Panel_add(super, (Object*) CheckItem_newByRef("Count CPUs from 1 instead of 0", &(settings->countCPUsFromOne)));
+   Panel_add(super, (Object*) CheckItem_newByRef("Label CPUs based on SMT topology (e.g. 0a, 0b) instead of CPU index", &(settings->showCPUSMTLabels)));
    Panel_add(super, (Object*) CheckItem_newByRef("Update process names on every refresh", &(settings->updateProcessNames)));
    Panel_add(super, (Object*) CheckItem_newByRef("Add guest time in CPU meter percentage", &(settings->accountGuestInCPUMeter)));
    Panel_add(super, (Object*) CheckItem_newByRef("Also show CPU percentage numerically", &(settings->showCPUUsage)));
--- a/Settings.c
+++ b/Settings.c
@@ -468,6 +468,8 @@ static bool Settings_read(Settings* this, const char* fileName, const Machine* h
       } else if (String_eq(option[0], "cpu_count_from_zero")) {
          // old (inverted) naming also supported for backwards compatibility
          this->countCPUsFromOne = !atoi(option[1]);
+      } else if (String_eq(option[0], "show_cpu_smt_labels")) {
+         this->showCPUSMTLabels = atoi(option[1]);
       } else if (String_eq(option[0], "show_cpu_usage")) {
          this->showCPUUsage = atoi(option[1]);
       } else if (String_eq(option[0], "show_cpu_frequency")) {
@@ -705,6 +707,7 @@ int Settings_write(const Settings* this, bool onCrash) {
    printSettingInteger("screen_tabs", this->screenTabs);
    printSettingInteger("detailed_cpu_time", this->detailedCPUTime);
    printSettingInteger("cpu_count_from_one", this->countCPUsFromOne);
+   printSettingInteger("show_cpu_smt_labels", this->showCPUSMTLabels);
    printSettingInteger("show_cpu_usage", this->showCPUUsage);
    printSettingInteger("show_cpu_frequency", this->showCPUFrequency);
    #ifdef BUILD_WITH_CPU_TEMP
@@ -810,6 +813,7 @@ Settings* Settings_new(const Machine* host, Hashtable* dynamicMeters, Hashtable*
    this->highlightMegabytes = true;
    this->detailedCPUTime = false;
    this->countCPUsFromOne = false;
+   this->showCPUSMTLabels = false;
    this->showCPUUsage = true;
    this->showCPUFrequency = false;
    #ifdef BUILD_WITH_CPU_TEMP
--- a/dragonflybsd/DragonFlyBSDMachine.c
+++ b/dragonflybsd/DragonFlyBSDMachine.c
@@ -373,3 +373,15 @@ bool Machine_isCPUonline(const Machine* host, unsigned int id) {
    // TODO: Support detecting online / offline CPUs.
    return true;
 }
+
+int Machine_getCPUPhysicalCoreID(const Machine* host, unsigned int id) {
+   assert(id < host->existingCPUs);
+   (void) host;
+   return (int)id;
+}
+
+int Machine_getCPUThreadIndex(const Machine* host, unsigned int id) {
+   assert(id < host->existingCPUs);
+   (void) host; (void) id;
+   return 0;
+}
--- a/linux/LinuxMachine.c
+++ b/linux/LinuxMachine.c
@@ -715,6 +715,28 @@ static void LinuxMachine_assignCCDs(LinuxMachine* this, int ccds) {
    }
 }
 
+static void LinuxMachine_computeThreadIndices(LinuxMachine* this) {
+   /* For SMT/Hyperthreading: compute the thread index for each CPU.
+      CPUs sharing the same physicalID and coreID are SMT siblings.
+      threadIndex is 0 for the first thread, 1 for the second, etc. */
+
+   const Machine* super = &this->super;
+   CPUData* cpus = this->cpuData;
+
+   /* CPU 0 is the average, skip it. For each CPU, count how many
+      lower-indexed CPUs share the same physicalID and coreID. */
+   for (size_t i = 1; i <= super->existingCPUs; i++) {
+      int threadIndex = 0;
+      for (size_t j = 1; j < i; j++) {
+         if (cpus[i].physicalID == cpus[j].physicalID &&
+             cpus[i].coreID == cpus[j].coreID) {
+            threadIndex++;
+         }
+      }
+      cpus[i].threadIndex = threadIndex;
+   }
+}
+
 #endif
 
 static void LinuxMachine_scanCPUFrequency(LinuxMachine* this) {
@@ -799,6 +821,7 @@ Machine* Machine_new(UsersTable* usersTable, uid_t userId) {
    LinuxMachine_fetchCPUTopologyFromCPUinfo(this);
    int ccds = LibSensors_countCCDs();
    LinuxMachine_assignCCDs(this, ccds);
+   LinuxMachine_computeThreadIndices(this);
    #endif
 
    return super;
@@ -826,4 +849,32 @@ bool Machine_isCPUonline(const Machine* super, unsigned int id) {
 
    assert(id < super->existingCPUs);
    return this->cpuData[id + 1].online;
+}
+
+int Machine_getCPUPhysicalCoreID(const Machine* super, unsigned int id) {
+#ifdef HAVE_SENSORS_SENSORS_H
+   const LinuxMachine* this = (const LinuxMachine*) super;
+
+   assert(id < super->existingCPUs);
+
+   const CPUData* cpu = &this->cpuData[id + 1];
+   return cpu->physicalID * (this->maxCoreID + 1) + cpu->coreID;
+#else
+   (void) super;
+   /* Fall back to CPU id when topology unavailable */
+   return (int)id;
+#endif
+}
+
+int Machine_getCPUThreadIndex(const Machine* super, unsigned int id) {
+#ifdef HAVE_SENSORS_SENSORS_H
+   const LinuxMachine* this = (const LinuxMachine*) super;
+
+   assert(id < super->existingCPUs);
+   return this->cpuData[id + 1].threadIndex;
+#else
+   (void) super;
+   (void) id;
+   return 0;
+#endif
 }
--- a/linux/LinuxMachine.c
+++ b/linux/LinuxMachine.c
@@ -623,7 +623,6 @@ static void scanCPUFrequencyFromCPUinfo(LinuxMachine* this) {
    }
 }
 
-#ifdef HAVE_SENSORS_SENSORS_H
 static void LinuxMachine_fetchCPUTopologyFromCPUinfo(LinuxMachine* this) {
    const Machine* super = &this->super;
 
@@ -737,8 +736,6 @@ static void LinuxMachine_computeThreadIndices(LinuxMachine* this) {
    }
 }
 
-#endif
-
 static void LinuxMachine_scanCPUFrequency(LinuxMachine* this) {
    const Machine* super = &this->super;
 
@@ -816,13 +813,14 @@ Machine* Machine_new(UsersTable* usersTable, uid_t userId) {
    // Initialize CPU count
    LinuxMachine_updateCPUcount(this);
 
-   #ifdef HAVE_SENSORS_SENSORS_H
    // Fetch CPU topology
+   int ccds = 0;
    LinuxMachine_fetchCPUTopologyFromCPUinfo(this);
-   int ccds = LibSensors_countCCDs();
+   #ifdef HAVE_SENSORS_SENSORS_H
+   ccds = LibSensors_countCCDs();
+   #endif
    LinuxMachine_assignCCDs(this, ccds);
    LinuxMachine_computeThreadIndices(this);
-   #endif
 
    return super;
 }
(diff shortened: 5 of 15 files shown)


Explorer09:
I have a concern about using the abbreviation "SMT" as I [don't know what it refers to](https://en.wikipedia.org/wiki/SMT_(disambiguation)). Maybe you could just say "label CPU by physical cores".

Your idea should be about grouping CPU meter numbers by their physical cores, right? Whether the threads would be labelled as a, b, c, etc. is second.

natoscott:
SMT is a well-known acronym in industry, its perfectly fine to use here - it's short and precise, people who might use this feature know what it means.

Looking at the linked wikipedia page also suggests it is a good term to use too - there's clearly only one possible meaning listed there that could be relevant in the context of htop.

Explorer09:
> SMT is a well-known acronym in industry, its perfectly fine to use here - it's short and precise, people who might use this feature know what it means.
> 
> Looking at the linked wikipedia page also suggests it is a good term to use too - there's clearly only one possible meaning listed there that could be relevant in the context of htop.

I would just say "multithread" or "hyperthread" rather than throw in an industry acronym that users need to look up for what it means.

The users of htop could be very new to computers and if we have to explain an acronym in the UI, it fails at being user-friendly.

Note that in the linked Wikipedia page SMT could also refer to surface-mount technology (although I've heard of SMD being used more often than SMT in that context). At first glance I was confused with "smt" being an ugly shorthand for "smart", but anyway I would vote for spelling out the word "multithread" (or "multithreading").

tjk213:
I would agree that SMT is a standard term & that there is really only one possible fit on that wiki page. With that said, happy to rename the flag if there is consensus on a better name.

Note that the description in the UI uses the term HyperThreads as well, for users who find that more familiar:
`Show SMT/HyperThread siblings (e.g. 0a, 0b) instead of CPU number`

dragonflybsd/DragonFlyBSDMachine.c
```diff
@@ -368,8 +368,20 @@ void Machine_scan(Machine* super) {
 
 bool Machine_isCPUonline(const Machine* host, unsigned int id) {
    assert(id < host->existingCPUs);
-   (void)host; (void)id;
+   (void) host; (void) id;
```
BenBE: Unrelated change.

linux/LinuxMachine.c
```diff
@@ -827,3 +850,31 @@ bool Machine_isCPUonline(const Machine* super, unsigned int id) {
...
+
+   assert(id < super->existingCPUs);
+
+   const CPUData* cpu = &this->cpuData[id + 1];
+   return cpu->physicalID * (this->maxCoreID + 1) + cpu->coreID;
+   #else
+   (void) super;
+   /* Fall back to CPU id when topology unavailable */
+   return (int)id;
+   #endif
```
BenBE:
```suggestion
#ifdef HAVE_SENSORS_SENSORS_H
   const LinuxMachine* this = (const LinuxMachine*) super;

   assert(id < super->existingCPUs);

   const CPUData* cpu = &this->cpuData[id + 1];
   return cpu->physicalID * (this->maxCoreID + 1) + cpu->coreID;
#else
   (void) super;
   /* Fall back to CPU id when topology unavailable */
   return (int)id;
#endif
```

Best to not indent preprocessor defines (unfortuantely inconsistently done in this file).

linux/LinuxMachine.c
```diff
@@ -827,3 +850,31 @@ bool Machine_isCPUonline(const Machine* super, unsigned int id) {
...
+   #ifdef HAVE_SENSORS_SENSORS_H
+   const LinuxMachine* this = (const LinuxMachine*) super;
+
+   assert(id < super->existingCPUs);
+   return this->cpuData[id + 1].threadIndex;
+   #else
+   (void) super;
+   (void) id;
+   return 0;
+   #endif
```
BenBE:
```suggestion
#ifdef HAVE_SENSORS_SENSORS_H
   const LinuxMachine* this = (const LinuxMachine*) super;

   assert(id < super->existingCPUs);
   return this->cpuData[id + 1].threadIndex;
#else
   (void) super;
   (void) id;
   return 0;
#endif
```

Same as above.

netbsd/NetBSDMachine.c
```diff
@@ -286,8 +286,20 @@ void Machine_scan(Machine* super) {
 
 bool Machine_isCPUonline(const Machine* host, unsigned int id) {
    assert(id < host->existingCPUs);
-   (void)host; (void)id;
+   (void) host; (void) id;
```
BenBE: Unrelated whitespace change.

CPUMeter.c
```diff
@@ -48,7 +48,14 @@ static void CPUMeter_init(Meter* this) {
       Meter_setCaption(this, "Avg");
    } else if (host->activeCPUs > 1) {
       char caption[10];
-      xSnprintf(caption, sizeof(caption), "%3u", Settings_cpuId(host->settings, cpu - 1));
+      if (host->settings->showSmtCpuLabels) {
+         int coreID = Machine_getCPUPhysicalCoreID(host, cpu - 1);
+         int threadIndex = Machine_getCPUThreadIndex(host, cpu - 1);
+         char threadLetter = 'a' + (char)threadIndex;
+         xSnprintf(caption, sizeof(caption), "%2d%c", coreID, threadLetter);
```
BenBE: Properly handle more than 26 threads per core …

I know this is rare, but … ;-)

Settings.c
```diff
@@ -472,6 +472,8 @@ static bool Settings_read(Settings* this, const char* fileName, const Machine* h
          this->showCPUUsage = atoi(option[1]);
       } else if (String_eq(option[0], "show_cpu_frequency")) {
          this->showCPUFrequency = atoi(option[1]);
+      } else if (String_eq(option[0], "show_smt_cpu_labels")) {
+         this->showSmtCpuLabels = atoi(option[1]);
```
BenBE: Personally I'd go for `show_cpu_smt_labels` and `showCPUSMTLabels`

DisplayOptionsPanel.c
```diff
@@ -183,6 +183,7 @@ DisplayOptionsPanel* DisplayOptionsPanel_new(Settings* settings, ScreenManager*
    Panel_add(super, (Object*) CheckItem_newByRef("Add guest time in CPU meter percentage", &(settings->accountGuestInCPUMeter)));
    Panel_add(super, (Object*) CheckItem_newByRef("Also show CPU percentage numerically", &(settings->showCPUUsage)));
    Panel_add(super, (Object*) CheckItem_newByRef("Also show CPU frequency", &(settings->showCPUFrequency)));
+   Panel_add(super, (Object*) CheckItem_newByRef("Show SMT/HyperThread siblings (e.g. 0a, 0b) instead of CPU number", &(settings->showSmtCpuLabels)));
```
BenBE: Suggestion for more descriptive label for the option; also avoiding Intel's proprietary naming:

```suggestion
   Panel_add(super, (Object*) CheckItem_newByRef("Label CPUs based on SMT topology (e.g. 0a, 0b) instead of CPU index", &(settings->showSmtCpuLabels)));
```

Furthermore I'd suggest putting this directly below `Count CPUs from 1 instead of 0`.

BenBE (changes requested):
> I would agree that SMT is a standard term & that there is really only one possible fit on that wiki page. With that said, happy to rename the flag if there is consensus on a better name.

You can keep SMT. No need for using Intel's proprietary name. Also multithreading can be mistaken for running multiple threads (LWP) of a process, thus also something you'd better avoid.

> Note that the description in the UI uses the term HyperThreads as well, for users who find that more familiar: `Show SMT/HyperThread siblings (e.g. 0a, 0b) instead of CPU number`

I'd suggest to hint to the CPU topology in addition to SMT, which should make the intent of this setting more obvious even to people who don't know what SMT is.

Explorer09:
> > I would agree that SMT is a standard term & that there is really only one possible fit on that wiki page. With that said, happy to rename the flag if there is consensus on a better name.
> 
> You can keep SMT. No need for using Intel's proprietary name. Also multithreading can be mistaken for running multiple threads (LWP) of a process, thus also something you'd better avoid.

When I googled the term "SMT", the first few results I got was pages about surface mount technology. It suggested me that the "SMT" is not _that_ universally used as you would think.

Also, I don't think user would confuse CPU threads with process threads. Those are different things.

<s>As I said, I don't like throwing in a new term without explanation. Even "CPU thread" is better term than "SMT" for the UI (user interface) here. Since you are bothered to spell out the word "Hyperthreads", then "CPU threads" could be a shorter term for conveying the same thing.</s>

EDIT: Read this [comment](https://github.com/htop-dev/htop/pull/1902#discussion_r2815717428) for my updated opinion. I no longer object the term SMT but I would like clarification through surrounding text.

CPUMeter.c
```diff
@@ -48,7 +48,14 @@ static void CPUMeter_init(Meter* this) {
       Meter_setCaption(this, "Avg");
    } else if (host->activeCPUs > 1) {
       char caption[10];
-      xSnprintf(caption, sizeof(caption), "%3u", Settings_cpuId(host->settings, cpu - 1));
+      if (host->settings->showCPUSMTLabels) {
+         int coreID = Machine_getCPUPhysicalCoreID(host, cpu - 1);
+         int threadIndex = Machine_getCPUThreadIndex(host, cpu - 1);
+         char threadLetter = 'a' + (char)(threadIndex % 26);
```
tjk213, line 54: @BenBE I added a wrap so any systems with more than 26 threads per core will just start reusing letters. does that work for you, or would you prefer something else?

natoscott: Most I've ever heard of is 8 per core, so this seems a sensible approach.  Consider adding an assert() so that some hapless developer someday will hit this if ever that limit is reached.

Explorer09: If we distinguish letter cases (`a-z` and `A-Z`) then the number of addressable threads can be bumped up to 52. However, I am not sure what's the use case. Since we took a character in the CPU meter label to number the thread, then the maxinum addtressable physical cores went down to 99, and that's another limit worth concerning (and discussion).

EDIT: By the way, would there be users who prefer labeling the cores by letters and threads by numbers?

DisplayOptionsPanel.c
```diff
@@ -179,7 +179,7 @@ DisplayOptionsPanel* DisplayOptionsPanel_new(Settings* settings, ScreenManager*
    Panel_add(super, (Object*) CheckItem_newByRef("Leave a margin around header", &(settings->headerMargin)));
    Panel_add(super, (Object*) CheckItem_newByRef("Detailed CPU time (System/IO-Wait/Hard-IRQ/Soft-IRQ/Steal/Guest)", &(settings->detailedCPUTime)));
    Panel_add(super, (Object*) CheckItem_newByRef("Count CPUs from 1 instead of 0", &(settings->countCPUsFromOne)));
-   Panel_add(super, (Object*) CheckItem_newByRef("Label CPUs based on SMT topology (e.g. 0a, 0b) instead of CPU index", &(settings->showSmtCpuLabels)));
+   Panel_add(super, (Object*) CheckItem_newByRef("Label CPUs based on SMT topology (e.g. 0a, 0b) instead of CPU index", &(settings->showCPUSMTLabels)));
```
Explorer09, line 182: How about "Label CPUs by SMT cores and threads (e.g. 1a, 1b, 2a, 2b)"?

Note: The SMT term is kept here, but I add "cores and threads" to clarify the meaning and what the option does.

CPUMeter.c
BenBE: While wrapping around (and throwing an assert) mitigates the immediate problem, I think a proper fallback like `%d:%d` might be a more sustainable way to go.

Two options available: Use single letter for first like 26 (or 16) threads, numbers for anything above, or always use numbers if there's some core with more than 26 (or 16) threads. Doing the latter is more complexity to implement, but gives a cleaner UI/UX.

BenBE:
@tjk213 Please rebase and squish your commits to show the necessary changes instead of the path taken to get there. Doing like 3 commits (one introducing the setting, one introducing the data collection required, and one implementing the display) fully suffices. No need to spam like 14 commits at this change. ;-)

CPUMeter.c
Explorer09: @BenBE There are only three characters of space for a label in a meter, so the thread numbering has to constrain to one or two characters. The extra character like a colon `:` is not an option.

I think a much easier, temporary solution is to mark threads greater than 26 as `*` instead of a letter. Yes, that would create duplicate labels, but as long as the threads are grouped together, I don't see that would create a big problem.

tjk213: if there is a clear consensus on what you folks would like to see I'm happy to implement any of these suggestions.

+1 to @Explorer09's point that expanding beyond 3 characters creates more problems than it solves.

My take is that we should leave this as-is, with just a wrap around the lower-case letters and no assert. The main intent here is just to serve as a reminder that "not all cores are created equal" and make it clear how the cores are paired. with the default setting, one might naively assume that hyperthreads are co-located (even/odd) rather than the more common hi/lo pattern. This intent is still accomplished even if some letters are repeated. at SMT-26, you'd have many columns & rows of meters, the repeated letters would be so far apart from each other that you'd hardly notice.

And of course the setting is off by default, so if there is some system out there with SMT-26 and the user doesn't like the repeated letters, they can just not use the setting.

Explorer09: I am against repeating letters, by the way. Imagine when you communicate with someone and point out that threads "16a" to "16d" are always 100% busy, and your friend asks you which "16a" thread you are referring to.

An asterisk would be better if we run out of letters.

BenBE: Let's use CJK characters in Unicode-enabled versions of htop. That way we don't need to consider multi-character identifier variants until we get to well above 100k threads …

Hey, core 42炎 is in flames!

/jk

Okay, back to the serious side of things: Priority should be on making things distinguishable. That's an issue that we already face with systems with more than 999 cores (which exist in data centers). So regardless of what notation we choose, there will be a limit on how many cores we can uniquely identify.

And to give a point of relevance: On UltraSPARC Niagara systems it is feasible to run htop where it could encounter topologies with 32 or more threads per core.  Similarly libvirt/QEMU (KVM) will present you with about anything you ask it to show to a guest. So keeping this in mind when devising a solution for marking the topology is crucial.

Explorer09: > Let's use CJK characters in Unicode-enabled versions of htop. That way we don't need to consider multi-character identifier variants until we get to well above 100k threads …
> 
> Hey, core 42炎 is in flames!

As a native Chinese speaker I can't understand this joke. Besides, CJK characters occupy two terminal columns per character, and we have only three terminal columns for a meter...

(On the other hand, it is indeed possible to [label a thousand CPU cores](https://en.wikipedia.org/wiki/Thousand_Character_Classic) with two terminal columns of space if you indeed use CJK. The catch is, almost no one memorises the order of those characters.)

> 
> Okay, back to the serious side of things: Priority should be on making things distinguishable. That's an issue that we already face with systems with more than 999 cores (which exist in data centers). So regardless of what notation we choose, there will be a limit on how many cores we can uniquely identify.
> 
> And to give a point of relevance: On UltraSPARC Niagara systems it is feasible to run htop where it could encounter topologies with 32 or more threads per core. Similarly libvirt/QEMU (KVM) will present you with about anything you ask it to show to a guest. So keeping this in mind when devising a solution for marking the topology is crucial.

Good info for a 32 thread system. But then, what prevents us from using both uppercase and lowercase letters? That way we can label up to 52 threads.

tjk213: i pushed an update to add capital letters

tjk213: thanks for the info about sparc & qemu, that's super interesting!

tjk213: @BenBE what do you think? can we merge?

linux/LinuxMachine.c
```diff
@@ -827,3 +850,31 @@ bool Machine_isCPUonline(const Machine* super, unsigned int id) {
    assert(id < super->existingCPUs);
    return this->cpuData[id + 1].online;
 }
+
+int Machine_getCPUPhysicalCoreID(const Machine* super, unsigned int id) {
+#ifdef HAVE_SENSORS_SENSORS_H
```
natoscott: @tjk213 why is this section protected by ifdef HAVE_SENSORS_SENSORS_H?  Seems an unrelated guard, and there's nothing here needing ifdef protection anyways (no external API calls)...?

linux/LinuxMachine.c
```diff
@@ -827,3 +850,31 @@ bool Machine_isCPUonline(const Machine* super, unsigned int id) {
...
+   return cpu->physicalID * (this->maxCoreID + 1) + cpu->coreID;
+#else
+   (void) super;
+   /* Fall back to CPU id when topology unavailable */
+   return (int)id;
+#endif
+}
+
+int Machine_getCPUThreadIndex(const Machine* super, unsigned int id) {
+#ifdef HAVE_SENSORS_SENSORS_H
```
natoscott: Likewise... am I missing something?

natoscott (changes requested):
LGTM except for the ifdef issue needing clarification (pretty sure we can just remove those ifdefs).

linux/LinuxMachine.h
```diff
@@ -53,6 +53,7 @@ typedef struct CPUData_ {
    int physicalID;      /* different for each CPU socket */
    int coreID;          /* same for hyperthreading */
    int ccdID;           /* same for each AMD chiplet */
+   int threadIndex;     /* SMT thread index: 0 for first thread, 1 for second, etc. */
```
natoscott, line 57: @tjk213  so, I think the ifdef below this line can move up 4 lines to just below temperature - AFAICS this PR is not dependent on libsensors right?  (the new information and other IDs here all come from /proc).

natoscott:
@tjk213 could you take a look into the ifdef issue?  It'd be good to get this feature into the pending htop-3.5.0 release - thanks!

linux/LinuxMachine.c
tjk213: yep, you're right. we can trim the guarded regions down to just the temperature.

currently the topology parsing code like `LinuxMachine_fetchCPUTopologyFromCPUinfo()` is guarded and i didn't want to change that. but it's only called once at startup, so it should be fine. I'll push an update

tjk213:
@natoscott done! thanks for the ping!

lilydjwg:
For multi-CPU systems the result doesn't seem to be great:

图片

This is a dual-CPU system with 12 cores / 24 threads on each CPU. Core numbers 6, 7, 20 and 21 are not used.

tjk213:
@lilydjwg can you post the output of `lscpu` and `cat /proc/cpuinfo`? I will take a look

lilydjwg:
<details>
<summary>lscpu</summary>

```
Architecture:                    x86_64
CPU op-mode(s):                  32-bit, 64-bit
Address sizes:                   46 bits physical, 48 bits virtual
Byte Order:                      Little Endian
CPU(s):                          48
On-line CPU(s) list:             0-47
Vendor ID:                       GenuineIntel
BIOS Vendor ID:                  Intel
Model name:                      Intel(R) Xeon(R) CPU E5-2678 v3 @ 2.50GHz
BIOS Model name:                 Intel(R) Xeon(R) CPU E5-2678 v3 @ 2.50GHz  CPU @ 2.5GHz
BIOS CPU family:                 179
CPU family:                      6
Model:                           63
Thread(s) per core:              2
Core(s) per socket:              12
Socket(s):                       2
Stepping:                        2
CPU(s) scaling MHz:              82%
CPU max MHz:                     3300.0000
CPU min MHz:                     1200.0000
BogoMIPS:                        4999.57
Flags:                           fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx pdpe1gb rdtscp lm constant_tsc arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc cpuid aperfmperf pni pclmulqdq dtes64 monitor ds_cpl vmx smx est tm2 ssse3 sdbg fma cx16 xtpr pdcm pcid dca sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand lahf_lm abm cpuid_fault epb invpcid_single pti ssbd ibrs ibpb stibp tpr_shadow vnmi flexpriority ept vpid ept_ad fsgsbase tsc_adjust bmi1 avx2 smep bmi2 erms invpcid cqm xsaveopt cqm_llc cqm_occup_llc dtherm ida arat pln pts md_clear flush_l1d
Virtualization:                  VT-x
L1d cache:                       768 KiB (24 instances)
L1i cache:                       768 KiB (24 instances)
L2 cache:                        6 MiB (24 instances)
L3 cache:                        60 MiB (2 instances)
NUMA node(s):                    2
NUMA node0 CPU(s):               0,2,4,6,8,10,12,14,16,18,20,22,24,26,28,30,32,34,36,38,40,42,44,46
NUMA node1 CPU(s):               1,3,5,7,9,11,13,15,17,19,21,23,25,27,29,31,33,35,37,39,41,43,45,47
Vulnerability Itlb multihit:     KVM: Mitigation: Split huge pages
Vulnerability L1tf:              Mitigation; PTE Inversion; VMX conditional cache flushes, SMT vulnerable
Vulnerability Mds:               Mitigation; Clear CPU buffers; SMT vulnerable
Vulnerability Meltdown:          Mitigation; PTI
Vulnerability Mmio stale data:   Mitigation; Clear CPU buffers; SMT vulnerable
Vulnerability Retbleed:          Not affected
Vulnerability Spec store bypass: Mitigation; Speculative Store Bypass disabled via prctl
Vulnerability Spectre v1:        Mitigation; usercopy/swapgs barriers and __user pointer sanitization
Vulnerability Spectre v2:        Mitigation; Retpolines, IBPB conditional, IBRS_FW, STIBP conditional, RSB filling, PBRSB-eIBRS Not affected
Vulnerability Srbds:             Not affected
Vulnerability Tsx async abort:   Not affected
```
</details>

[cpuinfo.txt](https://github.com/user-attachments/files/26368644/cpuinfo.txt)

tjk213:
@lilydjwg i think the code is working as expected.
the missing CPU ids in the GUI are also missing your cpuinfo.txt.
presumably, those IDs are missing from your cpuinfo because your silicon is a chip that actually has 14 cores per socket, but 2 of them have been disabled.

I posted a fix in #1940 that will render this more cleanly with contiguous IDs

lilydjwg:
I see, thank you. I didn't expect disabled cores still occupy ids.

PR #1903, opened 2026-02-17
pcp/PCPDynamicMeter.c: guard against underflow of buffer[] index

kmcdonell:
On the error paths, the index (bytes) was unconditionally decremented but it may be 0 (for failure on the first metric), leading to underflow and a huge value (bytes is size_t and hence unsigned).

Fixes Coverity CID 426719 over at
https://scan4.scan.coverity.com/#/project-view/38378/14885?selectedIssue=426719

### DIFF
--- a/pcp/PCPDynamicMeter.c
+++ b/pcp/PCPDynamicMeter.c
@@ -330,7 +330,8 @@ void PCPDynamicMeter_updateValues(PCPDynamicMeter* this, Meter* meter) {
       pmAtomValue atom, raw;
 
       if (!Metric_values(base, &raw, 1, desc->type)) {
-         bytes--; /* clear the separator */
+	 if (bytes > 0)
+	     bytes--; /* clear the separator */
          continue;
       }
 
@@ -342,7 +343,8 @@ void PCPDynamicMeter_updateValues(PCPDynamicMeter* this, Meter* meter) {
       if (desc->type == PM_TYPE_STRING)
          atom = raw;
       else if (pmConvScale(desc->type, &raw, &desc->units, &atom, &conv) < 0) {
-         bytes--; /* clear the separator */
+	 if (bytes > 0)
+	     bytes--; /* clear the separator */
          continue;
       }
 


pcp/PCPDynamicMeter.c
```diff
@@ -330,7 +330,8 @@ void PCPDynamicMeter_updateValues(PCPDynamicMeter* this, Meter* meter) {
       pmAtomValue atom, raw;
 
       if (!Metric_values(base, &raw, 1, desc->type)) {
-         bytes--; /* clear the separator */
+	 if (bytes > 0)
```
BenBE:
```suggestion
      if (bytes > 0)
```

pcp/PCPDynamicMeter.c
```diff
@@ -342,7 +343,8 @@ void PCPDynamicMeter_updateValues(PCPDynamicMeter* this, Meter* meter) {
       if (desc->type == PM_TYPE_STRING)
          atom = raw;
       else if (pmConvScale(desc->type, &raw, &desc->units, &atom, &conv) < 0) {
-         bytes--; /* clear the separator */
+	 if (bytes > 0)
```
BenBE:
```suggestion
      if (bytes > 0)
```

BenBE:
That patch is incomplete AFAICT. I'd propose the following variant instead:

```diff
diff --git a/pcp/PCPDynamicMeter.c b/pcp/PCPDynamicMeter.c
index 5bf1870d..b034a39c 100644
--- a/pcp/PCPDynamicMeter.c
+++ b/pcp/PCPDynamicMeter.c
@@ -316,11 +316,14 @@ void PCPDynamicMeter_enable(PCPDynamicMeter* this) {
 }
 
 void PCPDynamicMeter_updateValues(PCPDynamicMeter* this, Meter* meter) {
-   char* buffer = meter->txtBuffer;
-   size_t size = sizeof(meter->txtBuffer);
+   char* const buffer = meter->txtBuffer;
+   const size_t size = sizeof(meter->txtBuffer);
    size_t bytes = 0;
+   size_t bytes_old;
 
    for (size_t i = 0; i < this->totalMetrics; i++) {
+      bytes_old = bytes;
+
       if (i > 0 && bytes < size - 1)
          buffer[bytes++] = '/';  /* separator */
 
@@ -330,7 +333,7 @@ void PCPDynamicMeter_updateValues(PCPDynamicMeter* this, Meter* meter) {
       pmAtomValue atom, raw;
 
       if (!Metric_values(base, &raw, 1, desc->type)) {
-         bytes--; /* clear the separator */
+         bytes = bytes_old; /* clear the separator */
          continue;
       }
 
@@ -342,7 +345,7 @@ void PCPDynamicMeter_updateValues(PCPDynamicMeter* this, Meter* meter) {
       if (desc->type == PM_TYPE_STRING)
          atom = raw;
       else if (pmConvScale(desc->type, &raw, &desc->units, &atom, &conv) < 0) {
-         bytes--; /* clear the separator */
+         bytes = bytes_old; /* clear the separator */
          continue;
       }
 
@@ -390,6 +393,8 @@ void PCPDynamicMeter_updateValues(PCPDynamicMeter* this, Meter* meter) {
          bytes += xSnprintf(buffer + bytes, size - bytes, "%s", metric->suffix);
    }
 
+   buffer[CLAMP(bytes, 0u, size - 1)] = '\0';
+
    if (!bytes)
       xSnprintf(buffer, size, "no data");
 }
```

@natoscott Can you take a look?

BenBE:
@kmcdonell Your link to Coverity doesn't work for me.

kmcdonell:
@BenBE I guess you need some sort of login at scan4.coverity.com ... the essence of the issue there is:
```
332      if (!Metric_values(base, &raw, 1, desc->type)) {
     	5. known_value_assign: bytes = bytes, its value is now 18446744073709551615.
     	
CID 426719: (#1 of 1): Overflowed constant (INTEGER_OVERFLOW)
6. overflow_const: Expression bytes--, where bytes is known to be equal to 0, underflows the type of bytes--, which is type size_t.
333         bytes--; /* clear the separator */
334         continue;
```
Coverity missed the second bytes-- case.

I am only trying to fix this issue ... other than the null-byte termination (which is outside the scope of what I was trying to fix), your change looks to be functionally equivalent to mine, but yours has an extra variable and assignment in the loop.

natoscott:
> Your link to Coverity doesn't work for me.

Only project maintainers (well, invited people) can view covscan results as there's at times security aspects involved.  I'll look into getting htop setup like pcp for regular scans.

natoscott:
> [...] change looks to be functionally equivalent

I think both fix the covscan issue and @BenBE has found (and fixed) a subtle second issue, where an error on the final loop iteration could leave a trailing separator behind unintentionally.

PR #1909, opened 2026-02-19
Allow for optimizing out bound check on 64 bit systems

BenBE:
This re-wording of the previous snippet hints to the compiler more directly, that this condition can be dropped if the type of `size_t` is narrower than the one used for `OEISprimes` (`uint64_t`).

This optimization applies for 32 bit systems.

NB: This saves some precious CPU cycles on 64 bit systems … ;-)


Explorer09:
I've updated my PR #1842 to reflect the changes done here.

PR #1912, opened 2026-02-21
Harden and optimize CPU time scanning on Linux

BenBE:
Addresses covscan 644309.

First off: Do the overflow checking early to catch an issue with handling very large CPU thread counts.

Second, optimize the way marking of offline CPU threads is done.

Following this optimization, we can reduce our memory usage by 1 (in words: ONE) byte. ;-)

Overall this avoids putting the CPU timing data into the wrong spot for some edge cases …

### DIFF
--- a/linux/LinuxMachine.c
+++ b/linux/LinuxMachine.c
@@ -408,9 +408,9 @@ static void LinuxMachine_scanCPUTime(LinuxMachine* this) {
    if (!file)
       CRT_fatalError("Cannot open " PROCSTATFILE);
 
-   // Add an extra phantom thread for a later loop
-   assert(super->existingCPUs < UINT_MAX - 2);
-   bool adjCpuIdProcessed[super->existingCPUs + 2];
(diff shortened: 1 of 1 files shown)


linux/LinuxMachine.c
```diff
@@ -483,15 +486,10 @@ static void LinuxMachine_scanCPUTime(LinuxMachine* this) {
    }
 
    // Set the extra phantom thread as checked to make sure to mark trailing offline threads correctly in the loop
```
fasterit: That comment can go the way of the dodo as well

fasterit (changes requested):
LGTM, tested functional

The one stray comment should go, then GTM from me.
/DLange

linux/LinuxMachine.c
BenBE: Comment removed and updated the comment mentioning the phantom thread at the declaration site …

PR #1916, opened 2026-02-24
XUtil: const-correctnes for strchrnul

BenBE:
This contains multiple parts:

1. Fix a few warnings related to const-correctness and `strchrnul`
2. Fix the buffer overflow handling in `LinuxProcessTable_readCGroupFile` by handling the `snprintf` retur values
3. Actually advance our write location allowing for multiple entries to be read

### DIFF
--- a/linux/LinuxProcessTable.c
+++ b/linux/LinuxProcessTable.c
@@ -1072,6 +1072,7 @@ static void LinuxProcessTable_readCGroupFile(LinuxProcess* process, openat_arg_t
          break;
       }
 
+      at += (size_t)wrote;
       left -= (size_t)wrote;
    }
    fclose(file);


linux/LinuxProcessTable.c
```diff
@@ -1047,16 +1047,17 @@ static void LinuxProcessTable_readCGroupFile(LinuxProcess* process, openat_arg_t
...
          group = String_strchrnul(group, ':');
          if (!*group)
             break;
          group++;
       }
 
-      char* eol = String_strchrnul(group, '\n');
-      *eol = '\0';
+      const char* eol = String_strchrnul(group, '\n');
+      char* eol_w = &buffer[eol - buffer];
```
Explorer09, line 1060: How does this `&buffer[eol - buffer];` not trigger warnings at all?

Update: I have a better idea to the problem:

```c
size_t eol_offset = (size_t)(eol - (const char*)buffer);
buffer[eol_offset] = '\0';
```

BenBE: Based on my reading of the ISO C99 standard:

- `&p[i]` returns a pointer to the `i`th element in `p` using the type of `p`
- `b - a` returns `uintptr_t` in units of the pointed-to types for any `a` and `b` that point inside the same object (array) in memory, if the pointed-to types are compatible, and not incomplete; irrespective of cv-qualifiers.

Explorer09: `b - a` should return a `ptrdiff_t`, but if they are in different types a compiler may label it as a warning (a strict aliasing violation, maybe). A compiler wasn't told that the pointer returned from strchrnul is derived from the first argument passed into it, and so it doesn't know that `b - a` would work unless it can examine the body of strchrnul function.

Note: I don't like the constness thing at all. Part of the PR #1915 that I was recently doing had been dealing with `-Wcast-qual` warnings from GCC that over-warns things. (Not strictly a false positive, but GCC wasn't taught that `execv()` is special that it has to make an exception and cast away the const qualifier.)

PR #1918, opened 2026-03-04
update mem info on every refresh on Solaris platform

ViryBe:
# The bug
`htop` crashes on illumos when the cpu count changes.

Tested with latest main branch on an illumos system. I unfortunately don't have a Solaris machine to test on it, but according to the man pages the bug should be the same.
 
How to reproduce:

1. Open `htop` in a terminal
2. Disable CPU0:
```sh
# as root
psradm -f 0
# to re-enable CPU0
psradm -Fn 0
```

`htop` crashed in first terminal.

# Issue and Fix

The memory info was previously read only once and kept in a static variable. However, the pointer to the memory info is invalidated when the kstat chain is updated, such as when the cpu count changes.

From the Illumos `kstat_chain_update` man page:

```
[...] any kstat_t structures retrieved by the
kstat_lookup(3KSTAT) function or data pointers obtained through the
kstat_data_lookup(3KSTAT) function are invalidated by a call to
kstat_chain_update(). [...]
```

So, re-reading the struct `kstat_t` on every memory scan ensure that the pointer `meminfo` is always valid.

Note that I also replace the 2 calls to `kstat_open` and `kstat_close` by a single `kstat_chain_update`.

### DIFF
--- a/solaris/SolarisMachine.c
+++ b/solaris/SolarisMachine.c
@@ -61,10 +61,9 @@ static void SolarisMachine_updateCPUcount(SolarisMachine* this) {
    }
 
    if (change) {
-      kstat_close(this->kd);
-      this->kd = kstat_open();
-      if (!this->kd)
-         CRT_fatalError("Cannot open kstat handle");
+      kid_t update_kid = kstat_chain_update(this->kd);
+      if (update_kid < 0)
+         CRT_fatalError("Cannot update kstat chain");
    }
 }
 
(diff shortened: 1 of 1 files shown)


natoscott:
@ViryBe thanks for looking into it.

> I unfortunately don't have a Solaris machine to test on it

It'd be really good to test this no a Solaris machine before proceeding further - is there anyone out there who can help?

ViryBe:
yeah, I'll try to install solaris in a VM to test the patch if I have some time in the following days

solaris/SolarisMachine.c
```diff
@@ -30,6 +30,7 @@ static void SolarisMachine_updateCPUcount(SolarisMachine* this) {
    Machine* super = &this->super;
    long int s;
    bool change = false;
+   kid_t update_kid = -1;
```
BenBE: Reduce scope to minimum needed.

ViryBe:
Thanks for the feedback @BenBE.

I tested in a Solaris 11.4 VM and I can confirm that the bug is also present and the fix works as intended.

PR #1919, opened 2026-03-15
pipe() & fork() code quality improvements

Explorer09:
* The 'fdpairs' for pipe() are better initialized with values {-1, -1} as FD 0 is a valid FD (stdin). This change mainly helps debuggers of htop as they won't see junk values when pipe() fails.
* Make the return value checks of pipe() and fork() consistent throughout the codebase.

### DIFF
--- a/OpenFilesScreen.c
+++ b/OpenFilesScreen.c
@@ -101,13 +101,13 @@ static OpenFiles_ProcessData* OpenFilesScreen_getProcessData(pid_t pid) {
    pdata->cols[getIndexForType('i')] = 8;
 
    int fdpair[2] = {-1, -1};
-   if (pipe(fdpair) == -1) {
+   if (pipe(fdpair) < 0) {
       pdata->error = 1;
       return pdata;
    }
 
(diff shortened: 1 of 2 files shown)


linux/OpenRCMeter.c
```diff
@@ -48,8 +48,8 @@ static void updateViaExec(bool user) {
    if (Settings_isReadonly())
       return;
 
-   int fdpair[2];
-   if (pipe(fdpair) < 0)
+   int fdpair[2] = {-1, -1};
+   if (pipe(fdpair) != 0)
```
BenBE: Check `< 0` for failure; consistent with usual kernel error checks.

Explorer09: Just wondering, is there a document about POSIX's convention of return values with these system calls? Such as whether the standard does reserve all nonnegative return values for successful operations.

BenBE: The negative return value for error situations is a Linux kernel convention AFAIR.

PR #1921, opened 2026-03-15
Fix CPU virtualization bar color and help text in non-detailed mode

### LINKED ISSUE #1920: Help is missing description of orange-colored CPU bar
The "Help" screen is missing a description of orange-colored CPU bar. Thus, I don't understand what the orange bar refers to.

**Help screen:**

Image

**CPU bars:**

Image

I think the solution is to add a description of that to the Help screen for CPU bars.

I rarely see the orange bars. They've come up now especially as I'm running a QEMU VM though, which makes me think they're maybe related to the "guest" bar (or something similar).

natoscott:
The overlaying of CPU virt times (guest+steal) into the IRQ slot is problematic when displaying CPU time in bar mode. It results in the IRQ color being used for virt, which doesn't match up with the Help text for CPU util when in the non-detailed CPU time mode.  The text 'guest' displayed on the Help window is also misleading because the value is actually the sum of guest+steal time - use 'virt' instead, to indicate this is not guest time only.

Resolves #1920

### DIFF
--- a/Action.c
+++ b/Action.c
@@ -744,8 +744,8 @@ static Htop_Reaction actionHelp(State* st) {
       addbartext(CRT_colors[CPU_IOWAIT], "/", "io-wait");
       addbartext(CRT_colors[BAR_SHADOW], " ", "used%");
    } else {
-      addbartext(CRT_colors[CPU_GUEST], "/", "guest");
-      addbartext(CRT_colors[BAR_SHADOW], "                            ", "used%");
+      addbartext(CRT_colors[CPU_GUEST], "/", "virt");
+      addbartext(CRT_colors[BAR_SHADOW], "                             ", "used%");
    }
    addattrstr(CRT_colors[BAR_BORDER], "]");
 
--- a/CPUMeter.c
+++ b/CPUMeter.c
@@ -36,6 +36,13 @@ static const int CPUMeter_attributes[] = {
    CPU_IOWAIT
 };
 
+static const int CPUMeter_attributes_summary[] = {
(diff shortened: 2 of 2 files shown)


Explorer09:
Changing the attributes of CPUMeter during the middle of htop running can break graph display (perhaps you won't see it until the color graph (#714) mode feature is added).

Another problem is there is a time discrepancy between updating the settings and the actual CPU time is refetched. So there could be a short time frame that the CPUMeter data is desynchronized.

natoscott:
> Changing the attributes of CPUMeter during the middle of htop running can break graph display (perhaps you won't see it until the color graph (#714) mode feature is added).
> 

That PR is already broken then, as this is how the Load Average meter works already.

> Another problem is there is a time discrepancy between updating the settings and the actual CPU time is refetched. So there could be a short time frame that the CPUMeter data is desynchronized.

That may be a different bug (I doubt it TBH), but its not relevant to the (color, not value) fix here.

Explorer09:
I would like you to review and comment on #1922 which is my approach to the problem.

> > Changing the attributes of CPUMeter during the middle of htop running can break graph display (perhaps you won't see it until the color graph (#714) mode feature is added).
>
> That PR is already broken then, as this is how the Load Average meter works already.

Not quite. The Load Average meter still works with #714 because the color change only affects bar display and not graph display at all.

natoscott:
> I would like you to review and comment on #1922 which is my approach to the problem.

I prefer the simpler fix proposed here.

PR #1922, opened 2026-03-17
Use CPU_METER_STEAL for 'virtualized' CPU time in non-detailed mode

### LINKED ISSUE #1920: Help is missing description of orange-colored CPU bar

Explorer09:
The virtualized CPU time will now write to the `CPU_METER_STEAL` item in the non-detailed CPU meter mode. Also fix the CPU meter text in the help screen.

This pull request also includes code cleanups such as unnecessary assignments of `v[CPU_METER_IRQ]` in NetBSD and OpenBSD platforms.

I think my approach is better than what's proposed in #1921.

Resolves #1920.

### DIFF
--- a/pcp/Platform.c
+++ b/pcp/Platform.c
@@ -576,10 +576,12 @@ static double Platform_setOneCPUValues(Meter* this, const Settings* settings, pm
       v[CPU_METER_KERNEL] = values[CPU_SYSTEM_ALL_PERIOD].ull / total * 100.0;
       this->curItems = 3;
 
+      v[CPU_METER_IRQ] = 0.0;
+      v[CPU_METER_SOFTIRQ] = 0.0;
       value = values[CPU_STEAL_PERIOD].ull + values[CPU_GUEST_PERIOD].ull;
-      v[CPU_METER_IRQ] = value / total * 100.0;
+      v[CPU_METER_STEAL] = value / total * 100.0;
       if (settings->accountGuestInCPUMeter) {
-         this->curItems = 4;
+         this->curItems = 6;
       }
    }
 
--- a/pcp/Platform.c
+++ b/pcp/Platform.c
@@ -574,9 +574,13 @@ static double Platform_setOneCPUValues(Meter* this, const Settings* settings, pm
       v[CPU_METER_IOWAIT]  = values[CPU_IOWAIT_PERIOD].ull / total * 100.0;
    } else {
       v[CPU_METER_KERNEL] = values[CPU_SYSTEM_ALL_PERIOD].ull / total * 100.0;
+      this->curItems = 3;
+
       value = values[CPU_STEAL_PERIOD].ull + values[CPU_GUEST_PERIOD].ull;
       v[CPU_METER_IRQ] = value / total * 100.0;
-      this->curItems = 4;
+      if (settings->accountGuestInCPUMeter) {
+         this->curItems = 4;
+      }
    }
 
    percent = sumPositiveValues(v, this->curItems);
--- a/Action.c
+++ b/Action.c
@@ -804,8 +804,8 @@ static Htop_Reaction actionHelp(State* st) {
       addbartext(CRT_colors[CPU_IOWAIT], "/", "io-wait");
       addbartext(CRT_colors[BAR_SHADOW], " ", "used%");
    } else {
-      addbartext(CRT_colors[CPU_GUEST], "/", "guest");
-      addbartext(CRT_colors[BAR_SHADOW], "                            ", "used%");
+      addbartext(CRT_colors[CPU_GUEST], "/", "virtualized");
+      addbartext(CRT_colors[BAR_SHADOW], "                      ", "used%");
    }
    addattrstr(CRT_colors[BAR_BORDER], "]");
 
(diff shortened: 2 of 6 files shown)


pcp/Platform.c
```diff
@@ -560,9 +560,15 @@ static double Platform_setOneCPUValues(Meter* this, const Settings* settings, pm
...
    } else {
       v[CPU_METER_KERNEL] = values[CPU_SYSTEM_ALL_PERIOD].ull / total * 100.0;
+      this->curItems = 3;
+
+      v[CPU_METER_IRQ] = 0.0;
+      v[CPU_METER_SOFTIRQ] = 0.0;
       value = values[CPU_STEAL_PERIOD].ull + values[CPU_GUEST_PERIOD].ull;
-      v[CPU_METER_IRQ] = value / total * 100.0;
-      this->curItems = 4;
+      v[CPU_METER_STEAL] = value / total * 100.0;
```
natoscott, line 582: "value" here is not "steal" time though - its the sum of steal+guest now.  So this hasn't really solved things for your unmerged graphing PR.  IOW, there's a problem switching on/off detailed CPU mode because the historical data in the steal slot transitions between sum/not-sum/sum/not-sum ... so the "steal" value is incorrect whenever detailed mode changes state.

Explorer09: I didn't get it… What are you suggesting? I think the historical data having "sum/not-sum/sum/not-sum", will not present any visual display glitch.

Because "steal" and "guest" are presented as the same color in all color schemes of htop

natoscott: | Because "steal" and "guest" are presented as the same color

Good point, it'll work by luck until someone wants to separate out those categories like in most tools.

Explorer09: > 
> Good point, it'll work by luck until someone wants to separate out those categories like in most tools.

I won't call it working "by luck", because I can anticipate that the "steal" and "guest" colors be separate like this:

```c
[CPU_STEAL] = A_BOLD | ColorPair(Cyan, Black),
[CPU_GUEST] = ColorPair(Cyan, Black),
```

But then, how is it wrong by merging the "steal" and "guest" into a single item in the non-detailed CPU meter mode?

Isn't that the whole point of the non-detailed mode? That the unnecessary details of CPU time items can be merged (just like "irq" and "sort-irq" merged with "kernel")?

I really don't get what you are suggesting.

natoscott: | I won't call it working "by luck",

It is just luck - we very rarely have two items sharing a color in a Meter.

| But then, how is it wrong by merging the "steal" and "guest" into a single item

Its not wrong - it just remains obfuscated - like the existing code has always been, overloading one CPU values slot with multiple values from another.  The bug can be fixed by a trivial change compared to this proposal, so its a nack for me.

Explorer09: > 
> | But then, how is it wrong by merging the "steal" and "guest" into a single item
> 
> Its not wrong - it just remains obfuscated - like the existing code has always been, overloading one CPU values slot with multiple values from another. The bug can be fixed by a trivial change compared to this proposal, so its a nack for me.

I would claim the use of IRQ slot for virtual CPU time is plain wrong despite your fix looks trivial. The CPU meter code has fixed ("fixing" for the "making things unmovable" meaning) that item/slot as `CPU_METER_IRQ` and it shouldn't be overloaded with another use such as "virtualized". It breaks forward compatibility - such as with the not-yet-merged #714.

#714 relies on an assumption that the meanings of item indices of a meter to remain stable. It's also so for the meter character allocation (`|#*@$%&.`) in the monochrome mode. It would be more of a UX surprise when the `@` character in CPU meter suddenly means different things with the "detailed CPU time" toggle. I would rather have "virtual" CPU time permanently represented by `%` and `&` and document such in the manual.

natoscott:
| It breaks forward compatibility - such as with the not-yet-merged [#714](https://github.com/htop-dev/htop/pull/714)

Feel free to queue your proposed change up in that PR then - if that's ever merged, that'd be an appropriate time to add this complexity.  In the meantime its not warranted.

Explorer09:
> | It breaks forward compatibility - such as with the not-yet-merged [#714](https://github.com/htop-dev/htop/pull/714)
> 
> Feel free to queue your proposed change up in that PR then - if that's ever merged, that'd be an appropriate time to add this complexity. In the meantime its not warranted.

I already proposed this PR for the exact problem. I don't need to file another one. And this PR effectively supercedes your PR (if I make the changes on top of your PR they would effectively revert your changes anyway).

natoscott:
OK, sounds good.

Explorer09:
> > | It breaks forward compatibility - such as with the not-yet-merged [#714](https://github.com/htop-dev/htop/pull/714)
> > Feel free to queue your proposed change up in that PR then - if that's ever merged, that'd be an appropriate time to add this complexity. In the meantime its not warranted.
> 
> I already proposed this PR for the exact problem. I don't need to file another one. And this PR effectively supercedes your PR (if I make the changes on top of your PR they would effectively revert your changes anyway).

Sigh... I've told that this PR (#1922) requires reverting #1921 in order to build on top of it. What's the silly management decision that decided to merge #1921 first?

Issue #1923, opened 2026-03-18
Create color scheme(s) for full color terminal emulations


fasterit:
Forward from Debian: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1131144

RGB color terminal emulations are detectable by their -direct suffix.
I.e.
TERM=xterm-direct
TERM=konsole-direct
TERM=st-direct
TERM=mlterm-direct
TERM=iterm2-direct (iTerm2, MacOS)

We could do color schemes that are more beautiful than the default ones and not error prone to "having the wrong background color" as with the 16 / 256 color TERMs.

PR #1926, opened 2026-03-25
Add true‑color‑aware color schemes for *-direct terminals #1923

patil-rushikesh:
- Detect true color support via *-direct TERM values and terminfo Tc/RGB capabilities.
- Add dark and light true‑color‑aware schemes and automatically map Default / Light / Black on White to them when true color is available, keeping existing behavior on 16/256‑color terminals.
- Expose the new schemes in the Colors panel so they can also be selected manually.

### DIFF
--- a/pcp-htop.c
+++ b/pcp-htop.c
@@ -23,7 +23,7 @@ int main(int argc, char** argv) {
    /*
     * set OPTION_NOVALUE to map undefined metric operands to novalue()
     */
-    pmSetDerivedControl(PCP_DERIVED_OPTION_NOVALUE, 1);
+    (void)pmSetDerivedControl(PCP_DERIVED_OPTION_NOVALUE, 1);
 #endif
 
    /* extract environment variables */


CRT.c
```diff
@@ -1130,6 +1378,55 @@ void CRT_setMouse(bool enabled) {
...
+   if (!str || !suffix) {
+      return false;
+   }
+   size_t strLen = strlen(str);
+   size_t suffixLen = strlen(suffix);
+   if (strLen < suffixLen) {
+      return false;
+   }
+   return strcmp(str + strLen - suffixLen, suffix) == 0;
+}
```
BenBE: Should go to `XUtils.h`. Cf. `strncmp` optimization for `String_startsWith`.

Could even use `memcmp` given we know the exact lengths for both strings.

CRT.c
```diff
@@ -1130,6 +1378,55 @@ void CRT_setMouse(bool enabled) {
...
+   // Check for -direct suffix: xterm-direct, konsole-direct, st-direct, etc.
+   // The "-direct" suffix is a standard indicator of true color support in
+   // modern terminal emulators.
+   if (stringEndsWith(termType, "-direct")) {
+      return true;
+   }
+
+   // Optional: Check terminfo capabilities for true color support
+   // The "Tc" and "RGB" capabilities are official terminfo indicators of true color
+   #ifdef HAVE_TIGETSTR
```
BenBE: IT seems like this is never checks for in `configure.ac`, thus never defined. Or where is this defined?

CRT.c
```diff
@@ -1130,6 +1135,55 @@ void CRT_setMouse(bool enabled) {
...
+   char* tcCap = tigetstr("Tc");
+   char* rgbCap = tigetstr("RGB");
+   
+   // tigetstr returns -1 if capability doesn't exist, 0 if canceled/absent,
+   // and a valid pointer if present
+   if ((tcCap && tcCap != (char*)(-1) && tcCap != (char*)(0)) ||
+       (rgbCap && rgbCap != (char*)(-1) && rgbCap != (char*)(0))) {
+      return true;
+   }
+   #endif
```
Explorer09: Firstly, you need to check for `tigetstr(3)` in `configure.ac` or else `HAVE_TIGETSTR` could be left undefined.

Second, I prefer the logic to look like this:

```c
   // The "Tc" and "RGB" capabilities are official terminfo indicators of true color
   // tigetstr(3) returns -1 if capability doesn't exist, 0 if canceled/absent,
   // and a valid pointer if present
   const char* capStr = tigetstr("Tc");
   if (capStr && capStr != (const char*)-1) {
      return true;
   }
   const char* capStr = tigetstr("RGB");
   if (capStr && capStr != (const char*)-1) {
      return true;
   }
```

CRT.c
```diff
@@ -1331,24 +1636,50 @@ void CRT_setColors(int colorScheme) {
...
 
    short int grayBlackFg = COLORS > 8 ? 8 : 0;
-   short int grayBlackBg = (colorScheme != COLORSCHEME_BLACKNIGHT) ? -1 : 0;
+   short int grayBlackBg = (CRT_colorScheme != COLORSCHEME_BLACKNIGHT) ? -1 : 0;
    init_pair(ColorIndexGrayBlack, grayBlackFg, grayBlackBg);
 
    init_pair(ColorIndexWhiteDefault, White, -1);
 
-   CRT_colors = CRT_colorSchemes[colorScheme];
+   CRT_colors = CRT_colorSchemes[CRT_colorScheme];
```
BenBE: Unrelated noise; could be better handled by updating `colorScheme` in the block above.

configure.ac
```diff
@@ -879,7 +879,7 @@ if test "$my_htop_platform" = "solaris"; then
    # Since ncurses macros use the ERR macro, we cannot use another name.
    AC_DEFINE([ERR], [(-1)], [Predefine ncurses macro.])
 fi
-AC_CHECK_FUNCS([set_escdelay])
+AC_CHECK_FUNCS([set_escdelay tigetstr])
 AC_CHECK_FUNCS([getmouse])
```
Explorer09: Style suggestion. Let's merge these multiple `curses` functions check into one call:

```m4
AC_CHECK_FUNCS([ \
   getmouse \
   set_escdelay \
   tigetstr \
])
```

CRT.c
```diff
@@ -1406,20 +1393,21 @@ static bool terminalSupportsTrueColor(const char* termType) {
    // Check for -direct suffix: xterm-direct, konsole-direct, st-direct, etc.
    // The "-direct" suffix is a standard indicator of true color support in
    // modern terminal emulators.
-   if (stringEndsWith(termType, "-direct")) {
+   if (String_endsWith(termType, "-direct")) {
```
Explorer09: Is it possible for the "-direct" token to appear in the middle of the terminal string and not the end? For example "xterm-direct-something"?

fasterit: Well it can but the color suffixes don't really make sense:
`$ find /usr/share/terminfo/ -iname "*direct*"`

patil-rushikesh: “-direct” is used as a suffix in ncurses’ truecolor entries (e.g. xterm-direct, konsole-direct, st-direct) and not in the middle of TERM names. So checking for it only at the end matches existing conventions; a name like “xterm-direct-something” would be a non-standard custom entry.

CRT.c
```diff
@@ -1393,24 +1636,43 @@ void CRT_setColors(int colorScheme) {
       colorScheme = COLORSCHEME_DEFAULT;
    }
 
-   CRT_colorScheme = colorScheme;
+   int effectiveScheme = colorScheme;
```
Explorer09: I don't like overriding the color scheme here. If the user's htop config specifies a non-true-color default scheme, it should be kept as non-true-color.

If you want to change the default color scheme to "auto detection" or something, the best place to add the code is in the `Settings_new` function. (Look for the `this->colorScheme = 0;` line and start working from there.)

fasterit:
I like the implementation and it technically works.

The color-schemes themselves need to be improved though, they basically only have small changes for the dynamic meters and I see true RGB values used nowhere.

In a `xterm -tn xterm-direct` therefore the greyed USER columns are invisible:
grafik

configure.ac
BenBE: Or have one per function each …

Explorer09:
> I like the implementation and it technically works.
> 
> The color-schemes themselves need to be improved though, they basically only have small changes for the dynamic meters and I see true RGB values used nowhere.
> 
> In a `xterm -tn xterm-direct` therefore the greyed USER columns are invisible: grafik

This screenshot doesn't do justice to the potential true color capability of a color scheme. Can someone test it on a better terminal emulator with a better color scheme design?

fasterit:
> This screenshot doesn't do justice to the potential true color capability of a color scheme. Can someone test it on a better terminal emulator with a better color scheme design?

Sure. But it still needs to work with `xterm-direct` esp. if the 24bit color schemes are auto-selected at runtime (which you objected to earlier).

patil-rushikesh:
> > I like the implementation and it technically works.
> > The color-schemes themselves need to be improved though, they basically only have small changes for the dynamic meters and I see true RGB values used nowhere.
> > In a `xterm -tn xterm-direct` therefore the greyed USER columns are invisible: grafik
> 
> This screenshot doesn't do justice to the potential true color capability of a color scheme. Can someone test it on a better terminal emulator with a better color scheme design?

Screenshot 2026-03-26 at 12 00 23 AM

BenBE:
Please rebase your changes onto main instead of merging. Also incorporate fixes into the commits where they belong. Find more details in the style guide.

CRT.c
```diff
@@ -947,10 +947,270 @@ static int CRT_colorSchemes[LAST_COLORSCHEME][LAST_COLORELEMENT] = {
...
+      [METER_VALUE_OK] = ColorPair(Green, Black),
+      [METER_VALUE_WARN] = A_BOLD | ColorPair(Yellow, Black),
+      [LED_COLOR] = ColorPair(Green, Black),
+      [TASKS_RUNNING] = A_BOLD | ColorPair(Green, Black),
+      [PROCESS] = A_NORMAL,
+      /*
+       * On true-color terminals like xterm-direct, the old gray-on-black
+       * combination for PROCESS_SHADOW made columns such as USER nearly
+       * invisible. Use a higher-contrast gray by reusing the default
+       * foreground while keeping the dark background.
```
Explorer09: I can't understand this comment. If everything you've done to make htop support true color is to display a brighter gray for dimmed out text, then it would be better in the UX perspective to not do it as a separate color scheme, but as an extra color option that can apply to _all_ color schemes that htop ever supports.

The "Broken Gray" color scheme in htop is, in my opinion, a bad precedent and a UX mistake.

CRT.c
```diff
@@ -1330,25 +1639,31 @@ void CRT_setColors(int colorScheme) {
...
+   short int grayBlackBg = (CRT_colorScheme != COLORSCHEME_BLACKNIGHT) ? -1 : 0;
    init_pair(ColorIndexGrayBlack, grayBlackFg, grayBlackBg);
 
    init_pair(ColorIndexWhiteDefault, White, -1);
 
-   CRT_colors = CRT_colorSchemes[colorScheme];
+   CRT_colors = CRT_colorSchemes[CRT_colorScheme];
+}
+
+bool CRT_trueColorSupported(void) {
```
Explorer09: Perhaps remove this accessor method and instead expose the variable as a public symbol?

BenBE: Keeping the variable local is fine.

Settings.c
```diff
@@ -896,6 +897,25 @@ Settings* Settings_new(const Machine* host, Hashtable* dynamicMeters, Hashtable*
...
+    * Auto-select a true-color-aware default scheme on *-direct terminals
+    * only when no explicit color_scheme was configured anywhere.
+    *
+    * This preserves existing behavior for users with a color_scheme in
+    * htoprc (including non-true-color defaults) while still opting into
+    * the new TRUECOLOR scheme by default on capable terminals.
+    */
+   if (!this->colorSchemeExplicit) {
+      const char* termType = getenv("TERM");
+      if (termType && String_endsWith(termType, "-direct")) {
```
Explorer09: Why not use `CRT_trueColorSupported`? It seems this part duplicates what you have been done in CRT.c

Explorer09:
@patil-rushikesh You can convert the pull request into [a draft](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests#draft-pull-requests) if you feel you need to work on it a little bit and don't want maintainers to merge right now. It's better than closing and reopening repeatedly.

Also, use the [git rebase](https://git-scm.com/docs/git-rebase) feature instead of merge to incorporate changes done in the `main` branch.

patil-rushikesh:
> @patil-rushikesh You can convert the pull request into [a draft](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests#draft-pull-requests) if you feel you need to work on it a little bit and don't want maintainers to merge right now. It's better than closing and reopening repeatedly.
> 
> Also, use the [git rebase](https://git-scm.com/docs/git-rebase) feature instead of merge to incorporate changes done in the `main` branch.

I will reopen the pull request as a draft and continue working on it.
Apologies for the inconvenience, and thank you for the guidance.

PR #1927, opened 2026-03-26
Fixes for PCP derived metric handling

kmcdonell:
Details are in the commit log messages, but briefly

- use (new) PCP feature to turn on automapping of undefined metrics to "no values available"
- fix multiple problems associated with '=' appearing the in the expression for a derived metric
- clean up some of the config files

### DIFF
--- a/XUtils.c
+++ b/XUtils.c
@@ -178,7 +178,7 @@ char** String_split(const char* s, char sep, size_t* n) {
 }
 
 /* same as String_split() but only split on first occurrence of sep */
-char** String_split_one(const char* s, char sep, size_t* n) {
+char** String_split_first(const char* s, char sep, size_t* n) {
    char** out = xCalloc(3, sizeof(char*));
    size_t ctr = 0;
    const char* where;
--- a/XUtils.h
+++ b/XUtils.h
@@ -81,7 +81,7 @@ ATTR_NONNULL_N(1) ATTR_RETNONNULL
 char** String_split(const char* s, char sep, size_t* n);
 
 ATTR_NONNULL_N(1) ATTR_RETNONNULL
-char** String_split_one(const char* s, char sep, size_t* n);
+char** String_split_first(const char* s, char sep, size_t* n);
 
 void String_freeArray(char** s);
 
--- a/pcp/PCPDynamicColumn.c
+++ b/pcp/PCPDynamicColumn.c
@@ -146,7 +146,7 @@ static void PCPDynamicColumn_parseFile(PCPDynamicColumns* columns, const char* p
       }
 
       size_t n;
-      char** config = String_split_one(trimmed, '=', &n);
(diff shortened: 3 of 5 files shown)


XUtils.c
```diff
@@ -177,6 +177,30 @@ char** String_split(const char* s, char sep, size_t* n) {
    return out;
 }
 
+/* same as String_split() but only split on first occurrence of sep */
+char** String_split_one(const char* s, char sep, size_t* n) {
```
natoscott: Bike shedding here, but String_split_first reads a little clearer to me.  Not a strong opinion either way though, pick whichever you prefer IMO.

kmcdonell: OK, done.

XUtils.c
```diff
@@ -177,6 +177,30 @@ char** String_split(const char* s, char sep, size_t* n) {
...
+      size_t size = (size_t)(where - s);
+      out[ctr] = xStrndup(s, size);
+      ctr++;
+      s += size + 1;
+   }
+   if (s[0] != '\0') {
+      out[ctr] = xStrdup(s);
+      ctr++;
+   }
+   out = xRealloc(out, sizeof(char*) * (ctr + 1));
```
Explorer09, line 195: This `xRealloc` call is unnecessary. The `out` buffer space is at most the size of three pointers (with the last pointer space, `out[2]`, always NULL), so it can be allocated all at the first `xCalloc` call.

XUtils.c
```diff
@@ -177,6 +177,30 @@ char** String_split(const char* s, char sep, size_t* n) {
    return out;
 }
 
+/* same as String_split() but only split on first occurrence of sep */
+char** String_split_first(const char* s, char sep, size_t* n) {
```
Explorer09, line 181: I think the htop style guide would prefer the function to be named `String_splitFirst`.

check-pcp-style.sh
```diff
@@ -65,11 +65,14 @@ check_file() {
...
-      /^[^=]+ = [^=]+$/ {
-         split($0, pair, " = ")
+      /^[^=]+ = .+$/ {
+         npart = split($0, pair, " = ")
 
          name = trim(pair[1])
          value = trim(pair[2])
+	 for (i = 3; i <= npart; i++) {
+	     value = value " " trim(pair[i])
+	 }
```
BenBE, line 75: Indentation seems off here …

XUtils.c
BenBE: Just leave the array at its initial size, as the amount of over-allocation here doesn't really matter.

An alternative would be doing the splitting first and create the proper allocation once you know the number of needed segments to store. In any case: Avoid repeated calls to the allocator.

natoscott:
I've incorporated all review feedback, rebased and merged (we're keen to get this into PCP ASAP for a release next week).

PR #1928, opened 2026-03-26
Extend events system, make mouse handling consistent

### LINKED ISSUE #1760: Multiple Screens not working

fasterit:
Fixes: #1760
Improves over: #1783

Extends the events with a lost focus one and makes the screens, meters and columns editing handle that properly.
Since we have that consistent then, I spliced out the second commit to also make the function bar consistent between the meters and the screens setup. The meters one was a bit out-of-the-usual. The shortcuts (space to change meter mode ("style"), enter to accept, cursor keys to move) all still work.

When playing around in Setup for hours the UI inconsistency ... hurts ... so I fixed that quite early but I want to give you all the bikeshedding options :smile:. Commit #1 builds fine separately.

### DIFF
--- a/ColumnsPanel.c
+++ b/ColumnsPanel.c
@@ -1,6 +1,7 @@
 /*
 htop - ColumnsPanel.c
 (C) 2004-2011 Hisham H. Muhammad
+(C) 2020-2026 htop dev team
 Released under the GNU GPLv2+, see the COPYING file
 in the source distribution for its full text.
 */
@@ -33,6 +34,17 @@ static void ColumnsPanel_delete(Object* object) {
    free(this);
 }
 
+static void ColumnsPanel_cancelMoving(ColumnsPanel* this) {
+   Panel* super = &this->super;
+   for (int i = 0; i < Panel_size(super); i++) {
+      ListItem* item = (ListItem*) Panel_get(super, i);
+      if (item)
+         item->moving = false;
+   }
+   this->moving = false;
+   Panel_setSelectionColor(super, PANEL_SELECTION_FOCUS);
+}
+
 static HandlerResult ColumnsPanel_eventHandler(Panel* super, int ch) {
    ColumnsPanel* const this = (ColumnsPanel*) super;
 
@@ -44,16 +56,27 @@ static HandlerResult ColumnsPanel_eventHandler(Panel* super, int ch) {
       case 0x0a:
       case 0x0d:
       case KEY_ENTER:
-      case KEY_MOUSE:
       case KEY_RECLICK:
          if (selected < size) {
-            this->moving = !(this->moving);
-            Panel_setSelectionColor(super, this->moving ? PANEL_SELECTION_FOLLOW : PANEL_SELECTION_FOCUS);
-            ListItem* selectedItem = (ListItem*) Panel_getSelected(super);
-            if (selectedItem)
-               selectedItem->moving = this->moving;
+            if (this->moving) {
+               ColumnsPanel_cancelMoving(this);
+            } else {
+               this->moving = true;
+               Panel_setSelectionColor(super, PANEL_SELECTION_FOLLOW);
+               ListItem* selectedItem = (ListItem*) Panel_getSelected(super);
+               if (selectedItem)
+                  selectedItem->moving = true;
+            }
             result = HANDLED;
          }
+         break;
+      case KEY_MOUSE:
+         if (this->moving) {
+            /* Single click while in move mode: cancel move mode */
+            ColumnsPanel_cancelMoving(this);
+            result = HANDLED;
+         }
+         /* else: just select the item, do not enter move mode */
          break;
       case KEY_UP:
          if (!this->moving)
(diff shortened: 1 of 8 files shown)


BenBE (approved):
LGTM.

ColumnsPanel.c
```diff
@@ -33,6 +34,17 @@ static void ColumnsPanel_delete(Object* object) {
    free(this);
 }
 
+static void ColumnsPanel_cancelMoving(ColumnsPanel* this) {
```
Explorer09, line 37: Does this function work only if `this->moving` is true?

I see a room to improve the quality of this code. Maybe I should file a PR for it. For example, moving the `this->moving` conditionals from the callers to here.

fasterit: not worth it imo. The "local" checks are unlikely to succeed and faster than unconditional function calls with all the stack magic and `call`/`ret`.

Explorer09: > not worth it imo. The "local" checks are unlikely to succeed and faster than unconditional function calls with all the stack magic and `call`/`ret`.

I consider it's the compiler's job to move the `this->moving` flag check to outside the function for performance. My reason for putting it in the function is to make it safer (think of when a future developer calls the function without checking the state first). Also it's less redundancy in the source code level.

fasterit: ... on the price of reduced readability. We both have good arguments but this is a matter of taste in the end.

PR #1931, opened 2026-03-27
Track oldFocus correctly in ScreensPanel

### LINKED ISSUE #1930: In Screens Setup, mouse click to select another screen doesn't update the "Active Columns" panel
This issue exists since the first commit that introduced screen tabs in htop (72ba20fa5f077bab4286dd78f9944df15de3b3b4).

1. Go to "Setup" -> "Screens"
2. Make sure you have at least two screens created. If not, use "(F5)New" to create the second screen right now.
3. Mouse click on a screen in the "Screens" panel that is not currently selected/highlighted.

Actual behavior: The "Active Columns" panel still lists the columns of the previous screen. It's not updated.

Expected behavior: The "Active Columns" panel gets updated to list the active columns of the newly selected screen.

https://github.com/user-attachments/assets/78350d32-9552-4e93-b77b-257e0fc77bd6

fasterit:
Fixes: #1930

Track the previously selected ScreenListItem_ within ScreensPanel_ and use that for oldFocus so the event handler does not always come too late to the party.

### DIFF
--- a/ScreensPanel.c
+++ b/ScreensPanel.c
@@ -224,7 +224,7 @@ static HandlerResult ScreensPanel_eventHandlerNormal(Panel* super, int ch) {
    ScreensPanel* const this = (ScreensPanel*) super;
 
    int selected = Panel_getSelectedIndex(super);
-   ScreenListItem* oldFocus = (ScreenListItem*) Panel_getSelected(super);
+   ScreenListItem* oldFocus = this->prevSelected;
    bool shouldRebuildArray = false;
    HandlerResult result = IGNORED;
 
@@ -339,6 +339,8 @@ static HandlerResult ScreensPanel_eventHandlerNormal(Panel* super, int ch) {
       result = HANDLED;
    }
 
+   this->prevSelected = newFocus;
+
    if (shouldRebuildArray)
       rebuildSettingsArray(super, selected);
    if (result == HANDLED)
@@ -385,6 +387,7 @@ ScreensPanel* ScreensPanel_new(Settings* settings) {
    this->renamingNewItem = false;
    super->cursorOn = false;
    this->cursor = 0;
+   this->prevSelected = NULL;
    Panel_setHeader(super, "Screens");
 
    for (unsigned int i = 0; i < settings->nScreens; i++) {
@@ -392,6 +395,7 @@ ScreensPanel* ScreensPanel_new(Settings* settings) {
       char* name = ss->heading;
       Panel_add(super, (Object*) ScreenListItem_new(name, ss));
    }
+   this->prevSelected = (ScreenListItem*) Panel_getSelected(super);
    return this;
 }
 
--- a/ScreensPanel.h
+++ b/ScreensPanel.h
@@ -36,6 +36,7 @@ typedef struct ScreensPanel_ {
    size_t cursor;
    ListItem* renamingItem;
    bool renamingNewItem;
+   struct ScreenListItem_* prevSelected;
 } ScreensPanel;
 
 typedef struct ScreenListItem_ {


BenBE:
Just a small note while looking at the patch:

Having one field live in the Panel class (active selection), while the other (previous selection) lives in the ScreensPanel class is kinda strange. Wouldn't it make sense to have this kind of tracking in the Panels class, so clicks that change the selection can be triggered through the overall panel logic?

fasterit:
Hm, do you think this functionality could be useful outside the ScreensPanel?

Moving the `prevSelected` is rather easy:

``` patch
diff --git a/Panel.c b/Panel.c
index d8734a51..2b876d0d 100644
--- a/Panel.c
+++ b/Panel.c
@@ -60,6 +60,7 @@ void Panel_init(Panel* this, int x, int y, int w, int h, const ObjectClass* type
    this->scrollH = 0;
    this->selected = 0;
    this->oldSelected = 0;
+   this->prevSelected = NULL;
    this->selectedLen = 0;
    this->needsRedraw = true;
    this->cursorOn = false;
diff --git a/Panel.h b/Panel.h
index ab163b0c..37cad2c9 100644
--- a/Panel.h
+++ b/Panel.h
@@ -68,6 +68,7 @@ struct Panel_ {
    Vector* items;
    int selected;
    int oldSelected;
+   Object* prevSelected;
    size_t selectedLen;
    void* eventHandlerState;
    int scrollV;
diff --git a/ScreensPanel.c b/ScreensPanel.c
index e8dee0cd..b9be8803 100644
--- a/ScreensPanel.c
+++ b/ScreensPanel.c
@@ -238,7 +238,7 @@ static HandlerResult ScreensPanel_eventHandlerNormal(Panel* super, int ch) {
    ScreensPanel* const this = (ScreensPanel*) super;
 
    int selected = Panel_getSelectedIndex(super);
-   ScreenListItem* oldFocus = this->prevSelected;
+   ScreenListItem* oldFocus = (ScreenListItem*) super->prevSelected;
    bool shouldRebuildArray = false;
    HandlerResult result = IGNORED;
 
@@ -353,7 +353,7 @@ static HandlerResult ScreensPanel_eventHandlerNormal(Panel* super, int ch) {
       result = HANDLED;
    }
 
-   this->prevSelected = newFocus;
+   super->prevSelected = (Object*) newFocus;
 
    if (shouldRebuildArray)
       rebuildSettingsArray(super, selected);
@@ -401,7 +401,6 @@ ScreensPanel* ScreensPanel_new(Settings* settings) {
    this->renamingNewItem = false;
    super->cursorOn = false;
    this->cursor = 0;
-   this->prevSelected = NULL;
    Panel_setHeader(super, "Screens");
 
@@ -409,7 +408,7 @@ ScreensPanel* ScreensPanel_new(Settings* settings) {
       char* name = ss->heading;
       Panel_add(super, (Object*) ScreenListItem_new(name, ss));
    }
-   this->prevSelected = (ScreenListItem*) Panel_getSelected(super);
+   super->prevSelected = Panel_getSelected(super);
    return this;
 }
 
diff --git a/ScreensPanel.h b/ScreensPanel.h
index d3ca0b03..5b889fa4 100644
--- a/ScreensPanel.h
+++ b/ScreensPanel.h
@@ -36,7 +36,6 @@ typedef struct ScreensPanel_ {
    size_t cursor;
    ListItem* renamingItem;
    bool renamingNewItem;
-   struct ScreenListItem_* prevSelected;
 } ScreensPanel;
 
 typedef struct ScreenListItem_ {
```

fasterit:
^ @BenBE, @natoscott: Made a draft PR #1938 so you can comment there

BenBE:
We have the per-screen display options, we have the screen columns selection, … Two places that come to mind where we have display screens that depend on panel context

Second aspect is semantic similarity; splitting these fields across classes just feels "wrong".

PR #1932, opened 2026-03-27
Only issue KEY_RECLICK when the focussed item has not changed

### LINKED ISSUE #1929: In Screens Setup, mouse click on another panel and then click on the current screen enters rename mode
After dca23ae85a5654e73d06473449e3344753630813, I discover some UI behavior in htop's Setup screen undesirable.

1. Enter "Setup" -> "Screens" in htop.
2. Mouse click on an item in "Active Columns" panel.
3. Mouse click on the current, selected screen in the "Screens" panel.

Expected behavior: The current screen item and the "Screens" panel gets focus, but the current screen shouldn't go to rename mode.
Actual behavior: The current screen item goes to rename mode.

https://github.com/user-attachments/assets/8ac47b56-30a2-4e93-b918-977ced7d45dd

fasterit:
Fixes: #1929


Explorer09:
I can confirm this fixes the problem I reported in #1929. I have no comments with the code because the patch looks trivial.

PR #1933, opened 2026-03-28
Implement line editor for Search, Filter and (re)naming Screens

fasterit:
* Line editor with the most common movement and delete key-shortcuts
* History for Search and Filter (shared and saved as `htop_history`)

The Ctrl-left / Ctrl-right handling is xterm-compatible terminals only for now. Can be extended to other terminals if people care.
The Shift-left / Shift-right are stock ncurses from the extended set, so that should work everywhere™ but nobody used these for movement any more, I guess.

### DIFF
--- a/CommandLine.c
+++ b/CommandLine.c
@@ -425,6 +425,20 @@ int CommandLine_run(int argc, char** argv) {
    if (flags.commFilter)
       setCommFilter(&state, &(flags.commFilter));
 
+   /* Set up shared search/filter history, stored next to the config file */
+   const char* rcPath = settings->filename;
+   const char* lastSlash = strrchr(rcPath, '/');
+   char historyPath[PATH_MAX];
+   if (lastSlash) {
+      int dirLen = (int)(lastSlash - rcPath + 1);
+      xSnprintf(historyPath, sizeof(historyPath), "%.*s" "htop_history", dirLen, rcPath);
+   } else {
+   /* no history file saved unless we have a sane rcPath */
+      historyPath[0] = '\0';
+   }
+
+   IncSet_setHistoryFile(panel->inc, historyPath);
+
    ScreenManager* scr = ScreenManager_new(header, host, &state, true);
    ScreenManager_add(scr, (Panel*) panel, -1);
 
--- a/History.c
+++ b/History.c
@@ -1,0 +1,152 @@
+/*
+htop - History.c
+(C) 2004-2026 htop dev team
+Released under the GNU GPLv2+, see the COPYING file
+in the source distribution for its full text.
+*/
+
+#include "config.h" // IWYU pragma: keep
+
+#include "History.h"
+
+#include <fcntl.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+#include "Macros.h"
+#include "XUtils.h"
+
+
+static void History_load(History* this) {
+   if (!this->filename)
+      return;
+   FILE* fp = fopen(this->filename, "r");
+   if (!fp)
+      return;
+
+   char line[LINEEDITOR_MAX + 2];
+   while (fgets(line, sizeof(line), fp)) {
+      size_t len = strlen(line);
+      /* strip trailing newline */
+      while (len > 0 && (line[len - 1] == '\n' || line[len - 1] == '\r'))
+         line[--len] = '\0';
+      if (len == 0)
+         continue;
+
+      History_add(this, line);
+   }
+   fclose(fp);
+}
+
+History* History_new(const char* filename) {
+   History* this = xCalloc(1, sizeof(History));
+   this->capacity = 64;
+   this->entries = xCalloc(this->capacity, sizeof(char*));
+   this->count = 0;
+   this->position = 0;
+   this->saved[0] = '\0';
+   this->filename = filename ? xStrdup(filename) : NULL;
+
+   if (this->filename)
+      History_load(this);
+
+   this->position = this->count;
+
+   return this;
+}
+
+void History_delete(History* this) {
+   for (size_t i = 0; i < this->count; i++)
+      free(this->entries[i]);
+   free(this->entries);
+   free(this->filename);
+   free(this);
+}
+
+void History_save(const History* this) {
+   if (!this->filename)
+      return;
+   /* Settings_write writes things via a temp file & rename, we do it less robust but faster here: */
+   int fd = open(this->filename, O_WRONLY | O_CREAT | O_TRUNC, 0600);
+   if (fd == -1)
+      return;
+   FILE* fp = fdopen(fd, "w");
+   if (!fp) {
+      close(fd); // fd not consumed on failure, so close it
+      return;
+   }
+   size_t start = (this->count > HISTORY_MAX_ENTRIES) ? this->count - HISTORY_MAX_ENTRIES : 0;
+   for (size_t i = start; i < this->count; i++)
+      fprintf(fp, "%s\n", this->entries[i]);
+   fclose(fp);
+}
+
+void History_add(History* this, const char* entry) {
+   if (!entry || entry[0] == '\0')
+      return;
+
+   /* Deduplicate: remove previous identical entry if present */
+   for (size_t i = 0; i < this->count; i++) {
+      if (String_eq(this->entries[i], entry)) {
+         free(this->entries[i]);
+         memmove(this->entries + i, this->entries + i + 1, (this->count - i - 1) * sizeof(char*));
+         this->count--;
+         break;
+      }
+   }
+
+   /* Grow array if needed */
+   if (this->count >= this->capacity) {
+      if (this->capacity < HISTORY_MAX_ENTRIES) {
+         this->capacity = MINIMUM(this->capacity * 2, (size_t)HISTORY_MAX_ENTRIES);
+         this->entries = xReallocArray(this->entries, this->capacity, sizeof(char*));
+      } else {
+         /* Drop oldest entry */
+         free(this->entries[0]);
+         memmove(this->entries, this->entries + 1, (this->count - 1) * sizeof(char*));
+         this->count--;
+      }
+   }
+
+   this->entries[this->count++] = xStrdup(entry);
+
+   /* Reset position to "at new input" */
+   this->position = this->count;
+   this->saved[0] = '\0';
+}
+
+const char* History_navigate(History* this, LineEditor* editor, bool back) {
+   if (this->count == 0)
+      return NULL;
+
+   if (back) {
+      /* Going back (up arrow) */
+      if (this->position == this->count) {
+         /* Save current editor content before entering history */
+         strncpy(this->saved, LineEditor_getText(editor), LINEEDITOR_MAX);
+         this->saved[LINEEDITOR_MAX] = '\0';
+      }
+      if (this->position > 0) {
+         this->position--;
+         return this->entries[this->position];
+      }
+      return NULL; /* Already at oldest entry */
+   } else {
+      /* Going forward (down arrow) */
+      if (this->position >= this->count)
+         return NULL; /* Already at newest */
+      this->position++;
+      if (this->position == this->count) {
+         /* Restore saved input */
+         return this->saved;
+      }
+      return this->entries[this->position];
+   }
+}
+
+void History_resetPosition(History* this) {
+   this->position = this->count;
+   this->saved[0] = '\0';
+}
--- a/History.h
+++ b/History.h
@@ -1,0 +1,48 @@
+#ifndef HEADER_History
+#define HEADER_History
+/*
+htop - History.h
+(C) 2004-2026 htop dev team
+Released under the GNU GPLv2+, see the COPYING file
+in the source distribution for its full text.
+*/
+
+#include <stdbool.h>
+#include <stddef.h>
+
+#include "LineEditor.h"
+
+
+#define HISTORY_MAX_ENTRIES 512
+
+typedef struct History_ {
+   char** entries;    /* array of history strings, oldest first */
+   size_t count;      /* current number of entries */
+   size_t capacity;   /* allocated capacity */
+   size_t position;      /* current browse position: count = "at new input" */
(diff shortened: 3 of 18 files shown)


History.c
```diff
@@ -0,0 +1,163 @@
...
+   fclose(fp);
+
+   /* Trim to max entries (keep newest) */
+   if (this->count > HISTORY_MAX_ENTRIES) {
+      size_t excess = this->count - HISTORY_MAX_ENTRIES;
+      for (size_t i = 0; i < excess; i++)
+         free(this->entries[i]);
+      memmove(this->entries, this->entries + excess, HISTORY_MAX_ENTRIES * sizeof(char*));
+      this->count = HISTORY_MAX_ENTRIES;
+   }
```
BenBE: This might drop newer entries if things are read as:

```
a
b
c
d
-
e
f
```

Should retain `c`, `d`, `e`, and `f`, but will retain `a`, `b`, `c`, and `d` (if `HISTORY_MAX_ENTRIES` was 4.

CommandLine.c
```diff
@@ -425,6 +425,20 @@ int CommandLine_run(int argc, char** argv) {
...
+   {
+      const char* rcPath = settings->filename;
+      const char* lastSlash = strrchr(rcPath, '/');
+      char historyPath[PATH_MAX];
+      if (lastSlash) {
+         int dirLen = (int)(lastSlash - rcPath + 1);
+         snprintf(historyPath, sizeof(historyPath), "%.*s%s", dirLen, rcPath, "htop_history");
+      } else {
+         snprintf(historyPath, sizeof(historyPath), "htop_history");
+      }
```
BenBE:
```suggestion
      if (lastSlash) {
         int dirLen = (int)(lastSlash - rcPath + 1);
         xSnprintf(historyPath, sizeof(historyPath), "%.*s" "htop_history", dirLen, rcPath);
      } else {
         xSnprintf(historyPath, sizeof(historyPath), "htop_history");
      }
```

History.h
```diff
@@ -0,0 +1,48 @@
...
+#include "LineEditor.h"
+
+
+#define HISTORY_MAX_ENTRIES 512
+
+typedef struct History_ {
+   char** entries;    /* array of history strings, oldest first */
+   size_t count;      /* current number of entries */
+   size_t capacity;   /* allocated capacity */
+   int position;      /* current browse position: count = "at new input" */
```
BenBE: Why not `size_t`?

History.c
```diff
@@ -0,0 +1,163 @@
...
+   if (!fp)
+      return;
+
+   char line[LINEEDITOR_MAX + 2];
+   while (fgets(line, sizeof(line), fp)) {
+      size_t len = strlen(line);
+      /* strip trailing newline */
+      while (len > 0 && (line[len - 1] == '\n' || line[len - 1] == '\r'))
+         line[--len] = '\0';
+      if (len == 0)
```
BenBE, line 35: Why not use `History_add` while loading? Would automagically dedup.

IncSet.c
```diff
@@ -233,11 +293,25 @@ const char* IncSet_getListItemValue(Panel* panel, int i) {
    return l ? l->value : "";
 }
 
-void IncSet_drawBar(const IncSet* this, int attr) {
+void IncSet_drawBar(IncSet* this, int attr) {
```
BenBE: Could this draw call be kept with `const IncSet*` somehow?

LineEditor.c
```diff
@@ -0,0 +1,243 @@
...
+#include "config.h" // IWYU pragma: keep
+
+#include "LineEditor.h"
+
+#include <ctype.h>
+#include <string.h>
+
+#include "ProvideCurses.h"
+#include "CRT.h"
+#include "Panel.h"
```
BenBE:
```suggestion
#include "CRT.h"
#include "Panel.h"
#include "ProvideCurses.h"
```

LineEditor.c
```diff
@@ -0,0 +1,243 @@
...
+   this->scroll = 0;
+}
+
+void LineEditor_setText(LineEditor* this, const char* text) {
+   if (!text)
+      text = "";
+   size_t copyLen = this->maxLen;
+   strncpy(this->buffer, text, copyLen);
+   this->buffer[copyLen] = '\0';
+   this->len = strlen(this->buffer);
```
BenBE:
```suggestion
   this->len = strnlen(this->buffer, this->maxLen);
```

BenBE (changes requested):
Mind to split the Line Editor and History changes into two commits?

fasterit:
> Mind to split the Line Editor and History changes into two commits?

Yes, too much work. Line editor never existed without history.

History.c
fasterit: Shit. Testing confirms you are right :exploding_head:. Hm, I'll use your suggestion to re-use `History_add`. Thanky!

CommandLine.c
fasterit: oki, makes sense

History.h
fasterit: muscle memory, but yes, you are right. Will fix.

IncSet.c
fasterit: That'll be a bit involved, need to get rid of fieldStartX. I'll give it a try.

LineEditor.c
fasterit: thanks, forget to sort ... as usually

LineEditor.c
fasterit: ack

History.c
```diff
@@ -0,0 +1,163 @@
...
+      free(this->entries[i]);
+   free(this->entries);
+   free(this->filename);
+   free(this);
+}
+
+void History_save(const History* this) {
+   if (!this->filename)
+      return;
+   FILE* fp = fopen(this->filename, "w");
```
Explorer09: I feel uncomfortable when you create a history file without setting a umask first. The umask here should be `(S_IXUSR | S_IRWXG | S_IRWXO)`. In other words, no other user should be able to read the history when it's created, because, it's privacy.

History.h
```diff
@@ -0,0 +1,48 @@
...
+   char saved[LINEEDITOR_MAX + 1]; /* saved current input while browsing */
+   char* filename;    /* path to history file (may be NULL) */
```
Explorer09: How about documenting what would happen if this `filename` field is NULL then?

History.c
```diff
@@ -0,0 +1,163 @@
...
+   fclose(fp);
+}
+
+void History_add(History* this, const char* entry) {
+   if (!entry || entry[0] == '\0')
+      return;
+
+   /* Deduplicate: remove previous identical entry if present */
+   for (size_t i = 0; i < this->count; i++) {
+      if (strcmp(this->entries[i], entry) == 0) {
```
Explorer09: `String_eq`. (Looks like you let an AI generate this without a review.)

History.c
```diff
@@ -0,0 +1,163 @@
...
+   /* Grow array if needed */
+   if (this->count >= this->capacity) {
+      if (this->capacity < HISTORY_MAX_ENTRIES) {
+         this->capacity = MINIMUM(this->capacity * 2, (size_t)HISTORY_MAX_ENTRIES);
+         this->entries = xReallocArray(this->entries, this->capacity, sizeof(char*));
+      } else {
+         /* Drop oldest entry */
+         free(this->entries[0]);
+         memmove(this->entries, this->entries + 1, (this->count - 1) * sizeof(char*));
+         this->count--;
```
Explorer09, line 109: It look like this part of code duplicated the functionality of the `Vector` module that htop has already. I may be wrong, but this part is similar to the `Vector_take` function.

History.c
```diff
@@ -0,0 +1,163 @@
...
+   }
+
+   this->entries[this->count++] = xStrdup(entry);
+
+   /* Reset position to "at new input" */
+   this->position = (int)this->count;
+   this->saved[0] = '\0';
+}
+
+const char* History_navigate(History* this, LineEditor* editor, bool back) {
```
Explorer09, line 120: Suggest replacing the `back` boolean parameter with an integer `offset`. `offset` can be -1 or +1. The code would look more intuitive with it.

History.c
fasterit: Hm, that's an interesting view. `~-/config/` is 0700 (unless changed by the user). Why would we need to protect a list of search strings beyond that?

History.h
fasterit: It would gracefully not be read / written (but CommandLine.c will setup the filename so it won't be empty). I'll document it.

History.c
fasterit: I think AI would remember the random stuff in XUtils better than I do :smile: .
Fixed.

History.c
fasterit: Replaced after the review from @BenBE with `History_add`.

History.c
fasterit: And then we do a "which direction does your mouse wheel scroll" setting? :rofl: 
I think the bool is just as fine as an int.

History.c
Explorer09: First, the config directory might be located in an unexpected location (due to `XDG_CONFIG_HOME` being configurable), thus you can't expect it always has the mode 0700.

Second, the search strings can suggest what kind of apps that a user typically runs on their system. And this information is not supposed to be public. (Imagine you tell other people what queries you search in Google every day...)

History.c
Explorer09: > And then we do a "which direction does your mouse wheel scroll" setting? 🤣 I think the bool is just as fine as an int.

I didn't get what you say. I mean there's a possibility of skipping multiple items when scrolling, and the data structure permits it. So passing an integer offset wouldn't hurt.

fasterit:
Added a separate commit for implementing the review feedback, pls. squash on merge

History.c
fasterit: I made it 0600 as we do that with `htoprc`  as well. Consistency++

History.c
BenBE: And I didn't even run that code … ;-)

fasterit: /me bows deeply

CommandLine.c
```diff
@@ -426,19 +426,19 @@ int CommandLine_run(int argc, char** argv) {
...
-      const char* lastSlash = strrchr(rcPath, '/');
-      char historyPath[PATH_MAX];
-      if (lastSlash) {
-         int dirLen = (int)(lastSlash - rcPath + 1);
-         snprintf(historyPath, sizeof(historyPath), "%.*s%s", dirLen, rcPath, "htop_history");
-      } else {
-         snprintf(historyPath, sizeof(historyPath), "htop_history");
-      }
-      IncSet_setHistoryFile(panel->inc, historyPath);
+   const char* rcPath = settings->filename;
```
Explorer09, line 429: Observe these lines in the `Settings_new` function:

```c
   this->filename = xMalloc(PATH_MAX);
   if (!realpath(this->initialFilename, this->filename))
      free_and_xStrdup(&this->filename, this->initialFilename);
```

So a call like `dirname(settings->filename)` might not give you the htop directory you wanted, if the htoprc file is a symlink.

I was thinking if it's necessary to add the `htopConfigDir` field in the Settings object. It might be better in the long run if there's a central place to grab the `~/.config/htop` directory path.

fasterit: If people link to a central `htoprc` that does not mean they want a central history, does it?

As of now symlinking `htop_history` would not work and we'd have all kinds of issues (like the .bash_history being overwritten by terminal sessions and not combined, same thing here).

fasterit:
^ typo fix in commit message

History.c
BenBE: The `String_eq` in XUtils is mostly for convenience to avoid the negation you'd have to do with `strcmp`, both giving cleaner types (bool instead of int) and much more expressive when read as "English text". And compilers inline the referenced code anyway, so there's not even any overhead at runtime. :)

History.c
BenBE: I think that was sarcasm/joking. And no, while having an offset (negative for backwards in time, positive for the future) or an enum (with explicit values for backwards/forwards) would semantically make more sense, a bool is sufficient here; and easy enough to change later if we decided to do so …

History.c
Explorer09: @BenBE Note that I reviewed this for readability. A contributor with familiarity with htop codebase shouldn't have made this mistake (and it has been in the style guide). So the chance of this slipped through is someone not reviewing the AI code thoroughly (not a good sign for open source software's quality).

natoscott: @Explorer09 Nonsense.  People work across many projects, and have many years of experience using libc APIs.  Its easy for this to happen and for experienced htop devs too, whether AI is involved or not.

Explorer09: @natoscott I wouldn't make that accusation simply because of a wrong style in one place. But it's already part of htop's policy to disclose the use of AI, and (sigh…) it's the submitter's job to correct any sloppy code that AI could generate, isn't it?

Don't tell me you have never seen AI slop code.

BenBE: Let's stay on-topic. And htop's AI policy ain't it rn …

fasterit:
Squashed and rebased onto current main

BenBE:
There's a minor glitch with the current PR. Nothing too wild, but might be confusing to users:

Steps to reproduce:
1. Start `htop`
2. Hit <kbd>/</kbd> to search
3. Type `htop`
4. Press <kbd>Ctrl</kbd>+<lbd>U</kbd> to clear the editor
5. Notice the highlighting is not removed (empty search)
6. Hit <kbd>Esc</kbd> to cancel search
7. Notice the highlighting is not removed (state before the search was entered)
8. Move your cursor
9. The highlight is removed (as expected)
10. Hit </kbd>/</kdb> to enter search again
11. Just hit <kbd>Esc</kbd> to cancel
12. The highlighting re-appears

Nothing big, but nonetheless worth looking into.

fasterit:
> There's a minor glitch with the current PR. Nothing too wild, but might be confusing to users:

Great find, thank you @BenBE. Fixed in the last push.

fasterit:
Should be RTM™ now :hand_with_index_finger_and_thumb_crossed:

PR #1935, opened 2026-03-29
Add digits editing for numeric options

fasterit:
Also add NumPad +/- keys for increment / decrement in numeric options

### DIFF
--- a/CRT.h
+++ b/CRT.h
@@ -181,6 +181,13 @@ void CRT_handleSIGSEGV(int signal) ATTR_NORETURN;
 #define KEY_FOCUS_IN   (KEY_MAX + 'I')
 #define KEY_FOCUS_OUT  (KEY_MAX + 'O')
 #define KEY_DEL_MAC    127
+#ifndef PADPLUS
+# define KEY_PADPLUS    583
+# define KEY_PADMINUS   588
+#else
+# define KEY_PADPLUS    PADPLUS
+# define KEY_PADMINUS   PADMINUS
+#endif
 
 extern char CRT_degreeSign[];
 
--- a/DisplayOptionsPanel.c
+++ b/DisplayOptionsPanel.c
@@ -85,7 +85,7 @@ static HandlerResult DisplayOptionsPanel_eventHandler(Panel* super, int ch) {
             break;
          }
          /* When not editing, fall through to space/toggle */
-         /* FALLTHROUGH */
+         /* fallthrough */
       case ' ':
          if (numItem && numItem->editing) {
             /* Space while editing: apply pending edit, then toggle */
@@ -94,19 +94,14 @@ static HandlerResult DisplayOptionsPanel_eventHandler(Panel* super, int ch) {
             }
             super->cursorOn = false;
          }
-         switch (OptionItem_kind(selected)) {
-            case OPTION_ITEM_TEXT:
-               break;
-            case OPTION_ITEM_CHECK:
-               CheckItem_toggle((CheckItem*)selected);
-               result = HANDLED;
-               settingsChanged = true;
-               break;
-            case OPTION_ITEM_NUMBER:
-               NumberItem_toggle((NumberItem*)selected);
-               result = HANDLED;
-               settingsChanged = true;
-               break;
+         if (OptionItem_kind(selected) == OPTION_ITEM_NUMBER) {
+            NumberItem_toggle((NumberItem*)selected);
+            result = HANDLED;
+            settingsChanged = true;
+         } else if (OptionItem_kind(selected) == OPTION_ITEM_CHECK) {
+            CheckItem_toggle((CheckItem*)selected);
+            result = HANDLED;
+            settingsChanged = true;
          }
          break;
       case '-':
--- a/DisplayOptionsPanel.c
+++ b/DisplayOptionsPanel.c
@@ -37,58 +37,134 @@ static HandlerResult DisplayOptionsPanel_eventHandler(Panel* super, int ch) {
    DisplayOptionsPanel* this = (DisplayOptionsPanel*) super;
 
    HandlerResult result = IGNORED;
+   bool settingsChanged = false;
    OptionItem* selected = (OptionItem*) Panel_getSelected(super);
 
    if (!selected) {
       return result;
    }
 
+   NumberItem* numItem = (OptionItem_kind(selected) == OPTION_ITEM_NUMBER) ? (NumberItem*)selected : NULL;
+
+   /* Helper: position the hardware cursor right after the edit buffer.
+    * +1 on Y for the panel header row; +1 on X for the leading '[' bracket. */
+   #define SET_EDIT_CURSOR() do { \
+      super->cursorY = super->y + 1 + (super->selected - super->scrollV); \
+      super->cursorX = super->x + 1 + numItem->editLen; \
+      super->cursorOn = true; \
+   } while (0)
+
    switch (ch) {
+      case 27: /* Escape: cancel editing */
+         if (numItem && numItem->editing) {
+            NumberItem_cancelEditing(numItem);
+            super->cursorOn = false;
+            return HANDLED;
+         }
+         break;
+      case KEY_BACKSPACE:
+      case 127:
+         if (numItem && numItem->editing) {
+            NumberItem_deleteChar(numItem);
+            SET_EDIT_CURSOR();
+            return HANDLED;
+         } else if (numItem) {
+            NumberItem_startEditingFromValue(numItem);
+            NumberItem_deleteChar(numItem);
+            SET_EDIT_CURSOR();
+            return HANDLED;
+         }
+         break;
       case '\n':
       case '\r':
       case KEY_ENTER:
+         if (numItem && numItem->editing) {
+            if (NumberItem_applyEditing(numItem)) {
+               settingsChanged = true;
+            }
+            super->cursorOn = false;
+            result = HANDLED;
+            break;
+         }
+         /* When not editing, fall through to space/toggle */
+         /* FALLTHROUGH */
       case ' ':
+         if (numItem && numItem->editing) {
+            /* Space while editing: apply pending edit, then toggle */
+            if (NumberItem_applyEditing(numItem)) {
+               settingsChanged = true;
+            }
+            super->cursorOn = false;
+         }
          switch (OptionItem_kind(selected)) {
             case OPTION_ITEM_TEXT:
                break;
             case OPTION_ITEM_CHECK:
                CheckItem_toggle((CheckItem*)selected);
                result = HANDLED;
+               settingsChanged = true;
                break;
             case OPTION_ITEM_NUMBER:
                NumberItem_toggle((NumberItem*)selected);
                result = HANDLED;
+               settingsChanged = true;
                break;
          }
          break;
       case '-':
+      case KEY_PADMINUS:
       case KEY_F(7):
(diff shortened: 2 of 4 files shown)


OptionItem.h
```diff
@@ -76,5 +82,11 @@ int NumberItem_get(const NumberItem* this);
 void NumberItem_decrease(NumberItem* this);
 void NumberItem_increase(NumberItem* this);
 void NumberItem_toggle(NumberItem* this);
+void NumberItem_startEditing(NumberItem* this);
+void NumberItem_startEditingFromValue(NumberItem* this);
+void NumberItem_cancelEditing(NumberItem* this);
+bool NumberItem_applyEditing(NumberItem* this);
+bool NumberItem_addChar(NumberItem* this, char c);
+void NumberItem_deleteChar(NumberItem* this);
```
Explorer09, line 89: So many functions being added. I'm becoming skeptical of what the AI is changing here.

OptionItem.c
```diff
@@ -215,3 +229,97 @@ void NumberItem_toggle(NumberItem* this) {
...
+   int newValue;
+   char* endptr;
+   if (this->scale < 0) {
+      double displayValue = strtod(this->editBuffer, &endptr);
+      if (endptr == this->editBuffer) {
+         this->editLen = 0;
+         this->editBuffer[0] = '\0';
+         return false;
+      }
+      newValue = (int) round(displayValue / pow(10, this->scale));
```
Explorer09, line 278: Since `this->scale` is negative in this conditional, why not use multiplication (`* pow(10, -this->scale)`) for the scale instead of division? And what's the rationale for the rounding direction?

OptionItem.c
```diff
@@ -215,3 +229,97 @@ void NumberItem_toggle(NumberItem* this) {
...
+   } else {
+      newValue = (int) strtol(this->editBuffer, &endptr, 10);
```
Explorer09, line 280: Better move the `char* endptr;` declaration to inside the `if` block and `else` block.

OptionItem.c
```diff
@@ -215,3 +229,97 @@ void NumberItem_toggle(NumberItem* this) {
...
+}
+
+void NumberItem_startEditingFromValue(NumberItem* this) {
+   this->savedValue = NumberItem_get(this);
+   this->editing = true;
+   char tmp[NUMBERITEM_EDIT_MAX + 1];
+   int len;
+   /* Format the initial edit buffer using scale for decimal display */
+   if (this->scale < 0) {
+      len = xSnprintf(tmp, sizeof(tmp), "%.*f", -this->scale, pow(10, this->scale) * this->savedValue);
```
Explorer09, line 247: Potential rounding error. It looks like AI didn't care about this.

OptionItem.c
```diff
@@ -55,14 +56,19 @@ static void NumberItem_display(const Object* cast, RichString* out) {
...
       written = xSnprintf(buffer, sizeof(buffer), "%.*f", -this->scale, pow(10, this->scale) * NumberItem_get(this));
+      RichString_appendnAscii(out, CRT_colors[CHECK_MARK], buffer, written);
    } else if (this->scale > 0) {
       written = xSnprintf(buffer, sizeof(buffer), "%d", (int) (pow(10, this->scale) * NumberItem_get(this)));
+      RichString_appendnAscii(out, CRT_colors[CHECK_MARK], buffer, written);
    } else {
       written = xSnprintf(buffer, sizeof(buffer), "%d", NumberItem_get(this));
+      RichString_appendnAscii(out, CRT_colors[CHECK_MARK], buffer, written);
    }
-   RichString_appendnAscii(out, CRT_colors[CHECK_MARK], buffer, written);
```
Explorer09, line 65: This should be done a separate commit. I can't tell why we should move the function calls.

DisplayOptionsPanel.c
```diff
@@ -37,58 +37,134 @@ static HandlerResult DisplayOptionsPanel_eventHandler(Panel* super, int ch) {
...
+         if (numItem && numItem->editing) {
+            NumberItem_deleteChar(numItem);
+            SET_EDIT_CURSOR();
+            return HANDLED;
+         } else if (numItem) {
+            NumberItem_startEditingFromValue(numItem);
+            NumberItem_deleteChar(numItem);
+            SET_EDIT_CURSOR();
+            return HANDLED;
+         }
```
Explorer09: Redundancy

```suggestion
         if (numItem) {
            if (!numItem->editing) {
               NumberItem_startEditingFromValue(numItem);
            }
            NumberItem_deleteChar(numItem);
            SET_EDIT_CURSOR();
            return HANDLED;
         }
```

DisplayOptionsPanel.c
```diff
@@ -37,58 +37,134 @@ static HandlerResult DisplayOptionsPanel_eventHandler(Panel* super, int ch) {
...
    switch (ch) {
+      case 27: /* Escape: cancel editing */
+         if (numItem && numItem->editing) {
+            NumberItem_cancelEditing(numItem);
+            super->cursorOn = false;
+            return HANDLED;
+         }
+         break;
+      case KEY_BACKSPACE:
+      case 127:
```
Explorer09: One less magic number to deal with

```suggestion
      case KEY_DEL_MAC:
```

CRT.h
```diff
@@ -181,6 +181,13 @@ void CRT_handleSIGSEGV(int signal) ATTR_NORETURN;
 #define KEY_FOCUS_IN   (KEY_MAX + 'I')
 #define KEY_FOCUS_OUT  (KEY_MAX + 'O')
 #define KEY_DEL_MAC    127
+#ifndef PADPLUS
+# define KEY_PADPLUS    583
+# define KEY_PADMINUS   588
```
Explorer09, line 186: Why these numbers and where are they from?

OptionItem.c
fasterit: len is max ten chars, where do you see an issue here?

OptionItem.h
fasterit: Waldorf and Statler, are you doing the code reviews here?

OptionItem.c
fasterit: That scale nonsense isn't of my making, cf. 267014cb.

On the div: Isn't your mantra to leave such tiny optimization to the compiler?

`round` gets the nearest integer, seems reasonable, not? I did not even think of alternatives. Just need to clamp the user, that types in 1.52334 for a x.x field.

OptionItem.c
fasterit: And declare it two times that way? That seems wrong?

OptionItem.c
fasterit: I need to introduce another variable if I want to get the
`RichString_appendnAscii(out, CRT_colors[CHECK_MARK], buffer, written);` out of the nested if-construct.
It's again a matter of taste, both would work for me. I wrote the one above.

CRT.h
fasterit: Because that is what xterm compatible terminals send when you press + / - on the Numpad without Numlock active. If you have some Mac values for that, happy to add.

DisplayOptionsPanel.c
```diff
@@ -97,25 +171,66 @@ static HandlerResult DisplayOptionsPanel_eventHandler(Panel* super, int ch) {
...
             OptionItem* previous = selected;
             Panel_onKey(super, ch);
             selected = (OptionItem*) Panel_getSelected(super);
             if (previous != selected) {
                result = HANDLED;
+               settingsChanged = true;
             }
          }
-         /* fallthrough */
+         /* FALLTHROUGH */
```
BenBE, line 184: Unrelated change. I think the rest of the code all uses lowercase for this.

DisplayOptionsPanel.c
```diff
@@ -37,58 +37,132 @@ static HandlerResult DisplayOptionsPanel_eventHandler(Panel* super, int ch) {
...
                CheckItem_toggle((CheckItem*)selected);
                result = HANDLED;
+               settingsChanged = true;
                break;
             case OPTION_ITEM_NUMBER:
                NumberItem_toggle((NumberItem*)selected);
                result = HANDLED;
+               settingsChanged = true;
                break;
          }
```
BenBE: The remaining branches of this function use if/elseif chains. Probably good to settle on one style for all.

OptionItem.c
BenBE: Note that the rounding mode may be surprising depending on what is (implicitly) set.

OptionItem.h
```diff
@@ -11,6 +11,8 @@ in the source distribution for its full text.
 
 #include "Object.h"
 
+#define NUMBERITEM_EDIT_MAX 10
+
```
BenBE, line 15:
```suggestion

#define NUMBERITEM_EDIT_MAX 10
```

BenBE (commented):
Minor stylistic notes …

OptionItem.c
Explorer09: @BenBE The FP rounding mode is per-thread, and is not carried over to another executable image loaded by [execve(2)](https://man7.org/linux/man-pages/man2/execve.2.html), so in other words, the rounding mode is in our control. However, I noticed a [round(3)](https://man7.org/linux/man-pages/man3/round.3.html) call here and the reason for the rounding direction is not given.

The context seems to be that, a binary floating point is used to parse a string with decimals, and then it tries to convert the floating point to a fixed point. This raises a question: why not parse to a fixed point directly?

DisplayOptionsPanel.c
fasterit: I think they should all be uppercase but fine to revert this one and may be we can change them all in one go if you lot agree.

OptionItem.c
fasterit: 1.56 gets rounded to 1.6 in the "Update Interval" I consider that most correct. And it was the most easy solution, too.

BenBE: I though more on the lines of 1.45 vs 1.55, which depending on rounding mode may both be 1.5 (even/odd rounding vs. mathematically correct rounding) instead of 1.5 vs 1.6.

DisplayOptionsPanel.c
fasterit: Yes, makes aesthetic sense. Changed.

fasterit:
Please squash on merge

DisplayOptionsPanel.c
BenBE: According to [GCC documentation](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wimplicit-fallthrough) the match is made case-insensitive, thus both are fine.

Current distribution on main is:
```console
$ grep -Ri 'fallthr'
DisplayOptionsPanel.c:         /* fallthrough */
OpenFilesScreen.c:         } /* FALLTHRU */
MetersPanel.c:         /* else fallthrough */
MetersPanel.c:         /* else fallthrough */
ScreensPanel.c:         /* FALLTHRU */
ScreensPanel.c:         /* FALLTHRU */
ColumnsPanel.c:         /* else fallthrough */
ColumnsPanel.c:         /* else fallthrough */
```

`-Wimplicit-fallthrough=3` is basically an extension of `-Wimplicit-fallthrough=2` with some stricter case matching; but both variants are covered.

fasterit: To me it is more about the visual clue. I spot FALLTHRU much better than fallthrough.

BenBE (approved):
LGTM. Works as expected.

PR #1937, opened 2026-03-30
README: update instructions for Arch

eworm-de:
We have a meta package `base-devel` that defines the basic tools to build Arch Linux packages. Just install that and its dependencies, and `ncurses`.

Also add `--needed` to skip packages that are installed already.


eworm-de:
This is similar to `build-essential` for Debian. Guess it makes sense this way.

PR #1938, opened 2026-03-31
Move prevSelected from ScreensPanel to Panel

fasterit:
cf. comment from @BenBE at https://github.com/htop-dev/htop/pull/1931#issuecomment-4158609986

### DIFF
--- a/Panel.c
+++ b/Panel.c
@@ -60,6 +60,7 @@ void Panel_init(Panel* this, int x, int y, int w, int h, const ObjectClass* type
    this->scrollH = 0;
    this->selected = 0;
    this->oldSelected = 0;
+   this->prevSelected = NULL;
    this->selectedLen = 0;
    this->needsRedraw = true;
    this->cursorOn = false;
--- a/Panel.c
+++ b/Panel.c
@@ -60,7 +60,7 @@ void Panel_init(Panel* this, int x, int y, int w, int h, const ObjectClass* type
    this->scrollH = 0;
    this->selected = 0;
    this->oldSelected = 0;
-   this->prevSelected = NULL;
+   this->prevSelected = -1;
    this->selectedLen = 0;
    this->needsRedraw = true;
    this->cursorOn = false;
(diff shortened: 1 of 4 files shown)


Panel.h
```diff
@@ -68,6 +68,7 @@ struct Panel_ {
    Vector* items;
    int selected;
    int oldSelected;
+   Object* prevSelected;
```
BenBE: Two questions:

1. What's the usage difference between `oldSelected` and `prevSelected`?
2. Why not store `prevSelected` as an `int` too?

Panel.c
BenBE, line 215: Could `prevSelected`/`oldSelected` be updated here to avoid duplication across the code?

BenBE (commented):
Let's use this PR for some cleanup. Looks like we have some duplicated logic here.

Also, having `prevSelected` tracked by external code risks UAF/dangling ptr issues.

Panel.c
fasterit: I tried, did not come up with a good solution.

Panel.h
fasterit: Because the int can be recycled when items are deleted in between

natoscott (approved):
@fasterit agreed, this is a more logical home & clearer to me.

Panel.h
Explorer09: > Because the int can be recycled when items are deleted in between

How about this: you set `prevSelected` to a value of -1 within the routine of deleting the item. The problem of using a pointer for `prevSelected` is you open a risk for a use-after-free. Note that you didn't avoid the use-after-free problem whether you use an integer or a pointer for `prevSelected`, so it's best to fix the problem properly (by removing the references after deleting the objects).

fasterit:
So, we either do it as proposed or I can keep the whole shebang in ScreensPanel only and even simplify that a bit more. @BenBE [pointed out other potential use cases](https://github.com/htop-dev/htop/pull/1931#issuecomment-4161074303) which leans towards merging the Panel solution. I'm fine with both.

Panel.c
BenBE: This will take a bit for me to look into; fine either way, thus no blocker.

BenBE:
Will probably take til after Easter for having a closer look at this, thus if this is considered for merging ASAP, go ahead.

PR #1457, opened 2024-04-16
Add HeaderLayout format for one single 100%-width column

peterdey:
On small screens (e.g. 320x240 TFT screens), there aren't enough columns on the tty to display a two-column header in a useful format.

A one-column header, with 100% width, is useful in this situation.
htop-onecol-20240416_210048

PR #1939, opened 2026-03-31
NetworkIOMeter: shorten the label

eworm-de:
This brings it in line with other meters, for example the DiskIOMeter which comes with just "Dsk:". Also it allows for more information on limited width.

### DIFF
--- a/NetworkIOMeter.c
+++ b/NetworkIOMeter.c
@@ -179,5 +179,5 @@ const MeterClass NetworkIOMeter_class = {
    .name = "NetworkIO",
    .uiName = "Network IO",
    .description = "Network bytes & packets received/sent per second",
-   .caption = "Network: "
+   .caption = "Net: "
 };


eworm-de:
I think this makes sense for meters where one may want to have text and graphical variant next to each other.

fasterit:
I think I would argue the other way around: We have `Network`, `Tasks`, `Systemd` etc. but `Dsk`. Spending an i for `Disk` would make it more consistent. We are not bound to 80x25 character screens any more for most users.
/DLange

fasterit:
We do have `Swp` though. Apparently somebody tried to make all the "typically bar-style" meters three characters.

eworm-de:
> I think I would argue the other way around: We have `Network`, `Tasks`, `Systemd` etc. but `Dsk`.

But I do not have graphical meters for `Tasks` & `Systemd`. 😉

> Spending an i for `Disk` would make it more consistent. We are not bound to 80x25 character screens any more for most users. /DLange

Most of the time I have `tmux` running, with `htop` in a pane with 50% screen width. With three columns inside `htop` the with is quite limited and could benefit from saving some chars.

But after all I am open to both ways. Having things more or less consistent would be great, though.

Explorer09:
> > I think I would argue the other way around: We have `Network`, `Tasks`, `Systemd` etc. but `Dsk`.
> 
> But I do not have graphical meters for `Tasks` & `Systemd`. 😉
> 
> > Spending an i for `Disk` would make it more consistent. We are not bound to 80x25 character screens any more for most users. /DLange
> 
> Most of the time I have `tmux` running, with `htop` in a pane with 50% screen width. With three columns inside `htop` the with is quite limited and could benefit from saving some chars.
> 

Just a reminder that I take a neutral stance on this change. Both "Network:" and "Net" labels work fine with me.

And htop should cater for 80x25 terminal screens anyway. There is a use case for even smaller screen (like this #1457). Thus the request for shortening the label isn't unreasonable.

By the way, the reason I proposed the "Dsk" label before is that in the bar mode, the label truncated from "Disk" to "Dis" wasn't intuitive to read. ("Dis" can mean many things: distribution, display, displacement, etc.) The shortening to "Dsk" makes it less ambiguous.

PR #1940, opened 2026-04-05
Normalize physical core ID when `showCPUSMTLabels` is set

tjk213:
### Summary                                                                                                                                                                           
                                                                                                                                                                                      
This PR normalizes the physical core IDs used when the `showCPUSMTLabels` option is set.                                                                                              
                                                                                                                                                                                      
### Changes                                                                                                                                                                           
                                                                                                                                                                                      
  - add `coreIndex` in `LinuxMachine` which tracks normalized ID
  - return `coreIndex` in `Machine_getCPUPhysicalCoreID()`

### Motivation

This fixes the IDs displayed on systems with non-contiguous core IDs such as the xeon system that @lilydjwg tested on in #1902
[cpuinfo.txt](https://github.com/user-attachments/files/26490538/cpuinfo.txt)

### Future Work

It would be cool to have a mode that renders the CPU IDs as `socket.core.thread`, but this requires more than the three characters of space currently available.

### DIFF
--- a/linux/LinuxMachine.c
+++ b/linux/LinuxMachine.c
@@ -748,6 +748,9 @@ static void LinuxMachine_computeThreadIndices(LinuxMachine* this) {
       cpus[i].coreIndex = coreIndex;
    }
 
+   /* Set core & thread indices to zero for cpu0 (average) */
+   cpus[0].coreIndex = 0;
+   cpus[0].threadIndex = 0;
 }
 
 static void LinuxMachine_scanCPUFrequency(LinuxMachine* this) {
--- a/linux/LinuxMachine.c
+++ b/linux/LinuxMachine.c
@@ -739,13 +739,13 @@ static void LinuxMachine_computeThreadIndices(LinuxMachine* this) {
       because CoreIDs are not contiguous or because cpus are
       enumerated in an alternative order, or both. */
    for (size_t i = 1; i <= super->existingCPUs; i++) {
-      int coreIndex = 0;
-      for (size_t j = 1; j < i; j++) {
+      cpus[i].coreIndex = 0;
+      for (size_t j = i - 1; j >= 1; j--) {
          if (cpus[i].threadIndex == cpus[j].threadIndex) {
-            coreIndex++;
+            cpus[i].coreIndex = cpus[j].coreIndex + 1;
+            break;
          }
       }
-      cpus[i].coreIndex = coreIndex;
    }
 
    /* Set core & thread indices to zero for cpu0 (average) */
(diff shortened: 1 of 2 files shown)


linux/LinuxMachine.c
```diff
@@ -731,6 +731,23 @@ static void LinuxMachine_computeThreadIndices(LinuxMachine* this) {
...
       cpus[i].threadIndex = threadIndex;
    }
+
+   /* Now compute a normalized physical core index for each CPU.
+      On many systems, this index will match the following:
+        physicalID*(maxPhysicalID+1)+coreID
+      But there are some systems where this is not true, either
+      because CoreIDs are not contiguous or because cpus are
+      enumerated in an alternative order, or both. */
+   for (size_t i = 1; i <= super->existingCPUs; i++) {
```
Explorer09, line 741: Better not let `cpus[0].coreIndex` undefined.

linux/LinuxMachine.c
```diff
@@ -731,6 +731,26 @@ static void LinuxMachine_computeThreadIndices(LinuxMachine* this) {
...
+      int coreIndex = 0;
+      for (size_t j = 1; j < i; j++) {
+         if (cpus[i].threadIndex == cpus[j].threadIndex) {
+            coreIndex++;
+         }
+      }
```
Explorer09: Do it faster:

```suggestion
      for (size_t j = i - 1; j >= 1; j--) {
         if (cpus[i].threadIndex == cpus[j].threadIndex) {
            cpus[i].coreIndex = cpus[j].coreIndex + 1;
            break;
         }
      }
```

tjk213: i had to also delete the line below which was overwriting this assignment. but i made that edit and pushed the changes, thanks for the optimization!

linux/LinuxMachine.c
```diff
@@ -731,6 +731,25 @@ static void LinuxMachine_computeThreadIndices(LinuxMachine* this) {
...
    }
+
+   /* Now compute a normalized physical core index for each CPU.
+      On many systems, this index will match the following:
+        physicalID*(maxPhysicalID+1)+coreID
+      But there are some systems where this is not true, either
+      because CoreIDs are not contiguous or because cpus are
+      enumerated in an alternative order, or both. */
+   for (size_t i = 1; i <= super->existingCPUs; i++) {
+      for (size_t j = i - 1; j >= 1; j--) {
```
Explorer09, line 743: It's my mistake of missing this, but you should add a line `cpus[i].coreIndex = 0;` before the `for` loop.

PR #1942, opened 2026-04-05
Align AvailableMeters function bar & key behaviors with AvailableColumnsPanel

Explorer09:
1. Make AvailableMeters functions bar same as in AvailableColumns functions.
2. Merge the F5 ("Add") and Enter events, and obsolete old key events ('L', 'R' and F6) in AvailableMeters panel.
3. Allow `KEY_RECLICK` to add a meter.

### DIFF
--- a/AvailableMetersPanel.c
+++ b/AvailableMetersPanel.c
@@ -58,21 +58,12 @@ static HandlerResult AvailableMetersPanel_eventHandler(Panel* super, int ch) {
    bool update = false;
 
    switch (ch) {
-      case KEY_F(5):
-      case 'l':
-      case 'L':
-         AvailableMetersPanel_addMeter(header, this->meterPanels[0], Platform_meterTypes[type], param, 0);
-         result = HANDLED;
-         update = true;
-         break;
       case 0x0a:
       case 0x0d:
       case KEY_ENTER:
-      case KEY_F(6):
-      case 'r':
-      case 'R':
+      case KEY_F(5):
          AvailableMetersPanel_addMeter(header, this->meterPanels[this->columns - 1], Platform_meterTypes[type], param, this->columns - 1);
-         result = (KEY_LEFT << 16) | SYNTH_KEY;
+         result = HANDLED;
          update = true;
          break;
    }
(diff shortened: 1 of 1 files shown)


fasterit:
I'd keep the "r" and "l" keys, people may have them in their muscle memory.

Explorer09:
> I'd keep the "r" and "l" keys, people may have them in their muscle memory.

The 'R' and 'L' keys don't make sense for the multi-column UI of the meters, and you would have to explain what 'R' and 'L' stand for.

They're unnecessary nuisance for the code. Once I realize there only needs to be one add button (just like the Available Columns panel has done), it's better to have the old hotkeys gone. For room of better UX.

fasterit:
Move (R)ight and move (L)eft

Explorer09:
> Move (R)ight and move (L)eft

Note: 'R' and 'L' still work for moving meters between columns once they are added. It's just they cannot be invoked from the "Available Meters" anymore.

fasterit:
See, "L", "L" to add and move seemed quite logical. Now you force it to be "F5", "L" for no good reason.
At least the UX does not get worsened by it, you cleaned up the "odd" function bar nicely.

Explorer09:
> See, "L", "L" to add and move seemed quite logical. Now you force it to be "F5", "L" for no good reason. At least the UX does not get worsened by it, you cleaned up the "odd" function bar nicely.

If you press "L" to add a meter from AvailableMeters, where should the text cursor get focused on, the AvailableMeters panel, or the Meter panel of the column?

Because the old behavior didn't suggest the workflow you described, and now I'm confused.

In the AvailableColumns panel, pressing Enter or F5 twice adds two copies of the same column. Although user might do so only as an accident, the cursor focus remains on the AvailableColumns, so they can pick other columns to add very smoothly.

fasterit:
If I use "L" now, it adds to the left column, "R" to the right column. That's historic, quite probably before we got three and four column layouts. And nobody added at least an "M" for the middle one.

Explorer09:
> If I use "L" now, it adds to the left column, "R" to the right column. That's historic, quite probably before we got three and four column layouts. And nobody added at least an "M" for the middle one.

Won't make sense if the header layout is one column only, or four columns or more.

fasterit:
Cherrypicked to 319d16ab1c61f1cf27e0ce2d1889da4bf2e25e0b and 1c53fa0806d16f09e0dc91396ea746aaec17ffa4
Thank you.

Issue #1943, opened 2026-04-05
MetersPanel in Setup always counts CPUs from 1 despite the setting.


Explorer09:
See this screenshot. When the "Count CPUs from 1 instead of 0" is off, the CPU numbering in the Meters panel ("Column 1" and "Column 2") is still 1 to 8. Should be 0 to 7.

<img width="1018" height="604" alt="htop-screenshot-meters-panel" src="https://github.com/user-attachments/assets/fbcdea8a-f228-40d7-9dd8-0d1f320e857f" />

fasterit:
Setting (also) not respected when "Label CPUs based on SMT topology (e.g. 0a, 0b) instead of CPU index" is on.
Cf. 772da756 f9b1695c

PR #1946, opened 2026-04-06
Fix CPU numbering in MetersPanel setup and when showing SMT labels

fasterit:
Two fixes for #1943 and comment 1 there


natoscott:
LGTM.

PR #1951, opened 2026-04-09
Add documentation for the versioning scheme

fasterit:


docs/understanding-htop-versions.md
```diff
@@ -0,0 +1,35 @@
...
+This version id will be ammended by
+
+```
+git describe --abbrev=7 --dirty --always --tags
+```
+
+which you can find as `htop_git_version` in `configure.ac`.
+
+A development version will identify in `htop --version` and the crash report as:
+- htop 10.9.8-dev-10.9.0-16-aabbccdd
```
BenBE: Minor note: It's `gabbccdd`, i.e. the first letter in teh git commit is always g followed by the actual commit hash.

fasterit: tru dat, thanks!

BenBE (approved):
LGTM.

PR #1952, opened 2026-04-10
darwin: fix memory meter showing ~64 TB on ARM64 due to unsigned underflow

EsDmitrii:
## Problem

On macOS with Apple Silicon, the memory meter can display wildly incorrect
values — typically ~64 TB used out of 48 GB total. This makes the memory
bar completely unusable.

The root cause is an unsigned integer underflow in `Platform_setMemoryValues()`
(`darwin/Platform.c`). The active page count is computed as:

    active_count - purgeable_count - external_page_count

All three variables are `natural_t` (unsigned 32-bit). On systems where
file-backed pages (`external_page_count`) exceed `active_count - purgeable_count`,
the subtraction wraps around to ~4,294,867,000 pages. Multiplied by
`page_K` (16 for ARM64's 16 KB pages), this produces ~64 TB.

Example from a live system (M4 Max, 48 GB RAM):

    active_count         = 1,111,998
    purgeable_count      =    66,505
    external_page_count  = 1,144,396
    result (signed)      =   -98,903
    result (unsigned)    = 4,294,868,393
    × 16 KB              ≈ 64.0 TB         ← displayed by htop

This is a normal state on macOS — the file cache often exceeds
active pages, especially after boot or heavy file I/O.

## Fix

Cast each page count to `double` before subtraction so the arithmetic
is signed, and clamp the result to zero using the existing `MAXIMUM()`
macro. This avoids the wraparound while preserving the original
calculation semantics.

Both the `showCachedMemory=true` and `showCachedMemory=false` code paths
are fixed.

## Before / After
Screenshot 2026-04-10 at 16 17 36
Screenshot 2026-04-10 at 16 18 31
Screenshot 2026-04-10 at 16 19 01

## Test

Built and ran on macOS Sequoia 15.x, Apple M4 Max (48 GB).
Memory meter now shows correct values consistent with `top` and `vm_stat`.

### DIFF
--- a/darwin/Platform.c
+++ b/darwin/Platform.c
@@ -432,14 +432,21 @@ void Platform_setMemoryValues(Meter* mtr) {
 
    mtr->total = dhost->host_info.max_mem / 1024;
    mtr->values[MEMORY_CLASS_WIRED]       = page_K * vm->wire_count;
+
+   /*
+    * Use saturatingSub() to prevent unsigned underflow: on macOS,
+    * external_page_count (file-backed pages) can exceed active_count,
+    * causing the result to wrap around to ~4 billion pages (~64 TB on
+    * 16K-page ARM64 systems).
+    */
    if (settings->showCachedMemory) {
       mtr->values[MEMORY_CLASS_SPECULATIVE] = page_K * vm->speculative_count;
-      mtr->values[MEMORY_CLASS_ACTIVE]      = page_K * (vm->active_count - vm->purgeable_count - external_page_count); // external pages are pages swapped out
-      mtr->values[MEMORY_CLASS_PURGEABLE]   = page_K * vm->purgeable_count; // purgeable pages are flagged in the active pages
+      mtr->values[MEMORY_CLASS_ACTIVE]      = page_K * saturatingSub(vm->active_count, (unsigned long long)vm->purgeable_count + external_page_count);
+      mtr->values[MEMORY_CLASS_PURGEABLE]   = page_K * vm->purgeable_count;
    }
-   else { // if showCachedMemory is disabled, merge speculative and purgeable into the active pages
+   else {
       mtr->values[MEMORY_CLASS_SPECULATIVE] = 0;
-      mtr->values[MEMORY_CLASS_ACTIVE]      = page_K * (vm->speculative_count + vm->active_count - external_page_count); // external pages are pages swapped out
+      mtr->values[MEMORY_CLASS_ACTIVE]      = page_K * saturatingSub((unsigned long long)vm->speculative_count + vm->active_count, external_page_count);
       mtr->values[MEMORY_CLASS_PURGEABLE]   = 0;
    }
    mtr->values[MEMORY_CLASS_COMPRESSED]  = page_K * compressor_page_count;


anitoanto:
Possible fix for https://github.com/htop-dev/htop/issues/1953

Explorer09:
htop has a macro called saturatingSub() (in Macros.h). Maybe it's better to use that instead of cast to `double`.

Explorer09:
By the way, how did you capture the numbers as reported here?

> ```text
> active_count         = 1,111,998
> purgeable_count      =    66,505
> external_page_count  = 1,144,396
> ```

I hope these numbers are not made up.

EsDmitrii:
> htop has a macro called saturatingSub() (in Macros.h). Maybe it's better to use that instead of cast to `double`.

Let me double check

EsDmitrii:
@Explorer09 Good call, updated. Using saturatingSub() with promotion to unsigned long long before the intermediate addition to stay safe across the full range.

About numbers these are live values captured from vm_stat on the affected system (M4 Max, 48 GB, macOS Sequoia). The command:

`vm_stat | awk '/Pages active/{print "active_count =", $NF+0} /Pages purgeable/{print "purgeable_count =", $NF+0} /File-backed pages/{print "external_page_count =", $NF+0}'`

The key condition is external_page_count > active_count - purgeable_count, which is a normal state on macOS when the file cache is large (e.g. right after boot or heavy file I/O). 
Anyone with an ARM64 Mac can reproduce by checking vm_stat — if file-backed pages exceed active pages, htop will show ~64 TB in the memory meter.

Explorer09:
@EsDmitrii 
> Anyone with an ARM64 Mac can reproduce by checking vm_stat 

I have an M2 MacBook, but unfortunately my use cases with a Mac didn't involve in a large number of file-backed pages. In other words I cannot reproduce the issue during my normal use.

BenBE:
Please squash/fixup the commits into one as per our style guide …

SuhasSrinivasan:
Also reported in #1958

diego-lipinski-de-castro:
Has this been updated in homebrew? I have 3.5.0 installed on my mac m1 max and still shows 64tb ram

EsDmitrii:
> Has this been updated in homebrew? I have 3.5.0 installed on my mac m1 max and still shows 64tb ram
idk because it works for me on 3.5.0
Screenshot 2026-04-21 at 15 03 22

I can't test or debug it on other proc(
but you can share here these numbers:
`vm_stat | awk '/Pages active/{print "active_count =", $NF+0} /Pages purgeable/{print "purgeable_count =", $NF+0} /File-backed pages/{print "external_page_count =", $NF+0}'`

diego-lipinski-de-castro:
Captura de Tela 2026-04-21 às 18 19 08

Captura de Tela 2026-04-21 às 18 19 16

EsDmitrii:
@diego-lipinski-de-castro Hey! The fix has been merged into htop-dev:main two weeks ago and is tracked under the 3.5.1 milestone, but no release has been cut yet so Homebrew is still serving 3.5.0 without the patch.
Screenshot 2026-04-22 at 12 08 08

To verify the fix works for you right now, you can build from source:
```
git clone https://github.com/htop-dev/htop.git
cd htop
./autogen.sh
./configure
make
sudo make install
```
This will pull the current main branch which already includes the fix. 
Once 3.5.1 is released, Homebrew will pick it up automatically.

jyxjjj:
> Once 3.5.1 is released, Homebrew will pick it up automatically.

May I ask if there is any approximate timeline for the 3.5.1 release?

I fully understand that release scheduling depends on many factors, so I would greatly appreciate any information you could share.

EsDmitrii:
> > Once 3.5.1 is released, Homebrew will pick it up automatically.
> 
> May I ask if there is any approximate timeline for the 3.5.1 release?
> 
> I fully understand that release scheduling depends on many factors, so I would greatly appreciate any information you could share.

I don't have any info(
I just user who found the issue and fixed it:)
It's better to ask upstream maintainer

SuhasSrinivasan:
The current PR is not a complete fix, please see issue found in htop v3.5.1 #1975 and the PR #1976

PR #1955, opened 2026-04-12
linux: show M_PRIV instead of M_SHARE in default columns

### LINKED ISSUE #492: Add column with showing per-process unique memory usage value (calculate RES - SHR)
Very often we need to know, how much unique RAM used by specific process. The unique memory usage of separate process can be easily calculated manually via `RES - SHR` formula (or `VIRT - SHR - SWAP`?), but online calculating manually for each row is too hard task.

For solve this problem, will be good to add "Process unique memory" (`PUM`?) that will calculate this automatically, alsy with ability to sort process list by this value. What do you think about this idea?

ugolnikovE:
Replaces M_SHARE with M_PRIV in the default Linux Main screen columns, as suggested in #492. M_PRIV (RES - SHR) is a more actionable per-process memory metric than shared pages alone.

M_RESIDENT is kept as it remains the de-facto standard memory column across process viewers. M_SHARE is removed since it becomes redundant with M_PRIV. Users who want either column back can re-add them via the Setup screen (F2).

PCP also exposes M_PRIV but its defaults are left unchanged to keep this PR focused on Linux; happy to extend in a follow-up if desired.

Also documents M_PRIV in the man page.

Tested on Debian aarch64 with a fresh config: PRIV column appears in place of SHR, sorting works, config saves and reloads cleanly.

Closes #492

### DIFF
--- a/htop.1.in
+++ b/htop.1.in
@@ -434,6 +434,10 @@ process's used physical memory).
 .B M_SHARE (SHR)
 The size of the process's shared pages.
 .TP
+.B M_PRIV (PRIV)
+The private memory size of the process (i.e. the resident set size minus the
+shared memory size).
+.TP
 .B M_TRS (CODE)
 The text resident set size of the process (i.e. the size of the process's
 executable instructions).
--- a/linux/Platform.c
+++ b/linux/Platform.c
@@ -90,7 +90,7 @@ bool Running_containerized = false;
 const ScreenDefaults Platform_defaultScreens[] = {
    {
       .name = "Main",
-      .columns = "PID USER PRIORITY NICE M_VIRT M_RESIDENT M_SHARE STATE PERCENT_CPU PERCENT_MEM TIME Command",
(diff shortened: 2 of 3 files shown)


natoscott (approved):
LGTM.  If you are happy to add the PCP change needed, please go ahead with that too (perhaps in a follow-up commit or even separate PR if you prefer).

ugolnikovE:
Added the PCP commit.
While testing, I noticed that PRIV and SHR display without unit suffixes in pcp-htop, while RES and VIRT show correctly (see screenshot).
Couldn't find an existing issue. Should I open one, or is it known?
Screenshot 2026-04-15 at 22 44 06

natoscott:
> Couldn't find an existing issue. Should I open one, or is it known?

@evelikov that sounds like the metric units may be incorrect - but from a quick look, doesn't seem so - both are in units of KiB.  I'll take a close look when I can.  It's not a known issue so please go ahead and open that new issue, thanks!

natoscott:
Ah, I see the problem - PR to follow, don't worry about a new issue.

PR #1956, opened 2026-04-12
Fix search function

fasterit:
Fixing two funny bugs that remained after the line editor introduction:

1. Clicking into the panel does not deactivate the search function
1. Search activates the following function on every second keypress. Toggle != activate :)

### DIFF
--- a/MainPanel.c
+++ b/MainPanel.c
@@ -105,7 +105,8 @@ static HandlerResult MainPanel_eventHandler(Panel* super, int ch) {
          reaction = HTOP_REFRESH | HTOP_REDRAW_BAR;
       }
       if (this->inc->found && this->inc->active && !this->inc->active->isFilter) {
-         reaction |= Action_follow(this->state);
+         host->activeTable->following = MainPanel_selectedRow(this);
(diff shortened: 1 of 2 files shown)


MainPanel.c
```diff
@@ -99,13 +99,16 @@ static HandlerResult MainPanel_eventHandler(Panel* super, int ch) {
       reaction |= Action_setScreenTab(this->state, x);
       result = HANDLED;
    } else if (ch != ERR && this->inc->active) {
-      bool filterChanged = IncSet_handleKey(this->inc, ch, super, MainPanel_getValue, NULL);
+      /* A click in the panel body aborts the search/filter */
+      int keyToHandle = (ch == KEY_MOUSE || ch == KEY_RECLICK) ? 27 : ch;
```
BenBE: LGTM.

But doesn't this decentralize key handling for the incset somewhat?

BenBE (approved):

MainPanel.c
fasterit: It falls through to the line editor now and that one does not handle it. So that i-t-e seemed fine :smile:.

fasterit: I moved it to IncSet, thank you for the nudge!

PR #1957, opened 2026-04-12
Stable tree view

### LINKED ISSUE #1905: Fix tree view jumparound once for all
I think I opened this issue back when this was under @hishamhm's care, and even that a lot changed since then, this issue is making `htop` so unusable that I am forced to use it less and less .. especially since competition is catching up ... which is a great pity.

Tree mode is most useful of modes for sysadmins, since it shows process hierarchies on the machine, which can deliver much more information, than pure COLUMN ordered process list.

Unfortunately as machines get bigger the process trees grow bigger as well. 

Due to process being relatively cheap resource, on modern machines, the process activity, especially creation and destruction rate, is little bit higher than during the days of old.

And therein lies the issue: when you put `htop` into ptree mode, it constantly jumps around on the screen as processes fork and exit. Besides quick glance, you cannot really do anything useful, because exactly when you want to do some process op, the screen jumps, and you select wrong process. Bummer.

This issue is ancient. And it's becoming mighty annoying.

What is worse, I was told, back then, that due to design of the tree renderer, this issue is "unfixable".

Unfortunately, since some years ago, `top`, from `procps-ng`, which is "native" to most distros, grew ability to draw trees.

Start `top` and then press `V` (as in `Shift`+`V`) and behold: a tree! What is more concerning: Look ma! No jumping!

How they do? Woof!? Woof!?

fasterit:
Improvements to the following function are needed to make the stable tree view make sense (otherwise it would be quite just a duplication of functionality). Hence commit #1.

The stable tree view comes in two flavors:
soft = htop tries to keep the tree view stable but does not allow blank space above the tree root (PID 0) to accommodate all possible tree refreshs
hard = htop allows blank lines above the tree root (PID 0) to keep the visual line at nearly all cost

The default is off, so retain existing jumpiness.

I did not implement a hot key for switching as I *think* people who like stable tree view will just switch it on and keep it that way.

Fixes #1905

### DIFF
--- a/CommandLine.c
+++ b/CommandLine.c
@@ -68,7 +68,7 @@ static void printHelpFlag(const char* name) {
           "-p --pid=PID[,PID,PID...]       Show only the given PIDs\n"
           "   --readonly                   Disable all system and process changing features\n"
           "-s --sort-key=COLUMN            Sort by COLUMN in list view (try --sort-key=help for a list)\n"
-          "-t --tree                       Show the tree view (can be combined with -s)\n"
+          "-t --tree[=MODE]                Show the tree view (MODE: classic|soft|hard); can be combined with -s\n"
           "-u --user[=USERNAME]            Show only processes for a given user (or $USER)\n"
           "-U --no-unicode                 Do not use unicode but plain ASCII\n"
           "-V --version                    Print version info\n");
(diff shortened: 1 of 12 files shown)


jtmoon79:
This is great!  Should there be a command-line option?
Maybe `--tree-view-stabilize` with enumerated string values like `stable, stable-no-blanks, off` ?

fasterit:
I like the idea, but I added it as an optional argument to `--tree`. That seems easier to discover and remember.

CommandLine.c
```diff
@@ -100,6 +101,25 @@ typedef struct CommandLineSettings_ {
    bool hideFunctionBar;
 } CommandLineSettings;
 
+static bool parseTreeStableMode(const char* arg, int* stableTreeView) {
+   if (String_eq(arg, "0") || String_eq(arg, "jumpy")) {
```
BenBE:
```suggestion
   if (String_eq(arg, "0") || String_eq(arg, "jumpy") || String_eq(arg, "legacy")) {
```

fasterit: @BenBE: How do you like "classic" ?

fasterit:
I would probably squash [7d7c55d](https://github.com/htop-dev/htop/pull/1957/commits/7d7c55d779d801a776b0ae479481f3379f5b3b5d) into https://github.com/htop-dev/htop/pull/1957/commits/74a3b1490c2c7db033c891d30b5e0c8d46ae427c before merge.

Waiting on feedback from @BenBE for the mode names.

CommandLine.c
BenBE: Also fine. ;-)

fasterit: Oki, thanks, then RTM

fasterit:
Partially squashed as per https://github.com/htop-dev/htop/pull/1957#issuecomment-4366414264

PR #637, opened 2021-05-23
Add option to shadow path prefixes

cgzones:
Shadow path prefixes which are used by distributions, like
  - /usr/bin/
  - /usr/sbin/
  - /bin/
  - /sbin/
  - /usr/libexec/

PR #1961, opened 2026-04-16
Add user-configurable distribution path prefixes (implements #822)

### LINKED ISSUE #822: Implement text input control for settings
For allowing the proposed feature in #637 to be user-configurable there needs to be some UI for the user to input some text, edit the existing value and possibly have the UI provide some validity feedback.

The new text input should extend the list of normal settings controls we already have for checkbox items and numbers by a new text input method. That text input method should allow for some basic validation of the user input as well as provide a sensible subset of the (basic) readline features for text cursor control, word/line editing and possibly tab completion (e.g. when the input setting is a list of paths).

The second part consists of an example implementation of this new control to allow for #637 to be user-configurable list of paths.

JonasPfi:
Fixes #822, depends on #637

### Problem

The 'Shadow distribution path prefixes' feature (#637) uses a hardcoded list of paths (`/usr/bin/`, `/bin/`, `/nix/store/`, `/run/current-system/`, etc.) that cannot be changed by the user. This makes the feature unusable for non-standard setups such as Flatpak (`/var/lib/flatpak/`), custom install prefixes, or other distributions with different filesystem layouts.

### Solution

This PR implements two things as requested in #822:

**1. A new `StringItem` settings control**

Extends the existing settings UI infrastructure (`OptionItem`) with a new text input type backed by the existing `LineEditor`. This gives users a readline-like editing experience with support for:
- Cursor movement (`←`/`→`, `Ctrl+A`, `Ctrl+E`)
- Backspace / Delete

The `StringItem` follows the same patterns as the existing `CheckItem` and `NumberItem` controls.

**2. User-configurable path prefix list**

Replaces the hardcoded `CHECK_AND_MARK_DIST_PATH_PREFIXES` switch macro in `Process.c` with a `matchesDistPrefix()` function that checks against a user-supplied colon-separated list of paths. When the list is empty, the original built-in defaults are used as fallback.

### Changes

- `OptionItem.h/.c` — new `StringItem` type with `LineEditor` integration
- `Settings.h` — new `char* distPathPrefixes` field
- `Settings.c` — read/write `dist_path_prefixes` from/to htoprc
- `DisplayOptionsPanel.c` — new text input field below the existing checkbox
- `Process.c` — replace hardcoded switch macro with `matchesDistPrefix()`

### Testing

- Custom paths are highlighted correctly after input
- Settings persist across restarts
- Empty input correctly falls back to built-in defaults
- `valgrind --leak-check=full` reports 0 definitely lost bytes
- Escape correctly cancels editing and restores previous value
- Tested on WSL2 (Ubuntu) with Linux build

### DIFF
--- a/DisplayOptionsPanel.c
+++ b/DisplayOptionsPanel.c
@@ -45,12 +45,19 @@ static HandlerResult DisplayOptionsPanel_eventHandler(Panel* super, int ch) {
    }
 
    NumberItem* numItem = (OptionItem_kind(selected) == OPTION_ITEM_NUMBER) ? (NumberItem*)selected : NULL;
+   StringItem* strItem = (OptionItem_kind(selected) == OPTION_ITEM_STRING) ? (StringItem*)selected : NULL;
 
    /* Helper: position the hardware cursor right after the edit buffer.
     * +1 on Y for the panel header row; +1 on X for the leading '[' bracket. */
    #define SET_EDIT_CURSOR() do { \
       super->cursorY = super->y + 1 + (super->selected - super->scrollV); \
       super->cursorX = super->x + 1 + numItem->editLen; \
+      super->cursorOn = true; \
+   } while (0)
+
+   #define SET_STR_CURSOR() do { \
+      super->cursorY = super->y + 1 + (super->selected - super->scrollV); \
+      super->cursorX = super->x + 1 + (int)LineEditor_getCursor(&strItem->editor); \
       super->cursorOn = true; \
    } while (0)
 
@@ -61,9 +68,19 @@ static HandlerResult DisplayOptionsPanel_eventHandler(Panel* super, int ch) {
             super->cursorOn = false;
             return HANDLED;
          }
+         if (numItem && numItem->editing) {
+            NumberItem_cancelEditing(numItem);
+            super->cursorOn = false;
+            return HANDLED;
+         }
          break;
       case KEY_BACKSPACE:
       case KEY_DEL_MAC:
+         if (strItem && strItem->editing) {
+            LineEditor_handleKey(&strItem->editor, KEY_BACKSPACE);
+            SET_STR_CURSOR();
+            return HANDLED;
+         }
          if (numItem) {
             if (!numItem->editing) {
                NumberItem_startEditingFromValue(numItem);
@@ -76,6 +93,19 @@ static HandlerResult DisplayOptionsPanel_eventHandler(Panel* super, int ch) {
       case '\n':
       case '\r':
       case KEY_ENTER:
+         if (strItem && strItem->editing) {
+            StringItem_applyEditing(strItem);
+            super->cursorOn = false;
+            settingsChanged = true;
+            result = HANDLED;
+            break;
+         }
+         if (strItem && !strItem->editing) {
+            StringItem_startEditing(strItem);
+            SET_STR_CURSOR();
+            result = HANDLED;
+            break;
+         }
          if (numItem && numItem->editing) {
             if (NumberItem_applyEditing(numItem)) {
                settingsChanged = true;
@@ -219,6 +249,12 @@ static HandlerResult DisplayOptionsPanel_eventHandler(Panel* super, int ch) {
                SET_EDIT_CURSOR();
                return HANDLED;
             }
+         } else if (strItem) {
+            if (strItem->editing) {
+               LineEditor_handleKey(&strItem->editor, ch);
+               SET_STR_CURSOR();
+               return HANDLED;
+            }
          }
          break;
    }
@@ -282,6 +318,7 @@ DisplayOptionsPanel* DisplayOptionsPanel_new(Settings* settings, ScreenManager* 
    Panel_add(super, (Object*) CheckItem_newByRef("Highlight program \"basename\"", &(settings->highlightBaseName)));
    Panel_add(super, (Object*) CheckItem_newByRef("Highlight out-dated/removed programs (red) / libraries (yellow)", &(settings->highlightDeletedExe)));
    Panel_add(super, (Object*) CheckItem_newByRef("Shadow distribution path prefixes", &(settings->shadowDistPathPrefix)));
+   Panel_add(super, (Object*) StringItem_newByRef("- Custom path prefixes (colon-separated)", &(settings->distPathPrefixes), NULL));
    Panel_add(super, (Object*) CheckItem_newByRef("Merge exe, comm and cmdline in Command", &(settings->showMergedCommand)));
    Panel_add(super, (Object*) CheckItem_newByRef("- Try to find comm in cmdline (when Command is merged)", &(settings->findCommInCmdline)));
    Panel_add(super, (Object*) CheckItem_newByRef("- Try to strip exe from cmdline (when Command is merged)", &(settings->stripExeFromCmdline)));
--- a/OptionItem.c
+++ b/OptionItem.c
@@ -76,6 +76,27 @@ static void NumberItem_display(const Object* cast, RichString* out) {
    RichString_appendWide(out, CRT_colors[CHECK_TEXT], this->super.text);
 }
 
+static void StringItem_display(const Object* cast, RichString* out) {
+   const StringItem* this = (const StringItem*)cast;
+   int labelAttr = CRT_colors[CHECK_TEXT];
+   int boxAttr   = CRT_colors[CHECK_BOX];
+   int valAttr   = this->valid ? CRT_colors[CHECK_MARK] : CRT_colors[FAILED_READ];
+
+   RichString_writeAscii(out, boxAttr, "[");
+   if (this->editing) {
+      RichString_appendAscii(out, valAttr, this->editor.buffer);
+   } else {
+      const char* val = (this->ref && *this->ref) ? *this->ref : NULL;
+      if (val) {
+         RichString_appendAscii(out, valAttr, val);
+      } else {
+         RichString_appendAscii(out, CRT_colors[PROCESS_SHADOW], "(empty)");
+      }
+   }
+   RichString_appendAscii(out, boxAttr, "]  ");
+   RichString_appendWide(out, labelAttr, this->super.text);
+}
+
 const OptionItemClass OptionItem_class = {
    .super = {
       .extends = Class(Object),
@@ -110,6 +131,15 @@ const OptionItemClass NumberItem_class = {
       .display = NumberItem_display
    },
    .kind = OPTION_ITEM_NUMBER
+};
+
+const OptionItemClass StringItem_class = {
+   .super = {
+      .extends = Class(OptionItem),
+      .delete = OptionItem_delete,
+      .display = StringItem_display
+   },
+   .kind = OPTION_ITEM_STRING
 };
 
 TextItem* TextItem_new(const char* text) {
@@ -322,4 +352,41 @@ void NumberItem_deleteChar(NumberItem* this) {
       this->editLen--;
       this->editBuffer[this->editLen] = '\0';
    }
+}
+StringItem* StringItem_newByRef(const char* text, char** ref, bool (*validate)(const char* text)) {
+   StringItem* this = AllocThis(StringItem);
+   this->super.text = xStrdup(text);
+   this->ref = ref;
+   this->editing = false;
+   this->valid = true;
+   this->validate = validate;
+   LineEditor_init(&this->editor);
+   if (ref && *ref)
+      LineEditor_setText(&this->editor, *ref);
+   return this;
+}
+
+void StringItem_startEditing(StringItem* this) {
+   this->editing = true;
+   LineEditor_setText(&this->editor, (this->ref && *this->ref) ? *this->ref : "");
+}
+
+void StringItem_cancelEditing(StringItem* this) {
+   this->editing = false;
+   LineEditor_setText(&this->editor, (this->ref && *this->ref) ? *this->ref : "");
+}
+
+bool StringItem_applyEditing(StringItem* this) {
+   this->editing = false;
+   const char* text = this->editor.buffer;
+   if (this->validate && !this->validate(text)) {
+      this->valid = false;
+      return false;
+   }
+   this->valid = true;
+   if (this->ref) {
+      free(*this->ref);
+      *this->ref = (*text != '\0') ? xStrdup(text) : NULL;
+   }
+   return true;
 }
--- a/OptionItem.h
+++ b/OptionItem.h
@@ -9,6 +9,7 @@ in the source distribution for its full text.
 
 #include <stdbool.h>
 
+#include "LineEditor.h"
 #include "Object.h"
 
 #define NUMBERITEM_EDIT_MAX 10
@@ -17,6 +18,7 @@ enum OptionItemType {
    OPTION_ITEM_TEXT,
    OPTION_ITEM_CHECK,
    OPTION_ITEM_NUMBER,
+   OPTION_ITEM_STRING,
 };
 
 typedef struct OptionItemClass_ {
@@ -62,10 +64,20 @@ typedef struct NumberItem_ {
    int savedValue;
 } NumberItem;
 
+typedef struct StringItem_ {
+   OptionItem super;
+   char**  ref;
+   bool    editing;
+   bool    valid;
+   LineEditor editor;
+   bool    (*validate)(const char* text);
+} StringItem;
+
 extern const OptionItemClass OptionItem_class;
 extern const OptionItemClass TextItem_class;
 extern const OptionItemClass CheckItem_class;
 extern const OptionItemClass NumberItem_class;
+extern const OptionItemClass StringItem_class;
 
 TextItem* TextItem_new(const char* text);
 
@@ -87,5 +99,10 @@ void NumberItem_cancelEditing(NumberItem* this);
 bool NumberItem_applyEditing(NumberItem* this);
 bool NumberItem_addChar(NumberItem* this, char c);
 void NumberItem_deleteChar(NumberItem* this);
+
+StringItem* StringItem_newByRef(const char* text, char** ref, bool (*validate)(const char* text));
+void StringItem_startEditing(StringItem* this);
+void StringItem_cancelEditing(StringItem* this);
+bool StringItem_applyEditing(StringItem* this);
 
 #endif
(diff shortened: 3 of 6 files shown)


fasterit:
Undeclared use of AI, please read and understand the [AI policy](https://github.com/htop-dev/htop/blob/main/docs/ai-contributions-policy.md). Review your code and only then submit PRs.

JonasPfi:
I added what I used to the commit. Since its my first contribution I used Claude mostly to get used to the repo and also for some logic.

fasterit:
Did you see the "review your code" part?

JonasPfi:
I've reviewed all changes. I removed a comment I missed earlier but did not find anything else.

Explorer09:
My question is: Why do we need this? Even though I believe there are demands for a custom list of path prefixes, I don't think the need for a _UI_ for it is guaranteed.
Also, the format for storing those and loading those in the config can be discussed further. (I personally prefer storing the list as a separate file, e.g. `~/.config/htop/pathprefixes.txt`, not in `htoprc`)

Process.c
```diff
@@ -256,51 +287,10 @@ void Process_makeCommandStr(Process* this, const Settings* settings) {
...
-            case 'n':                                                                         \
-               CHECK_AND_MARK(str_, "/nix/store/");                                           \
-               break;                                                                         \
-            case 'r':                                                                         \
-               CHECK_AND_MARK(str_, "/run/current-system/");                                  \
-               break;                                                                         \
+         size_t _plen = matchesDistPrefix(str_, settings);                                    \
+         if (_plen > 0) {                                                                     \
+            WRITE_HIGHLIGHT(0, _plen, CRT_colors[PROCESS_SHADOW],                             \
+                           CMDLINE_HIGHLIGHT_FLAG_PREFIXDIR);                                 \
```
Explorer09: The `matchesDistPrefix()` can be split to a separate commit. I like this change and it can applied without the custom list of prefixes.

Process.c
```diff
@@ -174,6 +174,37 @@ static inline char* stpcpyWithNewlineConversion(char* dstStr, const char* srcStr
    return dstStr;
 }
 
+static size_t matchesDistPrefix(const char* str, const Settings* settings) {
```
Explorer09: The function return type is unintuitive comparing to the function name. When I read `matchesDistPrefix()` I expect a boolean return type. Or else give us a reason for returning the string length instead.

EDIT: Alternatively, name the function `findDistPrefixLength`

OptionItem.h
```diff
@@ -62,10 +64,20 @@ typedef struct NumberItem_ {
...
 } NumberItem;
 
+typedef struct StringItem_ {
+   OptionItem super;
+   char**  ref;
+   bool    editing;
+   bool    valid;
+   LineEditor editor;
+   bool    (*validate)(const char* text);
+} StringItem;
```
Explorer09, line 74: I'm skeptical of this item. It looks like a duplicate functionality when we have "Screens" whose names are editable.

OptionItem.h
```diff
@@ -88,4 +100,9 @@ bool NumberItem_applyEditing(NumberItem* this);
 bool NumberItem_addChar(NumberItem* this, char c);
 void NumberItem_deleteChar(NumberItem* this);
 
+StringItem* StringItem_newByRef(const char* text, char** ref, bool (*validate)(const char* text));
```
Explorer09, line 103: Split the `bool (*validate)(const char* text)` function definition to its own typedef, please?

Process.c
```diff
@@ -174,6 +174,37 @@ static inline char* stpcpyWithNewlineConversion(char* dstStr, const char* srcStr
...
+   if (settings->distPathPrefixes && settings->distPathPrefixes[0] != '\0') {
+      char* list = xStrdup(settings->distPathPrefixes);
```
Explorer09: Duplicating the string for every `matchesDistPrefix` call creates a performance bottleneck. Also, what if you have a large pool of strings to match with and the system runs low on memory? htop can crash easily.

Process.c
```diff
@@ -174,6 +174,37 @@ static inline char* stpcpyWithNewlineConversion(char* dstStr, const char* srcStr
...
+      }
+      free(list);
+      return matched;
+   }
+
+   static const char* const prefixes[] = {
+      "/bin/", "/sbin/", "/lib/", "/lib32/", "/lib64/", "/libx32/",
+      "/usr/bin/", "/usr/sbin/", "/usr/lib/", "/usr/lib32/", "/usr/lib64/",
+      "/usr/libx32/", "/usr/libexec/", "/usr/local/bin/", "/usr/local/lib/",
+      "/usr/local/sbin/", "/nix/store/", "/run/current-system/",
```
Explorer09: If you are bothered to make a colon separated list, then the built-in list may be colon-separated, too, to save code size.

Process.c
```diff
@@ -174,6 +174,37 @@ static inline char* stpcpyWithNewlineConversion(char* dstStr, const char* srcStr
    return dstStr;
 }
 
+static size_t matchesDistPrefix(const char* str, const Settings* settings) {
+   if (settings->distPathPrefixes && settings->distPathPrefixes[0] != '\0') {
+      char* list = xStrdup(settings->distPathPrefixes);
+      char* saveptr;
+      char* token = strtok_r(list, ":", &saveptr);
```
Explorer09: I prefer using `strchr(3)` or `strpbrk(3)` than `strtok_r(3)`.
With `strchr(3)`, I can get the token length and the pointer to the next token almost simultaneously, without a separate strlen() call.

Another advantage to the former two functions is they don't modify the string.

OptionItem.h
```diff
@@ -62,10 +64,20 @@ typedef struct NumberItem_ {
    int savedValue;
 } NumberItem;
 
+typedef struct StringItem_ {
+   OptionItem super;
+   char**  ref;
+   bool    editing;
+   bool    valid;
+   LineEditor editor;
+   bool    (*validate)(const char* text);
```
Explorer09, line 73: Ditto. (Split the `bool (*validate)(const char* text)` function definition to its own typedef.)

OptionItem.c
```diff
@@ -76,6 +76,23 @@ static void NumberItem_display(const Object* cast, RichString* out) {
...
+   int labelAttr = CRT_colors[CHECK_TEXT];
+   int boxAttr   = CRT_colors[CHECK_BOX];
+   int valAttr   = this->valid ? CRT_colors[CHECK_MARK] : CRT_colors[FAILED_READ];
+
+   RichString_writeAscii(out, boxAttr, "[");
+   if (this->editing) {
+      RichString_appendAscii(out, valAttr, this->editor.buffer);
+   } else {
+      const char* val = (this->ref && *this->ref) ? *this->ref : "";
+      RichString_appendAscii(out, valAttr, *val ? val : "(empty)");
```
Explorer09: Don't make the `(empty)` placeholder indistinguishable from the string containing literally the text "(empty)"

fasterit:
> My question is: Why do we need this? Even though I believe there are demands for a custom list of path prefixes, I don't think the need for a _UI_ for it is guaranteed.

We do not encourage editing `htoprc` by hand.

> Also, the format for storing those and loading those in the config can be discussed further. (I personally prefer storing the list as a separate file, e.g. `~/.config/htop/pathprefixes.txt`, not in `htoprc`)

Inside `htoprc` is fine, we should not inflate the config files for such minor functionality.

fasterit:
Thank you for your review @Explorer09, @JonasPfi you see you have lots to do.

fasterit:
@BenBE, @natoscott, @cgzones:
Do we want this path list to be editable, or do we want to have one canonical list (and add whatever flatpak needs there)?

Explorer09:
> > My question is: Why do we need this? Even though I believe there are demands for a custom list of path prefixes, I don't think the need for a _UI_ for it is guaranteed.
> 
> We do not encourage editing `htoprc` by hand.
> 
> > Also, the format for storing those and loading those in the config can be discussed further. (I personally prefer storing the list as a separate file, e.g. `~/.config/htop/pathprefixes.txt`, not in `htoprc`)
> 
> Inside `htoprc` is fine, we should not inflate the config files for such minor functionality.

My concern is that, if someone needs to ever edit this list, they would likely comment what each path does. The UX requirement for that would be insufficient to fulfill with just a line editor.

Look at how long a PATH variable it is in my system, for example:
macos-26-path-env-var-example

fasterit:
Yes, this might need more than the line editor and a full editor panel would probably not be worth it.
At least I would like to keep the scope creep a bit under control.
Let's see what the others say.

Explorer09:
Another advantage for using an external list file is that distributions can ship with a preconfigured list of paths, freeing the user for needing to configure the list themselves.

If distributions want to ship with a preconfigured list, this list will likely reside in `/etc/htop`, separate from the per-user configurations.

fasterit:
That would still be in a distro-default `/etc/htoprc`, we don't read `/etc/htop/`.

JonasPfi:
> > > My question is: Why do we need this? Even though I believe there are demands for a custom list of path prefixes, I don't think the need for a _UI_ for it is guaranteed.
> > 
> > 
> > We do not encourage editing `htoprc` by hand.
> > > Also, the format for storing those and loading those in the config can be discussed further. (I personally prefer storing the list as a separate file, e.g. `~/.config/htop/pathprefixes.txt`, not in `htoprc`)
> > 
> > 
> > Inside `htoprc` is fine, we should not inflate the config files for such minor functionality.
> 
> My concern is that, if someone needs to ever edit this list, they would likely comment what each path does. The UX requirement for that would be insufficient to fulfill with just a line editor.
> 
> Look at how long a PATH variable it is in my system, for example: macos-26-path-env-var-example

I do agree with this. Currently you won't be able to recreate the default paths with the custom ones, and I think it looks off from a UX perspective to have such a long string in the display options.

Explorer09:
> That would still be in a distro-default `/etc/htoprc`, we don't read `/etc/htop/`.

Um... Putting a settings file inside `/etc` without a subdirectory is a bad idea... Just FYI.

(If we ever need a system-wide htoprc, it would be `/etc/htop/htoprc`)

OptionItem.h
JonasPfi: Could you clarify. Do you mean that StringItem is unnecessary as a new abstraction, and we should instead embed LineEditor directly in DisplayOptionsPanel like ScreensPanel does?

fasterit:
Use the source, Luke. It is what we do. And have been doing for decades.
Ofc, these days people use subdirs in /etc but that was not always the case.

OptionItem.h
Explorer09: No. I mean there's a significant overlap between the two string editing functionality. It's better to consolidate the two functionality into one code. How to do that is left for you to figure out.

Settings.h
```diff
@@ -91,6 +91,7 @@ typedef struct Settings_ {
    bool highlightBaseName;
    bool highlightDeletedExe;
    bool shadowDistPathPrefix;
+   char* distPathPrefixes;
```
Explorer09: I'm not a fan of putting a pointer member within group of `bool` members. The alignment would force this `struct` to waste some bytes in memory (up to 14 bytes, I guess).

OptionItem.h
Explorer09: Uhh. Forgot to mention this: `LINEEDITOR_MAX` is currently 128 (bytes). This is like not enough for editing a list of strings.

DisplayOptionsPanel.c
```diff
@@ -45,6 +45,7 @@ static HandlerResult DisplayOptionsPanel_eventHandler(Panel* super, int ch) {
    }
 
    NumberItem* numItem = (OptionItem_kind(selected) == OPTION_ITEM_NUMBER) ? (NumberItem*)selected : NULL;
+   StringItem* strItem = (OptionItem_kind(selected) == OPTION_ITEM_STRING) ? (StringItem*)selected : NULL;
```
Explorer09, line 48: RTTI (run-time type information). This is a smell of bad object/class modeling. Unfortunately I don't have any idea to fix it for now.

Process.c
JonasPfi: Done, split into two commits.

Process.c
JonasPfi: Renamed `matchedDistPrefix` to `findDistPrefixLength`.

Process.c
```diff
@@ -256,51 +274,10 @@ void Process_makeCommandStr(Process* this, const Settings* settings) {
...
-                  }                                                                           \
-               }                                                                              \
-               break;                                                                         \
-            case 'n':                                                                         \
-               CHECK_AND_MARK(str_, "/nix/store/");                                           \
-               break;                                                                         \
-            case 'r':                                                                         \
-               CHECK_AND_MARK(str_, "/run/current-system/");                                  \
-               break;                                                                         \
+         size_t _plen = findDistPrefixLength(str_);                                    \
```
Explorer09: What motivated you to use a variable name with an underscore prefix?

OptionItem.h
JonasPfi: Agreed. I'll hold off on refactoring until there's consensus on whether the feature is wanted.

Process.c
JonasPfi: Replaced xStrdup/strtok_r with direct iteration using String_strchrnul.

Process.c
JonasPfi: Done.

Process.c
JonasPfi: Switched to String_strchrnul.

OptionItem.c
JonasPfi: The placeholder now uses CRT_colors[PROCESS_SHADOW].

Process.c
```diff
@@ -174,6 +174,24 @@ static inline char* stpcpyWithNewlineConversion(char* dstStr, const char* srcStr
...
+
+   const char* token = builtinPrefixes;
+   while (*token) {
+      const char* end = String_strchrnul(token, ':');
+      size_t len = end - token;
+      if (strncmp(str, token, len) == 0)
+         return len;
+      token = *end ? end + 1 : end;
+   }
+   return 0;
```
Explorer09, line 196: There's still a performance concern with this new function. The old code had a switch-case that could make the lookup slightly faster, but now we're forced to perform a linear search...

I wish a slightly faster data structure such as a [trie](https://en.wikipedia.org/wiki/Trie).

Process.c
JonasPfi: Old habits. I'll change it to `plen`.

OptionItem.c
```diff
@@ -76,6 +76,27 @@ static void NumberItem_display(const Object* cast, RichString* out) {
...
+   int boxAttr   = CRT_colors[CHECK_BOX];
+   int valAttr   = this->valid ? CRT_colors[CHECK_MARK] : CRT_colors[FAILED_READ];
+
+   RichString_writeAscii(out, boxAttr, "[");
+   if (this->editing) {
+      RichString_appendAscii(out, valAttr, this->editor.buffer);
+   } else {
+      const char* val = (this->ref && *this->ref) ? *this->ref : NULL;
+      if (val) {
+         RichString_appendAscii(out, valAttr, val);
```
Explorer09, line 91: Don't use `RichString_appendAscii` for strings read from external sources. Use `RichString_appendWide` instead.

OptionItem.c
```diff
@@ -76,6 +76,27 @@ static void NumberItem_display(const Object* cast, RichString* out) {
    RichString_appendWide(out, CRT_colors[CHECK_TEXT], this->super.text);
 }
 
+static void StringItem_display(const Object* cast, RichString* out) {
+   const StringItem* this = (const StringItem*)cast;
+   int labelAttr = CRT_colors[CHECK_TEXT];
+   int boxAttr   = CRT_colors[CHECK_BOX];
+   int valAttr   = this->valid ? CRT_colors[CHECK_MARK] : CRT_colors[FAILED_READ];
```
Explorer09, line 83: What does the validate function inside a StringItem do? And what would happen (or should happen) if the validation fails?

OptionItem.c
```diff
@@ -323,3 +353,40 @@ void NumberItem_deleteChar(NumberItem* this) {
       this->editBuffer[this->editLen] = '\0';
    }
 }
+StringItem* StringItem_newByRef(const char* text, char** ref, bool (*validate)(const char* text)) {
+   StringItem* this = AllocThis(StringItem);
+   this->super.text = xStrdup(text);
+   this->ref = ref;
```
Explorer09, line 359: What does `ref` do here and how is this argument different from `text`?

OptionItem.c
JonasPfi: Currently its a placeholder. There was the idea of validating if the input path is a valid path in #822  but I did not implement any validation.

Explorer09: And what should be done if the "validation" fails?

JonasPfi: The input should be displayed in red as visual feedback.

OptionItem.c
JonasPfi: `text` is the label displayed next to the input field ( "- Custom path prefixes"). `ref` is a pointer to the settings field being edited (`&settings->distPathPrefixes`), so changes are written directly to the settings struct.

OptionItem.c
Explorer09: The "red" visual feedback would be not good enough if you have a _list_ of path prefixes to edit. Also, I saw no code the "discard" directories that don't exist or are exact duplicates of previously specified paths.

Settings.h
JonasPfi: Grouped 'char*' with other 'char*'.

OptionItem.c
```diff
@@ -323,3 +353,40 @@ void NumberItem_deleteChar(NumberItem* this) {
       this->editBuffer[this->editLen] = '\0';
    }
 }
+StringItem* StringItem_newByRef(const char* text, char** ref, bool (*validate)(const char* text)) {
```
Explorer09, line 356:
```c
typedef bool (*ValidateStringFn)(const char* str, size_t len, void* context);
```

JonasPfi: Will consolidate into a typedef. Will use your suggested signature with len and context if the feature moves forward.

Process.c
JonasPfi: In this specific case implementing a trie might be overkill given the small number of prefixes. A full implementation would add significant complexity for what I'd expect to be a minimal performance gain.

Explorer09: I dont know. If it's me I'd implement it. Otherwise I don't like to trade a performance loss for a feature that few people use.

PR #1962, opened 2026-04-18
Fix overflow use xSnprintf and count n for convert newlines to spaces

GermanAizek:
Affected functions:
- Process_makeCommandStr()
- stpncpyWithNewlineConversion()

### DIFF
--- a/Process.c
+++ b/Process.c
@@ -166,8 +166,8 @@ static size_t matchCmdlinePrefixWithExeSuffix(const char* cmdline, size_t* cmdli
 }
 
 /* stpcpy, but also converts newlines to spaces */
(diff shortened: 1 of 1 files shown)


Explorer09:
Why do this?
Also, note the `snprintf(3)` family of functions don't work well with strings whose length can exceed `INT_MAX`.

Explorer09:
My suggestion is if you ever need a "stpncpy"-like function, make a separate implementation rather than use `xSnprintf` directly.

BenBE:
Using `xSnprintf` for the task `stpcpy` can do is like using a sledgehammer to crack a nut …

fasterit:
we discussed this in the core team and don't think the approach is the right one, we'd more look for a (slim) XUtils wrapper if at all

PR #1968, opened 2026-04-19
Consolidate ClockMeter code with DateTimeMeter code

Explorer09:


DateTimeMeter.c
```diff
@@ -25,28 +29,38 @@ static const int DateTimeMeter_attributes[] = {
...
-
 static void DateTimeMeter_updateValues(Meter* this) {
    const Machine* host = this->host;
 
    struct tm result;
    const struct tm* lt = localtime_r(&host->realtime.tv_sec, &result);
-   strftime(this->txtBuffer, sizeof(this->txtBuffer), "%F %H:%M:%S", lt);
+   strftime(this->txtBuffer, sizeof(this->txtBuffer),
+      (As_Meter(this) == &ClockMeter_class ? "%H:%M:%S" : (As_Meter(this) == &DateMeter_class ? "%F" : "%F %H:%M:%S")),
+      lt);
```
BenBE: Probably best to split this into an `if`/`else if` cascade like this:
```suggestion
   if (As_Meter(this) == &ClockMeter_class) {
      strftime(this->txtBuffer, sizeof(this->txtBuffer), "%H:%M:%S", lt);
   } else if (As_Meter(this) == &DateMeter_class) {
      strftime(this->txtBuffer, sizeof(this->txtBuffer), "%F", lt);
   } else {
      strftime(this->txtBuffer, sizeof(this->txtBuffer), "%F %H:%M:%S", lt);
   }
```

Yes that repeats some stuff, but the intend is much clearer. Also modern compilers should be capable enough to fold this using CSE just fine.

BenBE (commented):
A minor code clarity note, but else this looks fine.

DateTimeMeter.c
Explorer09: Just tested a similar code in Compiler Explorer, and I have some interesting observations:
* LLVM/Clang (22.1.0) can fold the function call pretty well in `-O2`, `-O3` and `-Os`.
* GCC (15.2) can fold it in `-Os`, but somehow refuses to fold in `-O2` and `-O3`; that is, it intentionally makes duplicate `strftime()` calls in three different branches.
So, I can conclude that this code repeating is okay. The CSE is indeed working.

PR #1969, opened 2026-04-19
Focus the target column after adding a meter

### LINKED ISSUE #1948: AvailableMeters inconsistent behavior on "Add Lt", "Add Rt" UI focus
This is a follow up of #1942.

The "Add left" and "Add right" functions in the "Available meters" have inconsistent behaviors on which panel should get focus after the add.

I have a video capture for demonstration:

https://github.com/user-attachments/assets/6d32f85a-3abd-4433-80f6-d6c4dde77834

1. Go to "Setup" -> "Meters"
2. Press Right key repeatedly until the "Available meters" is in focus.
3. Select a meter to add and press F5. Observe the "Available meters" panel remains focused.
4. Select another meter to add and press F6. Observe the focused panel shifts to the "Column 2" panel.

The expected result is consistent behavior in steps 3 and 4.

I can point out that the problem likely happens in the two different `result` values in `AvailableMetersPanel_eventHandler()`: 

https://github.com/htop-dev/htop/blob/6b15bd73f600940e5829208628cc33830dc26fca/AvailableMetersPanel.c#L65

```c
         result = HANDLED;
```

https://github.com/htop-dev/htop/blob/6b15bd73f600940e5829208628cc33830dc26fca/AvailableMetersPanel.c#L76

```c
         result = (KEY_LEFT << 16) | SYNTH_KEY;
```

lawrence3699:
Fixes #1948.

Before this change, adding a meter from Available meters only moved focus when adding to the right-most column. `Add Lt` kept focus on Available meters, so the next keypress still acted on the source list even though the new meter was inserted into Column 1.

This requests focus for the destination column after adding a meter on either side, while keeping the existing bindings intact.

Validation:
- `./autogen.sh && ./configure --enable-werror && make -k -j4`
- `make distcheck DISTCHECK_CONFIGURE_FLAGS='--enable-werror'`\n- local PTY repro on 2-column and 3-column layouts: `Add Lt` and `Add Rt` both switch focus to the target column after insertion

### DIFF
--- a/AvailableMetersPanel.c
+++ b/AvailableMetersPanel.c
@@ -44,6 +44,12 @@ static inline void AvailableMetersPanel_addMeter(Header* header, MetersPanel* pa
    Panel_setSelected((Panel*)panel, Panel_size((Panel*)panel) - 1);
 }
 
+static inline void AvailableMetersPanel_addMeterAndFocus(AvailableMetersPanel* this, size_t column, const MeterClass* type, unsigned int param) {
+   MetersPanel* target = this->meterPanels[column];
+   AvailableMetersPanel_addMeter(this->header, target, type, param, column);
+   ScreenManager_setPanelFocus(this->scr, (Panel*) target);
+}
+
 static HandlerResult AvailableMetersPanel_eventHandler(Panel* super, int ch) {
    AvailableMetersPanel* this = (AvailableMetersPanel*) super;
    Header* header = this->header;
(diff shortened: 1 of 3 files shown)


fasterit:
This is way too much noise for a functionality that could be implemented in approx. three lines of code.
Go do that.

Explorer09:
@lawrence3699 The original report is about a UX (user experience) issue that should be discussed before implementing a code change. The AI surely didn't understand that fact, and the proposed patch messes up more than the real things it fixed.

PR #1973, opened 2026-04-25
CI: Force update package lists in Ubuntu

Explorer09:
There's a recent update of the libcap-dev package (from 2.66-5ubuntu2.2 to 1:2.66-5ubuntu2.4) in Ubuntu 24.04, causing the package list built into the runner image to be out of date, and the build job failure.

Always run `apt-get update` before installing packages to prevent such problem.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,9 @@ jobs:
     steps:
     - uses: actions/checkout@v6
     - name: Install Dependencies
-      run: sudo apt-get install --no-install-recommends libncursesw5-dev
+      run: |
+        sudo apt-get update -q
+        sudo apt-get install --no-install-recommends libncursesw5-dev
     - name: Bootstrap
       run: ./autogen.sh
     - name: Configure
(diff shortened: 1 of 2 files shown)


BenBE:
Something is still broken on FreeBSD …

Explorer09:
> Something is still broken on FreeBSD …

Yes, and I didn't yet know how to fix it...

BenBE:
If I read the log correctly it fails to setup the FreeBSD ports and thus fails to correctly initialize the VM, thus SSH not being able to connect … At least that's how I'd read things …

Explorer09:
> If I read the log correctly it fails to setup the FreeBSD ports and thus fails to correctly initialize the VM, thus SSH not being able to connect … At least that's how I'd read things …

We're not alone on the problem:

https://forums.freebsd.org/threads/pkg-problem.102489/

Explorer09:
Here's another project that experienced the same issue as us (FreeBSD CI broken)

https://github.com/libexpat/libexpat/issues/1207

Explorer09:
The FreeBSD part of the issue seems to have fixed upstream.

PR #1974, opened 2026-04-26
Switch primary timekeeping from struct timeval to struct timespec

natoscott:
The generic gettime implementation already calls clock_gettime() with struct timespec, then converts down to struct timeval — losing nanosecond precision.  This eliminates that wasteful conversion and modernizes the internal time API to use nanosecond precision throughout (struct timespec) instead of microsecond (struct timeval).

Core changes:
- Machine.realtime: struct timeval → struct timespec
- GraphData.time: struct timeval → struct timespec
- Generic_gettime_realtime: now populates struct timespec directly
- timespec_cmp() and timespec_add() helpers in generic/gettime.h

System API boundaries retained as struct timeval:
- BSD platform uptime (gettimeofday + kernel sysctl boottime)
- TraceScreen.c select() timeout
- generic/gettime.c gettimeofday() fallback (no clock_gettime)
- Kernel structs (BSD kinfo_proc boot time, dragonfly busy_time)

### DIFF
--- a/Machine.h
+++ b/Machine.h
@@ -11,7 +11,7 @@ in the source distribution for its full text.
 #include <limits.h>
 #include <stdbool.h>
 #include <stdint.h>
-#include <sys/time.h>
+#include <time.h>
 #include <sys/types.h>
 
 #include "Panel.h"
@@ -38,7 +38,7 @@ typedef unsigned long long int memory_t;
 typedef struct Machine_ {
    struct Settings_* settings;
 
-   struct timeval realtime;   /* time of the current sample */
+   struct timespec realtime;  /* time of the current sample */
    uint64_t realtimeMs;       /* current time in milliseconds */
    uint64_t monotonicMs;      /* same, but from monotonic clock */
    uint64_t prevMonotonicMs;  /* time in milliseconds from monotonic clock of previous scan */
--- a/Meter.c
+++ b/Meter.c
@@ -24,6 +24,7 @@ in the source distribution for its full text.
 #include "Row.h"
 #include "Settings.h"
 #include "XUtils.h"
+#include "generic/gettime.h"
 
 
 #ifndef UINT32_WIDTH
@@ -255,10 +256,10 @@ static void GraphMeterMode_draw(Meter* this, int x, int y, int w) {
 
    // Record new value if necessary
    const Machine* host = this->host;
-   if (!timercmp(&host->realtime, &(data->time), <)) {
+   if (timespec_cmp(&host->realtime, &(data->time)) >= 0) {
       int globalDelay = host->settings->delay;
-      struct timeval delay = { .tv_sec = globalDelay / 10, .tv_usec = (globalDelay % 10) * 100000L };
-      timeradd(&host->realtime, &delay, &(data->time));
+      struct timespec delay = { .tv_sec = globalDelay / 10, .tv_nsec = (globalDelay % 10) * 100000000L };
+      timespec_add(&host->realtime, &delay, &(data->time));
 
       memmove(&data->values[0], &data->values[1], (nValues - 1) * sizeof(*data->values));
 
--- a/ScreenManager.c
+++ b/ScreenManager.c
@@ -123,7 +123,7 @@ static void checkRecalculation(ScreenManager* this, double* oldTime, int* sortTi
    Machine* host = this->host;
 
    Platform_gettime_realtime(&host->realtime, &host->realtimeMs);
-   double newTime = ((double)host->realtime.tv_sec * 10) + ((double)host->realtime.tv_usec / 100000);
+   double newTime = ((double)host->realtime.tv_sec * 10) + ((double)host->realtime.tv_nsec / 100000000L);
 
    *timedOut = (newTime - *oldTime > host->settings->delay);
    *rescan |= *timedOut;
--- a/darwin/Platform.c
+++ b/darwin/Platform.c
@@ -18,6 +18,7 @@ in the source distribution for its full text.
 #include <net/if_types.h>
 #include <net/route.h>
 #include <sys/socket.h>
+#include <sys/time.h>
 #include <mach/port.h>
 #include <net/if.h> // After `sys/socket.h` for struct `sockaddr` (for iOS6 SDK)
 
--- a/generic/gettime.c
+++ b/generic/gettime.c
@@ -10,31 +10,32 @@ in the source distribution for its full text.
 #include "generic/gettime.h"
 
 #include <string.h>
-#include <time.h>
+
+#if !defined(HAVE_CLOCK_GETTIME)
+#include <sys/time.h>
+#endif
 
 
-void Generic_gettime_realtime(struct timeval* tvp, uint64_t* msec) {
+void Generic_gettime_realtime(struct timespec* tsp, uint64_t* msec) {
 
 #if defined(HAVE_CLOCK_GETTIME)
 
-   struct timespec ts;
-   if (clock_gettime(CLOCK_REALTIME, &ts) == 0) {
-      tvp->tv_sec = ts.tv_sec;
-      tvp->tv_usec = (suseconds_t)(ts.tv_nsec / 1000);
-      *msec = ((uint64_t)ts.tv_sec * 1000) + ((uint64_t)ts.tv_nsec / 1000000);
+   if (clock_gettime(CLOCK_REALTIME, tsp) == 0) {
+      *msec = ((uint64_t)tsp->tv_sec * 1000) + ((uint64_t)tsp->tv_nsec / 1000000);
    } else {
-      memset(tvp, 0, sizeof(struct timeval));
+      memset(tsp, 0, sizeof(*tsp));
       *msec = 0;
    }
 
-#else /* lower resolution gettimeofday(2) is always available */
+#else /* lower resolution gettimeofday(2) fallback */
 
    struct timeval tv;
    if (gettimeofday(&tv, NULL) == 0) {
-      *tvp = tv; /* struct copy */
+      tsp->tv_sec = tv.tv_sec;
+      tsp->tv_nsec = (long)tv.tv_usec * 1000L;
       *msec = ((uint64_t)tv.tv_sec * 1000) + ((uint64_t)tv.tv_usec / 1000);
    } else {
-      memset(tvp, 0, sizeof(struct timeval));
+      memset(tsp, 0, sizeof(*tsp));
       *msec = 0;
    }
 
@@ -50,7 +51,7 @@ void Generic_gettime_monotonic(uint64_t* msec) {
    else
       *msec = 0;
 
-#else /* lower resolution gettimeofday() should be always available */
+#else /* lower resolution gettimeofday() fallback */
 
    struct timeval tv;
    if (gettimeofday(&tv, NULL) == 0)
--- a/generic/gettime.h
+++ b/generic/gettime.h
@@ -8,11 +8,28 @@ in the source distribution for its full text.
 */
 
 #include <stdint.h>
-#include <sys/time.h>
+#include <time.h>
 
 
-void Generic_gettime_realtime(struct timeval* tvp, uint64_t* msec);
+void Generic_gettime_realtime(struct timespec* tsp, uint64_t* msec);
 
 void Generic_gettime_monotonic(uint64_t* msec);
+
+static inline int timespec_cmp(const struct timespec* a, const struct timespec* b) {
+   if (a->tv_sec != b->tv_sec)
+      return a->tv_sec < b->tv_sec ? -1 : 1;
+   if (a->tv_nsec != b->tv_nsec)
+      return a->tv_nsec < b->tv_nsec ? -1 : 1;
+   return 0;
+}
+
+static inline void timespec_add(const struct timespec* a, const struct timespec* b, struct timespec* result) {
+   result->tv_sec = a->tv_sec + b->tv_sec;
+   result->tv_nsec = a->tv_nsec + b->tv_nsec;
+   if (result->tv_nsec >= 1000000000L) {
+      result->tv_sec++;
+      result->tv_nsec -= 1000000000L;
+   }
+}
 
 #endif
--- a/pcp/Metric.c
+++ b/pcp/Metric.c
@@ -219,7 +219,7 @@ void Metric_enableThreads(void) {
    pmFreeResult(result);
 }
 
-bool Metric_fetch(struct timeval* timestamp) {
+bool Metric_fetch(struct timespec* timestamp) {
    if (pcp->result) {
       pmFreeResult(pcp->result);
       pcp->result = NULL;
@@ -243,9 +243,9 @@ bool Metric_fetch(struct timeval* timestamp) {
    }
    if (timestamp) {
 #if PMAPI_VERSION >= 3
-      pmtimespecTotimeval(&pcp->result->timestamp, timestamp);
-#else
       *timestamp = pcp->result->timestamp;
+#else
+      pmtimevalTotimespec(&pcp->result->timestamp, timestamp);
 #endif
    }
    return true;
--- a/pcp/PCPMachine.c
+++ b/pcp/PCPMachine.c
@@ -16,7 +16,7 @@ in the source distribution for its full text.
 #include <math.h>
 #include <stdlib.h>
 #include <string.h>
-#include <sys/time.h>
+#include <time.h>
 
 #include "Machine.h"
 #include "Macros.h"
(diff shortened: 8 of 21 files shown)


pcp/Platform.c
```diff
@@ -905,14 +905,14 @@ CommandLineStatus Platform_getLongOption(int opt, ATTR_UNUSED int argc, char** a
...
       /* shift by start offset to stay in lock-step with realtime (archives) */
-      if (pcp->offset.tv_sec || pcp->offset.tv_usec)
-         pmtimevalDec(tv, &pcp->offset);
-      *msec = ((uint64_t)tv->tv_sec * 1000) + ((uint64_t)tv->tv_usec / 1000);
+      if (pcp->offset.tv_sec || pcp->offset.tv_nsec)
+         pmtimespecDec(tv, &pcp->offset);
+      *msec = ((uint64_t)tv->tv_sec * 1000) + ((uint64_t)tv->tv_nsec / 1000000);
    } else {
-      memset(tv, 0, sizeof(struct timeval));
+      memset(tv, 0, sizeof(struct timespec));
```
Explorer09, line 928:
```suggestion
      memset(tv, 0, sizeof(*tv));
```

Machine.h
```diff
@@ -11,8 +11,8 @@ in the source distribution for its full text.
 #include <limits.h>
 #include <stdbool.h>
 #include <stdint.h>
-#include <sys/time.h>
 #include <sys/types.h>
+#include <time.h>
```
Explorer09, line 14: Wrong include order.

Meter.c
```diff
@@ -17,6 +17,7 @@ in the source distribution for its full text.
 #include <string.h>
 
 #include "CRT.h"
+#include "generic/gettime.h"
```
Explorer09: Wrong include order.

ScreenManager.c
```diff
@@ -123,7 +123,7 @@ static void checkRecalculation(ScreenManager* this, double* oldTime, int* sortTi
    Machine* host = this->host;
 
    Platform_gettime_realtime(&host->realtime, &host->realtimeMs);
-   double newTime = ((double)host->realtime.tv_sec * 10) + ((double)host->realtime.tv_usec / 100000);
+   double newTime = ((double)host->realtime.tv_sec * 10) + ((double)host->realtime.tv_nsec / 100000000L);
```
Explorer09, line 126: Since the result is cast to `double`, the `L` suffix for casting the literal doesn't make sense.
```suggestion
   double newTime = ((double)host->realtime.tv_sec * 10) + ((double)host->realtime.tv_nsec / 1e8);
```

pcp/PCPMachine.c
```diff
@@ -381,9 +381,9 @@ Machine* Machine_new(UsersTable* usersTable, uid_t userId) {
 
    Machine_init(super, usersTable, userId);
 
-   struct timeval timestamp;
-   gettimeofday(&timestamp, NULL);
-   this->timestamp = pmtimevalToReal(&timestamp);
+   struct timespec timestamp;
+   pmtimespecNow(&timestamp);
+   this->timestamp = pmtimespecToReal(&timestamp);
```
Explorer09, line 386: Non-existent function (AI hallucination?)

```text
/home/runner/work/htop/htop/pcp/PCPMachine.c:372: undefined reference to `pmtimespecToReal'
```

pcp/PCPMachine.c
```diff
@@ -364,12 +364,12 @@ void Machine_scan(Machine* super) {
...
    Metric_enable(PCP_PROC_SMAPS_SWAPPSS, host->smaps_flag);
 
-   struct timeval timestamp;
+   struct timespec timestamp;
    if (Metric_fetch(&timestamp) != true)
       return;
 
    double sample = host->timestamp;
-   host->timestamp = pmtimevalToReal(&timestamp);
+   host->timestamp = pmtimespecToReal(&timestamp);
```
Explorer09, line 372: Non-existent function (AI hallucination?)

```text
/home/runner/work/htop/htop/pcp/PCPMachine.c:372: undefined reference to `pmtimespecToReal'
```

pcp/Metric.c
```diff
@@ -243,9 +243,10 @@ bool Metric_fetch(struct timeval* timestamp) {
    }
    if (timestamp) {
 #if PMAPI_VERSION >= 3
-      pmtimespecTotimeval(&pcp->result->timestamp, timestamp);
-#else
       *timestamp = pcp->result->timestamp;
+#else
+      timestamp->tv_sec = pcp->result->timestamp.tv_sec;
+      timestamp->tv_nsec = pcp->result->timestamp.tv_usec * 1000L;
```
Explorer09: Any reason not to use pmtimespecTotimeval(3)? I think ideally that function should be declared as inline function in the PCP headers.

generic/gettime.h
```diff
@@ -8,11 +8,25 @@ in the source distribution for its full text.
...
-#include <sys/time.h>
+#include <time.h>
 
 
-void Generic_gettime_realtime(struct timeval* tvp, uint64_t* msec);
+void Generic_gettime_realtime(struct timespec* tsp, uint64_t* msec);
 
 void Generic_gettime_monotonic(uint64_t* msec);
 
+#define timespec_cmp(tsp, usp, cmp) \
```
Explorer09: Just follow the naming from BSD and name it `timespeccmp`?

Also, this should have a `#ifdef timespeccmp` guard.

generic/gettime.h
```diff
@@ -8,11 +8,25 @@ in the source distribution for its full text.
...
+   (((tsp)->tv_sec == (usp)->tv_sec) ? \
+      ((tsp)->tv_nsec cmp (usp)->tv_nsec) : \
+      ((tsp)->tv_sec cmp (usp)->tv_sec))
+
+static inline void timespec_add(const struct timespec* a, const struct timespec* b, struct timespec* result) {
```
Explorer09, line 26: BSD names it `timespecadd`.

Explorer09:
One more thing: AI tagging should use `Assisted-by` tag rather than `Co-authored-by` tag.

It looks like open source projects are slowly standardizing the `Assisted-by` label for AI use (see, for example, [Linux kernel AI policy](https://docs.kernel.org/process/coding-assistants.html)). [The `Co-authored-by` tag looks inappropriate for AI.](https://bence.ferdinandy.com/2025/12/29/dont-abuse-co-authored-by-for-marking-ai-assistance/)

pcp/Platform.c
natoscott: I prefer it as-is.

ScreenManager.c
natoscott: It doesn't hurt and the non-scientific notation is more readable IMO.

pcp/PCPMachine.c
natoscott: No, it has existed since the first version of libpcp.  It must be either missing from headers of older versions of PCP, or we've not included the right headers (no warnings here though, so probably the former).  I'll add a fwd declaration with a note I guess.

pcp/Metric.c
natoscott: No reason not to, will do.

generic/gettime.h
natoscott: Its a non-standard interface, so doing that would just complicate a trivial thing - I prefer it as-is, simpler.

generic/gettime.h
natoscott: I prefer it as-is

generic/gettime.h
Explorer09: When I googled the `timespec_cmp`, I got a different interface definition than the one defined here. And I don't think it's a good idea to let AI copy an existing code without attributing the source.

Yes I know. This macro is copied from a BSD header, token-by-token. While I don't believe this trivial piece would bring in a copyright suit, it still feels not ethical to me.

natoscott:
> [The Co-authored-by tag looks inappropriate for AI.]

What a bizarre rant.  This is Claude's default attribution - I'll leave it as-is.

pcp/PCPMachine.c
natoscott: And again, that assessment is incorrect.

generic/gettime.h
Explorer09: See the copying issue of `timespeccmp` I mentioned in another comment.

Speaking of, the `timespec_add` function from Gnulib comes with a different interface:
https://github.com/coreutils/gnulib/blob/master/lib/timespec-add.c

generic/gettime.h
natoscott: > This macro is copied from a BSD header, token-by-token.

That's not how LLM training works.

> [...] copyright suit [...]

Completely and utterly absurd to even suggest this.

> [...] it still feels not ethical to me

Are you saying that I'm behaving unethically using this AI-generated code in a PR?

generic/gettime.h
natoscott: Your "copying issue" is nonsense and we don't use gnulib in htop (and are unlikely to ever do so), so this is also irrelevant.

generic/gettime.h
Explorer09: > > This macro is copied from a BSD header, token-by-token.
> 
> That's not how LLM training works.
> 

I've found the source. It's here:
https://github.com/openbsd/src/blob/master/sys/sys/time.h#L117
It's token-by-token copying. Even the parameter names are exact.

In case you want to argue that LLM does not keep copyrighted data, well here is a paper saying otherwise:
https://arxiv.org/html/2601.02671v1

In fact I don't like to accuse you for breaking copyright law or something, but suggesting that LLM is all-ethical crossed the line for me.

> 
> > [...] it still feels not ethical to me
> 
> Are you saying that I'm behaving unethically using this AI-generated code in a PR?

Partially. I know it's impossible to ban all LLM contribution, but I do wish the copyright impact for LLM-generated code to be minimized. Just because the code is publicly available doesn't mean it's public domain code. There's another legal defense called "de minimis". That means code that's too small and trivial could be okay. I would keep caution on any LLM-generated code, anyway.

pcp/PCPMachine.c
Explorer09: I stand corrected. (But the CI did report the error, so the header includes probably need to be fixed.)

generic/gettime.h
natoscott: > suggesting that LLM is all-ethical crossed the line for me.

Noone suggested anything of the sort - you are making up comments to suit your argument.

> Partially.

Wow.

> copyright impact for LLM-generated code to be minimized

There is absolutely no cause to think there's any copyright issue in this trivial code added here - you are making a strawman argument here to suit this axe you have to grind against AI use.  Just stop.

pcp/PCPMachine.c
natoscott: My mistake - the libpcp exports file was sanitised with the soname 4 transition - so it looks like this function was there forever, but older (soname 3) versions have a different exports file and did not have this.  I'll get Claude to refactor it.

generic/gettime.h
```diff
@@ -8,11 +8,25 @@ in the source distribution for its full text.
...
 
-void Generic_gettime_realtime(struct timeval* tvp, uint64_t* msec);
+void Generic_gettime_realtime(struct timespec* tsp, uint64_t* msec);
 
 void Generic_gettime_monotonic(uint64_t* msec);
 
+#define timespec_cmp(tsp, usp, cmp) \
+   (((tsp)->tv_sec == (usp)->tv_sec) ? \
+      ((tsp)->tv_nsec cmp (usp)->tv_nsec) : \
+      ((tsp)->tv_sec cmp (usp)->tv_sec))
```
BenBE: Unless there's a good reason to avoid strict typing, I'd suggest to use a static inline function instead of a macro here.

pcp/Metric.c
```diff
@@ -243,9 +243,9 @@ bool Metric_fetch(struct timeval* timestamp) {
    }
    if (timestamp) {
 #if PMAPI_VERSION >= 3
-      pmtimespecTotimeval(&pcp->result->timestamp, timestamp);
-#else
       *timestamp = pcp->result->timestamp;
+#else
+      pmtimevalTotimespec(&pcp->result->timestamp, timestamp);
 #endif
```
BenBE, line 249: Is this inversion of logic intended? So if I read this correct, `pmtimevalTotimespec` is the old API and `pcp->result->timestamp` is the new one?

pcp/PCPMachine.c
BenBE: Grok's razor: Don't attribute to AI, what can be explained by broken code alone … ;-)

Machine.h
BenBE: No, `<time.h>` before `<sys/types.h>` (parent before subdir) is completely right …

darwin/Platform.c
```diff
@@ -18,6 +18,7 @@ in the source distribution for its full text.
 #include <net/if_types.h>
 #include <net/route.h>
 #include <sys/socket.h>
+#include <sys/time.h>
```
BenBE, line 21: Just for clarification: For pulling in which declaration is this include? As there's no further change in this file. Is this a stray edit by chance?

generic/gettime.c
```diff
@@ -10,31 +10,29 @@ in the source distribution for its full text.
 #include "generic/gettime.h"
 
 #include <string.h>
-#include <time.h>
+#include <sys/time.h>
```
BenBE, line 15: Which type is used from `<sys/time.h>`? AFAICT, both `timespec` and `clock_gettime` should come from `<time.h>`. Or am I missing something?

If this is for `struct timeval`, maybe 
```suggestion

#if !defined(HAVE_CLOCK_GETTIME)
#include <sys/time.h>
#endif
```

generic/gettime.h
BenBE: Ah, oops. saw the use of `cmp` too late.

pcp/Metric.c
natoscott: The pmResult timestamps changed to nanosecond-precision (timespec) in PMAPI_VERSION >= 3 so we can just pull the value directly.  In the older version we have a timeval, so need to convert it to a timespec via that function call.

generic/gettime.c
natoscott: LOL - yes its timeval - and that suggestion is exactly what Claude originally came up with.  I edited it for simplicity to just directly include sys/time.h without conditional, but I can put that back if you prefer?

generic/gettime.h
natoscott: I agree re inline - and Claude originally wrote it that way.  I changed it back to the macro to match what we were doing with timeval to minimise this change.  I'd be happy to ditch this macro - there's only one call site, this could easily be coded as an inline without using the cpp macro trickey with the comparison operator.

Or do you strongly feel we should stick with the macro @BenBE ?

generic/gettime.c
BenBE: Fine either way, but having the #ifdef there gives some hints about why this change was made (especially as this include was effectively swapped with the header; which was what originally made me look into this).

natoscott: Done.  Also went ahead with the macro -> inline function switcheroo and dropped the double negative at the one call site in Meter.c.

generic/gettime.h
Explorer09: @BenBE The `cmp` is an operator so this has to be implemented as a macro. However, if you want an inline function it's still possible by an alternative prototype. Look at [timespeccompare(3) in Solaris](https://docs.oracle.com/cd/E88353_01/html/E37843/timespeccompare-3c.html) to see what I mean (it can return `<0` for less-than, `>0` for greater than, and `==0` for equal).

pcp/PCPMachine.c
Explorer09: @BenBE I don't want to give credit to AI contribution (it's a robot after all, not a person). If I make a false claim against a robot then you can just point that out. Since, technically, the submitter is the one responsible with the code, not the AI.

BenBE: We are all human here and humans are (last time I checked¹) fallible …

And in the end I don't care if a human or a machine broke the code. What I care about is that the code gets fixed. And if there's a shitload of AI slop involved that makes this happen is secondary, as long as the overall quality of the result is acceptable. Minimum qualifying criterion is that it compiles. Further criteria as well as terms&conditions may apply.

¹May not apply to you; haven't verified that part …

darwin/Platform.c
natoscott: The Darwin uptime calculation uses gettimeofday() and struct timeval to compute uptime from boot time. Previously, sys/time.h was pulled in transitively through generic/gettime.h.  Now that we're switching gettime.h to struct timespec, that transitive include went away, so darwin/Platform.c needs to include it directly.

PR #1977, opened 2026-05-01
linux: avoid repeated zram ENOENT probes

Komzpa:
## Summary

This changes Linux zram accounting to enumerate existing `/sys/block/zramN` devices instead of probing `zram0`, `zram1`, and so on until a file open fails.

On systems without zram, htop otherwise repeats failed sysfs lookups for `/sys/block/zram0/disksize` and `/sys/block/zram0/mm_stat` on refresh. Enumerating `/sys/block` avoids that repeated `ENOENT` path and also naturally handles non-contiguous zram numbering.

The implementation follows the existing Linux compat style: each enumerated zram directory is opened relative to the `/sys/block` directory using the openat-style path, and `disksize` / `mm_stat` are read with `Compat_readfileat` rather than `fopen`.

## Validation

- `git diff --check`
- `./autogen.sh && ./configure && make -j2`

### DIFF
--- a/linux/LinuxMachine.c
+++ b/linux/LinuxMachine.c
@@ -274,47 +274,75 @@ static void LinuxMachine_scanHugePages(LinuxMachine* this) {
    closedir(dir);
 }
 
+static bool LinuxMachine_isZramBlockName(const char* name) {
+   if (!String_startsWith(name, "zram"))
+      return false;
+
+   name += strlen("zram");
+   if (*name == '\0')
+      return false;
+
+   for (; *name != '\0'; name++) {
+      if (*name < '0' || *name > '9')
+         return false;
+   }
+
+   return true;
+}
+
+static void LinuxMachine_scanZramDevice(openat_arg_t blockDirFd, const char* name, memory_t* totalZram, memory_t* usedZramComp, memory_t* usedZramOrig) {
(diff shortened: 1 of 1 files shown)


BenBE:
Please avoid the `fopen` API (they cause trouble in high-memory-pressure situations.

Also, given we are already enumerating a directory using its fd, try to use the `openat` style APIs that help the kernel with resolving paths quicker.

Also take a look at the `Compat_readfileat` function to read the entire file contents at once.

Regarding building: Try to `./autogen.sh && ./configure && make`, which should give a good starting point to check your changes.

Komzpa:
Thanks, fixed. The zram scan no longer uses `fopen` for the sysfs files: it opens each enumerated zram directory relative to `/sys/block` using the openat-style compat path and reads `disksize`/`mm_stat` with `Compat_readfileat`.

I also reran the requested build path successfully: `./autogen.sh && ./configure && make -j2`

eworm-de:
This breaks completely for me... With commit 9fe974b823207e0f19a7bc4ae9cc31eba6e6cd47 I do not have any values in my zram meter.

eworm-de:
Opened this as fix: https://github.com/htop-dev/htop/pull/2045

PR #1980, opened 2026-05-02
SELinuxMeter: improve SELinux detection in constrained environments

Theldus:
This PR proposes a third 'enabled' state for the SELinuxMeter:
```txt
    enabled; mode: unknown
```
and the reasoning is simple: in very constrained environments (such as [Android]), its not possible to read `/sys/fs/selinux/enforce` to gauge the current mode, but it is still possible to confirm SELinux is enabled. Today htop returns `disabled` in that case, which is wrong; with this change it returns what it could actually determine.

The patch drops the `/etc/selinux/config` existence check (libselinux itself skips this check under `#ifdef ANDROID`) and maps read/parse failures of `/sys/fs/selinux/enforce` to the new unknown state instead of silently rendering them as permissive.
                                                                                                                                                                                         
This behavior is similar to other meters such as Uptime, which also returns (unknown) when it can't read the underlying data. 

[Android]: https://source.android.com/docs/security/features/selinux#background

### DIFF
--- a/linux/SELinuxMeter.c
+++ b/linux/SELinuxMeter.c
@@ -11,6 +11,7 @@ in the source distribution for its full text.
 
 #include "CRT.h"
 
+#include <errno.h>
 #include <stdbool.h>
 #include <stdint.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


fasterit (approved):
LGTM, thank you @Theldus

PR #1978, opened 2026-05-02
Fix building htop for TCC compiler and experiments for Gentoo users

GermanAizek:
My experiments on 2 CPUs Xeon E5-2699v3 (72 threads):

Wow! I am surprised by the compilation speed on TCC, as well as the resource consumption on monitoring
- TCC 3s compilation
- Clang 12s compilation
- GCC 15s compialtion

Building with -DNDEBUG macro by default.

I think Gentoo users will idolize Fabrice Bellard.

Install `tcc` compiler easy `apt install tcc`

Wiki: https://en.wikipedia.org/wiki/Tiny_C_Compiler
Official site: https://bellard.org/tcc/

tcc_htop
clang_htop
gcc_htop

PR #1981, opened 2026-05-03
build: Simplify __builtin_ctz check code

Explorer09:
Follow-up of #1978.

@GermanAizek I believe this check code also works with TCC without a false positive, but I would like you to help me test it out.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -352,7 +352,10 @@ AC_MSG_CHECKING(for __builtin_ctz)
 AC_LINK_IFELSE(
    [AC_LANG_PROGRAM(
       [],
-      [[volatile int test_ctz = __builtin_ctz(1); (void)test_ctz; /* Supported in GCC 3.4 or later */]]
+      [[
+         /* Supported in GCC 3.4 or later */
+         return __builtin_ctz(1U); /* Should be 0 (exit success) */
+      ]]
    )],
    AC_DEFINE([HAVE_BUILTIN_CTZ], 1, [Define to 1 if the compiler supports '__builtin_ctz' function.])
    AC_MSG_RESULT(yes),


BenBE:
It's kinda non-obvious on first glance that this program would return exit status 0, unless you know the sematics of CTZ. Thus I'm not sure this actually is an improvement.

configure.ac
```diff
@@ -352,7 +352,10 @@ AC_MSG_CHECKING(for __builtin_ctz)
 AC_LINK_IFELSE(
    [AC_LANG_PROGRAM(
       [],
-      [[volatile int test_ctz = __builtin_ctz(1); (void)test_ctz; /* Supported in GCC 3.4 or later */]]
+      [[
+         /* Supported in GCC 3.4 or later */
+         return __builtin_ctz(1U); /* Exit success */
```
BenBE: The full explanation would be this:
```suggestion
         return __builtin_ctz(1U); /* CTZ(1) == 0 -> Exit success */
```

Explorer09:
> It's kinda non-obvious on first glance that this program would return exit status 0, unless you know the sematics of CTZ.

I slightly tweaked the style to add a comment that the exit status of 0 is intended.

By the time you ask this question, you should have looked up the semantics of CTZ and can tell that `__builtin_ctz(1) == 0`.

It's actually a sanity check to ensure the behavior of `__builtin_ctz()` is correct (the test program will be compiled but not run, but if someone does run `./conftest` they will get a exit status of 0).

BenBE:
With these semantics we could actually even run that test proper. But that would complicate cross-compilation if you can't run the target platform. Thus link is sufficient.

configure.ac
```diff
@@ -352,7 +352,10 @@ AC_MSG_CHECKING(for __builtin_ctz)
 AC_LINK_IFELSE(
    [AC_LANG_PROGRAM(
       [],
-      [[volatile int test_ctz = __builtin_ctz(1); (void)test_ctz; /* Supported in GCC 3.4 or later */]]
+      [[
+         /* Supported in GCC 3.4 or later */
+         return __builtin_ctz(1U); /* Should be 0 (exit success */
```
BenBE:
```suggestion
         return __builtin_ctz(1U); /* Should be 0 (exit success) */
```

Explorer09:
Minor note: In another branch of mine I would be adding configure checks for `__builtin_clz` and `__builtin_popcount`, and I wish the styles of test codes be consistent. (I would use `__builtin_clz(-1U)` and `__builtin_popcount(0U)` that both return 0.)

BenBE (approved):

PR #1984, opened 2026-05-05
Fix drawTab bounds handling and improve rendering safety

vatsanant:
This change improves the safety and clarity of drawTab() while preserving its behavior.

Key changes:

- Reworked bounds handling using a `remainingWidth` variable instead of repeated `*x >= l` checks
  - Simplifies reasoning about available space
  - Ensures no writes occur past the screen width
  
- Prevent potential overflow when applying `SCREEN_TAB_COLUMN_GAP`
  - The gap is now clamped to remaining width
  - Previously, `*x += 1 + gap` could advance beyond the limit

- Removed unnecessary pointer usage for `y`
  - `y` is not modified, so passing by value simplifies the API

- Extracted color selection into helper functions
  - Reduces duplication and improves readability

- Added basic null checks for `x` and `name`

Behavioral notes:

- Rendering is unchanged in normal cases
- Near the right boundary, tabs are now safely clipped instead of potentially exceeding bounds

This is primarily a correctness and maintainability improvement with no intended visual changes.

### DIFF
--- a/ScreenManager.c
+++ b/ScreenManager.c
@@ -168,27 +168,57 @@ static void checkRecalculation(ScreenManager* this, double* oldTime, int* sortTi
    *rescan = false;
 }
 
-static inline bool drawTab(const int* y, int* x, int l, const char* name, bool cur) {
-   assert(*x >= 0);
-   assert(*x < l);
+static inline int getBorderColor(bool cur) {
+   return CRT_colors[cur ? SCREENS_CUR_BORDER : SCREENS_OTH_BORDER];
+}
 
-   attrset(CRT_colors[cur ? SCREENS_CUR_BORDER : SCREENS_OTH_BORDER]);
-   mvaddch(*y, *x, '[');
+static inline int getTextColor(bool cur) {
+   return CRT_colors[cur ? SCREENS_CUR_TEXT : SCREENS_OTH_TEXT];
+}
+
+static inline bool drawTab(int y, int* x, int maxWidth, const char* name, bool cur) {
+
+   assert( x != NULL);
+   assert(name != NULL); 
+   assert(*x >= 0 && *x < maxWidth);
+
+   int remainingWidth = maxWidth - *x;
+
+   //drwa '['
+   attrset(getBorderColor(cur));
+   mvaddch(y, *x, '[');
    (*x)++;
-   if (*x >= l)
-      return false;
-   int nameWidth = (int)strnlen(name, l - *x);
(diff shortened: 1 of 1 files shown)


Explorer09:
> * `y` is not modified, so passing by value simplifies the API

Please revert this part. The compiler can optimize this well and it's not necessary for you to do the micro-optimization.

> * Extracted color selection into helper functions

Please revert this part, too. The two helper functions increased cognitive load without any other benefit. You should just add one-line comments to explain what they do instead.

ScreenManager.c
```diff
@@ -168,27 +168,57 @@ static void checkRecalculation(ScreenManager* this, double* oldTime, int* sortTi
...
-   *x += 1 + SCREEN_TAB_COLUMN_GAP;
-   if (*x >= l)
+   } 
+
+   //draw ']'
+   attrset(getBorderColor(cur));
+   mvaddch(y, *x, ']');
+   (*x)++;
+   remainingWidth--;
+   if (remainingWidth <= 0){
```
Explorer09, line 211: How about just add this:
```c
if (*x >= l - 1 - SCREEN_TAB_COLUMN_GAP)
   return false;
```

before the `*x += 1 + SCREEN_TAB_COLUMN_GAP;` line, and you can save all the trouble maintaining an extra `remainingWidth` variable?

ScreenManager.c
```diff
@@ -168,27 +168,57 @@ static void checkRecalculation(ScreenManager* this, double* oldTime, int* sortTi
...
-   assert(*x < l);
+static inline int getBorderColor(bool cur) {
+   return CRT_colors[cur ? SCREENS_CUR_BORDER : SCREENS_OTH_BORDER];
+}
+
+static inline int getTextColor(bool cur) {
+   return CRT_colors[cur ? SCREENS_CUR_TEXT : SCREENS_OTH_TEXT];
+}
+
+static inline bool drawTab(int y, int* x, int maxWidth, const char* name, bool cur) {
```
Explorer09, line 179: The `l` (lowercase L) parameter in the original code does not refer to maximum "width". It's more appropriate to name the parameter "xMax" or "xEnd", depending on your preference.

fasterit:
This is replacing good code with different code (that would still have to be made styleguide compliant).
We don't do that to keep the noise level low(er) and make life of distro packagers a bit easier (backporting relevant patches).

If you want to do a minimal change to achieve "Near the right boundary, tabs are now safely clipped instead of potentially exceeding bounds" please propose a separate PR. This one will go nowhere.

PR #1986, opened 2026-05-06
Extend the keybindings to pan the main tree panel to support urxvt

fasterit:


fasterit:
Tested positively in https://github.com/htop-dev/htop/issues/1905#issuecomment-4392137550
Merging...

PR #1993, opened 2026-05-10
Remember meter positions across layout changes

### LINKED ISSUE #1941: HeaderLayout UX: Remember the meters' original column numbers until exit
This can be a UX improvement to the "Header layout" settings in the Setup screen.

1. Start htop with the default htop settings. There should be 2 meter columns.
2. Go to Setup -> Header layout
3. Change the layout to "1 column - full width"
4. Change the layout back to "2 columns - 50/50 (default)"
5. Observe that the meters that originally reside in the second column move to the first column, and they don't move back.

Expected result: The meters' original column numbers are remembered, and move back to their original column once the HeaderLayout has enough number of columns to do so. The original column numbers do not need to be remembered forever, but they should be remembered until any meter is moved in the "Meters" setting, or until htop exits.

https://github.com/user-attachments/assets/19e8f149-b63e-4603-888f-b94a75d19ab5

ugolnikovE:
Fixes #1941

Based on @Explorer09's suggestion: instead of collapsing meters into the last visible column when the layout is reduced, keep all columns allocated in memory. The rendering layer merges hidden columns visually. Switching back to the original layout restores meter positions.

Hidden columns are permanently collapsed if the user modifies meters in the editor or exits htop.

## Summary by CodeRabbit

* **Bug Fixes**
  * Temporary header/column reorganizations are now reliably undone, preventing duplicated or hidden meters.
  * Header layout consistently collapses after adds/removes, avoiding stale header state.
  * Moving or deleting meters only triggers layout recalculation when the meter list actually changes, reducing unnecessary redraws.

* **Refactor**
  * Internal header and panel behaviors reorganized for more predictable layout and serialization.

[Review Change Stack](https://app.coderabbit.ai/change-stack/htop-dev/htop/pull/1993)

### DIFF
--- a/CategoriesPanel.c
+++ b/CategoriesPanel.c
@@ -36,6 +36,10 @@ static const char* const CategoriesFunctions[] = {"      ", "      ", "      ", 
 
 static void CategoriesPanel_delete(Object* object) {
    CategoriesPanel* this = (CategoriesPanel*) object;
+   if (this->header->metersCopied) {
+      Header_undoMetersCopy(this->header);
+      this->header->metersCopied = false;
+   }
    Panel_done(&this->super);
    free(this);
 }
@@ -44,6 +48,20 @@ static void CategoriesPanel_makeMetersPage(CategoriesPanel* this) {
    size_t columns = HeaderLayout_getColumns(this->scr->header->headerLayout);
    MetersPanel** meterPanels = xMallocArray(columns, sizeof(MetersPanel*));
    Settings* settings = this->host->settings;
+
+   size_t hiddenMeterCount = 0;
+   for (size_t i = columns; i < this->header->maxColumns; i++) {
+      hiddenMeterCount += Vector_size(this->header->columns[i]);
+   }
+
+   if (hiddenMeterCount > 0) {
+      this->header->columns[columns - 1]->owner = false;
+      for (size_t i = columns; i < this->header->maxColumns; i++) {
+         Vector_splice(this->header->columns[columns - 1], this->header->columns[i]);
+      }
+      this->header->columns[columns - 1]->owner = true;
+   }
+   this->header->metersCopied = hiddenMeterCount > 0;
 
    for (size_t i = 0; i < columns; i++) {
       char titleBuffer[32];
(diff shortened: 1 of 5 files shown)


CategoriesPanel.c
```diff
@@ -45,6 +47,15 @@ static void CategoriesPanel_makeMetersPage(CategoriesPanel* this) {
...
    Settings* settings = this->host->settings;
 
+   bool didCopy = false;
+   for (size_t i = columns; i < this->header->maxColumns; i++) {
+      for (int j = 0; j < Vector_size(this->header->columns[i]); j++) {
+         Vector_add(this->header->columns[columns - 1], Vector_get(this->header->columns[i], j));
+         didCopy = true;
+      }
+   }
+   this->header->metersCopied = didCopy;
```
BenBE: Repeatedly assigning `didCopy` here feels a bit wasteful, especially if we could get the same value by simply or'ing the `Vector_size` of each column instead. Also given `Vector_add` may repeatedly resize the backing array it might be preferable if we could somehow forward the needed amount of memory and reserve a properly sized array beforehand. Not much of an performance issue in practice probably, but at least feels cleaner.

Header.c
```diff
@@ -42,7 +46,7 @@ Header* Header_new(Machine* host, HeaderLayout hLayout) {
 }
 
 void Header_delete(Header* this) {
-   Header_forEachColumn(this, i) {
+   for (size_t i = 0; i < this->maxColumns; i++) {
```
BenBE, line 49: Why this change?

MetersPanel.c
```diff
@@ -10,6 +10,7 @@ in the source distribution for its full text.
 
 #include "MetersPanel.h"
 
+#include <stdbool.h>
```
BenBE: Already included via `MetersPanel.h`

BenBE (commented):
Interesting approach, but something doesn't quite feel right yet. Haven't looked into it too deeply though, thus mostly first impression rn.

MetersPanel.c
ugolnikovE: Fixed

Header.c
ugolnikovE: `Header_forEachColumn` only iterates over visible columns. Since hidden columns now stay allocated in memory, `Header_delete` needs to free all of them up to `maxColumns`

CategoriesPanel.c
ugolnikovE: Thanks, good point. Replaced `didCopy` with a pre-count of hidden meters and switched from manual `Vector_add` loop to `Vector_splice`. I considered dropping the count loop entirely (just checking `columns < maxColumns`), but kept it for clarity. It makes the intent explicit and avoids unnecessary owner flag toggling when there's nothing to copy

ugolnikovE:
@BenBE applied your suggestions (Vector_splice, preallocation, didCopy cleanup). One design note: I kept the current behavior on meter style changes - they change rendering, not layout position, so the collapse path stays bound to position-affecting actions only. Ready for another look when you have time.

PR #2000, opened 2026-05-17
helpers: fix kernel version parsing with error check

GermanAizek:
using sscanf without checking whether all expected values have been parsed successfully can lead to undefined behavior (UB). It is worth checking the sscanf return value here to make sure that all three numbers are received before using the results

Reference: https://stackoverflow.com/questions/61066625/sscanf-check-doesnt-work-when-given-more-input-than-required

## Changes
Fixed kernel version parsing in `darwin/PlatformHelpers.c` by adding validation of the `sscanf` return value. The `Platform_GetKernelVersion` function now checks that all three version components (major, minor, patch) were successfully parsed before using them. If `sscanf` returns fewer than 3 matched values, the function resets the parsed fields to -1, preventing undefined behavior from partially-initialized data.

## Implementation Quality
The fix is narrow and focused. It leverages existing error handling infrastructure (the function already initializes the version to -1 on syscall failure) and applies the same pattern consistently. The additional validation adds 3 lines of logic (capture return value, conditional check, and reset on failure).

## Commit Structure
Single, focused commit addressing a specific correctness issue. The change scope matches the PR objective precisely.

[Review Change Stack](https://app.coderabbit.ai/change-stack/htop-dev/htop/pull/2000?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/darwin/PlatformHelpers.c
+++ b/darwin/PlatformHelpers.c
@@ -30,7 +30,10 @@ void Platform_GetKernelVersion(KernelVersion* k) {
       size_t size = sizeof(str);
       int ret = sysctlbyname("kern.osrelease", str, &size, NULL, 0);
       if (ret == 0) {
-         sscanf(str, "%hd.%hd.%hd", &cachedKernelVersion.major, &cachedKernelVersion.minor, &cachedKernelVersion.patch);
+         int n = sscanf(str, "%hd.%hd.%hd", &cachedKernelVersion.major, &cachedKernelVersion.minor, &cachedKernelVersion.patch);
+         if (n != 3) {
+            cachedKernelVersion.major = cachedKernelVersion.minor = cachedKernelVersion.patch = -1;
+         }
       }
    }
    memcpy(k, &cachedKernelVersion, sizeof(cachedKernelVersion));


darwin/PlatformHelpers.c
```diff
@@ -30,7 +30,10 @@ void Platform_GetKernelVersion(KernelVersion* k) {
       size_t size = sizeof(str);
       int ret = sysctlbyname("kern.osrelease", str, &size, NULL, 0);
       if (ret == 0) {
-         sscanf(str, "%hd.%hd.%hd", &cachedKernelVersion.major, &cachedKernelVersion.minor, &cachedKernelVersion.patch);
+         int n = sscanf(str, "%hd.%hd.%hd", &cachedKernelVersion.major, &cachedKernelVersion.minor, &cachedKernelVersion.patch);
+         if (n != 3) {
+            cachedKernelVersion.major = cachedKernelVersion.minor = cachedKernelVersion.patch = -1;
```
natoscott, line 35: In theory, we could see n take values 1 or 2 here, e.g. if major and minor are present in the string without a patch level.  Is it worth catering for that if we're adding this extra checking anyway?

darwin/PlatformHelpers.c
```diff
@@ -30,7 +30,10 @@ void Platform_GetKernelVersion(KernelVersion* k) {
...
+         }
```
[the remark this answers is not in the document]
BenBE: @coderabbitai Omitting braces here is possible, but not required.

But what should be avoided is multiple assignments on the same line.

darwin/PlatformHelpers.c
BenBE: While I doubt there'd be versions with just the major version in them, I see a case for just having major.minor present. But even just major.minor is some serious hacking in the system I'd figure, that I doubt its worth addressing separately …

natoscott: Agreed - I also wonder if this proposed code path can ever be traversed, since success from sysctlbyname presumably indicates the presence of a well-formed string.  In which case - can sscanf fail?  I suspect not.

BenBE: Likely only for non-release versions or versions altered by external tooling (dtrace hugging the process).

So yes, there are ways to reach that code path, but for all of them simply ignoring the returned string is likely the better way to go.

@coderabbitai Are there legit cases for the version string returned by `sysctlbyname` (on success) for the Darwin kernel version to return a string not shaped as `%d.%d.%d`?

BenBE: An alternative would be a struct initializer `cachedKernelVersion = {-1, -1, -1};` to keep it short.

PR #2001, opened 2026-05-18
Fix security issues reported by AFINE Team

natoscott:
## Summary

- Cap VLA length in `RichString_writeFromWide()` and `RichString_appendnWideColumns()` to prevent stack exhaustion from processes with very long command lines (GHSA-mjc3-mc44-c23f)
- Guard title truncation in `InfoScreen_drawTitled()` against narrow terminal widths to prevent stack buffer underflow (GHSA-6f53-xw2h-ww59)
- Strip control characters from info screen titles before passing to `mvaddstr()` to prevent terminal escape sequence injection (GHSA-q64m-h5hc-c4qq)
- Add `Char_isControl()` and `String_stripControlChars()` helpers to XUtils, and use `Char_isControl()` in non-wide RichString rendering to also filter DEL (0x7F)

All three issues were reported by Michał Majchrowicz and Marcin Wyczechowski of the AFINE Team.

## Summary

Addresses three GHSA security issues (GHSA-mjc3-mc44-c23f, GHSA-6f53-xw2h-ww59, GHSA-q64m-h5hc-c4qq) via targeted fixes to VLA handling, buffer bounds checking, and control character sanitization.

## Changes

**InfoScreen.c** (+3/-1): Adds bounds check `COLS >= 3` alongside length check before truncating with trailing dots, preventing stack buffer underflow when terminal width is very narrow. Introduces call to `String_stripControlChars()` to sanitize title strings before passing to `mvaddstr()`, blocking terminal escape sequence injection.

**RichString.c** (+9/-1): Introduces `RICHSTRING_MAX_WIDE_LEN` constant (0x8000 / 32KB) and caps VLA allocations in `RichString_writeFromWide()` and `RichString_appendnWideColumns()` to prevent stack exhaustion from processes with extremely long command lines. Replaces simple `>= 32` printable-byte check with `Char_isControl()` predicate in non-wide rendering path to also filter DEL character (0x7F), maintaining consistency with wide-character path using `iswprint()`.

**XUtils.h** (+13/-0): Adds two static inline helpers: `Char_isControl()` classifying control characters as those below space or equal to 0x7F, and `String_stripControlChars()` iterating through mutable strings to replace detected control characters with '?'.

## Assessment

Implementation is clean and logically organized. All three fixes are properly isolated by concern: VLA bounds in RichString for stack exhaustion, width validation in InfoScreen for buffer underflow, and control character filtering for injection prevention. Helper functions are positioned inline in headers for efficiency. Bounds enforcement uses reasonable limits (32KB for wide-character operations) and consistent filtering logic across code paths.

[Review Change Stack](https://app.coderabbit.ai/change-stack/htop-dev/htop/pull/2001?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/RichString.c
+++ b/RichString.c
@@ -197,7 +197,7 @@ static inline int RichString_writeFromWide(RichString* this, int attrs, const ch
    size_t newLen = from + len;
    RichString_setLen(this, newLen);
    for (size_t i = from, j = 0; i < newLen; i++, j++) {
-      this->chptr[i] = (((unsigned char)data_c[j]) >= 32 ? ((unsigned char)data_c[j]) : '?') | attrs;
+      this->chptr[i] = (Char_isControl(data_c[j]) ? '?' : (unsigned char)data_c[j]) | attrs;
    }
    this->chptr[newLen] = 0;
 
--- a/RichString.c
+++ b/RichString.c
@@ -21,6 +21,8 @@ in the source distribution for its full text.
 
 #define charBytes(n) (sizeof(CharType) * (n))
 
+#define RICHSTRING_MAX_WIDE_LEN 0x8000
+
 static void RichString_extendLen(RichString* this, size_t len) {
    if (this->chptr == this->chstr) {
       // String is in internal buffer
@@ -101,6 +103,9 @@ static inline int RichString_writeFromWide(RichString* this, int attrs, const ch
    if (len < 1)
       return 0;
 
+   if (len > RICHSTRING_MAX_WIDE_LEN)
+      len = RICHSTRING_MAX_WIDE_LEN;
+
    wchar_t data[len];
    len = mbstowcs_nonfatal(data, data_c, len);
    if (len <= 0)
@@ -116,6 +121,9 @@ static inline int RichString_writeFromWide(RichString* this, int attrs, const ch
 }
 
 int RichString_appendnWideColumns(RichString* this, int attrs, const char* data_c, size_t len, int* columns) {
+   if (len > RICHSTRING_MAX_WIDE_LEN)
+      len = RICHSTRING_MAX_WIDE_LEN;
+
    wchar_t data[len];
    len = mbstowcs_nonfatal(data, data_c, len);
    if (len <= 0)
--- a/XUtils.h
+++ b/XUtils.h
@@ -131,6 +131,19 @@ static inline ssize_t full_write_str(int fd, const char* str) {
    return full_write(fd, str, strlen(str));
 }
 
+static inline bool Char_isControl(char c) {
+   return (unsigned char)c < ' ' || c == '\x7F';
+}
+
+/* Replace control characters (C0 and DEL) with a safe substitute. */
+ATTR_NONNULL
+static inline void String_stripControlChars(char* s) {
+   for (; *s; s++) {
+      if (Char_isControl(*s))
+         *s = '?';
+   }
+}
+
 /* Compares floating point values for ordering data entries. In this function,
    NaN is considered "less than" any other floating point value (regardless of
    sign), and two NaNs are considered "equal" regardless of payload. */
--- a/InfoScreen.c
+++ b/InfoScreen.c
@@ -59,6 +59,8 @@ void InfoScreen_drawTitled(InfoScreen* this, const char* fmt, ...) {
       memset(&title[COLS - 3], '.', 3);
    }
 
+   String_stripControlChars(title);
+
    attrset(CRT_colors[METER_TEXT]);
    mvhline(0, 0, ' ', COLS);
    mvaddstr(0, 0, title);
--- a/InfoScreen.c
+++ b/InfoScreen.c
@@ -55,7 +55,7 @@ void InfoScreen_drawTitled(InfoScreen* this, const char* fmt, ...) {
    int len = vsnprintf(title, sizeof(title), fmt, ap);
    va_end(ap);
 
-   if (len > COLS) {
+   if (len > COLS && COLS >= 3) {
       memset(&title[COLS - 3], '.', 3);
    }
 


XUtils.h
```diff
@@ -131,6 +131,19 @@ static inline ssize_t full_write_str(int fd, const char* str) {
...
 
+static inline bool Char_isControl(char c) {
+   return (unsigned char)c < ' ' || c == '\x7F';
+}
+
+/* Replace control characters (C0 and DEL) with a safe substitute. */
+ATTR_NONNULL
+static inline void String_stripControlChars(char* s) {
+   for (; *s; s++) {
+      if (Char_isControl(*s))
```
Explorer09, line 142: Why having a `Char_isControl` when it's more appropriate to just use `!isprint(*s)`?

Update: If we actually need this, the function name should be `isAsciiPrint` and add a comment line on why we don't use `isprint` from standard library `ctype.h`. (Locale-dependency issue, or for code size or speed reasons?)

XUtils.h
```diff
@@ -131,6 +131,19 @@ static inline ssize_t full_write_str(int fd, const char* str) {
    return full_write(fd, str, strlen(str));
 }
 
+static inline bool Char_isControl(char c) {
+   return (unsigned char)c < ' ' || c == '\x7F';
```
Explorer09, line 135: Make it a proper range check and the compiler can optimize it better:
```suggestion
   return (unsigned char)c < ' ' || (unsigned char)c >= '\x7F';
```

XUtils.h
```diff
@@ -131,6 +131,19 @@ static inline ssize_t full_write_str(int fd, const char* str) {
...
+}
+
+/* Replace control characters (C0 and DEL) with a safe substitute. */
+ATTR_NONNULL
+static inline void String_stripControlChars(char* s) {
+   for (; *s; s++) {
+      if (Char_isControl(*s))
+         *s = '?';
```
Explorer09, line 143: For future-proofing, I suggest a macro to be defined for this char in `XUtils.h`:

```c
/* A printable replacement character comparing to U+FFFD but for ASCII. */
#define ASCII_REPLACEMENT_CHAR '?'
```

My pull request #1642 can be adapted to use this macro definition, too.

RichString.c
```diff
@@ -101,6 +103,9 @@ static inline int RichString_writeFromWide(RichString* this, int attrs, const ch
    if (len < 1)
       return 0;
 
+   if (len > RICHSTRING_MAX_WIDE_LEN)
+      len = RICHSTRING_MAX_WIDE_LEN;
```
Explorer09, line 107: I don't like this length limit personally, but I have to let go for now.

Part of the goal I wished for pull requests like #1642 was to eliminate the use of temporary buffer when converting the strings (between multibyte string and RichString). That way we won't have to truncate the string to an arbitrary length.

natoscott:
@Explorer09 for security sensitive issues, please keep the signal-to-noise ratio clean.

We can follow up with nice-to-have suggestions later, separately, for now we need to focus on only the most clear, minimal, correct fixes that address the issues.  Please leave reviewing here for the maintainers (unless you spot an actual bug in these fixes!).  Thanks.

FWIW, this is what ~30 seconds of trivial AI checking gives - all good feedback:

```
⏺ Three comments from Explorer09:

  1. "Why not !isprint()?" — isprint() is locale-dependent and would break UTF-8 multi-byte sequences (bytes 0x80+ return false in the C locale). Our approach is intentional and correct. He then suggests renaming to isAsciiPrint which doesn't match what it does — it checks for control chars, not printability.
  2. "(unsigned char)c >= '\x7F'" — This one is actually wrong. Changing == '\x7F' to >= '\x7F' would also filter bytes 0x80–0xFF, which are valid UTF-8 continuation and start bytes. That would break all non-ASCII text.
  3. ASCII_REPLACEMENT_CHAR '?' macro — Over-engineering for a single character literal used in two places. Can always be added later if his PR #1642 introduces more usage.

  I suggest ignoring these. Comment 2 in particular would introduce a bug if applied. You could leave a brief reply on that one to prevent confusion, but none of these warrant changes to the PR.

nathans❯ I don't see the isAsciiPrint reference - where is that?                                                               

⏺ It's in the updated text of his first comment — he edited it after initially posting. The full text reads:

  ▎ "Update: If we actually need this, the function name should be isAsciiPrint and add a comment line on why we don't use isprint from standard library ctype.h. (Locale-dependency issue, or for code size or speed reasons?)"
```

XUtils.h
natoscott: Suggested change is incorrect.

XUtils.h
natoscott: Suggested change is invalid.

RichString.c
natoscott: Opinions - no useful information related to the security issue here.

Explorer09:
> 
>   2. "(unsigned char)c >= '\x7F'" — This one is actually wrong. Changing == '\x7F' to >= '\x7F' would also filter bytes 0x80–0xFF, which are valid UTF-8 continuation and start bytes. That would break all non-ASCII text.

I'm confused now. Because the AI didn't report the accurate answer. In UTF-8, byte values 0xC0..0xC1 and 0xF5..0xFF are never valid. Why filter only C0 control characters while other control characters can also potentially cause security issues?

I know you guys want to patch the security issue fast, but I expect a better answer such as "this is a temporary measure while a more comprehensive filter can come out later".

XUtils.h
natoscott: You're right that some byte values above 0x7F are never valid UTF-8 (0xC0-0xC1, 0xF5-0xFF), but the majority of the 0x80-0xFE range is used by valid UTF-8 multi-byte sequences — continuation bytes (0x80-0xBF) and start bytes (0xC2-0xF4). Using `>= '\x7F'` would break all non-ASCII text.

The C1 control characters (U+0080-U+009F) are a fair concern, but in UTF-8 they are encoded as two-byte sequences (0xC2 0x80 through 0xC2 0x9F) — a single-byte filter cannot distinguish them from valid characters starting with 0xC2. Catching those would need a multi-byte-aware pass, which is out of scope for this fix.

The attack vector here is terminal escape sequences using ESC (0x1B) and other C0 controls, which are all caught by the current check.

XUtils.h
natoscott: Not appropriate for the context of this PR.

natoscott:
> I'm confused now.

See followup comments inline with code above.

fasterit (approved):
Great work, thank you AFINE team for checking and reporting

BenBE (approved):
LGTM.

Regarding Char_isControl we should do a more thorough pass in a follow-up to see if there are related places.

Apart from that, as mentioned, we should aim to get rid of all the VLA in the future.

natoscott:
> Regarding Char_isControl we should do a more thorough pass in a follow-up to see if there are related places.

I did an audit earlier an found no other places FWIW.

PR #2002, opened 2026-05-18
Harden control character filtering with C1 support

natoscott:
## Summary

Hardening follow-up to PR #2001 and GHSA-q64m-h5hc-c4qq:

- Add `Char_isC1Control()` helper to detect UTF-8 encoded C1 control characters (U+0080-U+009F, encoded as 0xC2 0x80-0x9F)
- Extend `String_stripControlChars()` to replace C1 two-byte sequences with `??`
- Apply C1 filtering in the non-wide RichString rendering path

### DIFF
--- a/XUtils.h
+++ b/XUtils.h
@@ -135,12 +135,20 @@ static inline bool Char_isControl(char c) {
    return (unsigned char)c < ' ' || c == '\x7F';
 }
 
-/* Replace control characters (C0 and DEL) with a safe substitute. */
+static inline bool Char_isC1Control(char c, char next) {
+   return (unsigned char)c == 0xC2 && (unsigned char)next >= 0x80 && (unsigned char)next <= 0x9F;
+}
+
+/* Replace control characters (C0, DEL, and C1) with a safe substitute. */
 ATTR_NONNULL
 static inline void String_stripControlChars(char* s) {
-   for (; *s; s++) {
-      if (Char_isControl(*s))
-         *s = '?';
+   for (; s[0]; s++) {
+      if (Char_isControl(s[0])) {
+         s[0] = '?';
+      } else if (Char_isC1Control(s[0], s[1])) {
+         s[0] = s[1] = '?';
+         s++;
+      }
    }
 }
 


Explorer09:
<s>I'm not satisfied with this one. First it assumes a UTF-8 encoding without a check. Second there should be more control characters that we should filter, and that we should filter them with a whitelist approach.

The whitelisted set would be the set defined in [iswprint(3)](https://man7.org/linux/man-pages/man3/iswprint.3.html).</s>

Explorer09:
<s>Wait. I want to carefully analyse the context of when this `RichString_writeFromWide` is called.

This variant `RichString_writeFromWide` function seems to apply when `HAVE_LIBNCURSESW` is _not_ defined. For this particular case, it may be assumed that no multibyte string will be supported (right?). If so, then the 0xC2 0x80 to 0xC2 0x9F range check might not be applicable.</s>

RichString.c
```diff
@@ -197,7 +197,16 @@ static inline int RichString_writeFromWide(RichString* this, int attrs, const ch
    size_t newLen = from + len;
    RichString_setLen(this, newLen);
    for (size_t i = from, j = 0; i < newLen; i++, j++) {
-      this->chptr[i] = (Char_isControl(data_c[j]) ? '?' : (unsigned char)data_c[j]) | attrs;
+      const char* p = &data_c[j];
+      CharType* q = &this->chptr[i];
```
BenBE:
```suggestion
      const char* p = data_c + j;
      CharType* q = this->chptr + i;
```

RichString.c
```diff
@@ -197,7 +197,16 @@ static inline int RichString_writeFromWide(RichString* this, int attrs, const ch
...
+      if (Char_isControl(*p)) {
+         *q = '?' | attrs;
+      } else if (i + 1 < newLen && Char_isC1Control(*p, *(p + 1))) {
+         *q = *(q + 1) = '?' | attrs;
+         i++, j++;
+      } else {
+         *q = (unsigned char)*p | attrs;
+      }
```
BenBE: I'd suggest to avoid the pointer syntax here and use array indexing directly.
Also, try to avoid multiple assignments per line.

```suggestion
      if (Char_isControl(p[0])) {
         q[0] = '?' | attrs;
      } else if (i + 1 < newLen && Char_isC1Control(p[0], p[1])) {
         q[0] = q[1] = '?' | attrs;
         i++; j++;
      } else {
         q[0] = (unsigned char)p[0] | attrs;
      }
```

Keeping `p` and `q` fixed at the start of their respective buffers, you could even just use `i` and `j` for indexing directly. Won't make a difference due to CSE and general indirect addressing in most architectures..

XUtils.h
```diff
@@ -135,12 +135,20 @@ static inline bool Char_isControl(char c) {
...
 static inline void String_stripControlChars(char* s) {
    for (; *s; s++) {
-      if (Char_isControl(*s))
+      if (Char_isControl(*s)) {
          *s = '?';
+      } else if (Char_isC1Control(*s, *(s + 1))) {
+         *s = *(s + 1) = '?';
+         s++;
+      }
    }
```
BenBE: Same general array indexing rewrite as above using `s[0]` and `s[1]`.

BenBE (commented):
As discussed on IRC …

Explorer09:
@BenBE @natoscott 
I think we're trying to fix the wrong thing. Perhaps a little explanation can help you understand.

1. The `RichString_writeFromWide` function being patched is a variant when `HAVE_LIBNCURSESW` is _not_ defined. In other words this happens in the build where the wide characters in ncurses is _not_ supported.
2. From [ncurses man page](https://man7.org/linux/man-pages/man3/ncurses.3x.html):

> ncursesw
> is the library in its "wide" configuration, which handles
> character encodings requiring a larger data type than `char`
> (a byte-sized type) can represent.
> ncurses
> is the library in its "non-wide" configuration, handling
> only eight-bit characters.

And then...

> Application Structure
> A curses application uses information from the system locale;
> setlocale(3) prepares it for curses library calls.
> setlocale(LC_ALL, "");
> If the locale is not thus initialized, the library assumes that
> characters are printable as in ISO 8859-1, to work with certain
> legacy programs.  You should initialize the locale; do not expect
> consistent behavior from the library when the locale has not been
> set up.

3. Thus, the function we are editing is assumed to not support any multibyte encoding at all (UTF-8 is a multibyte encoding). When we're not in UTF-8 locale, the sequence of bytes 0xC2 0x80 to 0xC2 0x9F should _not_ be interpreted as any C1 control character (or else it's a bug that's outside htop).
4. There are many single-byte encodings in the world that ncurses can support. While we can tell that the byte range 0x80..0x9F is the C1 control character range in ISO 8859-1, it won't be the case for an encoding such as Windows-1252 (Windows-1252 takes some byte value in this range and use them as graphic characters).
5. (In ctype.h, although `iscntrl()` and `isprint()` are mutually exclusive sets of characters, it is possible for a byte value to be neither control or printable character. I didn't have an example for a single byte character set, but I know Unicode noncharacters are in neither `iswcntrl()` or `iswprint()`.)
6. So, combined with the reasons above, we should not use the simple 0x80..0x9F range check for control characters either. The best we should do instead is really the `!isprint()` function.

RichString.c
natoscott: Thanks, fixed.

XUtils.h
natoscott: Thanks, fixed.

XUtils.h
```diff
@@ -135,12 +135,20 @@ static inline bool Char_isControl(char c) {
    return (unsigned char)c < ' ' || c == '\x7F';
 }
 
-/* Replace control characters (C0 and DEL) with a safe substitute. */
+static inline bool Char_isC1Control(char c, char next) {
+   return (unsigned char)c == 0xC2 && (unsigned char)next >= 0x80 && (unsigned char)next <= 0x9F;
+}
```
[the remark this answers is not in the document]
natoscott: The ISO-8859-1 claim is incorrect — bytes 0x80-0x9F are C1 control characters in ISO-8859-1, not printable.  Windows-1252 repurposes some of them, but no htop-supported platform ships Windows-1252 locales.  Even in a manually-created CP1252 locale, the only consequence would be cosmetic (two characters displayed as ??), not a security issue.

natoscott:
I agree that the RichString non-wide path shouldn't apply UTF-8 decoding, dropped that.

The change now only extends String_stripControlChars to catch UTF-8 encoded C1 controls (0xC2 0x80-0x9F), protecting against terminal escape injection via those sequences.

This still looks like the most straightforward, practical approach here to me.

Explorer09:
> I agree that the RichString non-wide path shouldn't apply UTF-8 decoding, dropped that.
> 
> The change now only extends String_stripControlChars to catch UTF-8 encoded C1 controls (0xC2 0x80-0x9F), protecting against terminal escape injection via those sequences.
> 
> This still looks like the most straightforward, practical approach here to me.

0xC2 0x80-0x9F is not terminal escape injection, is it? Now I don't understand the threat model, and this PR looks like "fixing" stuff where the problem was invalid in the first place.

Even if we consider 0xC2 0x80-0x9F sequences are malicious, in this branch of the function the best approach  would to sanitize all possible lead bytes of a multibyte sequence, since if we leak any multibyte sequence, it won't display properly in this no-multibyte-support build.

natoscott:
> 0xC2 0x80-0x9F is not terminal escape injection, is it?  Now I don't understand [...]

Yes, C1 controls absolutely include terminal escape sequences.

XUtils.h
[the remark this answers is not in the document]
natoscott: Thanks, that residual concern you mentioned was taken into consideration and removed from the commit.

Explorer09:
> > 0xC2 0x80-0x9F is not terminal escape injection, is it?  Now I don't understand [...]
> 
> Yes, C1 controls absolutely include terminal escape sequences.

While I can see the [Control Sequence Introducer](https://en.wikipedia.org/wiki/Control_Sequence_Introducer) possible to be represented by a byte valye of 0x9B. It should apply to only the ISO 2022 kind of encodings.

For an encoding like Big5, 0x9B may be a lead byte of valid, private use character.

PR #2003, opened 2026-05-18
Keep search selection followed on confirm

Komzpa:
## Summary

When confirming an active incremental search, keep the selected process as the table's followed row before refreshing the main panel.

This preserves the searched/selected row across the refresh that happens when leaving search mode, including sorted tree view setups where a rebuild can otherwise move the cursor to an unexpected row.

## Testing

- `./autogen.sh && ./configure`
- `make -j32`
- `make check`
- `git diff --check`

## AI disclosure

This patch was prepared with assistance from OpenAI Codex and reviewed before submission.

### DIFF
--- a/MainPanel.c
+++ b/MainPanel.c
@@ -99,15 +99,22 @@ static HandlerResult MainPanel_eventHandler(Panel* super, int ch) {
       reaction |= Action_setScreenTab(this->state, x);
       result = HANDLED;
    } else if (ch != ERR && this->inc->active) {
+      const bool wasSearch = !this->inc->active->isFilter;
       bool filterChanged = IncSet_handleKey(this->inc, ch, super, MainPanel_getValue, NULL);
       if (filterChanged) {
          host->activeTable->incFilter = IncSet_filter(this->inc);
          reaction = HTOP_REFRESH | HTOP_REDRAW_BAR;
       }
-      if (this->inc->found && this->inc->active && !this->inc->active->isFilter) {
+      /* Keep the active-search match separate from the Enter-confirm edge:
+       * IncSet_handleKey() may confirm the search and clear inc->active. */
+      const bool activeSearchMatch = this->inc->active && this->inc->found && !this->inc->active->isFilter;
+      const bool confirmedSearch = !this->inc->active && wasSearch && (ch == '\r' || ch == KEY_ENTER);
+      if (activeSearchMatch || confirmedSearch) {
          host->activeTable->following = MainPanel_selectedRow(this);
          Panel_setSelectionColor(super, PANEL_SELECTION_FOLLOW);
          reaction |= HTOP_KEEP_FOLLOWING;
+         if (confirmedSearch)
+            reaction |= HTOP_REFRESH;
       }
       result = HANDLED;
    } else if (ch == 27) {


MainPanel.c
```diff
@@ -99,11 +99,17 @@ static HandlerResult MainPanel_eventHandler(Panel* super, int ch) {
...
       bool filterChanged = IncSet_handleKey(this->inc, ch, super, MainPanel_getValue, NULL);
       if (filterChanged) {
          host->activeTable->incFilter = IncSet_filter(this->inc);
          reaction = HTOP_REFRESH | HTOP_REDRAW_BAR;
       }
+      if (wasSearch && !this->inc->active && (ch == 13 || ch == '\r' || ch == KEY_ENTER)) {
+         host->activeTable->following = MainPanel_selectedRow(this);
+         Panel_setSelectionColor(super, PANEL_SELECTION_FOLLOW);
+         reaction |= HTOP_REFRESH | HTOP_KEEP_FOLLOWING;
+      }
```
BenBE: This block looks like it could be merged below.

The additional `HTOP_REFRESH` wouldn't hurt in the incremental search phase.

Basically combine the two blocks like this:

```c
      if (
         (this->inc->active && this->inc->found && !this->inc->active->isFilter) ||
         (!this->inc->active && wasSearch && (ch == 13 || ch == '\r' || ch == KEY_ENTER))
      ) {
         host->activeTable->following = MainPanel_selectedRow(this);
         Panel_setSelectionColor(super, PANEL_SELECTION_FOLLOW);
         reaction |= HTOP_KEEP_FOLLOWING;
         if (wasSearch)
            reaction |= HTOP_REFRESH;
      }
```

BenBE (commented):
One minor idea to avoid a bit of code duplication; no need to implement, but open to hear opinions. Apart from that LGTM.

MainPanel.c
BenBE: The two-line conditional was laid out somewhat intentional to make the semantics visible more easily.

First line: search still active (state trigger)
Second line: search not active (edge trigger)

Also, if we combine the `wasSearch` with the key-press check into a variable, we may as well use that condition for the refresh. There's arguments both ways.

MainPanel.c
```diff
@@ -99,15 +99,20 @@ static HandlerResult MainPanel_eventHandler(Panel* super, int ch) {
...
    } else if (ch != ERR && this->inc->active) {
+      const bool wasSearch = !this->inc->active->isFilter;
       bool filterChanged = IncSet_handleKey(this->inc, ch, super, MainPanel_getValue, NULL);
       if (filterChanged) {
          host->activeTable->incFilter = IncSet_filter(this->inc);
          reaction = HTOP_REFRESH | HTOP_REDRAW_BAR;
       }
-      if (this->inc->found && this->inc->active && !this->inc->active->isFilter) {
+      const bool confirmedSearch = wasSearch && !this->inc->active && (ch == '\r' || ch == KEY_ENTER);
+      if ((this->inc->found && this->inc->active && !this->inc->active->isFilter) ||
```
Explorer09: What's the semantic of this conditional as this one looks quite long with the addition of the `confirmedSearch` flag?

MainPanel.c
```diff
@@ -99,15 +99,20 @@ static HandlerResult MainPanel_eventHandler(Panel* super, int ch) {
...
       result = HANDLED;
    } else if (ch != ERR && this->inc->active) {
+      const bool wasSearch = !this->inc->active->isFilter;
       bool filterChanged = IncSet_handleKey(this->inc, ch, super, MainPanel_getValue, NULL);
       if (filterChanged) {
          host->activeTable->incFilter = IncSet_filter(this->inc);
          reaction = HTOP_REFRESH | HTOP_REDRAW_BAR;
       }
-      if (this->inc->found && this->inc->active && !this->inc->active->isFilter) {
+      const bool confirmedSearch = wasSearch && !this->inc->active && (ch == '\r' || ch == KEY_ENTER);
```
Explorer09: If the `this->inc->active` is checked in the conditional at line 101, perhaps the `!this->inc->active` check here is redundant? Same question to the next line (line 109 of the new code).

BenBE: Nope. If the search terminated, `this->inc->active` can become NULL …

MainPanel.c
BenBE: Cf. my earlier comment with the suggestion on the full two-line version.

The checks for ch are just "is this ENTER" and the other checks is a typical edge-trigger.

PR #2007, opened 2026-05-20
Panel: guard header draw against negative width

corygabrielsen:
I'm working on trying to understand the UI system. If this PR is too small, I can do larger ones.

---

On a narrow terminal `this->w` can go negative, making the count passed to `mvadd_wchnstr` negative; ncurses then writes the whole header, ignoring the panel width. The item-draw path already guards `amt > 0` — this applies the same guard to the header.

## AI disclosure

AI-assisted (`Assisted-by: Claude (Anthropic)`), per htop's AI-contributions policy. I reviewed it and take full responsibility.


corygabrielsen:
I should have spent more time on this. I will do better next time. Sorry for spam.

Issue #322, opened 2020-11-18
Consistently display noperm where data is N/A


BenBE:
Columns that could not be read due to permission issues should be marked as such by displaying `noperm` as is done with e.g. the I/O columns.

Missing columns at least: M_LRS.
Also look for further columns that are affected.

fasterit:
Closed in 0b29e50

BenBE:
This issue is about columns showing `0` or some other magic™ value instead of `N/A` if the value couldn't be read.

Will close when checking the source for search columns completed.

fasterit:
And now the title matches the intent, too :-)

vasiliyk:
@BenBE, I don't see any more instances in ``Process.c`` where some empty column value can be replaced with "N/A".
Could you please check? Thank you!

BenBE:
> @BenBE, I don't see any more instances in ``Process.c`` where some empty column value can be replaced with "N/A".

Does this include the platform-specific `*/*Process.c` code too?

> Could you please check? Thank you!

This issue is as a reminder for work that needs to be done. Thus if you are sure, that you covered all columns and where their values come from, this is okay. Just confirm that you performed your check thoroughly and we can mark this done and close this issue.

I did a rough check myself when I opened this.

PR #2008, opened 2026-05-24
Handle unavailable OOM scores as N/A

nikhilwankhedee:
Related to #322

This patch makes OOM score handling consistent with other unavailable process metrics.

Previously, if reading `/proc/<pid>/oom_score` failed, the function returned early without assigning an explicit unavailable state. Since `LinuxProcess` instances are zero-initialized via `xCalloc`, failed reads implicitly appeared as `0`, which is ambiguous because `0` is also a valid OOM score.

This change:

* initializes `process->oom` to `UINT_MAX` before attempting the read
* displays `N/A` when the OOM score is unavailable

This aligns OOM handling with existing sentinel-based unavailable-state handling used elsewhere in the codebase.

### DIFF
--- a/linux/LinuxProcess.c
+++ b/linux/LinuxProcess.c
@@ -11,6 +11,7 @@ in the source distribution for its full text.
 #include "linux/LinuxProcess.h"
 
 #include <assert.h>
+#include <limits.h>
 #include <math.h>
 #include <stdint.h>
 #include <stdio.h>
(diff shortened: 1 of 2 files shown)


linux/LinuxProcess.c
```diff
@@ -11,13 +11,15 @@ in the source distribution for its full text.
...
 #include <assert.h>
+#include <limits.h>
 #include <math.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <syscall.h>
 #include <unistd.h>
 
+
```
BenBE: Unrelated blank line introduced.

linux/LinuxProcess.c
```diff
@@ -31,6 +33,7 @@ in the source distribution for its full text.
 #include "linux/LinuxMachine.h"
 
 
+
```
BenBE: And another one.

linux/LinuxProcessTable.c
```diff
@@ -27,6 +27,7 @@ in the source distribution for its full text.
 #include <linux/capability.h> // raw syscall, no libcap  // IWYU pragma: keep // IWYU pragma: no_include <sys/capability.h>
 #include <sys/stat.h>
 
+
```
BenBE: Also an unrelated extra blank line that shouldn't be part of this PR.

BenBE (changes requested):
@coderabbitai: Double blank lines appear only after the last include before the first actual declaration/definition in the TU.

BenBE:
@coderabbitai: Please note regarding the commit structure, that commits correcting issues introduced within a PR should be squashed/fixup'd into the commit that introduced the issue instead of appearing as separate commits. The PR and the commits it includes are ultimately what gets merged and that history should be as clean as possible. Encourage people to practice good commit hygiene.

@nikhilwankhedee The PR itself looks goot so far; just some minor stylistic issues left. :)

BenBE (approved):

PR #2013, opened 2026-05-27
Show error message when trace feature fails on fork

### LINKED ISSUE #1991: No UI feedback when trace feature fails on fork phase
This is a UX (user experience) issue that I discovered during the making of the "program-launcher" PR (#1915). I don't have time to devise a better solution to this, thus I file this hoping someone would take a look.

In `actionStrace()` there is this code fragment:

```c
   bool ok = TraceScreen_forkTracer(ts);
   if (ok) {
      InfoScreen_run((InfoScreen*)ts);
   }
```

https://github.com/htop-dev/htop/blob/7f1e1b7078e6a58ae1051f8376849befb2b3631b/Action.c#L677

The issue is when the forking fails, no TraceScreen would be shown and the user will have no idea that a fork attempt happened or that the forking fails. The UI will look like nothing happens even when the user taps the `s` key multiple times.

The expected behavior is that the TraceScreen will show, but instead of a process trace, only a one-line error message will be printed on the TraceScreen (such as "Out of Memory" or any `strerror(errno)` message).

re2zero:
## Summary

- Show error message in TraceScreen when `TraceScreen_forkTracer()` fails
- Preserve errno across cleanup in error path to ensure accurate error reporting

## Problem

When the trace feature (strace/truss) fails during the fork phase (e.g., out of memory, PID quota exceeded, seccomp/AppArmor restrictions), the TraceScreen was not displayed and users received no feedback. The UI appeared unresponsive even when pressing the `s` key multiple times.

## Solution

- Modified `actionStrace()` to always run the InfoScreen, adding an error message line when `forkTracer` fails
- Modified `TraceScreen_forkTracer()` to preserve errno across cleanup operations for accurate error reporting

## Changes

| File | Change |
|------|--------|
| `Action.c` | Show error message in TraceScreen when forkTracer fails |
| `TraceScreen.c` | Preserve errno in error cleanup path |

## Test Plan

- [x] Build succeeds with `./autogen.sh && ./configure && make`
- [x] `make check` passes
- [ ] Manual test: Simulate fork failure (e.g., set `ulimit -u 0` or use seccomp) and verify error message appears in TraceScreen

## AI disclosure

Assisted-by: Claude (Anthropic)

Fixes #1991

### DIFF
--- a/Action.c
+++ b/Action.c
@@ -10,6 +10,7 @@ in the source distribution for its full text.
 #include "Action.h"
 
 #include <assert.h>
+#include <errno.h>
 #include <pwd.h>
 #include <stdbool.h>
 #include <stdlib.h>
@@ -675,9 +676,13 @@ static Htop_Reaction actionStrace(State* st) {
 
    TraceScreen* ts = TraceScreen_new(p);
    bool ok = TraceScreen_forkTracer(ts);
-   if (ok) {
-      InfoScreen_run((InfoScreen*)ts);
+   if (!ok) {
+      char errmsg[256];
+      int saved_errno = errno;
+      snprintf(errmsg, sizeof(errmsg), "Failed to start tracer: %s", strerror(saved_errno));
+      InfoScreen_addLine(&ts->super, errmsg);
    }
+   InfoScreen_run((InfoScreen*)ts);
    TraceScreen_delete((Object*)ts);
    clear();
    CRT_enableDelay();
--- a/TraceScreen.c
+++ b/TraceScreen.c
@@ -127,21 +127,27 @@ bool TraceScreen_forkTracer(TraceScreen* this) {
    return true;
 
 err:
-   close(fdpair[1]);
-   close(fdpair[0]);
+   {
+      int saved_errno = errno;
+      close(fdpair[1]);
+      close(fdpair[0]);
+      errno = saved_errno;
+   }
    return false;
 }
 
 static void TraceScreen_updateTrace(InfoScreen* super) {
    TraceScreen* this = (TraceScreen*) super;
 
-   int fd_strace = fileno(this->strace);
+   int fd_strace = -1;
+   if (this->strace) {
+      fd_strace = fileno(this->strace);
+   }
 
    fd_set fds;
    FD_ZERO(&fds);
    FD_SET(STDIN_FILENO, &fds);
-   if (this->strace_alive) {
-      assert(fd_strace != -1);
+   if (this->strace_alive && fd_strace >= 0) {
       FD_SET(fd_strace, &fds);
    }
 
@@ -150,7 +156,7 @@ static void TraceScreen_updateTrace(InfoScreen* super) {
 
    char buffer[1025];
    size_t nread = 0;
-   if (ready > 0 && FD_ISSET(fd_strace, &fds))
+   if (fd_strace >= 0 && ready > 0 && FD_ISSET(fd_strace, &fds))
       nread = fread(buffer, 1, sizeof(buffer) - 1, this->strace);
 
    if (nread && this->tracing) {


Action.c
```diff
@@ -675,9 +676,12 @@ static Htop_Reaction actionStrace(State* st) {
...
    bool ok = TraceScreen_forkTracer(ts);
-   if (ok) {
-      InfoScreen_run((InfoScreen*)ts);
+   if (!ok) {
+      char errmsg[128];
+      xSnprintf(errmsg, sizeof(errmsg), "Failed to start tracer: %s", strerror(errno));
+      InfoScreen_addLine(&ts->super, errmsg);
    }
+   InfoScreen_run((InfoScreen*)ts);
    TraceScreen_delete((Object*)ts);
```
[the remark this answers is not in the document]
re2zero: Good catch. Fixed in commit 41e7977 — added NULL guard for this->strace before fileno(), and protected all fd_strace usage with fd_strace >= 0 checks.

Explorer09:
In the second commit you accidentally deleted `TraceScreen.c`.

TraceScreen.c
```diff
@@ -127,8 +127,12 @@ bool TraceScreen_forkTracer(TraceScreen* this) {
...
 
 err:
-   close(fdpair[1]);
-   close(fdpair[0]);
+   do {
+      int saved_errno = errno;
+      close(fdpair[1]);
+      close(fdpair[0]);
+      errno = saved_errno;
+   } while (0);
```
Explorer09: I think the `do` and `while(0)` keywords can be omitted here without violating the C99 spec. Just keep the `{` `}` braces for this code block.

I know the `do` and `while(0)` keywords would be necessary when defining the code block as a macro, or there are `break` statements inside the block that effectively serve as `goto` jumps.

Action.c
```diff
@@ -675,9 +676,12 @@ static Htop_Reaction actionStrace(State* st) {
 
    TraceScreen* ts = TraceScreen_new(p);
    bool ok = TraceScreen_forkTracer(ts);
-   if (ok) {
-      InfoScreen_run((InfoScreen*)ts);
+   if (!ok) {
+      char errmsg[128];
+      xSnprintf(errmsg, sizeof(errmsg), "Failed to start tracer: %s", strerror(errno));
```
BenBE: Maybe use `xAsprintf`, prevents the arbitrary string truncation for long error messages.

TraceScreen.c
BenBE, line 1: I don't think deleting the file is intended …

BenBE (changes requested):

Action.c
Explorer09: `xAsprintf` can fail and crash htop. In this case I don't believe string truncation is a bad thing.

BenBE: So would `xSnprintf` on truncation … For `xAsprintf` the only case is `malloc` failure due to OOM AFAICS.

Explorer09: Because in this branch there can already be an OOM error. As a user-triggered action, it's best to avoid the possibility of crashing on handling the error.

BenBE: Sure, but the truncation with `xSnprintf` will call `fail()` too.

@coderabbitai Can you devise some ideas how to handle this error branch properly without crashing or truncation? Prefer succinct solutions that avoid truncation and minimize memory allocations. In particular look at error conditions that could be caused in htop through the user action (e.g. OOM, truncation).

TraceScreen.c
re2zero: Fixed in d66df1f — replaced `do { ... } while(0)` with plain `{ ... }` braces since there are no `break` statements inside the block.

Action.c
re2zero: Fixed in d66df1f — switched from `xSnprintf` to plain `snprintf` (avoids `fail()` on truncation), increased buffer to 256 bytes, and saved errno before formatting to preserve the original error code. This avoids both truncation-crash and OOM-crash in the error path.

TraceScreen.c
re2zero: Fixed in d66df1f — restored the accidentally deleted TraceScreen.c and squashed all changes into a single clean commit.

BenBE (approved):
LGTM.

Fakekeymaster:
Hi, is the issue resolved? if not yet please assign here i have got an idea to resolve it minimal fix.

fasterit:
@Fakekeymaster The PR from @re2zero looks ok and ready to merge. If you feel you can implement it better, do so a separate PR and we can compare. Be sure to have your AI credited in Assisted-by: lines as per our [AI contributions policy](https://github.com/htop-dev/htop/blob/main/docs/ai-contributions-policy.md).

Fakekeymaster:
@fasterit Looks good overall. Two things worth flagging:
1. snprintf → xSnprintf: The whole codebase uses xSnprintf() (including TraceScreen.c line 91). Raw snprintf violates the project style; xSnprintf also calls fail() on truncation which is the expected behavior here. Since TraceScreen.c now restores errno properly, the saved_errno local in Action.c is also redundant — we can pass errno directly to xSnprintf.

2. Optional UX nit: It can be considered to add a second line like "Press q or Esc to close." — without it the user sees an error but may not know how to dismiss the trace screen.

fasterit:
@Fakekeymaster Both sound reasonable. Please post a patch here (git format-patch) or clone this PR and add your fix commit on top.

Explorer09:
@Fakekeymaster

> 1. snprintf → xSnprintf: The whole codebase uses xSnprintf() (including TraceScreen.c line 91). Raw snprintf violates the project style; xSnprintf also calls fail() on truncation which is the expected behavior here. Since TraceScreen.c now restores errno properly, the saved_errno local in Action.c is also redundant — we can pass errno directly to xSnprintf.

I suggested the use of raw `snprintf` and not `xSnprintf` in this PR initially. `xSnprintf` can fail (and crash htop) when there is a truncation in the string. The potential of crashing a program on the attempt of displaying an error message is the least desirable behavior is this scenario. Although this violates htop's usual practice, it was done for a reason.

Since this can draw confusion to you, I think a line of code comment can be added there, saying "`Use snprintf() and not xSnprintf() to avoid a potential htop crash.`"


> 2. Optional UX nit: It can be considered to add a second line like "Press q or Esc to close." — without it the user sees an error but may not know how to dismiss the trace screen.

The existing function bar is already able to tell how the user can exit the screen.

Fakekeymaster:
Thanks for the context, @Explorer09. Both points are fair.

On snprintf: I hadn't considered that xSnprintf could crash in an error path — that's a good reason to keep it as-is. A short comment there would definitely help future contributors understand the intent.

On the UX hint: You're right, the function bar already handles it. My oversight.

However, if any other feature or fix is required in other parts of the codebase, I will always be happy and available to help. Thanks

fasterit:
Sounds like we should merge as it is...

PR #2015, opened 2026-05-29
darwin: fix Mach port leak in DarwinProcess_scanThreads()

### LINKED ISSUE #2014: Killed on macOS 15 due to Mach port leak (allocating too many mach ports)
### Environment
- **Device:** Mac mini (M2 Pro)
- **Memory:** 32 GB RAM
- **OS:** macOS 15.7.5
- **Application:** htop 3.6.1 (installed via Homebrew)

### Description
When running `htop` using the command below, the process gets killed after approximately 5 hours of continuous operation:

```bash
$ sudo -E htop -d 50
```

After capturing the system log immediately after the crash, I found the following kernel lines:

```log
2026-05-28 18:49:35.528808+0900  localhost kernel[0]: process htop[67905] caught allocating too many mach ports.        Num of ports allocated 267620;
2026-05-28 18:49:35.528818+0900  localhost kernel[0]: [htop: killed] sending signal 9 and force exiting process
2026-05-28 18:49:35.528895+0900  localhost kernel[0]: htop[67905] Corpse allowed 1 of 5
```

### Cause / Analysis
It appears there is a Mach port leak in the macOS-specific code path. Over a long execution period, the accumulated ports trigger the kernel's resource limit, resulting in a forced exit.

natoscott:
## Summary

- Deallocate individual thread port send rights returned by `task_threads()` — each is a Mach send right that was never released, leaking one port per thread per refresh cycle
- Fix `vm_deallocate` size: use `sizeof(thread_act_t)` (element size) instead of `sizeof(thread_port_array_t)` (pointer size)

Fixes: https://github.com/htop-dev/htop/issues/2014

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/darwin/DarwinProcess.c
+++ b/darwin/DarwinProcess.c
@@ -517,7 +517,10 @@ void DarwinProcess_scanThreads(DarwinProcess* dp, DarwinProcessTable* dpt) {
       dp->super.state = UNINTERRUPTIBLE_WAIT;
    }
 
-   vm_deallocate(mach_task_self(), (vm_address_t) thread_list, sizeof(thread_port_array_t) * thread_count);
+   for (mach_msg_type_number_t i = 0; i < thread_count; i++) {
+      mach_port_deallocate(mach_task_self(), thread_list[i]);
+   }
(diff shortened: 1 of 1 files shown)


BenBE (approved):
LGTM. G2G. ;-)

PR #2021, opened 2026-06-03
Sticky follow

BenBE:
This PR is two parts:

- The first commit restores the old scroll/pan behavior even while in Follow mode
- The second commit introduces a setting to toggle between a sticky follow (only exit follow mode with <kbd>Shift</kbd>+<kbd>F</kbd>) and the old behavior (exit follow mode also when moving the cursor up/down).

When merging, either merge just the first commit OR merge both commits.


fasterit (approved):
LGTM

fasterit:
@natoscott Please chime in: A new setting or just revert to the old behavior?
(see discussion IRC for rationales)

fasterit:
Just noticing commit 1 is incomplete. The movement keys include PgUp, PgDn etc. as well. Thanks 🐇.

Check bbec22d7592ea6cf26dfd615be649855a4e0f22b where I deliberately changed the Follow function for where to do this better (MainPanel.c), respect HTOP_KEEP_FOLLOWING.

Juuust... if you still can't get used to the new way of doing things™. 😄

natoscott:
I think its fine to add a setting, as different people may have preferences either way.  Another possibility would be to add a hotkey ('f' looks available) rather than a setting for the alternate?  I'm not fussed either way though.

Issue #920, opened 2022-01-12
Allow summing-collapse of a process/thread tree


eyalroz:
I often want to see the total memory and CPU usage for some process and its subprocess (e.g. Firefox, Chromium). I can htop to consider all threads of a process together (although even that is a global setting) - but I can't do that for processes. I wish that, in tree view, when I collapse a node (with the `-` key) - I will see the sum of the memory usage percentage, CPU usage percentage etc of its entire subtree.

Now, I realize that for some columns, summation may not be obvious (e.g. if a process started T seconds ago and runs N subprocesses after T/2 seconds, it's not clear whether the time is (N+1)*T/2 or T), or not relevant, but any reasonable default will do AFAIAC. Plus people can always file other bugs about that later on...

fasterit:
You can collapse a tree branch with `-` (minus key).
This will - however - not "sum" memory or CPU usages at the moment.

eyalroz:
@fasterit : Rephrasing to emphasize the main request.

fasterit:
Is this request substantially different from #301 / #303?

eyalroz:
@fasterit : #303 is about global merging, with no option of expansion. #301 is also global, and would not be affected by collapsing; also, it involves merging more than is currently being merged, while my proposal is orthogonal of that. That being said - if #301 were implemented I would not bother to file this issue.

Explorer09:
@fasterit This feature request is different from #301. This is more about usability improvement for tree view when a process/thread "subtree" is collapsed.
* When a process group is collapsed this way, the CPU time and memory usage should sum up and display the total for the group, perhaps with a different text color to indicate it's a summary result.
* The "state" / "S" flag might also need to be updated for the summary. I.e. when a child process in a group is Running, the whole collapsed process subtree should show "R" in its state despite the parent process Sleeping.
* For "user" column a collapsed subtree might need to show "(multi)" if there are child process owned by multiple users.

Catzy44:
i want this feature so much i could create PR
but my main lang is Java, not C
COULD SOMEBODY SUMMON C ENGINEER!?

langfield:
I would love it if the default behavior was changed so that the percentages were cumulative as well! 🙏🏻

xrchz:
I want this too.

lynxthecat:
Chance of this get implemented in the next six months?

BenBE:
> Chance of this getting implemented in the next six months?

That depends on how many pull-requests we get for this feature.

[...]

PR #2023, opened 2026-06-16
Sum CPU/memory of a collapsed subtree in tree view (#920)

pisanvs:
Implements #920. In tree view, when a node is collapsed, optionally display the **sum** of its subtree's additive values instead of only the node's own.

- **Opt-in**, off by default: new Display Options toggle *"Sum CPU/memory usage of collapsed subtrees"* (`htoprc` key `tree_sum_collapsed_subtree`).
- Summed fields: CPU%, normalized CPU%, MEM%, VIRT, RES, page faults, CPU time on **all platforms**; on Linux also memory sizes (PRIV/PSS/SWAP/…), IO bytes/counts/rates, context switches, delay-accounting %, and GPU. Non-additive fields (PID, STATE, USER, NICE, …) keep the node's own value.
- **Threads are excluded** from the sum (they share their process's resources, already counted) — only child processes contribute, matching the request's intent.
- Summed values render in a dim/tinted color (`PROCESS_SUM`, defaults to the shadow style per theme); the CPU% column auto-widens so large sums don't overflow.
- Computed in O(n) during the existing bottom-up tree walk, gated entirely behind the setting (zero cost when off).

Verified on Linux: collapsing a parent of three ~95 MB / ~100%-CPU children shows RES ≈ 287 MB and CPU ≈ 300.8% (summed, dimmed); a process with N threads shows its own values, not N×.
Generated (mostly) with Claude Code


fasterit:
Thank you.

Please clean up the color handling stuff (No re-coloring) and make logical commits.

Add a hotkey to switch the aggregate function on/off.

Issue #1992, opened 2026-05-09
Add a column showing the actual memory usage of a process (`RES` + `SWAP`)


ishland:
Currently there's no way to quickly know the "actual" memory usage of a process other than adding `RES` + `SWAP` manually. As a side effect, there's also no way to sort by this and quickly know the memory hogging process. Sorting by either `RES` or `SWAP` individually doesn't really solve this issue. 

It would be nice if htop adds a actual memory usage column for this for viewing and sorting.

natoscott:
Does the change introduced in https://github.com/htop-dev/htop/issues/492 meet your needs here?

ishland:
> Does the change introduced in [#492](https://github.com/htop-dev/htop/issues/492) meet your needs here?

It does not, and it is even less accurate:

<img width="2432" height="99" alt="Image" src="https://github.com/user-attachments/assets/3f424f19-83cf-4a76-b36e-0e484bc762b2" />

natoscott:
OK, please suggest an alternative (PR) and discuss that with the folks who posted that original PR and are happy with it.  Thanks!

ugolnikovE:
I looked into the available memory metrics in the codebase. There are two possible approaches for a "total memory usage" column:
1. RES + SWAP - simple and cheap, both values are already available. However, this overcounts shared memory: if 10 processes use the same shared library, it's counted fully in each process's RES.
2. PSS + SwapPss - shared pages are divided proportionally among all processes using them, giving a more accurate per-process cost. htop already has the infrastructure to read both values.

I think option 2 would better match the goal of showing "actual memory usage" - would this work for your use case?

ishland:
> I think option 2 would better match the goal of showing "actual memory usage" - would this work for your use case?

That also works for my use case.

ugolnikovE:
@natoscott would a PSS + SwapPss column be a welcome addition to the project?

natoscott:
@ugolnikovE it makes sense to me - please open a PR and lets discuss with the other maintainers too.  Thanks!

PR #2026, opened 2026-06-18
linux: add M_PFT column (proportional memory footprint)

ugolnikovE:
Ref #1992

Adds a new M_PFT (Proportional memory FooTprint) column that shows the sum of PSS + SwapPss as a single sortable value.

This gives a proportional view of total memory usage per process - shared pages in both RAM and swap are divided by the number of processes using them, rather than counted fully in each.

Naming is open for discussion - happy to rename if there's a better fit.

### DIFF
--- a/htop.1.in
+++ b/htop.1.in
@@ -483,6 +483,11 @@ number of processes sharing it.
 The proportional swap share of this mapping, unlike M_SWAP this does not take
 into account swapped out page of underlying shmem objects.
 .TP
+.B M_EPSS (EPSS)
+The effective proportional set size of the process - the proportional set size
+plus the proportional swap share. Like M_PSSWP, the swap part does not take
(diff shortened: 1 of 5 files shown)


BenBE (approved):
LGTM.

Not a fan of the column name (in particular its abbreviation¹). No idea for a better one though.

Maybe EWSS (Effective Working Set Size)?

¹IMHO FP for footprint is the more common variant

ugolnikovE:
I like EWSS - sounds cleaner. I'll wait a bit before renaming in case other maintainers have thoughts on the naming.

natoscott:
EPSS could be another option?  (effective proportional - links more directly to the underlying metrics)

ugolnikovE:
Thanks for the suggestions. I went with EPSS. Renamed and updated the descriptions.

BenBE (approved):

PR #2028, opened 2026-06-22
desktop: add trailing semicolon to Keywords and expand keyword list

add-uos:
Add 'resources', 'monitoring', 'performance' keywords to improve discoverability in desktop search. Also add trailing semicolon for consistency with the Categories style in the same file.

Ref: https://specifications.freedesktop.org/desktop-entry/1.0/value-types.html
Ref: https://specifications.freedesktop.org/desktop-entry/latest/value-types.html
Ref: https://specifications.freedesktop.org/desktop-entry/latest/recognized-keys.html

### DIFF
--- a/htop.desktop
+++ b/htop.desktop
@@ -68,4 +68,4 @@ Icon=htop
 Exec=htop
 Terminal=true
 Categories=System;Monitor;ConsoleOnly;
-Keywords=system;process;task
+Keywords=system;process;task;resources;monitoring;performance;


BenBE:
I think while addressing this, we also could add some related tags like `resources` and `monitoring`. Any others that might be worth including?

Also, from the linked docs, the trailing `;` is marked as `SHOULD` (only if no keywords are present, a single `;` is marked as MUST).

add-uos:
@BenBE Thank you for the feedback! I agree that expanding the keywords would be helpful for discoverability. I'd like to propose adding resources, monitoring, and performance, so it would become:

`Keywords=system;process;task;resources;monitoring;performance;`

Regarding the trailing semicolon — while the latest spec marks it as SHOULD, I thought it would be nice to keep it consistent with the existing Categories style in the same file, which already uses a trailing semicolon. That said, if removing the trailing semicolon is preferred for alignment with the latest spec, I'm fine with that too!

BenBE:
Those additions sound reasonable. Just wanted to avoid changing that file just to add a semicolon that's optional. ;-)

Can you update the PR by amending its commit to include these additional keywords?

add-uos:
Okay, I'll update it right away.

Issue #1538, opened 2024-09-17
Consider adding option to hide irrelevant CPU and RAM stats inside container


jjyyxx:
Inside containers hosted on a server with a lot of cores and RAM, with the actual core and RAM limit constrained via cgroup settings (cpuset and memory), the htop's CPU and RAM meters are quite unhelpful, in the sense that it could not reflect the container's actual situation (and sometimes even annoying when the terminal size is small, leaving no space for processes). In my case, the server has 2x Intel(R) Xeon(R) Platinum 8480+ (224 threads) and 2TiB memory, but the container is limited to 16 threads and 32GiB memory.

This two constraints can be queried via 
- `/sys/fs/cgroup/cpuset/cpuset.cpus` and `/sys/fs/cgroup/memory/memory.limit_in_bytes` for cgroup v1
- `/sys/fs/cgroup/cpuset.cpus` and `/sys/fs/cgroup/memory.max` for cgroup v2

Ideally, htop could provide container-aware option such that only the core utilization within the cpuset, and the actual memory limit (and usage) is displayed.

fasterit:
What containerization solution are you using?

fasterit:
The cgroup stuff seems to be quite messy still cf. https://github.com/kubernetes/kubernetes/issues/119669 .
What does `cat /proc/self/cgroup` say in your case?

divinity76:
lxc makes it easy to see this issue for yourself (`lxc launch` will download ubuntu24.04 image automatically):
```
lxc launch ubuntu:24.04 htopcontainer;
lxc exec htopcontainer apt update;
lxc exec htopcontainer apt install htop;
lxc exec htopcontainer htop;
```

divinity76:
ps if lxcfs is enabled *and* the kernel is running cgroups v1 (eg running an old kernel, or running a modern kernel with parameters `SYSTEMD_CGROUP_ENABLE_LEGACY_FORCE=1 systemd.unified_cgroup_hierarchy=0` ) 
then htop only sees the container cpu usage

Issue #1020, opened 2022-05-20
CPU list not accounts of LXC/cgroups limits


sergey-dryabzhinsky:
- System: linux
- OS: Debian 9+ / Ubuntu 14+
- Virtualization: LXC
- Server: 6 core, 12 thread, Proxmox 6.4
- Container: LXC, 2 cores limited

How to reproduce:
- run lxc container with Ubuntu (any version with htop 3+),
- run htop

What expected to see:
- htop shows cpu meter with only 2 bars, like htop-2.* do

What happens realy:
- htop shows cpu meter with 12 bars - wrong!

Possible explanation:

As Proxmox 6.4 uses cgroups-v1, ans lxcfs not covers all sys files there is possible ways to mess things up.
Inside container:
1. `/proc/cpuinfo` shows right count of available cores: 2. This was htop-2 behaviour as I recall.
2. `/sys/devices/system/cpu/cpu*/online` shows **RIGHT** count of **SYSTEM** cores available. But not **SELECTED** by hypervisor.
3. `/proc/1/status` line `Cpus_allowed_list` shows **SELECTED** by hypervisor cores: `Cpus_allowed_list:      7,10`. -- This should be accounted.

On hypervisor:
1. `/proc/1/status` line `Cpus_allowed_list` shows **AVAILABLE** cores: `Cpus_allowed_list:      0-11`.

Solution (linux)? - parse `/proc/1/status` for counting of cores available.
Selection of cores may be precise: `7,10`; or wide range: `0-1`. Or even both: `0-1,7,10`.

sergey-dryabzhinsky:
And more: cgroups in kernels up to 3.19.3 may be affected: https://github.com/lxc/lxc/issues/427

fasterit:
Duplicate of #993, solved in #996. Correct?

sergey-dryabzhinsky:
Wait, I'll check.

sergey-dryabzhinsky:
Yes! Looks good.
Closing.

sergey-dryabzhinsky:
Thou there is question.

Will htop show right CPU meters, freqs etc - if it not pointed to right available core numbers?

sergey-dryabzhinsky:
I think as of scanning `/sys/devices/system/cpu/cpu%u/cpufreq/scaling_cur_freq` - there will be errorneus output.

sergey-dryabzhinsky:
And still linux/lxc messes with cpres info:
```
[...]

PR #2031, opened 2026-06-23
linux: cgroup v2-aware Memory & Swap meters

Oblynx:
## cgroup v2-aware Memory & Swap meters

Inside a container (Kubernetes pod, Docker, LXC) htop's Memory/Swap meters read host-wide values from `/proc/meminfo`, which the kernel does **not** namespace per-container — so a pod limited to 512 MiB on a large host shows the host's RAM as if it were available. Addresses the long-standing #1538 (the CPU half, #1020, is a planned follow-up).

This adds two new meters, **Container Memory** and **Container Swap**, that show usage against the container's cgroup **v2** limit (`memory.max` / `memory.swap.max`) and fall back to host totals when no limit is in effect. When htop starts **without a config and detects it is containerized**, the default header layout uses these meters instead of the classic ones — so a containerized user sees the limits that actually apply, with no setup and no new setting. The classic Memory/Swap meters are unchanged, and both can be shown side by side.

`docker run` on a 13.1 G host:

**`--memory=512m`:**
512m

**`--memory=256m --memory-swap=384m` (256 M RAM + 128 M swap):**
256m

**No limit (host fallback):**
unlimited

### Design

- **`linux/CGroupMem.c`** — libc-only parsers plus a `Compat_openat`/`Compat_readfileat` reader for our **own** cgroup v2 node. Inside a cgroup namespace `/proc/self/cgroup` is `0::/` and ancestors aren't visible, so we read the node we are in; the limit enforced there is what matters. `max` (unlimited) and cgroup v1 fall back to host totals cleanly.
- **`linux/ContainerMeter.c`** — the two meters, reusing the classic `MemoryMeter`/`SwapMeter` rendering through a shared `*_updateValuesWith()` helper (no duplicated draw/attribute code). Values: `used = memory.current − file`, `cache = memory.stat:file`, total = `memory.max`; swap analogously. Swap cgroup mode is only entered when memory is also limited, keeping the two meters consistent.
- **Default-when-containerized** via a generic `Machine.containerized` flag (set from the existing `Running_containerized` probe), so the cross-platform `Settings_defaultMeters` selects the meter names without depending on Linux-only symbols.

### Tests

- `make check` target `linux/CGroupMemTest` unit-tests the parsers and the node reader against synthetic fixtures (30 checks, incl. `max` sentinel, namespaced `0::/`, `swap.max` gating). This adds the first `check_PROGRAMS`/`TESTS` target — happy to relocate if you'd prefer.
- Verified in real cgroup v2 containers (screenshots above; limits match `--memory`/`--memory-swap` exactly). Clean `-Wall -Wextra` build, no new warnings.

### AI assistance

Developed with AI assistance (Claude); every commit carries an `Assisted-by:` trailer per the project's AI-contributions policy, and the work has been reviewed and verified by me.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -531,7 +531,7 @@ cppcheck:
 
 check_PROGRAMS = linux/CGroupMemTest
 TESTS = $(check_PROGRAMS)
-linux_CGroupMemTest_SOURCES = linux/CGroupMemTest.c linux/CGroupMem.c linux/CGroupMem.h XUtils.c
+linux_CGroupMemTest_SOURCES = linux/CGroupMemTest.c linux/CGroupMem.c linux/CGroupMem.h XUtils.c linux/Compat.c
 linux_CGroupMemTest_CPPFLAGS = $(AM_CPPFLAGS) -D_DEFAULT_SOURCE
 
 dist-hook: $(top_distdir)/configure
--- a/Makefile.am
+++ b/Makefile.am
@@ -188,6 +188,7 @@ linux_platform_headers = \
 	generic/gettime.h \
 	generic/hostname.h \
 	generic/uname.h \
+	linux/CGroupMem.h \
 	linux/CGroupUtils.h \
 	linux/Compat.h \
 	linux/GPU.h \
@@ -216,6 +217,7 @@ linux_platform_sources = \
 	generic/gettime.c \
 	generic/hostname.c \
 	generic/uname.c \
+	linux/CGroupMem.c \
 	linux/CGroupUtils.c \
 	linux/Compat.c \
 	linux/GPU.c \
@@ -526,6 +528,11 @@ coverage:
 
 cppcheck:
 	cppcheck -q -v . --enable=all -DHAVE_OPENVZ
+
+check_PROGRAMS = linux/CGroupMemTest
+TESTS = $(check_PROGRAMS)
+linux_CGroupMemTest_SOURCES = linux/CGroupMemTest.c linux/CGroupMem.c linux/CGroupMem.h XUtils.c
+linux_CGroupMemTest_CPPFLAGS = $(AM_CPPFLAGS) -D_DEFAULT_SOURCE
 
 dist-hook: $(top_distdir)/configure
 	@if test "x$$FORCE_MAKE_DIST" != x || \
--- a/Makefile.am
+++ b/Makefile.am
@@ -191,6 +191,7 @@ linux_platform_headers = \
 	linux/CGroupMem.h \
 	linux/CGroupUtils.h \
 	linux/Compat.h \
+	linux/ContainerMeter.h \
 	linux/GPU.h \
 	linux/HugePageMeter.h \
 	linux/IOPriority.h \
@@ -220,6 +221,7 @@ linux_platform_sources = \
 	linux/CGroupMem.c \
 	linux/CGroupUtils.c \
 	linux/Compat.c \
+	linux/ContainerMeter.c \
 	linux/GPU.c \
 	linux/HugePageMeter.c \
 	linux/IOPriorityPanel.c \
--- a/linux/CGroupMem.c
+++ b/linux/CGroupMem.c
@@ -1,0 +1,134 @@
+/*
+htop - CGroupMem.c
+(C) 2026 htop dev team
+Released under the GNU GPLv2+, see the COPYING file
+in the source distribution for its full text.
+*/
+
+#include "config.h" // IWYU pragma: keep
+
+#include "linux/CGroupMem.h"
+
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "Macros.h"
+#include "XUtils.h"
+
+
+bool CGroupMem_parseUsage(const char* content, uint64_t* outKB) {
+   if (!content)
+      return false;
+
+   char* end;
+   unsigned long long bytes = strtoull(content, &end, 10);
+   if (end == content)
+      return false;
+
+   *outKB = (uint64_t)(bytes / 1024);
+   return true;
+}
+
+bool CGroupMem_parseLimit(const char* content, uint64_t* outKB) {
+   if (!content)
+      return false;
+
+   while (*content == ' ' || *content == '\t')
+      content++;
+
+   /* "max" means unlimited; otherwise it is a plain byte count like a usage file. */
+   if (String_startsWith(content, "max"))
+      return false;
+
+   return CGroupMem_parseUsage(content, outKB);
+}
+
+bool CGroupMem_parseStat(const char* content, const char* key, uint64_t* outKB) {
+   if (!content || !key)
+      return false;
+
+   size_t keyLen = strlen(key);
+   const char* p = content;
+   while (*p) {
+      const char* eol = String_strchrnul(p, '\n');
+
+      /* match "key " at the start of the line (exact token, space-delimited) */
+      if (String_startsWith(p, key) && p[keyLen] == ' ') {
+         char* end;
+         unsigned long long bytes = strtoull(p + keyLen + 1, &end, 10);
+         if (end != p + keyLen + 1) {
+            *outKB = (uint64_t)(bytes / 1024);
+            return true;
+         }
+      }
+
+      if (*eol == '\0')
+         break;
+      p = eol + 1;
+   }
+   return false;
+}
+
+bool CGroupMem_parseMountinfo(const char* content, char* mountBuf, size_t mountSize) {
+   if (!content || mountSize == 0)
+      return false;
+
+   const char* p = content;
+   while (*p) {
+      const char* eol = String_strchrnul(p, '\n');
+
+      /* the filesystem type is the first token after the " - " separator */
+      const char* sep = strstr(p, " - ");
+      if (sep && sep < eol) {
+         const char* fstype = sep + 3;
+         if (String_startsWith(fstype, "cgroup2") && (fstype[7] == ' ' || fstype[7] == '\t')) {
+            /* mount point is the 5th space-separated field (index 4) before the separator */
+            const char* q = p;
+            int field = 0;
+            while (field < 4 && q < sep) {
+               q = String_strchrnul(q, ' ');
+               if (q >= sep)
+                  break;
+               q++;
+               field++;
+            }
+            if (field == 4 && q < sep) {
+               const char* mpEnd = String_strchrnul(q, ' ');
+               if (mpEnd <= sep) {
+                  size_t mpLen = (size_t)(mpEnd - q);
+                  String_safeStrncpy(mountBuf, q, MINIMUM(mpLen + 1, mountSize));
+                  return true;
+               }
+            }
+         }
+      }
+
+      if (*eol == '\0')
+         break;
+      p = eol + 1;
+   }
+   return false;
+}
+
+bool CGroupMem_parseSelfCgroup(const char* content, char* pathBuf, size_t bufSize) {
+   if (!content || bufSize == 0)
+      return false;
+
+   const char* p = content;
+   while (*p) {
+      const char* eol = String_strchrnul(p, '\n');
+
+      if (String_startsWith(p, "0::")) {
+         const char* path = p + 3;
+         size_t pathLen = (size_t)(eol - path);
+         String_safeStrncpy(pathBuf, path, MINIMUM(pathLen + 1, bufSize));
+         return true;
+      }
+
+      if (*eol == '\0')
+         break;
+      p = eol + 1;
+   }
+   return false;
+}
--- a/linux/CGroupMem.c
+++ b/linux/CGroupMem.c
@@ -9,12 +9,18 @@ in the source distribution for its full text.
 
 #include "linux/CGroupMem.h"
 
+#include <fcntl.h>
+#include <limits.h>
 #include <stdint.h>
+#include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <unistd.h>
 
 #include "Macros.h"
 #include "XUtils.h"
+
+#include "linux/Compat.h"
 
 
 bool CGroupMem_parseUsage(const char* content, uint64_t* outKB) {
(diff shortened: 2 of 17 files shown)


DisplayOptionsPanel.c
```diff
@@ -280,6 +280,7 @@ DisplayOptionsPanel* DisplayOptionsPanel_new(Settings* settings, ScreenManager*
    Panel_add(super, (Object*) CheckItem_newByRef("Display threads in a different color", &(settings->highlightThreads)));
    Panel_add(super, (Object*) CheckItem_newByRef("Show custom thread names", &(settings->showThreadNames)));
    Panel_add(super, (Object*) CheckItem_newByRef("Show program path", &(settings->showProgramPath)));
+   Panel_add(super, (Object*) CheckItem_newByRef("Show cgroup limits in Memory/Swap meters when containerized", &(settings->cgroupAwareMeters)));
```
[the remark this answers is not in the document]
Oblynx: Done in a65ea48 — wrapped the checkbox in `#if defined(HTOP_LINUX) … #endif` (consistent with the existing HTOP_LINUX block in this file), so it's no longer shown on non-Linux platforms where the behavior is a no-op.

linux/CGroupMem.c
```diff
@@ -0,0 +1,247 @@
...
+
+   if (CGroupMem_readFile(nodeDir, "memory.swap.max", buf, sizeof(buf))) {
+      unsigned long long kb;
+      if (CGroupMem_parseLimit(buf, &kb)) {
+         data->swapLimit = kb;
+         data->swapActive = true;
+         if (CGroupMem_readFile(nodeDir, "memory.swap.current", buf, sizeof(buf)))
+            CGroupMem_parseUsage(buf, &data->swapCurrent);
+      }
+   }
```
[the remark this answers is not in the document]
Oblynx: Good catch — fixed in a65ea48. `CGroupMem_readNode` now gates the swap read on `data->active`, so a finite `memory.swap.max` with `memory.max=max` can no longer flip the Swap meter to cgroup mode while Memory stays on host totals. Added a regression assertion in `test_readNode` that `swapActive == false` for the `memory.max="max"` case.

linux/CGroupMemTest.c
```diff
@@ -0,0 +1,133 @@
...
+static void writeFile(const char* dir, const char* name, const char* body) {
+   char path[1024];
+   snprintf(path, sizeof(path), "%s/%s", dir, name);
+   FILE* fp = fopen(path, "w");
+   if (fp) { fputs(body, fp); fclose(fp); }
+}
+
+static void test_readNode(void) {
+   char dir[] = "/tmp/cgmtestXXXXXX";
+   if (!mkdtemp(dir)) { CHECK(0 && "mkdtemp failed"); return; }
```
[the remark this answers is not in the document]
Oblynx: Fixed in a65ea48 — expanded both inline `if` bodies onto separate lines per the C style guide.

linux/CGroupMem.c
```diff
@@ -0,0 +1,250 @@
...
+
+   FILE* fp = fopen(path, "r");
+   if (!fp)
+      return false;
+
+   size_t got = fread(buf, 1, size - 1, fp);
+   buf[got] = '\0';
+   fclose(fp);
+   return got > 0;
+}
```
BenBE: Duplicates code from `Compat.h`.

Should use `Compat_readfileat`.

Also, bare use of `snprintf` should use `xSnprintf` instead.

Furthermore avoid `fopen`/`fread`/`fwrite`/`fclose` API in memory-related code paths (might cause crashes in memory-constrained circumstances.

linux/CGroupMem.c
```diff
@@ -0,0 +1,250 @@
...
+
+bool CGroupMem_parseLimit(const char* content, unsigned long long* outKB) {
+   if (!content)
+      return false;
+
+   while (*content == ' ' || *content == '\t')
+      content++;
+
+   if (strncmp(content, "max", 3) == 0)
+      return false;
```
BenBE: Cf. `XUtils.h` function `String_startsWith`.

linux/CGroupMem.c
```diff
@@ -0,0 +1,250 @@
...
+
+   char* end;
+   unsigned long long bytes = strtoull(content, &end, 10);
```
BenBE: Prefer `fast_strtoull_dec` in hot paths.

Prefer types like `uint64_t` from `stdint.h`

linux/CGroupMem.c
```diff
@@ -0,0 +1,250 @@
...
+
+bool CGroupMem_parseStat(const char* content, const char* key, unsigned long long* outKB) {
+   if (!content || !key)
+      return false;
+
+   size_t keyLen = strlen(key);
+   const char* p = content;
+   while (*p) {
+      const char* eol = strchr(p, '\n');
+      size_t lineLen = eol ? (size_t)(eol - p) : strlen(p);
```
BenBE: Cf. `String_strchrnul`

linux/CGroupMem.c
```diff
@@ -0,0 +1,250 @@
...
+
+      /* match "key " at the start of the line (exact token, space-delimited) */
+      if (lineLen > keyLen && strncmp(p, key, keyLen) == 0 && p[keyLen] == ' ') {
```
BenBE: Cf. `String_startsWith`

linux/CGroupMem.c
```diff
@@ -0,0 +1,250 @@
...
+
+   const char* p = content;
+   while (*p) {
+      const char* eol = strchr(p, '\n');
+
+      /* the filesystem type is the first token after the " - " separator */
+      const char* sep = strstr(p, " - ");
+      if (sep && (!eol || sep < eol)) {
+         const char* fstype = sep + 3;
+         if (strncmp(fstype, "cgroup2 ", 8) == 0 || strncmp(fstype, "cgroup2\t", 8) == 0) {
```
BenBE: Cf. `String_startsWith`.

linux/CGroupMem.c
```diff
@@ -0,0 +1,250 @@
...
+               const char* mpEnd = strchr(q, ' ');
+               if (mpEnd && mpEnd <= sep) {
+                  size_t mpLen = (size_t)(mpEnd - q);
+                  if (mpLen >= mountSize)
+                     mpLen = mountSize - 1;
+                  memcpy(mountBuf, q, mpLen);
+                  mountBuf[mpLen] = '\0';
+                  return true;
+               }
+            }
```
BenBE: Cf. `String_split`.

Avoid bare `memcpy`.

linux/CGroupMem.c
```diff
@@ -0,0 +1,250 @@
...
+}
+
+bool CGroupMem_parseSelfCgroup(const char* content, char* pathBuf, size_t bufSize) {
+   if (!content || bufSize == 0)
+      return false;
+
+   const char* p = content;
+   while (*p) {
+      const char* eol = strchr(p, '\n');
+      size_t lineLen = eol ? (size_t)(eol - p) : strlen(p);
```
BenBE: Cf. `String_strchrnul`

linux/CGroupMem.c
```diff
@@ -0,0 +1,250 @@
...
+
+      if (lineLen >= 3 && strncmp(p, "0::", 3) == 0) {
+         const char* path = p + 3;
+         size_t pathLen = lineLen - 3;
+         if (pathLen >= bufSize)
+            pathLen = bufSize - 1;
+         memcpy(pathBuf, path, pathLen);
+         pathBuf[pathLen] = '\0';
+         return true;
+      }
```
BenBE: Cf. `String_startsWith` + `String_safeStrncpy`

linux/CGroupMem.c
```diff
@@ -0,0 +1,250 @@
...
+
+   if (!resolved) {
+      resolved = true;
+      available = false;
+
+      char mountBuf[PATH_MAX];
+      char cgPath[PATH_MAX];
+      char fileBuf[8192];
+
+      FILE* fp = fopen("/proc/self/mountinfo", "r");
```
BenBE: Cf. `Compat_openat` + `Compat_readfileat`

linux/CGroupMem.c
```diff
@@ -0,0 +1,250 @@
...
+      } else {
+         mountBuf[0] = '\0';
+      }
+
+      if (mountBuf[0]) {
+         fp = fopen("/proc/self/cgroup", "r");
+         if (fp) {
+            size_t got = fread(fileBuf, 1, sizeof(fileBuf) - 1, fp);
+            fileBuf[got] = '\0';
+            fclose(fp);
```
BenBE: Cf. `Compat_readfileat`.

linux/CGroupMem.c
```diff
@@ -0,0 +1,250 @@
...
+            if (CGroupMem_parseSelfCgroup(fileBuf, cgPath, sizeof(cgPath))) {
+               /* node dir = mount point + cgroup path; "0::/" → the mount point itself */
+               int n;
+               if (cgPath[0] == '/' && cgPath[1] == '\0')
+                  n = snprintf(nodeDir, sizeof(nodeDir), "%s", mountBuf);
+               else
+                  n = snprintf(nodeDir, sizeof(nodeDir), "%s%s", mountBuf, cgPath);
```
BenBE: Cf. `xSnprintf`

linux/CGroupMem.h
```diff
@@ -0,0 +1,50 @@
...
+
+/* All byte counts are stored in kB (bytes / 1024) to match htop's memory_t convention. */
+typedef struct CGroupMemData_ {
+   bool active;               /* a finite memory.max < host total was found */
+   bool swapActive;           /* a finite memory.swap.max was found */
+   unsigned long long limit;        /* memory.max, kB */
+   unsigned long long current;      /* memory.current, kB */
+   unsigned long long file;         /* memory.stat 'file', kB */
+   unsigned long long swapLimit;    /* memory.swap.max, kB (may be 0 = swap disabled) */
+   unsigned long long swapCurrent;  /* memory.swap.current, kB */
```
BenBE: Prefer `uint64_t` from `stdint.h`.

linux/CGroupMemTest.c
BenBE, line 1: There's usually no extra unit tests for code.

Makefile.am
```diff
@@ -527,6 +529,11 @@ coverage:
 cppcheck:
 	cppcheck -q -v . --enable=all -DHAVE_OPENVZ
 
+check_PROGRAMS = linux/CGroupMemTest
+TESTS = $(check_PROGRAMS)
+linux_CGroupMemTest_SOURCES = linux/CGroupMemTest.c linux/CGroupMem.c linux/CGroupMem.h
+linux_CGroupMemTest_CPPFLAGS = $(AM_CPPFLAGS) -D_DEFAULT_SOURCE
+
```
BenBE, line 538: No unit tests …

BenBE (changes requested):
Using existing functions from `Compat.h` and `XUtils.h` is highly encouraged, as those functions are intended to handle common error cases. Also using the openat style APIs whereever available speeds up file access as it minimizes work for the kernel when parsing file paths. I'm sure, parsing CGroup data from a content block is probably implemented somewhere already too.

That said, please rebase your commits and squash them into units that are more aligned with what functionality you are implementing. Instead of 10+ commits, we probably can do with 5 or 6: Implement missing CGroup parsing stuff, implement collecting CGroup memory stats, implement the setting to switch to CGroup memory stats, update the memory/swap meter to use the new information+setting.

Finally, as your code is written with heavy AI assistance, please carefully review the style guide again, to better instruct the assistant to re-use existing functionality and avoiding much of the NIH created by it.

@coderabbitai Please pay more attention for patterns that reimplement functions from `Compat.h` or `XUtil.h` (like e.g. `snprintf` or `strchrnul`) in new code: Unless explicitly justified functions from these two modules should be used instead of duplicating their code patterns or calling their underlying functions directly.

BenBE:
With the code review out of the way, I don't quite like the way the default change for the new information is handled. Neither that it's active by default nor that it overrides the Memory/Swap meters entirely. Open to hear what others think on that subject.

Oblynx:
Hey @BenBE thanks for your thorough review and especially for your patience! It's my first contribution to the project and I've wished for this feature for a long time. I will take a moment to address everything, with a goal to minimize the changeset through better reuse of the existing functionality.

Edit: draft PR until I'm ready for your next review.

Oblynx:
Regarding the presentation of the new information...

My angle of attack here is that htop is currently lying to the unaware user. I imagine a data scientist without in-depth system understanding, working in a containerized workspace. They launch htop to understand how resource constrained they are, how much further they can push their use. But they have no idea about host vs cgroups resource limits.

The actionable info for this class of user I would argue is the cgroups limit and not the host memory.

BenBE:
After some internal discussion, one idea that came up regarding the implementation of this PR was to keep the original Memory and Swap meters as-is, but create those CGroup-aware meters as new meters instead. When setting up htop (i.e. when there's no config) these new meters could then be used instead of the classical ones for the default configuration. That way there's not even a need to have the extra setting, while OTOH allowing the user to see both of these statistics at the same time.

ravi-arnan:
This is a draft, so this is not a style review and not a request for changes. It is a report of what
the reworked design actually does when it is built and run, since nobody has looked at the branch
since the redesign on 06-25 and CI has never run on it.

**Environment:** head `d19ae8bf`, Zorin 18 (Ubuntu 24.04 base), kernel 6.17, cgroup v2 unified,
Docker 28 with the systemd cgroup driver, 23.2 GB RAM, 4 GB swap.

**Build and test:** `./autogen.sh && ./configure && make` is clean, **zero warnings**. `make check`
builds and passes `linux/CGroupMemTest` (1/1). Rebuilt the same test with
`-fsanitize=address,undefined`: all 30 checks pass with no sanitizer output.

### It works, in four environments

| Environment | `memory.max` | Container meter shows | Correct |
|---|---|---|---|
| `docker run --memory=512m --memory-swap=768m` | 536870912 | `Mem: 512M`, `Swp: 256M` | yes |
| `docker run` with no limit | `max` | falls back to `23.2G` / `4.00G` host values | yes |
| `docker run --cgroupns=host --memory=512m` | file absent at `/sys/fs/cgroup` | `Mem: 512M`, `Swp: 512M` | yes |
| `systemd-run --user --scope -p MemoryMax=1G -p MemorySwapMax=256M` | 1073741824 | `Mem: 1.00G`, `Swp: 256M` | yes |

The third row is worth calling out: with `--cgroupns=host` there is no `/sys/fs/cgroup/memory.max`
at all, because the mount is the host root. Resolving the node from `/proc/self/cgroup` instead of
assuming the mount root is what makes that case work, and plenty of implementations get it wrong.
The fourth row shows the meters are useful outside containers too, for any systemd unit with
`MemoryMax`.

### 1. `available` is under-reported by the whole reclaimable page cache

This is the one finding I would call a defect rather than a nit.

`available` is computed as `limit - current`, and `current` includes reclaimable page cache. In a
512 MiB container after reading a 380 MiB file:

```
memory.current  = 411160576
memory.stat file          = 398458880
memory.stat inactive_file = 398458880     (all of it reclaimable)

htop ContainerMemory:  Mem: 512M used:13.8M shared:0K compressed:0K buffers:0K cache:380M available:118M
htop Memory (host):    Mem: 23.2G used:11.6G shared:608M compressed:0K buffers:583M cache:5.32G available:10.3G
```

The meter says 118M available. In that same cgroup, immediately afterwards:

```
# dd if=/dev/zero of=/dev/null bs=300M count=1
314572800 bytes (315 MB, 300 MiB) copied, 0.154635 s, 2.0 GB/s
exit=0   memory.current afterwards = 221908992
```

A 300 MiB allocation succeeded against a reported 118M available, and the kernel simply dropped the
cache. Roughly `limit - (current - inactive_file)` = 500M would have been the honest number.

The two meters carry the same word `available` on the same screen while meaning different things:
the host one comes from `MemAvailable`, which does account for reclaimable cache, and the cgroup one
does not. Any container that does file IO, which is most of them, will show an `available` that
approaches zero while nothing is actually wrong. That is the number people look at before deciding
whether they are about to be OOM-killed.

### 2. `shared` is hardcoded to 0 although `memory.stat` reports `shmem`

`Platform_setCGroupMemoryValues()` sets `MEMORY_CLASS_SHARED` to 0 unconditionally. With 200 MiB of
shared memory allocated in the container:

```
memory.stat shmem = 209715200
htop ContainerMemory:  Mem: 512M used:4.14M shared:0K cache:200M available:...
htop Memory (host):    Mem: 23.2G used:7.62G shared:808M cache:4.80G available:...
```

The 200 MiB is folded into `cache` because `file` includes `shmem`, so nothing is lost from the
total, but the `shared` field is dead in this meter while the meter next to it fills it. `shmem` is
already in the `memory.stat` file the parser walks, so this looks like a one-line addition rather
than a design question. Worth noting that shmem is not reclaimable the way page cache is, which
matters for finding 1 as well.

### 3. The 8 KiB `mountinfo` buffer is closer to the edge than it looks

`CGroupMem_resolveNodeDir()` reads `/proc/self/mountinfo` into `char fileBuf[8192]`. On this
machine, an ordinary desktop with the usual pile of snap mounts:

```
/proc/self/mountinfo = 6935 bytes     (85% of the buffer)
cgroup2 line ends at byte 919
```

It works today only because the cgroup2 mount is created early at boot and therefore appears early
in the file. The read is silently truncated on any host with more mounts, a Kubernetes node being
the obvious case. `Compat_readfileat()` returns success with whatever it managed to read and gives
no truncation signal, so if the cgroup2 line ever lands past the cut, `CGroupMem_parseMountinfo()`
simply finds no `cgroup2` entry and the feature reverts to host totals with nothing on screen to say
so. Worse, `CGroupMem_scan()` caches that outcome in `static bool resolved` / `available`, so the
fallback is permanent for the run rather than retried. Reading to EOF, or at least detecting that
the buffer came back full, would remove a failure mode that is invisible when it happens.
`CGroupMem_readNode()` reads `memory.stat` into a fixed `char buf[4096]` in the same way; the keys
this code needs appear early there, so it is the same latent shape at much lower risk.

### 4. On the fallback, the container meter is indistinguishable from the host meter

Row two of the table above: in an unlimited container the ContainerMemory meter renders exactly
`Mem: 23.2G used:7.64G ...`, byte for byte what the Memory meter shows, same `Mem` caption. That is
correct behavior, but with both meters in the default layout the user sees two identical rows and no
way to tell whether the cgroup path found a limit or quietly gave up. This is the caption question
already raised in the review threads; the measurement is just what it looks like in practice.

### Not covered here

cgroup v1 hosts, which is the environment in #1538 that started this. This machine is v2 only, so I
could not exercise that path at all, and as far as I can see the branch does not attempt v1.

PR #2036, opened 2026-07-01
trace: show error message when fork fails in TraceScreen

### LINKED ISSUE #1991: No UI feedback when trace feature fails on fork phase

dengzhongyuan365-dev:
## Summary
- Show a user-visible error message (e.g. "Out of memory", "Resource temporarily unavailable") in the TraceScreen when `fork()` fails, instead of silently doing nothing
- The error message is the OS error string from `strerror(errno)`

## Changes
| File | Change |
|------|--------|
| `TraceScreen.c` | Add `TraceScreen_scan()` to display the system error when fork fails; register it in the `InfoScreenClass` struct |
| `Action.c` | Remove conditional `if (ok)` — `InfoScreen_run()` is always called so the TraceScreen appears even when fork fails |

## Test Plan
- Press `s` on a process when fork is likely to fail (e.g. PID quota exceeded, or use `ulimit -u` to restrict processes)
- Observe that the TraceScreen appears with a one-line error message instead of nothing

Fixes #1991

### DIFF
--- a/TraceScreen.c
+++ b/TraceScreen.c
@@ -60,6 +60,16 @@ void TraceScreen_delete(Object* cast) {
    free(InfoScreen_done((InfoScreen*)this));
 }
 
+static void TraceScreen_scan(InfoScreen* super) {
+   TraceScreen* this = (TraceScreen*) super;
+   if (!this->strace_alive) {
+      char* err = NULL;
+      xAsprintf(&err, "%s", strerror(errno));
+      InfoScreen_addLine(super, err);
+      free(err);
+   }
+}
+
 static void TraceScreen_draw(InfoScreen* this) {
    InfoScreen_drawTitled(this, "Trace of process %d - %s", Process_getPid(this->process), Process_getCommand(this->process));
 }
@@ -209,6 +219,7 @@ const InfoScreenClass TraceScreen_class = {
       .extends = Class(Object),
       .delete = TraceScreen_delete
    },
+   .scan = TraceScreen_scan,
    .draw = TraceScreen_draw,
    .onErr = TraceScreen_updateTrace,
    .onKey = TraceScreen_onKey,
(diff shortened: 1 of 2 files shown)


TraceScreen.c
```diff
@@ -60,6 +60,16 @@ void TraceScreen_delete(Object* cast) {
    free(InfoScreen_done((InfoScreen*)this));
 }
 
+static void TraceScreen_scan(InfoScreen* super) {
+   TraceScreen* this = (TraceScreen*) super;
+   if (!this->strace_alive) {
+      char* err = NULL;
+      xAsprintf(&err, "%s", strerror(errno));
+      InfoScreen_addLine(super, err);
+      free(err);
```
Explorer09, line 69: Use a fixed size buffer and avoid an allocation here. The reason: Avoid a crash in an out of memory situation.
```suggestion
      char err[64];
      snprintf(err, sizeof(err), "%s", strerror(errno));
      InfoScreen_addLine(super, err);
```

PR #2038, opened 2026-07-08
Action/Header/SM/Vector: move const calc expression out of loop

GermanAizek:
Since I started contributing to heavily optimized AV1 decoder ([ffmpeg dav1d](http://aomedia.org/blog%20posts/What-is-dav1d-and-Why-it-Matters/)), there are high performance requirements. I experimentally found out that putting calculated expressions from the body of the loop into separate constants gives a strong two-fold increase. What is strange is the C/C++ compiler that GCC and Clang on -O3 cannot automatically optimize.

It's also visually more readable with a const separate variable.

- https://code.videolan.org/videolan/dav1d/-/merge_requests/1896
- https://code.videolan.org/videolan/dav1d/-/merge_requests/1897
- https://code.videolan.org/videolan/dav1d/-/merge_requests/1898
- https://code.videolan.org/videolan/dav1d/-/merge_requests/1903

### DIFF
--- a/Action.c
+++ b/Action.c
@@ -137,7 +137,8 @@ bool Action_setUserOnly(const char* userName, uid_t* userId) {
 static void tagAllChildren(Panel* panel, Row* parent) {
    parent->tag = true;
    int parent_id = parent->id;
-   for (int i = 0; i < Panel_size(panel); i++) {
+   const int n = Panel_size(panel);
+   for (int i = 0; i < n; i++) {
       Row* row = (Row*) Panel_get(panel, i);
       if (!row->tag && Row_isChildOf(row, parent_id)) {
          tagAllChildren(panel, row);
@@ -160,7 +161,8 @@ static bool collapseIntoParent(Panel* panel) {
       return false;
 
    int parent_id = Row_getGroupOrParent(r);
-   for (int i = 0; i < Panel_size(panel); i++) {
+   const int n = Panel_size(panel);
+   for (int i = 0; i < n; i++) {
       Row* row = (Row*) Panel_get(panel, i);
       if (row->id == parent_id) {
          row->showChildren = false;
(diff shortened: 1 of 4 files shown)


Action.c
```diff
@@ -641,9 +644,11 @@ static Htop_Reaction actionBacktrace(State *st) {
 
    Vector* processes = Vector_new(Class(Process), false, VECTOR_DEFAULT_SIZE);
    if (selectedProcess && !Process_isUserlandThread(selectedProcess)) {
-      for (int i = 0; i < Vector_size(allProcesses); i++) {
-         Process* process = (Process *)Vector_get(allProcesses, i);
-         if (process && Process_getThreadGroup(process) == Process_getThreadGroup(selectedProcess)) {
+      const int thgid = Process_getThreadGroup(selectedProcess);
+      const int pvecsize = Vector_size(allProcesses);
```
Explorer09, line 648: A more readable variable name:
```suggestion
      const int numProcesses = Vector_size(allProcesses);
```

Action.c
```diff
@@ -914,8 +920,10 @@ static Htop_Reaction actionHelp(State* st) {
 }
 
 static Htop_Reaction actionUntagAll(State* st) {
-   for (int i = 0; i < Panel_size((Panel*)st->mainPanel); i++) {
-      Row* row = (Row*) Panel_get((Panel*)st->mainPanel, i);
+   Panel* panel = (Panel*)st->mainPanel;
```
Explorer09, line 923: We probably don't need a separate variable for `(Panel*)st->mainPanel`. The pointer stays unchanged during the loop, so the compiler should be able to optimize it.

Header.c
```diff
@@ -66,7 +66,8 @@ void Header_setLayout(Header* this, HeaderLayout hLayout) {
    } else {
       // move meters from to-be-deleted columns into last one
       for (size_t i = newColumns; i < oldColumns; i++) {
-         for (int j = this->columns[i]->items - 1; j >= 0; j--) {
+         const int n = this->columns[i]->items;
+         for (int j = n - 1; j >= 0; j--) {
```
Explorer09, line 70: Unnecessary change for optimization.

Header.c
```diff
@@ -182,8 +183,10 @@ Meter* Header_addMeterByClass(Header* this, const MeterClass* type, unsigned int
 
 void Header_reinit(Header* this) {
    Header_forEachColumn(this, col) {
-      for (int i = 0; i < Vector_size(this->columns[col]); i++) {
-         Meter* meter = (Meter*) Vector_get(this->columns[col], i);
+      const Vector* meters = this->columns[col];
+      const int n = Vector_size(meters);
```
Explorer09, line 187: Just curious, how can a compiler fail to optimize the `Vector_size` call here?

`Vector_size` is an inline function when `NDEBUG` is defined (i.e. when htop is not built for debug), and `Vector_size` merely retrieves the `items` member of the `Vector` object, which should be a constant throughout the loop.

fasterit:
The core team consensus is to leave such optimizations to the compiler for better readability of the code to humans. Closing.

PR #2040, opened 2026-07-08
Screens/Tabs/Signals panels: move const calc expression out of loop

GermanAizek:
Reference: https://github.com/htop-dev/htop/pull/2038


fasterit:
The core team consensus is to leave such optimizations to the compiler for better readability of the code to humans. Closing.

PR #2041, opened 2026-07-08
ColorsPanel: for loop check condition simplify for optimize

GermanAizek:
@natoscott, isn't it easier to check each step in the loops with this PR condition?
Access by index in array heavy operation vs check i index with const var LAST_COLORSCHEME

Clion IDE have diff view Code and Assembly with compiler flags.
Out of interest, I can show how much less x86 instructions it should give on -O3.

### DIFF
--- a/ColorsPanel.c
+++ b/ColorsPanel.c
@@ -63,7 +63,7 @@ static HandlerResult ColorsPanel_eventHandler(Panel* super, int ch) {
          assert(mark >= 0);
          assert(mark < LAST_COLORSCHEME);
 
-         for (int i = 0; ColorSchemeNames[i] != NULL; i++)
+         for (int i = 0; i < LAST_COLORSCHEME; i++)
             CheckItem_set((CheckItem*)Panel_get(super, i), false);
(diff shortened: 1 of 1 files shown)


fasterit:
@GermanAizek: Please show how much runtime / code that change saves.

GermanAizek:
> @GermanAizek: Please show how much runtime / code that change saves.

I can show assembler listing on -O3 using x86-64 instructions, I can calculate how many processor cycles are spent on this cycle, visually it will look like in this PR https://github.com/htop-dev/htop/pull/2055

GermanAizek:
I checked the assembler listing on latest versions GCC 16.1.0 and Clang 21.1.8, and for now it's better to leave it the same, since on -O3 it unroll loop itself, but my version makes CPU branch predictor work with `jne` instruction, which is worse for ALU and SIMD blocks.

image

image

GermanAizek:
@fasterit indeed, checking for NULL makes the loop more vectorized and the compiler can optimize better for AVX instructions, I'll check other interesting loops for a similar case.

But usually basic simplest loop is best optimized by C/C++ compilers when the `int` is an index and compared to another number (for indexed loop)

PR #2043, opened 2026-07-10
Store htop_history in XDG state directory

### LINKED ISSUE #1944: `htop_history` storage location and `$XDG_STATE_HOME`
Since 19fc20c40d420d3fca2c6215097292762d635729 , a new `htop_history` file is introduced for recording history of search queries. However, it might not the most ideal location for storing the history file next to `htoprc` (wherever the `htoprc` file is located).

The [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir/latest/) (0.8) has specified a `$XDG_STATE_HOME` directory, which I could argue would be a better location. Unfortunately the spec from XDG is vague. I'm not sure it was intended to store what's analogous to `/var/log` but for user files. Anyway, this issue report is to encourage discussion about using `$XDG_STATE_HOME` for the thing.

To be precise, I suggest if the `$XDG_STATE_HOME` is to be used, then

1. the history file should be stored as `$XDG_STATE_HOME/htop/htop_history` (fallback to `$HOME/.local/state/htop/htop_history` if the `$XDG_STATE_HOME` variable is undefined);
2. for backward compatibility, the `$XDG_CONFIG_HOME/htop/htop_history` might still be read, but `$XDG_STATE_HOME/htop/htop_history` would take precedence if it exists.

P.S. About the reason of separating `$XDG_CONFIG_HOME` and `$XDG_STATE_HOME`:

This is my personal understanding, and not written within the XDG spec. The Unix systems usually separate configuration files and variable files into different directories (see the Filesystem Hierarchy Standard), `/etc` and `/var` respectively. The assumption was that applications might frequently update the files in `/var` but would only update files in `/etc` when there is a setting change system-wide. `/etc` could be mounted read-only by administrators if they don't intend to change the settings throughout the applications' runtime (e.g. kiosk machines), but `/var` should be writable unless there's a special need. That justified the separation of the config directory and the state directory.

Stunkymonkey:
fixes: #1944

The history file is runtime state, not configuration, so it belongs in $XDG_STATE_HOME/htop/htop_history (defaulting to ~/.local/state/htop/htop_history) rather than alongside htoprc. A one-time rename migration moves an existing history file from the config dir to the new state dir on first run.

Assisted-by: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/CommandLine.c
+++ b/CommandLine.c
@@ -472,19 +472,8 @@ int CommandLine_run(int argc, char** argv) {
    if (flags.commFilter)
       setCommFilter(&state, &(flags.commFilter));
 
-   /* Set up shared search/filter history, stored next to the config file */
-   const char* rcPath = settings->filename;
-   const char* lastSlash = strrchr(rcPath, '/');
-   char historyPath[PATH_MAX];
-   if (lastSlash) {
-      int dirLen = (int)(lastSlash - rcPath + 1);
-      xSnprintf(historyPath, sizeof(historyPath), "%.*s" "htop_history", dirLen, rcPath);
-   } else {
-   /* no history file saved unless we have a sane rcPath */
-      historyPath[0] = '\0';
-   }
-
-   IncSet_setHistoryFile(panel->inc, historyPath);
+   /* Set up shared search/filter history in XDG state directory */
+   IncSet_setHistoryFile(panel->inc, settings->historyFilename);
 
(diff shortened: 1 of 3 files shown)


fasterit:
We discussed this topic on IRC and the consensus is to keep the `.htop_history` location as is. Closing this PR and the related issue.

PR #2045, opened 2026-07-15
linux: fix reading zram values

eworm-de:
This is a follow-up on an earlier commit 'linux: avoid repeated zram ENOENT probes'.
Actually `/sys/block/zramN` are symlinks, so setting O_NOFOLLOW breaks this. Let's drop the option.

Fixes: 9fe974b823207e0f19a7bc4ae9cc31eba6e6cd47

### DIFF
--- a/linux/LinuxMachine.c
+++ b/linux/LinuxMachine.c
@@ -292,7 +292,7 @@ static bool LinuxMachine_isZramBlockName(const char* name) {
 
 static void LinuxMachine_scanZramDevice(openat_arg_t blockDirFd, const char* name, memory_t* totalZram, memory_t* usedZramComp, memory_t* usedZramOrig) {
 #ifdef HAVE_OPENAT
-   int zramDirFd = Compat_openat(blockDirFd, name, O_DIRECTORY | O_PATH | O_NOFOLLOW);
+   int zramDirFd = Compat_openat(blockDirFd, name, O_DIRECTORY | O_PATH);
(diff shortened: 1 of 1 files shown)


eworm-de:
Well, alternatively we could look into `/sys/devices/virtual/block/`...

Wondering if the paths and/or symlinks changed with recent linux versions?

Komzpa (approved):
Thanks for catching this and for opening htop-dev/htop#2045.

`/sys/block/zramN` has been zram's documented sysfs interface since the
[original 2010 sysfs conversion](https://github.com/torvalds/linux/commit/33863c21e69e9bd988b63580ada0b729c1d90215),
and the current kernel ABI still documents
[`disksize` and `mm_stat` there](https://github.com/torvalds/linux/blob/master/Documentation/ABI/testing/sysfs-block-zram).

On current kernels, entries under `/sys/block` are symlinks, so
`openat(..., O_DIRECTORY | O_PATH | O_NOFOLLOW)` fails with `ENOTDIR` before
the relative `disksize` and `mm_stat` reads can run. Dropping `O_NOFOLLOW`
fixes that while retaining the dirfd-relative reads.

I would keep `/sys/block`: switching to `/sys/devices/virtual/block` would bind
htop to the backing sysfs topology instead of the documented userspace ABI.
The one-line fix looks correct to me. Thank you!

eworm-de:
Thanks for revealing the details. Sounds like keeping the documented path is the correct approach.

PR #2047, opened 2026-07-18
Add zswap usage and compression meters

ABHIRAMSHIBU:
## Summary

Add Linux zswap visibility to htop without changing existing memory or swap accounting.

This introduces:

- A `Zswap` bar meter showing compressed pool usage against the configured pool limit.
- A `ZswapStats` meter showing compressed usage, original size, and compression ratio.
- Detection of zswap being disabled.
- Clear `disabled` and `unavailable` states when appropriate.
- Support for systems where the pool limit is unavailable.

The pool limit is derived from the configured `max_pool_percent` value and total system memory.

## Testing

- `make -j4`
- `make check`
- `git diff --check`
- Manually verified zswap-enabled and zswap-disabled states.

### DIFF
--- a/linux/LinuxMachine.c
+++ b/linux/LinuxMachine.c
@@ -141,6 +141,8 @@ static void LinuxMachine_scanMemoryInfo(LinuxMachine* this) {
    memory_t sreclaimableMem = 0;
    memory_t zswapCompMem = 0;
    memory_t zswapOrigMem = 0;
+   bool zswapCompAvailable = false;
+   bool zswapOrigAvailable = false;
 
    FILE* file = fopen(PROCMEMINFOFILE, "r");
    if (!file)
@@ -153,6 +155,15 @@ static void LinuxMachine_scanMemoryInfo(LinuxMachine* this) {
          if (String_startsWith(buffer, label)) {                             \
             memory_t parsed_;                                                \
             if (sscanf(buffer + strlen(label), "%llu kB", &parsed_) == 1) {  \
+               (variable) = parsed_;                                         \
+            }                                                                \
+            break;                                                           \
+         } else (void) 0 /* Require a ";" after the macro use. */
+      #define tryReadFlag(label, variable, flag)                             \
+         if (String_startsWith(buffer, label)) {                             \
+            memory_t parsed_;                                                \
+            (flag) = sscanf(buffer + strlen(label), "%llu kB", &parsed_) == 1; \
+            if (flag) {                                                       \
                (variable) = parsed_;                                         \
             }                                                                \
             break;                                                           \
@@ -186,11 +197,12 @@ static void LinuxMachine_scanMemoryInfo(LinuxMachine* this) {
             }
             break;
          case 'Z':
-            tryRead("Zswap:", zswapCompMem);
-            tryRead("Zswapped:", zswapOrigMem);
+            tryReadFlag("Zswap:", zswapCompMem, zswapCompAvailable);
+            tryReadFlag("Zswapped:", zswapOrigMem, zswapOrigAvailable);
             break;
       }
 
+      #undef tryReadFlag
       #undef tryRead
    }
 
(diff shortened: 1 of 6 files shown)


linux/LinuxMachine.c
```diff
@@ -214,10 +226,32 @@ static void LinuxMachine_scanMemoryInfo(LinuxMachine* this) {
...
 
+static void LinuxMachine_scanZswapInfo(LinuxMachine* this) {
+   ZswapStats* zswap = &this->zswap;
+   char buffer[64];
+
+   /* If the parameter is unavailable, counters from /proc/meminfo imply enabled. */
+   zswap->enabled = zswap->available;
+   ssize_t enabledRead = Compat_readfile("/sys/module/zswap/parameters/enabled", buffer, sizeof(buffer));
+   if (enabledRead > 0 && (buffer[0] == 'N' || buffer[0] == 'n'))
+      zswap->enabled = false;
```
BenBE: A shorter version would be:

```suggestion
   ssize_t enabledRead = Compat_readfile("/sys/module/zswap/parameters/enabled", buffer, sizeof(buffer));
   zswap->enabled = (enabledRead > 0 && (buffer[0] == 'N' || buffer[0] == 'n')) ? false : zswap->available;
```

But even that won't be ideal, because it still tries to read those files below when we already determined Zram to be unavailable.

linux/ZswapMeter.c
```diff
@@ -0,0 +1,200 @@
...
+      xSnprintf(this->txtBuffer, sizeof(this->txtBuffer), "disabled");
+      return;
+   }
+
+   this->values[ZSWAP_STATS_METER_COMPRESSED] = zswap->usedZswapComp;
+   this->values[ZSWAP_STATS_METER_ORIGINAL] = zswap->usedZswapOrig;
+
+   char compressedBuffer[50];
+   char originalBuffer[50];
+   char ratioBuffer[50];
```
BenBE: Reduce buffer size to what's actually needed here, which is at most 20 characters; actually even further because we optimize for short output. Even using 16 chars here should be fine, which saves about 100 bytes on the stack and has better alignment.

linux/ZswapMeter.c
```diff
@@ -0,0 +1,200 @@
...
+      RichString_writeAscii(out, CRT_colors[FAILED_READ], "unavailable");
+      return;
+   }
+
+   if (!zswap->enabled) {
+      RichString_writeAscii(out, CRT_colors[METER_SHADOW], "disabled");
+      return;
+   }
+
+   char buffer[50];
```
BenBE: Same here. 16 bytes should be sufficient.

BenBE (changes requested):

linux/LinuxMachine.c
ABHIRAMSHIBU: I am adding an early return for this.

linux/ZswapMeter.c
ABHIRAMSHIBU: Done

linux/ZswapMeter.c
ABHIRAMSHIBU: Done

linux/ZswapMeter.h
```diff
@@ -0,0 +1,27 @@
...
+#define HEADER_ZswapMeter
+/*
+htop - ZswapMeter.h
+(C) 2026 Abhiram Shibu
+Released under the GNU GPLv2+, see the COPYING file
+in the source distribution for its full text.
+*/
+
+#include "Meter.h"
+
```
BenBE, line 11: Second blank line after last header.

linux/LinuxMachine.c
```diff
@@ -186,12 +197,13 @@ static void LinuxMachine_scanMemoryInfo(LinuxMachine* this) {
...
          case 'Z':
-            tryRead("Zswap:", zswapCompMem);
-            tryRead("Zswapped:", zswapOrigMem);
+            tryReadFlag("Zswap:", zswapCompMem, zswapCompAvailable);
+            tryReadFlag("Zswapped:", zswapOrigMem, zswapOrigAvailable);
             break;
       }
 
       #undef tryRead
+      #undef tryReadFlag
```
BenBE: Optional: Undefine in reverse define order.

```suggestion
      #undef tryReadFlag
      #undef tryRead
```

linux/ZswapStats.h
```diff
@@ -7,9 +7,19 @@ Released under the GNU GPLv2+, see the COPYING file
...
 
 typedef struct ZswapStats_ {
+   /* whether both zswap counters are available from /proc/meminfo */
+   bool available;
+   /* whether zswap is enabled */
+   bool enabled;
+   /* whether totalZswapPool was obtained from the zswap module parameter */
+   bool hasPoolLimit;
+   /* maximum configured size of the zswap pool */
+   memory_t totalZswapPool;
```
BenBE: Prefer to put the `bool` fields after the `memory_t` fields to avoid unnecessary padding of the structure.

BenBE (approved):
LGTM. Only minor notes remaining; nothing critical.

linux/ZswapMeter.h
ABHIRAMSHIBU: Done

linux/LinuxMachine.c
ABHIRAMSHIBU: Swapped it.

linux/ZswapStats.h
ABHIRAMSHIBU: Makes sense, I should have done this. Added the fix in the latest patch.

ABHIRAMSHIBU:
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> ⚠️ Outside diff range comments (1)
> 
> > linux/LinuxMachine.c (1)> `162-170`: _📐 Maintainability & Code Quality_ | _🟠 Major_ | _⚡ Quick win_
> > > **Avoid adding another function-like macro.**
> > > `tryReadFlag` duplicates parsing logic and hides `break` control flow inside a macro. Replace it with a small static helper or explicit parsing code in the `Z` case so the control flow remains visible.
> > > As per coding guidelines, avoid function-like macros.
> > > 
> > > _Source: Coding guidelines_
> 
> ℹ️ Review info
> ⚙️ Run configuration
> 
> 📥 Commits
> 
> 📒 Files selected for processing (3)

Initially I was thinking to move it to a function above the current function as `LinuxMachine_tryReadFlag()`, then felt to duplicate as I didnt know if it was a good idea to do that because it didnt feel right spot to put it after reading the file.

BenBE:
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> ⚠️ Outside diff range comments (1)
> 
> > linux/LinuxMachine.c (1)> `162-170`: _📐 Maintainability & Code Quality_ | _🟠 Major_ | _⚡ Quick win_
> > > **Avoid adding another function-like macro.**
> > > `tryReadFlag` duplicates parsing logic and hides `break` control flow inside a macro. Replace it with a small static helper or explicit parsing code in the `Z` case so the control flow remains visible.
> > > As per coding guidelines, avoid function-like macros.
> > > 
> > > _Source: Coding guidelines_
> 
> ℹ️ Review info

@coderabbitai These function-like macros for parsing are fine when defined for a very limited, local purpose. Copying the existing macro and extending it for the task at hand is fine.

@ABHIRAMSHIBU Can you rebase/squash your commits. Bug fixes to previous parts of a PR should be absorbed into where those issues were introduced.

PR #2050, opened 2026-07-25
exit follow mode when search or filter is cancelled

OldUser101:
condition change in bbec22d7592ea6cf26dfd615be649855a4e0f22b caused panel state during search to persist after the search was cancelled (e.g. escape key). previous behaviour relied on the absence of the `HTOP_KEEP_FOLLOWING` reaction to stop following, which is no longer the case.

explicitly setting `host->activeTable->following` to `-1` now, has the same effect as not setting `HTOP_KEEP_FOLLOWING` did before bbec22d7592ea6cf26dfd615be649855a4e0f22b.

### DIFF
--- a/MainPanel.c
+++ b/MainPanel.c
@@ -121,6 +121,8 @@ static HandlerResult MainPanel_eventHandler(Panel* super, int ch) {
          reaction |= HTOP_KEEP_FOLLOWING;
          if (confirmedSearch)
             reaction |= HTOP_REFRESH;
+      } else {
+         host->activeTable->following = -1;
       }
       result = HANDLED;
    } else if (ch == 27) {


fasterit:
Hm, do we want that? Do we want to guard it with the ` stickyFollow` setting?

OldUser101:
The behaviour I would expect is not to follow unless there is either a search match, or the search has been confirmed (e.g. pressing Enter). 

At the very least cancelling a failed search shouldn't keep following whatever was selected by  it, probably in error.

in regard to `stickyFollow`, I'm not sure if it's really relevant here since it currently serves a different purpose of keeping the followed item visible.

I also feel the current behaviour contradicts the intended behaviour of 7d7baebdcb40de2573f8c32d628e0808ee91d3c4: a process being followed when confirmed in the search. Currently, every search match, failed or not, will be followed whether the search was confirmed or aborted.

PR #2057, opened 2026-08-04
Const var make more optimized loops in -O3 GCC/Clang

GermanAizek:
Many thanks @fasterit for making me check the loop, I found such a cool optimization on GCC and Clang compilers with the -O3 optimization flag. Amazing results, it's strange that the compiler itself doesn't optimize for -O3.

Example with `actionUntagAll()` function:

screen

Opt A my branch (with `const int size`):
```asm
 actionUntagAll:
    subq    $24, %rsp
    movq    %rbp, 8(%rsp)
    movq    %rdi, %rbp
    movq    8(%rdi), %rdi
    call    Panel_size@PLT
    testl    %eax, %eax
    jle    .L33
    movq    %rbx, (%rsp)
    xorl    %ebx, %ebx
    movq    %r12, 16(%rsp)
    movl    %eax, %r12d
.L34:
    movq    8(%rbp), %rdi
    movl    %ebx, %esi
    addl    $1, %ebx
    call    Panel_get@PLT
    movb    $0, 29(%rax)
    cmpl    %ebx, %r12d
    jne    .L34
    movq    (%rsp), %rbx
    movq    16(%rsp), %r12
.L33:
    movq    8(%rsp), %rbp
    movl    $1, %eax
    addq    $24, %rsp
    ret 
```

screen2

Op B master branch (without `const int size`):
```asm
 actionUntagAll:
    pushq    %rbp
    movq    %rdi, %rbp
    pushq    %rbx
    xorl    %ebx, %ebx
    subq    $8, %rsp
    jmp    .L33
.L34:
    movq    8(%rbp), %rdi
    movl    %ebx, %esi
    addl    $1, %ebx
    call    Panel_get@PLT
    movb    $0, 29(%rax)
.L33:
    movq    8(%rbp), %rdi
    call    Panel_size@PLT
    cmpl    %ebx, %eax
    jg    .L34
    addq    $8, %rsp
    movl    $1, %eax
    popq    %rbx
    popq    %rbp
    ret 
```

Why separate const int (option A) faster

In Option A, I saved the size to a constant size before the loop started.
In the assembler, we see that the call `Panel_size@PLT` occurs only once before the start of the loop (before the label `.L34`). The result is stored in the `%r12d` register, and a simple and very fast comparison with the register takes place inside the loop itself: `cmpl %ebx, %r12d`.

For each iteration of the loop in Variant A, there is 1 function call.:

    `call Panel_get@PLT`

Why Option B vanilla code slower

In Option B, the exit condition of the loop is calculated anew at each iteration.
There is a label in the assembly code `.L33` (which is a loop condition check) contains:
Code snippet
```asm
.L33:
    movq 8(%rbp), %rdi
    call Panel_size@PLT <--- Function call!
    cmpl    %ebx, %eax
    jg      .L34
```
This means that for each iteration of the loop in Option B, there are 2 function calls.:

    `call Panel_get@PLT`

    `call Panel_size@PLT`

### DIFF
--- a/Action.c
+++ b/Action.c
@@ -138,7 +138,8 @@ bool Action_setUserOnly(const char* userName, uid_t* userId) {
 static void tagAllChildren(Panel* panel, Row* parent) {
    parent->tag = true;
    int parent_id = parent->id;
-   for (int i = 0; i < Panel_size(panel); i++) {
+   const int size =  Panel_size(panel);
+   for (int i = 0; i < size; i++) {
       Row* row = (Row*) Panel_get(panel, i);
       if (!row->tag && Row_isChildOf(row, parent_id)) {
          tagAllChildren(panel, row);
@@ -160,8 +161,9 @@ static bool collapseIntoParent(Panel* panel) {
    if (!r)
       return false;
 
-   int parent_id = Row_getGroupOrParent(r);
-   for (int i = 0; i < Panel_size(panel); i++) {
+   const int parent_id = Row_getGroupOrParent(r);
+   const int size =  Panel_size(panel);
+   for (int i = 0; i < size; i++) {
       Row* row = (Row*) Panel_get(panel, i);
       if (row->id == parent_id) {
          row->showChildren = false;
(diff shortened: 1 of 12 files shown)


GermanAizek:
In short calling a function (especially via PLT — Procedure Linkage Table) is an expensive operation for CPU compared to simply reading from ALU register. It requires preparing arguments, jumping through memory, changing the stack pointer, and returning.

If htop panel has, for example, 10lines :

for loop A will call Panel_size 1 time
for loop B will call Panel_size 11 times

Increddible.

BenBE:
The const part is not the main issue here, but the repeated call.

Often times doing

```
for(int i = 0, size = Panel_size(panel); i < size; i++) {…
```

works just the same and also limits the scope of `size` …

fasterit:
I like @BenBE's style better as it limits the scope of the temp variable.

@GermanAizek: Did you do some benchmarking (perf) how much the loop optimizations save overall?
I.e. htop 10s wallclock time runs and compare CPU cycle counts?

Issue #1913, opened 2026-02-21
"A signal 6 (Aborted) was received" after adding  the CGROUP column to a screen


jpsalvesen:
Steps to reproduce:

1. Add the CGROUP column to a screen
2. Go to the screen and wait for a second

---

htop version: 3.4.1
uname output: Linux learner 6.18.12-200.fc43.x86_64 #1 SMP PREEMPT_DYNAMIC Mon Feb 16 18:58:26 UTC 2026 x86_64 GNU/Linux
Distro: Fedora 43

Backtrace information:
----------------------
htop(CRT_handleSIGSEGV+0x131) [0x562121f4e4f1]
/lib64/libc.so.6(+0x1a290) [0x7f8b443ba290]
/lib64/libc.so.6(+0x743cc) [0x7f8b444143cc]
/lib64/libc.so.6(gsignal+0x1e) [0x7f8b443ba15e]
/lib64/libc.so.6(abort+0x26) [0x7f8b443a16d0]
htop(+0x2a522) [0x562121f6a522]
htop(xSnprintf+0xe3) [0x562121f6b153]
htop(+0x309b1) [0x562121f709b1]
htop(Row_display+0x40) [0x562121f4f4a0]
htop(Panel_draw+0x30d) [0x562121f5639d]
htop(ScreenManager_run+0x16c) [0x562121f6629c]
htop(CommandLine_run+0x8b8) [0x562121f4cf98]
/lib64/libc.so.6(+0x35b5) [0x7f8b443a35b5]
/lib64/libc.so.6(__libc_start_main+0x88) [0x7f8b443a3668]
htop(_start+0x25) [0x562121f41dc5]

BenBE:
This looks like an issue trying to format some of the data involved, as CGroup entries can get kinda lengthy. Does the same issue happen with the CCGROUP column too?

Also, cann you build a debug version and break on the abort? It'd be nice to get a `bt full` stacktrace from GDB for this issue. TIA.

jpsalvesen:
Sorry for the silence - been doing other stuff.

I just built a debug build from commit 88e00aa53c1cbf727069ace4123072d4fa9ae7c3 and reproduced the issue right away.

Built using `/autogen.sh && ./configure --enable-debug  --prefix=/home/user/tmp && make`

Execution yielded

```
./htop


FATAL PROGRAM ERROR DETECTED
============================
Please check at https://htop.dev/issues whether this issue has already been reported.
[...]

PR #2059, opened 2026-08-05
lpt: fix misprint fgets buffer size in cgroup parsing

GermanAizek:
https://github.com/htop-dev/htop/blob/main/linux/LinuxProcessTable.c#L935-L941

LinuxProcessTable_readCGroupFile() declares an inner buffer of PROC_LINE_LENGTH + 1 bytes but passes PROC_LINE_LENGTH to fgets(), telling it that the buffer is one byte smaller than it actually is

### DIFF
--- a/linux/LinuxProcessTable.c
+++ b/linux/LinuxProcessTable.c
@@ -938,7 +938,7 @@ static void LinuxProcessTable_readCGroupFile(LinuxProcess* process, openat_arg_t
    size_t left = PROC_LINE_LENGTH;
    while (!feof(file) && left > 0) {
       char buffer[PROC_LINE_LENGTH + 1];
-      const char* ok = fgets(buffer, PROC_LINE_LENGTH, file);
+      const char* ok = fgets(buffer, sizeof(buffer), file);
       if (!ok)
          break;
 


GermanAizek:
Wtf `CI / build-solaris-latest-gcc` Cancelled after 20m

GermanAizek:
> Wtf `CI / build-solaris-latest-gcc` Cancelled after 20m

  Download: 2220/3041 items   910.2/1252.2MB  72% complete (809k/s)
  Download: 2220/3041 items   912.1/1252.2MB  72% complete (668k/s)
  Download: 2220/3041 items   914.4/1252.2MB  73% complete (464k/s)
  Download: 2220/3041 items   916.4/1252.2MB  73% complete (436k/s)
  Download: 2220/3041 items   919.8/1252.2MB  73% complete (563k/s)
  Download: 2220/3041 items   921.8/1252.2MB  73% complete (553k/s)
  Download: 2220/3041 items   924.4/1252.2MB  73% complete (460k/s)
  Download: 2220/3041 items   927.9/1252.2MB  74% complete (608k/s)
  Download: 2220/3041 items   929.6/1252.2MB  74% complete (567k/s)
  Error: The operation was canceled.

ravi-arnan (commented):
Correct, and safe either way: `buffer` is `PROC_LINE_LENGTH + 1` bytes, so
`sizeof(buffer)` is within bounds and the old form simply left the last byte
unused. No overflow before or after.

Since I had a Linux box to hand I checked when the extra byte can actually be
observed, which turns out to be narrower than it looks.

It only matters for a `/proc/<pid>/cgroup` line of 4096 bytes or more. Those are
constructible: a delegated user cgroup plus 21 nested 200-character components
gives a 4099-byte line. It has to be built with relative `chdir` because the
absolute path is past `PATH_MAX`, but the kernel prints the full line regardless:

```
$ python3 -c "..."   # chdir down 21 levels under user@1000.service, write pid
line bytes: 4099
```

On that process, both `main` and this branch abort before the byte matters:

```
main    exit=134   FATAL PROGRAM ERROR / signal 6
pr2059  exit=134   same
```

That is #1913, the CGROUP column width crash that #2051 fixes, not anything
introduced here. Rebuilding this branch with #2051's buffer fix on top, and
rebuilding the same tree with only that fix and the `fgets` line reverted, both
run clean and render the same 255 characters, since the column truncates there
anyway:

```
with fgets(buffer, PROC_LINE_LENGTH)  longest rendered cgroup run = 255
with fgets(buffer, sizeof(buffer))    longest rendered cgroup run = 255
```

So this is hygiene rather than a user-visible fix, and it can only ever be
exercised behind #2051. That is not an argument against merging it, just a note
in case the commit message wants to be accurate about impact.

One adjacent thing this exposed, clearly out of scope for a one-liner: the loop
has no partial-line handling. When a line does exceed the buffer, the next
`fgets` returns the remainder of that same line and the parser treats it as a
new cgroup entry, walking it for two `:` separators that are not there and
appending a separator plus an empty group to `output`.
`LinuxProcessTable_readSmapsFile()` in the same file handles exactly this case
with `skipEndOfLine()`. If anyone cares about >4KB cgroup lines, that is the
real bug in this function, and it would be a separate PR.

Issue #1657, opened 2025-03-31
GPU meter is always zero on a double GPU headed PC


va-al:
**A problem description:** The GPU meter added in my **htop** layout always stays zeroed.  Although at the same time **nvtop** shows existing real load on the GPU.

**Expected behavior:** the htop's GPU counter show real % load.

**Extra information:**

1. I've checked this behavior on three different PCs. On two different one-GPU-equipped PCs (Intel and AMD) everything works as expected and only on a dual-GPU setup (Intel + NVIDIA) I have experienced the unexpected behavior.
2.  Configuration details and screenshots below: 

<details><summary>PC configuration</summary>
<p>

System:
  Host: HOST Kernel: 6.13.8-arch1-1 arch: x86_64 bits: 64
  Desktop: GNOME v: 48.0 Distro: Arch Linux
Machine:
  Type: Desktop System: Micro-Star product: MS-7B48 v: 1.0
    serial: <superuser required>
  Mobo: Micro-Star model: Z370-A PRO (MS-7B48) v: 1.0
    serial: <superuser required> UEFI: American Megatrends v: 2.D3
    date: 11/18/2021
CPU:
  Info: 6-core model: Intel Core i7-8700K bits: 64 type: MT MCP cache:
    L2: 1.5 MiB
  Speed (MHz): avg: 800 min/max: 800/4700 cores: 1: 800 2: 800 3: 800 4: 800
    5: 800 6: 800 7: 800 8: 800 9: 800 10: 800 11: 800 12: 800
Graphics:
  Device-1: NVIDIA TU106 [GeForce RTX 2070] driver: nvidia v: 570.133.07
  Device-2: Logitech C922 Pro Stream Webcam driver: snd-usb-audio,uvcvideo
    type: USB
  Display: wayland server: X.Org v: 24.1.6 with: Xwayland v: 24.1.6
    compositor: gnome-shell driver: X: loaded: nvidia
    gpu: nvidia,nvidia-nvswitch resolution: 3440x1440~144Hz
  API: EGL v: 1.5 drivers: nvidia
    platforms: gbm,wayland,x11,surfaceless,device
  API: OpenGL v: 4.6.0 vendor: nvidia v: 570.133.07 renderer: NVIDIA
    GeForce RTX 2070/PCIe/SSE2
  API: Vulkan v: 1.4.309 drivers: N/A surfaces: xcb,xlib,wayland
  Info: Tools: api: eglinfo, glxinfo, vulkaninfo
[...]

PR #2069, opened 2026-08-14
Linux: read GPU usage from drm-cycles when there is no ns counter

bogdanr:
On my Lunar Lake laptop (Arc 140V, `xe` driver) the GPU meter and the GPU columns
are stuck at `0.0%` no matter what the GPU is doing, while nvtop and gputop show
it pegged.

The reason is that [drm-usage-stats](https://www.kernel.org/doc/html/latest/gpu/drm-usage-stats.html)
allows two ways of reporting engine utilisation:

* `drm-engine-<engine>: <uint> ns` - busy time in nanoseconds
* `drm-cycles-<engine>` plus `drm-total-cycles-<engine>` - busy cycles and
  elapsed cycles, whose ratio is the utilisation

We only parse the first one, and the Xe driver only implements the second one.
There is not a single `drm-engine-*` line to be found:

```
$ grep drm- /proc/$(pidof glxgears)/fdinfo/*
drm-driver:xe
drm-client-id:607
drm-pdev:0000:00:02.0
...
drm-cycles-rcs:29821070
drm-total-cycles-rcs:5044293186233
drm-cycles-ccs:571
drm-total-cycles-ccs:5044293186233
...
```

Since Xe is what recent Intel hardware gets by default, this will affect more
and more machines. `intel_gpu_top` has the same problem, it refuses to start and
tells you to use `gputop` instead.

### What the patch does

`GPU.c` now also parses `drm-cycles-*` and `drm-total-cycles-*`, reusing the
existing client-id/pdev deduplication. Utilisation is `Δcycles / Δtotal-cycles`,
scaled with the sampling interval so it can be added up together with the
nanosecond values everywhere else. Doing it as a ratio means there is no need to
know the GPU clock, which is good, because on this machine it swings between
400 MHz and 1950 MHz while sampling.

`drm-total-cycles` is a device wide free running counter that every client
repeats, so it is aggregated with max instead of sum, both per process and per
engine.

Because the raw counters are no longer always nanoseconds, `LinuxProcess` keeps
them separately (`gpu_timeRaw`, `gpu_cycles`, `gpu_totalCycles`) and `gpu_time`
accumulates the observed busy time. For `drm-engine-*` drivers the accumulated
value comes out the same as the old absolute one, except it no longer goes
backwards when a client closes its fds.

Two things in `Platform.c` that got in the way and are fixed here as well: the
meter residue is now derived from the per engine deltas instead of a second
absolute counter, and the first sample is skipped because there is no interval
to divide by yet. That first sample used to display things like `20679d` of GPU
time.

I also pulled the line parsing and the duplicate client bookkeeping into two
small helpers, they were about to be written a third time.

### Testing

One machine per reporting style.

**Arc 140V, `xe`, cycle counters**, compared against `gputop` from
igt-gpu-tools:

```
workload                    htop GPU%   gputop
glxgears (rcs)               46.9 %     47.1 %
compute benchmark (ccs)      99.9 %    100.0 %
idle                          0.0 %      0.0 %
```

With glxgears running, the meter total of 85.6-86.4 % matches gputop adding up
its clients for the same period (Xorg 40.3 % + glxgears 45.3 % + picom 1.7 %).

**Radeon 8060S (Strix Halo), `amdgpu`, nanosecond counters**, to check that the
existing path did not move. Two glmark2 processes rendering off-screen at
1920x1080, `gpu_busy_percent` pinned at 100 %, one sample per second, this
branch next to the same tree without the patch:

```
this branch   96.6 101.4  99.2 100.4 101.4  98.6  99.5 100.6  99.7 101.9  98.0 ...
main          96.7 100.9  99.9  99.6 102.0  97.8 100.8  99.5  99.7 102.6  97.9 ...
```

Per process both glmark2 clients read ~49 % in either build, while an
independent aggregation of the fdinfo counters says 50.08 % and 49.92 %. Killing
the load gives a flat 0.00 % for as long as I watched it.

Builds clean with `-Wall -Wextra` on both machines, and with `--enable-debug`,
where no assertion fired during the runs. Still untested on NVIDIA.

### Two things I noticed but did not touch

`GPUMeter_engineData` holds four engines while Xe exposes five (`rcs`, `vcs`,
`vecs`, `bcs`, `ccs`), so the fifth one always lands in the residue. On this
laptop compute work goes to `ccs`, which means the named bars stay empty while
the total is right. Bumping the array is easy, but it needs a fifth colour and I
did not want to mix that in here.

A process with fds on two devices whose `drm-total-cycles` counters differ will
be slightly off, as only one reference counter is kept per process. That is the
same kind of conflation as #1657.

### DIFF
--- a/linux/GPU.c
+++ b/linux/GPU.c
@@ -45,7 +45,7 @@ static bool is_duplicate_client(const ClientInfo* parsed, ClientID id, const cha
    return false;
 }
 
-static void update_machine_gpu(LinuxProcessTable* lpt, unsigned long long int time, const char* engine, size_t engine_len) {
+static GPUEngineData* get_machine_gpu_engine(LinuxProcessTable* lpt, const char* engine, size_t engine_len) {
    Machine* host = lpt->super.super.host;
    LinuxMachine* lhost = (LinuxMachine*) host;
    GPUEngineData** engineData = &lhost->gpuEngineData;
@@ -60,17 +60,86 @@ static void update_machine_gpu(LinuxProcessTable* lpt, unsigned long long int ti
    if (!*engineData) {
       GPUEngineData* newData = xMalloc(sizeof(*newData));
       *newData = (GPUEngineData) {
-         .prevTime = 0,
-         .curTime  = 0,
-         .key      = xStrndup(engine, engine_len),
-         .next     = NULL,
+         .prevTime        = 0,
+         .curTime         = 0,
+         .prevCycles      = 0,
+         .curCycles       = 0,
+         .prevTotalCycles = 0,
+         .curTotalCycles  = 0,
+         .key             = xStrndup(engine, engine_len),
+         .next            = NULL,
       };
 
       *engineData = newData;
    }
 
-   (*engineData)->curTime += time;
+   return *engineData;
+}
+
+static void update_machine_gpu(LinuxProcessTable* lpt, unsigned long long int time, const char* engine, size_t engine_len) {
+   LinuxMachine* lhost = (LinuxMachine*) lpt->super.super.host;
+
+   get_machine_gpu_engine(lpt, engine, engine_len)->curTime += time;
    lhost->curGpuTime += time;
+}
+
+static void update_machine_gpu_cycles(LinuxProcessTable* lpt, unsigned long long int cycles, const char* engine, size_t engine_len) {
+   get_machine_gpu_engine(lpt, engine, engine_len)->curCycles += cycles;
+}
+
+/* drm-total-cycles-* is a device global counter, all clients of a device
+ * report the same value, so aggregate by taking the maximum. */
+static void update_machine_gpu_total_cycles(LinuxProcessTable* lpt, unsigned long long int totalCycles, const char* engine, size_t engine_len) {
+   GPUEngineData* engineData = get_machine_gpu_engine(lpt, engine, engine_len);
+
+   if (totalCycles > engineData->curTotalCycles)
+      engineData->curTotalCycles = totalCycles;
+}
+
+static bool count_section(enum section_state* sstate, ClientID client_id, const char* pdev, const ClientInfo* parsed_ids) {
+   if (*sstate == SECST_UNKNOWN) {
+      if (client_id != INVALID_CLIENT_ID && !is_duplicate_client(parsed_ids, client_id, pdev))
+         *sstate = SECST_NEW;
+      else
+         *sstate = SECST_DUPLICATE;
+   }
+
+   return *sstate == SECST_NEW;
+}
+
+/*
+ * Parses a "<prefix><engine>: <value><unit>" line, e.g. "engine-rcs: 1234 ns".
+ * An empty unit requires the value to be the last item on the line.
+ */
+static bool parse_engine_value(const char* line, const char* prefix, const char* unit,
+                               const char** engine, size_t* engine_len, unsigned long long int* value) {
+   const char* engineStart = line + strlen(prefix);
+
+   const char* delim = strchr(engineStart, ':');
+   if (!delim)
+      return false;
+
+   const char* numStart = delim + 1;
+   while (isspace((unsigned char)*numStart))
+      numStart++;
+
+   /* strtoull() would accept a sign and wrap the result around */
+   if (!isdigit((unsigned char)*numStart))
+      return false;
+
+   char* endptr;
+   errno = 0;
+   unsigned long long int parsed = strtoull(numStart, &endptr, 10);
+   if (errno != 0)
+      return false;
+
+   if (unit[0] ? !String_startsWith(endptr, unit) : *endptr != '\0')
+      return false;
+
+   *engine = engineStart;
+   *engine_len = delim - engineStart;
+   *value = parsed;
+   return true;
 }
 
 /*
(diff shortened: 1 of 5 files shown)


bogdanr:
Force pushed one fixup after getting hold of an AMD machine (Strix Halo,
`amdgpu`), which reports nanoseconds, so it exercises the path I could not test
before.

It found a bug in the first version. `GPU_readProcessData()` only rescans a
process every 5s unless it is marked active, and I had moved that marking behind
"the busy counter moved since the last sample". Processes that own a DRM client
but are momentarily idle then dropped out of the scan, and since the machine wide
totals are summed from the cumulative counters of the processes visited in a
pass, the sum collapsed and came back:

```
GPUDBG dt=1031ms total=100.12%
GPUDBG dt=1031ms total=14580.86%   <- idle client rejoins, its whole history counted
GPUDBG dt=1048ms total=0.00%       <- and drops out again
GPUDBG dt=1030ms total=99.57%
```

Marking a process active whenever it has a counter at all, as on master, keeps
it in the scan and the totals steady. It is only visible on `drm-engine-*`
hardware, because there you have long lived clients sitting on a large
cumulative time; on Xe the ratio resets every sample.

Numbers before and after are in the description, the two builds now agree sample
for sample.

bogdanr:
Two of the three comments are worth fixing, so I did both in `a9b5ab0`.

Raw counters on the error path. 
Non-numeric values.

Consolidating the three branches. - Leaving this one. They look alike, but the
aggregation rules genuinely differ: `engine-` sums, `cycles-` sums,
`total-cycles-` takes a maximum because it is a device wide counter that every
client repeats. The shared five lines are already behind `parse_engine_value()`
and `count_section()`; folding the rest into a callback or a table would make
the difference between the three harder to see, not easier.

bogdanr:
Braces added in `85873cb` - the styleguide is explicit that if one block of an
`if`/`else` has them, all of them do, and I am touching that statement anyway.
(Master gets away without them there, but no reason to keep that going.)

Not doing the rename to `GPU_*`. The two helpers that were already in this file,
`is_duplicate_client()` and `update_machine_gpu()`, are snake_case, and the new
ones were written to match them. Renaming only the new ones would leave the file
half and half, which is worse than either convention on its own, and renaming
all of them is churn that does not belong in a bug fix. Happy to send a separate
patch that converts the whole file if a maintainer wants it.

bogdanr:
The AMD machine came back, so the `drm-engine-*` half is re-verified on
`85873cb` as well: Radeon 8060S / `amdgpu`, two off-screen glmark2 runs holding
`gpu_busy_percent` at 100, meter over eleven samples 97.03 / 101.04 / 99.52 /
100.13 / 101.12 / 98.72 / 100.07 / 100.01 / 99.57 / 102.42 / 97.60 %, and flat
0.00 % once the load is gone. Same picture as before the review fixes.

ravi-arnan (commented):
You have covered `xe` and `amdgpu`. I have an Alder Lake-P / Iris Xe laptop on
`i915`, which is a third driver on the `drm-engine-*` nanosecond path, so I
built the branch and exercised that half.

No regression there. Clean build, no new warnings. With `vblank_mode=0 glxgears`
under load, both upstream `main` and this branch report non-zero,
similar-magnitude GPU% and a monotonically increasing GPU_TIME for the same
process:

```
main     18.1  17.9  28.2  36.8  21.1  17.8  ...
pr2069   15.7  21.1  19.8  21.5  16.0  ...
```

(Samples are not aligned and the load is not steady, so read that as shape, not
as a numeric comparison.)

One thing worth recording for anyone else trying to check the numbers from
outside htop: a naive sum over `/proc/<pid>/fdinfo/*` multiplies the result. My
first cross-check script read roughly four times the real utilisation, because
glxgears holds four fds that are all the same DRM client:

```
/proc/522302/fdinfo/{4,5,6,7}
drm-client-id:  135        (identical in all four)
drm-pdev:       0000:00:02.0
drm-engine-render: 33641346752 ns / 33641937680 / 33642722984 / 33643895844
```

That is exactly what `is_duplicate_client()` already handles, so htop is right
and the script was wrong, but it means any external verification has to dedupe
on `(drm-pdev, drm-client-id)` first.

Two things in the diff looked risky enough to check, and both are fine. Noting
them so nobody else has to re-derive them:

- `parse_engine_value()` with an empty unit requires the number to be the last
  thing on the line (`*endptr != '\0'`). That holds only because lines come from
  `strsep(&buf, "\n")` and therefore carry no trailing newline. It is correct as
  written, but it is a non-local assumption; a comment on the empty-unit case
  would keep a future change to the line splitting from breaking the cycles path
  silently.
- `lp->gpu_time` changes meaning, from "last raw counter value" to "accumulated
  busy time". The displayed GPU_TIME survives that, because on the first sample
  `saturatingSub(new_gpu_time, 0)` is the whole lifetime counter, so the
  accumulator starts from the same place the old raw assignment did.

One actual question, on the cycles branch:

```c
gputimeDelta += (unsigned long long int)((double)cyclesDelta / totalCyclesDelta * monotonicTimeDelta * (1000 * 1000));
```

`monotonicTimeDelta` is the global sampling interval, but a process can have
been skipped for up to 5 seconds by the activity check at the top of the
function. On the `ns` path that skew is visible: the delta covers 5 seconds of
busy time and is divided by one interval, so the percentage overshoots. On the
cycles path it is silently smoothed instead, since the ratio is already
normalised over whatever period the counters covered, and it is then scaled by a
single interval. So the two paths disagree in that corner. Is that intended? I
can see the argument that the cycles behaviour is the more correct of the two,
in which case the `ns` path is the one worth revisiting later.

I cannot test the `drm-cycles-*` half here, since `i915` only exports
`drm-engine-*`.

PR #2076, opened 2026-08-19
Add hardware sensor meters

massimomazzariol:
## Summary

Add Linux hardware sensor meters backed by libsensors.

Each discovered temperature or fan sensor is exposed as an individual DynamicMeter in Available Meters, using the existing htop meter infrastructure rather than a separate status bar or Setup page.

The meters display the current value together with runtime minimum, average and maximum values

## Screenshot

Hardware sensors exposed as individual meters in the standard htop header:

htop-prototype-b-individual

## Implementation

- Discover temperature and fan sensors through libsensors
- Expose one DynamicMeter per sensor
- Use the canonical `chip:feature` sensor ID for identity and persistence
- Sample the additional hardware sensors only while at least one sensor meter is in use
- Re-discover the sensor list if the runtime topology no longer matches the initial list
- Respect the existing Celsius/Fahrenheit setting
- Align values using the longest currently visible sensor label
- Support system-wide and per-user htop sensor label overrides

Custom labels can be configured in:

- `SYSCONFDIR/htop/sensor-labels.conf`
- `$XDG_CONFIG_HOME/htop/sensor-labels.conf` or `~/.config/htop/sensor-labels.conf`

User labels override system-wide labels. If no htop-specific label exists, the meter falls back to the libsensors label, then the chip prefix, and finally the canonical sensor ID

## Testing

Tested on Raspberry Pi / AArch64 with real temperature and fan sensors

Building matrix:

- GCC + sensors + Unicode + `-Werror`
- GCC + no sensors + Unicode + `-Werror`
- Clang + sensors + no Unicode + `-Werror`
- Static GCC + sensors + Unicode + `-Werror`

Runtime testing included:

- temperature and fan meters
- current / min / average / max statistics
- adding and removing sensor meters
- alignment based only on visible sensor meters
- system and user label override precedence
- label configuration parser edge cases

## Current limitations

The existing DynamicMeter header parser limits persisted names, so sensor IDs longer than the supported DynamicMeter name length are not exposed

Please note that newly hot-plugged sensors require restarting htop before they appear in Available Meters.

Further testing would be welcome on more complex systems, especially machines with a broader libsensors layout and server-class hardware

Assisted-by: ChatGPT (OpenAI)


fasterit:
I think this should be a meter like other similar info is being displayed as a meter in `htop`. There is no need for his bar to be special and it breaks the design.

massimomazzariol:
> I think this should be a meter like other similar info is being displayed as a meter in `htop`. There is no need for his bar to be special and it breaks the design.

Thanks @fasterit for the feedback! Well, that makes sense

I had a look at the existing Meter/Header infrastructure, and I think I can rework this PR to use the normal htop meter design instead of introducing a separate status-bar row.

The libsensors discovery and sampling code can stay in the platform/backend layer while the custom status-bar UI, ScreenManager integration and Status Bar setup page can be removed and replaced by a Meter implementation using the existing header configuration.

But before I start reworking it, would you prefer a single Hardware Sensors meter displaying the selected sensors, or individual per-sensor meters exposed through Available Meters?

I'm happy to realign the implementation with whichever approach fits htop's design at its best.

Please let me know and thanks again for the consideration of this feature

fasterit:
I don't know, @massimomazzariol. I have no need for this type of a meter.
@BenBE @natoscott any preference on your sides for how to implement {a, multiple} sensors meter(s)?

natoscott:
@fasterit no strong preferences here (definitely agree this should be a Meter not another status bar though).  @massimomazzariol please post screenshots showing what the options being considered will look like?  Thanks!

BenBE:
Let me chime in here too. I agree with @fasterit and @natoscott that this best should be a meter. I think the most flexible way to go about this is by setting this up in a similar fashion as the screen tabs: In the main Setup menu there's an item where you could create one or more such Sensor Meters, which you then could add via the Meters screen to the header area. Internally those Sensor meters could be handled as Dynamic Meters. That way much of the existing functionality could be re-used.

massimomazzariol:
Thanks everyone for the feedback!

I put together three working UI prototypes to make the trade-offs easier to evaluate visually and all the three screenshots below use the same Raspberry Pi/AArch64 system I own, the same sensor set and the same terminal size.

### **Prototype A**: single aggregate meter, one line

This is the most compact vertically, but it scales poorly horizontally once multiple sensors and min/avg/max values are shown

htop-prototype-a-aggregate



### **Prototype B**: individual per-sensor meters

Each sensor is an independent Meter and can be added, removed and reordered separately.
I used a table-style layout so the values remain aligned even if, for example, a fan goes from 1.200 RPM to 10.000+ RPM:

current │ min / avg / max

htop-prototype-b-individual


### **Prototype C**: single multiline Hardware Sensors meter

This keeps all selected sensors inside one Meter, while preserving the table-style layout and avoiding the horizontal scaling issue of option A

htop-prototype-c-multiline



Personally, I think **protype C** is def the cleanest compromise: it behaves as a single logical Meter, keeps the sensors grouped together, and still gives enough room for current/min/avg/max without turning every sensor into a separate header item. Please bear in mind this is just my personal choice

That said, I’m happy to implement whichever model fits htop’s design best.

One limitation on my side at the moment is hardware coverage: I can currently test this on my Raspberry Pi/AArch64, but I don’t yet have direct access to an Intel/AMD x86_64 system or (even better) a server-class hardware with a broader libsensors layout.

Before considering the final implementation ready, I would definitely want it tested on Intel/AMD systems, systems exposing more sensors, server hardware, and systems without usable libsensors data. If anyone is able to help test the chosen implementation on those platforms, that would be very useful or I'll find a way to make it myself as soon as the design and the implementation are both decided and stable

@BenBE @natoscott @fasterit please let me know which of these three directions would you prefer for the final implementation?

Thanks again!

BenBE:
I think going the route of Option B is possibly the easiest way forward. Would need just one DynamicMeter for each Sensor and would manage entirely without the extra menu for the Setup screen.

What I described above was leaning towards Option C, where you could create your own set of sensors for each Meter; but with the meters on a single line instead.

In related topics: There once was a discussion regarding a Disk Meter for particular disks; which you would configure for which device you want to monitor, cf. #952 and #200 for reference.

massimomazzariol:
OK then! I went with Option B and have now reworked the PR accordingly

The previous status bar implementation has been removed completely. Each libsensors sensor is now exposed as its own `DynamicMeter` and can be managed through the existing Meters UI, so there is no additional Setup menu.

The branch history is now clean and the implementation has been rebased onto the current `main`

I tested the new implementation with GCC and Clang, with and without libsensors/Unicode support, persistence across restarts, runtime sensor updates, min/avg/max statistics, missing sensor handling, sanitizers, and static analysis.

Let me know if this architecture now looks right to you. Thanks a lot!

linux/HardwareSensorDynamicMeter.c
```diff
@@ -0,0 +1,472 @@
...
+   char temperatureUnit) {
+
+   char number[32];
+   xSnprintf(number, sizeof(number), "%.0f", value);
+
+   switch (type) {
+      case SENSOR_TEMPERATURE: {
+         const size_t degreeWidth = CRT_degreeSign[0] ? 1 : 0;
+         xSnprintf(buffer, size, "%s%s%c", number, CRT_degreeSign, temperatureUnit);
+         return strlen(number) + degreeWidth + 1;
```
BenBE: `xSnprintf` already returns the string length; no need for that additional `strlen`.

linux/HardwareSensorDynamicMeter.c
```diff
@@ -0,0 +1,472 @@
...
+static uint32_t HardwareSensorDynamicMeter_hash32(const char* value) {
+   uint32_t hash = UINT32_C(2166136261);
+
+   for (const unsigned char* p = (const unsigned char*)value; *p; p++) {
+      hash ^= *p;
+      hash *= UINT32_C(16777619);
+   }
+
+   return hash;
+}
```
BenBE: Why two independent hash implementations? Where are those magic numbers from?

linux/HardwareSensorDynamicMeter.c
```diff
@@ -0,0 +1,472 @@
...
+   size_t size) {
+
+   const HardwareSensor* sensor = &sensors[index];
+
+   char base[128];
+   HardwareSensor_formatName(base, sizeof(base),
+                             sensor->chip,
+                             sensor->label,
+                             sensor->feature,
+                             sensor->type);
```
BenBE: Can be one line.
```suggestion
   HardwareSensor_formatName(base, sizeof(base), sensor->chip, sensor->label, sensor->feature, sensor->type);
```

If split across multiple lines, only one level:
```suggestion
   HardwareSensor_formatName(base, sizeof(base),
      sensor->chip, sensor->label, sensor->feature, sensor->type);
```

Cf. style guide.

linux/HardwareSensorDynamicMeter.c
```diff
@@ -0,0 +1,472 @@
...
+   }
+
+   if (duplicate && sensor->id)
+      xSnprintf(buffer, size, "%s (%s)", base, sensor->id);
+   else
+      xSnprintf(buffer, size, "%s", base);
+}
+
+static const HardwareSensorDynamicMeter* HardwareSensorDynamicMeter_lookupDefinition(
+   const Meter* meter) {
```
BenBE: One line … There's no strict line length.

```suggestion
static const HardwareSensorDynamicMeter* HardwareSensorDynamicMeter_lookupDefinition(const Meter* meter) {
```

linux/HardwareSensorDynamicMeter.c
BenBE, line 1: Just LibSensorMeter_* for identifiers suffices; makes overall identifiers shorter.

linux/HardwareSensorDynamicMeter.c
```diff
@@ -0,0 +1,472 @@
...
+#ifdef BUILD_WITH_CPU_TEMP
+   if (type == SENSOR_TEMPERATURE && settings->degreeFahrenheit)
+      unit = 'F';
+#else
+   (void)settings;
+   (void)type;
+#endif
+
+   return unit;
+}
```
BenBE: Why extra functions for this? Freedom Units aren't important enough to warrant so much code. Also, the conversion should be available as a function somewhere already.

BenBE (changes requested):
This is just a quick look at things; got some more remarks I'll add later.

massimomazzariol:
Update: I rebased the branch onto current main and consolidated the reworked implementation into a single commit

The current code uses individual DynamicMeters for libsensors temperature/fan sensors; the earlier status-bar implementation is no longer the current architecture. The review points have been addressed in the updated commit

I also reran the final build matrix with GCC/Clang, sensors enabled/disabled, Unicode enabled/disabled as applicable, and --enable-werror

linux/HardwareSensorDynamicMeter.c
massimomazzariol: I now use the return value from xSnprintf directly instead of calling strlen afterwards

linux/HardwareSensorDynamicMeter.c
massimomazzariol: Those were the standard FNV-1a constants. I simplified this to a single 64-bit FNV-1a hash and named the offset basis and prime explicitly. The full canonical sensor ID is still stored separately; the hash is only used to keep the persisted DynamicMeter name compact and stable

linux/HardwareSensorDynamicMeter.c
massimomazzariol: Fixed. The call is now kept on one line

linux/HardwareSensorDynamicMeter.c
massimomazzariol: Fixed. I also applied the same style to the other relevant signatures/calls in the same file

linux/HardwareSensorDynamicMeter.c
massimomazzariol: Done. I renamed HardwareSensorDynamicMeter to LibSensorMeter and shortened the related identifiers accordingly

linux/HardwareSensorDynamicMeter.c
massimomazzariol: Agreed. I removed the extra helper functions and kept the Fahrenheit conversion inline, matching the simpler style used elsewhere in the code

linux/HardwareSensorDynamicMeter.c
natoscott: @massimomazzariol this is generic code, not directly related to the hardware sensor meters, could be moved to somewhere common (Hashtable.h?) and made static inline.

fasterit: This hash is not needed at all. The sensor adaptor + name is unique and not so long that we need another hash Implementation. It is just that AI coding bots **love** hashes.

linux/LibSensorsMeter.c
```diff
@@ -0,0 +1,303 @@
...
+   switch (type) {
+      case SENSOR_TEMPERATURE:
+         xSnprintf(buffer, size, "%*d%s%c", CRT_degreeSign[0] ? 7 : 8, (int)value, CRT_degreeSign, unit);
+         return;
+
+      case SENSOR_FAN:
+         xSnprintf(buffer, size, "%5.0f RPM", value);
+         return;
+   }
+}
```
BenBE, line 221: May leave ´buffer` uninitialized on fallthru.

linux/LibSensorMeter.c
```diff
@@ -0,0 +1,303 @@
...
+   }
+
+   return home ? String_cat(home, CONFIGDIR "/htop/sensor-labels.conf") : NULL;
+}
+
+static void LibSensorMeter_applyLabelOverrides(const char* path, HardwareSensor* sensors, size_t sensorCount) {
+   if (!path)
+      return;
+
+   FILE* file = fopen(path, "r");
```
BenBE: Maybe subject to CWE-59, CWE-200; beware CWE-367.

Compare how the history file is handled.

linux/LibSensorMeter.c
```diff
@@ -0,0 +1,303 @@
...
+
+      DynamicMeter* meter = xCalloc(1, sizeof(*meter));
+
+      String_safeStrncpy(meter->name, sensor->id, sizeof(meter->name));
+      const char* rawDisplayName = sensor->label ? sensor->label : sensor->id;
+      char* displayName = xStrndup(rawDisplayName, sensorDisplayLabelMax);
+
+      xAsprintf(&meter->caption, "%s: ", displayName);
+      xAsprintf(&meter->description, "%s [%s] (%s)",
+                displayName, LibSensorMeter_labelSourceName(sensor->labelSource), sensor->id);
```
BenBE:
```suggestion
      xAsprintf(&meter->description, "%s (%s: %s)",
         displayName, LibSensorMeter_labelSourceName(sensor->labelSource), sensor->id);
```

linux/LibSensorMeter.c
```diff
@@ -0,0 +1,303 @@
...
+      Hashtable_put(table, ++key, meter);
+   }
+
+   LibSensors_freeHardwareSensors(sensors, sensorCount);
+
+   return table;
+}
+
+static void LibSensorMeter_freeFields(ATTR_UNUSED ht_key_t key, void* value, ATTR_UNUSED void* data) {
+
```
BenBE:
```suggestion
```

linux/LibSensorMeter.c
```diff
@@ -0,0 +1,303 @@
...
+#include "HardwareSensor.h"
+#include "Macros.h"
+#include "Settings.h"
+#include "XUtils.h"
+#include "linux/LibSensors.h"
+#include "linux/LinuxMachine.h"
+
+static bool hardwareSensorSamplingRequested = false;
+static size_t hardwareSensorVisibleLabelWidth = 0;
+static uint64_t hardwareSensorLayoutVersion = 0;
```
BenBE: Those are static; no need to be overly verbose with naming them.

`lsm_samplingRequested`, `lsm_labelWidth`, and `lsm_layoutVersion` should suffice.

linux/LibSensorMeter.c
```diff
@@ -0,0 +1,303 @@
...
+      RichString_writeAscii(out, CRT_colors[METER_VALUE_ERROR], "unavailable");
+      return;
+   }
+
+   const Settings* settings = meter->host->settings;
+   const size_t displayNameLength = LibSensorMeter_displayNameLength(definition);
+   const size_t captionPadding =
+      hardwareSensorVisibleLabelWidth > displayNameLength
+         ? hardwareSensorVisibleLabelWidth - displayNameLength
+         : 0;
```
BenBE: Can be one line (cf. naming remark above).

linux/LibSensorsMeter.c
BenBE, line 1: Minor nitpick, but the lib is called `libsensors` (plural);

linux/LibSensors.c
```diff
@@ -129,7 +141,189 @@ int LibSensors_reload(void) {
...
+   if (chipLength < 0)
+      return NULL;
+
+   const size_t chipNameLength = (size_t)chipLength;
+   const size_t featureNameLength = strlen(feature->name);
+   char* id = xMalloc(chipNameLength + featureNameLength + 2);
+
+   sym_sensors_snprintf_chip_name(id, chipNameLength + 1, chip);
+   id[chipNameLength] = ':';
+   memcpy(id + chipNameLength + 1, feature->name, featureNameLength + 1);
```
BenBE: Prefer `xCalloc` and `memcpy` only `featureNameLength` bytes. This ensures a NULL is present, even if `feature->name` was not terminated within `featureNameLength` bytes.

Best option was to skip the `memcpy` entirely, and use `xSnprintf` or `xAsprintf` instead, but that's optional. As this code runs during initialization only, not on repeat during runtime, performance impact should be limited.

linux/LibSensors.c
```diff
@@ -129,7 +141,189 @@ int LibSensors_reload(void) {
...
+            sensor->sampleCount = 1;
+            continue;
+         }
+
+         if (value < sensor->min)
+            sensor->min = value;
+         if (value > sensor->max)
+            sensor->max = value;
+         sensor->sampleCount++;
+         sensor->average += (value - sensor->average) / (double)sensor->sampleCount;
```
BenBE, line 315:
```suggestion
         if (value < sensor->min)
            sensor->min = value;
         if (value > sensor->max)
            sensor->max = value;

         sensor->sampleCount++;
         sensor->average += (value - sensor->average) / (double)sensor->sampleCount;
```

Some minimal visual space for the code to breathe … ;-)

linux/Platform.c
```diff
@@ -245,6 +247,9 @@ const MeterClass* const Platform_meterTypes[] = {
    &SecondsUptimeMeter_class,
    &BatteryMeter_class,
    &HostnameMeter_class,
+#ifdef HAVE_SENSORS_SENSORS_H
+   &DynamicMeter_class,
+#endif
```
BenBE: Putting this after the Blank Meter should suffice.

BenBE (changes requested):

BenBE (approved):

PR #2077, opened 2026-08-19
Vector: use size_t for count comparison

massimomazzariol:
`Hashtable_count()` returns `size_t`, while `Vector_countEquals()` accepts an `unsigned int`. In debug builds, calls from `Table.c` therefore trigger `-Wshorten-64-to-32` with Clang.

Use `size_t` for both the `expectedCount` parameter and the internal counter in `Vector_countEquals()`.

Follow-up to #1673, which tracked `-Wshorten-64-to-32` warnings and related integer type mismatches

Tested with a debug build using Clang and:

    -Wshorten-64-to-32 -Werror

The full build completes without warnings

### DIFF
--- a/Vector.c
+++ b/Vector.c
@@ -54,8 +54,8 @@ static bool Vector_isConsistent(const Vector* this) {
    return true;
 }
 
-bool Vector_countEquals(const Vector* this, unsigned int expectedCount) {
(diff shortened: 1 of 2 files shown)


natoscott (commented):
LGTM, thanks @massimomazzariol

PR #2079, opened 2026-08-19
CRT: fix invisible gray color on direct-color terminals

### LINKED ISSUE #2078: Gray elements render as RGB(0,0,8) on direct-color terminals
On a direct-color terminal (`TERM=xterm-direct` and friends), every element htop draws in gray comes out as RGB(0, 0, 8), which is indistinguishable from black. On the default black background those elements are invisible.

### Cause

`CRT.c:1359` picks the gray foreground by colour count:

```c
short int grayBlackFg = COLORS > 8 ? 8 : 0;
short int grayBlackBg = (colorScheme != COLORSCHEME_BLACKNIGHT) ? -1 : 0;
init_pair(ColorIndexGrayBlack, grayBlackFg, grayBlackBg);
```

`COLORS > 8` is a reasonable test for "this terminal has the aixterm bright colours 8 to 15", and it holds on 256-colour terminals. On a direct-color terminal it is true as well (`COLORS` is 16777216), but there indices at or above 8 are no longer palette entries. ncurses' `xterm-direct` entry defines:

```
setaf=\E[%?%p1%{8}%<%t3%p1%d%e38:2::%p1%{65536}%/%d:%p1%{256}%/%{255}%&%d:%p1%{255}%&%d%;m
```

so an index below 8 takes the legacy `3<n>` branch, and anything from 8 up is decomposed into an RGB triple. Index 8 becomes `38:2::0:0:8`, that is red 0, green 0, blue 8.

### Reproduction

Same binary at `6f33ddd`, same `htoprc` (`color_scheme=0`, `show_cached_memory=0`, Memory meter in text mode so the cache classes use `METER_SHADOW`), only `TERM` differs:

```
TERM=xterm-256color   COLORS=256        158x  ESC[90m              <- real gray
TERM=xterm-direct     COLORS=16777216   175x  ESC[38:2::0:0:8m     <- RGB(0,0,8)
```

### Scope

Counting elements bound to `ColorPairGrayBlack` per scheme:

| scheme | elements affected |
|---|---|
| Default | 9 |
| Light Terminal | 8 |
| Midnight | 2 |
| Black Night | 5 |
| Nord | 18 |
| Monochrome, Black on White, Broken Gray | none |

In the default scheme that is `METER_SHADOW`, `PROCESS_SHADOW`, `BAR_SHADOW`, `SWAP_FRONTSWAP`, `MEMORY_3`, `HELP_SHADOW`, `CPU_IOWAIT`, `DYNAMIC_GRAY` and `DYNAMIC_DARKGRAY`.

### There is a workaround, and I got this wrong earlier

In [#2072](https://github.com/htop-dev/htop/pull/2072) I said a user could not dodge this by picking another scheme. That was wrong, and measuring the schemes individually is what showed it. On `TERM=xterm-direct`:

```
scheme 0 Default       139x RGB(0,0,8)
scheme 7 Nord          163x RGB(0,0,8)
scheme 6 Broken Gray     0x
scheme 1 Monochrome      0x
```

Broken Gray is immune for a reason that is almost on the nose: `CRT.c:1197-1198` builds it by walking the default scheme and replacing exactly `A_BOLD | ColorPairGrayBlack` with `ColorPair(White, Black)`. It already exists as the escape hatch for terminals whose gray does not work, so it happens to cover this case too. Users hitting this today can select it.

### On fixing it

Worth noting that the obvious fix needs the extended colour API rather than `init_pair`: a 24-bit value such as 0x808080 does not fit the `short` that `init_pair` takes, so a real gray on these terminals wants `init_extended_color` / `init_extended_pair`. That is the same machinery [#2072](https://github.com/htop-dev/htop/pull/2072) introduces, though the two are otherwise independent, and this one affects the schemes that already exist rather than adding a new one.

Filed separately at @rahawii's suggestion in that thread. Measured on Linux, ncurses 6.4, with a locally compiled `xterm-direct` terminfo entry carrying `RGB` and `colors#0x1000000`, since this box has no `ncurses-term` package.

rahawii:
This PR fixes the issue where gray elements render as `RGB(0,0,8)` (effectively black) on direct-color terminals, causing them to disappear against a black background.

## Changes

* Replaced the legacy `init_pair` fallback for `ColorIndexGrayBlack` with `init_extended_pair` when `COLORS >= 16777216`. This makes the fix completely independent.
* Passed `0x949494` (a highly visible 24-bit light gray) instead of the index `8`.
* Wrapped the extended API calls in preprocessor directives to ensure compatibility with older `ncurses` builds.

## Visual Changes

* **Before:** Gray elements (like bar shadows, memory cache lines, and process threads in standard schemes) were rendered as `RGB(0,0,8)`, making them virtually invisible.
* **After:** Elements now render correctly in a clear light gray (`0x949494`), restoring high contrast and readability for standard schemes on `-direct` terminals.

gray_fix

Fixes #2078

### DIFF
--- a/CRT.c
+++ b/CRT.c
@@ -1356,9 +1356,18 @@ void CRT_setColors(int colorScheme) {
       }
    }
 
-   short int grayBlackFg = COLORS > 8 ? 8 : 0;
-   short int grayBlackBg = (colorScheme != COLORSCHEME_BLACKNIGHT) ? -1 : 0;
-   init_pair(ColorIndexGrayBlack, grayBlackFg, grayBlackBg);
+#if defined(HAVE_INIT_EXTENDED_PAIR)
+   if (COLORS >= 16777216) {
+      int grayBlackBg = (colorScheme != COLORSCHEME_BLACKNIGHT) ? -1 : 0;
+      int grayBlackFg = (colorScheme == COLORSCHEME_LIGHTTERMINAL) ? 0x5F5F5F : 0x949494;
+      init_extended_pair(ColorIndexGrayBlack, grayBlackFg, grayBlackBg);
+   } else
+#endif
+   {
+      short int grayBlackFg = COLORS > 8 ? 8 : 0;
+      short int grayBlackBg = (colorScheme != COLORSCHEME_BLACKNIGHT) ? -1 : 0;
+      init_pair(ColorIndexGrayBlack, grayBlackFg, grayBlackBg);
+   }
 
    init_pair(ColorIndexWhiteDefault, White, -1);
 
--- a/configure.ac
+++ b/configure.ac
@@ -884,6 +884,7 @@ if test "$my_htop_platform" = "solaris"; then
 fi
 AC_CHECK_FUNCS([set_escdelay])
 AC_CHECK_FUNCS([getmouse])
+AC_CHECK_FUNCS([init_extended_pair])
 AC_DEFINE([NCURSES_ENABLE_STDBOOL_H], [1], [Define to enable stdbool.h in ncurses])
 
 


ravi-arnan:
Tested this on the same harness I offered in #2078: a locally compiled `xterm-direct` entry carrying `RGB` and `colors#0x1000000`, ncurses 6.4, Linux, counting the emitted SGR sequences rather than reading the screen. It does what it says, and I could not break it. Comment rather than approval, I am not a maintainer.

## The fix works, on every affected scheme

Same htoprc, same 40x160 screen, only `color_scheme` and `TERM` differ. Counting `ESC[38:2::0:0:8m` (the broken gray) against `ESC[38:2::148:148:148m` (the new one):

| scheme | `6f33ddd5` broken | `0c649cf8` broken | `0c649cf8` new gray |
|---|---|---|---|
| Default | 78 | 0 | 71 |
| Light Terminal | 83 | 0 | 79 |
| Black Night | 24 | 0 | 10 |
| Nord | 86 | 0 | 82 |
| Monochrome, Black on White, Midnight, Broken Gray | 0 | 0 | 0 |

Totals move run to run because the process list does, so read the zeros and not the differences. The 256-colour path is untouched: `xterm-256color` still emits `ESC[90m` and zero truecolor sequences on both builds.

## Four things I checked because they were the ways this could go wrong

**The bold attribute survives.** `ColorPairGrayBlack` is used as `A_BOLD | ColorPairGrayBlack` in these schemes, and an extended pair could have dropped it. It does not: every gray run is still preceded by `ESC[0;1m`, byte for byte the same as main on `xterm-256color`.

```
main   256color   <ESC>(B<ESC>[0;1m<ESC>[90m
main   direct     <ESC>(B<ESC>[0;1m<ESC>[38:2::0:0:8m
2079   direct     <ESC>(B<ESC>[0;1m<ESC>[38:2::148:148:148m
```

**The default background survives `-1`.** `init_extended_pair` taking -1 for the default colour is the other thing that could have quietly regressed, into an opaque black background. No `48:2:` sequence is emitted anywhere in the patched captures, on any scheme, so the background is still the terminal's.

**The scope is complete, not partial.** Index 8 really was the only palette value htop ever asks for above 7: the pair loop at `CRT.c:1350` runs `i < 8` and `j < 8`, and `ColorIndexWhiteDefault` uses `White`. Direct-colour `setaf` routes anything below 8 through its legacy `%{8}%<` branch, so those are unaffected by construction. The measurement agrees: the patched Default capture contains exactly one distinct truecolor triple and it is yours.

**All four direct variants, not just the one I built for.** Using ncurses 6.4's own terminfo database:

| TERM | new gray | broken |
|---|---|---|
| `xterm-direct` | 71 | 0 |
| `xterm-direct2` | 105 | 0 |
| `xterm-direct16` | 87 | 0 |
| `xterm-direct256` | 82 | 0 |

Worth knowing if you reproduce this: `xterm-direct2` emits `38:2:148:148:148` with two colons where `xterm-direct` emits `38:2::148:148:148` with three, an empty colourspace field. A grep written for one silently reports zero on the other, which is exactly what my first pass did.

## Two things to raise, neither of them blocking

**`COLORS >= 16777216` is fine, and I checked rather than assumed.** I scanned all 2817 entries in ncurses 6.4's terminfo database for a disagreement between `RGB` being set and `colors` reaching `0x1000000`, in both directions. There are none, so the two predicates are interchangeable in practice.

The coordination point is #2072, not correctness. That PR adds `CRT_hasDirectColors = (tigetflag("RGB") == 1)` guarded on `NCURSES_EXT_FUNCS` and `NCURSES_EXT_COLORS`, while this one adds an independent `COLORS >= 16777216` test plus a new `AC_CHECK_FUNCS([init_extended_pair])`. Whichever of the two lands second probably wants to collapse to one predicate and one capability check. On the capability half I would lean toward the macro test #2072 uses: on this ncurses the declaration of `init_extended_pair` sits inside the `curses.h` block that defines `NCURSES_EXT_FUNCS` (line 925 here), so that macro tracks whether the prototype exists, whereas `AC_CHECK_FUNCS` link-tests the symbol using its own decoy prototype. They agree here, and `HAVE_INIT_EXTENDED_PAIR` is detected correctly with a clean build under htop's full warning set.

**`0x949494` is theme-blind where `SGR 90` was not.** This is the one I would want an opinion on. Before the patch the gray was the terminal's own bright black, which the user themes; after it, on a direct terminal it is a fixed shade. On a dark background that is a clear improvement. Light Terminal is the case that gives me pause, since it exists for light backgrounds:

| foreground | on white | on black |
|---|---|---|
| `0x949494` | 3.03:1 | 6.92:1 |
| `0x808080`, xterm's usual index 8 | 3.95:1 | 5.32:1 |

3.03:1 is below the 4.5:1 WCAG AA threshold for text. I want to be clear that this is arithmetic and not observation: I have no light-background direct-colour terminal here and did not see it rendered. If the value ends up scheme-dependent, or simply darker for Light Terminal and Black on White, the measurement above is the reason. If it is deliberate, a one-line comment naming the value would stop the next reader asking.

Everything else looks right to me, and thank you for turning this around so quickly after #2078.

ravi-arnan:
Re-tested `8cf4e505` on the same harness. The contrast point is addressed, and I measured it rather than taking the constant on trust.

## Per scheme, `TERM=xterm-direct`

| scheme | `0x5F5F5F` | `0x949494` | `RGB(0,0,8)` | other triples |
|---|---|---|---|---|
| Default | 0 | 89 | 0 | none |
| Light Terminal | 96 | 0 | 0 | none |
| Midnight | 0 | 0 | 0 | none |
| Black Night | 0 | 16 | 0 | none |
| Nord | 0 | 85 | 0 | none |
| Monochrome, Black on White, Broken Gray | 0 | 0 | 0 | none |

So exactly one scheme moved, and it is the one that needed to. `xterm-256color` is still untouched on this head: `ESC[90m` and zero truecolor sequences on Default, Light Terminal and Nord. Clean build under htop's full warning set.

The counts above are what reaches the screen with my meter and column layout, not the number of elements in each scheme, which is why Midnight shows 0 despite having 2. Reading the scheme tables gives 9 / 0 / 0 / 8 / 2 / 5 / 0 / 18 references to `ColorPairGrayBlack`, matching the table in #2078 exactly.

## The contrast, now

| foreground | on white | on black |
|---|---|---|
| `0x5F5F5F`, light schemes | **6.39:1** | 3.29:1 |
| `0x949494`, everything else | 3.03:1 | **6.92:1** |

Light Terminal goes from 3.03:1 to 6.39:1 against white, which clears AA with room. That was the ask and it is answered.

Worth saying out loud, since it is the other half of the same trade: a user running Light Terminal on a *dark* terminal now gets 3.29:1 where this morning's head gave them 6.92:1. I do not think that is worth changing. The scheme name is the only statement about the background that htop has, `-1` leaves the actual background to the terminal, and someone picking Light Terminal on a dark terminal has already made a stranger choice than this. Recording the number so nobody discovers it as a surprise later.

## Two small things

**The `COLORSCHEME_BLACKONWHITE` arm is unreachable.** That scheme's table has zero references to `ColorPairGrayBlack`, and it emitted zero sequences of either shade at runtime. So the arm costs nothing and does nothing. Fine as future-proofing if that is the intent, in which case a short comment would stop the next reader testing it the way I just did; otherwise it can go and the condition reads shorter.

**Default on a light terminal is the case the heuristic still does not cover.** It keeps `0x949494` at 3.03:1 against white. I would leave it: the answer for that user is to pick Light Terminal, which is what the scheme is for. Noting it only so the boundary of the fix is written down.

Nothing here blocks anything from my side, and the answer on the maintainer's actual question, whether to converge with #2072 on one direct-colour predicate and one capability check, has not changed.

ravi-arnan:
@rahawii checking the current head `a471831889`, the force-push at 13:22 landed after my last re-test, so I re-read the delta against what I measured.

The only change from `8cf4e505` is one line at `CRT.c:1362`: `COLORSCHEME_BLACKONWHITE` no longer gets `0x5F5F5F`, it now falls through to `0x949494` with the other light schemes. That narrows the special case rather than widening it, so the per-scheme SGR counts I posted still describe this head. I did not re-run the harness on `a471831889` itself; the code delta is small enough to review by reading, and I would rather say that than imply a fresh measurement.

One thing worth your eyes before this sits longer: the PR carries the `invalid :wastebasket:` label. It was applied by `coderabbitai[bot]` one minute after you opened the PR, not by a maintainer, and I see no human comment behind it. If that label was not yours to add, it is likely a false positive from the bot's labeling, and it may be worth asking a maintainer to clear it, because an `invalid` label on an open PR can stall review or read as a rejection to a passer-by.

fasterit:
@rahawii: Do you want to pursue this one or #2072?

rahawii:
I’ve finished working on both. I don’t have any further improvements for either one at this point. Everything I have is pushed, and I’ve fixed all the issues pointed out by the bot and @ravi-arnan.

fasterit:
@rahawii Thank you, but do you think we should merge one OR the other?

We can either fix the default theme or support -direct explicitly. Both seems unnecessary, not?

ravi-arnan:
They answer different problems, so I would not frame them as either/or, but if only one is merged it should be this one.

What I measured on `TERM=xterm-direct` (ncurses 6.4, counting emitted SGR sequences): the `ColorIndexGrayBlack` pair is the only thing htop emits that breaks on a direct terminal. It resolves to `RGB(0,0,8)` instead of gray, on roughly ninety screen runs per full screen at my meter and column layout (89/96/85 on Default/Light Terminal/Nord; the count moves with the process list). Everything else passes through the legacy sub-8 branch of `setaf`, which the direct terminfo entries leave alone, so it is byte-identical to before. #2079 fixes exactly that pair: 12 lines plus one configure check, guarded by `HAVE_INIT_EXTENDED_PAIR`, and it repairs every existing scheme for anyone who starts htop with a `-direct` TERM, with no config change.

#2072 does not repair that pair, so a user on a direct terminal still gets the broken gray until they opt in and pick the new "Direct" scheme. What it adds is a new scheme with its own palette, falling back to Default when the `RGB` capability is absent. That is a taste question rather than a correctness one.

So: keep this one regardless, and weigh #2072 separately on whether the project wants a direct-color scheme at all. If both do land, the two predicates (`COLORS >= 16777216` here, `tigetflag(\"RGB\")` there) and the two capability checks want collapsing into one, as flagged earlier, but that is cleanup, not a conflict.

PR #2085, opened 2026-08-23
Implement a click handler for Meters and make the thr/kthr counters in TasksMeter toggle thread visibility

### LINKED ISSUE #2086: Allow to toggle thr/kthr display by clicking in the Tasks Meter
Based on a comment by @hishamhm in #2083:
> Here's an unsolicited wild idea that occurred to me the other day: what if clicking the thr and kthr counters of the Meter toggled their display on and off? I don't think anything in Meters is clickable yet (at least they weren't in my days), but seeing the thread gray out made this thought spring to my mind the other day.

_Originally posted by @hishamhm in https://github.com/htop-dev/htop/issues/2083#issuecomment-5381531509_

---

Note: For this to work, we'd need some way to tell where clicks in the header are going to, locate which Meter they affect and let that Meter decide what was hit. AFAICT the code for this doesn't exist yet, but shouldn't be too hard to implement.

fasterit:
Idea from Hisham in https://github.com/htop-dev/htop/pull/2083#issuecomment-5381531509

### DIFF
--- a/Header.c
+++ b/Header.c
@@ -180,6 +180,17 @@ Meter* Header_addMeterByClass(Header* this, const MeterClass* type, unsigned int
    return meter;
 }
 
+/* Meters in text mode may span into empty neighboring columns; keep the
+ * drawn and clickable widths consistent. */
+static int Header_meterWidth(const Header* this, const HeaderLayoutDimensions* colDims, const Meter* meter, size_t col, const int width) {
+   if (meter->mode == TEXT_METERMODE && !Meter_isMultiColumn(meter) && meter->columnWidthCount > 1) {
+      size_t spanCol = col + meter->columnWidthCount - 1;
+      return HeaderLayout_getColumnDimensions(this->headerLayout, this->pad, width, spanCol).x2 - colDims->x1;
+   }
+
+   return colDims->x2 - colDims->x1;
+}
+
 void Header_reinit(Header* this) {
    Header_forEachColumn(this, col) {
       for (int i = 0; i < Vector_size(this->columns[col]); i++) {
@@ -207,15 +218,7 @@ void Header_draw(const Header* this) {
 
       for (int y = (pad / 2), i = 0; i < Vector_size(meters); i++) {
          Meter* meter = (Meter*) Vector_get(meters, i);
-
-         int drawWidth = colDims.x2 - colDims.x1;
-
-         /* Let meters in text mode expand to the right on empty neighbors;
-            except for multi column meters. */
-         if (meter->mode == TEXT_METERMODE && !Meter_isMultiColumn(meter) && meter->columnWidthCount > 1) {
-            size_t spanCol = col + meter->columnWidthCount - 1;
-            drawWidth = HeaderLayout_getColumnDimensions(this->headerLayout, pad, width, spanCol).x2 - colDims.x1;
-         }
+         int drawWidth = Header_meterWidth(this, &colDims, meter, col, width);
 
          assert(meter->draw);
          meter->draw(meter, colDims.x1, y, drawWidth);
@@ -302,9 +305,12 @@ int Header_click(const Header* this, int x, int y) {
    Header_forEachColumn(this, col) {
       HeaderLayoutDimensions colDims = HeaderLayout_getColumnDimensions(this->headerLayout, this->pad, width, col);
 
-      if (x < colDims.x1 || x >= colDims.x2) {
+      if (x < colDims.x1) {
          continue;
       }
+
+      const bool columnOwnsX = (x < colDims.x2);
+      bool hit = false;
 
       Vector* meters = this->columns[col];
       for (int meterY = (this->pad / 2), i = 0; i < Vector_size(meters); i++) {
@@ -315,15 +321,24 @@ int Header_click(const Header* this, int x, int y) {
             continue;
          }
 
+         const int meterHitWidth = Header_meterWidth(this, &colDims, meter, col, width);
+         if (x >= colDims.x1 + meterHitWidth) {
+            /* beyond this meter's drawn extent -> try the next column */
+            break;
+         }
+
          Meter_Click clickFn = Meter_clickFn(meter);
          if (clickFn) {
             return clickFn(meter, x - colDims.x1, y - meterY);
          }
 
+         /* meter at this position has no click handler: stop the search */
+         hit = true;
          break;
       }
 
-      break;
+      if (hit || columnOwnsX)
+         break;
    }
 
    return HTOP_OK;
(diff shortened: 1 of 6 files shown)


Meter.h
```diff
@@ -100,6 +102,7 @@ typedef struct MeterClass_ {
 #define Meter_attributes(this_)        As_Meter(this_)->attributes
 #define Meter_name(this_)              As_Meter(this_)->name
 #define Meter_uiName(this_)            As_Meter(this_)->uiName
+#define Meter_clickFn(this_)           As_Meter(this_)->click
```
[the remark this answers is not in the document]
BenBE: These Meter_* macros defined on As_Meter(this_) are intentional architecture and provide an easy way to access the VTable of Meter classes.

Meter.h
```diff
@@ -55,6 +55,7 @@ typedef ATTR_NONNULL void (*Meter_UpdateValues)(Meter*);
 typedef ATTR_NONNULL void (*Meter_Draw)(Meter*, int, int, int);
 typedef ATTR_NONNULL const char* (*Meter_GetCaption)(const Meter*);
 typedef ATTR_NONNULL ATTR_ACCESS3_W(2, 3) void (*Meter_GetUiName)(const Meter*, char*, size_t);
+typedef int (*Meter_Click)(Meter*, int relX, int relY);
```
BenBE: We need to ensure a Meter is given; thus the ATTR marker.
```suggestion
typedef ATTR_NONNULL int (*Meter_Click)(Meter*, int relX, int relY);
```

ScreenManager.c
```diff
@@ -294,6 +294,23 @@ void ScreenManager_run(ScreenManager* this, Panel** lastFocus, int* lastKey, con
                      panelFocus->lastMouseBarClickX = mevent.x;
                      ch = KEY_MOUSE_BAR_CLICK;
                   }
+               } else if (!this->state->hideMeters && this->header &&
+                          mevent.y >= this->y1 && mevent.y < this->y1 + header_height(this)) {
```
BenBE, line 299: Should include `x1`/`x2` bounds check to account for "Leave margin around header" setting.

ScreenManager.c
```diff
@@ -294,6 +294,23 @@ void ScreenManager_run(ScreenManager* this, Panel** lastFocus, int* lastKey, con
...
+                  /* Click in the meters/header area */
+                  int meterReaction = Header_click(this->header, mevent.x, mevent.y - this->y1);
```
BenBE:
```suggestion
                  int meterReaction = Header_click(this->header, mevent.x - this->x1, mevent.y - this->y1);
```

Header.c
```diff
@@ -307,3 +305,41 @@ int Header_calculateHeight(Header* this) {
 
    return maxHeight;
 }
+
+int Header_click(const Header* this, int x, int y) {
+   const size_t numCols = HeaderLayout_getColumns(this->headerLayout);
+   const int width = COLS - 2 * this->pad - ((int)numCols - 1);
+   int colX = this->pad;
```
BenBE: Given that the calculation for the hitbox for the meters need to stay in sync with the draw rects we should have this calculation factored out into `HeaderLayout_getColumnDimension` or alike returning the rect for this header column and have all relevant places (Meter drawing, click handling) reuse these calculations. This also avoids using the local variable `roundingLoss` and speeds up the process by avoiding some FP math.

Header.c
```diff
@@ -307,3 +305,41 @@ int Header_calculateHeight(Header* this) {
...
+   Header_forEachColumn(this, col) {
+      float colWidth = (float)width * HeaderLayout_layouts[this->headerLayout].widths[col] / 100.0F;
+
+      roundingLoss += colWidth - floorf(colWidth);
+      if (roundingLoss >= 1.0F)
+         colWidth += 1.0F;
+
+      int colEnd = colX + (int)floorf(colWidth);
+
+      if (x >= colX && x < colEnd) {
```
BenBE: Invert this condition and use `continue` instead to reduce overall code indentation.

BenBE (changes requested):
OVerall handling LGTM; minor function split to help keep some location calculations in sync.

Unrelated note: The `, ` text should probably be printed in teal always, not just when the element after them (`thr`/`kthr`) are active. ATM they are grey when inactive.

hishamhm:
@fasterit I built the branch and yes, it's working! Nice!

Clicking threads on and off like this made me realize that I never kept the selection stable across these toggles (that's most noticeable in tree view). If I'm looking at a process and then enable threads, the list jumps around and I have to hunt for the process again. That was already the case when toggling via keyboard; this just made me notice it.

Implementation-wise, I understand it's a bit trickier to keep the selection stable when disabling rather than enabling: if the selection is in a process, it's easy to keep it in the same PID and have threads show up beneath it. If the current selection is a thread and one toggles threads off, then of course the selected PID disappears, so I guess the logical thing for the selection would be for it to go up to the parent process. But yeah, that would no longer be just flicking a boolean.

But then of course that's a general behavior pattern: htop doesn't preserve the selected PID when going in and out of tree view, etc. But I don't know if it's worth it: preserving the selection across toggles (without having to enable "follow") doesn't make a lot of sense unless one is in "tree view sorted by PID" (which tends to be the most stable view in terms of screen movement).

But that's just an aside, just thinking out loud about the UI behavior! The clickable Meter is already a nice improvement! Thank you!!

fasterit:
@hishamhm: you have stable tree mode (check Setup -> Display options -> Tree view is kept visually stable) and we have much improved the Follow function (<kbd>F</kbd>) and there is "Sticky follow" in the Display options too which allows you to get out of Follow mode with the <kbd>UP</kbd>/<kbd>DOWN</kbd>cursor keys. Or not. That may help a bit already.

I don't think people toggle the thread views that often. But if you (or somebody else, please chime in) think differently ... we _can_ make the view "stable" on the last PID selected.

Header.c
```diff
@@ -307,3 +294,37 @@ int Header_calculateHeight(Header* this) {
...
+
+         Meter_Click clickFn = Meter_clickFn(meter);
+         if (clickFn) {
+            return clickFn(meter, x - colDims.x1, y - meterY);
+         }
+
+         break;
+      }
+
+      break;
```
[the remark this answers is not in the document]
fasterit: Addressed in [fe87829](https://github.com/htop-dev/htop/pull/2085/commits/fe87829bc74af83ebe32cd844c3078966aac6087)

fasterit:
Stable PID re-selection has been implemented on top of this PR in https://github.com/htop-dev/htop/pull/2090 .
@hishamhm and @BenBE both wanted it. :tada:

PR #2091, opened 2026-08-28
darwin: Fetch CPU frequency via sysctl (hw.cpufrequency)

badger200:
Fixed setting Show CPU Frequency checkbox so it works now instead of displaying <N/A>.
Assisted-by: Antigravity <ai@google.com>


darwin/Platform.c
```diff
@@ -357,7 +365,18 @@ double Platform_setCPUValues(Meter* mtr, unsigned int cpu) {
...
    total = mtr->values[CPU_METER_NICE] + mtr->values[CPU_METER_NORMAL] + mtr->values[CPU_METER_KERNEL];
 
-   mtr->values[CPU_METER_FREQUENCY] = NAN;
+   double frequency = NAN;
+   if (mtr->host->settings->showCPUFrequency) {
+      uint64_t cpufreq = 0;
+      size_t len = sizeof(cpufreq);
+      if (sysctlbyname("hw.cpufrequency", &cpufreq, &len, NULL, 0) == 0 && cpufreq > 0) {
+         frequency = (double)cpufreq / 1E6;
+      } else if (sysctlbyname("hw.cpufrequency_max", &cpufreq, &len, NULL, 0) == 0 && cpufreq > 0) {
```
BenBE, line 374: is `len` guaranteed to be unchanged when the first of these two `sysctl`s fails?

darwin/Platform.c
```diff
@@ -357,7 +365,18 @@ double Platform_setCPUValues(Meter* mtr, unsigned int cpu) {
    /* Convert to percent and return */
    total = mtr->values[CPU_METER_NICE] + mtr->values[CPU_METER_NORMAL] + mtr->values[CPU_METER_KERNEL];
 
-   mtr->values[CPU_METER_FREQUENCY] = NAN;
+   double frequency = NAN;
+   if (mtr->host->settings->showCPUFrequency) {
+      uint64_t cpufreq = 0;
+      size_t len = sizeof(cpufreq);
+      if (sysctlbyname("hw.cpufrequency", &cpufreq, &len, NULL, 0) == 0 && cpufreq > 0) {
```
BenBE, line 372: Is this the per-core frequency? Because I don't see, where it follow's up on the `cpu` argument provided to this function.

PR #2093, opened 2026-08-30
Action/InfoScreen: replace octal character literals with key constants

GermanAizek:
Replace raw octal character escapes '\014' and '\177' with standard KEY_CTRL('L') and KEY_DEL_MAC / KEY_BACKSPACE macros to improve code clarity and avoid static analyzer warnings


Action.c
```diff
@@ -1006,8 +1006,9 @@ void Action_setBindings(Htop_Action* keys) {
 #endif
    keys['Z'] = actionTogglePauseUpdate;
    keys['['] = actionLowerPriority;
-   keys['\014'] = actionRedraw; // Ctrl+L
-   keys['\177'] = actionCollapseIntoParent;
+   keys[KEY_CTRL('L')] = actionRedraw;
+   keys[KEY_DEL_MAC] = actionCollapseIntoParent;
+   keys[KEY_BACKSPACE] = actionCollapseIntoParent;
```
fasterit, line 1011: AI artifact?
you replace two lines and invent one

fasterit (changes requested):

Action.c
GermanAizek: > AI artifact? you replace two lines and invent one

No, this for compability with MacOS.

An additional line was added to ensure compatibility with different terminals and operating systems (cross-platform support).

Different terminal emulators handle the Backspace key press in different ways:

KEY_DEL_MAC (formerly \177): On macOS and some older UNIX terminals, pressing the Backspace key sends the ASCII code 127 (the DEL character).
Previously, the code hard-coded this specific value (the magic number \177), which meant the action might not work on other systems.

KEY_BACKSPACE: In most modern Linux/Windows terminals, the Backspace key sends the standard Backspace code (ASCII 8 or a special code from the ncurses library).

fasterit: KEY_BACKSPACE was not there before

fasterit: 0dbedf95 is the introducing commit. @hishamhm ... did you mean to use <backspace> for collapsing a tree segment outside of Mac keyboards as well. Like 8 years ago and nobody noticed? :smile:

hishamhm: I definitely did not take Mac keyboards into account back then... but I do remember that different Linux terminals (xterm, rxvt, konsole, the Linux textmode console, etc.) had different default settings wrt handling Backspace, and \177 was the code for Backspace in the terminal I was using back then (which might have been any of the above :) ). I think there is even an stty setting or something to configure what Backspace emits, isn't there?

fasterit: yes there is, `stty erase <char>` to set it
thank you for the info @hishamhm

fasterit:
So <kbd>backspace</kbd> (on ancient Linux distros and Mac keyboard) can collapse the tree segment. We have <kbd>-</kbd> (Minus) and <kbd>*</kbd> (star, toggle all children of PIDs without parents) documented in `man htop` and the help screen but not the <kbd>backspace</kbd>. Shall we drop that binding? Or make it more available as per @GermanAizek's current commit?

Action.c
Explorer09: It's better to make a separate commit for this change for clarity.

BenBE:
No hard feelings to drop the backspace binding …

fasterit:
@GermanAizek: Can you make this two commits?

The first one drops
`keys['\177'] = actionCollapseIntoParent;`  as this keybinding is outdated and should not be supported any more.

And then a second one for your remaining octal -> macro beautifications.

Thanky!

GermanAizek:
> The first one drops `keys['\177'] = actionCollapseIntoParent;` as this keybinding is outdated and should not be supported any more.
> 
> And then a second one for your remaining octal -> macro beautifications.


Fix it

https://github.com/htop-dev/htop/pull/2093/commits/2f98336f4cbccee819acd1df2f5d098c4525c90f

https://github.com/htop-dev/htop/pull/2093/commits/332ec2403f5555e2e3f459b313c9c6878095711b

fasterit:
@GermanAizek Thank you!

PR #2094, opened 2026-08-30
OpenFilesScreen: use MINIMUM instead of CLAMP for unsigned column length

GermanAizek:
The CLAMP(x, low, high) macro expands to a range check that calls MAXIMUM(x, low), which results in the evaluation of the subexpression dlen > 0

dlen is of type size_t (unsigned) and dlen is known to be >= 0, the expression is inside the if (dlen > (size_t)pdata->cols[index]) block, where pdata->cols[index] >= 0 (therefore, dlen > 0 is guaranteed to be true), the lower bound (0) is redundant, and the analyzer correctly identifies the condition as constant


fasterit:
Does the compiler not optimize this out anyways?

GermanAizek:
> Does the compiler not optimize this out anyways?

If you'd like, I can disassemble this line code with -O3 optimization flag, and we can compare the changes.

BenBE:
Any progress regarding your comparison?

GermanAizek:
> Does the compiler not optimize this out anyways?
> Any progress regarding your comparison?

I conducted the testing, and you were right — it’s not worth merging the change just for the sake of -O0, as it improves code readability.

image

PR #2095, opened 2026-08-30
Process: remove redundant non-NULL checks

GermanAizek:
In Process_makeCommandStr(), the function returns early when procExe or procComm is NULL. The subsequent checks in conditional expressions are always true and trigger static analyzer warnings


Explorer09:
Why the heck can't we let the compiler optimize it out? It's safer to have redundant checks unless the code in question is performance critical.

BenBE:
The mentioned checks are there somewhat intentional as DiD, because the surrounding code has a certain degree of complexity that can't be avoided easily. Thus those checks will stay.

PR #2102, opened 2026-09-05
XUtils: ignore empty needles in a multi-term filter

rawsun007:
`String_contains_i()` splits the filter on `|` and hands each needle to `strcasestr()`, which reports a match against *anything* for an empty needle. `String_split()` emits an empty field for a leading or repeated separator but drops a trailing one, so the same typo behaved differently depending on where it landed:

| filter | before | after |
| --- | --- | --- |
| `nginx\|` | nothing extra | unchanged |
| `\|nginx` | **every row** | only `nginx` rows |
| `nginx\|\|postgres` | **every row** | `nginx` or `postgres` rows |

So while composing a multi-term filter, a stray or half-typed `|` silently stopped the filter from filtering — and the same mistake at the end of the line was harmless. That inconsistency is what convinced me it wasn't intentional: if an empty alternative were meant to match everything, a trailing `|` would do it too.

Empty needles are now skipped. A filter consisting only of separators carries no search term and still matches everything, the same as an empty filter — the incremental-search callers already treat a null filter that way.

Fixed in `String_contains_i()` rather than `String_split()`, since the empty-field behaviour there is relied on elsewhere and only the filter's semantics are wrong.

Verification, on darwin: linked a small driver against the tree's own `XUtils.o` and ran 14 cases covering a leading, trailing, doubled and interior separator on both the matching and non-matching side, plus case-insensitivity and separators alone. All pass; reverting the change fails exactly `|nginx` and `nginx||postgres`. Full `make` clean with no new warnings under the tree's flags (`-Wall -Wextra -pedantic` and the rest).

I could not add a regression test: the only harness here is `test_spec.lua`, which drives a terminal rather than calling into C, so there is nowhere for a unit test of this to live. Happy to write a `test_spec.lua` case that types a filter into a live htop if you would rather have coverage that way — it seemed worse than the C-level evidence above, but it is your call.

AI disclosure per `docs/ai-contributions-policy.md`: found and written with Claude Code (Claude Opus 5), recorded in the commit's `Assisted-by:` trailer. @rawsun007 authorised the `Signed-off-by:` and reviewed the change.


fasterit:
I'll merge this without the first AI comment. Today's AIs tend to work around their limited memory and context window by adding rather useless, verbose (to a human) commentary. So I'll cull that one. Otherwise thank you for the PR.

fasterit:
Merged as b9a2a6359967b918f427e6779ad78a1d83a761e2

PR #2103, opened 2026-09-10
Harden horizontal scroll offsets

fasterit:
Spliced out from #2099

This makes htop only scroll right if there is something to scroll right to.

### DIFF
--- a/Panel.c
+++ b/Panel.c
@@ -409,6 +409,7 @@ bool Panel_onKey(Panel* this, int key) {
       case KEY_CTRL('B'):
          if (this->scrollH > 0) {
             this->scrollH -= MAXIMUM(CRT_scrollHAmount, 0);
+            this->scrollH = MAXIMUM(this->scrollH, 0);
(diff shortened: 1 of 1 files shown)


BenBE:
Does this also limit the right scrolling to the longest line horizontally?

fasterit:
Yes, you have to be on a longer line to be able to scroll right.
Do we want that? Otherwise we could only merge the pinning of the lower bound (left).

BenBE:
The patch as is blocks scrolling at the right end of the "Command" header, thus making command lines effectively inaccessible ATM.

My previous question was about two possible behaviours:
a) scroll to the right infinitely (previous)
b) allow scrolling up to the end of the longest command line

both is fine; maybe a) is more consistent, as the longest command line could change, which might introduce sudden jumps.

fasterit:
It actually does b). You just have to be on a line that is extending right beyond the current view-panel.

I am not convinced, we want that, hence my Q.

PR #2017, opened 2026-05-30
Formalize security process

BenBE:
@cgzones @fasterit @ginggs @natoscott RFC.

PR #2104, opened 2026-09-11
Add hide_offline_cpus setting to hide offline CPUs in the CPU meters

### LINKED ISSUE #853: Offline CPUs
Hi. Would it be possible to add a configure flag (or even runtime option) to disable the showing of offline CPUs? It's taking up space in the terminal for no reason that I can see. Thanks.

Edznux:
Hello :)

## Context
I was working on a relatively large server where we disabled SMT and this caused the 64 core / 128 threads listing to have 64 `[offline]` CPU displayed. This was less than ideal when using a small terminal pane, especially to see the load on the other 64 cores, so I decided to take a look at it. I noticed #853 and decided to pick it up!

I don't know the codebase, and I'm not a particularly good C dev, but I've reviewed the code and steered my agent for a bit to achieve this. I do believe it went a bit too far with some changes, but I also think it made sense for the purpose of the PR.

I do believe I've followed the contribution guidance and the AI disclosure and all as well, but happy to change.

## Goal

On machines booted with nosmt (or with CPUs otherwise offlined), the multi-CPU meters draw a bar reading "offline" for every disabled thread, doubling the height of the header on large SMT systems for no benefit.

Add a "Hide offline CPUs in the CPUs meters" display option (htoprc key hide_offline_cpus, default off).
When enabled, the CPUs meters lay out the sequence of online CPUs instead of every existing CPU, so the first/second half and multi-column layouts split over the shown CPUs only. Sub-meters are now kept per CPU id and created when a CPU is first shown, so a CPU that is hidden or shifts to another slot keeps its graph history; the meter recomputes its own height whenever the shown sequence changes.

Header_updateData now recalculates the header height and reports whether it changed, so the refresh loop re-lays out the screen when CPUs go online or offline at runtime instead of leaving blank or clipped rows until the terminal is resized.

## Recorded run

Here's an example of how it looks on my 16C/32T machine, with a script that disables cores over time.

https://github.com/user-attachments/assets/76705922-d1c8-432c-8025-1b5b30dc6a93

Closes #853 

Assisted-by: Claude Fable 5.1 <noreply@anthropic.com>

### DIFF
--- a/CPUMeter.c
+++ b/CPUMeter.c
@@ -44,8 +44,11 @@ static const int CPUMeter_attributes_summary[] = {
 };
 
 typedef struct CPUMeterData_ {
-   unsigned int cpus;
-   Meter** meters;
+   unsigned int ncol;         /* columns the shown sub-meters are laid out in */
+   unsigned int cpus;         /* size of meters: one slot per existing CPU */
+   Meter** meters;            /* sub-meter per CPU id, created when first shown */
+   unsigned int shownCount;
+   Meter** shown;             /* the sub-meters currently shown, in display order */
 } CPUMeterData;
 
 static void CPUMeter_init(Meter* this) {
@@ -233,8 +236,23 @@ static void CPUMeter_display(const Object* cast, RichString* out) {
    #endif
 }
 
+/* Whether the multi-CPU meters show a slot for this CPU (0-based id) */
+static bool CPUMeter_isShown(const Machine* host, unsigned int cpu) {
+   return !host->settings->hideOfflineCPUs || Machine_isCPUonline(host, cpu);
+}
+
+static unsigned int CPUMeter_countShown(const Machine* host) {
+   unsigned int shown = 0;
+   for (unsigned int cpu = 0; cpu < host->existingCPUs; cpu++) {
+      if (CPUMeter_isShown(host, cpu))
+         shown++;
+   }
+   return shown;
+}
+
+/* Range of this meter within the sequence of shown CPUs */
 static void AllCPUsMeter_getRange(const Meter* this, unsigned int* start, unsigned int* count) {
-   unsigned int cpus = this->host->existingCPUs;
+   unsigned int cpus = CPUMeter_countShown(this->host);
    switch (Meter_name(this)[0]) {
       default:
       case 'A': // All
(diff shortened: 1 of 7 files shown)


ravi-arnan (commented):
Reviewed this closely since I did the adjacent CPU hotplug fix (#2064). Fetched the PR head (`f0ea593`) and built on Linux under htop's own warning set: exit 0, zero new warnings in any touched file. Also ran the built binary for a few seconds of live refreshes under a pty with asserts enabled: no crash, no assert trip. All 16 CPUs on my machine are online, so the hide path itself is compile-verified plus traced by hand, not runtime-tested; stating that boundary explicitly.

Things I verified that hold up:

- The per-CPU-id `meters[]` indexing subsumes the mid-range removal case from #2064: arbitrary online/offline gaps are handled by construction, and the Left/Right halves now partition the *shown* sequence, which is the sane semantic with hiding on.
- `Machine_isCPUonline()` exists on all 8 platform backends, so the new gate is portable with no extra ifdefs.
- `Header_updateData()` void-to-bool is source-compatible with the two other callers (`DisplayOptionsPanel.c:235`, `AvailableMetersPanel.c:86`); both use it as a bare statement.
- `AllCPUsMeter_updateMode()` calls `Meter_init()`, which is only a `->init` dispatch (Meter.h), so there is no updateMode/init recursion; mode propagation to the sub-meters through `commonInitSubMeters` is correct.
- The `xReallocArray` growth path for `shown` cannot expose uninitialized slots: when the array grows, `changed` is true and the compare short-circuits.
- Bonus this brings along: `Header_updateData` reporting height changes up to `ScreenManager_resize` means a live hotplug while htop runs now resizes the header instead of leaving a stale height.

One nit (non-blocking): the settings label says "(e.g. SMT siblings disabled by nosmt)", but offline CPUs equally come from hot-unplug. Suggest widening it, e.g. "(e.g. disabled by nosmt/nosmp or hot-unplugged)".

One question: hidden CPUs stop receiving `Meter_updateValues()`, so when a CPU comes back the first delta spans the whole hidden interval. For CPU% that reads as an average over the gap, which seems fine, but is that the intended semantic, or should a re-shown CPU reset its history instead?

No blocking issues found. Leaving approval to the maintainers.

Edznux:
@ravi-arnan you can test by temporarily disabling a core
For example, disabling core 12:
```
sudo sh -c 'echo 0 > /sys/devices/system/cpu/cpu12/online'
```

and set to 1 to re-enable it :)

ravi-arnan:
Follow-up on the testing boundary I flagged in my review: I took a real CPU offline
and ran the hide path, so it is runtime-verified rather than traced by hand.

**Method.** CPU 12 offline through the kernel (`echo 0 > /sys/devices/system/cpu/cpu12/online`,
back to 1 after every run; the box is back to 16/16). Rebuilt at `f0ea593` with
`--enable-debug`, so assertions are actually live this time. The binary ran in a pty
and I read the rendered screen after it settled.

**Results with cpu12 offline:**

- **hide=1, CPUs (1/1):** 15 rows, `0 1 ... 11 13 14 15`. cpu12 gone, the rest in place,
  and the header is exactly one row shorter (the process list moves up from line 19 to
  18). No assert trip, no crash.
- **hide=0:** 16 rows, cpu12 still there as `12[                        offline]`. So the
  setting only controls visibility; the offline marking is untouched.
- **hide=1, CPUs (1/2) + (2/2):** 8 + 7 rows. I put a meter between the two halves to make
  the boundary unambiguous: left = `0-7`, right = `8, 9, 10, 11, 13, 14, 15`. The halves
  partition the *shown* sequence, cpu12 is absent from both, and no slot is duplicated or
  dropped. This is the case I care about, since #2064 was about a mid-range offline CPU and
  this is exactly what the shown-index mapping has to get right.
- **hide=1, CPUs (1&2/2):** the 15 shown slots fill 8 rows in two columns
  (`0-7` | `8-11,13-15`), column-major, no hole where cpu12 would be.

**Live hotplug in one instance** (hide=1): started with cpu12 offline (15 rows), brought
it online (16 rows, cpu12 back, header one row taller), took it offline again (15 rows),
brought it back online (16). Four transitions in a single htop, no assert, no redraw
artifacts. That exercises `Header_updateData()` reporting the height change and the
`ScreenManager_resize()` behind it, which I could only check by reading the code last time.

**And the answer to my own question** (does a re-shown CPU average over the hidden gap):
no. The per-CPU periods are recomputed on every scan in `LinuxMachine_scanCPUData()`, not
in the meter, so hiding a meter does not stretch the next delta. There is a separate
pre-existing quirk for CPUs that are genuinely offline: `LinuxMachine.c:580-585` (upstream,
from #2017) zeroes `cpuData[i]` for any id missing from `/proc/stat`, so the first value
after a CPU returns is measured against a zero baseline, one frame of "average since
boot". Not introduced here and it happens with hide=0 too, where the row is simply
visible; in the run cpu12 came back at 0.0% and looked unremarkable. Meter graph history
does freeze while hidden, so a re-shown graph has a time discontinuity, which seems
acceptable.

The label nit from the review still stands if you want it (the setup text mentions nosmt
but not nosmp/hot-unplug). Nothing blocking from me; approval to the maintainers as
before.

fasterit:
```
CPUMeter.c:332:12: warning: Access to field 'h' results in a dereference of a null pointer [core.NullDereference]
  332 |    int h = data->shown[0]->h;
      |            ^~~~~~~~~~~~~~~~~
1 warning generated.
```

from the clang analyzer CI job

PR #2107, opened 2026-09-13
Very simple hwmon fan speed meter

foxpy:
This PR is a replacement for #2009, which implemented the same feature, but using a vendor-specific interface, only supporting Thinkpads running Linux. The new version should support all Linux devices, because it uses hwmon.

It only reports the speed of the first fan it finds, which is not great, but hey, it's something :)

@BenBE sorry for ignoring #2009 for the entire summer, I've been busy with personal stuff and completely forgot about this :D

I remember that we talked about adding support for minimum and maximum fan speed reporting, which I also decided to keep aside for now.


massimomazzariol:
Hi, just adding some context that might be useful here for this PR

While working on #2076 I also drafted a follow-up branch to make the hardware sensor code backend-independent and add direct Linux hwmon support as a fallback when libsensors is unavailable:

https://github.com/massimomazzariol/htop/tree/feature/hwmon-fallback

The draft currently discovers hwmon temperature, fan, PWM and voltage sensors and feeds them through the same generic `HardwareSensor` interface used by the meters. It also avoids relying on `hwmonN` numbering for persistent identity, since those numbers are not stable across boots.

The main hwmon backend commit is here:

https://github.com/massimomazzariol/htop/commit/fd395ce5fac9d5a0315b12d4e1a09c5707501273

I deliberately kept this out of #2076 because I did not want to expand the scope again after the sensor meter implementation had gone through review.

Since #2107 is now adding fan support through hwmon, I thought it was worth mentioning in case parts of that draft or the abstraction could be useful, or if the two approaches should be aligned before both implementations diverge.

If you want, feel free to adapt the draft depending on which direction you prefer

linux/Platform.c
```diff
@@ -441,6 +443,84 @@ void Platform_setGPUValues(Meter* this, double* totalUsage, unsigned long long*
    this->values[residueIndex] = residuePercentage;
 }
 
+static int Platform_getFanSpeedFromHwmonDevice(int parent_dfd, char const* hwmon_device) {
```
BenBE, line 446: Using `const char *` here is fine and overall more common in the codebase.
```suggestion
static int Platform_getFanSpeedFromHwmonDevice(int parent_dfd, const char *hwmon_device) {
```

linux/Platform.c
```diff
@@ -441,6 +443,84 @@ void Platform_setGPUValues(Meter* this, double* totalUsage, unsigned long long*
...
+   DIR* dirp;
+   int dfd;
+
+   dfd = openat(parent_dfd, hwmon_device, O_RDONLY | O_CLOEXEC | O_DIRECTORY);
```
BenBE, line 450: Cf. `Compat_openat`

linux/Platform.c
```diff
@@ -441,6 +443,84 @@ void Platform_setGPUValues(Meter* this, double* totalUsage, unsigned long long*
...
+         continue;
+      }
+
+      int fan_speed;
+      if (fscanf(fp, "%d", &fan_speed) != 1) {
+         fclose(fp);
+         continue;
+      }
+
+      fclose(fp);
```
BenBE, line 481: Cf. `Compat_readfileat`

linux/Platform.c
```diff
@@ -441,6 +443,84 @@ void Platform_setGPUValues(Meter* this, double* totalUsage, unsigned long long*
...
+
+   struct dirent* dir;
+   while ((dir = readdir(dirp)) != NULL) {
+      int fan_num;
+      if (sscanf(dir->d_name, "fan%d_input", &fan_num) != 1)
+         continue;
+
+      int fd = openat(dfd, dir->d_name, O_RDONLY | O_CLOEXEC);
+      if (fd == -1)
+         continue;
```
BenBE, line 468: Some blank line after the if statement doesn't hurt.
```suggestion
         continue;

```

linux/Platform.c
```diff
@@ -441,6 +443,84 @@ void Platform_setGPUValues(Meter* this, double* totalUsage, unsigned long long*
...
+   }
+
+   closedir(dirp);
+   return -1;
+}
+
+/*
+Traverse hwmon sensors, find first fan, read and return its speed.
+Returns -1 on failure.
+*/
```
BenBE, line 493:
```suggestion
/* Traverse hwmon sensors, find first fan, read and return its speed. */
```

Makefile.am
```diff
@@ -117,6 +117,7 @@ myhtopheaders = \
 	FileDescriptorMeter.h \
 	FunctionBar.h \
 	GPUMeter.h \
+	FanMeter.h \
```
BenBE, line 120: File order is sorted alphabetically

Makefile.am
```diff
@@ -214,6 +215,7 @@ linux_platform_headers = \
 
 linux_platform_sources = \
 	GPUMeter.c \
+	FanMeter.c \
```
BenBE, line 218: File order is sorted alphabetically

BenBE (changes requested):

BenBE:
@foxpy No problem. Sometimes PRs take a while to ripen. One of my earliest PRs to htop took more then 6 months and it mostly cleaned up some existing code to make it be shared by all platforms. Given the long-term maintainability is important with a huge code-base, having several approaches to a problem to compare what works best is a clear plus. This applies similarly to using synergies with existing code, thus re-using existing infrastructure to query information or realize a feature is much preferred.

@coderabbitai Removing braces around single-statement bodies is not required, but (strongly) encouraged. It's totally fine to have them, e.g. if the statements itself are somewhat complicated, but for most simple cases (i.e. just an assignment), the variant without braces is preferred/recommended. Furthermore: blank openat without `HAVE_OPENAT` guards MUST not appear; that's what the `Compat_openat` helpers are for.

@massimomazzariol Your PR #2076 is still on my list of PRs to review, but this might still take some time. As noted above: Having a shared infrastructure for accessing hwmon information without duplicating logic unnecessarily sounds good.

foxpy:
@BenBE thank you for your review. I am afraid @massimomazzariol 's work in PR #2076 is more useful and we both should focus on reviewing it instead of spending extra time on what I did. What do you think?

PR #2109, opened 2026-09-15
Support + / - (toggle tree expand and collapse) and * (same for all branches) on NumPad

fasterit:
follow up to #2093


CRT.h
```diff
@@ -179,20 +179,26 @@ void CRT_handleSIGSEGV(int signal) ATTR_NORETURN;
...
+# define KEY_PADSTAR   587
 #else
 # define KEY_PADPLUS   PADPLUS
 # define KEY_PADMINUS  PADMINUS
+# define KEY_PADSTAR   PADSTAR
 #endif
 
+/* Upper bound for MainPanel's key table; covers the keypad keys, which are
+   htop-internal values above ncurses' KEY_MAX (0777 oct = 511 dec). */
+#define KEY_MAX_HTOP  (KEY_MAX > KEY_PADMINUS ? KEY_MAX : KEY_PADMINUS + 1)
```
BenBE, line 200: Use `MAXIMUM` from `Macros.h`, which we include already anyway.
```suggestion
#define KEY_MAX_HTOP  MAXIMUM(KEY_MAX, KEY_PADMINUS + 1)
```

fasterit: I like the more verbose version better as the `KEY_MAX > KEY_PADMINUS` will remind me in a year wtf that exists. But I did add the comment, so if you get annoyed by it, I'll change to `MAXIMUM`.

PR #2110, opened 2026-09-18
XUtils: treat a carriage return as whitespace in String_trim

rawsun007:
`String_readLine` removes only the `\n`:

```c
char* newLine = strrchr(at, '\n');
if (newLine) {
   *newLine = '\0';
```

so a config file written with CRLF line endings hands every line to the caller with a trailing `\r`. `String_trim` skips spaces, tabs and newlines but not carriage returns, so the `\r` survives.

Driving the tree's own `XUtils.o` with a CRLF `htoprc`:

```
readLine               [column_meters_0=LeftCPUs2 Memory Swap\r]
  after String_trim    [column_meters_0=LeftCPUs2 Memory Swap\r]   before
  after String_trim    [column_meters_0=LeftCPUs2 Memory Swap]     after
```

Every caller of `String_trim` is parsing a config file — `Settings_splitLineToIDs` for `htoprc`, and `PCPDynamicMeter.c` / `PCPDynamicScreen.c` for their `=`-separated config — so the carriage return lands in the last field of each line. For `column_meters_0` that is the final meter name.

That `String_trim` already strips `\n`, which `String_readLine` has by then removed, reads as an intent to be tolerant of line-ending debris; `\r` looks like the case that was missed rather than one deliberately excluded.

The change adds `'\r'` to both loops. Case table, again against the tree's `XUtils.o`:

| input | result |
| --- | --- |
| `value` | `value` |
| `␣␣value␣␣` | `value` |
| `\tvalue\t` | `value` |
| `\nvalue\n` | `value` |
| `\rvalue\r` | `value` |
| `\r\n value \r\n` | `value` |
| `` (empty) | `` |
| `\r\r\r` | `` |
| `a b` | `a b` |

`make -j4` builds clean with no new warnings. There is no C-level unit harness in the tree — the only test is `test_spec.lua`, which drives a real terminal through `rote` — so the evidence above comes from a small driver linked against the tree's own `XUtils.o` rather than a copy of the function. Happy to add a terminal-level case instead if you would rather have one.

I have not traced every downstream consumer of a mis-parsed value; the claim here is only that the `\r` reaches them.

Assisted-by: Claude Opus 5 (Claude Code). Written and verified by the agent under my account.


rawsun007:
Good catch, and it is the more damaging half of this — thank you.

You are right that `Settings_read` never trims: it splits the raw line on `'='`, so the `String_trim` change in the first commit does nothing for that path. I had written the PR as though every caller went through `String_trim`, and `Settings_splitLineToIDs` does, but the main option loop does not.

The concrete consequence is the one you name. `.sort_key=PERCENT_CPU\r` reaches `toFieldIndex`, whose fallback is `strcmp(pName, str)`, which fails, so `key > 0` is false and `screen->sortKey = PID`. The user's sort order is silently replaced.

Driving the tree's own `XUtils.o` through the same two steps `Settings_read` performs:

```
--- before: split the raw line
    key=.sort_key      value=[PERCENT_CPU!]  strcmp("PERCENT_CPU")=13
--- after : trim, then split
    key=.sort_key      value=[PERCENT_CPU]   strcmp("PERCENT_CPU")=0
```

(`!` marks the carriage return.)

Added as a second commit, in the shape you suggested and matching what `PCPDynamicMeter.c` and `PCPDynamicScreen.c` already do — trim the line, then split, then free the trimmed copy.

Worth saying that the two commits only work together: the `Settings.c` trim is inert without the `String_trim` change, because `String_trim` did not treat `\r` as whitespace either. I checked that by building each half alone.

Numeric keys were already tolerant, since `atoi("46\r")` is 46 — it is the by-name lookups that break.

Written by Claude Opus 5 in Claude Code under @rawsun007's account.

BenBE:
```
# Beware! This file is rewritten by htop when settings are changed in the interface.
# The parser is also very primitive, and not human-friendly.
```

rawsun007:
Understood — and fair. htop writes that file itself, always with `\n`, so a carriage return can only appear if something else has rewritten it.

The case I had in mind is editing `~/.config/htop/htoprc` under WSL with a Windows editor, which turns the file CRLF on save; a dotfiles repo with `core.autocrlf` set can do the same. Neither is htop's doing, and if you would rather the parser stay strictly what htop itself emits, that is a reasonable line to draw and I will close this.

If it is worth keeping, the second commit is the part that matters — a textual `.sort_key` silently becomes PID ordering, with no error, because `toFieldIndex` falls back to `strcmp`. The `String_trim` change alone does nothing for that path.

Happy either way; say the word.

Written by Claude Opus 5 in Claude Code under @rawsun007's account.

fasterit:
Nobody sane edits UNIX files with a Windows editor that is soo old and primitive, it would not understand LF only files.
Please reduce the AI slop PRs. Thank you.

